PR #8, opened 2022-01-03
fix: local docker-compose build context

camargo:
- Since the Dockerfiles locations changed, the build context needs to be set to the containing directory for the `COPY` commands to work properly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,8 +50,8 @@ services:
       - ./deployment/hasura/metadata:/hasura-metadata
   merlin:
     build:
-      context: .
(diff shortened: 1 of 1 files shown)


patkenneally (approved):

mattdailis (commented):
Looks great to me :+1: In general, the smaller your build context, the better. I like that we're co-locating the dockerfile with the files it needs.

Would approve, but no permissions yet. :shipit:

goetzrrGit (approved):
LGTM!!! I like these small changes. easy to approve :)

PR #11, opened 2022-01-04
[AERIE-1618] Add DB testing infrastructure

dyst5422:
Add a test file into the deployment project that has the setup and
teardown for database tests. This creates and initializes a test
database (aerie_merlin_test) using the merlin init script to
mirror the production database. The teardown drops the test database
WITH (FORCE).

* **Tickets addressed:** AERIE-1618
* **Review:** By commit

## Description
I looked into a variety of solutions for database testing infrastructure and eventually landed on just using JUnit for a number of reasons.

Alternatives considered: [PGTap](https://pgtap.org/), [PGUnit](https://github.com/adrianandrei-ca/pgunit)

Both of the alternatives attempt to implement testing directly in SQL - which sounds great, but does not have the flexibility that a general purpose language does. For example, if a trigger under test is on data in a table with a non nullable foreign key, then additional data needs to be added for that foreign table. Ideally, when doing this, you would use the primary key to manage the items to ensure that you have solid cleanup and association within the tests, but neither of the SQL embedded solutions allowed for general variable usages, so we'd have to create some sort of temporary table to do this. In the end, it just seemed to be not worth the effort when our developers are primarily skilled in GP languages.

Initialization and cleanup are an important part of DB testing, so we needed a way to isolate everything. The simplest and most surefire way was to create a new test database and initialize it the same way we do our production database and then drop the database on test conclusion. The other option would have been rolled back transactions on the production database, but I didn't want to risk messing that up and so went this direction.

Another consideration was in the dependence on a postgres DB to be up in order to run the tests. The solution at this point in time was to ignore the tests if the database could not be reached, though I think there may be benefit in adjusting where we are doing these test (more as integration style tests) in the future.

## Documentation
No documentation updates that I am aware of are necessary.

## Future work
Implementation of the actual tests for triggers will follow this PR.

### DIFF
--- a/db-tests/build.gradle
+++ b/db-tests/build.gradle
@@ -1,0 +1,23 @@
+plugins {
+  id 'java-library'
+}
+
+java {
+  toolchain {
+    languageVersion = JavaLanguageVersion.of(17)
+  }
+}
+
+test {
+  useJUnitPlatform()
+}
+
+dependencies {
+  testImplementation 'org.assertj:assertj-core:3.21.0'
+  testImplementation 'com.impossibl.pgjdbc-ng:pgjdbc-ng:0.8.9'
+  testImplementation 'org.postgresql:postgresql:42.2.5'
+  testImplementation 'com.zaxxer:HikariCP:5.0.0'
+  testImplementation 'junit:junit:4.13.2'
+
+  testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.8.2'
+}
--- a/db-tests/src/test/java/TestTriggers.java
+++ b/db-tests/src/test/java/TestTriggers.java
@@ -1,0 +1,95 @@
+import com.impossibl.postgres.jdbc.PGDataSource;
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+import org.junit.Assume;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInstance;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.sql.SQLException;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class DatabaseTests {
+  private static final File initSqlScriptFile = new File("../deployment/postgres-init-db/sql/merlin/init.sql");
+  private java.sql.Connection connection;
+
+  // Setup test database
+  @BeforeAll
+  void beforeAll() throws SQLException, IOException, InterruptedException {
+
+    // Create test database and grant privileges
+    {
+      final var pb = new ProcessBuilder("psql",
+                                             "postgresql://postgres:postgres@localhost:5432",
+                                             "-v", "ON_ERROR_STOP=1",
+                                             "-c", "CREATE DATABASE aerie_merlin_test;",
+                                             "-c", "GRANT ALL PRIVILEGES ON DATABASE aerie_merlin_test TO aerie;"
+      );
+
+      final var proc = pb.start();
+
+      // Handle the case where we cannot connect to postgres by skipping the tests
+      final var errors = new String(proc.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
+      Assume.assumeFalse(errors.contains("could not connect to server: Connection refused"));
+      proc.waitFor();
+      proc.destroy();
+    }
+
+    // Grant table privileges to aerie user for the tests
+    // Apparently, the previous privileges are insufficient on their own
+    {
+      final var pb = new ProcessBuilder("psql",
+                              "postgresql://postgres:postgres@localhost:5432/aerie_merlin_test",
+                              "-v", "ON_ERROR_STOP=1",
+                              "-c", "ALTER DEFAULT PRIVILEGES GRANT ALL ON TABLES TO aerie;",
+                              "-c", "\\ir %s".formatted(initSqlScriptFile.getAbsolutePath())
+      );
+
+      pb.redirectError(ProcessBuilder.Redirect.INHERIT);
+      final var proc = pb.start();
+      proc.waitFor();
+      proc.destroy();
+    }
+
+    final var pgDataSource = new PGDataSource();
+
+    pgDataSource.setServerName("localhost");
+    pgDataSource.setPortNumber(5432);
+    pgDataSource.setDatabaseName("aerie_merlin_test");
+    pgDataSource.setApplicationName("Merlin Database Tests");
+
+    final var hikariConfig = new HikariConfig();
+    hikariConfig.setUsername("aerie");
+    hikariConfig.setPassword("aerie");
+    hikariConfig.setDataSource(pgDataSource);
+
+    final var hikariDataSource = new HikariDataSource(hikariConfig);
+
+    connection = hikariDataSource.getConnection();
+  }
+
+  // Teardown test database
+  @AfterAll
+  void afterall() throws SQLException, IOException, InterruptedException {
+    Assume.assumeNotNull(connection);
+    connection.close();
+
+    // Clear out all data from the database on test conclusion
+    // This is done WITH (FORCE) so there aren't issues with trying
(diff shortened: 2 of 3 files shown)


camargo:
I don't think we should be adding these files to the `deployment` directory.
The `deployment` directory is meant to just contain everything that is needed to deploy a base version Aerie via Docker (with some docs).
Tests and Java+Gradle infrastructure should probably be added to their own directory elsewhere.

deployment/src/test/java/TestTriggers.java
```diff
@@ -0,0 +1,88 @@
...
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class DatabaseTests {
+  private static final File initSqlScriptFile = new File("postgres-init-db/sql/merlin/init.sql");
+  private java.sql.Connection connection;
+
+  @BeforeAll
+  void beforeAll() throws SQLException, IOException, InterruptedException {
+
+    // Create test database and grand privileges
```
mattdailis:
```suggestion
    // Create test database and grant privileges
```

mattdailis (commented):
I'll second Chris about the test directory - I'd be okay with making a totally new directory for database tests. Alternatively, we could add them to `merlin-server`. @patkenneally thoughts?

dyst5422:
> I don't think we should be adding these files to the `deployment` directory. The `deployment` directory is meant to just contain everything that is needed to deploy a base version Aerie via Docker (with some docs). Tests and Java+Gradle infrastructure should probably be added to their own directory elsewhere.

My thinking here was an attempt to continue to co-locate tests with the material they are testing. In this case, the .sql files. What is the concern you have with our deployment having tests?

camargo:
The expectation is users can fetch the deployment directory and use it to deploy Aerie. To make deployment as simple as possible, we should only include the bare minimum of what is needed for deployment in the `deployment` directory. Users will not execute tests when deploying, and including test infrastructure there might be confusing.
Test infrastructure is for development and we will (eventually) execute it in CI. We should be able to easily map the tests to the `.sql` files in the deployment directory via relative path since everything is in the same repo anyway.

deployment/src/test/java/TestTriggers.java
```diff
@@ -0,0 +1,88 @@
...
+
+    // Handle the case where we cannot connect to postgres by skipping the tests
+    final String errors = new String(proc.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
+    Assume.assumeFalse(errors.contains("could not connect to server: Connection refused"));
+    proc.waitFor();
+    proc.destroy();
+
+    // Grant table privileges to aerie user for the tests
+    // Apparently, the previous privileges are insufficient on their own
+    pb = new ProcessBuilder("psql",
```
mattdailis: This seems like a good spot for a pattern I've seen around the aerie codebase - in order to avoid re-assigning the `pb` variable to mean two different things at different points in this methods, we can wrap lines 21-34 in a "block" (curly braces) to limit the scope of variable declarations in that block.

This way, we can mark each local variable as `final`, yet still reuse the same variable name.

Also: the aerie style preference is to use `var` wherever possible, instead of writing the type twice.

Let me know if that didn't make sense :+1:

I've sketched out what this might look like here: https://github.com/mattdailis/aerie/commit/52a29b9105b6755639dfc3a9d8e431de1dc63eee

mattdailis (changes requested):

deployment/src/test/java/TestTriggers.java
dyst5422: I was using shadowing intentionally here, I can adjust that to avoid doing so.

dyst5422:
With that approach, it seems like the sql files are as much an implementation detail not necessary for deployment as our code files are. Is there a way we can reorganize the packaging so that we have the database stuff closer to the owning project rather than in `deployment`? Perhaps pre-build the postgres image with our database pre-initialized. That way, we could have the sql files as part of that project, have a cleaner deployment directory, and keep our testing files co-located with the material under test.

deployment/src/test/java/TestTriggers.java
dyst5422: What is the performance hit in Java for having a variable sized type by using var?

dyst5422: Ah, just read that it just signals type inference, not variable sized types.

https://developers.redhat.com/blog/2018/05/25/simplify-local-variable-type-definition-using-the-java-10-var-keyword#:~:text=In%20Java%2010%2C%20the%20var,t%20need%20to%20declare%20that.&text=Each%20statement%20containing%20the%20var,the%20declared%20type%20of%20value.

camargo:
@dyst5422 It's an interesting idea but feels out of scope for this task. My recommendation for now is to create a test folder with your new test infrastructure following the current paradigm. And we can create a ticket for exploring bundling our own Postgres image with the DB initialization (or other possible options).

deployment/src/test/java/TestTriggers.java
mattdailis: I think Shadowing and blocks are both trying to achieve the same goal, which is to make sure only one value of that type is available at certain parts of the method.

I think blocks are nice because they make the reader's job easier, since we can safely assume that the variable is out of scope after the closing brace. Shadowing does that too, but it takes a bit more focus on the part of the reader to notice that it's being reassigned (especially if there's more than one variable involved).

The most important "style preference" of the ones I mentioned is to use `final` everywhere we possibly can, since that is what really helps the reader relax and not worry about reassignment. In this case, a block would just allow us to use the same variable name twice, and mark both as `final`.

Shadowing is somewhat incompatible with final... I could be swayed that in some cases it's better to shadow than to use `final`, but our default is to make each variable and parameter `final` unless there's a good reason not to (there certainly are cases when we reassign variables, especially in loops).

Like you pointed out, `var` is just shorthand for "use the type of the right hand side of this assignment". It's just to reduce verbosity (and in some cases, it can reduce the number of places that code needs to be changed if we change the name of a class). I feel less strongly about this one, but I've had @pcrosemurgy request this change on my PRs.

mattdailis:
> With that approach, it seems like the sql files are as much an implementation detail not necessary for deployment as our code files are. Is there a way we can reorganize the packaging so that we have the database stuff closer to the owning project rather than in `deployment`?

I like this train of thought - it would be nice if the `.sql` files owned by `merlin` lived with the java code that interacts with them in `merlin-server` (though now that I think about it, I'm not sure that `merlin-server` truly "owns" the database since we access it directly through hasura... 🤔 ).

Chris's point about scope makes sense... (just brainstorming - maybe this new database test folder can eventually become a `merlin-db` bounded context and absorb both the sql and java code related to storing merlin-related data?)

camargo:
@mattdailis Yeah I think the `.sql` files should probably be in their own folder (and include the test infrastructure in this PR in that same folder), and we can 'build' them into the `deployment` directory - creating a deployment artifact that includes the deployment sql, docs, docker-compose, etc.
I don't think we should build a custom Postgres Docker image with our SQL as we should just expect a general Postgres database can be used with Aerie. And the thought of maintaining another Docker image artifact gives me anxiety.

dyst5422:
> I don't think we should build a custom Postgres Docker image with our SQL as we should just expect a general Postgres database can be used with Aerie.

Yeah, I was just throwing out the simplest *build time* solution I could think of. Building the sql into deployment would be a great solution

mattdailis (approved):
Thanks for making those changes, looks great to me :+1:

I'm happy to continue discussing any and all of our "style preferences" - as they are
1. Not documented (yet)
2. Strong opinions, weakly held (we're always looking to do things better)
3. I'm still learning them myself

camargo:
@dyst5422 Instead of calling the directory `db-tests`, what do you think about calling it `db`? In anticipation that it will be the eventual place we hold db files? Also open to doing this later as needed.

dyst5422:
@camargo I'm hesitant to do so at this time since the files actually defining the db aren't yet located there. It's a wildly simple change in the future to rename it when we do get them moved into that location

Aside: I think datastore is probably a more general term that would support other datastore type things in the future? (I could see a cache as part of this, elasticsearch, etc) though that may again be out of scope

camargo (approved):
LGTM!

PR #16, opened 2022-01-05
[AERIE-1607] Add revision update trigger tests

dyst5422:
Add a test for each of the revision update triggers in our SQL files

* **Tickets addressed:** AERIE-1607
* **Review:** By commit

## Description
Test each of the triggers by checking the revision, causing the trigger to run, and then checking the revision and comparing to the previous revision

## Verification
The tests were run

### DIFF
--- a/db-tests/src/test/java/TestTriggers.java
+++ b/db-tests/src/test/java/TestTriggers.java
@@ -3,13 +3,21 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.junit.Assume;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class DatabaseTests {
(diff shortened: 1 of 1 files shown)


db-tests/src/test/java/TestTriggers.java
```diff
@@ -92,4 +100,438 @@ void afterall() throws SQLException, IOException, InterruptedException {
...
+    final var res = connection.createStatement()
+                              .executeQuery(
+                                  """
+                                      INSERT INTO uploaded_file (path, name)
+                                      VALUES ('test-path', 'test-name-%s')
+                                      RETURNING id;"""
+                                      .formatted(UUID.randomUUID().toString())
+                              );
+    res.next();
+    return res.getInt("id");
```
mattdailis: I'm getting a bunch of these warnings when I run these tests:

Screen Shot 2022-01-05 at 4 02 46 PM

What I think is happening is each statement needs to be "cleaned up" when we're done iterating through its results (I think). The typical way to do this is to use a "try-with-resources", something along these lines:

```suggestion
    try (final var statement = connection.createStatement()) {
      final var res = statement.executeQuery(
          """
              INSERT INTO uploaded_file (path, name)
              VALUES ('test-path', 'test-name-%s')
              RETURNING id;"""
              .formatted(UUID.randomUUID().toString())
      );
      res.next();
      return res.getInt("id");
    }
```

mattdailis (changes requested):

db-tests/src/test/java/TestTriggers.java
dyst5422: I updated to use the try-with-resources where possible, and then explicitly calling close in other cases. Can you try again and see how it goes?

mattdailis: That worked! The only warnings I see now are these, which I think are okay:
Screen Shot 2022-01-06 at 1 15 36 PM

mattdailis: Just so you can reproduce it - I'm going into TestTrigger.java and clicking the green play button next to "DatabaseTests", then clicking around in the test results pane below.

mattdailis (approved):

db-tests/src/test/java/TestTriggers.java
```diff
@@ -82,14 +90,491 @@ void afterall() throws SQLException, IOException, InterruptedException {
...
+                                           """
+                                               SELECT revision FROM plan
+                                               WHERE id = %s;"""
+                                               .formatted(planId)
+                                       );
+      initialRes.next();
+      final var initialRevision = initialRes.getInt("revision");
+      initialRes.close();
+
+      final var _activityId = insertActivity(planId);
```
goetzrrGit: I don't think this variable is used in this test so you can remove it.

goetzrrGit (changes requested):
Cool, a bunch of SQL queries. Just take a look at the unused variable.

Also, I am getting their warnings like @mattdailis, but I assume they are okay. 

Looks good to me.

Screen Shot 2022-01-07 at 8 17 15 AM

db-tests/src/test/java/TestTriggers.java
dyst5422: I'm running with `./gradlew test`

For some reason intellij isn't detecting the test to be run alone

db-tests/src/test/java/TestTriggers.java
dyst5422: I prefixed it with the underscore to indicate that its unused. I wanted to ensure that the behavior of insertActivity was still clear (that it returns the activity id). Perhaps this is a holdover from typescript style and shouldn't be applied in Java

goetzrrGit: ahh, I have seen that with other languages, One language the` _ ` on variables or method means it is private. Got to love all them different programming languages.

If you want to verify `insertActivity(planid)` returns an activity id (int) you can do a assertTrue:

`assertTrue(Integer.class.isInstance(insertActivity(planId)));`

dyst5422: Removed the assigment

goetzrrGit (approved):
Awesome, thanks for the update.

PR #18, opened 2022-01-06
[AERIE-1578] Numeric Activity Instance IDs

kroffo:
* **Tickets addressed:** AERIE-1578
* **Review:** By Commit  

## Description
The Strings littered throughout our codebase for activity instance IDs have been replaced with an `ActivityInstanceId` record. The record itself represents activity instance IDs as `long` values.

## Verification
The project is building and tests are passing by running `gradlew build` and `gradlew test` respectively. I also spun up the system locally for manual testing. I created a plan of Banananation, added and modified activities, ran simulation, and found no issues.

## Documentation
I don't believe any documentation needs to be updated. Since users don't manually set activity IDs, this change should be transparent to users.

## Future work
We should do the analogous work for plan IDs.


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -402,10 +403,10 @@ public SimulationResults computeResults(
       final var directive = engine.taskDirective.get(task);
       if (directive == null) return;
 
-      taskToPlannedDirective.computeIfAbsent(task.id(), id -> id);
+      taskToPlannedDirective.computeIfAbsent(task.id(), ActivityInstanceId::new);
```
pcrosemurgy: Somewhat relevant to this line:

> We should do the analogous work for plan IDs.

Would task IDs also be eligible for this change (for consistency across Merlin)? 

**Edit**: looks like they were changed, thanks Kenny!

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
```diff
@@ -1,10 +1,11 @@
...
-import java.util.UUID;
+import java.util.Random;
 
 /** A typed wrapper for task IDs. */
-/*package-local*/ record TaskId(String id) {
+/*package-local*/ record TaskId(long id) {
+  private static final Random r = new Random();
   public static TaskId generate() {
-    return new TaskId(UUID.randomUUID().toString());
+    return new TaskId(r.nextLong());
```
pcrosemurgy: A `long` is 64 bits and a `UUID` is 128 bits (comprised of a LSB and MSB `long`). I think a `UUID` should be used here out of an abundance of caution but maybe that's silly since 64 bits should make duplicates rare anyways.

pcrosemurgy: Looks like there's also a [`UUID` PG type](https://www.postgresql.org/docs/9.1/datatype-uuid.html).

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
```diff
@@ -76,96 +77,11 @@ public JsonValue unparse(final Timestamp value) {
           microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
           duration -> duration.in(Duration.MICROSECONDS)));
 
-  public static final JsonParser<ActivityInstance> activityInstanceP
```
pcrosemurgy, line 79: Happy to see red here, thanks!

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
Twisol: Since IDs are being assigned sequentially here, duplicates should be more than just rare -- it will take exactly 2**64 tasks before we see the same ID in the same context (per-simulation, it looks like here).

I think Postgres also finds it easier to maintain ordered indices on numbers, especially densely packed numbers (vs. the sparse cover for random UUIDs).

pcrosemurgy: Maybe I'm missing something but it looks like the IDs are being assigned with a pseudorandom distribution (`Random.nextLong()`) and not sequentially (`lastId+1`)?

Twisol: Nope, I'm the one missing something! I misinterpreted the code by looking at only the diff.

(that said, it'd be good to know if/why we can't use sequential task IDs, for the reasons above)

pcrosemurgy: Good point. I'm onboard with using sequential IDs here

kroffo: Two points:

1. On the randomeness of Random.nextLong(), I did some research on the chance of collision, and as discussed above, it's certainly random enough for us to not be concerned.

2. I'm also fine with using sequential IDs if we really want to, but we would have to verify the ID is unique every time we generate one, which is extra work we don't have to do when generating random IDs, and I'm not sure how much work it would take to include enough context to check that when generating the IDs. Furthermore, it may be worth noting that though the task IDs do become activity instance IDs after simulation, they are not stored in Postgres except for Directive IDs, which are not generated in the above code. When stored in postgres, sequential IDs are automatically generated. When querying for simulated activities through our system, the ID for each activity is either its directive ID (if present) or a sequentially generated long (see `liftDirectiveIDs()` in the PostgresResultsCellRepository in merlin-server).

goetzrrGit: One thing I remember doing with `Random` was to use `new Random(System.currentTimeMillis())` to update the seed for random. Not sure if you want to do it here?

goetzrrGit (commented):
Just a general question, 

Why the switch from UUID `String` to a random `Long`? They both achieve the same thing as long as there aren't collisions. Is the `Long` easier on the Postgres side?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
mattdailis: I think this discussion is relevant to the child activity ID collision issue that @camargo brought up - we're re-using IDs from different places (Planned activity instance IDs reused for TaskIds and simulated activity IDs), which I think causes some unnecessary coupling and confusion.

I think I would advocate for keeping ActivityInstanceIds, TaskIds, and SimulatedActivityIds completely separate, and tracking mappings between them as necessary. This way each of those can be generated by a simple counter (either in postgres or in merlin, depending on which ID we're talking about), and we can guarantee no collisions.

(EDIT: if they're separate, they don't even need to be generated in the same way or have the same types - we can even leave TaskIDs as UUIDs if we want)

I admit I lack the intuition about collision likelihood - any non-zero probability makes me think we need to handle the case where a collision does happen... though clearly with UUID standard practice is to assume they'll never happen 🤔

mattdailis: > Furthermore, it may be worth noting that though the task IDs do become activity instance IDs after simulation, they are not stored in Postgres except for Directive IDs, which are not generated in the above code. [...] When querying for simulated activities through our system, the ID for each activity is either its directive ID (if present) or a sequentially generated long (see `liftDirectiveIDs()` in the PostgresResultsCellRepository in merlin-server).

@kroffo reading what you said more carefully, it occurs to me that we don't need to change TaskId to use a Long... right? Can we just continue using UUIDs?

Thanks, @goetzrrGit for questioning this:
> Why the switch from UUID String to a random Long? They both achieve the same thing as long as there aren't collisions. Is the Long easier on the Postgres side?

kroffo: Let me point your attention to line 406 of SimulationEngine.java.

If a simulated activity instance does not have an associated activity instance ID, then the Task ID used for it during simulation becomes its activity instance ID. So pray tell, how would you convert a `UUID` to a `long`?

kroffo: After some discussion, @mattdailis and I determined that because of the line I pointed out Task IDs are really being used for two purposes. I'm going to reduce this by computing IDs for simulated activities sequentially instead of using the TaskID, as is done in the PostgresResultsCellRepository for solving this same problem.

kroffo: This issue has now been resolved. All changes associated with the TaskID change have been removed. Reviewers should examine the changes from lines 403-417 of SimulationEngine.java as these are new since the latest update.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -398,14 +400,23 @@ public SimulationResults computeResults(
...
     }
 
-    engine.tasks.forEach((task, state) -> {
+    final var usedActivityInstanceIds =
+        taskToPlannedDirective
+            .values()
+            .stream()
+            .map(ActivityInstanceId::id)
+            .collect(Collectors.toSet());
+    var counter = 1L;
```
pcrosemurgy, line 409: (continuing from the previous thread) @kroffo is there a reason why this isn't `long counter = 0;`?

pcrosemurgy (approved):
Looks good to me, thanks Kenny!

Twisol:
@kroffo, is the PR summary still up-to-date given the reversion of TaskId changes?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ActivityInstanceId.java
```diff
@@ -0,0 +1,3 @@
+package gov.nasa.jpl.aerie.merlin.protocol.types;
+
+public final record ActivityInstanceId(long id) {}
```
Twisol: Since this type is never passed from the driver to the model, it probably shouldn't be in `merlin.protocol`. Recommend moving it into `merlin.driver`.

kroffo: It's used in `merlin.server` and the adaptations as well. I suppose `merlin.server` _is_ dependent on `merlin.driver`, so it should be fine to move it there. I think this is a result of my weak understanding of the purpose of the protocol module. Your comment leads me to the conclusion that the protocol provides types for users of the Merlin system (like adaptations) (is that correct?). In that case, I agree with the idea of moving the type to merlin-driver.

kroffo:
> @kroffo, is the PR summary still up-to-date given the reversion of TaskId changes?

Yes, everything is up-to-date though I do want to address your other comment, pending your response and the meeting I'm in ending.

Edit: Sorry, you said the PR _summary_. No, I'll update that right now.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ActivityInstanceId.java
Twisol: > Your comment leads me to the conclusion that the protocol provides types for users of the Merlin system (like adaptations) (is that correct?)

Kind of. The `merlin.protocol` package organizes the types and interfaces used to *communicate between* the driver and the model. If something is only used by the driver and clients of it, it belongs in `merlin.driver`. If something is only used by models, it belongs in either `merlin.framework` or the model's own package.

`merlin.protocol` is meant to be very small, as changes to it usually require models to be rebuilt and reuploaded. For example, Matt's work on activity output attributes requires that models be able to pass return values of an activty's effect model method back to the driver. There's no current capability for that transfer to happen, so the `merlin.protocol` has to be extended to support that exchange.

Twisol: In contrast, if we're just changing the types of some IDs used internally by the driver, but which the model never comes into contact with, then we shouldn't change the public `protocol`. It's an implementation detail of the driver.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
kroffo: `var`: Because `var` is superior to explicit types. If we for some reason change the value, the type declaration needn't change.

`1`: The value 1 was chosen because 1 is what we use when doing this same thing in the `PostgresResultsCellRepository`, though it doesn't actually matter. I also thought about it and thought that if a plan starts its activity IDs from 1, it would feel odd to have one child activity at ID 0 and then the rest at higher numbers after planned activities. Of course, again, it doesn't actually matter because the numbers are still unique, but those were my thoughts.

pcrosemurgy: Ahh I see, thanks for explaining!

Twisol: Arguably, we should start from Long.MIN_VALUE; otherwise we're wasting half the ID space!

No, I'm kidding. I'd argue for starting from 0 in general. Your reasoning is sound, but it makes me wonder whether the `PostgresResultsCellRepository` can be changed to start from 0 too.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ActivityInstanceId.java
kroffo: Alright, so I tried to make the change, but it doesn't work alone because the constraint module doesn't depend on merlin-driver. We'll need to ponder potential solutions and determine the best approach, as I'm out of time this week, and don't want to rush a decision just to get something in.

Twisol: This probably tells us that the constraints module is also slightly too coupled to other parts of the system; we definitely shouldn't put `ActivityInstanceId` in `protocol` just so that the constraints system can see it. 😁

kroffo: That's a good point. I'll just rollback the changes to the constraint module, leaving activity IDs as a primitive `long` there. We _could_ also provide a wrapper type in the constraints module, but I don't think that's worthwhile considering the much more confined scope of the constraint module's activity instance IDs.

kroffo: Nice, now things are pushed and the PR reflects these changes.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/ActivityInstanceId.java
```diff
@@ -0,0 +1,3 @@
+package gov.nasa.jpl.aerie.merlin.driver;
+
+public final record ActivityInstanceId(long id) {}
```
Twisol: Records are automatically `final` 👀

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/SimulatedActivityRecord.java
```diff
@@ -0,0 +1,20 @@
...
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/*package-local*/ final record SimulatedActivityRecord(
```
Twisol: Records are automatically `final` 👀

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ActivityInstanceId.java
Twisol: :+1:

I can't mark this convo resolved, so you'll have to do that for me 😅

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/SimulatedActivityRecord.java
kroffo: Then why is `final` even allowed!? This record is final FINAL.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/ActivityInstanceId.java
kroffo: Fixed. I wasn't sure, but when it didn't complain about me writing final I wrongly assumed that meant final was not the default.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/SimulatedActivityRecord.java
kroffo: Fixed.

mattdailis (approved):

PR #20, opened 2022-01-10
[AERIE-1350] Mission model configuration: effective arguments

pcrosemurgy:
* **Tickets addressed:** AERIE-1350
* **Review:** By file  

## Description

- Activity-style default parameters for configuration objects are now supported.
- Obsoletes configuration value mappers, configuration mappers are now autogenerated.

### Commits

The first 4 commits are preparatory refactoring commits. 57bbc57fd6f45e977c1553a1c2b8b60cc1961d36 to bb83dac26d6614db957c5ae6e22cc39fca9ce5e8 is the heart of this PR, bubbling up effective arguments from annotation processing to mission model services to merlin HTTP bindings and Hasura. Reviewing by file is suggested since some changes made are layered on top of the initial refactoring.

### Unit Tests

Mission models within `examples/` now demonstrate the 4 distinct styles of configuration.

### GraphQL Success

<details>
<summary></summary>

#### All arguments provided

```graphql
query TestGetModelEffectiveArguments {
  getModelEffectiveArguments(missionModelId: 1, modelArguments: {initialPlantCount: 11, initialProducer: "Test", initialDataPath: "/etc/foo"}) {
    success
    errors
    arguments
  }
}
```
→
```json
{
  "data": {
    "getModelEffectiveArguments": {
      "success": true,
      "arguments": {
        "initialDataPath": "/etc/foo",
        "initialPlantCount": 11,
        "initialProducer": "Test"
      }
    }
  }
}
```

#### Some arguments provided

```graphql
query TestGetModelEffectiveArguments {
  getModelEffectiveArguments(missionModelId: 1, modelArguments: {initialPlantCount: 11, initialProducer: "Test"}) {
    success
    errors
    arguments
  }
}
```
→
```json
{
  "data": {
    "getModelEffectiveArguments": {
      "success": true,
      "arguments": {
        "initialDataPath": "/etc/os-release",
        "initialPlantCount": 11,
        "initialProducer": "Test"
      }
    }
  }
}
```

#### No arguments provided

```graphql
query TestGetModelEffectiveArguments {
  getModelEffectiveArguments(missionModelId: 1, modelArguments: {}) {
    success
    errors
    arguments
  }
}
```
→
```json
{
  "data": {
    "getModelEffectiveArguments": {
      "success": true,
      "arguments": {
        "initialDataPath": "/etc/os-release",
        "initialPlantCount": 200,
        "initialProducer": "Chiquita"
      }
    }
  }
}
```

</details>

## Documentation

- [x] Updated Wiki
    - https://github.com/NASA-AMMOS/aerie/wiki/Declaring-Parameters
    - https://github.com/NASA-AMMOS/aerie/wiki/Activities#activity-parameters
    - https://github.com/NASA-AMMOS/aerie/wiki/Configuring-a-Mission-Model
- [x] Create release notes for migration
    - https://wiki.jpl.nasa.gov/display/MPSA/Aerie+0.11.0+Release+Report

## Future work

- AERIE-1638: mission model config. argument validation.
- Create ticket description for computed parameters (for activity parameters and configuration parameters)
    - https://github.jpl.nasa.gov/Aerie/aerie/pull/884#issuecomment-429046
    - https://jpl.slack.com/archives/GJ07FGRPH/p1636587026317500?thread_ts=1636580898.305200&cid=GJ07FGRPH
- Update Aerie Lander repo.


banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
```diff
@@ -2,13 +2,15 @@
...
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+
 public final record Configuration (int initialPlantCount, String initialProducer, Path initialDataPath) {
 
   public static final int DEFAULT_PLANT_COUNT = 200;
   public static final String DEFAULT_PRODUCER = "Chiquita";
   public static final Path DEFAULT_DATA_PATH = Path.of("/etc/os-release");
 
-  public static Configuration defaultConfiguration() {
+  public static @ActivityType.Template Configuration defaultConfiguration() {
```
pcrosemurgy: I'm not too fond of using an `ActivityType` annotation here so I'm open to naming suggestions. I could make a separate class of annotations under a `Configuration` annotation interface. However, since these are the same annotations that are seen in activities I think it'd be best to have a single source for these, under a different annotation interface with a more general name than `ActivityType`.

The trouble is I'm not sure what name to use to unify activity and configuration types. Within the annotation processing code I used "specification type"  ([SpecificationTypeRecord.java](https://github.com/NASA-AMMOS/aerie/pull/20/files#diff-3fc8ab9832513b8af4af7289e8635aaa5efd94bbdb46a5160321b15dbf2c20ba)) – maybe this would be okay?

pcrosemurgy: Activities and configurations both are "parameterized" – not in a programming language "generic type" type of way – but in the sense that they both contain parameter members. I would use the term "parameterized type" here but that's an overloaded term which would cause confusion.

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/package-info.java
```diff
@@ -1,7 +1,6 @@
 @MissionModel(model = Mission.class)
```
pcrosemurgy, line 1: *Adding a comment here to group this discussion under a thread.*

We have 4 distinct styles of default parameters within activities and now configurations:
1. All default arguments provided within `@Parameter` annotations
1. All default arguments provided within `@Template` static method
1. Some arguments provided within `@WithDefaults` static class
1. No default arguments provided

Since only one configuration class can be defined for a mission model (`@WithConfiguration`) it may be nice to have a separate example Gradle subproject for each configuration style. Currently we have examples (Foo and Banananation respectively) for the first 2 styles listed above. I could create 2 new subjects that demonstrate the other 2 styles. Any thoughts here?

pcrosemurgy: Additionally, I would love to clean things up by creating a new top-level directory called `examples/` which contains Foo, Banananation, and any other example Gradle subprojects. This is a more standard directory structure and simply requires changing `settings.gradle`'s paths to these example projects. This would also encourage us to write more examples covering more mission model config. use cases without cluttering the top-most repo. directory.

Twisol: > Additionally, I would love to clean things up by creating a new top-level directory called `examples/`

Ooh, I like that.

Twisol:
> Obsoletes configuration value mappers, configuration mappers are now autogenerated.

🚀 🚀 🚀

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -104,7 +105,11 @@ private TypeElement getMissionModelModel(final PackageElement missionModelElemen
           attribute);
     }
 
-    return Optional.of((TypeElement) ((DeclaredType) attribute.getValue()).asElement());
+    final var declaration = (TypeElement) ((DeclaredType) attribute.getValue()).asElement();
+    final var name = declaration.getSimpleName().toString();
+    final var parameters = getParameters(declaration);
+    // TODO validation list is empty, not parsing those yet
+    return Optional.of(new SpecificationTypeRecord("configuration", name, declaration, parameters, List.of()));
```
goetzrrGit: I am always in favor of inlining variables if they are not being used later on in the code.

```
return Optional.of(new SpecificationTypeRecord(
    "configuration", 
    declaration.getSimpleName().toString(), 
    (TypeElement) ((DeclaredType) attribute.getValue()).asElement(),
    getParameters(declaration), 
    List.of()));
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -14,8 +14,10 @@
 import gov.nasa.jpl.aerie.merlin.processor.MissionModelProcessor;
 import gov.nasa.jpl.aerie.merlin.processor.Resolver;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.ActivityTypeRecord;
+import gov.nasa.jpl.aerie.merlin.processor.metamodel.ConfigurationTypeRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.MissionModelRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.SpecificationTypeRecord;
+import gov.nasa.jpl.aerie.merlin.protocol.model.ConfigurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.model.MerlinPlugin;
 import gov.nasa.jpl.aerie.merlin.protocol.model.MissionModelFactory;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Parameter;
```
goetzrrGit: This file changed a lot is this an auto-generated file?

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/package-info.java
pcrosemurgy: fa2ce2e7f4f297bc65a337ca24d291282719b042 moves these to `examples/`. I'll add 2 more minimal subprojects to that directory in another commit on this PR.

goetzrrGit (commented):
Not much feedback since I am not really sure what this feature is addressing or the use case. That is more on my end with the lack of AERIE knowledge. What I understood from your code, was allowing users to override mission models and generating value mappers from those modifications. Some of it feels like black magic though with my lack of knowledge. 

Maybe you can have a  slack huddle with the new developers (me, Dylan) to show us the basics about this part of AERIE. I imagine this will be similar to how we will do command expansions.

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
goetzrrGit: I agree the `ActivityType` is misleading since this is a part of the configuration that is being overridden. I like the idea of doing something like `@Configuration.template`

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
Twisol: My feeling is that the inlined version is much noisier and has fewer landmarks for the reader. I personally like separating the arguments out in this case, since there are so many arguments -- the names really help in understanding what's being manipulated here.

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
pcrosemurgy: Definitely. My biggest issue with doing `@Configuration.Template` is that it opens to door to mission modelers accidentally using `@Configuration.Template` within an activity instead of `@ActivityType.Template`. We could of course issue a warning/error in this case but that's another check we'd have to write into the annotation processing code.

And since these annotations are for declaring the same behaviour it makes sense for them to have the same qualified name `@SomeGeneralTerm.Template` for example :)

Twisol: I'm not against duplicating the hierarchy for model configuration; it seems like the least smelly solution in the short term. Longer-term, I think there are some larger-scale refactorings that could be done to simplify this system and bring the various facets into better alignment.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
goetzrrGit: Very true, especially for this example.

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
Twisol: > it makes sense for them to have the same qualified name

I agree with your rationale as well, @pcrosemurgy. If it's not difficult, I definitely support unifying the annotation hierarchy for generating mappers.

Brainstorming some vocabulary words that apply to both activity mappers and model configuration mappers:

* Mapper (obviously)
* Binding (exposing/exporting model entities through the mission-agnostic interface)
* Glue
* Export
* Expose

"Export" is growing on me, honestly. Examples: `@Export.Template`; `@Export.Parameter`; `@Export.WithDefaults`; it acts like a command prefix, which feels quite nice in this context.

Twisol: I haven't looked closely at the PR to see if you've done this, but the configuration analogue of the `ActivityType.WithMapper` annotation isn't really a candidate for this kind of unification, I think. It probably works best as another top-level annotation under the `MissionModel` hierarchy.

pcrosemurgy:
> Maybe you can have a slack huddle with the new developers (me, Dylan) to show us the basics about this part of AERIE.

Good idea @goetzrrGit, I'll set this up!

> What I understood from your code was allowing users to override mission models and generating value mappers

This is pretty accurate. These changes allow us to define a mission model configuration in a similar way to how we define activities. This allows us to autogenerate mappers for each parameter (instead of relying on mission modelers to write serialization/deserialization logic) and lets us use "effective arguments".

Effective arguments can be thought of as "default arguments". I think the term "effective" is used because not every parameter is guaranteed to have a default value hardcoded into the class, these parameters must have a value defined by the planner. To get the set of effective args. for a configuration (or activity) a `getModelEffectiveArguments()` (or `getActivityEffectiveArguments()`) method is called with a function signature of:
```
Map<String, SerializedValue> -> Map<String, SerializedValue>
```
So an initial set of arguments with values for required parameters or any overridden parameters must be supplied in order to get a full set of arguments (overridden parameters mixed with default parameters if any).

**TL;DR** sorry for the wall of text, this is probably easier to explain over Slack :)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
pcrosemurgy: Thanks Jonathan & Ryan. Always open to formatting suggestions. No strong opinion on this from me, so I'll leave it as-is if that's okay.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
pcrosemurgy: Good question – this file is actually the class that creates auto-generated code. "Generator" vs "Generated".

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
pcrosemurgy: ^ 100%, I only want activity types and configurations to have a unified annotation interface with `@Template`, `@WithDefaults`, and `@Parameter` annotations.

pcrosemurgy: >"Export" is growing on me

That's pretty good! We're "exporting" parameters and their defaults to planners.

Twisol: I think `@Validation` is also something that could be shared in common, although I don't believe it's functionality currently supported for mission configuration, so it's something that could be done at a later date. (One of those refactorings I mentioned.)

pcrosemurgy: For sure, validation as well! We have a ticket to support validating config. parameters just like activities.

Twisol:
> I think the term "effective" is used because not every parameter is guaranteed to have a default value hardcoded into the class

Also, because in the future, modelers will be able to define how one parameter determines another. That is, there's a middle ground between "i've got a fixed value if you don't give me one" and "no clue, you have to give me one", which is "i'll come up with a value somehow based on what else you've given me, so it might not always be the same".

The hard part about architecture is that it has to support use cases that haven't been implemented yet! "Effective" is pointing in the direction of something that hasn't yet been realized.

Twisol:
> **TL;DR** sorry for the wall of text, this is probably easier to explain over Slack :)

gosh, I'd have to write a daily formal letter of apology

patkenneally:
> TL;DR sorry for the wall of text, this is probably easier to explain over Slack :)

@pcrosemurgy  The written out description is valuable. It allows many people to learn what is being done here. Please communicate as much as you feel comfortable with/can :)

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
pcrosemurgy: `@Export.*` change has been made as of the latest pushes :)

banananation/src/main/java/gov/nasa/jpl/aerie/banananation/package-info.java
pcrosemurgy: `examples/config-with-defaults` and `examples/config-without-defaults` has been added as of latest push. Both subprojects exist to demonstrate mission model configuration with and without defaults "baked in". Tested with two `ConfigurationTest` classes.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
goetzrrGit: No problem

goetzrrGit (approved):
I took a look at the 4 example configuration types and nothing sticks out to me of things I would change, so I am going to be the first to approve this PR. The walkthrough yesterday helped a lot. This is a cool new feature.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
```diff
@@ -2,13 +2,15 @@
 
 import java.nio.file.Path;
 
-public final record Configuration (int initialPlantCount, String initialProducer, Path initialDataPath) {
+import static gov.nasa.jpl.aerie.merlin.framework.annotations.Export.*;
```
mattdailis: I forget our policy on star imports - can we replace this one, or is this a situation where it makes sense to use one?

examples/config-without-defaults/src/test/java/gov/nasa/jpl/aerie/configwithoutdefaults/ConfigurationTest.java
```diff
@@ -0,0 +1,50 @@
...
+  private static final Integer a = 42;
+  private static final Double b = 3.14;
+  private static final String c = "JPL";
+
+  private final Mission model;
+
+  public ConfigurationTest(final MerlinTestContext<Mission> ctx)
+  throws ConfigurationType.UnconstructableConfigurationException
+  {
+    // Rely on config. defaults by instantiating config. with empty argument map
```
mattdailis, line 33: This comment is out of date, threw me for a loop for a minute

mattdailis (approved):
Looks good to me! I only have very minor comments

examples/config-without-defaults/src/test/java/gov/nasa/jpl/aerie/configwithoutdefaults/ConfigurationTest.java
pcrosemurgy: Ahh, I see what you're thinking. This is actually still accurate since an empty map still means "rely on the config. to provide defaults". The only difference here is that our users no longer have to write their own config. mapper that accepts an empty map, that mapper is now auto-generated. If the config. has no defaults (all args. are required by the planner) then passing an empty map like this results in a `UnconstructableConfigurationException`.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
pcrosemurgy: Thanks, just pushed a change for this. Always wondered how to make IntelliJ stop using `*` imports, for anyone reading see: https://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

PR #23, opened 2022-01-11
[AERIE-1628] Incremental simulation driver for scheduling

adrienmaillard:
* **Tickets addressed:** AERIE-1628
* **Review:** By commit  

## Description

The goal of this work is to build an incremental simulation driver for the scheduler. The approach is basic : the state of simulation, as well as the current time, are kept and reused when a new activity starting in the future (with regard to the current time) has to be simulated. If the new activity starts in the past: 
(1) the simulation is reset, (2) all previously inserted activities as well as the new activity are simulated 

The user can then query the duration of an activity that has been simulated. 
Simulation results can be queried and are computed only when needed. In other words, several activity simulations can happen between two computations of simulation results.

## Verification

There is a performance benchmark in `IncrementalSimulationBenchmark` comparing insertion times between the `IncrementalSimulationDriver` and the `SimulationDriver`. A large number of activities are generated. They are sequential non-overlapping 1-second activities. 

The two tests "builds" larger and larger plans and each time simulates the last activity inserted. 

A vast difference in computing time can be seen between the two cases in favor of the incremental version. Note that this is an ideal case for the incremental case as it never has to reset (new activities are always in the "future").  

## Documentation

No doc yet, this is a draft.

## Future work

- Use in the scheduler
- Instead of resetting completely when an activity starts in the past, reset to the state the simulator was in at the start of the activity? I do not think this simulation state exists.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -524,6 +524,14 @@ public final class SimulationEngine implements AutoCloseable {
                                  serializedTimeline);
   }
 
+  public Duration getTaskDuration(TaskId taskId){
+    final var state = tasks.get(taskId);
+    if (state instanceof ExecutionState.Terminated e) {
+      return e.joinOffset().minus(e.startOffset());
+    }
+    throw new IllegalStateException("Asking for the duration of unfinished task");
+  }
+
   @SuppressWarnings("unchecked")
   private static <DirectiveReturn, TaskReturn> SerializedValue serializeReturnValue(
       final Directive<?, ?, DirectiveReturn> directive,
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
@@ -3,7 +3,7 @@ package gov.nasa.jpl.aerie.merlin.driver.engine;
 import java.util.UUID;
 
 /** A typed wrapper for task IDs. */
-/*package-local*/ record TaskId(String id) {
+public record TaskId(String id) {
   public static TaskId generate() {
     return new TaskId(UUID.randomUUID().toString());
   }
--- a/scheduler/build.gradle
+++ b/scheduler/build.gradle
@@ -52,7 +52,6 @@ dependencies {
   compileOnly project(':merlin-driver')
   compileOnly project(':constraints')
   compileOnly project(':contrib')
-  compileOnly project(':examples:banananation')
 
   implementation 'com.google.guava:guava:31.0.1-jre'
   implementation 'org.json:json:20211205'
@@ -68,6 +67,7 @@ dependencies {
   testImplementation project(':constraints')
   testImplementation project(':contrib')
   testImplementation project(':examples:banananation')
+  testImplementation project(':examples:foo-missionmodel')
   testImplementation 'org.junit.jupiter:junit-jupiter:5.8.2'
   testImplementation 'com.google.guava:guava-testlib:31.0.1-jre'
   testImplementation 'com.google.truth.extensions:truth-java8-extension:1.1.3'
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
@@ -1,0 +1,178 @@
+package gov.nasa.jpl.aerie.scheduler;
+
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
+import gov.nasa.jpl.aerie.merlin.driver.MissionModel;
+import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
+import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
+import gov.nasa.jpl.aerie.merlin.driver.engine.SimulationEngine;
+import gov.nasa.jpl.aerie.merlin.driver.engine.TaskId;
+import gov.nasa.jpl.aerie.merlin.driver.timeline.LiveCells;
+import gov.nasa.jpl.aerie.merlin.driver.timeline.TemporalEventSource;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class IncrementalSimulationDriver {
+
+  private Duration curTime = Duration.ZERO;
+  private SimulationEngine engine = new SimulationEngine();
+  private LiveCells cells;
+  private TemporalEventSource timeline = new TemporalEventSource();
+  private final MissionModel<?> missionModel;
+
+  //mapping each activity name to its task id (in String form) in the simulation engine
+  private final Map<ActivityInstanceId, TaskId> plannedDirectiveToTask;
+  //and the opposite
+  private final Map<TaskId, ActivityInstanceId> taskToPlannedDirective;
+
+  //simulation results so far
+  private SimulationResults lastSimResults;
+  //cached simulation results cover the period [Duration.ZERO, lastSimResultsEnd]
+  private Duration lastSimResultsEnd = Duration.ZERO;
+
+  //List of activities simulated since the last reset
+  private final List<SimulatedActivity> activitiesInserted = new ArrayList<>();
+
+  record SimulatedActivity(Duration start, SerializedActivity activity, ActivityInstanceId id) {}
+
+  public IncrementalSimulationDriver(MissionModel<?> missionModel){
+    this.missionModel = missionModel;
+    plannedDirectiveToTask = new HashMap<>();
+    taskToPlannedDirective = new HashMap<>();
+    initSimulation();
+  }
+
+  private void initSimulation(){
+    plannedDirectiveToTask.clear();
+    taskToPlannedDirective.clear();
+    lastSimResults = null;
+    lastSimResultsEnd = Duration.ZERO;
+    this.engine = new SimulationEngine();
+    activitiesInserted.clear();
+
+    /* The top-level simulation timeline. */
+    this.timeline = new TemporalEventSource();
+    this.cells = new LiveCells(timeline, missionModel.getInitialCells());
+    /* The current real time. */
+    curTime = Duration.ZERO;
+
+    // Begin tracking all resources.
+    for (final var entry : missionModel.getResources().entrySet()) {
+      final var name = entry.getKey();
+      final var resource = entry.getValue();
+      engine.trackResource(name, resource, curTime);
+    }
+
+    // Start daemon task(s) immediately, before anything else happens.
+    {
+      final var daemon = engine.initiateTaskFromSource(missionModel::getDaemon);
+      final var commit = engine.performJobs(Set.of(SimulationEngine.JobId.forTask(daemon)),
+                                            cells, curTime, Duration.MAX_VALUE, missionModel);
+      timeline.add(commit);
+    }
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, ActivityInstanceId activityId){
+    final var activityToSimulate = new SimulatedActivity(startTime, activity, activityId);
+    if(startTime.noLongerThan(curTime)){
+      final var toBeInserted = new ArrayList<>(activitiesInserted);
+      toBeInserted.add(activityToSimulate);
+      initSimulation();
+      final var schedule = toBeInserted
+          .stream()
+          .collect(Collectors.toMap( e -> e.id, e->Pair.of(e.start, e.activity)));
+      simulateSchedule(schedule);
+      activitiesInserted.addAll(toBeInserted);
+    } else {
+      final var schedule = Map.of(activityToSimulate.id,
+                                  Pair.of(activityToSimulate.start, activityToSimulate.activity));
+      simulateSchedule(schedule);
+      activitiesInserted.add(activityToSimulate);
+    }
+  }
+
+
+  /**
+   * Get the simulation results from the Duration.ZERO to the current simulation time point
+   * @return the simulation results
+   */
+  public SimulationResults getSimulationResults(){
+    return getSimulationResultsUntil(curTime);
+  }
+
+  /**
+   * Get the simulation results from the Duration.ZERO to a specified end time point.
+   * The provided simulation results might cover more than the required time period.
+   * @return the simulation results
+   */
+  public SimulationResults getSimulationResultsUntil(Duration endTime){
+    //if previous results cover a bigger period, we return do not regenerate
+    if(lastSimResults == null || endTime.longerThan(lastSimResultsEnd)) {
+      final Map<String, ActivityInstanceId> convertedTaskToPlannedDir = new HashMap<>();
+      taskToPlannedDirective.forEach((taskId, plannedDirective)->
+                                         convertedTaskToPlannedDir.put(taskId.id(), plannedDirective));
+      lastSimResults = engine.computeResults(
+          engine,
+          Instant.now(),
+          endTime,
+          convertedTaskToPlannedDir,
+          timeline,
+          missionModel);
+      lastSimResultsEnd = endTime;
+      //while sim results may not be up to date with curTime, a regeneration has taken place after the last insertion
+    }
+    return lastSimResults;
+  }
+
+  private void simulateSchedule(Map<ActivityInstanceId, Pair<Duration, SerializedActivity>> schedule){
+
+    for (final var entry : schedule.entrySet()) {
+      final var directiveId = entry.getKey();
+      final var startOffset = entry.getValue().getLeft();
+      final var directive = entry.getValue().getRight();
+
+      final var taskId = engine.initiateTaskFromInput(missionModel, directive);
+      engine.scheduleTask(taskId, startOffset);
+      plannedDirectiveToTask.put(directiveId,taskId);
+      taskToPlannedDirective.put(taskId, directiveId);
+    }
+
+    while (true) {
+      final var batch = engine.extractNextJobs(Duration.MAX_VALUE);
+      // Increment real time, if necessary.
+      final var delta = batch.offsetFromStart().minus(curTime);
+      curTime = batch.offsetFromStart();
+      timeline.add(delta);
+      // TODO: Advance a dense time counter so that future tasks are strictly ordered relative to these,
+      //   even if they occur at the same real time.
+
+      // Run the jobs in this batch.
+      final var commit = engine.performJobs(batch.jobs(), cells, curTime, Duration.MAX_VALUE, missionModel);
+      timeline.add(commit);
+
+      // Exit IFF all tasks are complete
+      if (taskToPlannedDirective.keySet().stream().allMatch(taskId -> engine.isTaskComplete(taskId))) {
+        break;
+      }
+
+    }
+    lastSimResults = null;
+  }
+
+  /**
+   * Returns the duration of a terminated simulated activity
+   * @param activityInstanceId the activity id
+   * @return its duration if the activity has been simulated and has finished simulating, an IllegalArgumentException otherwise
+   */
+  public Duration getActivityDuration(ActivityInstanceId activityInstanceId){
+    return engine.getTaskDuration(plannedDirectiveToTask.get(activityInstanceId));
+  }
+
+}
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
@@ -1,0 +1,42 @@
+package gov.nasa.jpl.aerie.scheduler;
+
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
+import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
+import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SimulationDriverFacade {
+
+  private final IncrementalSimulationDriver driver;
+
+  private final Map<String, ActivityInstanceId> activityNameToActivityId = new HashMap<>();
+
+  private static long count = 0;
+
+  public SimulationDriverFacade(IncrementalSimulationDriver driver){
+    this.driver = driver;
+  }
+
+  public void simulateActivity(SerializedActivity activity, String nameAct, Duration startTime){
+    final var activityId = new ActivityInstanceId(count++);
+    activityNameToActivityId.put(nameAct, activityId);
+    driver.simulateActivity(activity,startTime, activityId);
+  }
+
+  public Duration getActivityDuration(String activityName){
+    var activityId = activityNameToActivityId.get(activityName);
+    return driver.getActivityDuration(activityId);
+  }
+
+  public SimulationResults getSimulationResults(){
+    return driver.getSimulationResults();
+  }
+
+  public SimulationResults getSimulationResultsUntil(Duration endTime){
+    return driver.getSimulationResultsUntil(endTime);
+  }
+
+}
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationUtility.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationUtility.java
@@ -2,7 +2,6 @@ package gov.nasa.jpl.aerie.scheduler;
 
 import gov.nasa.jpl.aerie.banananation.Configuration;
 import gov.nasa.jpl.aerie.banananation.generated.ConfigurationMapper;
-import gov.nasa.jpl.aerie.banananation.generated.GeneratedMissionModelFactory;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModel;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModelBuilder;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
@@ -12,13 +11,23 @@ import java.nio.file.Path;
 public final class SimulationUtility {
 
   private static MissionModel<?> makeMissionModel(final MissionModelBuilder builder, final SerializedValue config) {
-    final var factory = new GeneratedMissionModelFactory();
+    final var factory = new gov.nasa.jpl.aerie.banananation.generated.GeneratedMissionModelFactory();
     final var model = factory.instantiate(config, builder);
     return builder.build(model, factory.getTaskSpecTypes());
   }
 
+  public static MissionModel<?>
+  getFooMissionModel() {
+    final var conf = new gov.nasa.jpl.aerie.foomissionmodel.Configuration();
+    final var serializedConfig = SerializedValue.of(new gov.nasa.jpl.aerie.foomissionmodel.generated.ConfigurationMapper().getArguments(conf));
+    final var factory = new gov.nasa.jpl.aerie.foomissionmodel.generated.GeneratedMissionModelFactory();
+    final var builder = new MissionModelBuilder();
+    final var model = factory.instantiate(serializedConfig, builder);
+    return builder.build(model, factory.getTaskSpecTypes());
+  }
 
-  public static MissionModel<?> getMissionModel(){
+
+  public static MissionModel<?> getBananaMissionModel(){
     final var config = new Configuration(Configuration.DEFAULT_PLANT_COUNT, Configuration.DEFAULT_PRODUCER, Path.of("/etc/hosts"));
     final var serializedConfig = SerializedValue.of(new ConfigurationMapper().getArguments(config));
     return makeMissionModel(new MissionModelBuilder(), serializedConfig);
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/BasicActivity.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/BasicActivity.java
@@ -1,0 +1,16 @@
+package gov.nasa.jpl.aerie.foomissionmodel.activities;
+
+import gov.nasa.jpl.aerie.foomissionmodel.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.delay;
+
+@ActivityType("BasicActivity")
+public final class BasicActivity {
+  @EffectModel
+  public void run(final Mission mission) {
+    delay(Duration.of(1, Duration.SECONDS));
+  }
+}
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
@@ -5,6 +5,7 @@
 @WithMappers(BasicValueMappers.class)
 @WithMappers(FooValueMappers.class)
 
+@WithActivityType(BasicActivity.class)
 @WithActivityType(FooActivity.class)
 @WithActivityType(BarActivity.class)
 @WithActivityType(DecompositionTestActivities.ParentActivity.class)
@@ -14,6 +15,7 @@ package gov.nasa.jpl.aerie.foomissionmodel;
 
 import gov.nasa.jpl.aerie.contrib.serialization.rulesets.BasicValueMappers;
 import gov.nasa.jpl.aerie.foomissionmodel.activities.BarActivity;
+import gov.nasa.jpl.aerie.foomissionmodel.activities.BasicActivity;
 import gov.nasa.jpl.aerie.foomissionmodel.activities.DecompositionTestActivities;
 import gov.nasa.jpl.aerie.foomissionmodel.activities.FooActivity;
 import gov.nasa.jpl.aerie.foomissionmodel.mappers.FooValueMappers;
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -128,6 +128,10 @@ public final class SimulationDriver {
     }
   }
 
+  public static ControlTask buildPlanTask(Map<ActivityInstanceId, Pair<Duration, SerializedActivity>> schedule){
+    return new ControlTask(schedule);
+  }
+
   private static final class ControlTask implements Task<Unit> {
     private final Map<ActivityInstanceId, Pair<Duration, SerializedActivity>> schedule;
 
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationBenchmark.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationBenchmark.java
@@ -1,0 +1,83 @@
+package gov.nasa.jpl.aerie.scheduler;
+
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
+import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
+import gov.nasa.jpl.aerie.merlin.driver.SimulationDriver;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
+
+public class IncrementalSimulationBenchmark {
+
+  /**
+   * This benchmark builds incrementally longer plans of sequential non-overlapping activities and simulates them with
+   * incremental and non-incremental simulation drivers. Each printed timing _left _right shows the time _right it took
+   * to simulate a plan with a number of _left activities. Note that this is an ideal case for the incremental driver as
+   * it never has to reset.
+   */
+  public static void main(String[] args){
+    System.out.println("Incremental");
+    benchmarkIncrementalSimulationDriver();
+    System.out.println("Non-incremental");
+    benchMarkSimulationDriver();
+  }
+
+  private static ArrayList<IncrementalSimulationTest.TestSimulatedActivity> getActivities(){
+    final int nbActs = 5000;
+    final var step = Duration.of(3,SECOND);
+    final var durSim = step.times(nbActs);
+    final var acts = new ArrayList<IncrementalSimulationTest.TestSimulatedActivity>();
+    for(var cur = Duration.ZERO; cur.shorterThan(durSim); cur = cur.plus(step) ){
+      var act = new IncrementalSimulationTest.TestSimulatedActivity(
+          cur,
+          new SerializedActivity("BasicActivity", Map.of()),
+          String.valueOf(cur.in(MICROSECONDS)));
+      acts.add(act);
+    }
+    return acts;
+  }
+
+
+  private static void benchMarkSimulationDriver(){
+    final var acts = getActivities();
+    final var fooMissionModel = SimulationUtility.getFooMissionModel();
+    int i = 0;
+    var sum = 0.;
+    final var alreadyIn = new HashMap<ActivityInstanceId, Pair<Duration, SerializedActivity>>();
+    //builds incrementally long plans and simulates them
+    for(var act:acts){
+      alreadyIn.put(new ActivityInstanceId(Long.parseLong(act.name())), Pair.of(act.start(), act.activity()));
+      final var task = SimulationDriver.buildPlanTask(alreadyIn);
+      final var start = System.nanoTime();
+      SimulationDriver.simulateTask(fooMissionModel, task);
+      final var dur = System.nanoTime() - start;
+      sum += dur;
+      final var curMean = sum / (++i);
+      System.out.println(i + " " + curMean);
+    }
+  }
+
+  private static void benchmarkIncrementalSimulationDriver(){
+    final var acts = getActivities();
+    final var fooMissionModel = SimulationUtility.getFooMissionModel();
+    final SimulationDriverFacade incrementalSimulationDriver = new SimulationDriverFacade(new IncrementalSimulationDriver(fooMissionModel));
+    int i = 0;
+    var sum = 0.;
+    for(var act : acts) {
+      final var start = System.nanoTime();
+      incrementalSimulationDriver.simulateActivity(act.activity(), act.name(), act.start());
+      final var dur = System.nanoTime() - start;
+      sum += dur;
+      final var curMean = sum / (++i);
+      System.out.println(i + " " + curMean);
+    }
+  }
+
+
+}
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationTest.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationTest.java
@@ -1,0 +1,70 @@
+package gov.nasa.jpl.aerie.scheduler;
+
+import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Objects;
+
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECONDS;
+
+public class IncrementalSimulationTest {
+
+  SimulationDriverFacade incrementalSimulationDriver;
+  Duration endOfLastAct;
+
+  @BeforeEach
+  public void init() {
+    final var acts = getActivities();
+    final var fooMissionModel = SimulationUtility.getFooMissionModel();
+    incrementalSimulationDriver =
+        new SimulationDriverFacade(new IncrementalSimulationDriver(fooMissionModel));
+    for (var act : acts) {
+      final var start = System.nanoTime();
+      incrementalSimulationDriver.simulateActivity(act.activity, act.name, act.start);
+    }
+  }
+  @Test
+  public void simulationResultsTest(){
+    //ensures that simulation results are generated until the end of the last act;
+    var simResults = incrementalSimulationDriver.getSimulationResults();
+    assert(simResults.realProfiles.get("/utcClock").get(0).getLeft().isEqualTo(endOfLastAct));
+    /*ensures that when current simulation results cover more than the asked period and that nothing has happened
+    between two requests, the same results are returned*/
+    var simResults2 = incrementalSimulationDriver.getSimulationResultsUntil(Duration.of(7,SECONDS));
+    assert(Objects.equals(simResults, simResults2));
+  }
+
+  @Test
+  public void durationTest(){
+    var act1Dur = incrementalSimulationDriver.getActivityDuration("first");
+    var act2Dur = incrementalSimulationDriver.getActivityDuration("second");
+    assert(act1Dur.isEqualTo(Duration.of(1, SECONDS)));
+    assert(act2Dur.isEqualTo(Duration.of(1, SECONDS)));
+  }
+
+  private ArrayList<TestSimulatedActivity> getActivities(){
+    final var acts = new ArrayList<TestSimulatedActivity>();
+    var act1 = new TestSimulatedActivity(
+        Duration.of(0, SECONDS),
+        new SerializedActivity("BasicActivity", Map.of()),
+        "first");
+    acts.add(act1);
+    var act2 = new TestSimulatedActivity(
+        Duration.of(14, SECONDS),
+        new SerializedActivity("BasicActivity", Map.of()),
+        "second");
+    acts.add(act2);
+
+    endOfLastAct = Duration.of(15,SECONDS);
+    return acts;
+  }
+
+
+  record TestSimulatedActivity(Duration start, SerializedActivity activity, String name){}
+
+
+}
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationFacadeTest.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationFacadeTest.java
@@ -60,7 +60,7 @@ public class SimulationFacadeTest {
 
   @BeforeEach
   public void setUp() {
-    missionModel = SimulationUtility.getMissionModel();
+    missionModel = SimulationUtility.getBananaMissionModel();
     wrappedMissionModel = new MissionModelWrapper(missionModel, horizon);
     facade = new SimulationFacade(horizon, missionModel);
   }


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -503,6 +503,14 @@ public SimulationResults computeResults(
                                  serializedTimeline);
   }
 
+  public Duration getTaskDuration(TaskId id){
```
Twisol: Suggest `getTaskDurationIfTerminated`, or `getTerminatedTaskDuration`, to clarify at calling sites that this isn't simply guessing the duration of an in-progress task.

Twisol: Recommend returning an `Optional<Duration>`, to avoid leaking `null`s outside component boundaries. (Nulls are infectious!)

Alternatively, we could throw an exception if the task isn't terminated, if we expect this method to only ever be called on a terminated task.

foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
```diff
@@ -4,6 +4,7 @@
 
 @WithMappers(BasicValueMappers.class)
 @WithMappers(FooValueMappers.class)
+@WithActivityType(BasicActivity.class)
```
Twisol: This annotation should be grouped with the other `@WithActivityType`s below. (That is, this line should be swapped with the empty line below it.)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+  private Map<SerializedActivity, TaskId> actToId;
+
+  //boolean stating whether the simulation results should be regenerated
+  private boolean areLastSimResultsDirty;
+
+  //simulation results so far
+  private SimulationResults lastSimResults;
+
+  //should the simulation be resetted everytime an activity is simulated
+  private boolean resetEveryTime;
```
Twisol: Based on the PR summary, I assume this flag is only used for performance testing? If so, and given the clear wins we get from not restarting after every change to the schedule, it would probably be best to remove this and replace it with `false` everywhere it was used.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+public class StepSimulation {
+
+  private Duration curTime;
+  private SimulationEngine engine;
+  private LiveCells cells;
+  private TemporalEventSource timeline;
+  private MissionModel<?> missionModel;
+
+  //used for looking up durations
+  private Map<SerializedActivity, TaskId> actToId;
```
Twisol: Comments should describe what a thing is, not what it is used for; it will too-easily get out of sync with its usage over time otherwise, as new uses for the data come along.

Twisol: This map might be ill-founded; it's possible to have two identical activities (even at the same time), in which case you will have two tasks for the same serialized activity. While two identical activities at the same time ought to have the same duration (thanks to the simulation's transactional semantics), that's not necessarily the case for identical activities at _different_ times, which are allowed by this map.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+  private SimulationResults lastSimResults;
+
+  //should the simulation be resetted everytime an activity is simulated
+  private boolean resetEveryTime;
+
+  //map from activity start time to serialized activity and its name
+  private LinkedHashMap<Duration, Pair<SerializedActivity, String>> activitiesInserted;
+
+  /*when we really want to reset the whole simulation*/
+  public void clearActInserted(){
```
Twisol: This comment isn't clear to me -- is this describing the method itself, or one particular caller's use case?

Is this method called from anywhere?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+import java.time.Instant;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.PriorityQueue;
+import java.util.Set;
+
+public class StepSimulation {
```
Twisol: "StepSimulation" is a better name for a method than a class. Is there a noun or noun-like phrase we can use, like `SimulationStepper`? Alternatively, `IncrementalSimulationDriver` (by symmetry with the existing driver), or `SimulatedSchedule` (representing a mutable schedule and its current results)?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+
+  /*when we really want to reset the whole simulation*/
+  public void clearActInserted(){
+    activitiesInserted.clear();
+  }
+
+  public StepSimulation(MissionModel<?> missionModel){
+    this.missionModel = missionModel;
+    resetEveryTime = false;
+    activitiesInserted = new LinkedHashMap<>();
```
Twisol: Could we move these constant values up to the field initializers? At least for me, it helps to separate the consistently-initialized fields from the ones that depend on constructor arguments; and in this case, it's good to see that the default is to reset every time, and that we always start with an empty list.

In particular, fields without initializers begin `null`, and it's hard to tell at a glance which things stay null and which things are immediately assigned to an actual value.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public void resetEveryTime(){
+    resetEveryTime= true;
```
Twisol: Consistency with the rest of the file:
```suggestion
    resetEveryTime = true;
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+        engine.trackResource(name, resource, curTime);
+      }
+
+      // Start daemon task(s) immediately, before anything else happens.
+      {
+        final var daemon = engine.initiateTaskFromSource(missionModel::getDaemon);
+        final var commit = engine.performJobs(Set.of(SimulationEngine.JobId.forTask(daemon)),
+                                              cells, curTime, Duration.MAX_VALUE, missionModel);
+        timeline.add(commit);
+      }
```
Twisol: This series of code should be indented by one less level.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+    this.missionModel = missionModel;
+    resetEveryTime = false;
+    activitiesInserted = new LinkedHashMap<>();
+  }
+
+  public void resetEveryTime(){
+    resetEveryTime= true;
+  }
+
+  public void initSimulation(){
```
Twisol: Can we call this from the constructor? This method also initializes some fields (like `actToId`) that are morally never non-null, and I don't yet see a use-case where a caller will ever construct this class without immediately calling `initSimulation` afterward.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      toBeReinserted.putAll(activitiesInserted);
+      initSimulation();
+      insertAllActsBefore(toBeReinserted, startTime);
+      simulateAct(activity, startTime,name);
+      insertAllActsAfter(toBeReinserted, startTime);
+    }
+    simulateAct(activity, startTime,name);
+  }
+
+  private void insertAllActsBefore(LinkedHashMap<Duration, Pair<SerializedActivity, String>> acts, Duration startTime){
```
Twisol: This method can accept just a `Map`, so the caller can provide whatever type of map they want (including the immutable kind produced by `Map.of()`).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      var start = act.getKey();
+      var activity = act.getValue().getLeft();
+      var name = act.getValue().getRight();
+      if(start.shorterThan(startTime)) {
+        this.simulateAct(activity, startTime, name);
+      }
+    }
+  }
+
+  private void insertAllActsAfter(LinkedHashMap<Duration, Pair<SerializedActivity, String>> acts, Duration startTime){
```
Twisol: This method can accept just a `Map`, so the caller can provide whatever type of map they want (including the immutable kind produced by `Map.of()`).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+        final var commit = engine.performJobs(Set.of(SimulationEngine.JobId.forTask(daemon)),
+                                              cells, curTime, Duration.MAX_VALUE, missionModel);
+        timeline.add(commit);
+      }
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, String name){
+    if(startTime.shorterThan(curTime) || resetEveryTime){
+      var toBeReinserted = new LinkedHashMap<Duration, Pair<SerializedActivity, String>>();
+      toBeReinserted.putAll(activitiesInserted);
```
Twisol: I think `var toBeReinserted = new LinkedHashMap<>(activitiesInserted);` suffices, thanks to the convenience constructor.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      initSimulation();
+      insertAllActsBefore(toBeReinserted, startTime);
+      simulateAct(activity, startTime,name);
+      insertAllActsAfter(toBeReinserted, startTime);
+    }
+    simulateAct(activity, startTime,name);
```
Twisol: Is it intentional that the activity passed to `simulateActivity` is simulated twice if resimulation is necessary? (That is, I'd expect this line to be in an `else`.)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      }
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, String name){
+    if(startTime.shorterThan(curTime) || resetEveryTime){
+      var toBeReinserted = new LinkedHashMap<Duration, Pair<SerializedActivity, String>>();
+      toBeReinserted.putAll(activitiesInserted);
+      initSimulation();
+      insertAllActsBefore(toBeReinserted, startTime);
+      simulateAct(activity, startTime,name);
```
Twisol: Here and a couple lines below, missing whitespace for consistency:
```suggestion
      simulateAct(activity, startTime, name);
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+  }
+
+  private void insertAllActsAfter(LinkedHashMap<Duration, Pair<SerializedActivity, String>> acts, Duration startTime){
+    for(var act : acts.entrySet()){
+      var start = act.getKey();
+      var activity = act.getValue().getLeft();
+      var name = act.getValue().getRight();
+      if(start.longerThan(startTime) || start.isEqualTo(startTime)) {
```
Twisol: This condition can be `start.noShorterThan(startTime)`, i.e. the exact negation of the analogous check in `insertAllActsBefore`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, String name){
+    if(startTime.shorterThan(curTime) || resetEveryTime){
+      var toBeReinserted = new LinkedHashMap<Duration, Pair<SerializedActivity, String>>();
+      toBeReinserted.putAll(activitiesInserted);
+      initSimulation();
+      insertAllActsBefore(toBeReinserted, startTime);
+      simulateAct(activity, startTime,name);
+      insertAllActsAfter(toBeReinserted, startTime);
```
Twisol: I'm a little nervous about this logic. If the new activity occurs at the same time as an existing ("reinserted") one, then this breakdown will violate the expectation that activities occurring at the same time are simulated concurrently. If I read this right, the given activity will be considered to occur _causally before_ (and hence visible to) the existing activities at this time. That's not good, because we'll get different simulation results and modeled task behaviors depending on whether we're running the whole schedule in one go (as the Merlin server does) or incrementally building up the schedule (as here).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      {
+        final var daemon = engine.initiateTaskFromSource(missionModel::getDaemon);
+        final var commit = engine.performJobs(Set.of(SimulationEngine.JobId.forTask(daemon)),
+                                              cells, curTime, Duration.MAX_VALUE, missionModel);
+        timeline.add(commit);
+      }
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, String name){
+    if(startTime.shorterThan(curTime) || resetEveryTime){
```
Twisol: Similarly to the sibling comment, this check probably ought to be `startTime.noLongerThan(curTime)`, as activities running at the same time ought to be invisible to each other. However, as written here, a new activity at `curTime` will causally observe the effects of other activities at `curTime`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+          curTime,
+          new HashMap<String, ActivityInstanceId>(),
+          timeline,
+          missionModel);
+    }
+    return lastSimResults;
+  }
+
+  //WARNING computes the results for the whole horizon until current time them from beginning
+  public SimulationResults getSimulationResultsUntil(Duration endTime){
```
Twisol: I'm not sure, but because of the caching in `lastSimResults`, if I call this method twice with different methods and no intervening changes, the second call will (erroneously?) return the same results set as from the first call, yes? Should `endTime` be cached alongside `lastSimResults`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      var activity = act.getValue().getLeft();
+      var name = act.getValue().getRight();
+      if(start.longerThan(startTime) || start.isEqualTo(startTime)) {
+        this.simulateAct(activity, startTime, name);
+      }
+    }
+  }
+
+  //WARNING computes the results for the whole horizon until current time them from beginning
+  public SimulationResults getSimulationResults(){
```
Twisol: Can this method just be `return getSimulationResultsUntil(curTime)`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+      timeline.add(commit);
+    }
+    areLastSimResultsDirty = true;
+    activitiesInserted.put(startTime, Pair.of(activity, name));
+  }
+
+  public Duration getActivityDuration(SerializedActivity act){
+    //peek into engine
+    return engine.getTaskDuration(actToId.get(act));
+  }
```
Twisol: Is this method only for testing purposes, or are there production use-cases for it? The `SimulationResults` class already provided a map of completed activities from which you can query activity durations -- I'm not sure we want to poke a(nother) hole in the engine just to get at this data.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public boolean activityHasFinished(SerializedActivity act){
```
Twisol: Is this method used anywhere, or are there any production use-cases expected for it?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationUtility.java
```diff
@@ -12,13 +12,23 @@
...
 
   private static MissionModel<?> makeMissionModel(final MissionModelBuilder builder, final SerializedValue config) {
-    final var factory = new GeneratedMissionModelFactory();
+    final var factory = new gov.nasa.jpl.aerie.banananation.generated.GeneratedMissionModelFactory();
     final var model = factory.instantiate(config, builder);
     return builder.build(model, factory.getTaskSpecTypes());
   }
 
+  public static MissionModel<?>
+  getFooMissionModel() {
```
Twisol, line 20: For consistency within file:
```suggestion
  public static MissionModel<?> getFooMissionModel() {
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationUtility.java
```diff
@@ -12,13 +12,23 @@
...
+    var conf = new gov.nasa.jpl.aerie.foomissionmodel.Configuration();
```
Twisol: For consistency within file:
```suggestion
    final var conf = new gov.nasa.jpl.aerie.foomissionmodel.Configuration();
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+  @Test
+  public void dontreset() {
+    manyacts(false);
+  }
+
+  public void manyacts(boolean reset){
+    var cur = Duration.ZERO;
+    final var step = Duration.of(3,SECOND);
+    // ten days
+    final var durSim = Duration.of(864000, SECOND);
```
Twisol: We can internalize the comment, e.g. as:
```suggestion
    final var DAY = Duration.HOUR.times(24);
    final var durSim = DAY.times(10);
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    final var acts = new ArrayList<Triple<Duration, SerializedActivity, String>>();
+    for(; cur.shorterThan(durSim); cur = cur.plus(step) ){
```
Twisol: `cur` isn't used outside this loop, so we can move it into the loop initializer:
```suggestion
    for(var cur = Duration.ZERO; cur.shorterThan(durSim); cur = cur.plus(step) ){
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+      var act = Triple.of(
+          cur,
+          new SerializedActivity("BasicActivity", Map.of()), String.valueOf(cur));
```
Twisol: Two of the components of this "triple" are on one line, making it hard to parse its components visually.
```suggestion
      var act = Triple.of(
          cur,
          new SerializedActivity("BasicActivity", Map.of()),
          String.valueOf(cur));
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+public class StepSimulationTest {
+
+  @Test
+  public void resetEveryTime() {
+    manyacts(true);
+  }
+  @Test
+  public void dontreset() {
+    manyacts(false);
+  }
```
Twisol: This class smells less like a test and more like a top-level program that happens to be a benchmark, especially because there are no assertions being checked. Since this is a benchmark, we don't really want ot ti be run alongside all the other unit tests; unit tests should be fast!

Recommend making this method a top-level `main`, and removing the other method (`resetEveryTime`), per earlier comment about removing the performance-test-only resetting toggle.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void manyacts(boolean reset){
+    var cur = Duration.ZERO;
+    final var step = Duration.of(3,SECOND);
+    // ten days
+    final var durSim = Duration.of(864000, SECOND);
+
+    final var acts = new ArrayList<Triple<Duration, SerializedActivity, String>>();
```
Twisol: Honestly, it might be good to extract this Triple out as a `record TestActivity(Duration offset, SerializedActivity activity, String name) {}`, so that you can work with named fields within the test loop. Makes it much easier to follow what's going on, and for minimal extra code.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    System.out.println(acts.size() + " activities to simulate");
+    int i = 0;
+    var curMean = 0.;
+    var sum = 0.;
+    for(var act : acts) {
+      //System.out.println("Sim act " + (i++));
+      var start = System.nanoTime();
+      stepSimulation.simulateActivity(act.getMiddle(), act.getLeft(), act.getRight());
+      stepSimulation.getActivityDuration(act.getMiddle());
```
Twisol: Can you explain what `getActivityDuration` is being used for in this test? Its return value is ignored, and its implementation proxies to a cheap data lookup in the engine. Can this call be removed?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
```diff
@@ -0,0 +1,64 @@
...
+      var dur = System.nanoTime() - start;
+      i++;
+      sum+=dur;
+      curMean = sum/i;
+      System.out.println(i + " " + curMean);
+      if(i>10000){
+        break;
+      }
```
Twisol: If we want to control the number of steps taken by the test, and `durSim` is only used to generate enough activities, perhaps it would be better to generate exactly 10,000 activities rather than however many are generated over the currently-specified 10 days?

Twisol (commented):
This is an exciting PR! It does feel very nascent, as you call out, but I can see the outlines of a really nice abstraction in here. In particular, this feels like a "live schedule" -- an object representing an editable schedule, which you can ask for results from at any time, and then make further edits. This seems like a great abstraction for use by the scheduler, and potentially even a useful way to make the Merlin server perform incremental simulations as well -- even if across an API boundary.

I've got a bunch of other comments inline, most of which aren't vastly important, but there are a few semantic correctness issues I wanted to call out around the order of activity execution.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: Coming back to this after completing a review, I'm not convinced it's necessary to add this method to the engine. Curious to hear your thoughts.

adrienmaillard: It has to do with how I project the use of this driver. Before, the `SimulationResults` were always computed and we would take the durations of activities from there but it is a lot of work to perform each time when I know that the scheduler just wants to know about the duration of the last activity inserted.

adrienmaillard: I believe an exception would tell us that the scheduler is not working properly which is good. Do we have a general policy for exceptions in Aerie ?

foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
adrienmaillard: Yes. IDE automated magic does not care about tidiness :)

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
adrienmaillard: Performance comparison
Just for you @Twisol (this is on Slack), this is the graphical version of what is printed during the tests (not sure this is clear), showing the difference in performance between the incremental and reset versions. This is exciting, I hope it translates when we use it with scheduling.

adrienmaillard: This is exactly what you are saying, a benchmark. I will make this a program. About the other method, I'll answer where you commented.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
adrienmaillard: I wanted to see if it made a difference in terms of performance but yes it will be removed.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
adrienmaillard: Totally. Part of the quick-draft-pr to start the discussion.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: In the long term, for sure. But before it is really mature, it feels very useful for debugging against a version that is behaving like the existing simulation driver.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
Twisol: Yes, the graph makes sense; you're not re-running the entire simulation every time you add an activity in the future. That behavior is clearly quadratic, and it only ever made sense for full runs of complete plans. This is an exciting optimization; I'm really hopeful that this can be refined, over time, to replace the existing driver altogether.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: Yes. In the scheduler, tasks have unique ids. I guess we need to import them here as well. Maybe the `StepSimulation` can take care of translating them to `SerializedActivity`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: It is not called, it is an artefact and should be removed.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: Not really; it's historically been a case-by-case justification, with some rather loose philosophy behind each choice.

Broadly, I think of exceptions in three cases: caller-callee contract failures, component integration contract failures, and faulty developer assumptions.

* In the first, the caller directly made an illegal request; either the arguments were bad, or the method was called in a program state in which it must not be. (That's the case here.)

* In the second, neither the caller nor callee are directly at fault, but the callee is unable to fulfill its side of the contract -- for instance, a network issue occurred behind an interface that doesn't expose the network details. In that case, whichever component brought the caller's component and callee's component into existence, and wired them together, is at fault -- or at least is in the best position to correct for the fault and move on.

* In the third, the program itself isn't really at fault. Rather, the developer(s) who built the program expected that a certain scenario would never happen, and it did. These are good candidates for `Error`-subtype exceptions, as we don't normally want these to be caught -- if we did, the fault would have been accounted for at some level. Best-case, this is treated the same as the previous class, but likely higher-up in the tree of component decomposition, and with a louder error message to the developers. Worst-case, the program terminates.

In all cases, unchecked exceptions are used. Checked exceptions are *much* more like an "alternate return value", and so don't really capture exceptional situations at all. These days, we have `sealed` interfaces that can be used instead to model alternatives, though niche use cases do occasionally still arise.

Twisol: In this case, we should document the contract that `getTaskDuration` (or alternate name recommended) must only be called when the task is known to have terminated; an exception of a specified type will be thrown otherwise.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: I like `IncrementalSimulationDriver`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: We can. Even on the subject of this method being public, I am more and more inclined to make it private and prevent the user to mess with the internal behavior. Only thing it would allow is not recreating a new object between "simulation runs".

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: It needs to be ordered by key (we want chronological insertion). So `OrderedMap` to be less specific ?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: See my comment for `insertAllActsBefore`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: This is a gross mistake and I am ashamed :)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
mattdailis: I think you can use `SortedMap` here, since that expects the map to be sorted by key (`Duration` in this case). `LinkedHashMap` is ordered by insertion order into the map, which I don't think is what you want

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
Twisol: 🤔 How hard do you expect it would be to write a benchmark that really does use the existing simulation driver? The `SimulationDriver#simulateTask` method doesn't compute simulation results, so it seems like it should be fairly comparable?

Honestly, I'd recommend running such a benchmark anyway, as the benchmarks you've composed are all against the new code. We have to trust that there are no bugs or other inconsistencies between `resetEveryTime = true` and the existing driver.

I'm certain we'd still see a change from (sharply!) quadratic to linear under this benchmark, but it feels like a much safer way to make inferences against the behavior of the existing system.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
Twisol: > Only thing it would allow is not recreating a new object between "simulation runs".

I don't think there's much value in that, yeah. `private` away!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: I suspected that, hence asking for checks on correctness. I suppose that, in this case, what we can do is bundle all the tasks together (before, current, after) and use the `ControlTask` as it is used in the `SimulationDriver`, what do you think? While we are on these aspects, you can see that the curTime is updated in the ControlTask but never set on the end of the last activity (I would say it stays at the beginning of this activity), do you see that as an issue ? If the next activity starts at the same time, I guess we fall again in the problem you describe here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: I've answered that in another comment, but basically, we do not want to re-build the `SimulationResults` every time we simulate an activity to get its duration (or test its placing), that's why this is present.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: No it can be removed for now.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
Twisol: I think Adrien has the caller inserting in sorted order, so it happens to work out. But you're right, @mattdailis, the method should document somehow (either by argument type or by comment) that it expects the provided map to be iterable in sorted order.

Honestly, the map provided to this method, `activitiesInserted`, is never really used as a map at all -- the only operations performed on it are essentially `put`, `entrySet`, and `clear`. This could be replaced with a sorted list (replace `put` with sorted insert, `entrySet` with regular `iterator()`, and `clear` stays the same). Then both `insertAllActsBefore` and `insertAllActsAfter` can be lifted into the lone caller, which itself can be significantly simplified.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
```diff
@@ -3,7 +3,7 @@
 import java.util.UUID;
 
 /** A typed wrapper for task IDs. */
-/*package-local*/ record TaskId(String id) {
+/*package-local*/ public record TaskId(String id) {
   public static TaskId generate() {
     return new TaskId(UUID.randomUUID().toString());
   }
```
goetzrrGit: Just so you and @mattdailis are aware, there is going to be a collision with this class.  @mattdailis changed the structure to this:

```
/** A typed wrapper for task IDs. */
public record TaskId(String id) implements Scheduler.TaskId {
  public static TaskId generate() {
    return new TaskId(UUID.randomUUID().toString());
  }
}
```

goetzrrGit (commented):
A lot of great feedback already. I just wanted to point out the collision to `TaskId.java`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: In the revision I am about to push, I let the ControlTask take care of the ordering now for scheduling multiple activities as it is done in the SimulationDriver. All of this has disappeared.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
mattdailis: Hmm... I'm trying to remember why I made it public. It might not need to be for computed attributes.

Also if we're getting rid of `waitFor` as suggested by Jonathan, it won't need to implement Scheduler.TaskId either 👍

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
adrienmaillard: The engine is now throwing an IllegalArgumentException when the task has not finished (or been scheduled).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: In the latest changes, you will see a benchmark against the `SimulationDriver`. I have removed this boolean and the associated method.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: You're right, it has been changed.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: It does now.

patkenneally:
I realize that this started as a draft PR to share. It's great that it has had so much work and discussion. If we anticipate turning this into fully fledged PR at some point, we need to be mindful that a PR's commit history should show a logical build up of the feature/change being presented. As a result commits should be logically atomic steps in walking the reviewer to that end. For the most recent commit with message "Address draft PR comments", these changes should be redistributed through the commit history of this PR via a `git rebase -i` . @adrienmaillard I'll leave it to your judgement as to when and how you want to refactor the history of this branch to show the incremental build up of this change.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
adrienmaillard: Peripheral question about this: why does `Scheduler.spawn` return a `String` version of the `TaskId` but not the object used in the engine? I had to add a nasty search in the `SimulationEngine` to be able to get the `TaskId` object corresponding to the same id.

mattdailis: `Scheduler.spawn` is the interface used by the model - we do not want the model to be able to use task ids for anything other than calling `waitFor(taskid)`.

Since TaskId is a record, you actually don't really need to find the original object - you can make `new TaskId(String)` and it will behave in all the same ways (same hashcode, same equals). Just don't use `==` on it.

We are getting rid of spawn returning a String - it's going to return `void` quite soon. The driver will need to get its task ids some other way 🤔

mattdailis: There's also the package dependency reason: `TaskId` is in `merlin-driver` (where the engine lives). The model does not have access to the engine's types. The only types that cross the boundary between the model and the engine live in `merlin-sdk`, in the `protocol` package.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: `IllegalStateException` might be more accurate, since it's not the task ID itself that's invalid, but rather the state of the simulation at which you're querying that task ID.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
Twisol: > The driver will need to get its task ids some other way 🤔

Yes; in [the branch](https://github.com/NASA-AMMOS/aerie/compare/feature/remove-waitFor) where I've prototyped this `void spawn()` change, I ended up not using a control task at all. Instead, the driver just loads all of the tasks into the engine up front, since the engine still returns proper task IDs from its own inherent methods.

This is technically a de-optimization, since the control task was meant to avoid polluting the engine's priority queue with a bunch of fixed tasks, but we can always address that in a different way.

adrienmaillard: That will probably change how **everything** is done in this incremental simulation, from ControlTask to getting durations back. We might want to wait for that change to be merged.

adrienmaillard: @mattdailis my use case was getting execution states from the SimulationEngine.tasks map. And creating new TaskId for that did not work because they are different objects.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,245 @@
...
+   */
+  public SimulationResults getSimulationResults(){
+    return getSimulationResultsUntil(curTime);
+  }
+
+  /**
+   * Get the simulation results from the Duration.ZERO to a specified end time point.
+   * The provided simulation results might cover more than the required time period.
+   * @return the simulation results
+   */  public SimulationResults getSimulationResultsUntil(Duration endTime){
```
Twisol: This method's first line has rolled up onto the end of its javadoc.
```suggestion
   */
   public SimulationResults getSimulationResultsUntil(Duration endTime){
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
Twisol: Something else might have been confounding things, because every `record` automatically generates an overridden `equals()` and `hashCode()` that invoke the same methods over each of their components. The `Map` classes use these methods, not reference equality, on their keys.

Twisol: We frequently construct new `TaskId` wrappers around the same `String` ID in the engine itself -- see [this line](https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java#L249). If it didn't work this way, tasks waiting on their children would never awaken.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+
+  private void initSimulation(){
+    actToId = new HashMap<>();
+    lastSimResults = null;
+    this.engine = new SimulationEngine();
+    activitiesInserted.clear();
+
+    /* The top-level simulation timeline. */
+    this.timeline = new TemporalEventSource();
+    areLastSimResultsDirty = false;
```
Twisol: Can we reset `lastSimResultsEnd` here as well? It happens to not matter the way it's currently used, but I'd like to make sure there's no spurious state left over from previous runs for future changes to trip over.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+          return TaskStatus.delayed(delta);
+        }
+
+        this.scheduledTasks.remove();
+
+        final var directiveId = nextTask.getMiddle();
+        final var specification = nextTask.getRight();
+
+        final var id = scheduler.spawn(specification.getTypeName(), specification.getParameters());
+        actToId.put(nextTask.getMiddle(), id);
```
Twisol: `getMiddle()` is already saved as a (more helpfully-named) variable, `directiveId`:
```suggestion
        actToId.put(directiveId, id);
```
In particular, this makes it much more obvious that `actToId` is the inverse of `taskToPlannedDirective`, which suggests we might want to rename it to `plannedDirectiveToTask` instead.

adrienmaillard: I'll replace the `getMiddle()` by `directiveId` but I find that `plannedDirective` makes little sense in the context of the incremental simulation (`actToId` is part of the `IncrementalSimulationDriver`), when it is really activity ids (coming originally from the scheduler) to task id (issued by the engine), `actIdToTaskId`?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
adrienmaillard: My issue was coming from something else then, thank you for the explanations.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
Twisol: Whatever terminology is settled on, I'd mostly just like the `xToY` and `yToX` pairing to be upheld. That means you might want/need to rename the existing `taskToPlannedDirective` map to match your intended convention.

Twisol: Also, the difference between "activity" and "directive" is simply that "directive" applies to both the activity modeling and sequence/command modeling domains, while "activity" is particular. Since this class is owned by activity-specific scheduling domain, I don't see a problem with using "activity", but I do think the `IncrementalSimulationEngine` could very easily be lifted out of the scheduler in the future -- it's not truly activity-specific on its own, it just caters to use-cases surfaced by the scheduler.

Twisol: (As a concrete suggestion, I'd be in favor of "activityToTask" and "taskToActivity". I tend to be less a fan of shortened words like "act", although it's unambiguous in this case.)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+   * The provided simulation results might cover more than the required time period.
+   * @return the simulation results
+   */  public SimulationResults getSimulationResultsUntil(Duration endTime){
+    //if previous results cover a bigger period, we return do not regenerate
+    if(areLastSimResultsDirty || lastSimResults == null || endTime.longerThan(lastSimResultsEnd)) {
+      lastSimResults = engine.computeResults(
+          engine,
+          Instant.now(),
+          endTime,
+          new HashMap<>(),
```
Twisol: I'm a little concerned about this empty `HashMap`. This map is used by `computeResults` to replace engine-internal task IDs with the activity IDs sourced from the original plan. Normally we pass `controlTask.extractTaskToPlannedDirective()` here, but this class doesn't hold onto the `controlTask`s it constructs.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -503,6 +503,23 @@ public SimulationResults computeResults(
                                  serializedTimeline);
   }
 
+  public Duration getTaskDuration(String id){
```
Twisol: If we're able to figure out whatever was [the problem here](https://github.com/NASA-AMMOS/aerie/pull/23#discussion_r783570163), this would be better to take a `TaskId`. The use of `String` as the concrete ID type is mostly an implementation detail (modulo some unfortunate boundary-crossing with the return type of `spawn()`). In particular, the various `initiate` methods on `SimulationEngine` _return_ a `TaskId`, so we should expect the same type back as a parameter here.

(You're right -- the search under this line really is very unfortunate.)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+  private MissionModel<?> missionModel;
+
+  //mapping each activity name to its task id (in String form) in the simulation engine
+  private Map<String, String> actToId;
+
+  //boolean stating whether the simulation results should be regenerated
+  private boolean areLastSimResultsDirty;
+
+  //simulation results so far
+  private SimulationResults lastSimResults;
```
Twisol: If I read the implementation right, we can remove `areLastSimResultsDirty` in favor of setting `lastSimResults = null` whenever the cached results are invalidated. (That is, if the results are dirty, we'll never use them again anyway.)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+  //cached simulation results cover the period [Duration.ZERO, lastSimResultsEnd]
+  private Duration lastSimResultsEnd = Duration.ZERO;
+
+  //map from activity start time to serialized activity and its name
+  private List<SimulatedActivity> activitiesInserted = new ArrayList<>();
+
+  record SimulatedActivity(Duration start, SerializedActivity activity, String name) implements Comparable<SimulatedActivity> {
+    @Override
+    public int compareTo(@NotNull final SimulatedActivity o) {
```
Twisol: I'm not sure `SimulatedActivity` needs to implement `Comparable` anymore.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,247 @@
...
+  //boolean stating whether the simulation results should be regenerated
+  private boolean areLastSimResultsDirty;
+
+  //simulation results so far
+  private SimulationResults lastSimResults;
+  //cached simulation results cover the period [Duration.ZERO, lastSimResultsEnd]
+  private Duration lastSimResultsEnd = Duration.ZERO;
+
+  //map from activity start time to serialized activity and its name
+  private List<SimulatedActivity> activitiesInserted = new ArrayList<>();
```
Twisol: This comment is inaccurate now that this is a `List`. Aside from that, this comment mostly repeats what the code says (or had said, as a `Map`). It may be helpful to add more semantic information, e.g.:
```suggestion
  // The collection of activities scheduled so far.
  private List<SimulatedActivity> activitiesInserted = new ArrayList<>();
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
adrienmaillard: I have replaced `IllegalArgumentException` exceptions with `IllegalStateException`. By the way, do you prefer me to resolve these conversations when appropriate or you want to do it ?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: In the latest commit I have actually went back to the original directive/task naming and let reference to activities in a facade for the scheduler.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: I have replaced this empty map by the `taskToDirective` map now locally kept by the `IncrementalSimulationDriver`.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
adrienmaillard: You will see I have anticipated the change of `spawn` and directly used one of the `initiate` method to get the `TaskId`. Also got rid of the `ControlTask`.

adrienmaillard: And also replaced this `String` id by`TaskId`.

Twisol:
> By the way, do you prefer me to resolve these conversations when appropriate or you want to do it ?

Since I'm an external collaborator, I can't guarantee that I'll always be able to come back to these PRs in a timely fashion. It's probably best to treat these comments as advisory as a matter of process.

I'll always try to resolve proactively when I can. Unfortunately, at the moment I'm neck-deep in anxiety preparing a presentation to a research group, so I'm not going to be able to get back to these before the end of the week.

adrienmaillard:
Good luck on this presentation @Twisol. To other potential reviewers > I have adressed all the comments so far in my commits.

foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
kroffo: It doesn't by default, but I'm pretty sure you can set up your IDE preferences/settings to care and group things more appropriately!

adrienmaillard: I'll look into that, thanks.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/StepSimulationTest.java
adrienmaillard: The way activities are generated has changed.

kroffo:
> I realize that this started as a draft PR to share. It's great that it has had so much work and discussion. If we anticipate turning this into fully fledged PR at some point, we need to be mindful that a PR's commit history should show a logical build up of the feature/change being presented. As a result commits should be logically atomic steps in walking the reviewer to that end. For the most recent commit with message "Address draft PR comments", these changes should be redistributed through the commit history of this PR via a `git rebase -i` . @adrienmaillard I'll leave it to your judgement as to when and how you want to refactor the history of this branch to show the incremental build up of this change.

I second this. When you add commits changing things added/modified by earlier commits, those commits should be fixup commits to the commits you're "fixing", and then squashed into the commits they are fixing up with a rebase.

adrienmaillard:
I will massively squash then. There'll be three commits, one for the driver, one for the scheduling facade, one for the benchmark.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: This map has disappeared from the step simulation. No concern anymore.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/StepSimulation.java
adrienmaillard: This runs differently now, we let the engine schedule the whole plan itself (in a way very similar to the current `SimulationDriver`) so this problem has disappeared.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/TaskId.java
```diff
@@ -3,7 +3,7 @@
 import java.util.UUID;
 
 /** A typed wrapper for task IDs. */
-/*package-local*/ record TaskId(String id) {
+/*package-local*/ public record TaskId(String id) {
```
pcrosemurgy: Should remove the `/*package-local*/` comment here since it's now `public`. :)

scheduler/build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
   compileOnly project(':constraints')
   compileOnly project(':contrib')
   compileOnly project(':examples:banananation')
+  compileOnly project(':examples:foo-missionmodel')
```
pcrosemurgy: Are `banananation` and `foo-missionmodel` being used outside of `src/test/...`? If not, I'd remove line 55 & 56 here. From a quick grep it looks like they're only expected to be used within `src/test/...`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,182 @@
...
+  record SimulatedActivity(Duration start, SerializedActivity activity, ActivityInstanceId id) {}
+
+  public IncrementalSimulationDriver(MissionModel<?> missionModel){
+    this.missionModel = missionModel;
+    initSimulation();
+  }
+
+  private void initSimulation(){
+    plannedDirectiveToTask = new HashMap<>();
+    taskToPlannedDirective = new HashMap<>();
```
pcrosemurgy: It's good that these member variables are being assigned a value at construction time but without setting these directly within the class member variable declaration (lines 31 & 33 above) or within the class constructor we cannot guarantee that these fields are assigned and not null. Could you make these member variables final and either merge this assignment with their declarations or move them to the constructor? Thanks :)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class SimulationDriverFacade {
+
+  private final IncrementalSimulationDriver driver;
+
+  private final Map<String, ActivityInstanceId> activityNameToActivityId = new HashMap<>();
+
+  private long count = 0;
```
pcrosemurgy: Maybe this should be `static`, otherwise `simulateActivity()` may produce falsely-identified `ActivityInstanceId`s if `SimulationDriverFacade` is instantiated and used more than once.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
mattdailis: Hmm, `initSimulation` can get called multiple times over the lifetime of an `IncrementalSimulationDriver`.

Maybe we could initialize the hashmaps in the constructor, then call `.clear()` in `initSimulation`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class IncrementalSimulationDriver {
+
+  private Duration curTime;
+  private SimulationEngine engine;
+  private LiveCells cells;
+  private TemporalEventSource timeline;
```
pcrosemurgy: I'd suggest making these `final` and refactoring this class to use some sort of builder pattern.

These 4 members are initialized to `null` by default and there's an assumption that an assignment occurs before their first real use. That's usually not too complicated to get right in your head but when other people have to maintain that logic it's easy for `NullPointerException`s to arise out of forgetfulness.

pcrosemurgy: Basically I try to avoid "lazily assigned" variables (non-`final` and without a default) as much as possible.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
pcrosemurgy: Yeah, that'd make sense. Or some sort of builder pattern to rein in and consolidate the mutable member variables.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: These four are assigned each time we reset the simulation, which is not controlled by the user. I do not see how to make them final. Could you elaborate ?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
adrienmaillard: Will do.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
pcrosemurgy: Ahh I see. I'd at least change this to:
```java
  private Duration curTime = Duration.ZERO;
  private SimulationEngine engine = new SimulationEngine();
  private TemporalEventSource timeline = new TemporalEventSource();
```
to avoid `null` defaults. 

Another idea would be to make a `static` inner-class here called `Builder` to store all of the variables that are non-`final`. This way class members are at least organized by mutability/immutability and referring to any mutable variable would require qualifying with a `builder.*`. More of a way to signal to future maintainers that these variables are mutable rather than creating anything really enforceable, but better than no signal maybe.

mattdailis: Hmm... is anything being built here? I like your idea about grouping them in a static inner class, I'm just not sure I'd use the word "builder". Would something like this be similar to what you're describing, @pcrosemurgy?
```
record SimulationDetails(SimulationEngine engine, Duration curTime, TemporalEventSource timeline)
```

EDIT: I like your first suggestion as well - just making sure the fields have sensible defaults.

pcrosemurgy: Ahh true,`SimulationDetails` makes more sense. Either way, Adrien, this is all more of a stylistic thing, it doesn't give any new compile-time guarantees so I can definitely see an argument against it. That said it may be nice to have

adrienmaillard: I am not so convinced
- because `cells` needs a `TemporalSource` and the mission model, we cant' do anything about it
- we double initialize these variables when creating the object (default initializers + `initSimulation()`) 

Some classes are stateless, this class is very stateful and `initSimulation()` makes it clear and provides an initialization that has some meaning. 
 
I will put default initializers as you first suggested if that's fine.

pcrosemurgy: That's fair, thanks Adrien!

adrienmaillard: Thank you !

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,184 @@
...
+      var exit = true;
+      for(var taskId : taskToPlannedDirective.keySet()){
+        if(!engine.isTaskComplete(taskId)){
+          exit = false;
+          break;
+        }
+      }
+      if(exit){
+        break;
+      }
```
pcrosemurgy, line 163: Another stylistic change but I'd probably write this as:
```java
      // Exit IFF all tasks are complete
      if (taskToPlannedDirective.keySet().stream().allMatch(taskId -> engine.isTaskComplete(taskId))) {
        break;
      }
```

pcrosemurgy (approved):
This looks good to me. There are a few formatting changes I'd make before merge, but nothing that I consider to be blocking.

For example: 2 empty newlines, missing space between method sig. and curly braces (`public void someMethod(){`), missing space before & after `:` operator in for-each loops (`for(var act:acts)` in the benchmark test), missing space after control flow statements (`for(...)` instead of `for (...)`). All mostly aesthetic but it's generally good to have a consistent style, even if everyone doesn't agree on the exact style.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -524,6 +524,14 @@ public SimulationResults computeResults(
                                  serializedTimeline);
   }
 
+  public Duration getTaskDuration(TaskId taskId){
+    var state = tasks.get(taskId);
```
goetzrrGit: `final var state = task.get(taskId);`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,184 @@
...
+    {
+      final var daemon = engine.initiateTaskFromSource(missionModel::getDaemon);
+      final var commit = engine.performJobs(Set.of(SimulationEngine.JobId.forTask(daemon)),
+                                            cells, curTime, Duration.MAX_VALUE, missionModel);
+      timeline.add(commit);
+    }
+  }
+
+  public void simulateActivity(SerializedActivity activity, Duration startTime, ActivityInstanceId activityId){
+    var activityToSimulate = new SimulatedActivity(startTime, activity, activityId);
```
goetzrrGit: `final var activityToSimulate = new SimulatedActivity(startTime, activity, activityId);`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,184 @@
...
+    if(startTime.noLongerThan(curTime)){
+      var toBeInserted = new ArrayList<>(activitiesInserted);
```
goetzrrGit: `final var toBeInserted = new ArrayList<>(activitiesInserted);`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Get the simulation results from the Duration.ZERO to a specified end time point.
+   * The provided simulation results might cover more than the required time period.
+   * @return the simulation results
+   */
+  public SimulationResults getSimulationResultsUntil(Duration endTime){
+    //if previous results cover a bigger period, we return do not regenerate
+    if(lastSimResults == null || endTime.longerThan(lastSimResultsEnd)) {
+      Map<String, ActivityInstanceId> convertedTaskToPlannedDir = new HashMap<>();
```
goetzrrGit: `final  Map<String, ActivityInstanceId> convertedTaskToPlannedDir = new HashMap<>();`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
```diff
@@ -0,0 +1,42 @@
...
+  private final Map<String, ActivityInstanceId> activityNameToActivityId = new HashMap<>();
+
+  private static long count = 0;
+
+  public SimulationDriverFacade(IncrementalSimulationDriver driver){
+    this.driver = driver;
+  }
+
+  public void simulateActivity(SerializedActivity activity, String nameAct, Duration startTime){
+    var activityId = new ActivityInstanceId(count++);
```
goetzrrGit: `final var activityId = new ActivityInstanceId(count++);`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationDriverFacade.java
```diff
@@ -0,0 +1,42 @@
...
+    activityNameToActivityId.put(nameAct, activityId);
+    driver.simulateActivity(activity,startTime, activityId);
+  }
+
+  public Duration getActivityDuration(String activityName){
+    var activityId = activityNameToActivityId.get(activityName);
```
goetzrrGit, line 30: `    final var activityId = activityNameToActivityId.get(activityName);`

goetzrrGit (approved):
Looks good just a minor style change by adding the `final` keyword to some variables.

mattdailis (approved):
I'm excited at the possibilities this opens up. We'll need a follow-up PR where we try to hook this up to the actual scheduling algorithm, and we'll need to run some performance and correctness tests then. But I think for a first cut this PR is ready to merge :+1:

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: beautiful, I'll take it.

PR #24, opened 2022-01-11
[AERIE-1639] Plan ID Record

kroffo:
* **Tickets addressed:** AERIE-1639
* **Review:** By file  

## Description
This branch introduces the `PlanId` record, analogous to the `ActivityInstanceId` record introduced in [the other PR (UPDATE)]. There should be no changes to the system from an external perspective, as this merely changes the mechanism by which Plan IDs are passed around inside Merlin.

## Verification
System successfully builds, and simulation of a Banananation plan with the parent activity ran successfully. Plan was then modified to ensure updates to the plan by ID were successful, and simulations re-run. Everything worked as expected. I would also like to wait for the go ahead from @camargo before merging to prevent the need for any hotfixes as was found with the `ActivityInstanceId` record PR.

## Documentation
All changes internal to the system, no documentation update required.

## Future work
Perhaps we'll do this again for mission model ID? We could also consider just doing that now and add it onto this PR.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/exceptions/NoSuchActivityInstanceException.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/exceptions/NoSuchActivityInstanceException.java
@@ -1,13 +1,14 @@
 package gov.nasa.jpl.aerie.merlin.server.exceptions;
 
 import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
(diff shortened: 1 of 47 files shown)


mattdailis (approved):

camargo (approved):
Tried this locally against the UI and it looks good!
Opened https://github.com/NASA-AMMOS/aerie-ui/pull/1 and will merge after this is merged to account for the simulate function accepting a number now instead of a string.

pcrosemurgy (approved):

PR #25, opened 2022-01-12
[AERIE-1629] Computed Attributes in Simulation Results

mattdailis:
* **Tickets addressed:** AERIE-1629
* **Review:** By commit  

## Description
I've taken #17 and split it into three branches. I'll PR them sequentially, so we can focus on one at a time.

EDIT: If it would help declutter, I can make a separate PR for the refactoring commits. Let me know if anyone would prefer that.

The focus of this PR is reporting a value as part of simulation results upon completion of an activity instance.

The commits in this PR fall into one of the following phases
1. Preliminary refactoring (first 5 commits)
1. Use Supplier instead of Runnable to represent an effect model
1. Generate mapper and supplier for EffectModels with non-void return types
1. Reporting:
    - Include ValueSchema in ActivityType
    - Include return value in SimulatedActivity
    - Persist all of that in postgres

We continue to support `void` return types as before. These activities will have a `SerializedValue.UNIT` return value.

## Verification

I've done some simple manual testing by giving BiteBanana a non-void return type and viewing it via hasura. Some more thorough verification is needed prior to merging.

- [ ] more thorough testing. (unit test?)

## Documentation

This work is part of the Activity Computed Attributes working group: https://wiki.jpl.nasa.gov/display/MPSA/Working+Group%3A+Activity+Computed+Attributes

- [ ] We should update some user-facing documentation to explain how to use this feature.

## Future work

Step 2: (AERIE-1630) Make this value available to the caller
Step 3. (AERIE-1631) Tag events with activity ids

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
@@ -69,7 +69,7 @@ public final class MissionModel<Model> {
       @Override
       public TaskStatus step(final Scheduler scheduler) {
         MissionModel.this.daemons.forEach(daemon -> scheduler.spawn(daemon.create()));
-        return TaskStatus.completed();
+        return TaskStatus.completed(Unit.UNIT);
       }
 
       @Override
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
@@ -3,7 +3,6 @@ package gov.nasa.jpl.aerie.merlin.driver;
 import gov.nasa.jpl.aerie.merlin.driver.engine.Directive;
 import gov.nasa.jpl.aerie.merlin.driver.timeline.LiveCells;
 import gov.nasa.jpl.aerie.merlin.protocol.driver.Initializer;
-import gov.nasa.jpl.aerie.merlin.protocol.driver.Query;
 import gov.nasa.jpl.aerie.merlin.protocol.driver.Scheduler;
 import gov.nasa.jpl.aerie.merlin.protocol.model.ConfigurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.model.Resource;
@@ -12,8 +11,6 @@ import gov.nasa.jpl.aerie.merlin.protocol.model.TaskSpecType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.TaskStatus;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
-import org.apache.commons.lang3.tuple.Pair;
-import org.apache.commons.lang3.tuple.Triple;
 
 import java.util.Collections;
 import java.util.List;
@@ -64,7 +61,7 @@ public final class MissionModel<Model> {
   public Directive<Model, ?> instantiateDirective(final SerializedActivity specification)
   throws TaskSpecType.UnconstructableTaskSpecException
   {
-    return Directive.instantiate(this.taskSpecTypes.get(specification.getTypeName()), specification.getParameters());
+    return Directive.instantiate(this.taskSpecTypes.get(specification.getTypeName()), specification.getArguments());
   }
 
   public Task getDaemon() {
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
@@ -24,18 +24,18 @@ public final class MissionModel<Model> {
   private final LiveCells initialCells;
   private final Map<String, Resource<?>> resources;
   private final List<SerializableTopic<?>> topics;
-  private final Map<String, TaskSpecType<Model, ?>> taskSpecTypes;
+  private final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes;
   private final Optional<ConfigurationType<?>> configurationType;
-  private final List<Initializer.TaskFactory> daemons;
+  private final List<Initializer.TaskFactory<?>> daemons;
 
   public MissionModel(
       final Model model,
       final LiveCells initialCells,
       final Map<String, Resource<?>> resources,
       final List<SerializableTopic<?>> topics,
-      final List<Initializer.TaskFactory> daemons,
+      final List<Initializer.TaskFactory<?>> daemons,
       final Optional<ConfigurationType<?>> configurationType,
-      final Map<String, TaskSpecType<Model, ?>> taskSpecTypes)
+      final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes)
   {
     this.model = Objects.requireNonNull(model);
     this.initialCells = Objects.requireNonNull(initialCells);
@@ -54,20 +54,20 @@ public final class MissionModel<Model> {
     return this.configurationType;
   }
 
-  public Map<String, TaskSpecType<Model, ?>> getTaskSpecificationTypes() {
+  public Map<String, TaskSpecType<Model, ?, ?>> getTaskSpecificationTypes() {
     return this.taskSpecTypes;
   }
 
-  public Directive<Model, ?> instantiateDirective(final SerializedActivity specification)
+  public Directive<Model, ?, ?> instantiateDirective(final SerializedActivity specification)
   throws TaskSpecType.UnconstructableTaskSpecException
   {
     return Directive.instantiate(this.taskSpecTypes.get(specification.getTypeName()), specification.getArguments());
   }
 
-  public Task getDaemon() {
-    return new Task() {
+  public Task<Unit> getDaemon() {
+    return new Task<>() {
       @Override
-      public TaskStatus step(final Scheduler scheduler) {
+      public TaskStatus<Unit> step(final Scheduler scheduler) {
         MissionModel.this.daemons.forEach(daemon -> scheduler.spawn(daemon.create()));
         return TaskStatus.completed(Unit.UNIT);
       }
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelBuilder.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelBuilder.java
@@ -63,7 +63,7 @@ public final class MissionModelBuilder implements Initializer {
   }
 
   @Override
-  public String daemon(final TaskFactory task) {
+  public <Return> String daemon(final TaskFactory<Return> task) {
     return this.state.daemon(task);
   }
 
@@ -73,7 +73,7 @@ public final class MissionModelBuilder implements Initializer {
   }
 
   public <Model>
-  MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?>> taskSpecTypes) {
+  MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes) {
     return this.state.build(model, taskSpecTypes);
   }
 
@@ -81,14 +81,14 @@ public final class MissionModelBuilder implements Initializer {
     MissionModelBuilderState withConfigurationType(ConfigurationType<?> configurationType);
 
     <Model>
-    MissionModel<Model> build(Model model, Map<String, TaskSpecType<Model, ?>> taskSpecTypes);
+    MissionModel<Model> build(Model model, Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes);
   }
 
   private final class UnbuiltState implements MissionModelBuilderState {
     private final LiveCells initialCells = new LiveCells(new CausalEventSource());
 
     private final Map<String, Resource<?>> resources = new HashMap<>();
-    private final List<TaskFactory> daemons = new ArrayList<>();
+    private final List<TaskFactory<?>> daemons = new ArrayList<>();
     private final List<MissionModel.SerializableTopic<?>> topics = new ArrayList<>();
 
     private ConfigurationType<?> configurationType;
@@ -144,7 +144,7 @@ public final class MissionModelBuilder implements Initializer {
     }
 
     @Override
-    public String daemon(final TaskFactory task) {
+    public <Return> String daemon(final TaskFactory<Return> task) {
       this.daemons.add(task);
       return null;  // TODO: get some way to refer to the daemon task
     }
@@ -157,7 +157,7 @@ public final class MissionModelBuilder implements Initializer {
 
     @Override
     public <Model>
-    MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?>> taskSpecTypes) {
+    MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes) {
       final var missionModel = new MissionModel<>(
           model,
           this.initialCells,
@@ -208,7 +208,7 @@ public final class MissionModelBuilder implements Initializer {
     }
 
     @Override
-    public String daemon(final TaskFactory task) {
+    public <Return> String daemon(final TaskFactory<Return> task) {
       throw new IllegalStateException("Daemons cannot be added after the schema is built");
     }
 
@@ -219,7 +219,7 @@ public final class MissionModelBuilder implements Initializer {
 
     @Override
     public <Model>
-    MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?>> taskSpecTypes) {
+    MissionModel<Model> build(final Model model, final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes) {
       throw new IllegalStateException("Cannot build a builder multiple times");
     }
   }
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -81,8 +81,8 @@ public final class SimulationDriver {
     }
   }
 
-  public static <Model>
-  void simulateTask(final MissionModel<Model> missionModel, final Task task) {
+  public static <Model, Return>
+  void simulateTask(final MissionModel<Model> missionModel, final Task<Return> task) {
     try (final var engine = new SimulationEngine()) {
       /* The top-level simulation timeline. */
       var timeline = new TemporalEventSource();
@@ -128,7 +128,7 @@ public final class SimulationDriver {
     }
   }
 
-  private static final class ControlTask implements Task {
+  private static final class ControlTask implements Task<Unit> {
     private final Map<ActivityInstanceId, Pair<Duration, SerializedActivity>> schedule;
 
     /* The directive that caused a task (if any). */
@@ -152,7 +152,7 @@ public final class SimulationDriver {
     }
 
     @Override
-    public TaskStatus step(final Scheduler scheduler) {
+    public TaskStatus<Unit> step(final Scheduler scheduler) {
       while (true) {
         var nextTask = this.scheduledTasks.peek();
         if (nextTask == null) break;
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -169,7 +169,7 @@ public final class SimulationDriver {
         final var directiveId = nextTask.getMiddle();
         final var specification = nextTask.getRight();
 
-        final var id = scheduler.spawn(specification.getTypeName(), specification.getParameters());
+        final var id = scheduler.spawn(specification.getTypeName(), specification.getArguments());
         this.taskToPlannedDirective.put(id, directiveId);
       }
 
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -173,7 +173,7 @@ public final class SimulationDriver {
         this.taskToPlannedDirective.put(id, directiveId);
       }
 
-      return TaskStatus.completed();
+      return TaskStatus.completed(Unit.UNIT);
     }
 
     @Override
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -55,7 +55,7 @@ public final class SimulationEngine implements AutoCloseable {
   private final Subscriptions<Topic<?>, ResourceId> waitingResources = new Subscriptions<>();
 
   /** The execution state for every task. */
-  private final Map<TaskId, ExecutionState> tasks = new HashMap<>();
+  private final Map<TaskId, ExecutionState<?>> tasks = new HashMap<>();
   /** The getter for each tracked condition. */
   private final Map<ConditionId, Condition> conditions = new HashMap<>();
   /** The profiling state for each tracked resource. */
@@ -67,14 +67,14 @@ public final class SimulationEngine implements AutoCloseable {
   @DerivedFrom("taskParent")
   private final Map<TaskId, Set<TaskId>> taskChildren = new HashMap<>();
   /** The instantiated input provided to the task. Missing entries indicate tasks without input. */
-  private final Map<TaskId, Directive<?, ?>> taskDirective = new HashMap<>();
+  private final Map<TaskId, Directive<?, ?, ?>> taskDirective = new HashMap<>();
 
   /** Construct a task defined by the behavior of a model given a type and arguments. */
   public <Model>
   TaskId initiateTaskFromInput(final MissionModel<Model> model, final SerializedActivity input) {
     final var task = TaskId.generate();
 
-    final Directive<Model, ?> directive;
+    final Directive<Model, ?, ?> directive;
     try {
       directive = model.instantiateDirective(input);
     } catch (final TaskSpecType.UnconstructableTaskSpecException ex) {
@@ -84,23 +84,23 @@ public final class SimulationEngine implements AutoCloseable {
       return task;
     }
 
-    this.tasks.put(task, new ExecutionState.NotStarted(() -> directive.createTask(model.getModel())));
+    this.tasks.put(task, new ExecutionState.NotStarted<>(() -> directive.createTask(model.getModel())));
     this.taskDirective.put(task, directive);
 
     return task;
   }
 
   /** Define a task given a factory method from which that task can be constructed. */
-  public TaskId initiateTaskFromSource(final TaskSource source) {
+  public <Return> TaskId initiateTaskFromSource(final TaskSource<Return> source) {
     final var task = TaskId.generate();
-    this.tasks.put(task, new ExecutionState.NotStarted(source));
+    this.tasks.put(task, new ExecutionState.NotStarted<>(source));
     return task;
   }
 
   /** Define a task given a black-box task state. */
-  public TaskId initiateTask(final Duration startTime, final Task state) {
+  public <Return> TaskId initiateTask(final Duration startTime, final Task<Return> state) {
     final var task = TaskId.generate();
-    this.tasks.put(task, new ExecutionState.InProgress(startTime, state));
+    this.tasks.put(task, new ExecutionState.InProgress<>(startTime, state));
     return task;
   }
 
@@ -203,14 +203,24 @@ public final class SimulationEngine implements AutoCloseable {
     //   for putting an updated lifecycle back into the task set.
     var lifecycle = this.tasks.remove(task);
 
+    stepTaskHelper(task, frame, currentTime, model, lifecycle);
+  }
+
+  private <Return> void stepTaskHelper(
+      final TaskId task,
+      final TaskFrame<JobId> frame,
+      final Duration currentTime,
+      final MissionModel<?> model,
+      final ExecutionState<Return> lifecycle)
+  {
     // Extract the current modeling state.
-    if (lifecycle instanceof ExecutionState.IllegalSource) {
+    if (lifecycle instanceof ExecutionState.IllegalSource<Return>) {
       // pass -- uninstantiable tasks never progress or complete
-    } else if (lifecycle instanceof ExecutionState.NotStarted e) {
+    } else if (lifecycle instanceof ExecutionState.NotStarted<Return> e) {
       stepEffectModel(task, e.startedAt(currentTime), frame, currentTime, model);
-    } else if (lifecycle instanceof ExecutionState.InProgress e) {
+    } else if (lifecycle instanceof ExecutionState.InProgress<Return> e) {
       stepEffectModel(task, e, frame, currentTime, model);
-    } else if (lifecycle instanceof ExecutionState.AwaitingChildren e) {
+    } else if (lifecycle instanceof ExecutionState.AwaitingChildren<Return> e) {
       stepWaitingTask(task, e, frame, currentTime);
     } else {
       // TODO: Log this issue to somewhere more general than stderr.
@@ -219,9 +229,9 @@ public final class SimulationEngine implements AutoCloseable {
   }
 
   /** Make progress in a task by stepping its associated effect model forward. */
-  private void stepEffectModel(
+  private <Return> void stepEffectModel(
       final TaskId task,
-      final ExecutionState.InProgress progress,
+      final ExecutionState.InProgress<Return> progress,
       final TaskFrame<JobId> frame,
       final Duration currentTime,
       final MissionModel<?> model
@@ -234,16 +244,26 @@ public final class SimulationEngine implements AutoCloseable {
     // TODO: Report which topics this activity wrote to at this point in time. This is useful insight for any user.
     // TODO: Report which cells this activity read from at this point in time. This is useful insight for any user.
 
+    this.stepEffectModelHelper(task, progress, currentTime, state, status);
+  }
+
+  private <Return> void stepEffectModelHelper(
+      final TaskId task,
+      final ExecutionState.InProgress<Return> progress,
+      final Duration currentTime,
+      final Task<Return> state,
+      final TaskStatus<Return> status)
+  {
     // Based on the task's return status, update its execution state and schedule its resumption.
-    if (status instanceof TaskStatus.Completed s) {
+    if (status instanceof TaskStatus.Completed<Return> s) {
       final var children = new LinkedList<>(this.taskChildren.getOrDefault(task, Collections.emptySet()));
 
       this.tasks.put(task, progress.completedAt(currentTime, s.returnValue(), children));
       this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(currentTime));
-    } else if (status instanceof TaskStatus.Delayed s) {
+    } else if (status instanceof TaskStatus.Delayed<Return> s) {
       this.tasks.put(task, progress.continueWith(state));
       this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(currentTime.plus(s.delay())));
-    } else if (status instanceof TaskStatus.AwaitingTask s) {
+    } else if (status instanceof TaskStatus.AwaitingTask<Return> s) {
       this.tasks.put(task, progress.continueWith(state));
 
       final var target = new TaskId(s.target());
@@ -256,7 +276,7 @@ public final class SimulationEngine implements AutoCloseable {
       } else {
         this.waitingTasks.subscribeQuery(task, Set.of(SignalId.forTask(new TaskId(s.target()))));
       }
-    } else if (status instanceof TaskStatus.AwaitingCondition s) {
+    } else if (status instanceof TaskStatus.AwaitingCondition<Return> s) {
       final var condition = ConditionId.generate();
       this.conditions.put(condition, s.condition());
       this.scheduledJobs.schedule(JobId.forCondition(condition), SubInstant.Conditions.at(currentTime));
@@ -271,7 +291,7 @@ public final class SimulationEngine implements AutoCloseable {
   /** Make progress in a task by checking if all of the tasks it's waiting on have completed. */
   private <Return> void stepWaitingTask(
       final TaskId task,
-      final ExecutionState.AwaitingChildren<?> awaiting,
+      final ExecutionState.AwaitingChildren<Return> awaiting,
       final TaskFrame<JobId> frame,
       final Duration currentTime
   ) {
@@ -285,7 +305,7 @@ public final class SimulationEngine implements AutoCloseable {
       }
 
       final var nextChild = awaiting.remainingChildren().getFirst();
-      if (!(this.tasks.get(nextChild) instanceof ExecutionState.Terminated)) {
+      if (!(this.tasks.get(nextChild) instanceof ExecutionState.Terminated<?>)) {
         this.tasks.put(task, awaiting);
         this.waitingTasks.subscribeQuery(task, Set.of(SignalId.forTask(nextChild)));
         break;
@@ -444,7 +464,7 @@ public final class SimulationEngine implements AutoCloseable {
 
       final var activityId = taskToPlannedDirective.get(task.id());
 
-      if (state instanceof ExecutionState.Terminated e) {
+      if (state instanceof ExecutionState.Terminated<?> e) {
         simulatedActivities.put(activityId, new SimulatedActivity(
             directive.getType(),
             directive.getArguments(),
@@ -631,9 +651,9 @@ public final class SimulationEngine implements AutoCloseable {
     }
 
     @Override
-    public String spawn(final Task state) {
+    public <Return> String spawn(final Task<Return> state) {
       final var task = TaskId.generate();
-      SimulationEngine.this.tasks.put(task, new ExecutionState.InProgress(this.currentTime, state));
+      SimulationEngine.this.tasks.put(task, new ExecutionState.InProgress<>(this.currentTime, state));
       SimulationEngine.this.taskParent.put(task, this.activeTask);
       SimulationEngine.this.taskChildren.computeIfAbsent(this.activeTask, $ -> new HashSet<>()).add(task);
       this.frame.signal(JobId.forTask(task));
@@ -684,24 +704,24 @@ public final class SimulationEngine implements AutoCloseable {
   }
 
   /** The lifecycle stages every task passes through. */
-  private sealed interface ExecutionState {
+  private sealed interface ExecutionState<Return> {
     /** The task has an invalid source for its behavior. */
     // TODO: Provide more details about the instantiation failure.
-    record IllegalSource()
-        implements ExecutionState {}
+    record IllegalSource<Return>()
+        implements ExecutionState<Return> {}
 
     /** The task has not yet started. */
-    record NotStarted(TaskSource source)
-        implements ExecutionState
+    record NotStarted<Return>(TaskSource<Return> source)
+        implements ExecutionState<Return>
     {
-      public InProgress startedAt(final Duration startOffset) {
-        return new InProgress(startOffset, this.source.createTask());
+      public InProgress<Return> startedAt(final Duration startOffset) {
+        return new InProgress<>(startOffset, this.source.createTask());
       }
     }
 
     /** The task is in its primary operational phase. */
-    record InProgress(Duration startOffset, Task state)
-        implements ExecutionState
+    record InProgress<Return>(Duration startOffset, Task<Return> state)
+        implements ExecutionState<Return>
     {
       public <Return> AwaitingChildren<Return> completedAt(
           final Duration endOffset,
@@ -710,8 +730,8 @@ public final class SimulationEngine implements AutoCloseable {
         return new AwaitingChildren<>(this.startOffset, endOffset, returnValue, remainingChildren);
       }
 
-      public InProgress continueWith(final Task newState) {
-        return new InProgress(this.startOffset, newState);
+      public InProgress<Return> continueWith(final Task<Return> newState) {
+        return new InProgress<>(this.startOffset, newState);
       }
     }
 
@@ -721,7 +741,7 @@ public final class SimulationEngine implements AutoCloseable {
         Duration endOffset,
         Return returnValue,
         LinkedList<TaskId> remainingChildren
-    ) implements ExecutionState
+    ) implements ExecutionState<Return>
     {
       public Terminated<Return> joinedAt(final Duration joinOffset) {
         return new Terminated<>(this.startOffset, this.endOffset, joinOffset, this.returnValue);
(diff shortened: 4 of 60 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -152,11 +152,12 @@ public static JsonValue serializeSimulatedActivity(final SimulatedActivity simul
     return Json
         .createObjectBuilder()
         .add("type", simulatedActivity.type)
-        .add("parameters", serializeArgumentMap(simulatedActivity.parameters))
+        .add("parameters", serializeArgumentMap(simulatedActivity.arguments))
```
mattdailis: I stopped short of making this change, since it affects the API. I do think we'll want it though, unless there are any objections.

```suggestion
        .add("arguments", serializeArgumentMap(simulatedActivity.arguments))
```

Twisol: gosh, please, yes! @camargo is the arbiter here IMHO

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
```diff
@@ -11,9 +11,11 @@
...
-  String spawn(String type, Map<String, SerializedValue> arguments);
+  TaskIdentifier spawn(Task task);
+  TaskIdentifier spawn(String type, Map<String, SerializedValue> arguments);
 
-  String defer(Duration delay, Task task);
-  String defer(Duration delay, String type, Map<String, SerializedValue> arguments);
+  TaskIdentifier defer(Duration delay, Task task);
+  TaskIdentifier defer(Duration delay, String type, Map<String, SerializedValue> arguments);
+
+  interface TaskIdentifier {}
```
mattdailis: I called this `TaskIdentifier` so that I wouldn't have to use fully qualified names in the `SimulationEngine`. If anyone thinks `TaskId` (or any other name) would be better, happy to change it.

pcrosemurgy: Not to make more work for you but I'm on team `TaskId` :) This would follow the `ActivityInstanceId` and `PlanId` convention we have.

Twisol: Have there been any discussions yet on removing `waitFor`? If we did that, we could remove the need to return a task identifier in the `Scheduler` interface altogether.

Twisol: Meanwhile, I'm also on team `TaskId` -- qualified name collisions are the name of the game when it comes to complex component boundaries like this. You could call the engine's version an `EngineTaskId` or `ConcreteTaskId` instead, if you prefer.

(I can be rather easily swayed, especially if `TaskIdentifier` is on its way out due to removal of `waitFor`, anyway.)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
camargo: Yes I fully support this change. It will require a change in the UI here: https://github.com/NASA-AMMOS/aerie-ui/blob/develop/src/utilities/requests.ts#L1118
Notice how we are already changing the property name to `arguments` as that is what the rest of the UI expects :)

camargo: @mattdailis Let me know if you are going to change it and I will submit and link a quick PR

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
mattdailis: > Have there been any discussions yet on removing `waitFor`? If we did that, we could remove the need to return a task identifier in the `Scheduler` interface altogether.

Not yet - punting on that until "Step 2: (AERIE-1630) Make effect model return value available to the caller."

Sounds good - thanks, team TaskId 😉

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
mattdailis: @camargo awesome, I'll make the change now :+1:

camargo: Here is the PR: We can merge it once this PR is ready and merged: https://github.com/NASA-AMMOS/aerie-ui/pull/2

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
Twisol: I chatted with Pat a bit this evening, and threw together a branch for removing `waitFor` at `feature/remove-waitFor`. This would get rid of the need to return task identifiers at all, which is *lovely*.

We just need to make sure this won't adversely impact customers; Pat has kindly offered to do a cursory investigation on this front.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -512,14 +512,14 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
-                            .map(effectModel -> switch (effectModel.getRight()) {
+                            .map(effectModel -> switch (effectModel.executor()) {
                               case Threaded -> CodeBlock
                                   .builder()
                                   .addStatement(
                                       "return $T.threaded(() -> $L.$L($L.model())).create($L.executor())",
                                       gov.nasa.jpl.aerie.merlin.framework.ModelActions.class,
                                       "activity",
-                                      effectModel.getLeft(),
+                                      effectModel.methodName(),
```
goetzrrGit: I like this, It is more clear what right and left mean now.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -512,29 +512,21 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
-                                  .build();
-                            })
+                            .map(effectModel -> CodeBlock
+                                .builder()
+                                .addStatement(
+                                    "return $T.$L(() -> $L.$L($L.model())).create($L.executor())",
+                                    gov.nasa.jpl.aerie.merlin.framework.ModelActions.class,
+                                    switch(effectModel.executor()) {
+                                      case Threaded -> "threaded";
+                                      case Replaying -> "replaying";
```
goetzrrGit: I first I was trying to figure out why the switch statement was removed. Then I realized you compressed the switch statement here. :P

goetzrrGit (approved):
I am starting to understand more about this part of AERIE. It is all those walkthroughs you gave me :) Just a general question what are `activity computed attributes`? Visually I add `BiteBanana` to the plan, simulate, and get back values that affect the graph/resource at the bottom of the plan. Are those values you get back from the simulation `activity computed attributes`

Maybe we don't have an example yet?

I give this PR a loose `approve` someone with more experience can veto this though

mattdailis:
> I am starting to understand more about this part of AERIE. It is all those walkthroughs you gave me :)

🥳 yay, let's do more of those, for all the different things people are working on

> Just a general question what are `activity computed attributes`? Visually I add `BiteBanana` to the plan, simulate, and get back values that affect the graph/resource at the bottom of the plan. Are those values you get back from the simulation `activity computed attributes`

Yeah, sort of. The charter for our working group described it as:
> Currently, the only observable behaviors of an activity are its duration, decomposition, resources affected, and events emitted. This working group seeks to identify use cases for allowing an activity to have a more directly observable "output", and come to a consensus for how this new feature could best fit into Aerie.

So, what you currently see in the UI is `resources`. Activities emit `events`, which are interpreted into `effects`, some of which are serialized and output as `resource profiles` in simulation results. These resources are "global" in that there is one continuous "profile" that describes the value of that resource at any given time in the simulation.

There are some use cases for which this "global" property of resources is inconvenient - if you have multiple overlapping activities affecting the same resource, it can be difficult to discern the behavior of each individual activity.

The idea of "activity computed attributes" is that we are allowing an activity to return some arbitrary value upon completion. This value will be included in simulation results alongside other attributes of the activity, such as its input parameters, start time duration, etc.

> Maybe we don't have an example yet?

Correct - I haven't included an example in this PR, perhaps I should. I've been testing by going into `BiteBanana`, and changing the return type of `run` to something else (I usually do `boolean` for simplicity, but it can also be more complex). Then I recompile banananation and try running a simulation (either via the banananation test, or via the UI). The `returnValue` should be part of the simulation results, which you can look at through hasura.

Having written that up, I acknowledge that testing this PR is a bit of a pain 😅 . If I have some time later this week I'll see if I can add an example and make it more streamlined.

goetzrrGit:
@mattdailis thanks for the clarification. This helps a lot!!!

pcrosemurgy:
>more thorough testing. (unit test?)

Would be cool to test this out in an example adaptation like `foo-missionmodel`. `FooActivityTest` maybe?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -106,7 +106,7 @@ public MissionModelJar getMissionModelById(String id) throws NoSuchMissionModelE
     try {
       // TODO: [AERIE-1516] Teardown the missionModel after use to release any system resources (e.g. threads).
       return this.loadConfiguredMissionModel(missionModelId)
-                 .validateActivity(activityParameters.getTypeName(), activityParameters.getParameters());
+                 .validateActivity(activityParameters.getTypeName(), activityParameters.getArguments());
```
kroffo: It seems like the `activityParameters` argument should be renamed `activity`, and the method should be renamed `validateActivityArguments`, right?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/EffectModelRecord.java
```diff
@@ -0,0 +1,6 @@
+package gov.nasa.jpl.aerie.merlin.processor.metamodel;
+
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+
+public record EffectModelRecord(String methodName, ActivityType.Executor executor) {
```
kroffo: Why not just call it `EffectModel`?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/TaskSpecType.java
```diff
@@ -18,6 +20,8 @@ Specification instantiate(Map<String, SerializedValue> arguments)
   List<String> getValidationFailures(Specification taskSpec);
 
   Task createTask(Model model, Specification taskSpec);
+  ValueSchema getReturnValueSchema();
+  SerializedValue serializeReturnValue(Object returnValue);
```
pcrosemurgy: Thinking out loud: my initial reaction is to suggest changing this to
```java
<T> SerializedValue serializeReturnValue(T returnValue);
```
since I generally avoid using a bare `Object` parameter type. But I can't think of a good reason why `Object` isn't fine here.

Any chance you could run through the changes made in this PR sometime next week? Would be cool to go over the generated code that's implementing these methods.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/EffectModelRecord.java
pcrosemurgy: Whichever way we go I think we should stick to one convention. The `metamodel` package contains:
```
ActivityTypeRecord.java
ConfigurationTypeRecord.java
ExportDefaultsStyle.java
ExportTypeRecord.java
MapperRecord.java
MissionModelRecord.java
ParameterRecord.java
ParameterValidationRecord.java
TypeRule.java
```
so a `Record` suffix seems to be the norm. I'm kind of partial to this suffix since it indicates that the class is "meta", not a true `EffectModel` or `Mapper` for example.

pcrosemurgy: Another example is `ActivityType`, we already have 3 distinct `ActivityType`s:
```
./merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ActivityType.java
./scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityType.java
./merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/ActivityType.java
```
Which is okay but when possible it's nice to have a unique name.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/TaskSpecType.java
Twisol: Since return types are symmetric with parameter types in many ways, I'd expect to have a generic type parameter added to `TaskSpecType` specifically for the return type.

```java
public interface TaskSpecType<Model, Specification, Return> {
  // ...
  Task<Return> createTask(Model model; Specification taskSpec);
  // ...
  SerializedValue serializeReturnValue(Return returnValue);
}
```

```java
public interface Task<Return> {
  // ...
  TaskStatus<Return> step(Scheduler scheduler);
  // ...
}
```

```java
public interface TaskStatus<Return> {
  record Completed(Return returnValue) implements TaskStatus {}
  // ...
}
```

This type parameter would "infect" a number of other places, but it would more faithfully capture the specific ways we may obtain and use a task's return value. `Object` can be almost anything, so it provides no guidance as to where you ought to get a value to feed to `serializeReturnValue`. The generic `Return` is an identifier you can trace the usage of to get a complete picture of its flow through the system.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/EffectModelRecord.java
Twisol: I agree with @pcrosemurgy. For context, I think the term "record" was used by analogy with databases, as the annotation processor essentially builds up a knowledge base _about_ the model, then runs over all of the records it has collected (into tables!) and acts on that knowledge.

(That is, we didn't use "Record" simply because it was implemented as a Java `record`; that would have been completely unnecessary!)

kroffo: My main concern was indeed that we were using it simply because the class is a record, but I can jive with it being more representative of the fact that it's a record of something without being the actual something.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -195,6 +195,19 @@ public static ValueSchema ofVariant(final List<Variant> variants) {
...
+   * @return A new {@link ValueSchema} representing the Unit type.
+   */
+  public static ValueSchema ofUnit() {
+    return new ValueSchema() {
+      public <T> T match(final Visitor<T> visitor) {
+        return visitor.onVariant(List.of(new Variant("Unit", "UNIT")));
+      }
+    };
+  }
+
```
kroffo: My understanding of `Unit` is that it is a placeholder to fill the `Optional` returned from each `asType()` method. `Unit` itself isn't meaningful, so I'm surprised to see this change. I admittedly am not fully confident my understanding is whole and correct, but if it is correct then I'm not sure this change makes sense. Let me know what you're thinking here.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -503,6 +505,14 @@ public SimulationResults computeResults(
                                  serializedTimeline);
   }
 
+  @SuppressWarnings("unchecked")
+  private static <DirectiveReturn, TaskReturn> SerializedValue serializeReturnValue(
+      final Directive<?, ?, DirectiveReturn> directive,
+      final TaskReturn returnValue) {
+    // SAFETY: Tasks always return the same type as the TaskSpecType that declares them
+    return directive.directiveType().serializeReturnValue((DirectiveReturn) returnValue);
```
Twisol: I feel like there must be some way to make `DirectiveReturn` and `TaskReturn` manifestly the same type, i.e. make this method take one type parameter. 🤔 I might look at this more carefully later today.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -333,7 +335,14 @@ private ActivityMapperRecord getActivityMapper(final PackageElement missionModel
       final var executorAnnotation = element.getAnnotation(ActivityType.EffectModel.class);
       if (executorAnnotation == null) continue;
 
-      return Optional.of(new EffectModelRecord(element.getSimpleName().toString(), executorAnnotation.value()));
+      if (!(element instanceof ExecutableElement executableElement)) continue;
+
+      final var returnType = executableElement.getReturnType();
+      final Optional<TypeMirror> nonVoidReturnType = returnType.getKind() == TypeKind.VOID
+          ? Optional.empty()
+          : Optional.of(returnType);
```
kroffo: You should be able to use `var` here, I think by using `Optional.<TypeMirror>empty()`.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
```diff
@@ -17,6 +17,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
```
kroffo: This import seems to have been added accidentally.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ActivityType.java
```diff
@@ -1,6 +1,14 @@
 package gov.nasa.jpl.aerie.merlin.server.models;
 
 import gov.nasa.jpl.aerie.merlin.protocol.types.Parameter;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
+
 import java.util.List;
+import java.util.Optional;
```
kroffo: Accidentally added import?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
```diff
@@ -28,7 +29,9 @@
   public static final String NONEXISTENT_ACTIVITY_TYPE = "no-activity";
   public static final ActivityType EXISTENT_ACTIVITY = new ActivityType(
       EXISTENT_ACTIVITY_TYPE,
-      List.of(new Parameter("Param", ValueSchema.STRING)), List.of());
+      List.of(new Parameter("Param", ValueSchema.STRING)),
+      List.of(),
+      ValueSchema.UNIT);
```
kroffo: This makes me uncomfortable. We discussed earlier how we are sort of adding a second responsibility onto UNIT with these changes, and here using UNIT as an explicit activity effect model return type feels weird. I don't have a well thought out alternative or solution, but perhaps we need to continue/finish our discussion on UNIT.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
```diff
@@ -17,7 +17,6 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
```
kroffo: This was added in an earlier commit and should be removed there. I put a comment on it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -293,7 +293,8 @@ private static SimulationResults getSimulationResults(
             record.duration(),
             record.parentId().map(pgIdToSimId::get).orElse(null),
             record.childIds().stream().map(pgIdToSimId::get).collect(Collectors.toList()),
-            record.directiveId()
+            record.directiveId(),
+            SerializedValue.UNIT // TODO retrieve return value from database
```
kroffo: TODO before we merge, right?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ActivityAttributesRecord.java
```diff
@@ -0,0 +1,13 @@
...
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.util.Map;
+import java.util.Optional;
+
+public record ActivityAttributesRecord(
+    Optional<ActivityInstanceId> directiveId,
+    Map<String, SerializedValue> arguments,
+    SerializedValue returnValue
+) {}
```
kroffo: Nice addition!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateActivityTypeAction.java
```diff
@@ -1,18 +1,25 @@
...
   private static final @Language("SQL") String sql = """
-    insert into activity_type (model_id, name, parameters, required_parameters)
-    values (?, ?, ?::json, ?::json)
-    on conflict (model_id, name) do update set parameters = ?::json, required_parameters = ?::json
+    insert into activity_type (model_id, name, parameters, required_parameters, return_value_schema)
+    values (?, ?, ?::json, ?::json, ?::json)
+    on conflict (model_id, name) do update
+      set parameters = ?::json,
+      required_parameters = ?::json,
+      return_value_schema = ?::json
```
kroffo: I wanted to suggest you rework this so you don't have to specify the values a second time for these, but apparently it's not possible :(

https://stackoverflow.com/questions/13586012/using-same-prepared-statement-parameter-multiple-times-in-a-sql

kroffo (changes requested):
I put some comments for discussion, and brought up some minor issues to address. I would've put approval honestly, but I want to make sure I explicitly approve once we've discussed things, especially the UNIT thing.

pcrosemurgy (approved):
Matt gave a great walkthrough earlier. All my initial comments/questions are cleared up, this looks good!

Approving but I'm assuming you'll wrap up the documentation & verification with a short GraphQL blurb.

mattdailis:
Updates in the last few force pushes:
- Rebased onto develop
- Added a whole lot of generics (basically anything that has to do with Tasks has a `<Return>` generic now)
- Removed the commit to hide TaskId implementation from mission model. I'll happily make another PR for this one, but it was cluttering up this PR, and isn't actually related to this simulation-results-only view of computed attributes
- Split "Introduce ActivityAttributes record" into its own commit
- Added a commit to "Use jsonb instead of merlin_attribute_set for spans"
- Removed unused Optional imports as requested by Kenny
- Fixed a bug where I forgot to handle a TODO statement (great catch, Kenny)
- Used var instead of an explicit type where Kenny requested

Remaining tasks before merging:
- [ ] Test via hasura (and save the code snippet for documentation)
- [x] Figure out whether the use of ValueSchema.UNIT alongside ValueSchema.Unit is appropriate
- [ ] Address Jonathan's comment about statically enforcing that DirectiveReturnType matches TaskReturnType (I _think_ this may require making TaskIds generic, which I'm a bit hesitant to do)

Paul and I discussed unit testing, and decided that for this change, API testing would be more appropriate, so I plan to write up some instructions on how to manually test this via the hasura API.

patkenneally (approved):
Solid work!

Twisol:
> (I _think_ this may require making TaskIds generic, which I'm a bit hesitant to do)

If we merge #28 first, we can reduce a *lot* of the surface area that `TaskId` shows up in, which should make it much more palatable to generify.

patkenneally:
I'm testing #28 this afternoon, so we could get it in today.

mattdailis:
This morning's force push rebases onto develop - notably this was rebasing onto Paul's #20 , which involved a refactor of the annotation processing code.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: Yep, in the last commit.

pcrosemurgy:
@mattdailis changes made to `MissionModelGenerator` look good to me after your rebase. The generators could eventually use some more refactoring (`generateCommonMapperMethods()` in particular) but that's a separate issue.

mattdailis:
Updates over the past couple days:
- Rebased onto #28  
- Tweaked verbage: Activities have computed attributes, and Tasks have return values. Previously my code conflated the two concepts.
- Moved "Unit" out of the protocol and into the framework - from the Sim engine's perspective, all Tasks have return types, and it's up to the framework to decide how to represent the void return type.
- Since it may be visible to mission modelers, who may confuse the word "Unit" for a unit of measure, I've called the class "enum VoidEnum { VOID }" and included a javadoc comment linking to the wikipedia article on the Unit type. I still think this is better than using Java's capital V Void, because I want a type that has exactly one non-null value. Ideally I want any runtime casts to fail if VoidEnum is passed where it shouldn't be - in the case of null, these casts would quietly succeed.
- The mission model generator now uses EnumValueMapper(VoidEnum.class) to implement `getReturnValueSchema` and `serializeReturnValue`

I'll take another stab at eliminating the cast Jonathan mentioned, and then I think I'm ready to merge this. I'm getting tired of rebasing 😅

kroffo (approved):

Twisol:
> Since it may be visible to mission modelers, who may confuse the word "Unit" for a unit of measure, I've called the class "enum VoidEnum { VOID }" and included a javadoc comment linking to the wikipedia article on the Unit type. I still think this is better than using Java's capital V Void, because I want a type that has exactly one non-null value. Ideally I want any runtime casts to fail if VoidEnum is passed where it shouldn't be - in the case of null, these casts would quietly succeed.

This makes sense to me -- it's "void" in the sense of C/Java return types, not in the sense of type theory. ("Unit" is traditionally used in functional programming to mean the same thing; a return value that has no informational content, because it's always the same value. "Void" is traditionally used to mean a type with no inhabitants at all; it should be impossible to return from a function outputting "Void".)

Twisol:
That said, if there's even the slightest chance we'll need the type-theoretic "Void" at some point, we might want to proactively think about what we'd call it instead. It's not so unlikely -- type-theoretic "Unit" is the identity for product types, and type-theoretic "Void" is the identity for sum types, so in a situation where we want to allow variants to be encoded cleanly in a schema (which we currently only have for enums specifically), "Void" often comes up naturally.

Twisol:
~~By the way, it just occurred to me that a (Java-theoretic) "Void" return type ought to be equivalent to a `ValueSchema.ofStruct(Map.of())` i.e. a struct with no fields (an "empty product"). Is it possible, then, to avoid introducing "Void" entirely?~~

Ignore me! "Empty product" and "enum with one variant" are both perfectly legitimate and equivalent encodings of type-theoretic unit, and you've used the second one. We need *some* return type for activities, and isolating it to the framework as you've done makes perfect sense.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -152,7 +152,7 @@ public static JsonValue serializeSimulatedActivity(final SimulatedActivity simul
     return Json
         .createObjectBuilder()
         .add("type", simulatedActivity.type)
-        .add("parameters", serializeArgumentMap(simulatedActivity.arguments))
+        .add("arguments", serializeArgumentMap(simulatedActivity.arguments))
```
Twisol, line 155: I really appreciate that this API-facing change is separated out into its own commit 😄

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
```diff
@@ -43,28 +43,28 @@ public TaskStatus step(final Scheduler scheduler) {
       this.hostToTask.put(new TaskRequest.Resume(scheduler));
       final var response = this.taskToHost.take();
 
-      if (response instanceof TaskResponse.Success) {
-        final var status = ((TaskResponse.Success) response).status;
+      if (response instanceof TaskResponse.Success successResponse) {
```
Twisol: Personally, I tend to use very short names like `r` for the instanceof-introduced variables, because they tend to have very constraint scope. It also means that every branch gets the same name, which avoids introducing lots of names for what really is the same object, even as we interact with it differently in each branch.

In the future ([JEP 405](https://openjdk.java.net/jeps/405)), we'll be able to extract the fields of this records inline in the pattern, so even this name will go away. (Something like `response instanceof TaskResponse.Success(var status)`.)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -531,29 +531,20 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+                        .map(effectModel -> CodeBlock
                               .builder()
                               .addStatement(
-                                  "return $T.threaded(() -> $L.$L($L.model())).create($L.executor())",
+                                  "return $T.$L(() -> $L.$L($L.model())).create($L.executor())",
                                   gov.nasa.jpl.aerie.merlin.framework.ModelActions.class,
+                                  switch (effectModel.executor()) {
+                                    case Threaded -> "threaded";
+                                    case Replaying -> "replaying";
+                                  },
```
Twisol, line 621: oh yes, that's much nicer 👀

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -516,15 +516,13 @@ public Boolean onVariant(final List<Variant> variants) {
...
-    /** The user-facing presentation of this variant. */
-    public final String label;
-
-    public Variant(final String key, final String label) {
-      this.key = Objects.requireNonNull(key);
-      this.label = Objects.requireNonNull(label);
+  public static record Variant(
+      String key,  // The unique internal name of this variant
+      String label // The user-facing presentation of this variant
+  ) {
```
Twisol: I think I prefer the style you have here, but note that we could also use `@param` javadocs on the `Variant` itself:

```java
/**
 * @param key The unique internal name of this variant
 * @param label The user-facing presentation of this variant
 */
public static record Variant(String key, String label)
```

Technically speaking, this is *the* way to associate javadocs to these fields, if we want IDEs to recognize them as such.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -516,15 +516,13 @@ public Boolean onVariant(final List<Variant> variants) {
...
-  public static final class Variant {
-    /** The unique internal name of this variant. */
-    public final String key;
-    /** The user-facing presentation of this variant. */
-    public final String label;
-
-    public Variant(final String key, final String label) {
-      this.key = Objects.requireNonNull(key);
-      this.label = Objects.requireNonNull(label);
+  public static record Variant(
```
Twisol: Records are `static` by default 😁

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -516,15 +516,13 @@ public Boolean onVariant(final List<Variant> variants) {
...
+      String key,  // The unique internal name of this variant
+      String label // The user-facing presentation of this variant
+  ) {
+    public Variant {
+      Objects.requireNonNull(key);
+      Objects.requireNonNull(label);
```
Twisol: I'm much less bullish on keeping these `requireNonNull`s around these days, especially when the `record` could otherwise be a one-liner. They certainly don't hurt though 😅

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
```diff
@@ -46,8 +49,10 @@ public TaskStatus step(final Scheduler scheduler) {
       if (response instanceof TaskResponse.Success successResponse) {
         final var status = successResponse.status;
 
-        if (status instanceof TaskStatus.Completed) {
+        if (status instanceof TaskStatus.Completed completed) {
           this.lifecycle = Lifecycle.Terminated;
+          // SAFETY: This TaskStatus is constructed by this class, using the same ReturnType.
+          this.returnValue = (ReturnType) completed.returnValue();
```
Twisol: It's not clear how this cast is legal. The type of `TaskStatus.Completed#returnValue` is `TaskReturnValue<T>` for some `T`. By chasing the `SAFETY` clause, `T` is `ReturnType`, but this cast would require that `TaskReturnValue<ReturnType> = ReturnType`, which is unlikely?

(In case this is changed in later commits, I'm looking at 13b6ad2a4b076067f83bdc6f2d20c0d444828e2d.)

mattdailis: Yikes, that doesn't seem right, looks like I overlooked ThreadedTask on one of my many rebases. I'm surprised it didn't blow up during testing 🤔 

I don't quite remember why it's being saved in a field - I think for the purposes of this PR (return values in sim results only), there shouldn't be a need to cache it in a field. Taking a look now.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
```diff
@@ -137,15 +142,15 @@ public TaskResponse run(final TaskRequest request) {
             this);
 
         try (final var restore = ThreadedTask.this.rootContext.set(context)) {
-          ThreadedTask.this.task.run();
-          return new TaskResponse.Success(TaskStatus.completed());
+          ThreadedTask.this.returnValue = ThreadedTask.this.task.get();
+          return new TaskResponse.Success(TaskStatus.completed(ThreadedTask.this.returnValue));
         } catch (final TaskAbort ex) {
-          return new TaskResponse.Success(TaskStatus.completed());
+          return new TaskResponse.Success(TaskStatus.completed(VoidEnum.VOID));
```
Twisol: Now that we're providing return values on completion, it looks a little concerning to use `TaskStatus.completed()` when the task is being asked to *abort*. Perhaps we should add a separate `TaskResponse` variant for aborting without a return value.

More specifically, by potentially using either `VoidEnum` or whatever type the task *actually* returns, it becomes much less obvious that we're licensed to *extract* a particular type from the `Success` variant elsewhere

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/TaskReturnValue.java
```diff
@@ -0,0 +1,3 @@
+package gov.nasa.jpl.aerie.merlin.protocol.types;
+
+public record TaskReturnValue<T>(T returnValue) {}
```
Twisol: `TaskReturnValue<T>` seems to be equivalent to `T` itself. Is there a particular reason we want a wrapper here?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
```diff
@@ -69,7 +69,7 @@ public Task getDaemon() {
       @Override
       public TaskStatus step(final Scheduler scheduler) {
         MissionModel.this.daemons.forEach(daemon -> scheduler.spawn(daemon.create()));
-        return TaskStatus.completed();
+        return TaskStatus.completed(null);
```
Twisol: Why `null` and not `VoidEnum.VOID`?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -173,7 +173,7 @@ public TaskStatus step(final Scheduler scheduler) {
         this.taskToPlannedDirective.put(id, directiveId);
       }
 
-      return TaskStatus.completed();
+      return TaskStatus.completed(null);
```
Twisol: Why `null` and not `VoidEnum.VOID`?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -235,10 +236,10 @@ private void stepEffectModel(
     // TODO: Report which cells this activity read from at this point in time. This is useful insight for any user.
 
     // Based on the task's return status, update its execution state and schedule its resumption.
-    if (status instanceof TaskStatus.Completed) {
+    if (status instanceof TaskStatus.Completed completed) {
```
Twisol: Recalling my earlier comment, notice that other branches here just name the concrete variant `s`. At least for symmetry, we should probably name this one `s` as well. (Or change all the others, but I advocate in my earlier comment for `s`.)

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/TaskReturnValue.java
mattdailis: Hmm... I originally had a separate `Map<TaskId, TaskReturnValue<?>> taskReturns` map, and I didn't like the look of a `Map<TaskId, Object>`. However, now TaskReturnValue is folded into ExecutionState, so maybe it isn't adding much anymore.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -713,16 +717,24 @@ public InProgress continueWith(final Task newState) {
     }
 
     /** The task has completed its primary operation, but has unfinished children. */
-    record AwaitingChildren(Duration startOffset, Duration endOffset, LinkedList<TaskId> remainingChildren)
-        implements ExecutionState
+    record AwaitingChildren(
+        Duration startOffset,
+        Duration endOffset,
+        LinkedList<TaskId> remainingChildren,
+        TaskReturnValue<?> returnValue
```
Twisol: I know it's totally non-obvious, but the ordering the parameters here is meant to separate things into two groups: the "fixed" data that's merely been logged, and the "active" data that's modifed as the task state machine moves along. There's a lot more that could be done to make this obvious, but here I'd simply recommend making sure that `remainingChildren` is last.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
mattdailis: I moved VoidEnum.VOID into merlin-framework, where it's not accessible from the driver. Do you figure it would be worth moving it back into protocol? Or maybe defining a similar Unit enum internally for the driver's use? 🤔

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -713,16 +717,24 @@ public InProgress continueWith(final Task newState) {
...
+    ) implements ExecutionState
     {
-      public Terminated joinedAt(final Duration joinOffset) {
-        return new Terminated(this.startOffset, this.endOffset, joinOffset);
+      public Terminated joinedAt(final Duration joinOffset, final TaskReturnValue<?> returnValue) {
+        return new Terminated(this.startOffset, this.endOffset, joinOffset, returnValue);
```
Twisol: `AwaitingChildren` already has a `returnValue` field it can forward to the `Terminated` variant. Since `joinedAt` accepts a whole other `returnValue`, this field is (locally) unused. I imagine calling code extracts it out and then passes it back in, but this is unnecessary.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
Twisol: Good point, but yes, I'd avoid `null` wherever possible.

To be honest, I'm of the opinion that combining all of the daemons into one top-level task was actually a mistake born of convenience. Each daemon will automatically have its own return type anyway, and there's no good reason for us to group them up like this.

And `ControlTask` (the other place this happens) is already gone on one of my branches.

So yeah, I'd recommend adding a local `Unit` for the sake of avoiding `null`. No reason everybody has to use the same unit type.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
```diff
@@ -12,18 +13,38 @@ protected ModelActions() {}
   static final Scoped<Context> context = Scoped.create();
 
 
+  public static <T> Context.TaskFactory threaded(final Supplier<T> task) {
+    return executor -> new ThreadedTask<>(executor, ModelActions.context, task);
+  }
```
Twisol: In order to spawn a lambda that has a return value, we'll also need to know how to serialize/deserialize that return value. It's not clear where that information comes from here.

Given the title of this PR, I suspect it would be okay to defer supporting lambdas-with-returns to a later pass on the topic?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
```diff
@@ -12,18 +13,38 @@ protected ModelActions() {}
...
+  public static String spawn(final Runnable task) {
+    return spawn(supplierOfRunnable(task));
+  }
+
+  private static Supplier<VoidEnum> supplierOfRunnable(final Runnable task) {
+    return () -> {
+      task.run();
+      return VoidEnum.VOID;
+    };
+  }
```
Twisol: Personally, I'd be in favor of inlining this method at its call sites. I find `() -> { task.run(); return VoidEnum.VOID; }` more informative than `supplierOfRunnable`, without being *too* verbose.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelBuilder.java
```diff
@@ -63,7 +63,7 @@ public void resource(final String name, final Resource<?> resource) {
   }
 
   @Override
-  public String daemon(final TaskFactory task) {
+  public <Return> String daemon(final TaskFactory<Return> task) {
```
Twisol, line 66: You can use a wildcard here, too.

```suggestion
  public String daemon(final TaskFactory<?> task) {
```

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
mattdailis: By adding a generic to TaskResponse, this cast is no longer necessary :+1:

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/TaskReturnValue.java
mattdailis: TaskReturnValue has been eliminated :+1:

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModel.java
mattdailis: I added a local Unit :+1: The only `completed(null)` calls left are:
- ReplayingTask status field inits to `TaskStatus.completed(null)`
- ThreadedTask aborts with a `completed(null)`

These have to be null, since they must be values of the `Return` type, but the Supplier has not returned any value.

Future work: handle ThreadedTask aborting with a different variant of either TaskStatus or TaskResponse.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
mattdailis: Fixed :+1:

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Point taken 👍

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Reordered :+1:

Twisol:
I might not be able to re-review, so feel no need to wait on me 😃 If you've addressed what you think you need to address, 🚀

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelBuilder.java
mattdailis: I'll make a ticket for revisiting daemon tasks, particularly to replace the "lump into one parent task" strategy

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
mattdailis: Only tasks that come from TaskSpecTypes will be serialized in computeResults (by virtue of having a "Directive"). So, tasks defined by lambdas are supported, (and in the future their return types can be made available to the caller), without any need for serialization.

Now, if it turns out we do want to include lambda tasks in serialization results... that's another problem entirely, I think.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Future work

mattdailis:
Final force-push was a minor field-name change from "ActivityType.returnValueSchema" to "ActivityType.computedAttributesValueSchema".

Thanks for all the feedback and numerous iterations, everyone! I think we're in a significantly better place than we were when I first opened this. Looking forward to the possibilities this change opens up in the future 🚀 

Merging now.

PR #26, opened 2022-01-12
[AERIE-1605] Add plan_dataset trigger tests

dyst5422:
Add a test for each of the plan_dataset triggers in our SQL files

* **Tickets addressed:** AERIE-1605
* **Review:** By commit

## Description
Simple SQL manipulation based tests.

## Verification
N/A

## Documentation
N/A - minor code comments added to SQL files for clarity

## Future work
None (hopefully)

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
@@ -1,9 +1,11 @@
+package gov.nasa.jpl.aerie.database;
+
 import com.impossibl.postgres.jdbc.PGDataSource;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
-import org.junit.Assume;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
@@ -14,9 +16,13 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
+import java.time.Duration;
 import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+record PlanDatasetRecord(int plan_id, int dataset_id) {}
 
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
@@ -41,7 +47,7 @@ class DatabaseTests {
 
       // Handle the case where we cannot connect to postgres by skipping the tests
       final var errors = new String(proc.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
-      Assume.assumeFalse(errors.contains("Connection refused"));
+      Assumptions.assumeFalse(errors.contains("Connection refused"));
       proc.waitFor();
       proc.destroy();
     }
@@ -82,7 +88,7 @@ class DatabaseTests {
   // Teardown test database
   @AfterAll
   void afterAll() throws SQLException, IOException, InterruptedException {
-    Assume.assumeNotNull(connection);
+    Assumptions.assumeTrue(connection != null);
     connection.close();
 
     // Clear out all data from the database on test conclusion
@@ -126,7 +132,7 @@ class DatabaseTests {
     }
   }
 
-  int insertMissionModel(int fileId) throws SQLException {
+  int insertMissionModel(final int fileId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -151,15 +157,19 @@ class DatabaseTests {
     }
   }
 
-  int insertPlan(int missionModelId) throws SQLException {
+  int insertPlan(final int missionModelId) throws SQLException {
+    return insertPlan(missionModelId, "2020-1-1 00:00:00");
+  }
+
+  int insertPlan(final int missionModelId, final String start_time) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
               """
                   INSERT INTO plan (name, model_id, duration, start_time)
-                  VALUES ('test-plan-%s', '%s', '0', '2020-1-1 00:00:00')
+                  VALUES ('test-plan-%s', '%s', '0', '%s')
                   RETURNING id;"""
-                  .formatted(UUID.randomUUID().toString(), missionModelId)
+                  .formatted(UUID.randomUUID().toString(), missionModelId, start_time)
           );
       res.next();
       return res.getInt("id");
@@ -176,7 +186,7 @@ class DatabaseTests {
     }
   }
 
-  int insertActivity(int planId) throws SQLException {
+  int insertActivity(final int planId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -202,7 +212,7 @@ class DatabaseTests {
     }
   }
 
-  int insertSimulationTemplate(int modelId) throws SQLException {
+  int insertSimulationTemplate(final int modelId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -227,7 +237,7 @@ class DatabaseTests {
     }
   }
 
-  int insertSimulation(int simulationTemplateId, int planId) throws SQLException {
+  int insertSimulation(final int simulationTemplateId, final int planId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -252,12 +262,62 @@ class DatabaseTests {
     }
   }
 
+  int insertDataset() throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      final var res = statement
+          .executeQuery(
+              """
+                  INSERT INTO dataset
+                  DEFAULT VALUES
+                  RETURNING id;"""
+          );
+      res.next();
+      return res.getInt("id");
+    }
+  }
+
+  void clearDatasets() throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      statement
+          .executeUpdate(
+              """
+                  TRUNCATE dataset CASCADE;"""
+          );
+    }
+  }
+
+  PlanDatasetRecord insertPlanDataset(final int planId) throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      final var res = statement
+          .executeQuery(
+              """
+                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
+                  VALUES ('%s', '0')
+                  RETURNING plan_id, dataset_id;"""
+                  .formatted(planId)
+          );
+      res.next();
+      return new PlanDatasetRecord(res.getInt("plan_id"), res.getInt("dataset_id"));
+    }
+  }
+
+  void clearPlanDatasets() throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      statement
+          .executeUpdate(
+              """
+                  TRUNCATE plan_dataset CASCADE;"""
+          );
+    }
+  }
+
   int fileId;
   int missionModelId;
   int planId;
   int activityId;
   int simulationTemplateId;
   int simulationId;
+  PlanDatasetRecord planDatasetRecord;
 
   @BeforeEach
   void beforeEach() throws SQLException {
@@ -267,6 +327,7 @@ class DatabaseTests {
     activityId = insertActivity(planId);
     simulationTemplateId = insertSimulationTemplate(missionModelId);
     simulationId = insertSimulation(simulationTemplateId, planId);
+    planDatasetRecord = insertPlanDataset(planId);
   }
 
   @AfterEach
@@ -277,6 +338,8 @@ class DatabaseTests {
     clearActivities();
     clearSimulationTemplates();
     clearSimulations();
+    clearDatasets();
+    clearPlanDatasets();
   }
 
   @Nested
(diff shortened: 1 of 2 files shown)


pcrosemurgy (approved):
Thanks Dylan, looks sane to me :)

db-tests/src/test/java/TestTriggers.java
```diff
@@ -2,6 +2,7 @@
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.junit.Assume;
+import org.apache.commons.lang3.tuple.Pair;
 import org.junit.jupiter.api.AfterAll;
```
Twisol: Imports are a little unordered here -- this `org.apache` is sandwiched between two `org.junit`.

dyst5422: Addressed

db-tests/src/test/java/TestTriggers.java
```diff
@@ -1,6 +1,7 @@
 import com.impossibl.postgres.jdbc.PGDataSource;
```
patkenneally, line 3: 🔶 I think the class and the file should be named the same DatabaseTests.java and DatabaseTests. Or if there's a more apt naming that would work just as well.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -1,6 +1,7 @@
...
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
+import org.apache.commons.lang3.tuple.Pair;
 import org.junit.Assume;
```
patkenneally: 🔻 My IntelliJ is telling me that we should favor a static import of the classes we need from org.junit.jupiter.api.Assumptions rather than using the older org.junit.Assume. I think there are two instances in this PR.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
+                  VALUES (%s, '0')
+                  RETURNING dataset_id;"""
+                  .formatted(planId)
+          );
+      res.next();
+      final var newDatasetId = res.getInt("dataset_id");
+      res.close();
+
+      assertInstanceOf(Integer.class, newDatasetId);
```
patkenneally: 🔶 I'm interested to discuss this assertion choice. I may be missing something here but it seems like this is always true given the return type of the `res.getInt`. Could it be more descriptive to test something about the value like maybe check `newDatasetId > 0` or whatever we expect it to be?

patkenneally (dismissed):
Legend:
🔻 **Issues** to address before merge
🔶 **Requests** that should not block merge, but should at least be discussed
🔵 **Recommendations** that can be skipped if desired, but should be acknowledged.
&nbsp;&nbsp;&nbsp;&nbsp; **Observations** that don't even need to be acknowledged

🔶 Are we able to move these tests into a java package path, something like merlin-database/src/test/java/gov/nasa/jpl/aerie/database ?

db-tests/src/test/java/TestTriggers.java
kroffo: I second this ^

db-tests/build.gradle
```diff
@@ -13,6 +13,7 @@ test {
 }
 
 dependencies {
+  testImplementation 'org.apache.commons:commons-lang3:3.12.0'
```
kroffo: This dependency is solely so we can use `Pair` right? I think it would probably be better to just make a record in place of the pairs we are using, primarily for increased visibility into what the pairs are, and secondly to avoid increasing the size of dependencies.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -252,12 +259,63 @@ void clearSimulations() throws SQLException {
...
+      final var res = statement
+          .executeQuery(
+              """
+                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
+                  VALUES ('%s', '0')
+                  RETURNING plan_id, dataset_id;"""
+                  .formatted(planId)
+          );
+      res.next();
+      return Pair.of(res.getInt("plan_id"), res.getInt("dataset_id"));
```
kroffo: Could use a record like `PlanDatasetRecord` to represent these instead of a `Pair`. Looking at other modules such as `merlin-server`, we do use a `model/` package for such purposes in our testing, so we should do that here too.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -267,6 +325,8 @@ void beforeEach() throws SQLException {
     activityId = insertActivity(planId);
     simulationTemplateId = insertSimulationTemplate(missionModelId);
     simulationId = insertSimulation(simulationTemplateId, planId);
+    datasetId = insertDataset();
+    planDatasetId = insertPlanDataset(planId);
```
kroffo: Replacing the pair with a record would also make this field more clearly a record of data.

db-tests/src/test/java/TestTriggers.java
kroffo: A better check might be to call `assertFalse(res.wasNull());`

`res.getInt()` returns 0 if the value was null, which gets by the current check without issue: https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getInt(int)

However, we can check that the value `getInt()` determined it's return value from was not null using `res.wasNull()`: https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#wasNull()

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+          .executeQuery(
+              """
+                  SELECT * FROM dataset
+                  WHERE id = %s;"""
+                  .formatted(newDatasetId)
+          );
+
+      datasetRes.next();
+      assertEquals(newDatasetId, datasetRes.getInt("id"));
+      assertEquals(0, datasetRes.getInt("revision"));
```
kroffo, line 672: If we go with the `res.wasNull()` check on the ID above, we could do the same here. However, I also wonder if we should make it a separate test to check the revision. I think when we are testing more than one thing it indicates the presence of two tests in one. If one thing is wrong, we can't tell if it's truly one thing or both things that are broken.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+      final var newPlanRes = connection.createStatement()
+          .executeQuery(
+              """
+                  SELECT * from plan
+                  WHERE id = %s;"""
+                  .formatted(newPlanId)
+          );
+      newPlanRes.next();
+      final var newPlanStartTime = newPlanRes.getTimestamp("start_time");
+      newPlanRes.close();
```
kroffo: All of this is setting up initial conditions for the actual test which is that an inserted plan_dataset with a dataset_id should automatically compute the offset_from_plan_start.

It would be better to have the initial state before this state have two plans and one plan_dataset already in the database, and have the required information (start times, and offset) stored in variables so you don't have to read the database to get them. You're already doing this anyway with a bit of other information, it won't hurt to add some more initial state.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+                  INSERT INTO plan_dataset (plan_id, dataset_id)
+                  VALUES (%s, %s)
+                  RETURNING *;"""
+                  .formatted(newPlanId, planDatasetId.getRight())
+          );
+      planDatasetInsertRes.next();
+      final var newOffsetFromPlanStart = Duration.parse(planDatasetInsertRes.getString("offset_from_plan_start"));
+      planDatasetInsertRes.close();
+
+      assertEquals(offsetFromPlanStart.minus(Duration.ofMillis(newPlanStartTime.getTime() - planStartTime.getTime())), newOffsetFromPlanStart);
```
kroffo: Would you mind storing the result of the computation in a variable before this so that the assertion is more clear at a glance?

db-tests/src/test/java/TestTriggers.java
```diff
@@ -152,14 +155,18 @@ void clearMissionModels() throws SQLException {
   }
 
   int insertPlan(int missionModelId) throws SQLException {
+    return insertPlan(missionModelId, "2020-1-1 00:00:00");
+  }
+
+  int insertPlan(int missionModelId, String start_time) throws SQLException {
```
kroffo: Would you mind making all function parameters `final` unless there's a good reason not to? We try to default to `final` as much as possible unless there's a good reason not to. You can also set your editor up to flag this for you if you would like.

kroffo (changes requested):
Good work on the tests! Some points to address, but overall great stuff!

db-tests/src/test/java/TestTriggers.java
dyst5422: My thinking here was to just test that the fields of the dataset are the defaults - hence the revision assertion here.

db-tests/src/test/java/TestTriggers.java
dyst5422: Order of events here:
get currently existing plan start time
get the offset for the associated plan from the plan_dataset currently existing
insert a new plan with a different start time
grab the start time from the new plan
insert a new plan_dataset, which fires the trigger, and get the offset from it
assert the new offset is correct

Your suggestion is to add the second plan with the different time as part of test setup? I did it directly as part of this test since it's specific to this test and not really a shared setup. We can definitely have an additional beforeEach specific to this test, but that seemed like misdirection and overhead.

db-tests/build.gradle
dyst5422: We can do that. I was honestly surprised that tuple types weren't part of the language, but a bolt on class.

Twisol: Java was, and still is in some ways, a very restricted language. 😅 We've only recently got some more niceties.

db-tests/src/test/java/TestTriggers.java
kroffo: I wish we could have a `@Before` annotation that applies only to this specific test... I checked before I wrote my above comment and it's not included in the API we are using, so that's not an option.

Maybe you could just move the setup to a separate function just above this and call it with a comment explaining that the function just sets up initial state for the test? Then if a failure happens in the initialization, at least it's clear from the stack trace it's in the initialization function, and not the test itself.

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/build.gradle
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
kroffo: Sorry to be so picky (though at the same times, we _should_ hold ourselves to high standards) but the name `PlanDatasetId` for the record seems misleading, as it's not the ID of a PlanDataset, but rather a record of a PlanDataset, which is at its most basic form just two IDs, the association of a plan ID with a Dataset ID. If you have opposing thoughts I'm happy to hear you out, but I do think we should rename it to `PlanDatasetRecord` (to be clear, not because it's a Java record, but because it's a literal record of data).

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
```diff
@@ -252,12 +262,63 @@ void clearSimulations() throws SQLException {
...
+  }
+
   int fileId;
   int missionModelId;
   int planId;
   int activityId;
   int simulationTemplateId;
   int simulationId;
+  int datasetId;
+  PlanDatasetId planDatasetId;
```
kroffo: The datasetID should be included in the `PlanDatasetId` record, which is not just holding a plan ID (see comment about renaming the record). In fact, the field `datasetId` is never used throughout the tests, and can be removed (see comment on `beforeEach()` method)

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
```diff
@@ -267,6 +328,8 @@ void beforeEach() throws SQLException {
     activityId = insertActivity(planId);
     simulationTemplateId = insertSimulationTemplate(missionModelId);
     simulationId = insertSimulation(simulationTemplateId, planId);
+    datasetId = insertDataset();
+    planDatasetId = insertPlanDataset(planId);
```
kroffo: The insertion of a plan dataset automatically creates the dataset, so you shouldn't explicitly create a dataset unless you _also_ want one that isn't associated with the plan, but you aren't using it in your tests so I don't think that is the case.

db-tests/src/test/java/TestTriggers.java
kroffo: Resolving, but there are new comments to address in this area as well.

db-tests/src/test/java/TestTriggers.java
kroffo: Yeah, it doesn't seem like a big deal to me, just thinking about the principles more than the specific situation. I'm personally fine with checking the two things here, but if it ever does come up as an issue we should probably split it.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
dyst5422: Agreed. I'm going to leave the dataset insert method in there for possible future use cases, but just remove the call and assignment

db-tests/src/test/java/TestTriggers.java
dyst5422: I don't think the benefits outweigh the extra code necessary for that. Since we need the return value of a bunch of that "initialization", in addition to the new plan id, we'd have an additional method (not too bad here) and an additional record to get that data out of the scope of the function that really only exists to get that data out of the function - which is again a very specific initialization function for this particular test.

kroffo: Well, since there's no elegant solution here I'm fine with keeping it how it is, so I'll resolve.

kroffo (approved):

PR #28, opened 2022-01-16
Implement defer() in terms of other primitives

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
Back when `defer()` was introduced (which was very early, IIRC!), we didn't have the ability to spawn lambdas. Now we do, so `defer(t, a)` can simply be implemented as `spawn(() -> { delay(t); spawn(a); })`. This simplifies the model-engine interface, which needed to duplicate the `spawn` variants just for the extra `Duration` argument.

This technically costs an extra `Task` per `defer()` (since it's implemented with two `spawn`s). This cost is reduced on the model side by explicitly using a `ReplayingTask` for the delaying lambda. On the engine side, the extra task does have to be tracked, but ideally this overhead should be minimal. If we remove `waitFor` as well (see Future Work), we should be able to proactively prune completed task subtrees, and the number of tasks that need to be searched when propagating completion up the tree will also be reduced, so we should eventually see more wins than losses overall.

## Verification
I don't actually have access to any models that use `defer`; I'd like to request that somebody with access to the InSight model check that things come out the same. (Especially if there's any obvious change to performance.)

That said, semantically we've always intended that `defer` behave like a delay prefixing the spawned activity, so if this change breaks anything, there are deeper issues to be addressed.

## Documentation
This change does not impact any user-facing APIs or contracts.

## Future work
* I'd like to get rid of the `type, arguments` variant of `spawn` as well, since that incurs an extra serialization+deserialization that shouldn't *really* be necessary. That's not going to be quite as straightforward as removing `defer`, but at least this PR reduces the amount of spawn-related surface to cover in the first place.
* I'd also like to get rid of `waitFor`. I've discussed the rationale with @mattdailis and @patkenneally before, but the long-and-short is that `waitFor` interacts poorly with some desirable semantics around activity return values, specifically that we'd like `call(act)` to return the child activity's return value to its parent, or rethrow any uncaught exceptions the child activity threw. But, as noted above, since removing `waitFor` means that no model code can explicitly depend on the existence of a task, we can also eagerly prune completed tasks whose IDs have not been otherwise exposed from the engine. Since results computation still occurs at the end of simulation, we can't take full advantage of this yet -- we need to hold on to all tasks until results can be computed -- but when streaming results is implemented, we won't need to keep around any information not relevant to the ongoing simulation.

### DIFF
--- a/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
+++ b/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
@@ -45,27 +45,27 @@ public /*non-final*/ class ModelActions {
   }
 
   public static String defer(final Duration duration, final Runnable task) {
-    return defer(duration, threaded(task));
+    return spawn(replaying(() -> { delay(duration); spawn(task); }));
   }
 
   public static String defer(final Duration duration, final Context.TaskFactory task) {
-    return context.get().defer(duration, task);
+    return spawn(replaying(() -> { delay(duration); spawn(task); }));
   }
 
   public static String defer(final Duration duration, final String type, final Map<String, SerializedValue> arguments) {
-    return context.get().defer(duration, type, arguments);
+    return spawn(replaying(() -> { delay(duration); spawn(type, arguments); }));
   }
 
   public static String defer(final long quantity, final Duration unit, final Runnable task) {
-    return defer(unit.times(quantity), threaded(task));
-  }
-
-  public static String defer(final long quantity, final Duration unit, final String type, final Map<String, SerializedValue> arguments) {
-    return defer(unit.times(quantity), type, arguments);
+    return spawn(replaying(() -> { delay(quantity, unit); spawn(task); }));
   }
 
   public static String defer(final long quantity, final Duration unit, final Context.TaskFactory task) {
-    return context.get().defer(unit.times(quantity), task);
+    return spawn(replaying(() -> { delay(quantity, unit); spawn(task); }));
+  }
+
+  public static String defer(final long quantity, final Duration unit, final String type, final Map<String, SerializedValue> arguments) {
+    return spawn(replaying(() -> { delay(quantity, unit); spawn(type, arguments); }));
   }
 
 
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -650,31 +650,6 @@ public final class SimulationEngine implements AutoCloseable {
 
       return task.id();
     }
-
-    @Override
-    public String defer(final Duration delay, final Task state) {
-      if (delay.isNegative()) throw new IllegalArgumentException("Cannot schedule a task in the past");
-
-      final var task = TaskId.generate();
-      SimulationEngine.this.tasks.put(task, new ExecutionState.InProgress(this.currentTime, state));
-      SimulationEngine.this.taskParent.put(task, this.activeTask);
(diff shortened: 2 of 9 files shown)


pcrosemurgy (approved):
Thanks! This looks good to me.

Twisol:
Removed some freshly-unused imports I missed.

mattdailis (approved):
Looks reasonable to me :+1:

Twisol:
@patkenneally, is it alright if I merge this today? Want to make sure I'm not stepping on any delivery cycle considerations.

No rush in particular, though.

Twisol:
Actually, I'd still like to wait until somebody is able to sanity-check this against the InSight model. 🤔

patkenneally:
I've updated the Aerie Lander model (formerly known as the InSight model) to use the SNAPSHOT libraries from the Aerie github packages repo. I'll be able to run this branch against than in the next 48hrs. As you suggest, I agree let's hold off on merging this until we have that confirmation.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
```diff
@@ -45,27 +45,27 @@ public static void call(final String type, final Map<String, SerializedValue> ar
   }
 
   public static String defer(final Duration duration, final Runnable task) {
-    return defer(duration, threaded(task));
+    return spawn(replaying(() -> { delay(duration); spawn(task); }));
```
kroffo, line 48: If you recall, what was the motivation for `threaded(task)` in the old code? Also, what's the motivation to change to `replaying()` in the update?

kroffo (changes requested):
Before we merge I want to make sure I fully understand the implications of the changes here. I put a comment on ModelActions to discuss.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java
Twisol: It's certainly not clear from the diff, but the `threaded` didn't go away. The inner `spawn()` invokes `spawn(Runnable)`, which itself invokes `spawn(threaded(task))`.

The outer `spawn`, which is new, replaces the primitive call to `defer` with a primitive call to `spawn()` that simply causes a delay and then the inner spawn. Because it's so small, we want this outer `spawn` to use a replaying task to avoid thread-switching overhead.

Twisol: The motivation for `threaded` in the old code, which remains in the new code, is that replaying semantics is surprising to modelers, who are used to methods being run from start to finish. Replaying tasks rerun earlier parts of the task more than later parts, since they need to replay up to the point they last yielded at, so logging statements and such will give surprising results. So `threaded` is simply a sane, correct default for arbitrary tasks. Modelers can invoke any of these helpers with a `replaying(task)` wrapper to opt in to replaying behavior.

kroffo: Okay, after reading your responses and looking at the code in IntelliJ with the surrounding context and thinking about it for a few minutes, I get it and am on board. 

To restate my understanding (open for correction if I have any misconceptions), we use `threaded` tasks as a default to allow callers to spawn without having to worry about the task being created, but they can also explicitly set the task type if they wish by calling `threaded()` or `replaying()`. In the case of `defer()`, that is still the case, but we're wrapping it in a replaying task with a delay first. We _could_ use a threaded task for this, but we choose `replaying` simply because we expect better performance for this limited scenario by using replaying tasks.

kroffo (approved):

patkenneally:
I've tested this on the aerie lander repo and all looks in order.

PR #29, opened 2022-01-20
[HOTFIX] Fix scheduling simulation

adrienmaillard:
* **Review:** By commit  

## Description

Some recent changes made to the SimulationFacade broke the `SimulationFacade` interface with the `SimulationEngine`. Activity IDs are created on the fly but not saved and thus cannot be retrieved later when the user wants to query activity durations. I fix that by creating a structure, mapping from activity names used in scheduling to these activity ids.

## Verification

Verified with AerieLander tests. 

## Documentation

None

## Future work

None

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
@@ -71,6 +71,8 @@ public class SimulationFacade {
   //simulation results from the last simulation, as output directly by simulation driver
   private SimulationResults lastSimDriverResults;
 
+  private Map<String, ActivityInstanceId> activityNameToActivityId= new HashMap<>();;
+
   /**
    * Accessor for integer resource feeders
    *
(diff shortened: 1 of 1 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
```diff
@@ -71,6 +71,8 @@
   //simulation results from the last simulation, as output directly by simulation driver
   private SimulationResults lastSimDriverResults;
 
+  private Map<String, ActivityInstanceId> activityNameToActivityId;
```
pcrosemurgy: I'd make this:
```java
  private final Map<String, ActivityInstanceId> activityNameToActivityId = new HashMap<>();
```
to unify this declaration with its assignment (on line 184), that way you can sidestep any possible nullability runtime exceptions. Line 184 could be changed to `activityNameToActivityId.clear()` if this map is expected to be populated multiple times.

adrienmaillard: You're right, I have done that.

patkenneally (approved):
Approved in the interest of having a working scheduler. A subsequent ticket has been recorded for refactoring of the ActivityInstance class.

pcrosemurgy (approved):

PR #30, opened 2022-01-20
[AERIE-1603]  Simulation dataset trigger tests

dyst5422:
* **Tickets addressed:** AERIE-1603
* **Review:** By commit

## Description
Set up and then fire triggers and use select statements to assert the triggers properly fired

## Verification
Inspection and running the tests

## Documentation
Test addition only - no production code changes

## Future work
Need to ensure this is merged on top of some of the other test PR's so that we get those changes in as the base for this. Will need to rebase this prior to merge to bring in other PR comment changes.

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
@@ -23,12 +23,13 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+record SimulationDatasetRecord(int simulation_id, int dataset_id){}
 record PlanDatasetRecord(int plan_id, int dataset_id) {}
 
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class DatabaseTests {
-  private static final File initSqlScriptFile = new File("../deployment/postgres-init-db/sql/merlin/init.sql");
+  private static final File initSqlScriptFile = new File("../merlin-server/sql/merlin/init.sql");
   private java.sql.Connection connection;
 
   // Setup test database
@@ -238,7 +239,7 @@ class DatabaseTests {
     }
   }
 
-  int insertSimulation(final int simulationTemplateId, final int planId) throws SQLException {
+  int insertSimulationWithTemplateId(final int simulationTemplateId, final int planId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -247,6 +248,21 @@ class DatabaseTests {
                   VALUES ('%s', '%s', '{}')
                   RETURNING id;"""
                   .formatted(simulationTemplateId, planId)
+          );
+      res.next();
+      return res.getInt("id");
+    }
+  }
+
+  int insertSimulationWithoutTemplateId(final int planId) throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      final var res = statement
+          .executeQuery(
+              """
+                  INSERT INTO simulation (plan_id, arguments)
+                  VALUES ('%s', '{}')
+                  RETURNING id;"""
+                  .formatted(planId)
           );
       res.next();
       return res.getInt("id");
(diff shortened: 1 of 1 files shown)


db-tests/src/test/java/TestTriggers.java
```diff
@@ -252,12 +259,63 @@ void clearSimulations() throws SQLException {
...
+      final var res = statement
+          .executeQuery(
+              """
+                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
+                  VALUES ('%s', '0')
+                  RETURNING plan_id, dataset_id;"""
+                  .formatted(planId)
+          );
+      res.next();
+      return Pair.of(res.getInt("plan_id"), res.getInt("dataset_id"));
```
kroffo: We should use a record, mirroring the SimulationDatasetId record created for the existing tests. We don't need to include the Pair dependency, and the record gives more information about what is being returned, and what information is inside.

db-tests/build.gradle
```diff
@@ -13,6 +13,7 @@ test {
 }
 
 dependencies {
+  testImplementation 'org.apache.commons:commons-lang3:3.12.0'
```
kroffo: This dependency is not needed if we use a record for PlanDatasetId, as done for SimulationDatasetId.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
+                  VALUES (%s, '0')
+                  RETURNING dataset_id;"""
+                  .formatted(planId)
+          );
+      res.next();
+      final var newDatasetId = res.getInt("dataset_id");
+      res.close();
+
+      assertInstanceOf(Integer.class, newDatasetId);
```
kroffo: This is asserting that the value was not null, right? It would be more clear to use `assertNotNull()`, since `res.getInt()` returns an integer.

goetzrrGit (approved):
It ran for me and all test passed

patkenneally (approved):

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+          .executeQuery(
+              """
+                  SELECT * FROM plan_dataset
+                  WHERE plan_id = %s and dataset_id = %s;"""
+                  .formatted(planDatasetId.getLeft(), planDatasetId.getRight())
+          );
+      planDatasetSelectRes.next();
+      final var offsetFromPlanStart = Duration.parse(planDatasetSelectRes.getString("offset_from_plan_start"));
+      planDatasetSelectRes.close();
+      assertEquals(Duration.ofMillis(0), offsetFromPlanStart);
```
kroffo: This is asserting the initial state that is setup before each test, so if we needed to change the offset for any reason, this test would break here. This test shouldn't care if the offset is zero or not, but rather just use the offset it gets to calculate what the new plan_dataset's offset should be.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -577,4 +639,124 @@ void shouldIncrementSimulationRevisionOnSimulationUpdate() throws SQLException {
...
+                  INSERT INTO plan_dataset (plan_id, dataset_id)
+                  VALUES (%s, %s)
+                  RETURNING *;"""
+                  .formatted(newPlanId, planDatasetId.getRight())
+          );
+      planDatasetInsertRes.next();
+      final var newOffsetFromPlanStart = Duration.parse(planDatasetInsertRes.getString("offset_from_plan_start"));
+      planDatasetInsertRes.close();
+
+      assertEquals(offsetFromPlanStart.minus(Duration.ofMillis(newPlanStartTime.getTime() - planStartTime.getTime())), newOffsetFromPlanStart);
```
kroffo: May be better to perform the computations and store in named variables to make the purpose of this assertion more clear. This will also be beneficial in addressing the other comment on this test, as you can determine the dataset start time here.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -759,4 +761,200 @@ SELECT COUNT(*) FROM dataset
...
+                  AND table_name = 'event_%s'
+                );"""
+                .formatted(datasetId)
+        )
+        ) {
+          res.next();
+          assertTrue(res.getBoolean("exists"));
+        }
+
+        insertDataset();
```
kroffo: Was this a copy/paste error? Why is a dataset being inserted here?

kroffo (changes requested):
Some issues to address, but good work, the tests themselves are pretty good!

db-tests/src/test/java/TestTriggers.java
dyst5422: Assert conditions prior to insert, do insert, assert conditions after insert

dyst5422:
@kroffo can you go back over and ensure I addressed all of your concerns so we can get this merged in

kroffo (approved):

PR #31, opened 2022-01-21
[AERIE-1604] Simulation dataset cancel triggers tests

dyst5422:
* **Tickets addressed:** AERIE-1604
* **Review:** By commit

## Description
Tests setup the db, runs the action that causes the trigger, then checks that the desired changes were effected

## Verification
Inspection and running the tests (also did quick negative test checks)

## Documentation
None - Testing change only

## Future work
Need to rebase before merge

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTests.java
@@ -631,7 +631,8 @@ class DatabaseTests {
       connection.createStatement()
                 .executeUpdate(
                     """
-                        UPDATE simulation_template SET description = 'test-description-updated'
+                        UPDATE simulation_template
+                        SET description = 'test-description-updated'
                         WHERE id = %s;"""
                         .formatted(simulationTemplateId)
                 );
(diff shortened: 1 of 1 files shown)


db-tests/src/test/java/TestTriggers.java
```diff
@@ -2,6 +2,7 @@
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.junit.Assume;
```
patkenneally: My only comment for this PR 😆:  You could flip the order of these two imports and then you'd have a lovely alphabetically ordered list.

patkenneally (approved):

goetzrrGit (approved):
They all passed

db-tests/src/test/java/TestTriggers.java
```diff
@@ -1067,5 +1068,105 @@ SELECT COUNT(*)
         }
       }
     }
+
+    @Test
+    void shouldCancelOnMissionModelUpdate() throws SQLException {
```
kroffo: Should cancel what? Can you rename this to be more clear? Since all the trigger tests are in this one Java class, it's helpful to have very descriptive names as you've been doing in previous work. This applies to the tests below as well.

db-tests/src/test/java/TestTriggers.java
```diff
@@ -1067,5 +1068,105 @@ SELECT COUNT(*)
...
+      try (final var statement = connection.createStatement()) {
+        statement
+            .executeUpdate(
+                """
+                    UPDATE mission_model
+                    SET name = 'updated-name-%s'
+                    WHERE id = %s;"""
+                    .formatted(UUID.randomUUID().toString(), missionModelId)
```
kroffo: What was the motivation to generate a UUID rather than just using a hardcoded string for this test?

db-tests/src/test/java/TestTriggers.java
```diff
@@ -1067,5 +1068,105 @@ SELECT COUNT(*)
...
+
+        try (final var res = statement.executeQuery(
+            """
+                SELECT canceled
+                FROM simulation_dataset
+                WHERE simulation_id = %s;"""
+                .formatted(simulationDatasetId.simulation_id())
+        )) {
+          res.next();
+          assertTrue(res.getBoolean("canceled"));
```
kroffo: Hmmm. Should it be verified at the start of the test that canceled is false too, so that we know canceled was actually set to true by the update? This applies to the tests below as well.

kroffo (changes requested):
Some minor issues to discuss/address. I only reviewed the last commit, as the first three are on another PR.

db-tests/src/test/java/TestTriggers.java
dyst5422: Done

db-tests/src/test/java/TestTriggers.java
dyst5422: Removed Assume in favor of Assumptions

db-tests/src/test/java/TestTriggers.java
dyst5422: Theres a uniqueness constraint across mission name and version that I am avoiding here.

db-tests/src/test/java/TestTriggers.java
dyst5422: Did this on all these tests

kroffo (approved):

PR #35, opened 2022-02-02
[AERIE-1647] Refactor scheduler ActivityInstance

adrienmaillard:
* **Tickets addressed:** AERIE-1647 refactor scheduler activity instance
* **Review:** By commit  

## Description

There are three kinds of updates in this PR
(1) the refactor of ActivityInstance ids (scope of the ticket)
(2) the refactoring of how resource/parameters are typed in the scheduler 
(3) style/spring cleaning 

About (1) string attributes have been replaced with a statically generated incremental long id. These changes are in the first commit `259a583`. When looking more closely at how parameters have been modelled so far in activity instances, we saw the potential for a larger refactor.

(2) Before, activity parameters were understood as a `Map<String, Object>`. In this PR, we replace this with a `Map<String, SerializedValue>` which lead to bigger changes related to typing. Sometimes, activity instance parameters are dynamically computed at execution and thus now return `SerializedValue`. For that, the user can directly pass a resource and it will be queried at the appropriate time. Thus, scheduler resources that were typed before now return SerializedValue. The same idea has been applied throughout the scheduler code and led to some simplifications in the interface with the simulation as we do not have to manage different types of resources anymore. Also, the scheduler's ActivityType has been refactored to be linked with the mission model and allow that passed parameters (when the user is writing a rule) are correct with regard to the mission model definition. All of these changes are visible in commit`131ead2`. 

(3) Then, in commit `d4f2790`, `e4d618f`, and `be8d890`, a bunch of useless classes have been removed and a style pass has been performed to improve the quality of the code. The last commit is a small bugfix for the merlinsight/aerielander tests. 

## Verification

Scheduler tests have been modified with the new typing strategy and they pass. Merlinsight rules schedule the correct activities.  

## Documentation

None

## Future work

Refactor the MissionModelWrapper.


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PlanInMemory.java
```diff
@@ -70,16 +70,16 @@ public void add(ActivityInstance act) {
       throw new IllegalArgumentException(
           "adding activity with null start time to plan");
     }
-    final var name = act.getName();
+    final var name = act.getId();
```
patkenneally: 🔻 We should rename this variable to id or instanceId or something that conveys it's true usage.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
```diff
@@ -71,7 +71,7 @@
   //simulation results from the last simulation, as output directly by simulation driver
   private SimulationResults lastSimDriverResults;
 
-  private Map<String, ActivityInstanceId> activityNameToActivityId= new HashMap<>();;
+  private Map<Long, ActivityInstanceId> activityNameToActivityId= new HashMap<>();;
```
patkenneally: 🔶 Is this truly a map of name and id now that an instance no longer has a name field? It seems like this is more a map of plan activity ids to simulated driver ids. Also this line has two semi-colons.

patkenneally:
@adrienmaillard I'm in the middle of a review, but I was wondering if you wouldn't object to squashing commits 44e5814 and f1651a9 together and moving the resulting commit to be first in your commit history? this would group the class/file deletions up front, and make redundant a lot of the editing done in those files on account the changes in other commits. Also there is an edit in scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java in commit f1651a9 that should be extracted to sit with likewise changes in commit d40418c. As always I'm happy to help walk through a `git rebase -i` session with you if desired.

adrienmaillard:
> @adrienmaillard I'm in the middle of a review, but I was wondering if you wouldn't object to squashing commits [44e5814](https://github.com/NASA-AMMOS/aerie/commit/44e5814864bcd3e569e2412c8d1dd6a9918b284c) and [f1651a9](https://github.com/NASA-AMMOS/aerie/commit/f1651a952a9b7fe4f3346b1b07c1395146ee71fc) together and moving the resulting commit to be first in your commit history? this would group the class/file deletions up front, and make redundant a lot of the editing done in those files on account the changes in other commits. Also there is an edit in scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java in commit [f1651a9](https://github.com/NASA-AMMOS/aerie/commit/f1651a952a9b7fe4f3346b1b07c1395146ee71fc) that should be extracted to sit with likewise changes in commit [d40418c](https://github.com/NASA-AMMOS/aerie/commit/d40418c8908aff7904ff8b2bd863571d2758428b). As always I'm happy to help walk through a `git rebase -i` session with you if desired.

done

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
```diff
@@ -108,7 +106,6 @@ public Builder basedOn(ActivityCreationTemplate template) {
       endsIn = template.endRange;
       durationIn = template.durationRange;
       startsOrEndsIn = template.startOrEndRange;
-      nameMatches = (template.nameRE != null) ? template.nameRE.pattern() : null;
```
patkenneally, line 111: :large_orange_diamond: This change seems to belong in a later commit.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
```diff
@@ -126,8 +123,6 @@ protected ActivityCreationTemplate fill(ActivityCreationTemplate template) {
       template.startRange = startsIn;
       template.endRange = endsIn;
       template.startOrEndRange = startsOrEndsIn;
-      template.nameRE = (nameMatches != null)
```
patkenneally, line 129: :large_orange_diamond: This change seems to belong in a later commit.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplateDisjunction.java
```diff
@@ -103,7 +103,6 @@ OrBuilder basedOn(@NotNull ActivityCreationTemplateDisjunction template) {
       endsIn = template.endRange;
       durationIn = template.durationRange;
       startsOrEndsIn = template.startOrEndRange;
-      nameMatches = (template.nameRE != null) ? template.nameRE.pattern() : null;
       parameters = template.parameters;
```
patkenneally: :small_red_triangle_down: While were refactoring the `ActivityInstance`, we would benefit from bringing our usage of the terms parameter and argument in line with other Aerie components. In a separate commit,  interleaved into the branch history after this one, let's rename the `parameters` member variable and associated usages/accessor methods to use `argument`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityType.java
```diff
@@ -37,7 +37,7 @@ public ActivityType(String name) {
    *
    * @param name IN the identifier of the activity type
    */
-  public ActivityType(String name, TaskSpecType<?,?> specType) {
```
patkenneally: :small_red_triangle_down: If this change (and it's associated changes to the TaskSpecType type signature) is relocated to an earlier commit f917e5f then it will compile.

patkenneally (changes requested):

Some findings that don't relate to specific line of code. 

:small_red_triangle_down: db9b5d67b8e8e5161c2743bd86bf566fa962f043
- ~~Does not compile.~~
- ~~`final ActivityInstance act = new ActivityInstance(name, type);`~~

:small_red_triangle_down: f917e5fdb186e7b1b166bcafbfc9623f6473cd38
- ~~Does not compile~~.
- ~~It seems to me that this commit is two pieces of work. Can they be separated into 1. replace almost all notion of resource/parameter typing in the scheduler by SerializedValue/Parameter, and 2. Introduce variable parameters for activity instances? Having the two together makes it difficult to separate the interactions between co-mingled changes.~~

General comments:
- We need to make sure that each commit 
- compiles and run sensibly.
- contains only the code relating to the stated change in the commit message.
- commit message should be a concise sentence. Further description can be included in a separate paragraph below the initial summary. Please review this article (linked to in our CONTRIBUTING.md) for how to write a commit message https://cbea.ms/git-commit/

adrienmaillard:
> It seems to me that this commit is two pieces of work. Can they be separated into 1. replace almost all notion of resource/parameter typing in the scheduler by SerializedValue/Parameter, and 2. Introduce variable parameters for activity instances? Having the two together makes it difficult to separate the interactions between co-mingled changes.

I do not see how because when arguments become `SerializedValue`, variable parameters such as `StateQueryParam` need a place to sit (`Object` allowed both arguments and uninstantiated arguments).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PlanInMemory.java
adrienmaillard: It has been renamed id.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
adrienmaillard: It has been renamed `activityIdToActivityInstanceId`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
adrienmaillard: it has been moved to commit d113e2ea

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
adrienmaillard: it has been moved to commit d113e2ea

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplateDisjunction.java
adrienmaillard: commit d1433e10 does that. We only keep `parametricDuration` for now.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityType.java
adrienmaillard: done

patkenneally:
> > It seems to me that this commit is two pieces of work. Can they be separated into 1. replace almost all notion of resource/parameter typing in the scheduler by SerializedValue/Parameter, and 2. Introduce variable parameters for activity instances? Having the two together makes it difficult to separate the interactions between co-mingled changes.
> 
> I do not see how because when arguments become `SerializedValue`, variable parameters such as `StateQueryParam` need a place to sit (`Object` allowed both arguments and uninstantiated arguments).

That's true, you are right. My ordering was inverted. Rather, introduce variable parameters machinery, then use it to support the change to `SerializedValue`. I'll let you make the judgment on whether to make that distinction to this PRs story.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
patkenneally: Please do tell me I'm confused, if you think I am 😄. I think this name doesn't yet accurately communicate this map's relation. Looking at the code, it seems to me that it's planActInstanceIdToSimulationActInstanceId or something similar.

patkenneally:
Thanks for massaging the commits into a more cogent PR story. It helps a lot! I particularly liked that you took a commit and did a linting pass in c658066e23edfd4c9f4c227935aab19e2a0c3bf3 🤓

kroffo:
In the first commit a bunch of unused files are removed. I'm curious what the purpose of these files was to begin with, and why they are no longer needed? Bonus points if you add such a description to the commit itself (as a commit description, not a long commit message).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
kroffo: I am confused, because the commit you shared @adrienmaillard is _this_ commit. Was the change in a different commit and you moved it to this one? Perhaps Github moved this conversation to follow the change, if so?

adrienmaillard:
> In the first commit a bunch of unused files are removed. I'm curious what the purpose of these files was to begin with, and why they are no longer needed? Bonus points if you add such a description to the commit itself (as a commit description, not a long commit message).

This summer, an intern, Kevin (I'll keep family names out of here) worked on a way of generating these classes from a mission model so we don't have to write some other code (like `AerieController`). But we do not want to depend on a mission model, it was unclear at the time. Steve might have more details for you.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
mattdailis: ☝️ I am confused about this as well. In scope for this PR would be to make sure we're on the same page about whether we're talking about planned activity instances or simulated activity instances.

Out of scope for this PR: I think it would be helpful to follow up on Kenny's work to introduce ActivityInstanceIds, to further separate Planned Activity Instances vs Simulated Activity Instances throughout aerie. Verbose as that is, I think "Activity Instance" by itself is ambiguous, and muddles our ability to talk about what data is available at what parts of the workflow.

adrienmaillard: There is a third type of activity instances, the scheduling activity instances that are the subject of this PR. This container maps from scheduling activity instances ids to simulated activity instances ids. Do you want me to box the Activity Instance long id into a `SchedulingActivityInstanceId` to make that more clear ?

adrienmaillard: > Please do tell me I'm confused, if you think I am 😄. I think this name doesn't yet accurately communicate this map's relation. Looking at the code, it seems to me that it's planActInstanceIdToSimulationActInstanceId or something similar.

Yes I can do that.

mattdailis: I see… so there are really 3 different ID sources -
1. The scheduler autoincrements a counter for every ActivityInstance object it constructs
1. SimulationFacade generates ActivityInstanceIds for every simulation run
1. The simulation engine may assign new IDs to child activities that were not part of the plan

So yeah, I think it would be helpful to wrap the scheduler-specific id in a record :+1:

adrienmaillard: There is now a `SchedulingActivityInstanceId` record and the name of the map has been changed to `planActInstanceIdToSimulationActInstanceId`.

adrienmaillard:
> > > It seems to me that this commit is two pieces of work. Can they be separated into 1. replace almost all notion of resource/parameter typing in the scheduler by SerializedValue/Parameter, and 2. Introduce variable parameters for activity instances? Having the two together makes it difficult to separate the interactions between co-mingled changes.
> > 
> > 
> > I do not see how because when arguments become `SerializedValue`, variable parameters such as `StateQueryParam` need a place to sit (`Object` allowed both arguments and uninstantiated arguments).
> 
> That's true, you are right. My ordering was inverted. Rather, introduce variable parameters machinery, then use it to support the change to `SerializedValue`. I'll let you make the judgment on whether to make that distinction to this PRs story.

I ended up not doing that.

mattdailis (approved):
All of this looks like forward progress to me :+1:

Seeing all those deletions feels really good:
Screen Shot 2022-02-08 at 2 25 12 PM

pcrosemurgy (approved):

patkenneally (approved):

PR #39, opened 2022-02-08
[AERIE-1679] SQL tables to postgres for Command Expansions

goetzrrGit:
[AERIE-1679]

* **Tickets addressed:** AERIE-1679
* **Review:** By commit  

## Description
During the 'Command Dictionary WG' the team has come up with a data flow diagram for command expansions. The data models are defined in the document below and I converted the mock tables into SQL tables for Postgres. 

**Command Expansion Data Flow Diagram**
https://docs.google.com/drawings/d/1ZShnHzeWt7UVPVpVWSSYNPij7TWdtM33oMIxWMV7nbA/edit?usp=sharing

## Verification
~~I nuked everything in docker to verify that the tables were created in postgres.~~

~~- `docker system prune --all --volumes  `~~
~~- `./gradlew build; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up`~~
~~- Verify you see 'aerie-expansion` table in postgres and the tables defined in the SQL schema ( I used pgAdmin 4)~~

No verification as there is no service using this schema at the moment. I tested locally just to make sure the table were going to be created correctly in Postgres

## Documentation
Data flow diagram

## Future work
Work out the command expansion services.

### DIFF
--- a/deployment/postgres-init-db/init-aerie.sh
+++ b/deployment/postgres-init-db/init-aerie.sh
@@ -26,6 +26,7 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-E
   CREATE DATABASE aerie_ui;
   GRANT ALL PRIVILEGES ON DATABASE aerie_ui TO aerie;
   \echo 'Done!'
+
 EOSQL
 
 export PGPASSWORD=aerie
--- a/deployment/postgres-init-db/sql/expansion/tables/expansion_rules.sql
+++ b/deployment/postgres-init-db/sql/expansion/tables/expansion_rules.sql
@@ -1,0 +1,17 @@
+create table expansion_rules (
+  id integer generated always as identity,
+
+  activity_type text not null,
+  expansion_logic text not null,
+
+  constraint expansion_rules_primary_key
+  primary key (id)
+);
+comment on table expansion_rules is e''
+  'The user defined logic to expand an activity type.';
+comment on column expansion_rules.id is e''
+  'The synthetic identifier for this expansion rule.';
+comment on column expansion_rules.activity_type is e''
+  'The user selected activity type.';
+comment on column expansion_rules.expansion_logic is e''
+  'The expansion logic used to generate commands.';
--- a/deployment/postgres-init-db/sql/expansion/tables/expansion_set.sql
+++ b/deployment/postgres-init-db/sql/expansion/tables/expansion_set.sql
@@ -1,0 +1,21 @@
+create table expansion_set (
+  id integer generated always as identity,
+
+  command_dict_id integer not null,
+  mission_model_id integer not null,
+
+  constraint expansion_set_primary_key
+    primary key (id),
+
+  foreign key (command_dict_id)
+  references command_dictionary (id)
+);
+
+comment on table expansion_set is e''
+  'A binding of a command dictionary to a mission model.';
+comment on column expansion_set.id is e''
+  'The synthetic identifier for the set.';
+comment on column expansion_set.command_dict_id is e''
+  'The ID of a command dictionary.';
+comment on column expansion_set.mission_model_id is e''
+  'The ID of a mission model.';
--- a/deployment/postgres-init-db/sql/expansion/tables/expansion_set_to_rule.sql
+++ b/deployment/postgres-init-db/sql/expansion/tables/expansion_set_to_rule.sql
@@ -1,0 +1,19 @@
+create table expansion_set_to_rule (
+  set_id integer not null,
+  rule_id integer not null,
+
+  constraint expansion_set_to_rule_primary_key
+  primary key (set_id,rule_id),
+
+  foreign key (set_id)
+  references expansion_set (id),
+
+  foreign key (rule_id)
+  references expansion_rules (id)
+);
+comment on table expansion_set_to_rule is e''
+  'The join table between expansion_set and expansion_rules';
+comment on column expansion_set_to_rule.set_id is e''
+  'The id for an expansion_set.';
+comment on column expansion_set_to_rule.rule_id is e''
+  'the id for an expansion_rule.';
--- a/deployment/postgres-init-db/sql/expansion/init.sql
+++ b/deployment/postgres-init-db/sql/expansion/init.sql
@@ -1,0 +1,13 @@
+-- The order of inclusion is important! Tables referenced by foreign keys must be loaded before their dependants.
+
+begin;
+
+  -- Command Expansion Tables.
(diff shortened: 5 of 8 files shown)


deployment/postgres-init-db/sql/expansion/tables/expansion_rules.sql
```diff
@@ -0,0 +1,22 @@
+create table expansion_rules (
+  id integer generated always as identity,
+  activity_type text not null,
+  expansion_logic text not null,
+  expansion_set_id integer not null,
```
dyst5422: I think this is a many-to-many association as an expansion function could be associated with any number of expansion sets as command dictionaries and mission models evolve.

Or actually, we might not want the expansion set id here at all as we would be working in the opposite direction of expansion set to expansion rules and want the expansion set to have a list of ids of expansion rules

deployment/postgres-init-db/sql/expansion/tables/expansion_set.sql
```diff
@@ -0,0 +1,20 @@
+create table expansion_set (
+  id integer generated always as identity,
+  command_dict_id integer not null,
+  mission_model_id integer not null,
+
```
dyst5422, line 6: I think we probably want the expansion set to have references to each of the expansions in the set.

dyst5422 (changes requested):

deployment/postgres-init-db/sql/expansion/tables/expansion_set.sql
mattdailis: Sounds like we might want a join table to establish this relationship. Something like `expansion_set_membership` with two columns `expansion_set_id` and `expansion_rule_id`.

goetzrrGit: `expansion_set.sql` has this reference. Let me know if that is what you were looking for? This schema is the expansion configuration used when you expand activities in the plan. The `package/bundle` idea we have talked about in our meetings

deployment/postgres-init-db/sql/expansion/tables/expansion_rules.sql
goetzrrGit: I schedule a slack huddle to discuss this further with folks.

deployment/postgres-init-db/sql/expansion/tables/expansion_set.sql
goetzrrGit: I schedule a slack huddle to discuss this further with folks.

kroffo: Be sure to update this thread with the result of that huddle!

goetzrrGit: After the meeting, we went with the many-to-many approach between `expansion_set` and `expansion_rule` and will be using a join table (`expansion_set_to_rule`) to link these two tables together. This will allow for users to update the mission model or command dictionary and select which expansion logic to carry over to the new set.

deployment/postgres-init-db/sql/expansion/tables/expansion_set_to_rule.sql
```diff
@@ -0,0 +1,21 @@
+create table expansion_set_to_rule (
+  id integer generated always as identity,
```
kroffo: This ID is not necessary in this case, as the primary key is given by the combination of the set_id and rule_id. You got the primary key right below though, so just remove this and it will be good 👍🏻

deployment/postgres-init-db/init-aerie.sh
```diff
@@ -21,6 +21,12 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-E
   CREATE DATABASE aerie_ui;
   GRANT ALL PRIVILEGES ON DATABASE aerie_ui TO aerie;
   \echo 'Done!'
+
+  \echo 'Initializing aerie_expansion database...'
+  CREATE DATABASE aerie_expansion;
+  GRANT ALL PRIVILEGES ON DATABASE aerie_expansion TO aerie;
+  \echo 'Done!'
```
kroffo: I am not nearly as knowledgeable in this area as some of our other team members (@pcrosemurgy), but I think there's more to be done to add a new database to our system. Comparing to https://github.com/NASA-AMMOS/aerie/pull/42/files it seems that there's some Hasura configuration and other stuff to add.

kroffo (changes requested):
One issue and one concern to address, but other than those this is looking good!

deployment/postgres-init-db/init-aerie.sh
patkenneally: :small_red_triangle_down: Let's remove this from this PR as there are no components using the database. Let's keep this PR solely about bring the initial schemas into version control.

deployment/postgres-init-db/init-aerie.sh
```diff
@@ -36,3 +42,9 @@ psql -v ON_ERROR_STOP=1 --username "aerie" --dbname "aerie_ui" <<-EOSQL
   \ir /docker-entrypoint-initdb.d/sql/ui/init.sql
   \echo 'Done!'
 EOSQL
+
```
patkenneally: :small_red_triangle_down: Similarly as noted above

patkenneally (changes requested):

goetzrrGit:
@kroffo and @patkenneally I have made the requested changes

deployment/postgres-init-db/init-aerie.sh
```diff
@@ -26,6 +26,7 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-E
   CREATE DATABASE aerie_ui;
   GRANT ALL PRIVILEGES ON DATABASE aerie_ui TO aerie;
   \echo 'Done!'
+
```
patkenneally, line 29: :small_red_triangle_down: Got a rogue line here.

patkenneally (approved):

dyst5422:
An element we didn't entirely talk about, but I think warrants a discussion, is handling of the types autogenerated from the mission model. We can do this dynamically, but it might make sense to cache it in the same way we do the command dictionary types.

kroffo (approved):
My issues have been addressed, and I am happy to give my approval now
👍🏻

goetzrrGit:
Going to 'Hold Merge' on the PR as I will be moving these files to a different part of Aerie. The PR was to approve the first cut of the SQL schema and since it has been approved I am going to close this issue.

PR #40, opened 2022-02-08
[AERIE-1582] Fix concurrent composition of register effects

patkenneally:
* **Tickets addressed:** AERIE-1582
* **Review:** By commit  

## Description
The previous logic would produce an incorrect result for the expression `(set(1) | set(2)) | set(3)`, namely `set(3)`. The expected result is a pure conflict. As a result, @Twisol produced this fix int he logic and added a truth table of sorts to document explicitly the behavior of the RegisterEffect. 

## Verification
Unit test to come

## Documentation
I have created a documentation ticket AERIE-1681 for explaining the various cell types in the Merlin's contrib package.

## Future work
Possible formal tests.

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java
@@ -61,12 +61,36 @@ public final class RegisterEffect<T> {
 
     @Override
     public RegisterEffect<T> concurrently(final RegisterEffect<T> left, final RegisterEffect<T> right) {
-      // If neither is null, there's a conflict. Otherwise, pick the one that isn't null.
-      if (left.newValue == null) {
+      /*
+       left.hasValue | left.isConflicted | right.hasValue | right.isConflicted || ( left | right )
+      ---------------+-------------------+----------------+--------------------++--------------------
+      # The "one of them is a no-op" cluster
+               false |             false |          true  |              true  || right
+               false |             false |          true  |              false || right
+               false |             false |          false |              true  || conflict (right)
+               false |             false |          false |              false || noop (left/right)
+               false |             true  |          false |              false || conflict (left)
+               true  |             false |          false |              false || left
+               true  |             true  |          false |              false || left
+      # The "neither of them is a no-op" cluster
+               false |             true  |          true  |              true  || conflict (left)
+               false |             true  |          true  |              false || conflict (left)
+               false |             true  |          false |              true  || conflict (left/right)
+               true  |             false |          false |              true  || conflict (right)
+               true  |             true  |          false |              true  || conflict (right)
+               true  |             false |          true  |              true  || conflict
+               true  |             false |          true  |              false || conflict
+               true  |             true  |          true  |              true  || conflict
+               true  |             true  |          true  |              false || conflict
+       */
+      if (left.newValue == null && !left.conflicted) {
+        // Left is a no-op; take the right.
         return right;
-      } else if (right.newValue == null) {
+      } else if (right.newValue == null && !right.conflicted) {
+        // Right is a no-op; take the left.
         return left;
       } else {
+        // Left and right are both doing *something*, causing a pure conflict.
         return conflict();
       }
     }
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java
@@ -54,7 +54,7 @@ public final class RegisterEffect<T> {
         // The suffix is a pure conflict, and the prefix performed a valid write.
         return new RegisterEffect<>(prefix.newValue, suffix.conflicted);
       } else {
-        // Both suffix and prefix are pure conflicts, so the suffix dominates the prefix.
+        // The suffix is a pure conflict, and the prefix performs no write, so the suffix dominates the prefix.
         return suffix;
       }
     }


Twisol:
> However, this still feels like a bit of a cop out and that formalized tests might be best.

How would you feel about a mechanized proof of correctness in [Agda](https://agda.readthedocs.io/en/v2.6.2.1/getting-started/what-is-agda.html)? 👀

Of course, there's no formal connection between the Java code under discussion and the model of it in Agda, so it wouldn't make for a very good regression test. But maybe it at least gives us confidence in the current implementation?

(For what it's worth, I now have a proof that `apply(r, e1); apply(r, e2)` is equivalent to `apply(r, sequentially(e1, e2))`. This doesn't tell us that `sequentially`, `concurrently`, or `apply` actually do what we intend them to do; it's more of a coherence condition. But it's something!)

mattdailis (approved):
This hurts my brain, but I believe it's correct.

I considered requesting a refactor with variants like `NoOp`, `PureConflict`, `ConflictedWrite`, and `UnconflictedWrite`... but then I would expect the two combinators to have 16 case statements each, for every pairing of two variants, and that doesn't sound like much of a win over the two booleans 🤔 .

mattdailis:
> I considered requesting a refactor with variants like `NoOp`, `PureConflict`, `ConflictedWrite`, and `UnconflictedWrite`... but then I would expect the two combinators to have 16 case statements each, for every pairing of two variants, and that doesn't sound like much of a win over the two booleans 🤔 .

Here's what I had in mind - I tried to mirror all of the if statements in this PR exactly. Let me know what you think https://github.com/NASA-AMMOS/aerie/compare/hotfix/fix-register-concurrently-review

mattdailis:
I'm realizing that this comment is not exactly correct. It should say

> The prefix is either a no-op or a pure conflict, and the suffix is a pure conflict, so the suffix dominates the prefix

https://github.com/NASA-AMMOS/aerie/blob/d6c4ce6c8c73d41592d9c672e79039f68970f2d9/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java#L57

The functionality is correct though - `return suffix` is fine in both cases.

mattdailis (changes requested):
Requesting a change to the comment in the else branch of `sequentially` to acknowledge that `prefix` could be either a no-op, or a pure conflict.

Twisol:
> > I considered requesting a refactor with variants like `NoOp`, `PureConflict`, `ConflictedWrite`, and `UnconflictedWrite`... but then I would expect the two combinators to have 16 case statements each, for every pairing of two variants, and that doesn't sound like much of a win over the two booleans 🤔 .
> 
> Here's what I had in mind - I tried to mirror all of the if statements in this PR exactly. Let me know what you think https://github.com/NASA-AMMOS/aerie/compare/hotfix/fix-register-concurrently-review

I think I like it. Have you tried comparing it in performance to this branch? I think it's comparable in number of allocations, and there shouldn't be any significant difference between reading the class tag vs. a field. Although I'm not sure how reading the class tag might interact with the JIT.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/cells/register/RegisterEffect.java
```diff
@@ -61,12 +61,36 @@ public String toString() {
...
+      if (left.newValue == null && !left.conflicted) {
+        // Left is a no-op; take the right.
         return right;
-      } else if (right.newValue == null) {
+      } else if (right.newValue == null && !right.conflicted) {
+        // Right is a no-op; take the left.
         return left;
       } else {
+        // Left and right are both doing *something*, causing a pure conflict.
         return conflict();
```
kroffo, line 94: Based on your table, I don't think this is right. Consider the row on line 70. It looks like the first `if` condition passes and returns `right`, but `right` is conflicted, so shouldn't it actually return conflict?

I expected I may simply be confused, in which case would you mind explaining the intended behavior?

kroffo: @mattdailis and I discussed this and believe this makes sense now. For me, I think I got confused because I thought the problem was more complex than it actually seems to be. I now understand that the if conditions are simply saying that if the left one is irrelevant (by being empty and not conflicted) then just pass up the right regardless of what's in it, and vice-versa. If neither is irrelevant, then we can't combine them, so we pass up a conflict.

kroffo (approved):
I'm happy with this change, so I'm going to approve. I am not against Matt's suggestion either, but since it's not my suggestion I thought it worth giving my approval here too.

Twisol:
Anything I can do to help get this branch in? @patkenneally, would you like me to rebase this branch and incorporate @mattdailis' feedback?

patkenneally:
@Twisol I would be very happy for you to ply your wares on this to bring it up to date and include the changes from @mattdailis 🙇

Twisol:
Okay, I've rebased on latest `develop` and updated the comment @mattdailis noted for accuracy. I'd like to pose that @mattdailis' proposed redesign be evaluated separately from this PR, as (1) I think the correctness issue should be fixed ASAP, and (2) I'm not in a position to benchmark and make a meaningful comparison without the lander model in my hands.

mattdailis (approved):
Happy to defer stylistic refactoring in favor of correct functionality, especially as @Twisol pointed out, it would be best to do some performance testing before making any major changes. Approving now 👍

patkenneally:
Thank you @Twisol ! The task for adding unit tests for this behavior has been captured in internal JPL ticket AERIE-1740 and is scheduled to be completed in the coming Aerie 0.12.0 release window. I will go ahead and merge this when the checks have finished.

PR #41, opened 2022-02-08
[REFACTOR] Establish scheduler data flow interfaces

pcrosemurgy:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Establishes scheduler data flow interfaces by replacing `merlin-server` models/logic with scheduler-specific models/logic.

This is needed to provide a common working interface for @pcrosemurgy, @kroffo, and @adrienmaillard through scheduler database/data flow work.

## Verification
All tests passing.

## Documentation
None.

## Future work
@kroffo will fill in the stubbed out `LocalSpecificationService`, @adrienmaillard will complete the stubbed out `SynchronousSchedulerAgent`, @pcrosemurgy will hook up the scheduler DB within the scheduler app driver.

Eventually it'd also be nice to unify some of the common serialization methods/classes into a Gradle subproject that both `merlin-server` and `scheduler-server` depend on. This should be fairly straightforward – a good starting point would be to recursively find all identical `*.java` filenames within both projects, check if each matching file has a significant `diff`, and then relocate the file to the common subproject.


mattdailis (approved):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/SchedulerAppDriver.java
```diff
@@ -35,11 +36,12 @@ public static void main(final String[] args) {
 
     //create objects in each service abstraction layer (mirroring MerlinApp)
     final var merlinService = new GraphQLMerlinService(config.merlinGraphqlURI());
```
goetzrrGit, line 38: Is this `scheduler` of a new service? Should we not use the `merlin` in `GraphQLMerlinService`

goetzrrGit: Maybe `GraphQLSchedulerService`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -3,12 +3,12 @@
...
-import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
-import gov.nasa.jpl.aerie.merlin.server.models.Timestamp;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchPlanException;
+import gov.nasa.jpl.aerie.scheduler.server.http.InvalidEntityException;
+import gov.nasa.jpl.aerie.scheduler.server.http.InvalidJsonException;
+import gov.nasa.jpl.aerie.scheduler.server.http.SerializedValueJsonParser;
+import gov.nasa.jpl.aerie.scheduler.server.models.PlanId;
+import gov.nasa.jpl.aerie.scheduler.server.models.Timestamp;
 import gov.nasa.jpl.aerie.scheduler.AerieController;
 import gov.nasa.jpl.aerie.scheduler.MissionModelWrapper;
```
goetzrrGit, line 13: I was thinking this class can be rename since this isn't part of the merlin server?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
```diff
@@ -1,8 +1,8 @@
 package gov.nasa.jpl.aerie.scheduler.server.services;
 
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
-import gov.nasa.jpl.aerie.merlin.server.exceptions.NoSuchPlanException;
-import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchPlanException;
```
goetzrrGit, line 4: Similar comment about the 'Merline' name

goetzrrGit (commented):
Nothing sticks out as blockers. I just had comments about the naming of some classes that have 'Merlin' in them. I thought this service isn't part of the merlin-server so it might not need to have the 'Merlin' name in it.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/SchedulerAppDriver.java
pcrosemurgy: Thanks for looking through this! I think I see where there may be some confusion.

So the line below it is the actual scheduler service (`specificationService`, where `specification` is referring to a scheduler spec.). The `GraphQLMerlinService` is a GraphQL service that queries Merlin's GraphQL API.

pcrosemurgy: Currently `GraphQLMerlinService` is used to query for plans from Merlin. Kenny & Adrien will be working on removing this `GraphQLMerlinService` altogether and relying on `SpecificationService` to handle Merlin queries internally.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: The class is there to allow the scheduler server to communicate with a merlin server via GraphQL, it makes sense to me.

kroffo: That's right, the name is _service_ because this class acts as a service that provides Merlin's capabilities to the scheduler. This is the contract declared by the `MerlinService` interface, and this implementation carries out the contractual obligations via GraphQL.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
kroffo: Just be sure to resolve this comment too if you become satisfied with the discussion on the other one.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/SchedulerAppDriver.java
kroffo: @pcrosemurgy I don't think the second comment you said above is quite true. The specification service is a layer for interacting with the scheduler's database, and has nothing to do with Merlin. The GraphQLMerlinService will still be used by Adrien to load plans, mission models, etc. the way it already does.

The boundary between the Merlin and the Scheduler services is firm, with GraphQL the mechanism by which they may communicate. This is the purpose of the `GraphQLMerlinService`, and I don't think we're replacing that.

goetzrrGit: Thanks for the clarification guys.

goetzrrGit (approved):
Thanks for clarifying things for me.

adrienmaillard (approved):

kroffo (approved):

PR #43, opened 2022-02-10
[AERIE-1672] Scheduler Request Workflow

kroffo:
* **Tickets addressed:** AERIE-1672
* **Review:** By commit  

## Description
This work addresses the processing of scheduling requests to provide the proper input to the scheduling system. This primarily involves the implementation of postgres repositories and actions for pulling data out of the database. Rules and specifications are read as input for the scheduler, while requests are created to connect specifications to their results (which are not yet stored, see Future Work below)

## Verification
Upon spinning up the system, all database tables load without issue. Apart from that, I'm not sure how we can actually run the postgres actions right now, because there isn't a complete workflow. I think this is fine, as if there are any issues, we'll be fixing them when we get a more complete workflow together and test the system as a whole. 

## Documentation
No documentation yet, as we are just building the system up now, and it is not ready for use.

## Future work
Next up I will be implementing the `PostgresResultsCell` to get scheduling results written to the database (AERIE-1674).


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SpecificationRevisionData.java
```diff
@@ -0,0 +1,33 @@
...
+    List<RuleRecord> ruleRecords
+) implements RevisionData {
+  @Override
+  public MatchResult matches(final RevisionData other) {
+    if (!(other instanceof SpecificationRevisionData o)) return MatchResult.failure("");
+
+    if (this.specificationRevision != o.specificationRevision) {
+      return MatchResult.failure("Specification Revision mismatch");
+    } else if (this.ruleRecords().size() != o.ruleRecords.size()) {
+      return MatchResult.failure("Different number of rules");
```
pcrosemurgy: Not sure how often this failure case is expected to be hit (depends on if it's user-"triggerable" or implementation-"triggerable") but if it's expected somewhat frequently then it may be nice to add some string formatting here with the values in conflict. For example: `Expected %d rules, found %d.` or `Different number of rules: %d and %d`, etc.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/IntegrationFailureException.java
```diff
@@ -0,0 +1,17 @@
+package gov.nasa.jpl.aerie.scheduler.server.remotes.postgres;
+
+public abstract /*sealed*/ class IntegrationFailureException extends RuntimeException {
```
pcrosemurgy: Can uncomment sealed and add `permits DatabaseException, FailedInsertException` before the `extends` keyword :)

Or you can feel free to remove the `sealed` comment altogether, but the `sealed`/`permits` pattern may be nice to have if we ever want to do some sort of double-dispatch (`instanceof` pattern matching) on a plain `IntegrationFailureException` type.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -2,28 +2,140 @@
...
 
 import gov.nasa.jpl.aerie.scheduler.server.ResultsProtocol;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchRequestException;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchSpecificationException;
 import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
 import gov.nasa.jpl.aerie.scheduler.server.remotes.ResultsCellRepository;
+import gov.nasa.jpl.aerie.scheduler.server.services.ScheduleResults;
+import org.apache.commons.lang3.NotImplementedException;
 
 public record PostgresResultsCellRepository(DataSource dataSource) implements ResultsCellRepository {
```
pcrosemurgy: When I initially stubbed `PostgresResultsCellRepository` I made it a `record` to save myself some boilerplate. We don't need to expose `dataSource()` publicly, so if you wanted to you could make this a `class` with a `private final Datasource dataSource`. Thought it was worth mentioning, no obligation to change this.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
```diff
@@ -1,25 +1,72 @@
...
 import gov.nasa.jpl.aerie.scheduler.server.remotes.SpecificationRepository;
 import gov.nasa.jpl.aerie.scheduler.server.services.RevisionData;
 
-public record PostgresSpecificationRepository(DataSource dataSource) implements SpecificationRepository {
+import javax.sql.DataSource;
+import java.sql.SQLException;
+import java.util.stream.Collectors;
+
+public final class PostgresSpecificationRepository implements SpecificationRepository {
+  private final DataSource dataSource;
```
pcrosemurgy, line 24: Thanks for reworking this, sorry for being lazy with the `record` again here 😂

kroffo: Oh, did I replace it? I didn't mean to, I just took my changes over yours when I was resolving the conflicts without paying too much attention. I forget records exist, or I would probably use them. I have to remind myself every time I see them that they are just classes with extra conditions. I always think of them as nothing more than data, and forget that.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: Good to know, don't resolve this comment until next week so that I hopefully see it again then, otherwise I'll surely forget.

pcrosemurgy: For sure, not a big deal if it's not changed.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SpecificationRevisionData.java
kroffo: We honestly should remove the check on rules altogether, and add a trigger to the database to update the specification revision whenever a rule associated with the specification is updated. Then we can remove all the stuff about rules from `SpecificationRevisionData` altogether.

pcrosemurgy: That'd be nice. As part of this PR or a follow-up?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/IntegrationFailureException.java
kroffo: Alright, this should be fixed now. I've never worked with sealed anything before, so thanks for explaining how to implement it to me!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: This should be addressed now too. I made it a record to keep that variable private!

pcrosemurgy: Thanks! I think you meant to type `class`. While you're at it, can you make that class `final` too? So:
```java
public final class PostgresResultsCellRepository ...
```
No need to jump on it today, have a nice weekend!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SpecificationRevisionData.java
kroffo: Let me see how easily it works out right now.

kroffo: Turns out it was not difficult at all. Implemented and pushed.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: Taken care of!

kroffo:
Alright, I've updated the commits on this PR to exclusively use the term Goal instead of Rule, and everything should be good now.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SpecificationRevisionData.java
pcrosemurgy: Thanks!!

pcrosemurgy:
Thanks @kroffo, checking this out again now.

pcrosemurgy (approved):
This looks good to me! I'm curious if we have a ticket/plan to expand `db-tests/` to include some `scheduling` SQL tests.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
```diff
@@ -1,25 +1,69 @@
...
+          final var getSpecificationAction = new GetSpecificationAction(connection);
+          final var getSpecificationRulesAction = new GetSpecificationRulesAction(connection)
+      ) {
+        final var specificationRecord = getSpecificationAction.get(specificationId);
+        final var goals = getSpecificationRulesAction
+            .get(specificationId)
+            .stream()
+            .map(g -> new GoalRecord(
+                new GoalId(g.id()),
+                new CardinalityGoal.Builder().named(g.definition()).build()))
```
adrienmaillard: This will not work, the builder will tell you other fields (thereExists, inPeriod...) are missing. This : 
 ```
 CardinalityGoal goal = new CardinalityGoal.Builder()
      .inPeriod(ActivityExpression.ofType(new ActivityType("")))
      .thereExistsOne(new ActivityCreationTemplate.Builder()
                          .ofType(new ActivityType(""))
                          .build())
      .named("TestCardGoal")
      .forAllTimeIn(Window.at(Duration.SECONDS))
      .owned(ChildCustody.Jointly)
      .build();
```
is the minimalistic definition of a CardinalityGoal. And it should be much more restrictive than that, we should not be able to create such an ill-formed goal.

**Review by adrienmaillard — CHANGES_REQUESTED** [review]
I have not checked SQL constraints, I am not qualified and I trust others for that. All of this looks good to me but one item that should be changed.

File: scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java (hunk 5 above)
Comment (kroffo) [reply scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java]: Ah, yes this is the part I'm the least knowledgeable on. The need I was trying to address is simply that I need to create a Goal with the name instantiated, but I don't have any other information about the goal, because it's just a placeholder until we load the goals from the JAR. Might you have a suggestion for how to better handle this?

File: scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationRulesAction.java [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
+import org.intellij.lang.annotations.Language;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+/*package-local*/ final class GetSpecificationRulesAction implements AutoCloseable {
```
mattdailis: I'm still on the fence on the naming (and I will be for a long time, so no need to wait for me!)

However, for the sake of consistency with your other code - this should probably be renamed `GetSpecificationGoalsAction`.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
adrienmaillard: Either you can use the bit of code I have provided to create the minimal goal (with the appropriate name) or you will have to skip the creation of the fake goal and explicitly check the loaded jar goal name against the text content of the goal record. I have no preference as all of this is a temporary workaround.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/IntegrationFailureException.java
kroffo: @pcrosemurgy Just pinging you to resolve or comment on this.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
kroffo: Alright, thanks, I'll go with the code you have provided, as I would like to keep the interfaces accurate (`Specification` contains `GoalRecord`s which contains `Goal`s), so that replacing the workaround requires minimal change. Thanks Adrien!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationRulesAction.java
kroffo: Absolutely, I just missed it while renaming things! Thank you for your careful examination!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
kroffo: Alright, I implemented it and moved it to a function so the workaround portion is isolated from the code that won't need to be changed when removing the workaround.

kroffo:
I've made a couple of force pushes today. These updates include the following:
- Set name of goals in workaround function properly (was hardcoded string from copy/paste error)
- Make non-overridden PostgresResultCellRepository functions `static`
- Pull ID records out of postgres actions
- Return Optional types from postgres actions, instead of throwing exceptions. Exceptions are now thrown by callers when empty optionals are returned. This was motivated by the fact that the exceptions are based on the server model, but the actions work on the postgres model (which models postgres data more faithfully).

adrienmaillard (approved):

kroffo:
Latest force-push adds Hasura metadata.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationRulesAction.java
kroffo: Resolving as this was a simple update and don't want to get held up on this not being marked resolved.

PR #44, opened 2022-02-10
Support Hasura v2.2 action JSON `request_query` field

pcrosemurgy:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Supports new Hasura action handler fields (mainly `request_query`): https://hasura.io/docs/latest/graphql/core/actions/action-handlers.html#http-handler

## Verification
Simulation run through UI runs without issue. This serves as a decent end-to-end test for this PR.

## Documentation
None.

## Future work
None.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
@@ -89,24 +89,25 @@ public abstract class MerlinParsers {
           untuple((role, userId) -> new HasuraAction.Session(role, userId.orElse(""))),
           $ -> tuple($.hasuraRole(), Optional.ofNullable($.hasuraUserId()))));
 
-  private static <S> JsonParser<Pair<Pair<String, S>, HasuraAction.Session>> hasuraActionP(final JsonParser<S> inputP) {
+  private static <I> JsonParser<Pair<Pair<Pair<String, I>, HasuraAction.Session>, String>> hasuraActionP(final JsonParser<I> inputP) {
     return productP
-        .  field("action", productP.field("name", stringP))
-        .  field("input", inputP)
-        .  field("session_variables", hasuraActionSessionP);
+        .field("action", productP.field("name", stringP))
+        .field("input", inputP)
+        .field("session_variables", hasuraActionSessionP)
+        .field("request_query", stringP);
   }
 
   public static final JsonParser<HasuraAction<HasuraAction.MissionModelInput>> hasuraMissionModelActionP
       = hasuraActionP(productP.field("missionModelId", stringP))
       . map(Iso.of(
-          untuple((name, missionModelId, session) -> new HasuraAction<>(name, new HasuraAction.MissionModelInput(missionModelId), session)),
-          $ -> tuple($.name(), $.input().missionModelId(), $.session())));
+          untuple((name, missionModelId, session, requestQuery) -> new HasuraAction<>(name, new HasuraAction.MissionModelInput(missionModelId), session)),
+          $ -> tuple($.name(), $.input().missionModelId(), $.session(), "")));
 
   public static final JsonParser<HasuraAction<HasuraAction.PlanInput>> hasuraPlanActionP
       = hasuraActionP(productP.field("planId", planIdP))
       . map(Iso.of(
-          untuple((name, planId, session) -> new HasuraAction<>(name, new HasuraAction.PlanInput(planId), session)),
-          $ -> tuple($.name(), $.input().planId(), $.session())));
+          untuple((name, planId, session, requestQuery) -> new HasuraAction<>(name, new HasuraAction.PlanInput(planId), session)),
+          $ -> tuple($.name(), $.input().planId(), $.session(), "")));
 
   public static final JsonParser<HasuraMissionModelEvent> hasuraMissionModelEventTriggerP
       = productP
@@ -136,8 +137,8 @@ public abstract class MerlinParsers {
   public static final JsonParser<HasuraAction<HasuraAction.MissionModelArgumentsInput>> hasuraMissionModelArgumentsActionP
       = hasuraActionP(hasuraMissionModelArgumentsInputP)
       .map(Iso.of(
-          untuple(HasuraAction::new),
-          $ -> tuple($.name(), $.input(), $.session())));
+          untuple((name, input, session, requestQuery) -> new HasuraAction<>(name, input, session)),
+          $ -> tuple($.name(), $.input(), $.session(), "")));
 
   private static final JsonParser<HasuraAction.ActivityInput> hasuraActivityInputP
       = productP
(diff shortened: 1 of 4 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
```diff
@@ -89,24 +89,25 @@ public JsonValue unparse(final Timestamp value) {
...
-  private static <S> JsonParser<Pair<Pair<String, S>, HasuraAction.Session>> hasuraActionP(final JsonParser<S> inputP) {
+  private static <I> JsonParser<Pair<Pair<Pair<String, I>, HasuraAction.Session>, Optional<String>>> hasuraActionP(final JsonParser<I> inputP) {
     return productP
-        .  field("action", productP.field("name", stringP))
-        .  field("input", inputP)
-        .  field("session_variables", hasuraActionSessionP);
+        .field("action", productP.field("name", stringP))
+        .field("input", inputP)
+        .field("session_variables", hasuraActionSessionP)
+        .optionalField("request_query", stringP);
```
mattdailis: 🔶 A "discussion question" of sorts is - is `requestQuery` truly optional, or does `v2.2` always provide it?

The optional seems like a reasonable way to support both `v2.0.10` and `v2.2`, but I just want to make sure I understand the motivation.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
```diff
@@ -136,8 +137,8 @@ public JsonValue unparse(final Timestamp value) {
   public static final JsonParser<HasuraAction<HasuraAction.MissionModelArgumentsInput>> hasuraMissionModelArgumentsActionP
       = hasuraActionP(hasuraMissionModelArgumentsInputP)
       .map(Iso.of(
-          untuple(HasuraAction::new),
-          $ -> tuple($.name(), $.input(), $.session())));
+          untuple((name, input, session, requestQuery) -> new HasuraAction<>(name, input, session)),
```
mattdailis, line 140: 🔶 Would it make any sense to add the `requestQuery` as a field on `HasuraAction`? Or do we prefer to discard it, since we have no use for it?

mattdailis (approved):
Looks good to me.

Legend:
🔻 Issues to address before merge
🔶 Requests that should not block merge, but should at least be discussed
🔵 Recommendations that can be ignored if desired

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
pcrosemurgy: At first that's what I did, but until we have a use for it I think it's probably best to discard it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
pcrosemurgy: Yeah it's nice that this would allow us to downgrade to v2.0 if we ever needed to. Am open to making this a plain old `field()` and discarding it instead. No strong opinion here so lmk if you have any preference :)

camargo (approved):
Tested simulation action and it works as expected. LGTM.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
mattdailis: 🔻 I think it's fine, but write a comment explaining the reasoning, and I think we'll want to eventually pick one version to support.

goetzrrGit:
I seem to have broken something on my end so I couldn't test it. No activity_types show up in Hasura and I can't add anything to a plan. No need to block merging though.

pcrosemurgy:
@goetzrrGit thanks for testing this out, I'll double check and make sure I don't see the same issue. You may need to remove all your Docker images, volumes, etc. (since we've upgraded the Hasura image).

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
pcrosemurgy: Just pushed a change to make this a `field()` :)

goetzrrGit:
@pcrosemurgy I nuked everything. Here is how I ran it on my computer.

`docker system prune --all --volumes   `
`./gradlew build; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up
`

pcrosemurgy:
Hmm, what's your output when running `docker images`?
I'm thinking you may still have the old Hasura image. My output looks like:
```
REPOSITORY                    TAG                        IMAGE ID       CREATED         SIZE
aerie_merlin                  latest                     c02801e69e00   3 seconds ago   578MB
ghcr.io/nasa-ammos/aerie-ui   develop                    ba1ec4c473b6   2 hours ago     363MB
hasura/graphql-engine         v2.2.0.cli-migrations-v3   ebdf522b135f   8 days ago      462MB
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
pcrosemurgy: Also retested with this change. Looking good

goetzrrGit:
I think I have the latest:

```
REPOSITORY                         TAG                        IMAGE ID       CREATED          SIZE
aerie_merlin                       latest                     c37b031f2657   25 minutes ago   578MB
aerie_scheduler                    latest                     03c139cbeaeb   25 minutes ago   608MB
ghcr.io/nasa-ammos/aerie-ui        develop                    ba1ec4c473b6   2 hours ago      363MB
ghcr.io/nasa-ammos/aerie-gateway   develop                    c36efce8e44f   4 hours ago      147MB
hasura/graphql-engine              v2.2.0.cli-migrations-v3   ebdf522b135f   8 days ago       462MB
postgres                           14.1                       da2cb49d7a8d   2 weeks ago      374MB
```

pcrosemurgy:
Sorry Ryan, really not too sure here. To be honest, when spinning up a local Docker stack I completely remove all images, volumes, containers, networks, etc. It takes ~3 minutes but while waiting I can generally work on other things. I am wondering if your `aerie_merlin` image is too old? It looks like it was created 25 mins. ago (from when you ran `docker images`) but I think you tested things out ~10 mins. prior to checking `docker images`. @camargo is my goto for Docker stack issues, he may be able to spot the issue.

**Edit** assuming that the `aerie_merlin` image was created after checking out this branch and running `gradle clean build` then my comment about its `CREATED` timestamp is probably irrelevant.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinParsers.java
mattdailis: Looks fine to me - as long as we don't need to support older versions of hasura (I'm not the right person to make that decision)

mattdailis (approved):
Giving a second approval, since I ended up requesting changes. 

I think it's okay to break support for older versions of hasura. We're still an alpha project, and don't want to bloat our code with too much baggage just yet.

PR #46, opened 2022-02-11
[AERIE-1673]  Compute Scheduling Results

adrienmaillard:
* **Tickets addressed:** AERIE-1673
* **Review:** By commit  

## Description

In this draft PR, I am exposing the current changes I am making to the scheduler to accommodate the new types of scheduling results that we have to provide. 

The main challenge is to implement and track which goals create which activities because a distinction between the two is made in the results. 

To do that, I had to make this distinction in the `Evaluation` object, which tracks the progress and success of scheduling goals. 

Before, we did not track associations when activity A created by goal G1 were satisfying goal G2. We only tracked that A was created by G1. It also gave me the opportunity to actually start enforcing the ownership relationship that exists between activities and goals. A goal can decide that the activities it has created cannot be used to satisfy other goals. 

To track both creations and associations, I have created a new conflict that a goal can produce when an existing activity satisfies this goal but is not yet associated with this goal. The goals must check beforehand that they actually can piggyback on the existing activity. Then, the solver proceeds to associate the activity to the goal.

All the goals had to be modified to produce these conflicts. 

Then, when all the information had been made available, I have changed the server so it can package it. Because we have to return maps of `GoalId` to `ActivityInstanceId`, I had to make the server keep track of the created/associated activity instances in the plans we send/receive from Merlin.  

Note that the server gets rules from the database but actually still loads the Merlinsight/AerieLander jar. This will have to be finished later when we actually store goals in the database and we can translate them to objects. 

This work is evolving with the changes that @kroffo and @pcrosemurgy are making to the server. I also have to write some new tests for the scheduler. 

## Verification

All tests pass including AerieLander tests 

## Documentation

None
## Future work

None

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
@@ -1,14 +1,18 @@
 package gov.nasa.jpl.aerie.scheduler.server.services;
 
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.scheduler.ActivityInstance;
 import gov.nasa.jpl.aerie.scheduler.Plan;
 import gov.nasa.jpl.aerie.scheduler.Problem;
 import gov.nasa.jpl.aerie.scheduler.Time;
 import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchPlanException;
 import gov.nasa.jpl.aerie.scheduler.server.models.PlanId;
 import gov.nasa.jpl.aerie.scheduler.server.models.PlanMetadata;
+import org.apache.commons.lang3.tuple.Pair;
 
 import java.io.IOException;
+import java.util.Map;
 
 /**
  * interface for retrieving / storing details to merlin
@@ -54,7 +58,7 @@ public interface MerlinService {
    * @return the database id of the newly created aerie plan container
    * @throws NoSuchPlanException when the plan container could not be found in aerie after creation
    */
-  PlanId createNewPlanWithActivities(final PlanMetadata planMetadata, final Plan plan)
+  Pair<PlanId, Map<ActivityInstance, ActivityInstanceId>> createNewPlanWithActivities(final PlanMetadata planMetadata, final Plan plan)
   throws IOException, NoSuchPlanException;
 
   /**
@@ -89,8 +93,9 @@ public interface MerlinService {
    * @param planId aerie database identifier of the target plan to synchronize into
    * @param plan plan with all activity instances that should be stored to target merlin plan container
    * @throws NoSuchPlanException when the plan container does not exist in aerie
+   * @return
    */
-  void updatePlanActivities(final PlanId planId, final Plan plan) throws IOException, NoSuchPlanException;
+  Map<ActivityInstance, ActivityInstanceId> updatePlanActivities(final PlanId planId, final Plan plan) throws IOException, NoSuchPlanException;
 
   /**
    * confirms that the specified plan exists in the aerie database, throwing exception if not
@@ -121,8 +126,9 @@ public interface MerlinService {
    * @param planId the database id of the plan container to populate with new activity instances
    * @param plan the plan from which to copy all activity instances into aerie
    * @throws NoSuchPlanException when the plan container does not exist in aerie
+   * @return
    */
-  void createAllPlanActivities(final PlanId planId, final Plan plan) throws IOException, NoSuchPlanException;
+  Map<ActivityInstance, ActivityInstanceId> createAllPlanActivities(final PlanId planId, final Plan plan) throws IOException, NoSuchPlanException;
 
 
 }
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
@@ -1,26 +1,37 @@
 package gov.nasa.jpl.aerie.scheduler.server.services;
 
+import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModel;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModelLoader;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
-import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchSpecificationException;
+import gov.nasa.jpl.aerie.scheduler.ActivityInstance;
 import gov.nasa.jpl.aerie.scheduler.GlobalConstraint;
 import gov.nasa.jpl.aerie.scheduler.Goal;
 import gov.nasa.jpl.aerie.scheduler.HuginnConfiguration;
 import gov.nasa.jpl.aerie.scheduler.JarClassLoader;
 import gov.nasa.jpl.aerie.scheduler.Plan;
+import gov.nasa.jpl.aerie.scheduler.PlanningHorizon;
 import gov.nasa.jpl.aerie.scheduler.PrioritySolver;
 import gov.nasa.jpl.aerie.scheduler.Problem;
 import gov.nasa.jpl.aerie.scheduler.Solver;
+import gov.nasa.jpl.aerie.scheduler.Time;
 import gov.nasa.jpl.aerie.scheduler.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.scheduler.server.config.PlanOutputMode;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchPlanException;
+import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchSpecificationException;
 import gov.nasa.jpl.aerie.scheduler.server.exceptions.ResultsProtocolFailure;
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalRecord;
+import gov.nasa.jpl.aerie.scheduler.server.models.PlanId;
 import gov.nasa.jpl.aerie.scheduler.server.models.PlanMetadata;
+import gov.nasa.jpl.aerie.scheduler.server.models.Specification;
 
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -31,7 +42,7 @@ import java.util.stream.Collectors;
  *
  * @param merlinService interface for querying plan details from merlin
  * @param modelJarsDir path to parent directory for mission model jars (interim backdoor jar file access)
- * @param rulesJarPath path to jar file to load scheduling rules from (interim solution for user input rules)
+ * @param goalsJarPath path to jar file to load scheduling goals from (interim solution for user input goals)
  * @param outputMode how the scheduling output should be returned to aerie (eg overwrite or new container)
  */
 //TODO: will eventually need scheduling goal service arg to pull goals from scheduler's own data store
@@ -39,7 +50,7 @@ public record SynchronousSchedulerAgent(
     SpecificationService specificationService,
     GraphQLMerlinService merlinService,
     Path modelJarsDir,
-    Path rulesJarPath,
+    Path goalsJarPath,
     PlanOutputMode outputMode
 )
     implements SchedulerAgent
@@ -47,7 +58,7 @@ public record SynchronousSchedulerAgent(
   public SynchronousSchedulerAgent {
     Objects.requireNonNull(merlinService);
     Objects.requireNonNull(modelJarsDir);
-    Objects.requireNonNull(rulesJarPath);
+    Objects.requireNonNull(goalsJarPath);
   }
 
   /**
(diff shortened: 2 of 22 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -65,13 +88,22 @@ public void associate(java.util.Collection<ActivityInstance> acts) {
...
+      return java.util.Collections.unmodifiableSet(acts.entrySet().stream().filter((a)-> a.getValue().equals(true)).map(
+          Map.Entry::getKey).collect(
+          Collectors.toSet()));
     }
-
     /**
      * the set of all activities that contributed to the evaluation
      */
-    protected final java.util.Set<ActivityInstance> acts = new java.util.HashSet<>();
+    protected final java.util.Map<ActivityInstance, Boolean> acts = new java.util.HashMap<>();
```
kroffo: Would you mind updating this documentation too? It appears to now be a map where the value simply tells if the activity was created to satisfy this goal, right?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
```diff
@@ -0,0 +1,28 @@
...
+    this.instances = instancesToChooseFrom;
+  }
+
+  public Collection<ActivityInstance> getActivityInstancesToChooseFrom(){
+    return instances;
+  }
+
+  @Override
+  public Windows getTemporalContext() {
+    return null;
```
kroffo, line 27: Why is this null? If it's only unclear to me because I'm unfamiliar with this area of our system, a simple answer here will suffice, but if it's less obvious then a code comment may be helpful too.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+   * ctor creates a new conflict
+   *
+   * @param goal IN STORED the dissatisfied goal that issued the conflict
+   */
+  public MissingAssociationConflict(final Goal goal, final Collection<ActivityInstance> instancesToChooseFrom) {
+    super(goal);
+    this.instances = instancesToChooseFrom;
+  }
```
kroffo, line 19: What are the `instancesToChooseFrom` for? I'm not understanding (likely due to lack of context) why this is a list, but in use it seems to be provided a list of a single instance, for example in Cardinality goal (see comment there).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/CardinalityGoal.java
```diff
@@ -173,24 +178,32 @@ protected CardinalityGoal fill(CardinalityGoal goal) {
...
       }
+    }
 
+    //at this point, have thrown exception if not satisfiable
+    //TODO:should not throw an exception but something more of an explanation
+    //compute the missing association conflicts
+    for(var act:acts){
+      if(!associatedActivitiesToThisGoal.contains(act) && planEvaluation.canAssociateMoreToCreatorOf(act)){
+        //they ALL have to be associated
+        conflicts.add(new MissingAssociationConflict(this, List.of(act)));
```
kroffo, line 195: For my understanding, could you offer a bit of explanation as to what the conflict's purpose is? It makes me curious, seeing a loop creating a bunch of lists of single things (not saying it's wrong, just that I would like to understand what's going on here).

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
```diff
@@ -54,7 +58,7 @@ Plan getPlanActivities(final PlanMetadata planMetadata, final MissionModelWrappe
    * @return the database id of the newly created aerie plan container
    * @throws NoSuchPlanException when the plan container could not be found in aerie after creation
    */
-  PlanId createNewPlanWithActivities(final PlanMetadata planMetadata, final Plan plan)
+  Pair<PlanId, Map<ActivityInstance, ActivityInstanceId>> createNewPlanWithActivities(final PlanMetadata planMetadata, final Plan plan)
```
kroffo, line 61: What's the motivation for IDs to be the values instead of the keys?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -64,32 +74,112 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
     try {
       //confirm requested plan to schedule from/into still exists at targeted version (request could be stale)
       //TODO: maybe some kind of high level db transaction wrapping entire read/update of target plan revision
+
+      // TODO: Remove workaround function wrapper once goals are read from database
+      final var specificationAndPlanMetadata = loadSpecificationGoalsFromJAR(specificationService.getSpecification(request.specificationId()));
       ensureRequestIsCurrent(request);
+      final var planMetadata = specificationAndPlanMetadata.getValue();
+      final var specification = specificationAndPlanMetadata.getKey();
```
kroffo: This feels a little odd. Rather than getting the PlanMetadata in `loadSpecificationGoalsFromJAR()` and returning it alongside the specification, I suggest loading the PlanMetadata here, and passing it into the `loadSpecificationGoalsFromJAR()` method. As-is we are giving this function a hidden secondary purpose.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -64,32 +74,112 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
...
+      ensurePlanRevisionMatch(specification,planMetadata.planRev());
```
kroffo: Good, I was telling myself all week that I need to make sure you added this when I review this PR! You didn't need me too!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -257,10 +332,16 @@ private void storeFinalPlan(final PlanMetadata planMetadata, final MissionModelW
    * @param plan the target plan after the scheduling run has completed
    * @return summary of the state of the plan after scheduling ran; eg goal success metrics, associated instances, etc
    */
-  private ScheduleResults collectResults(final Plan plan) {
-    final var activityCount = plan.getActivities().size();
-    //TODO: complete stub
-    return new ScheduleResults(Map.of()); // TODO Adrien: complete this stubbed instance
+  private ScheduleResults collectResults(final Plan plan, final Map<ActivityInstance, ActivityInstanceId> instancesToIds, Map<Goal, GoalId> rulesToIds) {
+    Map<GoalId, ScheduleResults.GoalResult> ruleResults = new HashMap<>();
```
kroffo: I suggest renaming all occurrences of `rule` to `goal`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PrioritySolver.java
```diff
@@ -193,29 +193,17 @@ private void solve() {
    * @return a descending-priority ordered queue of goals from the input
    *     problem, ready for processing
    */
-  private java.util.PriorityQueue<Goal> getGoalQueue() {
+  private java.util.LinkedList<Goal> getGoalQueue() {
```
kroffo: Since `LinkedList` is imported, you can remove the `java.util.` here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
```diff
@@ -34,7 +34,7 @@ public Problem(MissionModelWrapper mission) {
   /**
    * container of all goals in the problem, indexed by name
    */
-  protected final java.util.HashMap<String, Goal> goalsByName = new java.util.HashMap<>();
+  protected final java.util.List<Goal> goalsOrderedByPriority = new java.util.ArrayList<>();
```
kroffo: No need for the `java.util.` here, and you can add the import for `ArrayList` as well.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -23,16 +26,17 @@ private ActivityType getActivityType(final String name) {
...
 
   @Override
-  public java.util.Collection<Goal> getGoals() {
-    Collection<Goal> goals = new ArrayList<>();
-    goals.add(generateDSNVisibilityAllocationGoal());
-    goals.addAll(getFirstRuleGoals());
-    goals.addAll(getSecondRuleGoals());
-    goals.addAll(getThirdRuleGoals());
-    return goals;
+  public java.util.List<Goal> getGoals() {
```
kroffo: Another unnecessary `java.util.`

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
```diff
@@ -21,7 +24,7 @@ public void testRecurrence() {
     MissionModelWrapper missionModel = new MissionModelWrapper();
     Problem problem = new Problem(missionModel);
 
-    problem.add(goal);
+    problem.addAll(List.of(goal));
```
kroffo: Why not just add a wrapper function for `addAll()` on `Problem` that takes a single goal and wraps it in a list?

kroffo (changes requested):
Mostly questions to help my understanding here, and some small suggestions, but nothing major.

Some more general PR feedback:
Some of your commit messages run over the character limit, but also don't fully explain motivation where it is unclear at times. Commit descriptions are a great way to remedy this, as you can put a short commit message on top, and have a paragraph of description explaining the details, motivation and context of the commit.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -41,22 +44,42 @@
      */
     protected double score = 0.0;
```
goetzrrGit: Move to the top of the `GoalEvaluation` class

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -41,22 +44,42 @@
...
 
+    /**
+     *
+     */
+    protected Integer nbConflictsDetected = null;
```
goetzrrGit: This one too

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -41,22 +44,42 @@
...
      */
-    public void associate(ActivityInstance act) { acts.add(act); }
+    public void associate(ActivityInstance act, boolean createdByThisGoal) { acts.put(act, createdByThisGoal);}
 
     /**
      * flags all given activities as contributing to the goal's (dis)satisfaction
      *
      * @param acts IN container of activities that contributed to the goal's
      *     evaluation
      */
```
goetzrrGit: It might be good to update the javadoc with the `createdByThisGoal` parameter

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -41,22 +44,42 @@
...
+      }
+      return Optional.of(nbConflictsDetected);
+    }
+
     /**
      * flags given activity as contributing to the goal's (dis)satisfaction
      *
      * @param act IN the activity instance that contributed to the goal's
      *     evaluation
      */
```
goetzrrGit: It might be good to update the javadoc with the `createdByThisGoal` parameter

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PrioritySolver.java
```diff
@@ -237,42 +238,57 @@ private void satisfyOptionGoal(OptionGoal goal) {
...
-          actsToInsert = listOfActivitiesInserted;
-          currentSatisfiedGoal = subgoal;
+        if(evaluation.forGoal(subgoal).getScore() == 0 || subgoal.isPartiallySatisfiable()) {
+          var associatedActivities = evaluation.forGoal(subgoal).getAssociatedActivities();
+          var insertedActivities = evaluation.forGoal(subgoal).getInsertedActivities();
+          var aggregatedActivities = new ArrayList<ActivityInstance>();
+          aggregatedActivities.addAll(associatedActivities);
+          aggregatedActivities.addAll(insertedActivities);
+          if (aggregatedActivities.size() > 0 && (goal.optimizer.isBetterThanCurrent(new ArrayList<>(
+              aggregatedActivities)) || currentSatisfiedGoal == null)) {
```
goetzrrGit: Is there a reason you are creating a new ArrayList object instead of passing in the `aggregatedActivities` ArrayList

`goal.optimizer.isBetterThanCurrent(aggregatedActivities)`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PrioritySolver.java
```diff
@@ -237,42 +238,57 @@ private void satisfyOptionGoal(OptionGoal goal) {
...
-        //TODO: a score equal to 0 is perfect for this notation so we want to do something else probably
-        evaluation.forGoal(subgoal).setScore(0);
+        rollback(subgoal);
       }
       //we should have the best solution
       if (currentSatisfiedGoal != null) {
+        for(var act: actsToAssociateWith){
+          //we do not care about ownership here as it is not really a piggyback but just the validation of the supergoal
+          evaluation.forGoal(goal).associate(actsToInsert, false);
+        }
```
goetzrrGit: Not sure if this loop is needed since you are not using the `var act` unless `evaluation.forGoal(goal).associate(actsToInsert, false);` needs to be called multiple times?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Collection;
+
+public class MissingAssociationConflict extends Conflict {
+  private final Collection<ActivityInstance> instances;
+
+  /**
+   * ctor creates a new conflict
+   *
+   * @param goal IN STORED the dissatisfied goal that issued the conflict
+   */
```
goetzrrGit: Update javadoc with `Collection<ActivityInstance> instancesToChooseFrom`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
```diff
@@ -98,8 +77,8 @@ public void add(Goal goal) {
    *
    * @return an un-modifiable container of the goals requested for this plan
    */
-  public java.util.Collection<Goal> getGoals() {
-    return java.util.Collections.unmodifiableCollection(goalsByName.values());
+  public java.util.List<Goal> getGoals() {
```
goetzrrGit: Ditto Kenny suggestions remove `java.util.`

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -70,18 +74,18 @@ public Goal generateDSNVisibilityAllocationGoal(){
 
     ProceduralCreationGoal dsnGoal = new ProceduralCreationGoal.Builder()
         .named("Schedule DSN contacts for initial setup")
-        .withPriority(50)
+        ////.withPriority(50)
```
goetzrrGit: If you don't need this line you can remove it.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -108,12 +112,12 @@ public Goal generateDSNVisibilityAllocationGoal(){
                              Duration.of(1,Duration.MINUTE)));
     ProceduralCreationGoal pro = new ProceduralCreationGoal.Builder()
         .named("TurnOnAndOFFMonitoring")
-        .withPriority(10)
+        //.withPriority(10)
```
goetzrrGit: Ditto

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -129,14 +133,14 @@ public Goal generateDSNVisibilityAllocationGoal(){
 
     RecurrenceGoal goal1a = new RecurrenceGoal.Builder()
         .named("1a")
-        .withPriority(9)
+        //.withPriority(9)
```
goetzrrGit: Remove

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -152,20 +156,20 @@ public Goal generateDSNVisibilityAllocationGoal(){
         .named("1c")
         .inPeriod(new TimeRangeExpression.Builder().from(new Windows(planningHorizon.getHor())).build())
         .forAllTimeIn(planningHorizon.getHor())
-        .withPriority(8)
+        //.withPriority(8)
```
goetzrrGit: Remove

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -479,11 +472,11 @@ public Goal generateDSNVisibilityAllocationGoal(){
                           .withArgument("compQuality", SerializedValue.of(95))
                           .build())
       .forEach(secondIdaMoveArm)
-      .withPriority(4)
+      //.withPriority(4)
```
goetzrrGit: remove

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -507,11 +500,11 @@ public Goal generateDSNVisibilityAllocationGoal(){
                           .duration(Duration.of(15, Duration.MINUTE))
                           .build())
       .forEach(firstICCImages)
-      .withPriority(4)
+      //.withPriority(4)
```
goetzrrGit: Remove

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRules.java
```diff
@@ -713,12 +704,12 @@ public Goal generateDSNVisibilityAllocationGoal(){
                             .withArgument("xbandAntSel", SerializedValue.of("EAST_MGA"))
                             .build())
         .forEach(tre2)
-        .withPriority(3)
+        //.withPriority(3)
```
goetzrrGit: Remove

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestCardinalityGoal.java
```diff
@@ -31,14 +44,14 @@ public void testone() {
         .named("TestCardGoal")
         .forAllTimeIn(period)
         .owned(ChildCustody.Jointly)
-        .withPriority(7.0)
+        //.withPriority(7.0)
```
goetzrrGit, line 47: Remove the all of the //.withPriority

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
goetzrrGit: +1

goetzrrGit (changes requested):
Nice job!!1 I haven't really been involved with the scheduler side of Aerie, but I have some code changes that you might want to look at.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
adrienmaillard: This conflict says that there is a missing association. And sometimes you can associate to several instances (because several corresponds) and sometimes only one. and the decision is made by the scheduler, not the conflict itself.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
adrienmaillard: It is because the temporal context here is irrelevant. We could probably throw an exception instead because you should not use the temporal context of a `MissingAssociationConflict`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/CardinalityGoal.java
adrienmaillard: The idea is that a goal detects an activity that can satisfy it but see that it is not associated to it yet so it says to the scheduler that it is missing an association only. And then the scheduler ultimately decides what to do. A goal does not make decisions but only suggests. And sometimes you can choose the association from multiple activities (think of a recurrence goal when you want 1 activity for a period and you have 3), you let the scheduler decide which one to associate with.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/MerlinService.java
adrienmaillard: because when we get the plan back from the scheduler and we want to compute the results, you get a collection of activity instances and have to find which merlin activity instance it corresponds to.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
adrienmaillard: So the idea was to signify that the only way to pass goals to the scheduler is to pass an ordered list. But I implemented it wrong. I am replacing this Problem::addAll by Problem::setGoals that clears the current list of goals and adds all the passed goals. And still no adding a single goal because that might give the wrong idea.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PrioritySolver.java
adrienmaillard: You are right.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/PrioritySolver.java
adrienmaillard: You spotted a bug, I have corrected it.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
adrienmaillard: I did what you proposed.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
adrienmaillard: I'll let you resolve on this one.

adrienmaillard:
I have adressed all of your comments in my recent force-push.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
kroffo: Ah, that does make sense then. I appreciate the name `setGoals()` much more!

kroffo (approved):
All my concerns have been addressed, so I can now approve 👍🏻

goetzrrGit (approved):
All my concerns have been addressed. Feel free to merge.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java
Twisol: Just as a note, it's usually good to [avoid using `null`](http://www.javapractices.com/topic/TopicAction.do?Id=134) in situations where it can leak out beyond its context. 

If you shouldn't be calling the method (and you have a way to know this from the outside), an exception is good. Otherwise, an [informationless instance](http://www.javapractices.com/topic/TopicAction.do?Id=59) of Windows is better, e.g. `Windows.forever()` or `new Windows()` (empty).

dyst5422:
Disclaimer - I have no idea what I’m talking about.

There is a distinct difference between a window that covers everything and no window at all - which might be the case here.


On Feb 18, 2022, at 21:57, Jonathan Castello ***@***.***> wrote:

﻿

@Twisol commented on this pull request.

________________________________

In scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/MissingAssociationConflict.java<https://urldefense.us/v3/__https://github.com/NASA-AMMOS/aerie/pull/46*discussion_r810452134__;Iw!!PvBDto6Hs4WbVuu7!Z8JBoHWL7RohhuDUWtjOw1wwkLKd9ypT9pA6lvEH7D8byHk_o1Anh66uJwD7ZItNARgCJ8kkl0f7$>:

> +   * ctor creates a new conflict
+   *
+   * @param goal IN STORED the dissatisfied goal that issued the conflict
+   */
+  public MissingAssociationConflict(final Goal goal, final Collection<ActivityInstance> instancesToChooseFrom) {
+    super(goal);
+    this.instances = instancesToChooseFrom;
+  }
+
+  public Collection<ActivityInstance> getActivityInstancesToChooseFrom(){
+    return instances;
+  }
+
+  @Override
+  public Windows getTemporalContext() {
+    return null;


Just as a note, it's usually good to avoid using null<https://urldefense.us/v3/__http://www.javapractices.com/topic/TopicAction.do?Id=134__;!!PvBDto6Hs4WbVuu7!Z8JBoHWL7RohhuDUWtjOw1wwkLKd9ypT9pA6lvEH7D8byHk_o1Anh66uJwD7ZItNARgCJ02h7gMO$> in situations where it can leak out beyond its context.

If you shouldn't be calling the method (and you have a way to know this from the outside), an exception is good. Otherwise, an informationless instance<https://urldefense.us/v3/__http://www.javapractices.com/topic/TopicAction.do?Id=59__;!!PvBDto6Hs4WbVuu7!Z8JBoHWL7RohhuDUWtjOw1wwkLKd9ypT9pA6lvEH7D8byHk_o1Anh66uJwD7ZItNARgCJ8OQbiNg$> of Windows is better, e.g. Windows.forever().

Twisol:
> Disclaimer - I have no idea what I’m talking about. There is a distinct difference between a window that covers everything and no window at all - which might be the case here.

Yep, but they're both suitably "universal" objects -- there's no specific information to go on. Which one is actually a good replacement for `null` is up to context... which is kind of the point of why `null` isn't great ^_^

Twisol:
I just noticed that you replied to the email notif, meaning you didn't see my edit! I edited in `new Windows()` as an option before you replied.

PR #45, opened 2022-02-10
[AERIE-1648] Extract MissionModelWrapper and redistribute its functionalities

adrienmaillard:
* **Tickets addressed:** AERIE-1648
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

Motivation: many components of the `MissionModelWrapper` class are more appropriately encapsulated by the other existing classes and data structures.

Its functionalities have been mostly redistributed into `Problem`. 

There are two commits: the first is the refactoring, the second one contains the consequences of a linting pass on the scheduler code.

## Verification

Validated against `MerlinsightRules`. 

## Documentation

None

## Future work

None

PR #48, opened 2022-02-16
[HOTFIX] Update AerieLander scheduler tests

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Recent updates to the way mission model configuration works (notably #20) required changes to the aerie lander mission model. The scheduler tests were not updated to match, and this eluded notice because they are excluded from compilation. This PR brings them back into compatibility with aerie lander.

~The second commit of this PR renames the tests to use "AerieLander" instead of deprecated terminology.~

## Verification
I'm still not 100% sure what the proper workflow is for running tests against the aerielander repository. I cloned it and saved it in its own subproject of aerie, and updated `settings.gradle` as well as `scheduler/build.gradle` to depend on it. I then commented out the compile exclusions in `scheduler/build.gradle` and ran the tests, and saw that they passed.

## Documentation
No documentation is necessary for these changes at this time, in my opinion.

## Future work
- use the name AerieLander consistently
- streamline the process for running these tests, so we can detect these regressions sooner

### DIFF
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlinSightTestUtility.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlinSightTestUtility.java
@@ -1,6 +1,7 @@
 package gov.nasa.jpl.aerie.scheduler;
 
 import gov.nasa.jpl.aerie.merlin.driver.MissionModelBuilder;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


kroffo (approved):
Looks good to me, though I'm trusting your run of the tests since running them myself is not straightforward. We should just make sure other PRs that will be affected by this (via conflicts) are discussed before merging. I believe @adrienmaillard has a PR that touches these files a bit.

adrienmaillard (approved):
Fine with me as well. I have been applying the test process you are describing over and over since aerielander has disappeared from the repo.

mattdailis:
> Looks good to me, though I'm trusting your run of the tests since running them myself is not straightforward. We should just make sure other PRs that will be affected by this (via conflicts) are discussed before merging. I believe @adrienmaillard has a PR that touches these files a bit.

You're right Kenny, this does cause merge conflicts with #45 . What I'm going to do is delete my second commit, which includes the renames. Without that commit, there are no merge conflicts with #45 . After merging #45 we can make a PR just for the rename 👍

PR #49, opened 2022-02-16
[Aerie-1686]  Setup JS Command Expansion Server

goetzrrGit:
* **Tickets addressed:** AERIE-1686
* **Review:** By commit  

## Description

Instead of having another GitHub repo for the command expansion service, it was decided to keep all the Aerie services under one project. I have added a command-expansion-server directory containing a basic node server we can build upon.

- [x] Gradle build processes
- [x] Docker compose and env variables to connect to db
- [x] PR building
- [x] Publishing build products (docker image to ghcr)
- [x] Identify a testing harness (we can chat about this more if you like)

## Verification

- `docker system prune --all --volumes  <-- Nuke everything `
- `./gradlew build; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up`
- Verify you have a new docker container `aerie_commanding` running 
- Login to the new container and see that the node server is running
- Enter localhost:3000 on your browser to connect with the server

## Documentation
Added Command Expansion Service -  https://github.com/NASA-AMMOS/aerie/wiki/Product-Guide

## Future work
[AERIE-1687](https://jira.jpl.nasa.gov/browse/AERIE-1687)
Setup database schema intialization for command expansion server

### DIFF
--- a/command-expansion-server/build.gradle
+++ b/command-expansion-server/build.gradle
@@ -18,11 +18,11 @@ task build(type: NpmTask) {
   args = ['run', 'build']
 }
 
-task test(type: NpmTask) {
-  dependsOn processResources
-  dependsOn npmInstall
-  args = ['run', 'test']
-}
+//task test(type: NpmTask) {
+//  dependsOn processResources
+//  dependsOn npmInstall
+//  args = ['run', 'test']
+//}
 
 task clean(type: Delete) {
   delete 'build', 'node_modules', 'package-lock.json'
--- a/command-expansion-server/build.gradle
+++ b/command-expansion-server/build.gradle
@@ -1,0 +1,19 @@
+
+plugins {
+  id "com.github.node-gradle.node" version "3.1.1"
+}
+
+node {
+  version = '16.4.0'
+  npmVersion = '8.5.0'
+  download = true
+}
+
+task build(type: NpmTask) {
+  dependsOn npmInstall
+  args = ['run','build']
+}
+
+task clean(type: Delete) {
+  delete 'build','node_modules', 'package-lock.json'
+}
--- a/command-expansion-server/build.gradle
+++ b/command-expansion-server/build.gradle
@@ -1,6 +1,5 @@
-
 plugins {
-  id "com.github.node-gradle.node" version "3.1.1"
+  id 'com.github.node-gradle.node' version '3.1.1'
 }
 
 node {
@@ -9,11 +8,22 @@ node {
   download = true
 }
 
+task processResources {
+  // Currently does nothing, exists as a hook for `distributeSql`
+}
+
 task build(type: NpmTask) {
+  dependsOn processResources
   dependsOn npmInstall
-  args = ['run','build']
+  args = ['run', 'build']
+}
+
+task test(type: NpmTask) {
+  dependsOn processResources
+  dependsOn npmInstall
+  args = ['run', 'test']
 }
 
 task clean(type: Delete) {
-  delete 'build','node_modules', 'package-lock.json'
+  delete 'build', 'node_modules', 'package-lock.json'
 }
--- a/command-expansion-server/src/packages/db/db.ts
+++ b/command-expansion-server/src/packages/db/db.ts
@@ -1,0 +1,40 @@
+import type { Pool, PoolConfig } from "pg";
+import pg from "pg";
+import { getEnv } from "../../env.js";
+
+const { Pool: DbPool } = pg;
+
+const {
+  POSTGRES_AERIE_EXPANSION_DB,
+  POSTGRES_HOST: host,
+  POSTGRES_PASSWORD: password,
+  POSTGRES_PORT: port,
+  POSTGRES_USER: user,
+} = getEnv();
+
+export class DbExpansion {
+  private static pool: Pool;
+
+  static getDb(): Pool {
+    return DbExpansion.pool;
+  }
+
+  static init() {
+    try {
+      const config: PoolConfig = {
+        database: POSTGRES_AERIE_EXPANSION_DB,
+        host,
+        password,
+        port: parseInt(port, 10),
+        user,
+      };
+
+      console.log(`Postgres Config:`);
+      console.log(config);
+
+      DbExpansion.pool = new DbPool(config);
+    } catch (error) {
+      console.log(error);
+    }
+  }
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -64,6 +64,24 @@ services:
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store
+  aerie_commanding:
+    build:
+      context: ./command-expansion-server
+      dockerfile: Dockerfile
+    container_name: aerie_commanding
+    depends_on: [ "postgres" ]
+    environment:
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      COMMANDING_SERVER_PORT : 3000
+      COMMANDING_DB: "aerie_expansion"
+      COMMANDING_DB_PASSWORD: "aerie"
+      COMMANDING_DB_PORT: 5432
+      COMMANDING_DB_SERVER: "postgres"
+      COMMANDING_DB_USER: "aerie"
+      COMMANDING_LOCAL_STORE: /usr/src/app/commanding_file_store
+    image: aerie_commanding
+    ports: [ "3000:3000" ]
+    restart: always
   aerie_ui:
     container_name: aerie_ui
     depends_on: ["postgres"]
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,6 +14,7 @@ include 'parsing-utilities'
 // Services for deployment within the Aerie infrastructure
 include 'merlin-server'
 include 'scheduler-server'
+include 'command-expansion-server'
 
 // Command-line interface direct to services
 include 'constraints'
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -22,3 +22,5 @@ jobs:
         run: ./gradlew test
       - name: Assemble
         run: ./gradlew assemble
+      - name: Assemble Node Project
+        run: ./gradlew command-expansion-server:build
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,6 +38,8 @@ jobs:
           reporter: java-junit
       - name: Assemble
         run: ./gradlew assemble
+      - name: Assemble Node Project
+        run: ./gradlew command-expansion-server:build
       - name: Login to the Container Registry
         uses: docker/login-action@v1
         with:
@@ -70,6 +72,19 @@ jobs:
           push: true
           tags: ${{ steps.aerieScheduler.outputs.tags }}
           labels: ${{ steps.aerieScheduler.outputs.labels }}
+      # Build aerie-commanding Docker artifacts.
+      - name: Extract metadata (tags, labels) for aerie-commanding Docker image
+          id: aerieCommanding
+          uses: docker/metadata-action@v3
+          with:
+            images: ${{ env.REGISTRY }}/nasa-ammos/aerie-commanding
+      - name: Build and push aerie-commanding Docker image
+          uses: docker/build-push-action@v2
+          with:
+            context: ./command-expansion-server
+            push: true
+            tags: ${{ steps.aerieCommanding.outputs.tags }}
+            labels: ${{ steps.aerieCommanding.outputs.labels }}
       # Publish via Gradle.
       - name: Publish Package
         run: ./gradlew publish
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@
 /nbbuild/
 /nbdist/
 /nbproject/private/
+node_modules
 
 # Ignore any SQL files copied from build tasks
 deployment/postgres-init-db/sql/**/*.sql
--- a/build.gradle
+++ b/build.gradle
@@ -11,40 +11,32 @@ task archiveDeployment(type: Tar) {
 
 // Add `distributeSql` task to all subprojects with `sql/` child directory
 configure(subprojects.findAll { it.projectDir.toPath().resolve('sql').toFile().exists() }) { sp ->
-  // Specify only subprojects with `processResources` tasks,
-  // otherwise Gradle cannot see `processResources` as a valid task
-  ['java', 'com.github.node-gradle.node'].each { pluginId ->
-    plugins.withId(pluginId) {
 
-      task distributeSql(type: Copy) {
-        into "$rootDir/deployment/postgres-init-db/sql"
-        from fileTree(dir: "${sp.projectDir}/sql", include: '**/*.sql')
-      }
+  // Distribute SQL as part of resource processing
+  task distributeSql(type: Copy, dependsOn: 'processResources') {
+    into "$rootDir/deployment/postgres-init-db/sql"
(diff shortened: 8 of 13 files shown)


command-expansion/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    // "emitDecoratorMetadata": true,                    /* Emit design-type metadata for decorated declarations in source files. */
+    // "jsxFactory": "",                                 /* Specify the JSX factory function used when targeting React JSX emit, e.g. 'React.createElement' or 'h' */
+    // "jsxFragmentFactory": "",                         /* Specify the JSX Fragment reference used for fragments when targeting React JSX emit e.g. 'React.Fragment' or 'Fragment'. */
+    // "jsxImportSource": "",                            /* Specify module specifier used to import the JSX factory functions when using `jsx: react-jsx*`.` */
+    // "reactNamespace": "",                             /* Specify the object invoked for `createElement`. This only applies when targeting `react` JSX emit. */
+    // "noLib": true,                                    /* Disable including any library files, including the default lib.d.ts. */
+    // "useDefineForClassFields": true,                  /* Emit ECMAScript-standard-compliant class fields. */
+
+    /* Modules */
+    "module": "commonjs",                                /* Specify what module code is generated. */
```
dyst5422: We should target ESNext so we run it as a module. Lets tag up quick about this at some point and talk it through.

command-expansion/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    /* Projects */
+    // "incremental": true,                              /* Enable incremental compilation */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./",                          /* Specify the folder for .tsbuildinfo incremental compilation files. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2016",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
dyst5422: We should have this be ESNext to track any JS api updates - since we're pulling the latest node version in the docker image

command-expansion/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
dyst5422: `"lib": ["node"],`

command-expansion/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    // "jsxFragmentFactory": "",                         /* Specify the JSX Fragment reference used for fragments when targeting React JSX emit e.g. 'React.Fragment' or 'Fragment'. */
+    // "jsxImportSource": "",                            /* Specify module specifier used to import the JSX factory functions when using `jsx: react-jsx*`.` */
+    // "reactNamespace": "",                             /* Specify the object invoked for `createElement`. This only applies when targeting `react` JSX emit. */
+    // "noLib": true,                                    /* Disable including any library files, including the default lib.d.ts. */
+    // "useDefineForClassFields": true,                  /* Emit ECMAScript-standard-compliant class fields. */
+
+    /* Modules */
+    "module": "commonjs",                                /* Specify what module code is generated. */
+     "rootDir": "./src",                                  /* Specify the root folder within your source files. */
+    // "moduleResolution": "node",                       /* Specify how TypeScript looks up a file from a given module specifier. */
```
dyst5422: Uncomment this

command-expansion/tsconfig.json
```diff
@@ -0,0 +1,101 @@
...
+    // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
+    // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
+    // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
+    // "typeRoots": [],                                  /* Specify multiple folders that act like `./node_modules/@types`. */
+    // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
+    // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
+    // "resolveJsonModule": true,                        /* Enable importing .json files */
```
dyst5422: Uncomment this for easy building in of JSON configs

settings.gradle
```diff
@@ -14,6 +14,7 @@ include ':parsing-utilities'
 // Services for deployment within the Aerie infrastructure
 include ':merlin-server'
 include ':scheduler-server'
+include ':command-expansion'
```
mattdailis: Would it make sense to follow this naming convention and call it `command-expansion-server`?

dyst5422 (changes requested):

dyst5422:
Regarding a test runner, Jest is definitely the way to go nowadays. There are a couple steps to setting it up with TypeScript

Additional dev-dependencies for that - `jest`, `jest-circus`, `jest-html-reporters`, `ts-jest`

Probably also want to include `ts-doc` for auto documentation generation

command-expansion/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+node {
+  version = '16.0.0'
+  npmVersion = '8.5.0'
+  download = true
+}
+
+task build(type: NpxTask) {
+  dependsOn npmInstall
+
+  command = 'npx'
```
mattdailis: I'm curious about the pros/cons of putting together the command here, vs using the npm run tasks directly.
(see [docs](https://github.com/node-gradle/gradle-node-plugin/blob/master/docs/usage.md#executing-npm-tasks))

For example, depending on the `npm_build` task would be equivalent to running `npm run build`, which is defined to be `npx tsc` in package.json.

One possible advantage of your way is that specifying the input files could prevent unnecessary re-builds? But I'm just guessing here.

command-expansion/tsconfig.json
dyst5422: Part of this will be renaming files to `.mts` so they transpile to `.mjs` and node recognizes them as module-type files

command-expansion/src/app.js
```diff
@@ -0,0 +1,11 @@
+"use strict";
```
camargo: Since this `.js` file is built (generated by `tsc`) we can/should remove it

command-expansion/tsconfig.json
camargo: For completeness, another option here is to use the `"type": "module"` flag in your `package.json` file:
https://nodejs.org/docs/latest-v16.x/api/packages.html#type
If you want to drop back to CommonJS you can use the `.cjs` extension.

command-expansion/tsconfig.json
camargo: Note we are using the LTS Node version for all our Docker base images (v16.4.0).

command-expansion/build.gradle
goetzrrGit: I wanted to make things simpler and consistent by using the same command `./gradlew build` which will build all the other Aerie services. You can `./gradlew npm_build` too, but I felt that was one extra thing to remember so I defined a `build` task instead.

mattdailis: Good point :+1: I like the consistency as well.

I'm noticing that the first line is `dependsOn npmInstall`. Would it still achieve what we're trying to do if we added another line `dependsOn npm_build`?

I'm not super familiar with gradle, so I'm not sure this is the "right way" to call one task from another task. I'm curious to hear your thoughts.

command-expansion/tsconfig.json
goetzrrGit: @dyst5422 @camargo Will `ESNext` use the LTS version or the latest version which could be something higher in the future?

settings.gradle
goetzrrGit: Sure no objections to the name

command-expansion/tsconfig.json
dyst5422: The support in the package.json file is sporadic among tooling.

command-expansion/tsconfig.json
dyst5422: It tracks the latest advancements to the language (that have been stabilized)

`Additional APIs available in ESNext - This changes as the JavaScript specification evolves`

camargo: So I believe `ESNext` currently points to `ES2022+`: https://www.typescriptlang.org/tsconfig#lib
According to this site https://node.green/#ES2022 Node LTS for `ES2022` is ~70% feature complete.
Where `ES2021` is 100% feature complete: https://node.green/#ES2021 so I'd go with that

command-expansion/tsconfig.json
camargo: What tooling? This feature is very widely used in big open source JS projects

command-expansion/src/app.js
goetzrrGit: Actually, this file shouldn't be here so I removed it.

command-expansion/tsconfig.json
dyst5422: Jest is the one that jumps out at me (although it's been a year or so since i've tried doing that).

command-expansion/tsconfig.json
dyst5422: Sounds good

command-expansion/build.gradle
goetzrrGit: Thanks for pointing this out. I reworked the task into npm type which is a lot cleaner

command-expansion/tsconfig.json
goetzrrGit: @dyst5422 there is no `"lib": ["node"]` available. I can set it to `ES2021` though. Would that work? Since the `target` and `module` are `ES2021`

command-expansion/tsconfig.json
dyst5422: Sorry, I don't know why its tagging ALL these lines, I meant just the last one with `resolveJsonModule`

The others we can leave out

command-expansion/tsconfig.json
dyst5422: Right! sorry, I am tired today and not thinking straight. Yes thats fine. I think we just want the latest stable specification for each of these.

command-expansion-server/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+  download = true
+}
+
+task build(type: NpmTask) {
+  dependsOn npmInstall
+  args = ['run','build']
+}
+
+task clean(type: Delete) {
+  delete '.gradle', 'build','node_modules', 'package-lock.json'
```
pcrosemurgy: I think you can safely remove `'.gradle'` from this list here. `.gradle/` should only contain cached Gradle dependencies, and in this Gradle subproject the only cached dep. should be the `node-gradle` plugin.

pcrosemurgy: I also think `'build'` can be removed since this task (`clean`) is not overriding the default Gradle `clean` task (which removes `build/`). I'd double check me on this though.

goetzrrGit: Thanks for the suggestion. I should make a task called `nuke` which will remove everything. :P

goetzrrGit: I just verified and `'build'` is needed. `.gradle clean` doesn't remove the `build/` directory in the javascript project.

pcrosemurgy: Thanks for checking! :)

command-expansion-server/src/app.mts
```diff
@@ -0,0 +1,18 @@
+import express, { Application, Request, Response } from "express";
```
camargo: We lost syntax highlighting here with the `.mts` extension. 
Can we consider the other option of using `"type": "module"` in `package.json`?

mattdailis: I don't have an opinion on `.mts` vs "type": "module", since all this is very new to me, and I'm still learning about what javascript modules even are.

However, the syntax highlighting doesn't seem to be an issue on my end - VSCode was able to recognize `.mts` as typescript out of the box, and IntelliJ let me add `*.mts` to the list of patterns associated with Typescript under `Preferences/Editor/File Types`.

dyst5422: We had a discussion on slack (@camargo @goetzrrGit @dyst5422) and resolved all concerns with type: module, so I think we'll go that direction to preserve highlighting in GH

pcrosemurgy:
Making a note to myself: on #51 https://github.com/NASA-AMMOS/aerie/pull/51/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R16 we'll need to add a line here to detect `processResources` for the `command-expansion-server` subproject. Kind of confusing but basically `build.gradle` has no knowledge of whether each subproject has an actual `processResources` task unless a narrowing `plugins.withId()` context is used.

Ryan, I'll work with you to make sure we get this right :)

pcrosemurgy (approved):
My JS is not too strong but the DB connection setup/logic seems to make sense.

command-expansion-server/src/app.mts
goetzrrGit: @dyst5422 it should be fixed now.

command-expansion-server/src/packages/db/db.ts
```diff
@@ -0,0 +1,40 @@
...
+} = getEnv();
+
+export class DbExpansion {
+  private static pool: Pool;
+
+  static getDb(): Pool {
+    return DbExpansion.pool;
+  }
+
+  static async init(): Promise<void> {
```
dyst5422: I don't see any async in this function right now. I assume we have this as an async method assuming that there is potential for it to be async in the future.

docker-compose.yml
```diff
@@ -64,6 +64,24 @@ services:
...
+    build:
+      context: ./command-expansion-server
+      dockerfile: Dockerfile
+    container_name: aerie_commanding
+    depends_on: [ "postgres" ]
+    environment:
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      COMMANDING_SERVER_PORT : 3000
+      COMMANDING_DB: "aerie_expansion"
+      COMMANDING_DB_PASSWORD: "aerie"
```
dyst5422, line 77: Is there security vulnerability with this password being on GH?

goetzrrGit: If you look at the docker-compose we have embedded multiple usernames and passwords for our other services. It is a valid point but maybe not a concern as of now. What do you think @patkenneally

command-expansion-server/src/packages/db/db.ts
goetzrrGit: Right now we are not doing anything async so I will remove it. If we do later, I can add it back in.

docker-compose.yml
patkenneally: This is on our radar. We should endeavor to remove these in favor of say [Docker Compose Secrets](https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images) . Additionally, our documentation regarding such deployment considerations is absent. We should provide users with guidance of typical configuration and security considerations. We'll take up this work as a separate ticket.

camargo (approved):

kroffo (commented):
I'm going to just comment to say I have looked at this, but I don't have enough expertise in this area to feel like my approval would carry any weight at this time. Since you already have the required number of approvals, this should not be an issue.

dyst5422 (approved):

PR #51, opened 2022-02-17
[AERIE-1666] Distribute SQL with Gradle

pcrosemurgy:
* **Tickets addressed:** AERIE-1666
* **Review:** By commit  

## Description
1. Automatically copies SQL files from their Gradle subproject `sql/` directory to the central `deployment/postgres-init-db/sql/` directory. 
1. Relocates centralized SQL files to their respective subprojects (under a `sql/` directory).
1. Adds GitHub action to upload `deployment.tar` archive as part of the `publish` workflow.

## Verification
Items 1 and 2 above were tested locally by running `gradle assemble` and checking that all distributed SQL files were copied over correctly and correctly git-ignored.

Item 3 will be verified upon merging this PR (running the `publish` workflow).

## Documentation
None.

## Future work
None.


.github/workflows/publish.yml
```diff
@@ -72,6 +72,14 @@ jobs:
           labels: ${{ steps.aerieScheduler.outputs.labels }}
       # Publish via Gradle.
       - name: Publish Package
-        run: gradle publish
+        run: ./gradlew publish
```
goetzrrGit, line 75: Just curious why ./gradlew and not Gradle. Still new to Gradle and my interaction with it has only been `./gradlew build `, `clean`, and creating a `task`

scheduler/build.gradle
```diff
@@ -23,23 +22,24 @@ sourceSets {
...
-    from sourceSets.test.output
-    include 'gov/nasa/jpl/aerie/scheduler/MerlInsightRules.class'
-    archiveFileName = nameMerlinSightJar
+def nameMerlinSightJar = 'merlinsight-rules.jar'
+def testResourceDir = 'src/test/resources'
+
+task merlinsightJar(type: Jar, dependsOn: compileTestJava)  {
+  from sourceSets.test.output
+  include 'gov/nasa/jpl/aerie/scheduler/MerlInsightRules.class'
+  archiveFileName = nameMerlinSightJar
```
pcrosemurgy, line 31: Just a note, no semantic changes were made in this file. I cleaned up some formatting, quotes, and variable names.

build.gradle
```diff
@@ -1,8 +1,35 @@
...
+task archiveDeployment(type: Tar) {
+  destinationDirectory = file('.')
+  archiveFileName = 'deployment.tar'
+  into 'deployment'
+  from 'deployment'
+}
+
+// Add `distributeSql` task to all subprojects with `sql/` child directory
+configure(subprojects.findAll { it.projectDir.toPath().resolve('sql').toFile().exists() }) {
+  pluginManager.withPlugin('java') {
```
goetzrrGit: The new Command Expansion Server is a node/javascript project. This project will have SQL schemas as well. Will this be overlooked as it is not java?

settings.gradle
```diff
@@ -1,29 +1,29 @@
...
 // DB tests
-include ':db-tests'
+include 'db-tests'
 
 // Example mission models
-include ':examples:banananation'
-include ':examples:foo-missionmodel'
-include ':examples:config-with-defaults'
-include ':examples:config-without-defaults'
+include 'examples:banananation'
```
pcrosemurgy, line 26: Another change that's just syntactic. The leading `:` is unnecessary here.

goetzrrGit (commented):
I tested locally on my system and the SQL files are copied over correctly.

build.gradle
pcrosemurgy: Aha! You beat me to it 😂 
Mentioned this on your PR: https://github.com/NASA-AMMOS/aerie/pull/49#issuecomment-1043251309

.github/workflows/publish.yml
pcrosemurgy: Mostly for consistency here since the other actions that use gradle use the gradle wrapper (`./gradlew`). In general I think it's best practice to use the wrapper, that you know exactly what version of gradle is running, despite what the machine/VM has installed.

When using gradle locally I always just use my system's `gradle` and make sure all projects I'm contributing to use the latest version of gradle – so it's usually not a problem locally, but when running on servers you never know I guess.

build.gradle
pcrosemurgy: Basically just need to also consider the `com.github.node-gradle.node` plugin. I would preemptively add this but I'd rather add it to your PR, that way I can verify it works.

goetzrrGit: I have another ticket [AERIE-1687](https://jira.jpl.nasa.gov/browse/AERIE-1687) we can work on instead of adding it to my current PR. The other ticket will utilize this PR.

.github/workflows/publish.yml
goetzrrGit: Thank you for the explanation.

goetzrrGit (approved):

.gitignore
```diff
@@ -36,7 +37,7 @@
 /nbproject/private/
 
 # Ignore any SQL files copied from build tasks
-deployment/postgres-init-db/sql/*/*.sql
+deployment/postgres-init-db/sql/**/*.sql
```
kroffo, line 40: What does _two_ stars mean?

kroffo (approved):
I'm no gradle expert at all, but looks good to me!

.gitignore
pcrosemurgy: It's a recursive wildcard, so any arbitrarily nested directory structure containing `.sql` files will be git-ignored (unless explicitly un-git-ignored).

scheduler-server/sql/scheduler/tables/scheduling_rule.sql
```diff
@@ -0,0 +1,51 @@
+create table scheduling_rule (
```
patkenneally, line 1: :large_orange_diamond: Random question: is there a reason the path to these files is `scheduler-server/sql/scheduler/tables/*`, rather than just `scheduler-server/sql/tables/*` ? Does the addition of `scheduler` provide some delineation I'm not aware of?

patkenneally (approved):
I left one question that I'd like to now about, but dropping an approval here still.

scheduler-server/sql/scheduler/tables/scheduling_rule.sql
pcrosemurgy: Good question! Yes, the `scheduler` directory is needed to maintain parity with the directory structure that `deployment/postgres-init-db/sql/<table_name>` needs. Since `<table_name>` may differ from the Gradle subproject name (`merlin-server` vs. `merlin` for example) this extra layer is needed :)

pcrosemurgy: Could rework this to just be `scheduler-server/sql/tables/*` and then have `scheduler-server/build.gradle` define a DB table name variable, but this feels a little more natural than relying on the build system for that.

patkenneally: Thanks for the explanation! We can leave it as you have it. I can't think of a strong reason to do otherwise at the moment.

build.gradle
pcrosemurgy: Going rouge and resolving this comment ;)

.gitignore
pcrosemurgy: Also resolving this one to merge, feel free to reach out about this Kenny :)

kroffo: Ironically I clicked resolve and it gave me an error because you clicked it just before I did!

PR #52, opened 2022-02-17
[AERIE-1674] Persist Scheduling Results

kroffo:
* **Tickets addressed:** AERIE-1674
* **Review:** By commit  

## Description
This branch provides the implementation of the scheduler-server's `PostgresResultsCellRepository`. The commits on this branch define the Postgres actions that will be used by the repository, and in the final commit the actions are hooked up tot he repository. Do note that some commits have descriptions explaining the motivation for their changes that reviewers may find useful.

## Verification
Each commit was tested by running `./gradlew classes` to make sure all commits compile. `./gradlew test` was also run at the end to ensure that no tests were broken. I also made sure the database initializes properly, since the scheduling_request table was modified.

As for testing the postgres actions themselves, I haven't run them since the scheduling agent is still in development.

## Documentation
No documentation updates needed. These features are in development, and not yet documented.

## Future work
Once the scheduler agent is hooked up we should do some integration testing and make sure a scheduling request can be created and run through our system, even if we have to manually insert the data into the database since the UI isn't hooked up yet. As careful as we try to be, it's difficult to be sure our code that interacts with the database is accurate without actually testing. Since none of this is in use yet, though, it shouldn't be a big deal if there's any issues to fix.


deployment/postgres-init-db/sql/scheduler/init.sql
```diff
@@ -2,5 +2,15 @@
```
goetzrrGit: Just a note that @patkenneally PR changes the location of each service's SQL schema. Everything will be eventually moved to:
`<your-gradle-subproject>/sql/<table-name>
`

goetzrrGit: And I just realized this is drafted :P

kroffo: If that PR gets merged first I'll move the files 👍🏻

kroffo:
Force-pushed rebase off new develop to incorporate the SQL file location changes. No conflicts arose as git correctly applied the update to the scheduling_request table to the moved file.

deployment/postgres-init-db/sql/scheduler/init.sql
kroffo: @goetzrrGit This is now addressed 👍🏻

kroffo:
Force-pushed a rebase off develop that also fixes the Error messages thrown by the cancel and state setting actions (they referred to the dataset table from merlin-server).

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -96,53 +108,185 @@ private static RequestRecord createRequest(
...
     public ResultsProtocol.State get() {
-      // TODO
-      throw new NotImplementedException("PostgresResultsCell not yet implemented");
+      try (final var connection = dataSource.getConnection()) {
+        final var request = getRequest(connection, specId, specRevision);
+        return switch(request.status()) {
+          case "incomplete" -> new ResultsProtocol.State.Incomplete();
+          case "failed" -> new ResultsProtocol.State.Failed(request.failureReason());
+          case "success" -> new ResultsProtocol.State.Success(getResults(connection, request.analysisId()));
+          default -> throw new Error(String.format("Unexpected scheduling request state: %s", request.status()));
```
pcrosemurgy: I'd be nice if `ResultRecord.status` was an `enum` type. Within the get request action you could parse the string. Just worries me since the `String status` does not offer compile-type guarantees and these strings are used in more than one place.

pcrosemurgy:
Will look through more next week but just wanted to say I really appreciate how consistent you are with your DB actions.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: I would rather not parse the string in the action, because I believe the actions are merely for data retrieval. They provide the data that's in the database, but don't operate on it. The interpretation of that data is the responsibility of the relevant postgres repository, and I believe this boundary is beneficial to us (please let me know if your thoughts differ, I only just settled on my understanding of this boundary last week, and am always happy to hear other ideas/suggestions).

Now that said, I agree that having a switch statement on these string values feels less than ideal, but the string-to-State conversion is entirely contained in the `PostgresResultsCellRepository` in both directions*. In other words, the serialization/deserialization using strings for the State of a scheduling request is no more than an implementation detail of the `PostgresResultsCellRepository`.

`ResultsProtocol.State` is an interface (which could be made sealed) with three implementations, each with their own unique associated data. We can't do that with an enum, since all three variants have to have the same structure. We _could_ explore ways to move the strings into the interface implementations themselves, but I think this would cause more complexity than it's worth.

*the `"incomplete"` state is never explicitly set, but is the default value for new rows  added to the `scheduling_request` table

kroffo:
> Will look through more next week but just wanted to say I really appreciate how consistent you are with your DB actions.

Konsistency is key!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
pcrosemurgy: I'm cool with whatever you decide on here since at the end of the day it's fairly benign but thought I should give my 2 cents:

>I would rather not parse the string in the action, because I believe the actions are merely for data retrieval. They provide the data that's in the database, but don't operate on it. 

That's generally true but the boundary is not very clear for very discrete types like enums or ID records. For things like `PlanId` we already do parsing within actions, so I don't think an `incomplete`/`failed`/`success` parse is much of a stretch.

> In other words, the serialization/deserialization using strings for the State of a scheduling request is no more than an implementation detail of the PostgresResultsCellRepository.

Agreed, it is an implementation detail but the issue I see is exactly that: the implementation is error prone because we're relying on a custom and case-sensitive string in more than 1 place in the same class. `succeedRequest()` for example:
https://github.com/NASA-AMMOS/aerie/pull/52/files/edd39ea3767ee578e8c0165bb3ffeaddd90f688d#diff-cc4e1dc19800e82575716e95ec48d6a045df8df9af90ab257d7f56dd9293b829R159

>ResultsProtocol.State is an interface (which could be made sealed) with three implementations, each with their own unique associated data. We can't do that with an enum, since all three variants have to have the same structure.

This can be solved with double-dispatch (visitor pattern or `sealed` & `instanceof`).

kroffo: Regarding the boundary between the server code and postgres action code, you're right, in the merlin-server, we build/parse `PlanId` records in postgres actions. We shouldn't. If you look at the new actions for the scheduler-server here, you'll see that I took care to avoid this, because I feel that we should keep our server models and postgres models separate. There are places where it almost feels silly, but keeping this boundary firm allows us to examine either side of it without having to keep the other in mind and reduces coupling between the server code and postgres code to a single location (the repository class itself). I do think if time permits it would actually be worth refactoring the merlin-server postgres actions to conform to this boundary as well.

As for the `ResultProtocol.State` implementation, I'm happy to explore other implementations that enable a more reliable postgres repository implementation, so I'll see what things would look like based on your suggestions and get back to you on that part.

pcrosemurgy: Cool, and completely with you on the `PlanId` parsing stuff.

kroffo: @pcrosemurgy Here's the thing about the `ResultsProtocol.State` problem. We need to go from string value to State instance and vice-versa. We could _get_ strings from State objects by adding a method to the State interface, but we can't go in the other direction without basically having a switch statement as we do now. And the fact that we need to compute results for a success variant, or a failure reason for a failed variant makes it even less clear how we would do such a thing.

camargo:
> Once the scheduler agent is hooked up we should do some integration testing and make sure a scheduling request can be created and run through our system, even if we have to manually insert the data into the database since the UI isn't hooked up yet. 

Can you please share with me (Wiki page maybe?) the GQL requests you make to insert the data when you run these manual integration tests? It will help me move on the UI more quickly if I have some concrete examples to play with.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSatisfyingActivitiesAction.java
```diff
@@ -0,0 +1,50 @@
...
+  public Map<GoalId, List<ActivityInstanceId>> get(final long analysisId) throws SQLException {
+    this.statement.setLong(1, analysisId);
+    final var resultSet = this.statement.executeQuery();
+
+    final var satisfyingActivities = new HashMap<GoalId, List<ActivityInstanceId>>();
+    while (resultSet.next()) {
+      final var goalId = new GoalId(resultSet.getLong("goal_id"));
+      final var activityId = new ActivityInstanceId(resultSet.getLong("activity_id"));
+
+      if (!satisfyingActivities.containsKey(goalId)) satisfyingActivities.put(goalId, new ArrayList<>());
```
mattdailis: This works just fine, but may be worth considering the "computeIfAbsent" method on maps.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostGoalSatisfactionAction.java
```diff
@@ -0,0 +1,42 @@
...
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
+import org.intellij.lang.annotations.Language;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Map;
+
+/*package-local*/ final class PostGoalSatisfactionAction implements AutoCloseable {
```
mattdailis: Did we ever reach a consensus on "Post" vs "Create" nomenclature? I think we had discussed renaming everything to Create in merlin-server, but had never gotten around to it.

EDIT: Or was it Insert?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
mattdailis: Passing thought: we should consider defining the valid values as a postgres enum, so the database will reject any strings that aren't valid.

mattdailis (approved):
Looks good to me - I have no specific changes to request, just a few comments expressing some passing thoughts that should not block merge, especially since there are some tasks that will depend on this PR very soon.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSatisfyingActivitiesAction.java
Twisol: For the particular use here (create a list and add to it), `computeIfAbsent` is truly lovely, since it returns the value whether you had to add it or not.

```java
      satisfyingActivities
        .computeIfAbsent(goalId, () -> new ArrayList<>())
        .add(activityId);
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: Ya know, I actually explored that way back when working on merlin-server's simulation_dataset table, but I was still so new to postgres I didn't feel confident in what I was doing and gave up 😅

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostGoalSatisfactionAction.java
kroffo: We have a mix, and the reason seems to be that I naturally went for `create` when creating a new entity, such as creating a scheduling request, and `post` when inserting data that is associated with an existing thing, such as when posting scheduling results, which are associated with an existing request. Technically these "post" actions are still creating rows of data, and technically the "create" actions are posting data as well. I'm not sure if the distinction is actually beneficial, especially since it's undocumented (just in my head, I suppose). I'm happy to make things more consistent if we agree on an approach. I do feel that `insert` is the neutral ground between `create` and `post` here.

Twisol: FWIW, "post" feels more like an HTTP / REST verb than a database / Postgres verb, though maybe that's just me. "insert" feels more generally DB-y. The verb "create" feels closer to the domain for some cases, but since this Action class is an implementation detail of a Repository (being package-private), maybe "create" makes more sense for the method on the Repository, and "insert" makes more sense for the low-level actions.

My two cents 😅

mattdailis:
I noticed that `LocalSpecificationService` still throws UnsupportedOperationExceptions. It's not strictly speaking part of this PR, since it's not part of "Scheduling Results", but I just want to highlight it since I personally forgot about it until now.

https://github.com/NASA-AMMOS/aerie/blob/edd39ea3767ee578e8c0165bb3ffeaddd90f688d/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/LocalSpecificationService.java#L8-L23

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetRequestAction.java
```diff
@@ -12,11 +10,14 @@
...
-      where
-        req.specification_id = ? and
-        req.specification_revision = ?
+      r.analysis_id,
+      r.status,
+      r.failure_reason,
+      r.canceled
+    from scheduling_request as r
+    where
+      r.specification_id = ? and
```
mattdailis, line 19: 🔻 Oop! We should make a call once and for all between `specification` and `spec`. This column name does not exist.

(My vote is to use the full word everywhere)

mattdailis (changes requested):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetRequestAction.java
mattdailis: There's a similar issue in `GetSpecificationAction`, which refers to the table `scheduling_specification`, which in reality is called `scheduling_spec`.

https://github.com/NASA-AMMOS/aerie/blob/edd39ea3767ee578e8c0165bb3ffeaddd90f688d/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationAction.java#L26

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SetRequestStateAction.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+/*package-local*/ final class SetRequestStateAction implements AutoCloseable {
+  private static final @Language("SQL") String sql = """
+    update scheduling_request
+      set
+        state = ?,
```
mattdailis:
```suggestion
        status = ?,
```

mattdailis (changes requested):
One more small column name mismatch

kroffo:
> > Once the scheduler agent is hooked up we should do some integration testing and make sure a scheduling request can be created and run through our system, even if we have to manually insert the data into the database since the UI isn't hooked up yet.
> 
> Can you please share with me (Wiki page maybe?) the GQL requests you make to insert the data when you run these manual integration tests? It will help me move on the UI more quickly if I have some concrete examples to play with.

I can figure some out for you if you would like, but I haven't been using GraphQL for testing. I've been putting data into the database with TablePlus for now. Typically I'll add a few rows to scheduling_goal, then create a row of scheduling_spec, and add a row to scheduling_spec_goals for each scheduling_goal to associate the goals with the spec.

dyst5422:
Sounds like a great opportunity to flex some more db-tests in.

Sent from my iPhone

On Feb 23, 2022, at 10:51, Kenny Roffo ***@***.***> wrote:

﻿

Once the scheduler agent is hooked up we should do some integration testing and make sure a scheduling request can be created and run through our system, even if we have to manually insert the data into the database since the UI isn't hooked up yet.

Can you please share with me (Wiki page maybe?) the GQL requests you make to insert the data when you run these manual integration tests? It will help me move on the UI more quickly if I have some concrete examples to play with.

I can figure some out for you if you would like, but I haven't been using GraphQL for testing. I've been putting data into the database with TablePlus for now. Typically I'll add a few rows to scheduling_goal, then create a row of scheduling_spec, and add a row to scheduling_spec_goals for each scheduling_goal to associate the goals with the spec.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSatisfyingActivitiesAction.java
kroffo: Sure thing, I never think to use these functions, thanks for the heads up, I'll try to keep them in mind!

mattdailis:
> > Can you please share with me (Wiki page maybe?) the GQL requests you make to insert the data when you run these manual integration tests? It will help me move on the UI more quickly if I have some concrete examples to play with.
> 
> I can figure some out for you if you would like, but I haven't been using GraphQL for testing. I've been putting data into the database with TablePlus for now. Typically I'll add a few rows to scheduling_goal, then create a row of scheduling_spec, and add a row to scheduling_spec_goals for each scheduling_goal to associate the goals with the spec.

I'm collecting a few graphql queries as I try to get an end to end test going to make sure I didn't screw up jar loading in #54 . I'd be happy to share those out once they're in a usable state.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetGoalSatisfactionAction.java
```diff
@@ -0,0 +1,47 @@
...
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+
+/*package-local*/ final class GetGoalSatisfactionAction implements AutoCloseable {
+  private final static @Language("SQL") String sql = """
+    select
+      goal.goal_id,
+      goal.satisfied
+    from schedulign_goal_analysis as goal
```
mattdailis:
```suggestion
    from scheduling_goal_analysis as goal
```

mattdailis (changes requested):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetRequestAction.java
kroffo: I'm going to probably go with specification for all except low-scope things like local variables, or cases of `as spec` for database tables in queries as above. I'll update you here once I've addressed this.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetGoalSatisfactionAction.java
kroffo: Are you having trouble finding time to hangout with friends? Does it constantly seem like every time you have a free moment everyone you know is busy? If you're tired of this problem, then it's time to align! Align your schedules with SCHEDULIGN!

Twisol: report spam

unsubscribe

dyst5422:
These are the threads I came here for.

Sent from my iPhone

On Feb 23, 2022, at 14:27, Jonathan Castello ***@***.***> wrote:

﻿

@Twisol commented on this pull request.

________________________________

In scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetGoalSatisfactionAction.java<https://urldefense.us/v3/__https://github.com/NASA-AMMOS/aerie/pull/52*discussion_r813379717__;Iw!!PvBDto6Hs4WbVuu7!cTiA7yKufMbgGrzIpR80GEQ2NNmYSFdQVREcrg1K5kv45BEYxqyK4xj3YqAvz9ivk0KVI6j3xzdj$>:

> +import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
+import gov.nasa.jpl.aerie.scheduler.server.services.ScheduleResults;
+
+import java.sql.Connection;
+import java.util.HashMap;
+import java.util.Map;
+
+/*package-local*/ final class GetGoalSatisfactionAction implements AutoCloseable {
+  private final static @Language("SQL") String sql = """
+    select
+      goal.goal_id,
+      goal.satisfied
+    from schedulign_goal_analysis as goal


report spam

unsubscribe

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostGoalSatisfactionAction.java
kroffo: It seems that we generally lean toward "insert" here, so I've gone ahead and changed the "Post" actions over to "Insert" actions.

kroffo:
> I noticed that `LocalSpecificationService` still throws UnsupportedOperationExceptions. It's not strictly speaking part of this PR, since it's not part of "Scheduling Results", but I just want to highlight it since I personally forgot about it until now.
> 
> https://github.com/NASA-AMMOS/aerie/blob/edd39ea3767ee578e8c0165bb3ffeaddd90f688d/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/LocalSpecificationService.java#L8-L23

That needed to be implemented, so I added a commit doing so.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetRequestAction.java
kroffo: Alright, I added a commit to the start of this branch changing all instances of `spec` to `specification` in the database files. I stopped at every commit along the way in that rebase to make sure all postgres actions reflected this change.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SetRequestStateAction.java
kroffo: Nice catch!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetGoalSatisfactionAction.java
kroffo: Fixed!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: @pcrosemurgy Check out the latest commit on top of this branch. I implemented a Status enum in the RequestRecord to hide the string-comparison details away in a dedicated spot. This was based off some work done with @patkenneally and @mattdailis on the merlin-server yesterday. If desired, I can squash this commit into the relevant commits on the rest of this branch as well.

kroffo:
Alright, after this morning I've addressed all comments. There's some new commits at both the front and back of this branch, so please check them out! The changes to existing commits were either directly related to a comment above, or an update to make the commit consistent with one of the additional commits.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: Furthermore, I've added a postgres enum `status_t` for the request status, so the only values it can ever take are the ones we allow (else postgres will give an error).

pcrosemurgy: Thanks a lot, this looks great :)

pcrosemurgy (approved):

scheduler-server/sql/scheduler/init.sql
```diff
@@ -5,8 +5,8 @@ begin;
   \ir tables/scheduling_goal.sql
   \ir tables/scheduling_template.sql
   \ir tables/scheduling_template_goals.sql
-  \ir tables/scheduling_spec.sql
-  \ir tables/scheduling_spec_goals.sql
+  \ir tables/scheduling_specification.sql
```
mattdailis, line 8: Glad to see this renaming. We still need to update the public_scheduling_spec.yaml hasura metadata

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
```diff
@@ -1,6 +1,5 @@
 package gov.nasa.jpl.aerie.scheduler.server.remotes.postgres;
 
-import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
```
mattdailis, line 3: The SQL in this file is invalid - there should be no comma after the closing parenthesis, and the table is called `scheduling_goal` singular.

mattdailis (changes requested):
I only got about halfway through my test, but I'll give you these comments now so as not to keep you waiting

scheduler-server/sql/scheduler/init.sql
mattdailis: Update: I've finished my test, and I did not find any more errors in files relevant to this PR. I'll make a separate hotfix PR for the other errors I found 👍

kroffo: Nice catch, should be fixed 👍🏻

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
kroffo: Fixed this in the "Remove unused import" commit (and updated the commit message)

mattdailis (approved):
Great work, Kenny 🚀 , thanks for addressing my concerns.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetRequestAction.java
kroffo: I'll take the spaceship as a sign this is good to go, especially since the issue has been addressed, and Matt reached out to me to ask when I'm going to merge.

PR #54, opened 2022-02-18
[AERIE-1656]  Activities opt-in to controllable duration

mattdailis:
* **Tickets addressed:** AERIE-1656
* **Review:** By commit  

Currently, the scheduler assumes that all activities have a `duration` parameter, which can be used to control their duration. This is not, in general, true.

One option to reconcile this is to have the scheduler inspect the parameters of an activity type, and check if there is a `duration` parameter with the correct type. However, this feels wrong to me, since it ascribes a meaning to a particular parameter name - something we've managed to avoid thus far. Parameter names should be completely up to the mission modelers.

The option I chose to pursue is to add an annotation, `@ControllableDuration(parameterName = "myDurationParam")` that allows an activity definition to opt-in to being controllable by the scheduler. I added a `validateControllableDurationParameter` method to the `MissionModelParser` that checks that the string provided:
1. corresponds to an existing parameter name
2. that parameter has type Duration

The scheduler will be able to call `TaskSpecType::getDurationType` on a particular activity's TaskSpecType, and get back whether the activity has opted-in to having a controllable duration:

```java
public sealed interface DurationType {
  record Controllable(String parameterName) implements DurationType {}
  record Uncontrollable() implements DurationType {}
}
```

The contract between the scheduler and the mission model is that after simulating any activity marked with `@ControllableDuration`, its actual duration is expected to be equal to the value of the declared duration parameter. The scheduler will rely on this fact to satisfy certain goals (such as lining up the end time of one activity with the start time of another activity).

I modified the scheduler to use `getDurationType` to look up what parameter to set for a desired duration. For activities with Uncontrollable duration, the scheduler prints a warning, but does not throw an exception. The main reason here is I do not want to impede uncontrollable activities in existing plans. Future work is to test that the scheduler works with an existing plan.

## Verification
I ran the aerie lander tests, and they passed (though they don't seem to have very clear success criteria, so I want to spend some time in a future PR making the tests stricter).

I'll make a corresponding PR to aerielander momentarily.

## Documentation
This will, of course, need to be documented, but it's still in flux. We have a meeting with some stakeholders next week, where we'll get some feedback.

## Future work
- Focused testing of controllable activity durations
- Testing of the scheduler against existing plans (i.e. not just plans generated from procedural rules, as the tests do now)
- It may be desirable to schedule activities whose duration is uncontrollable. This will place restrictions on the kinds of goals that can place this activity.
- There are potential performance benefits from getting more fine grained than `Uncontrollable`. For example, if the scheduler knew that an activity's duration is "Context Independent", it would know that the activity's duration is invariant under translation in time, which could reduce the necessity to re-simulate.
- It may also turn out to be useful to allow gradations of "Controllable". Some activities cannot uphold the contract of matching the requested duration exactly. Perhaps they have a lower bound for how long they will take. Perhaps the activity can only take multiples of a certain amount of time, so requesting a duration between two multiples would require the activity to decide whether to treat the requested duration as a lower or upper bound. (E.g. take "N" pictures takes "N" times the amount of time it takes to take a picture).

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
@@ -13,8 +13,7 @@ import gov.nasa.jpl.aerie.merlin.processor.metamodel.ParameterValidationRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.EffectModelRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.MissionModelRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.TypeRule;
-
-import org.apache.commons.lang3.tuple.Pair;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
@@ -246,6 +245,12 @@ import java.util.function.Predicate;
     final var parameters = this.getExportParameters(activityTypeElement);
     final var effectModel = this.getActivityEffectModel(activityTypeElement);
 
+    final var durationParameterName = effectModel.flatMap(EffectModelRecord::durationParameter);
+    if (durationParameterName.isPresent()) {
+      validateControllableDurationParameter(name, parameters, durationParameterName.get());
+    }
+
+
     /*
     The following parameter was created as a result of AERIE-1295/1296/1297 on JIRA
     In order to allow for optional/required parameters, the processor
@@ -258,6 +263,37 @@ import java.util.function.Predicate;
     final var defaultsStyle = this.getExportDefaultsStyle(activityTypeElement);
 
     return new ActivityTypeRecord(name, activityTypeElement, parameters, validations, mapper, defaultsStyle, effectModel);
+  }
+
+  private void validateControllableDurationParameter(
+      final String activityName,
+      final List<ParameterRecord> parameters,
+      final String durationParameterName)
+  throws InvalidMissionModelException
+  {
+    final var durationParameterRecord = lookupParameterByName(parameters, durationParameterName);
+    if (durationParameterRecord.isEmpty()) {
+      throw new InvalidMissionModelException(
+          "In activity " + activityName + ", \"" + durationParameterName + "\"" +
+          " is declared as the ControllableDuration parameter, but there is no parameter with that name");
+    }
+    final var record = durationParameterRecord.get();
+    if (!this.typeUtils.isSameType(record.type, this.elementUtils.getTypeElement(Duration.class.getName()).asType())) {
+      throw new InvalidMissionModelException(
+          "In activity " + activityName +
+          ", parameter \"" + record.name + "\"" +
+          " is declared as the ControllableDuration parameter, but does not have type " + Duration.class.getName() + "." +
+          " Instead, it has type " + record.type);
+    }
+  }
+
+  private static Optional<ParameterRecord> lookupParameterByName(final Iterable<ParameterRecord> parameters, final String name) {
+    for (final var param : parameters) {
+      if (param.name.equals(name)) {
+        return Optional.of(param);
+      }
+    }
+    return Optional.empty();
   }
 
   private ExportDefaultsStyle getExportDefaultsStyle(final TypeElement exportTypeElement)
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
@@ -354,12 +354,15 @@ import java.util.function.Predicate;
 
       if (!(element instanceof ExecutableElement executableElement)) continue;
 
+      final var durationTypeAnnotation = element.getAnnotation(ActivityType.ControllableDuration.class);
+      final var durationParameter = Optional.ofNullable(durationTypeAnnotation).map(ActivityType.ControllableDuration::parameterName);
+
       final var returnType = executableElement.getReturnType();
       final var nonVoidReturnType = returnType.getKind() == TypeKind.VOID
           ? Optional.<TypeMirror>empty()
           : Optional.of(returnType);
 
-      return Optional.of(new EffectModelRecord(element.getSimpleName().toString(), executorAnnotation.value(), nonVoidReturnType));
+      return Optional.of(new EffectModelRecord(element.getSimpleName().toString(), executorAnnotation.value(), nonVoidReturnType, durationParameter));
     }
 
     return Optional.empty();
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
@@ -308,10 +308,14 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
                                 activityTypeRecord ->
                                     CodeBlock
                                         .builder()
-                                        .addStatement("result.put(\"$L\", $T.$L())",
+                                        .addStatement("result.put(\"$L\", $T.$L)",
                                                       activityTypeRecord.name(),
                                                       DurationType.class,
-                                                      "uncontrollable"))
+                                                      activityTypeRecord
+                                                          .effectModel()
+                                                          .flatMap(EffectModelRecord::durationParameter)
+                                                          .map(durationParameter -> CodeBlock.of("controllable(\"$L\")", durationParameter))
+                                                          .orElse(CodeBlock.of("uncontrollable()"))))
                             .reduce((x, y) -> x.add("$L", y.build()))
                             .orElse(CodeBlock.builder()).build())
                     .addStatement("return result")
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
@@ -24,6 +24,9 @@ import gov.nasa.jpl.aerie.merlin.processor.metamodel.ExportTypeRecord;
 import gov.nasa.jpl.aerie.merlin.protocol.model.ConfigurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.model.MerlinPlugin;
 import gov.nasa.jpl.aerie.merlin.protocol.model.MissionModelFactory;
+import gov.nasa.jpl.aerie.merlin.protocol.model.SchedulerModel;
+import gov.nasa.jpl.aerie.merlin.protocol.model.SchedulerPlugin;
+import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Parameter;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
@@ -46,7 +49,7 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
 
   /** Generate `GeneratedMerlinPlugin` class. */
   public JavaFile generateMerlinPlugin(final MissionModelRecord missionModel) {
-    final var typeName = missionModel.getPluginName();
+    final var typeName = missionModel.getMerlinPluginName();
 
     final var typeSpec =
         TypeSpec
@@ -67,6 +70,37 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
                     .addStatement(
                         "return new $T()",
                         missionModel.getFactoryName())
+                    .build())
+            .build();
+
+    return JavaFile
+        .builder(typeName.packageName(), typeSpec)
+        .skipJavaLangImports(true)
+        .build();
+  }
+
+  /** Generate `GeneratedSchedulerPlugin` class. */
+  public JavaFile generateSchedulerPlugin(final MissionModelRecord missionModel) {
+    final var typeName = missionModel.getSchedulerPluginName();
+
+    final var typeSpec =
+        TypeSpec
+            .classBuilder(typeName)
+            .addAnnotation(
+                AnnotationSpec
+                    .builder(javax.annotation.processing.Generated.class)
+                    .addMember("value", "$S", MissionModelProcessor.class.getCanonicalName())
+                    .build())
+            .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
+            .addSuperinterface(SchedulerPlugin.class)
+            .addMethod(
+                MethodSpec
+                    .methodBuilder("getSchedulerModel")
+                    .addAnnotation(Override.class)
+                    .addModifiers(Modifier.PUBLIC)
+                    .returns(missionModel.getSchedulerModelName())
+                    .addStatement("return new $T()",
+                                  missionModel.getSchedulerModelName())
                     .build())
             .build();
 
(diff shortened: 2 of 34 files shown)


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/DurationType.java
```diff
@@ -1,5 +1,6 @@
 package gov.nasa.jpl.aerie.merlin.protocol.types;
 
 public sealed interface DurationType {
+  record Controllable(String parameterName) implements DurationType {}
   record Uncontrollable() implements DurationType {}
 }
```
kroffo: I need to get comfortable with this pattern. Sealed interfaces and records still feel pretty new to me, so I tend to not think of them in my own designs too much, but this seems to work well.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/ActivityType.java
```diff
@@ -25,4 +25,10 @@
   @interface EffectModel {
     Executor value() default Executor.Threaded;
   }
+
+  @Retention(RetentionPolicy.CLASS)
+  @Target(ElementType.METHOD)
+  @interface ControllableDuration {
+    String parameterName();
+  }
```
kroffo, line 33: Couldn't the annotation get the parameter name by looking at the parameter declared right below it? Annotations are attached to the line of code below them, aren't they? This is fine for now, but it may be an improvement to explore later if our users find it obnoxious to say the parameter name twice.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -260,6 +264,37 @@ private ActivityTypeRecord parseActivityType(final PackageElement missionModelEl
...
+  throws InvalidMissionModelException
+  {
+    final var durationParameterRecord = lookupParameterByName(parameters, durationParameterName);
+    if (durationParameterRecord.isEmpty()) {
+      throw new InvalidMissionModelException(
+          "In activity " + activityName + ", \"" + durationParameterName + "\"" +
+          " is declared as the ControllableDuration parameter, but there is no parameter with that name");
+    }
+    final var record = durationParameterRecord.get();
+    if (!record.type.toString().equals("gov.nasa.jpl.aerie.merlin.protocol.types.Duration")) {
```
kroffo: Can't you import the Duration class and then compare to `Duration.getClass().getName()` or something like that? As is, if we ever rearrange our package structure here we'll have to change this hardcoded line.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -640,8 +641,19 @@ private record ComputedAttributesCodeBlocks(TypeName typeName, FieldSpec fieldDe
             .build());
   }
 
-  private static MethodSpec makeGetDurationTypeMethod(final ActivityTypeRecord activityType) {
+  private MethodSpec makeGetDurationTypeMethod(final ActivityTypeRecord activityType) {
     final var durationParameter = activityType.effectModel().flatMap(EffectModelRecord::durationParameter);
+    if (durationParameter.isPresent()) {
+      for (final var param : activityType.parameters()) {
+        if (param.name.equals(durationParameter.get())) {
+          if (!param.type.toString().equals("gov.nasa.jpl.aerie.merlin.protocol.types.Duration")) {
```
kroffo: Another instance of the hardcoded full type name, if you decide to tackle the other comment. Resolve when the other comment is resolved.

kroffo (changes requested):
Looks good overall, but I have some minor things I'd like to at least discuss.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/ActivityType.java
Twisol: Not "below them", but you're basically correct. (It's more "to the right", where whitespace is ignored as it it usually is.)

One benefit of a top-level `@ControllableDuration` annotation is that you can only add one of it (by default) to a class. That is, you can't accidentally mark two parameters as controllable durations for the activity. However, I don't think performing this check manually in the processor is complicated enough to tip the scales; I think it'd still be reasonable to associate the annotation more tightly to the the parameter.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
Twisol: You should be able to use `Types#isAssignable` or `Types#isSameType` here; `Types` is one of the top-level utility objects provided to the processor by the compiler.

Twisol:
I'm fine with the user-facing approach (mod Kenny's concerns), but I'm a little nervous about bundling a scheduler-facing capability into the simulation-facing API. There are a bunch of knobs that a scheduler could want to turn regarding a model, and a lot of it comes down to knowing "more" about a model than is necessary to simply run a simulation. I've called this a "scheduling model" in conversation with @patkenneally.

Would it be feasible to extract this `getDurationType` method out into a separate API surface that isn't reachable from the existing `MerlinPlugin` interface? At this point, it could be as simple as a `SchedulerPlugin` with a method `Map<String, String> getDurationControllableTaskTypes()` returning a map from activity type name to parameter name.

Twisol:
(Or, appreciating the modeling work you've done, `Map<String, DurationType>`, where absence in the map should imply `DurationType.Uncontrollable`.)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: Whoops, I don't need to be doing this check twice - I'll go ahead and delete this one.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/ActivityType.java
mattdailis: I agree - the main reason I went with this approach is that we have several different ways to provide parameters (as records, as `@Parameter`, etc), and I wasn't sure how to handle all of them. I'll ping @pcrosemurgy to see if I can get some help on this.

kroffo: I think due to the lack of resources (such as time) we are fine to move forward with your approach, so I'm going to resolve this. After all, this will work just fine and isn't really cumbersome at all.

kroffo (approved):
I do have one comment left to address, but since my workday is at its end I'm going to give my approval on good faith that you'll address the remaining comment (either by incorporating the suggested change, or convincing another team member with some expertise in the related area that an alternative approach is better).

mattdailis:
Force push addresses Kenny's comments, since they were straightforward fixes.

Still mulling over the "interface per client" idea, which I do very much agree with.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
mattdailis: I ended up with this:

```java
this.typeUtils.isSameType(record.type, this.elementUtils.getTypeElement(Duration.class.getName()).asType())
```

Verbose, but robust to moving/renaming `Duration`.

Twisol: Yes, that looks about as verbose as I remember it being. 😅

adrienmaillard:
All of this looks good. I guess I would also add some guards in `ActivityCreationTemplate::createInstanceForReal` to prevent setting an activity instance duration when it is not controllable. And in `ActivityCreationTemplate` builders as well when the user tries to specify a duration parameter.

mattdailis:
> I'm a little nervous about bundling a scheduler-facing capability into the simulation-facing API. There are a bunch of knobs that a scheduler could want to turn regarding a model, and a lot of it comes down to knowing "more" about a model than is necessary to simply run a simulation. I've called this a "scheduling model" in conversation with @patkenneally.
> 
> Would it be feasible to extract this `getDurationType` method out into a separate API surface that isn't reachable from the existing `MerlinPlugin` interface? At this point, it could be as simple as a `SchedulerPlugin` with a method `Map<String, DurationType> getDurationControllableTaskTypes()` returning a map from activity type name to parameter name.

I've taken a stab at this. I've introduced `SchedulerPlugin`:
```java
public interface SchedulerPlugin {
  SchedulerModel getSchedulerModel();
}
```

and `SchedulerModel`:
```java
public interface SchedulerModel {
  Map<String, DurationType> getDurationTypes();
}
```

There is a new generated file, `GeneratedSchedulerModel`, that looks like this:
```java
@Generated("gov.nasa.jpl.aerie.merlin.processor.MissionModelProcessor")
public final class GeneratedSchedulerModel implements SchedulerModel {
  @Override
  public Map<String, DurationType> getDurationTypes() {
    final var result = new HashMap<String, DurationType>();
    final var uncontrollable = new DurationType.Uncontrollable();
    result.put("BiteBanana", uncontrollable);
    result.put("PeelBanana", uncontrollable);
    result.put("ThrowBanana", uncontrollable);
    result.put("GrowBanana", new DurationType.Controllable("growingDuration"));
    result.put("LineCount", uncontrollable);
    return result;
  }
}
```

Ideally (from what I can tell - please contradict me if you have any opinions), the SchedulerPlugin should be a separate service exposed by the mission model jar. I had some trouble setting that up, so in this first cut, the `GeneratedMerlinPlugin` implements both `MerlinPlugin` and `SchedulerPlugin`, and the scheduler casts it to `SchedulerPlugin` at runtime before calling `getSchedulerModel`.

kroffo:
What do you think of using static factory methods i.e.
`static DurationType DurationType.uncontrollable()` and `static DurationType controllable(String paramName)`?

Twisol:
> Ideally (from what I can tell - please contradict me if you have any opinions), the SchedulerPlugin should be a separate service exposed by the mission model jar.

Yes, that's right; the Plugin has no type parameters, while the Model may, and we want it to be very clear that those type parameters are not controllable by the driver, but instead selected by the model. The plugin is just a very thin registered interface that bootstraps access to the provided domain.

Twisol:
> I had some trouble setting that up, so in this first cut, the `GeneratedMerlinPlugin` implements both `MerlinPlugin` and `SchedulerPlugin`,

I think this serves just fine, so long as the JAR contains entries in `META-INF/services/` for both interfaces independently. So long as no important work is done when the plugin is allocated (as is the intent, as a thin registration point), it doesn't much matter if it's allocated twice via different interface lookups. There's no need to change this later unless we have a driving need for scheduler models to be defined without an associated simulation model.

> and the scheduler casts it to `SchedulerPlugin` at runtime before calling `getSchedulerModel`.

I'm less fond of this, but I understand the reason. After this PR goes through, I'd be happy to take a look at removing the assumed downcast myself if you'd like.

mattdailis:
> What do you think of using static factory methods i.e. `static DurationType DurationType.uncontrollable()` and `static DurationType controllable(String paramName)`?

Done :+1: I do like the static factory methods. I kind of wish there was a way to enforce that they are used, instead of using `new`, but that runs counter to how "records" are meant to represent "the data, and only the data".



> All of this looks good. I guess I would also add some guards in `ActivityCreationTemplate::createInstanceForReal` to prevent setting an activity instance duration when it is not controllable. And in `ActivityCreationTemplate` builders as well when the user tries to specify a duration parameter.

Also done :+1: Though @adrienmaillard and I had a conversation and decided that updating `createInstanceForReal` is out of the scope of this PR, since it will require some non-local refactoring of the flow of data through the scheduler.




> > I had some trouble setting that up, so in this first cut, the `GeneratedMerlinPlugin` implements both `MerlinPlugin` and `SchedulerPlugin`,
> 
> I think this serves just fine, so long as the JAR contains entries in `META-INF/services/` for both interfaces independently. So long as no important work is done when the plugin is allocated (as is the intent, as a thin registration point), it doesn't much matter if it's allocated twice via different interface lookups. There's no need to change this later unless we have a driving need for scheduler models to be defined without an associated simulation model.
> 
> > and the scheduler casts it to `SchedulerPlugin` at runtime before calling `getSchedulerModel`.
> 
> I'm less fond of this, but I understand the reason. After this PR goes through, I'd be happy to take a look at removing the assumed downcast myself if you'd like.

I've added the scheduler plugin to `META-INF/services/` for each of our example mission models, and updated the scheduler to load the `SchedulerPlugin` through a separate codepath from the `MerlinPlugin`. This eliminates the need for a cast.

I haven't been able to fully test that the scheduler can load the jar successfully. #52 will enable end-to-end tests, so I think after that is merged, it will be appropriate to try and test the jar loading codepaths.

Twisol:
> I do like the static factory methods. I kind of wish there was a way to enforce that they are used, instead of using `new`, but that runs counter to how "records" are meant to represent "the data, and only the data".

I think making the constructors private would accomplish this, but I think that's too much ceremony for something that, as you say, represents "the data, and only the data". Static factories are just a style/consistency convention in this setting, and I don't know that it does any good to limit the constructors themselves.

> I've added the scheduler plugin to `META-INF/services/` for each of our example mission models, and updated the scheduler to load the `SchedulerPlugin` through a separate codepath from the `MerlinPlugin`. This eliminates the need for a cast.

Thank you for digging into this! I'm glad it wasn't too burdensome :+1:

settings.gradle
```diff
@@ -28,3 +28,4 @@ include 'examples:banananation'
 include 'examples:foo-missionmodel'
 include 'examples:config-with-defaults'
 include 'examples:config-without-defaults'
+include 'aerielander'
```
adrienmaillard: aerielander has not been cleared yet (to the best of my knowledge) so building would fail for someone who would not have it in their directory

adrienmaillard:
> > All of this looks good. I guess I would also add some guards in `ActivityCreationTemplate::createInstanceForReal` to prevent setting an activity instance duration when it is not controllable. And in `ActivityCreationTemplate` builders as well when the user tries to specify a duration parameter.
> 
> Also done 👍 Though @adrienmaillard and I had a conversation and decided that updating `createInstanceForReal` is out of the scope of this PR, since it will require some non-local refactoring of the flow of data through the scheduler.

Looks good. Thank you for improving the quality of the code while you were there.

settings.gradle
mattdailis: Good catch, thanks @adrienmaillard . I will fix this asap. Status on this PR is that I would like to run it through the tests on `prototype/api-tests` described in #67 before merging, since I'm not confident that I got the jar loading logic correct. My goal is to get that done before the end of the day today.

adrienmaillard (approved):

mattdailis:
I have tested the JAR loading, and it seems to work fine. Main note to consider is that the mission models, the merlin server, and the scheduler server, ALL need to be recompiled in order for this to work. (It's an unfortunate side effect of making GeneratedMerlinPlugin implement both MerlinPlugin and SchedulerPlugin).

Removing the `don't merge` label now.

Twisol:
Wait, why do the service components all need to be recompiled? If a component only uses `MerlinPlugin`, they shouldn't be affected, should they?

mattdailis:
> Wait, why do the service components all need to be recompiled? If a component only uses `MerlinPlugin`, they shouldn't be affected, should they?

I stand slightly corrected - it's merlin driver, not merlin server that needs to be recompiled.

I was getting an error to the effect of:
```
java.lang.NoClassDefFoundError: gov/nasa/jpl/aerie/merlin/protocol/model/SchedulerPlugin
```

My hunch is that the class file for the GeneratedMerlinPlugin refers to the SchedulerPlugin interface, which caused it to blow up, since I hadn't recompiled merlin-server or merlin-driver.

Twisol:
Oh, I see what went down. The model JAR doesn't bundle `merlin-sdk`; it expects it to be provided at runtime by the JVM that's loading it. If you don't recompile and redeploy the server, then it doesn't have the version of `merlin-sdk` with the new plugin interface in it, and so the JAR screams.

mattdailis:
> Oh, I see what went down. The model JAR doesn't bundle `merlin-sdk`; it expects it to be provided at runtime by the JVM that's loading it. If you don't recompile and redeploy the server, then it doesn't have the version of `merlin-sdk` with the new plugin interface in it, and so the JAR screams.

Yup. I suppose ideally we would separate the GeneratedMerlinPlugin into two classes to avoid this issue... but I'm not sure it's a real issue. When the protocol updates, everyone needs to be recompiled, and that's okay 🤷 

I guess I'm thinking if we weren't an alpha project I would worry about this more, since we would potentially have different versions of aerie trying to be compatible with each other out in the wild.

Twisol:
I agree with you on the alpha part, but the problem is a little more insidious than that. A model JAR built with both a scheduler plugin and a simulation plugin can only be used in an environment with both interfaces present on the CLASSPATH, even if that environment only needs to use (or be aware of) one of them. The only way around this is to bundle `merlin-sdk` with the model JAR (and pray that the bundled version match the one in the environment) or to deploy completely separate JARs per interface.

Twisol:
Backstepping a little bit, it's only a problem if using one interface would cause the JAR to end up needing the other. This is trivially true because the `GeneratedMerlinPlugin` implements both interfaces, but it can be less obviously true depending on which classes reference what in the JAR. Ugh.

Twisol:
Which, of course, is what you said. My brain is just catching up to yours, apparently.

mattdailis:
I went ahead and split the Generated plugins into two files - there is now a GeneratedSchedulerPlugin alongside the GeneratedMerlinPlugin.

I tested using a newly compiled mission model against a docker container containing an old version of the merlin server, and saw no errors :+1:

As you rightly pointed out, Jonathan, this is a potentially fragile way to split up a design, so in principle we should still expect all clients of merlin-sdk to be recompiled, but in this case the SchedulerPlugin is so limited in scope, and so completely separated from the MerlinPlugin, I think it goes okay.

Twisol:
Aah, that's awesome! I really hope I didn't put you to any trouble there; I wasn't critiquing the PR as it stands so much as reflecting on the difficulties of a plugin architecture. 😅 But that looks great!

mattdailis:
> Aah, that's awesome! I really hope I didn't put you to any trouble there; I wasn't critiquing the PR as it stands so much as reflecting on the difficulties of a plugin architecture. 😅 But that looks great!

It was bothering me personally, so I feel better having fixed it 😅 . I really appreciate your input - especially the idea to make a separate API for scheduler-specific things.

Merging now, as promised to @adrienmaillard to unblock #71 and his work on scheduling activities with uncontrollable durations.

PR #63, opened 2022-02-25
[AERIE-1715]  Typescript evaluation in `scheduler-server`

mattdailis:
* **Tickets addressed:** AERIE-1715
* **Review:** By commit  

## Description

<details>
<summary>See outdated description here</summary>
This PR adds Nashorn as a dependency to `scheduler-server`, and adds a SchedulerDSLCompilationService. Note that here I'm using the word "Compilation" to mean "Compilation of Typescript to Javascript, followed by execution of Javascript to produce JSON, followed by parsing of JSON".

The flow is as follows:
1. Typescript program is provided as a string
2. The program string is prepended with a preface, and followed by an epilogue
3. This combined string is saved to a temporary file
4. `tsc` (the typescript compiler) is run on this temporary file, producing another temporary file
5. The newly generated file is read in as a string
6. This string is executed in Nashorn
7. The epilogue contains a "JSON.stringify()" of the result, so we get back a serialized JSON string
8. This string is run through a JSON parser that we define, the same way we define all of our JSON parsers.

I tried and failed to get a standalone tsc js file that successfully runs under Nashorn. Rather than let that be a blocker, I've gone ahead and installed the typescript compiler in the `scheduler-server` docker container.

Currently, it is assumed that the scheduling goal defines a `goal()` function that returns a particular type. This is in part because I could not figure out how to get modules to work.

</details>

#### New description after many iterations on this draft PR

This PR adds a `scheduling-dsl-compiler` node program that lives alongside the `scheduler-server`. Its job is to turn scheduling goals defined in typescript into their JSON representations, which can then be parsed and converted into `Goal`s that the scheduler can understand.

The `SchedulingDSLCompilationService` is the java class that manages communication with the node subprocess. And instance of this class is created at server startup, and the `close()` method is called on server shutdown.

This PR also lays the groundwork for parsing the JSON representation of the scheduling DSL into scheduler `Goal` objects.

Lastly, this PR adds the ability to generate typescript representing activity types.

This PR supports only RecurrenceGoals that span the entire plan duration. Expanding the scheduling DSL should happen in future PRs.

To do before merging:
- [x] handle errors gracefully. A goal compilation error is no reason to crash the server
- [x] completely remove the jar goals workaround

> **NOTE**: All code that was previously part of this draft PR is still available on the `prototype/AERIE-1715--js-engine-in-scheduler--first-draft` branch. Most notably, there was a lot more of the scheduling DSL defined before - but I deleted it from this PR, since we didn't have all that functionality hooked up yet.

## Verification
I added some unit tests as I went, feel free to check those out and execute them.

> **Note** to run the tests, you'll need to have `tsc` on your path. I did this by running `npm install --global typescript`
> 
> I toyed with using gradle-node-plugin to install typescript for testing purposes, but again, I didn't want to spend all day figuring that out. Any suggestions are welcome!

## Documentation

I'm not aware of any documentation that this would directly affect - perhaps it will deserve mention in our general scheduling EDSL documentation when we get around to writing that.

## Future work
The goal of this PR was to establish the workflow. Next steps are to expand the EDSL definition, provide an actual library, and write a parser that corresponds to that library.


dyst5422:
If we need node for the typescript compiler, why not just run the whole typescript to ast in node and remove the additional Nashorn dependency?

Additionally, we should try to find a way that leaves the typescript compiler running so we don't have to release everything each compilation

mattdailis:
> If we need node for the typescript compiler, why not just run the whole typescript to ast in node and remove the additional Nashorn dependency?

That's a good point - I'll prototype that option and see how we feel about maintaining this functionality in typescript :+1: (one advantage is it will be very similar to command expansion that way)

> Additionally, we should try to find a way that leaves the typescript compiler running so we don't have to release everything each compilation

That's certainly on the table as we consider performance optimizations. In the first cut I think I'd rather just start it up from scratch every time for simplicity's sake.

Regardless of which approach we settle on for this initial version, my main goal with this PR is to "connect the dots" across our end-to-end workflow. To that end, what I think is most important for us to iron out over the next week or two is the details of the API provided by the `SchedulingGoalDSLCompilationService`, which right now is just this method:

https://github.com/NASA-AMMOS/aerie/blob/dde5bf8ada5a4ce279a68612ba0331b0744cb3f0/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingGoalDSLCompilationService.java#L24

The `GoalDefinition` in particular will need to expand to encompass a usable subset of the AST definition, and the `schedulingJsonP` parser will need to be updated accordingly. We'll also eventually need to make activity type information available somehow.

dyst5422:
I’ll take a look/poke at the PR and might do a branch off it to propose some changes addressing some of this.

On Feb 25, 2022, at 08:25, Matt Dailis ***@***.***> wrote:

﻿

If we need node for the typescript compiler, why not just run the whole typescript to ast in node and remove the additional Nashorn dependency?

That's a good point - I'll prototype that option and see how we feel about maintaining this functionality in typescript 👍 (one advantage is it will be very similar to command expansion that way)

Additionally, we should try to find a way that leaves the typescript compiler running so we don't have to release everything each compilation

That's certainly on the table as we consider performance optimizations. In the first cut I think I'd rather just start it up from scratch every time for simplicity's sake.

Regardless of which approach we settle on for this initial version, my main goal with this PR is to "connect the dots" across our end-to-end workflow. To that end, what I think is most important for us to iron out over the next week or two is the details of the SchedulingGoalDSLCompilationService, which right now is just this method:

https://github.com/NASA-AMMOS/aerie/blob/dde5bf8ada5a4ce279a68612ba0331b0744cb3f0/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingGoalDSLCompilationService.java#L24<https://urldefense.us/v3/__https://github.com/NASA-AMMOS/aerie/blob/dde5bf8ada5a4ce279a68612ba0331b0744cb3f0/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingGoalDSLCompilationService.java*L24__;Iw!!PvBDto6Hs4WbVuu7!ftIDug_fpywCsv3jLiQssG7A4prE2iYCd_OEScmAqQP78KGvN0sCji4S8KI7WICbDRW6cp3_aS-h$>

The GoalDefinition in particular will need to expand to encompass a usable subset of the AST definition, and the schedulingJsonP parser will need to be updated accordingly. We'll also eventually need to make activity type information available somehow.

mattdailis:
> I’ll take a look/poke at the PR and might do a branch off it to propose some changes addressing some of this

Sounds good :+1: In my last couple commits, I took a step in that direction by adding a node script that just calls the compiler (twice, because I couldn't figure out how to output error messages from transpileSource). Feel free to start with that, if you find that's a reasonable starting point.

EDIT: Notice that the tests are passing now 😃 . It took some gradle hackery to add node to the path for the test environment.

dyst5422:
Yeah. Do the tests you’ve written cover the calling of the node stuff? If so, I might also play with setting it up as a multiple-call service where we just spit the scheduling script into stdin and snag the ast from stdout to cover the performance initially.

Sent from my iPhone

On Feb 25, 2022, at 12:06, Matt Dailis ***@***.***> wrote:

﻿

I’ll take a look/poke at the PR and might do a branch off it to propose some changes addressing some of this

Sounds good 👍 In my last couple commits, I took a step in that direction by adding a node script that just calls the compiler (twice, because I couldn't figure out how to output error messages from transpileSource). Feel free to start with that, if you find that's a reasonable starting point.

mattdailis:
> Yeah. Do the tests you’ve written cover the calling of the node stuff? If so, I might also play with setting it up as a multiple-call service where we just spit the scheduling script into stdin and snag the ast from stdout to cover the performance initially.

Yup - I slotted the node call into the same place I called `tsc` from before:

https://github.com/NASA-AMMOS/aerie/blob/b19a8505e48030f811835770db23540d4eed4baf/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptExecutionService.java#L70-L73

So main.ts expects to be called like this:
```
node main.js INFILE.ts --outfile OUTFILE.js
```

At the moment it's still running the javascript in `nashorn`. `main.ts` is just a thin wrapper around `ts.transpileModule`.

The tests will use your local node install, if it's on the PATH. If it's not on the PATH, gradle will install its own node executable in `scheduler-server/.gradle/nodejs/node-v14.15.4-darwin-x64/bin/node`.

Realizing as I write this that I should probably force gradle to grab node 16 instead.

scheduler-server/scheduling-dsl-compiler/.gitignore
```diff
@@ -0,0 +1,5 @@
+build/
+node_modules/
+tsconfig.tsbuildinfo
+scheduling-dsl-compiler-log.txt
```
pcrosemurgy: Lines 2 and 3 can be moved to the top-level `.gitignore`, then all git-ignore rules can be centralized and the `build/` rule won't be needed.

Also, why/when is `scheduling-dsl-compiler-log.txt` being generated? Just curious

build.gradle
```diff
@@ -36,11 +36,13 @@ configure(subprojects.findAll { it.projectDir.toPath().resolve('sql').toFile().e
 
   // For all Node subprojects
   sp.plugins.withId('com.github.node-gradle.node') {
-    // Distribute SQL as part of resource processing
-    tasks.create('processResources')
-    tasks.findAll { it.name == 'processResources' }.each { it.dependsOn distributeSql }
-    // Remove distributed SQL as part of `clean` tasks
-    // 'clean' is not visible here, use the subproject .gradle 'clean' task
+    if (!sp.projectDir.toString().contains('scheduler-server')) { // scheduler-server is both a java and a node project
```
pcrosemurgy: Hmm does `if (!sp.plugins.hasPlugin('java'))` not work here?

scheduler-server/build.gradle
```diff
@@ -9,8 +10,48 @@ java {
...
+
 test {
+  dependsOn(assembleSchedulingDSLCompiler)
   useJUnitPlatform()
+  def architectureSpecificDir = new ByteArrayOutputStream()
+  exec{
+    commandLine "ls", projectDir.toPath().resolve(".gradle/nodejs")
+    standardOutput = architectureSpecificDir;
+    ignoreExitValue true
+  }
```
pcrosemurgy, line 48: So `architectureSpecificDir` contains the output from `ls .gradle/nodejs`? Not sure I fully understand what's being done above and below here, could you briefly describe what's needed here? Thanks :)

scheduler-server/build.gradle
```diff
@@ -9,8 +10,48 @@ java {
...
 }
 
+node {
+  nodeProjectDir = file("${project.projectDir.toPath().resolve("scheduling-dsl-compiler")}")
+  version = '16.4.0'
+  npmVersion = '8.5.0'
+  download = true
+}
+
+task assembleSchedulingDSLCompiler(type: NpmTask) {
```
pcrosemurgy, line 20: Minor naming note – I usually prefer to follow camel case for abbreviations/acronyms. Microsoft has a little doc. about this: https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

Basically this helps avoid situations where you have `parse DBM XML` written as `parseDBMXML` by using `parseDbmXml`. Still a bit ugly but personally I find it easier to read/understand.

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -0,0 +1,53 @@
+import Piscina from 'piscina';
+import * as fs from 'fs'
+import * as AST from './libs/scheduler-ast.js';
+
+async function main() {
+
+  var access = fs.createWriteStream('scheduling-dsl-compiler-log.txt', {flags: 'a'});
+
+  // @ts-ignore
```
pcrosemurgy: Can a comment explaining what's being ignored be added here? Thanks :)

scheduler-server/scheduling-dsl-compiler/.gitignore
mattdailis: Will do :+1: 

scheduling-dsl-compiler-log.txt is generated whenever the `scheduling-dsl-compiler/src/main.ts` file is run - I mainly use this log file for debugging the typescript, since as a subprocess of java, its standard error doesn't get output to the console. Right now every single request and response is logged in this file - I might reduce the verbosity to just errors.

`tsconfig.tsbuildinfo` is how typescript keeps track of previously compiled files so it doesn't have to recompile the whole project every time. The project is so small that I think it would be better to disable this functionality, so this file is never created. I'll look into that.

Action items for me:
- [x] Remove build and node_modules from this gitignore
- [x] Remove the need for tsconfig.tsbuildinfo and remove it from this gitignore
- [x] Reduce verbosity in scheduling-dsl-compiler-log.txt

build.gradle
mattdailis: Didn't work for me 😅  Happy to be proven wrong. I think it has to do with the order with which the build.gradle files are evaluated - the top level file gets evaluated before the subprojects do, so hasPlugin returns false... actually I'm not sure that's correct either. 🤷

scheduler-server/build.gradle
mattdailis: Good question - if node is not available on a system, the gradle plugin will install node locally at this location: 

```
scheduler-server/.gradle/nodejs/node-v16.4.0-darwin-x64/bin/node
```

I want to add `scheduler-server/.gradle/nodejs/node-v16.4.0-darwin-x64/bin/` to the PATH, so that java code can just expect `node` to be on the PATH.

Rather than hard-coding `node-v16.4.0-darwin-x64`, I get it via `ls .gradle/nodejs/`.

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -0,0 +1,53 @@
...
+      }) as Promise<{ ast: AST.GoalSpecifier }>;
+      let stringified = JSON.stringify(result);
+      if (stringified === undefined) {
+        throw Error(result + " was not JSON serializable")
+      }
+      console.error("success\n" + stringified + "\n")
+      process.stdout.write("success\n" + stringified + "\n");
+    } catch (error: any) {
+      console.error("error\n" + error.message + '\n')
+      process.stdout.write("error\n" + error.message + '\n');
```
pcrosemurgy: Is there a reason why we're duplicating these log messages? Sorry, may be a silly question

build.gradle
pcrosemurgy: Gotcha. I'll play around with this for a minute, otherwise what you have will have to do. Thanks

scheduler-server/build.gradle
mattdailis: I'm guessing you'll say the same thing about `SchedulingDSL` and `SchedulingDSLCompilationService` 😝

pcrosemurgy: Lol yup, no obligation to change this though. Pros/cons to both conventions.

scheduler-server/scheduling-dsl-compiler/src/main.ts
mattdailis: That's a good question, and pretty key to how this node-java interaction is set up. Java is listening for a response on standard output, so `process.stdout.write` is actually the important functionality here.

`console.error` is just going into the `scheduling-dsl-compiler-log.txt` for debugging purposes.
I think I will get rid of the `console.error`, just because I think it's unnecessarily verbose

scheduler-server/build.gradle
pcrosemurgy: Got it. I think this should be possible without bailing out to the shell. I'll mess with this and try to comment with a block that should work.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+  {
+    final SchedulingDSL.GoalSpecifier goalSpecifier;
+    try {
+      goalSpecifier = schedulingDSLCompilationService.compileSchedulingGoalDSL(
+          planId,
+          pgGoal.definition(),
+          "goals don't have names?");
+    } catch (SchedulingDSLCompilationService.SchedulingDSLCompilationException | IOException e) {
+      e.printStackTrace();
+      throw new Error("");
```
pcrosemurgy: Even if this is an unlikely exception I'd suggest putting some description within the `Error` constructor.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
```diff
@@ -0,0 +1,102 @@
...
+      final var requestJson = reader.readValue();
+      final var result = parser.parse(requestJson);
+      return result.getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)));
+    } catch (JsonParsingException e) {
+      throw new InvalidJsonException(e);
+    }
+  }
+
+  public static class SchedulingDSLCompilationException extends Exception {
+    SchedulingDSLCompilationException(final String message, final Exception e) {
```
pcrosemurgy, line 96: Initially the name of this exception made me think it was a true DSL compilation exception, not a DSL compilation *service* exception. Could maybe just make this `ServiceException` since it's a static class of `SchedulingDSLCompilationService` already. Not sure

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @NotNull
+  private static String generateActivityParameterTypes(final ActivityType activityType) {
+    final var result = new ArrayList<>();
+    for (final var param : activityType.parameters().entrySet().stream().sorted(Map.Entry.comparingByKey()).toList()) {
+      final var name = param.getKey();
+      final var valueSchema = param.getValue();
+      result.add(String.format("%s: %s,", name, valueSchemaToTypescriptType(valueSchema)));
+    }
+    return String.join("\n", result.toArray(new String[0]));
```
pcrosemurgy: No need to change this but this could also be written as:
```java
result.stream().collect(Collectors.joining(System.getProperty("line.separator")))
```
Which is a longer one-liner 😂 ... so maybe nevermind.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
```diff
@@ -0,0 +1,102 @@
...
+  public SchedulingDSLCompilationService(final TypescriptCodeGenerationService typescriptCodeGenerationService) throws SchedulingDSLCompilationException, IOException {
+    this.typescriptCodeGenerationService = typescriptCodeGenerationService;
+    final var schedulingDslCompilerRoot = System.getenv("SCHEDULING_DSL_COMPILER_ROOT");
+    final var schedulingDslCompilerCommand = System.getenv("SCHEDULING_DSL_COMPILER_COMMAND");
+    this.nodeProcess = Runtime.getRuntime().exec(new String[]{"node", schedulingDslCompilerCommand}, null, new File(schedulingDslCompilerRoot));
+
+    try {
+      this.nodeProcess.getInputStream();
+      this.nodeProcess.getOutputStream();
+    } catch (Exception e) {
```
pcrosemurgy: Can you replace this with the expected exception(s) here?

scheduler-server/scheduling-dsl-compiler/.gitignore
pcrosemurgy: >I mainly use this log file for debugging the typescript, since as a subprocess of java, its standard error doesn't get output to the console.

When creating a sub-process you can redirect the stderr to the parent process's stdout/stderr. Can you point me to where this sub-process is being spawned?

scheduler-server/build.gradle
pcrosemurgy: > if node is not available on a system, the gradle plugin will install node locally at this location

And if node is available on the system then `scheduler-server/.gradle/nodejs/node-v16.4.0-darwin-x64` should not exist, right? If so, I think line 51 would result in an exception.

pcrosemurgy: Maybe `.gradle/nodejs/<node>` is generated regardless of whether node is installed on the system or not.

dyst5422: Yeah, I think I ran into this issue at some point and got around it eventually by removing my local node from the path temporarily for it to provoke the install

pcrosemurgy: @mattdailis maybe we can simplify things here by assuming that the user has node@16 installed on their system. After-all, it is a req. for aerie development (https://github.com/NASA-AMMOS/aerie/blob/develop/docs/building.md should probably be updated to reflect this).

mattdailis: > Yeah, I think I ran into this issue at some point and got around it eventually by removing my local node from the path temporarily for it to provoke the install

Hmm, that's not ideal - I was hoping it would "just work" if you already had node.



> @mattdailis maybe we can simplify things here by assuming that the user has node@16 installed on their system. After-all, it is a req. for aerie development (https://github.com/NASA-AMMOS/aerie/blob/develop/docs/building.md should probably be updated to reflect this).

I'd be okay with this - I'd just need your help making sure that when our tests run as a github action, node is available.

pcrosemurgy: Created https://github.com/NASA-AMMOS/aerie/pull/76 which I think should resolve/simplify this.

build.gradle
pcrosemurgy: Created https://github.com/NASA-AMMOS/aerie/pull/77 to generalize this.

scheduler-server/scheduling-dsl-compiler/.gitignore
pcrosemurgy: If you have ~15 mins. tomorrow it may be easiest to just walk through some of this, probably quicker than responding to all the comments I dropped :)

pcrosemurgy (approved):
Nice work on this! Went through changes made with Matt on a call and we resolved the comments I had.

scheduler-server/scheduling-dsl-compiler/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "type": "module",
+  "scripts": {
+    "build": "tsc",
+    "test": "npm run build && node build/main.js",
+    "generate-ast-schema": "npx typescript-json-schema src/libs/scheduler-ast.ts \"*\" --aliasRefs --out ./build/scheduler.schema.json"
+  },
+  "author": "",
+  "dependencies": {
+    "commander": "^9.0.0",
+    "piscina": "^3.2.0",
```
dyst5422: Not needed anymore

mattdailis: Good catch

dyst5422: commander as well i think

mattdailis: Removed. (piscina)

I'll get commander next.

mattdailis: I don't even think we're using source-map yet?

EDIT: nevermind, we are

mattdailis: Latest force push removed commander.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
```diff
@@ -0,0 +1,97 @@
...
+    final var schedulingDslCompilerCommand = System.getenv("SCHEDULING_DSL_COMPILER_COMMAND");
+    this.nodeProcess = new ProcessBuilder("node", schedulingDslCompilerCommand)
+        .redirectError(ProcessBuilder.Redirect.INHERIT)
+        .directory(new File(schedulingDslCompilerRoot))
+        .start();
+
+    final var inputStream = this.nodeProcess.outputWriter();
+    inputStream.write("ping\n");
+    inputStream.flush();
+    if (!"pong".equals(this.nodeProcess.inputReader().readLine())) {
```
kroffo: I know this is pretty minor, but it feel uncomfortable to have the hardcoded string first here. I guess it just feels more natural to say "hey, this thing here should be equal to this value", rather than "this value should be equal to this thing here."

mattdailis: I agree - the main reason to put the string literal first is to avoid a null pointer exception if the value is `null`.

mattdailis: There's a method provided by `Objects` that handles the null case - I'm okay with writing it this way if you find it more readable:

```suggestion
    if (!Objects.equals(this.nodeProcess.inputReader().readLine(), "pong")) {
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public enum GoalKinds {
+    ActivityRecurrenceGoal
+  }
+
+  public sealed interface GoalSpecifier {
+    record GoalDefinition(
+        GoalKinds kind,
+        ActivityTemplate activityTemplate,
+        Duration interval
```
mattdailis, line 87: I want to call out that `interval` only applies to recurrence goals. We'll need to split out `GoalDefinition` into variants when we add more goal types to the DSL.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -0,0 +1,164 @@
...
+      if (type instanceof TSString) {
+        return "string";
+      } else if (type instanceof TSDouble) {
+        return "AST.Double";
+      } else if (type instanceof TSBoolean) {
+        return "boolean";
+      } else if (type instanceof TSInt) {
+        return "AST.Integer";
+      } else if (type instanceof TSDuration) {
+        return "AST.Duration";
```
mattdailis: Latest force push adds the `AST.` prefix to `Integer` and `Duration`. @dyst5422 is there a more preferable way to refer to these types? I could import them all explicitly from `scheduler-ast.ts`...?

dyst5422: We can supply these to the global typespace as just `Integer`, `Duration` etc and not have to prefix them, which I think is a better option

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
```diff
@@ -0,0 +1,70 @@
...
+export type I32 = I<32>;
+export type I64 = string;
+export type VarString<PrefixBitLength extends number, MaxBitLength extends number> = string;
+export type F<BitLength extends 32 | 64> = number;
+export type F32 = F<32>;
+export type F64 = F<64>;
+export type Duration = F64;
+export type Integer = I32;
+export type Double = F64;
+export type Time = string;
```
dyst5422: Wrap these in a `declare global {...}` and remove the export and they will be in the global typespace

mattdailis: Done. Upon further reflection, we probably don't need all of these types - my impression is that numeric types at the scheduler level will be simpler (basically just `Integer` and `Double`), unlike the command dictionaries which care a lot about how many bytes the numbers take up

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
kroffo: That's a good point, I hadn't thought of that. I do like the `Objects.equals` way.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
dyst5422: Agreed. If we don't need adherence on the scheduler side of things, thats fine. We should check number ranges coming back for integers to ensure they are between `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER` to avoid any issues of unexpected truncation/rounding

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
mattdailis: Done. Latest force push uses the proposed `Objects.equals` instead of string literal `.equals()`.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/SpecificationRepository.java
```diff
@@ -3,10 +3,12 @@
...
 import gov.nasa.jpl.aerie.scheduler.server.models.Specification;
 import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
+import gov.nasa.jpl.aerie.scheduler.server.remotes.postgres.PostgresSpecificationRepository;
 import gov.nasa.jpl.aerie.scheduler.server.services.RevisionData;
 
 public interface SpecificationRepository {
   // Queries
-  Specification getSpecification(SpecificationId specificationId) throws NoSuchSpecificationException;
+  Specification getSpecification(SpecificationId specificationId)
+  throws NoSuchSpecificationException, PostgresSpecificationRepository.GoalBuildFailureException;
```
kroffo: It seems like we probably shouldn't be putting implementation details onto the interface. Could we at least maybe make a more generic exception that the `GoalBuildFailureException` extends? Something like `SpecificationLoadException`?

mattdailis: Sure :+1: Yeah, I wasn't sure which part is an implementation detail - since it is known that Specifications have goals. I'll do as you suggested though, brb

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -108,13 +109,13 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
       //collect results and notify subscribers of success
       final var results = collectResults(solutionPlan,instancesToIds, goals);
       writer.succeedWith(results);
-    } catch (final ResultsProtocolFailure | NoSuchSpecificationException e) {
+    } catch (final ResultsProtocolFailure | NoSuchSpecificationException | PostgresSpecificationRepository.GoalBuildFailureException e) {
```
kroffo: If the exception on the interface is updated, this one should be too.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/SpecificationRepository.java
mattdailis: Done.

kroffo: It's not the goal part that I had a problem with, it's the PostgresSpecificationRepository being mentioned in the SpecificationRepository interface.

mattdailis: Right - it should be gone now :+1: I got rid of the GoalBuilder-specific exception, and just included a message in the `SpecificationLoadException`. Would it make sense to have both?

kroffo (approved):
You've already addressed all the issues I had commented on along my review, so I can now approve 👍🏻

dyst5422 (approved):

PR #64, opened 2022-02-27
[REFACTOR] Move lorem_ipsum.txt into the package hierarchy

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
When viewing the project in IntelliJ's "Packages" view (my preferred Project view), the `data/lorem_ipsum.txt` resource is shown at the top level. This is because it's located directly under the `src/main/resources/` directory, rather than further nested into packages like the parallel source hierarchy.

This misplacement has a knock-on effect in the code, in that an extra `getClassLoader()` call is necessary in order to address resources relative to the root of the package hierarchy. By moving the resource into the same package as the `SimulationUtility` class that references it, we can use the class's own (reflective) `getResource()` method, with a path relative to the class's own position in the hierarchy.

Lastly, I noticed that this data file is only referenced from `test` code, but the resource is included in `main`. I moved the file to the `src/test/resources/` tree, so this file is no longer included in built Banananation JARs.

## Verification
This change was verified by (a) observing an exception when the file had been moved without adjusting the code (so, we know it throws if it can't find it), and then (b) observing no exception after adjusting the code. I used the following snippet (in `SimulationUtility`) before and after to give more evidence that it was finding the correct file.

```java
    try {
      System.out.println(dataPath);
      System.out.println(Files.size(dataPath));
    } catch (final IOException ignored) {}
```

## Documentation
No documentation impact; this is just a matter of the internal organization of the Banananation model's own files.

## Future work
None (I hope).


Twisol:
\*sigh\* Note to self, don't accidentally hold Cmd while trying to add a newline; it submits the PR.

Apologies; editing. (**EDIT**: Finished editing.)

pcrosemurgy (approved):
Thank you! Much appreciated ☺️

patkenneally (approved):

PR #65, opened 2022-02-28
Typescript service generates AST

dyst5422:
The typescript will now generate the AST in full before sending back to
Java. Additionally, it holds it open as a long running process to pipe
data into and out of to avoid the overhead of loading the
typescript compiler/node executable multiple times.

* **Tickets addressed:** N/A
* **Review:** By commit

## Description
Tried to get the typescript to do the full typescript -> AST conversion in a long running process that responds to multiple stdin inputs of the form `{ "source": string, "filename": string }`.

## Verification
I wasn't able to get the tests to run through and am hoping to get some help from @mattdailis to finish this out. Seems like they are hanging, so I no doubt messed up the interop between node and Java. Tests will need to be updated when we resolve this to handle the new return format.

I did a command line test of the node service with 
```
npm run build && node build/main.js
{"source": "export default function myGoal() { return Goal.ActivityRecurrenceGoal({ windowSet: WindowSet.entirePlanWindow, activityTemplate: ActivityTemplates.PeelBanana('some goal', { peelDirection: 'fromStem' }), rangeToGenerate: [1,1]}) }", "filename": "somefile"}
```

## Documentation
None

## Future work
Complete this and get it into @mattdailis's main branch for this ticket.
Add in documentation autogeneration for the eDSL api for users (using typedoc)


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingGoalDSLCompilationService.java
```diff
@@ -7,51 +7,75 @@
...
+    var outputReader = new BufferedReader(new InputStreamReader(this.inputStream));
+    var errorReader = new BufferedReader(new InputStreamReader(this.errorStream));
 
-    final var executionResult = executeTypescript(preface, goalTypescript, epilogue);
-    if (!(executionResult instanceof final String executionResultString)) {
-      throw new SchedulingGoalDSLCompilationException("Expected result of evaluating typescript to be a string, but got " + executionResult);
+    inputWriter.write("{ \"source\": \"" + goalTypescript + "\", \"filename\": \"" + goalName + "\" }\n");
+
+    var error = errorReader.readLine();
+    var output = outputReader.readLine();
```
mattdailis, line 65: I poked around at it this morning, and isolated the issue to these three lines.

First, I needed to add a `inputWriter.flush()` call after the write - the subprocess was never triggering its `on('data')` callback.

Next, there's an issue that I couldn't figure out how to fix without changing the protocol - both `readLine` calls are blocking, but only one of them is expected to actually have data (the subprocess will _either_ write to standard out, or to standard error).

The solution that worked for me was to change the protocol to make the subprocess always write to standard out, but it would always write two lines, the first of which was either "success" or "error".

Would you like me to commit these changes directly to your branch? Or merge this one into mine and keep working from there?

My only gut concern about the long-running subprocess approach is whether we'd ever have to restart it if it craps out for some reason. We can flesh that scenario out in some more testing.

dyst5422: Let's merge this into yours and work from there. Do you have time for some pair programming on this today?

PR #66, opened 2022-02-28
Model configuration interface cleanup

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR comes in three pieces:

* The first six commits are basic cleanups I noticed while doing other things, grouped together at the front. The largest is the first commit, an imports cleanup across the entire (Java) codebase using IntelliJ's "Optimize imports" automation. (It originally contained just a few imports fixes I noticed myself, but I realized I could hit everything pretty easily.)
* The next commit ensures that custom mappers (declared with the `@WithMapper` action) implement the `TaskSpecType` interface that generated code relies on. If a user made use of custom mapper at this point, it's likely that it was compile-time broken, so I suspect nobody's using this functionality. But now they can, again, if they need to.
* The last seven commits smooth out some of the rough edges in the current support for model configuration.
    * Instead of an potentially-missing `Optional` configuration type, the framework now provides a configuration type for `VoidEnum`, representing an absence of configuration. (This is a good example of the [empty object pattern](http://www.javapractices.com/topic/TopicAction.do?Id=59).)
    * Config deserialization is now the driver's responsibility; `MissionModelFactory#instantiate` now expects a concrete instance of its configuration type, just like `TaskSpecType#instantiate`.
    * `MissionModelFactory#getParameters` no longer exists, as the same information can be obtained through `getConfigurationType()`.

Most of the sites that used `MissionModelFactory#instantiate` were test support code. Previously, these would have to construct a configuration object and then serialize it. Because of the changes to `#instantiate`, this extra step is no longer necessary.

Some branching logic in the annotation processor, especially in the implementation of `MissionModelFactory#instantiate`, have been drastically simplified, since all models have an associated `ConfigurationType`, regardless of whether the top-level model actually accepts configuration. (If it does not, its `ConfigurationType` is simply `EmptyConfigurationType`.)

## Verification
I ran the `SimulateMapSchedule` test driver and observed a successful run. Most changes were the type that the compiler would fail; some things that were not covered by the compiler (or by the `SimulateMapSchedule` run) might include:
* Code generation paths not exercised by the example models, such as models that don't accept configuration (we have none!), or activity types that use `@WithMapper` (we have none!)
* Configuration deserialization through the Merlin server. This logic was only relocated, and the changes were such that the compiler should catch any issues, but I have not actually exercised that code path.

## Documentation
* Our user documentation for `@WithMapper` likely deserves some attention, seeing as the functionality may have been broken for a while.

## Future work
* The annotation processor has always been hairy, but it's becoming quite difficult to trace the origins of certain data, and code duplication is becoming a problem. I've punted on this a bit (and maybe added to the duplication issue), but it would be good to set in place some architectural support for coordinating information derived from the input data. In particular, viewed through a web application lens, our views (JavaPoet templates) have too much logic and derived state, and most computations should be factored out of the view layer entirely.


pcrosemurgy:
Awesome, I'll look through this in more detail later but just wanted to say thanks @Twisol!

>Code generation paths not exercised by the example models, such as models that don't accept configuration (we have none!), or activity types that use @WithMapper (we have none!)

True, we should really add some activity/config. that tests this path out.

Twisol:
This PR will conflict in interesting ways with #86. They're independent PRs, but they touch similar regions of code, so care must be taken when rebasing one once the other lands.

pcrosemurgy:
Haven't forgotten about this btw, hoping to have some time to review this by end of next week.

Twisol:
No prob. Thanks for the heads-up!

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -586,6 +513,13 @@ private record ComputedAttributesCodeBlocks(TypeName typeName, FieldSpec fieldDe
     return generateCommonMapperMethods(missionModel, activityType).map(typeSpec -> typeSpec.toBuilder()
         .addMethod(makeGetReturnValueSchemaMethod())
         .addMethod(makeSerializeReturnValueMethod(activityType))
+        .addMethod(
+            MethodSpec
+                .methodBuilder("getName")
```
mattdailis: 🔶  Would it be possible to add an `@Override` to this `getName` method? It does override the ActivityMapper interface after all.

Speaking of which - would it make sense to reword this commit to say `Generate a getName method only for activity mappers`, since `getName` is not part of `TaskSpecType` anymore?

I am curious why this method is necessary at all, since it's only used from generated code, and the name of the activity is known statically. I suppose it's good to have a complier-checkable "single source of truth" rather than generating a lot of string literals...

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
```diff
@@ -33,17 +33,22 @@
         return loadMissionModel(missionModelConfig, factory, builder);
     }
 
-    private static <Model>
+    private static <Config, Model>
```
mattdailis, line 36: I accidentally made this method public in #54 , when rebasing you can just ignore my changes here.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -100,6 +98,9 @@ public JavaFile generateMissionModelFactory(final MissionModelRecord missionMode
             .addSuperinterface(
                 ParameterizedTypeName.get(
                     ClassName.get(MissionModelFactory.class),
+                    missionModel.modelConfigurationType
+                        .map($ -> ClassName.get($.declaration()))
+                        .orElse(ClassName.get(gov.nasa.jpl.aerie.merlin.framework.VoidEnum.class)),
```
mattdailis, line 138: Cool use of VoidEnum

mattdailis (approved):
Approving - though I do have request with the tag:
🔶 _Requests that should not block merge, but should at least be discussed_

I rebased this branch locally - let me know if you'd like me to push that up, or if you'd prefer to do the rebase yourself. In particular, I'm not sure how you did the project-wide optimize imports - I dropped that commit, and re-did it by right clicking on each java subproject and clicking "Optimize imports". I pushed the rebased version up to `feature/model-config-refactor--dailis-rebase`

While I am in favor of broad sweeping clean-ups, they are a recipe for merge conflicts with any ongoing work. Fortunately I have no ongoing java work at this exact moment, but @adrienmaillard does, so it may be good to get his consent before making him rebase onto this PR 😅

Twisol:
> In particular, I'm not sure how you did the project-wide optimize imports - I dropped that commit, and re-did it by right clicking on each java subproject and clicking "Optimize imports".

That's exactly how I did it, yes :+1: If that's the only thing you changed in your rebase, I'm happy for you to push your branch; it's definitely more likely to be right than whatever Git does automatically to interleave changes.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
Twisol: > Would it be possible to add an `@Override` to this `getName` method? It does override the ActivityMapper interface after all.

Ah, good point. Yes, that would be possible.

> Speaking of which - would it make sense to reword this commit to say `Generate a getName method only for activity mappers`, since `getName` is not part of `TaskSpecType` anymore?

True! I can fix this after you've re-pushed your imports.

Twisol: > I am curious why this method is necessary at all, since it's only used from generated code, and the name of the activity is known statically. I suppose it's good to have a complier-checkable "single source of truth" rather than generating a lot of string literals...

Only used from generated code, but might have to be implemented by user code. We allow users to define custom activity mappers, and I'd like them to be able to code to an interface rather than having to know somehow that there also needs to be a magical `public static final String NAME`.

Twisol: Amusingly, `getName` will drop out entirely in #86, so if you'd like we can just remove these changes and let #86 deal with `getName` more holistically. (EDIT: Well, nothing *uses* it in #86; I'm just missing a commit there to remove it and its generation code.)

Twisol: > and the name of the activity is known statically.

Ohh, I could just do the thing I did with the `Map` in `ActivityTypes` to statically use the `name` associated with the activity type. That's totally true, and it would be independent of whether an activty type has a user-defined mapper. Well, see above; I think shifting this change to #86 will be more holistic anyway!

Twisol: Okay, I've removed a total of five commits relating to `getName`, and will move the removal of `getName` to #86, where it will amount to one commit of deleting code. 🚀

Deleted commits:

```
pick 7355e78d7 Remove intermediate list in generated ActivityTypes
pick 97bd2f9d8 Use statically-known activity type name
pick 17fa2bede Remove an unnecessary use of TaskSpecType::getName
pick 67a969b22 Remove getName from ConfigurationType and TaskSpecType
pick 3301533d2 Generate a getName method only for task spec types
```

Twisol:
I've removed some commits thanks to @mattdailis' eagle eye, and have updated the PR summary in turn. (If you're already read the PR summary, I've only removed information relevant to the removed commits.)

patkenneally (approved):
I've taken a note on the refactoring to reduce the duplication in the annotation's processor.

Twisol:
@patkenneally, am I clear to merge once I rebase and check that all commits build? 😃

patkenneally:
Yes please feel free to merge when you like.

PR #69, opened 2022-02-28
[HotFix] Can't copy sql from node projects

goetzrrGit:
* **Review:** By commit  

## Description
For a strange reason, Gradle is unable to see the command-expansion-server's `build.gradle` tasks. This causes the Aerie `build.gradle` to ignore moving the SQL files over. I tweaked Aerie `build.gradle` a little to get it working. 

## Verification
run `./gradlew clean` and verify `aerie/deployment/postgres-init-db/sql/commanding` is removed
run `./gradlew assemble` and verify `aerie/deployment/postgres-init-db/sql/commanding` is added

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,9 +37,10 @@ configure(subprojects.findAll { it.projectDir.toPath().resolve('sql').toFile().e
   // For all Node subprojects
   sp.plugins.withId('com.github.node-gradle.node') {
     // Distribute SQL as part of resource processing
+    tasks.create('processResources')
     tasks.findAll { it.name == 'processResources' }.each { it.dependsOn distributeSql }
     // Remove distributed SQL as part of `clean` tasks
-    tasks.findAll { it.name == 'clean' }.each { it.dependsOn undoDistributeSql }
+    // 'clean' is not visible here, use the subproject .gradle 'clean' task
   }
 
   // Distribute SQL prior to creating deployment archive
--- a/command-expansion-server/build.gradle
+++ b/command-expansion-server/build.gradle
@@ -8,9 +8,6 @@ node {
   download = true
 }
 
-task processResources {
-  // Currently does nothing, exists as a hook for `distributeSql`
-}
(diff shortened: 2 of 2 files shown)


camargo:
With this change, do the `merlin-sever` and `scheduler-server` packages need a clean in their `build.gradle`'s as well?
Tagging @pcrosemurgy

camargo (approved):
Tested locally. Building and clean up works throughout! Thanks for the quick fix.

pcrosemurgy (changes requested):
>Gradle is unable to see the command-expansion-server's build.gradle tasks. This causes the Aerie build.gradle to ignore moving the SQL files over.

I'm not sure I'm able to replicate this. With `gradle distributeSql` I see:
```
deployment/postgres-init-db/sql
├── commanding
│   ├── tables
│   │   ├── activity_instance_commands.sql
│   │   ├── command_dictionary.sql
│   │   ├── expansion_rules.sql
│   │   ├── expansion_run.sql
│   │   ├── expansion_set.sql
│   │   └── expansion_set_to_rule.sql
│   └── init.sql
```

**Edit**: am able to replicate this when using `gradle processResources`. Very confused, the exact same test I did in https://github.com/NASA-AMMOS/aerie/pull/60 is no longer working.

goetzrrGit:
@pcrosemurgy `gradle distributeSql` does work but when you run `./gradlew assemble`, the  `distrubitSql`  isn't being called. Also `gradle clean` doesn't remove the commanding SQL files.

This is because the node's `processResource` isn't visible to aeries's build.gradle so it isn't being invoked. My code change makes the processResource visible and allows for it to work.

pcrosemurgy (approved):
Tested this out and it looks good to me. Thanks a lot @goetzrrGit!

Really crazy that the only way (that I/we know of) to accomplish this is by dynamically creating a `processResources` task. That's like dynamically generating a Makefile target upon building.

PR #70, opened 2022-02-28
[AERIE-1717]  Upload a command dictionary

goetzrrGit:
* **Tickets addressed:** AERIE-1717
* **Review:** By commit  

## Description
The first part of the command expansion servicer. This will allow the user to upload a command dictionary, parse, and create the command-lib.ts. The path to this file will be saved in the command_dictionary table. 

There will be a separate PR for the Aerie-Gateway, which you can use to upload the command dictionary via the gateway. It will not affect this PR though.

- [x] Add endpoint to upload the command dictionary. Future work: move to the Aerie Gateway
- [x] Import command dictionary parse library to parse the command dictionary
- [x] Generate command-lib.ts and save to the filesystem
- [x] Save (path, version, and mission) to Postgres
- [x] Support upload of the same command dictionary file

## Verification

1.  ` docker system prune --all --volumes`
2.  `./gradlew clean; ./gradlew assemble;  docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up`
3.  Download the command dictionary: https://jira.jpl.nasa.gov/browse/AERIE-1633
4.  Copy the XML content to https://www.base64encode.org/ and encode to `base64`
5.  In Hasura use the graphQL to send the command dictionary to the command service

Take the base64 string and add it to the call at `<INSERT>`
```
mutation MyMutation {
  uploadDictionary(dictionary: "<INSERT>") {
    id
  }
}
```
6.  Verify the query returns an ID to the Postgres table
7.  Look at Hasura and verify you see the entries in the command-dictionary table

## Documentation

## Future work
[AERIE-1718](https://jira.jpl.nasa.gov/browse/AERIE-1718) Create an activity-type library from a mission model.

### DIFF
--- a/command-expansion-server/package.json
+++ b/command-expansion-server/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "type": "module",
   "scripts": {
-    "build": "tsc;",
+    "build": "rm -rf ./build/packages/lib/CommandEDSLPreface.ts; tsc; cp ./src/packages/lib/CommandEDSLPreface.ts ./build/packages/lib/CommandEDSLPreface.ts;",
     "start": "node ./build/app.js",
     "ts-watch": "tsc --watch",
     "dev-server": "supervisor -- --inspect=0.0.0.0 ./build/app.js",
--- a/command-expansion-server/sql/commanding/tables/command_dictionary.sql
+++ b/command-expansion-server/sql/commanding/tables/command_dictionary.sql
@@ -5,9 +5,9 @@ create table command_dictionary (
   mission text not null,
   version text not null,
 
-  constraint command_dictionary_primary_key
+  constraint command_dictionary_synthetic_key
       primary key (id),
-  constraint pair unique (mission,version)
+  constraint command_dictionary_natural_key unique (mission,version)
 );
 
 comment on table command_dictionary is e''
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -47,9 +47,12 @@ app.post("/dictionary", async (req, res) => {
 
   if (rows.length < 0) {
     console.error(`POST /dictionary: No command dictionary was updated in the database`);
+    res.status(400).json({ message: `POST /dictionary: No command dictionary was updated in the database` });
+  } else {
+    const id = rows[0].id;
+    res.status(200).json({ id });
   }
-  const id = rows[0].id;
-  res.status(200).json({ id });
+  return;
 });
 
 app.put("/expansion/:activityTypeName", async (req, res) => {
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -9,35 +9,35 @@ import { processDictionary } from "./packages/lib/CommandTypeCodegen.js";
 const PORT: number = parseInt(getEnv().PORT, 10) ?? 3000;
 
 const app: Application = express();
-app.use(bodyParser.json({ limit: '25mb' }));
+app.use(bodyParser.json({ limit: "25mb" }));
 
 DbExpansion.init();
 const db = DbExpansion.getDb();
-
 
 app.get("/", (req: Request, res: Response) => {
   res.send("Aerie Command Service");
 });
 
-app.put("/dictionary", async (req, res) => {
-  const dictionary = Buffer.from(req.body.input.dictionary, 'base64').toString();
-
-
+app.post("/dictionary", async (req, res) => {
+  const base64Dictionary: string = req.body.input.dictionary;
+  const dictionary = Buffer.from(base64Dictionary, "base64").toString("utf8");
   console.log(`Dictionary received`);
+
   const parsedDictionary = ampcs.parse(dictionary);
   console.log(
     `Dictionary parsed - version: ${parsedDictionary.header.version}, mission: ${parsedDictionary.header.mission_name}`
   );
+
   const commandDictionaryPath = await processDictionary(parsedDictionary);
   console.log(`command-lib generated - path: ${commandDictionaryPath}`);
 
   const sqlExpression = `
-    insert into command_dictionary (command_types, mission, version)
-    values ($1, $2, $3)
-    on conflict (mission, version) do update
-    set command_types = $1
-    returning id;
-  `;
+      insert into command_dictionary (command_types, mission, version)
+      values ($1, $2, $3)
+      on conflict (mission, version) do update
+      set command_types = $1
+      returning id;
+    `;
 
   const { rows } = await db.query(sqlExpression, [
     commandDictionaryPath,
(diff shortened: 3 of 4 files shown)


command-expansion-server/src/packages/ampcs/parser.ts
```diff
@@ -0,0 +1,553 @@
+import { xml2js } from "xml-js";
```
camargo: This effectively open sources this library. Can we just add it as a new repo to NASA-AMMOS and throw it up on NPM? It is also used in our sequence editor so I don't think we should be duplicating the code like this if we can help it.

goetzrrGit: True, I just wanted something to build temporary and possibly merge to develop so we don't block other parts of the command expansion service. 

We can remove it later with the library once it is on `NASA-AMMOS`

@patkenneally what do you think? Maybe this is something that someone on the team can do in parallel?

camargo: I am happy to do this. Can do it in 10 minutes. Just say the word :)

goetzrrGit: oh, I thought it was a long process? Not sure if you need @patkenneally approval but I am happy if you do it now.

camargo: Yeah it should be a very quick process once we have the green light. Copying the project to a new repo in NASA-AMMOS. And then pushing it to the @nasa-jpl NPM org (which I already have access to).

command-expansion-server/src/app.ts
```diff
@@ -13,6 +19,46 @@ app.get("/", (req: Request, res: Response) => {
...
+    const base64Dictionary: string = req.body.input.dictionary;
+    const dictionary = Buffer.from(base64Dictionary, "base64").toString("utf8");
+    console.log(`Dictionary received`);
+
+    const parsedDictionary = ampcs.parse(dictionary);
+    console.log(
+      `Dictionary parsed - version: ${parsedDictionary.header.version}, mission: ${parsedDictionary.header.mission_name}`
+    );
+
+    const commandDictionaryPath = await Promise.resolve(processDictionary(parsedDictionary));
```
dyst5422: The Promise.resolve isn't needed here as processDictionary returns a promise. And in fact the additional resolved promise its wrapped in could delay its return by an additional event loop.

command-expansion-server/data/preface.txt
```diff
@@ -0,0 +1,66 @@
+/** START Preface */
```
dyst5422: Lets move this file into src and rename it to CommandEDSLPreface.ts so that we can import the types from it for type checking the expansion execution code down the line - we need to adjust the command codegen to pick up that new file instead of the txt one

command-expansion-server/package.json
```diff
@@ -14,16 +14,18 @@
   "license": "ISC",
   "dependencies": {
     "@microsoft/tsdoc": "^0.13.2",
+    "@nasa-jpl/aerie-ampcs": "^1.0.0",
     "express": "^4.17.2",
-    "multer": "^1.4.4",
+    "graphql": "^16.3.0",
```
dyst5422: This is unused, so lets remove it

command-expansion-server/sql/commanding/tables/command_dictionary.sql
```diff
@@ -6,7 +6,8 @@ create table command_dictionary (
   version text not null,
 
   constraint command_dictionary_primary_key
-      primary key (id)
+      primary key (id),
+  constraint pair unique (mission,version)
```
dyst5422: lets rename this to `command_dictionary_synthetic_key` like our other tables

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -0,0 +1,216 @@
...
+
+function versionBuilder(dictionary: ampcs.CommandDictionary) {
+  return `output/command_lib_${dictionary.header.mission_name.toLowerCase()}_${dictionary.header.version}.ts`;
+}
+
+export async function processDictionary(dictionary: ampcs.CommandDictionary) {
+  const prefaceUrl = new URL("../../../data/preface.txt", import.meta.url);
+  const folder = `${getEnv().STORAGE}/${dictionary.header.mission_name.toLowerCase()}/`;
+  const fileName = `command_lib.${dictionary.header.version}.ts`;
+  const prefaceString = await Promise.all([fs.promises.readFile(prefaceUrl.pathname, "utf8")]);
```
dyst5422: Promise.all not needed here

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -0,0 +1,216 @@
...
+      throw new Error(`Unsupported argument type: ${argument.arg_type}`);
+  }
+}
+
+function versionBuilder(dictionary: ampcs.CommandDictionary) {
+  return `output/command_lib_${dictionary.header.mission_name.toLowerCase()}_${dictionary.header.version}.ts`;
+}
+
+export async function processDictionary(dictionary: ampcs.CommandDictionary) {
+  const prefaceUrl = new URL("../../../data/preface.txt", import.meta.url);
```
dyst5422: Ensure we update this location

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -0,0 +1,216 @@
...
+  const folder = `${getEnv().STORAGE}/${dictionary.header.mission_name.toLowerCase()}/`;
+  const fileName = `command_lib.${dictionary.header.version}.ts`;
+  const prefaceString = await Promise.all([fs.promises.readFile(prefaceUrl.pathname, "utf8")]);
+
+  const { values, declarations } = generateTypescriptCode(dictionary); //update sql table store seprate
+
+  await fs.promises.mkdir(folder, { recursive: true });
+
+  return await Promise.resolve([
```
dyst5422: promise.resolve not needed/undesirable here

dyst5422 (changes requested):

command-expansion-server/data/preface.txt
goetzrrGit: It was added to packages/lib/CommandEDSLPreface.ts so I can use that path

command-expansion-server/package.json
goetzrrGit: Once I rebased with `develop` it was removed

command-expansion-server/src/app.ts
goetzrrGit: I fixed this, Thanks

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -199,7 +199,7 @@ function versionBuilder(dictionary: ampcs.CommandDictionary) {
 }
 
 export async function processDictionary(dictionary: ampcs.CommandDictionary) {
-  const prefaceUrl = new URL("./CommandEDSLPreface.ts", import.meta.url);
+  const prefaceUrl = new URL("./packages/lib/CommandEDSLPreface.js", import.meta.url);
```
dyst5422: Hmmm....interesting problem I didn't think about...we want the typescript version of this for type checking but the output format that would be relative to the file being executed at runtime would be a .js file ...hmmm

command-expansion-server/src/app.ts
```diff
@@ -9,93 +9,94 @@ import { processDictionary } from "./packages/lib/CommandTypeCodegen.js";
...
-  ]);
-
-
-
-  if (rows.length < 0) {
-    console.error(`POST /dictionary: No command dictionary was updated in the database`);
-    res.status(500).send(`POST /dictionary: No command dictionary was updated in the database`);
-    return;
+app.post("/dictionary", async (req, res) => {
+  try {
```
dyst5422: We can remove the try-catch here as we have a global error handler down at lines 97-100

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
goetzrrGit: It was resovled fixed once I rebased with `develop`

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
goetzrrGit: The path has been updated.

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
dyst5422: Might need to have a build step to copy over the .ts file for that purpose. That can be handled later in the execution PR though.

goetzrrGit: Added a copy command to the `npm run build`

command-expansion-server/sql/commanding/tables/command_dictionary.sql
goetzrrGit: I have changed `pair` to `command_dictionary_synthetic_key`  thanks for the suggestion.

mattdailis: 🔻 Hmm, I'm going to be a bit pedantic here and unpack what I think we mean by `synthetic` and `natural`.

- **natural key**: a key formed by some combination of columns that make up the data in this table
- **synthetic key**: an additional column whose sole purpose is to give each row a unique identifier

So the `pair` key is really a *natural* key, because it is formed by columns that are already in the data: (mission, version)

The key above - the `command_dictionary_primary_key` - is in fact a synthetic key, since it is based on a generated integer whose sole purpose is to give each row a unique identifier.

Take a look at the mission model sql table for an example:
https://github.com/NASA-AMMOS/aerie/blob/bcb2bae8de0357ccc8da2c2aaf76da8e7f49bf0b/merlin-server/sql/merlin/tables/mission_model.sql#L12-L15

So for the sake of consistency with the mission model table, I vote we rename `command_dictionary_primary_key` to `command_dictionary_synthetic_key`, and rename what used to be called `pair` to `command_dictionary_natural_key`.

Let me know if that doesn't make sense or if you see reasons to use different words here.

goetzrrGit: Thank for the explanation @mattdailis This makes sense to me and I will make the change to match that of the mission model.

dyst5422 (approved):

command-expansion-server/src/app.ts
```diff
@@ -9,93 +9,89 @@ import { processDictionary } from "./packages/lib/CommandTypeCodegen.js";
...
     parsedDictionary.header.version,
   ]);
 
-
-
   if (rows.length < 0) {
     console.error(`POST /dictionary: No command dictionary was updated in the database`);
-    res.status(500).send(`POST /dictionary: No command dictionary was updated in the database`);
-    return;
+    res.status(400).json({ message: `POST /dictionary: No command dictionary was updated in the database` });
```
mattdailis, line 50: This commit, "Minor tweaks to Upload endpoint to comply with Hasura action", mixes style refactoring with functional changes, and as a result the meat of the change gets lost in the noise.

Would it be possible to split it into two commits? The first commit would only contain the style changes, and the second one only the functional changes?

goetzrrGit: Hi @mattdailis the style change was a side-affect from the prettier javascript extension that runs every time I save. I agree this adds some noise, but what you highlighted is the change reflected in the commit. I had a lot of trouble rebasing and minimizing the PR into a smaller size? Can I pass on splitting this up into 2 commits but will take this forward to the other PRs?

dyst5422: @goetzrrGit Can you do the rebase to split this apart?

goetzrrGit: let @mattdailis and @dyst5422  let me try that

dyst5422:
@patkenneally theres a security warning for rate limiting to prevent DDOS. I figure since this is an internal service that we would have any rate limiting on the gateway/hasura instead of this service and we can dismiss it

command-expansion-server/src/app.ts
goetzrrGit: @mattdailis and @dyst5422 I got it split now

patkenneally:
@dyst5422 Agreed. We need to have this discussion to ensure we have a clear design for this. But for now, we can let the warning remain.

command-expansion-server/src/app.ts
mattdailis: Thanks for splitting the commits Ryan - I really appreciate it. (And sorry for missing your message from an hour ago)

Looks like this line was lost in the rebase:
```
res.status(400).json({ message: `POST /dictionary: No command dictionary was updated in the database` });
```

You'll probably want to add that back into the commit called "Don't return an 'undefined' if there is an error processing the command dictionary."

goetzrrGit: Sigh....stupid git

Good caught. Thanks @mattdailis

mattdailis: Fantastic, thanks Ryan

mattdailis (approved):

PR #71, opened 2022-02-28
[AERIE-1692] Connect Step Simulation to the scheduler

adrienmaillard:
* **Tickets addressed:** AERIE-1692
* **Review:** By commit  

## Description
In the first commit, the `IncrementalSimulationDriver` is fixed:
- the exit condition in `simulateSchedule` has been changed: it does not only wait for all tasks to be finished but also all the events that might be immediately/instantly triggered by these activities (changing the value of a register for example). This bug has been found after running a test involving the `banananation` mission model. 
- a private method has been added to simulate until a certain time, without new activities. This method is used to generate `simulationResults` up to a certain time. 

Second commit is making the duration getters return `Optional` so it is cleaner from the outside. 

Third commit is adding a `SimulationFacade` member to `SimResource` to prepare for changes in 5th commit. 

Fourth commit adds the possibility of passing a `SimulationFacade` to the `Problem` ctor as there can be a need for a `SimulationFacade` defined outside the `Problem`, when building a problem P2 from another problem P1 for example as goals defined in P1 are referencing resources from the simulation facade contained in P1.

Fifth commit is the actual connection between step simulation and scheduling : 1) replace `simulatePlan` in `SimulationFacade` by incremental operations `simulateActivity` and `removeActivityFromSimulation` 2) resources are updated only when necessary depending on the queried time domain in addition to the current validity of simulation results making the simulation of empty plans at initialization now useless.

Sixth commit is removing the `SimulationDriverFacade` whose role is handled by `SimulationFacade`. 

Seventh commit is removing the last traces of the merlinsight jar from the gradle file after merge with #63.

## Verification
Scheduler tests, including MerlinsightTests (that require the AerieLander mission model), were run successfully.

## Documentation

Check the step simulation engine for accuracy. 

## Future work

- Caching for resources: when querying SimulationResults, it would be great to be able to query them only for a certain time window. Then, querying the simulation results could also become more incremental.    
- Using a dependency graph to avoid triggering simulation only to get back simulation results


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -79,6 +81,27 @@ private void initSimulation(){
     }
   }
 
+  //
+  private void simulateUntil(Duration endTime){
+    assert(endTime.noShorterThan(curTime));
```
pcrosemurgy, line 86: I'd avoid using `assert` since assertion checking can be toggled by the runner of the JVM.
I think it may be best to replace this with: `if (...) throw new Error("...")`.

Also, line 84 seems to have a stray empty comment.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -171,8 +206,8 @@ private void simulateSchedule(Map<ActivityInstanceId, Pair<Duration, SerializedA
    * @param activityInstanceId the activity id
    * @return its duration if the activity has been simulated and has finished simulating, an IllegalArgumentException otherwise
    */
-  public Duration getActivityDuration(ActivityInstanceId activityInstanceId){
-    return engine.getTaskDuration(plannedDirectiveToTask.get(activityInstanceId));
+  public Optional<Duration> getActivityDuration(ActivityInstanceId activityInstanceId){
+    return Optional.of(engine.getTaskDuration(plannedDirectiveToTask.get(activityInstanceId)));
```
pcrosemurgy: >Second commit is making the duration getters return Optional so it is cleaner from the outside.

Looks like an optional is unnecessary here. Maybe I'm missing some context?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
```diff
@@ -18,7 +17,7 @@
    */
   private final MissionModel<?> missionModel;
 
-  private final SimulationFacade simulationFacade;
+  private SimulationFacade simulationFacade;
```
pcrosemurgy: I think the removal of `final` here may have been on accident?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
pcrosemurgy: Ahh, it looks like you've made lower-level functions return `Optional` to make it easier on higher-level functions. I think any `Optional` wrapping should be done client-side (within the higher-level code), unless `engine.getTaskDuration()` returned a `null` or `Optional` itself.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: I'll replace the assert with an exception and remove the comment.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
adrienmaillard: Yes it has been an accident, I'll put it again.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
pcrosemurgy: Cool, thanks Adrien!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: So actually, my code is wrong because the SimulationEngine throws an exception when the duration is not available. I should try/catch it. And what you are saying is that it should be passed up to the simulation facade instead of treated in the driver ?

pcrosemurgy: Ohh, good catch, I missed that.

```java
  public Duration getTaskDuration(TaskId taskId){
    final var state = tasks.get(taskId);
    if (state instanceof ExecutionState.Terminated e) {
      return e.joinOffset().minus(e.startOffset());
    }
    throw new IllegalStateException("Asking for the duration of unfinished task");
  }
```

It may be a good idea to change this exception to a custom checked exception (instead of an unchecked runtime exception like it is now). Then you can choose to either bubble up the exception or do a try/catch within `IncrementalSimulationDriver` to wrap it as an `Optional` if the exception is thrown.

All depends on whether the incremental sim. driver is responsible for gracefully handling the exception case or not. I have a feeling that the incremental sim. driver should bubble the exception up and have clients (callers) worry about how it should be handled (`Optional` wrapping, logging an error, etc.).

mattdailis: Hmm... I'm not arguing either way, but I do think unconditionally wrapping a value in an Optional is suspicious.

I propose having the simulation engine return an optional.

Just so I understand - under what circumstances would we call getActivityDuration and expect it to be empty?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
```diff
@@ -60,7 +59,15 @@ public Problem(MissionModel<?> mission, PlanningHorizon planningHorizon) {
         this.add(new ActivityType(taskType.getKey(), taskType.getValue()));
       }
     }
-    simulationFacade = new SimulationFacade(planningHorizon, missionModel);
+    this.simulationFacade = simulationFacade;
+  }
+
+  public Problem(PlanningHorizon planningHorizon){
+    this(null, planningHorizon, null);
```
mattdailis: Just trying to understand - why do we want to allow defining a Problem with no mission model, and no simulation facade?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
```diff
@@ -80,66 +89,65 @@ public PlanningHorizon getPlanningHorizon(){
...
-      schedule.put(activityIdSim, Pair.of(
-          act.getStartTime(),
-          new SerializedActivity(act.getType().getName(), params)));
     }
+    //reset incremental simulation
+    if(atLeastOne){
+      var oldInsertedActivities = new HashMap<>(insertedActivities);
+      insertedActivities.clear();
+      planActInstanceIdToSimulationActInstanceId.clear();
+      driver = new IncrementalSimulationDriver(missionModel);
```
mattdailis, line 121: Just checking - is creating a new IncrementalSimulationDriver the current expected way to reset the simulation?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
adrienmaillard: For toy problems, for tests.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/SimulationFacade.java
adrienmaillard: The other way was to call `IncrementalSimulation.initSimulation` but it is currently private. Both solutions are very close in terms of what's done.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: @mattdailis this can happen when the simulation of an activity has started but has not finished and you are asking for its duration.

adrienmaillard: Querying the duration of an unfinished task is not a normal operation/a normal interaction to have with the engine, it denotes an algorithmic failure on the client side, that's why I put a runtime exception. What do you think about that ?

pcrosemurgy: So it's not expected that any client (UI or otherwise) will ask for an unfinished task's duration, right? If so then I agree, the `IllegalStateException` runtime ex. sounds okay. But if there's any UI/client-side workflow that regularly queries for an unfinished task's duration then a checked exception would probably be best.

adrienmaillard: It is not expected no. I'll let the exception in the engine and remove the Optional in the client.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
mattdailis: I would like to find a way to make the interface consistent for "real" problems and "test" problems. Not sure off the top of my head how to do that.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: Actually... querying a duration of an unfinished task is kind of the only way of knowing whether a task has finished or not. The other way is computing `SimulationResults` but this is very costly. 
I retract my previous statement and I am going with Matt's proposition, the engine returning an `Optional` directly.  
Sorry for the back and forth.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Problem.java
mattdailis: I made a ticket for this, no need to do this in this PR: AERIE-1748

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
mattdailis: Looks good to me

mattdailis (approved):
I have no remaining concerns for this PR

mattdailis:
The plan is to merge #54 before this PR, so that @adrienmaillard can rebase onto it.

pcrosemurgy:
@mattdailis looks like #54 is marked as "do not merge"

pcrosemurgy (approved):
Thanks Adrien!

mattdailis:
> @mattdailis looks like #54 is marked as "do not merge"

Yes - you're right. I marked it with that because although it has two approvals, I haven't tested a key piece of its functionality yet. In particular, I want to make sure I got the JAR loading logic correct for the new `SchedulerPlugin`. I'm working on testing that right now, and if it works I'll remove the label and merge that PR :+1:

pcrosemurgy:
Ahh got it, good thinking!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -79,6 +81,27 @@ private void initSimulation(){
     }
   }
 
+  //
+  private void simulateUntil(Duration endTime){
+    assert(endTime.noShorterThan(curTime));
+    while (true) {
+      final var batch = engine.extractNextJobs(Duration.MAX_VALUE);
+      // Increment real time, if necessary.
+      final var delta = batch.offsetFromStart().minus(curTime);
```
kroffo: This variable is computed up here, but not used until line 95, the only place it is used. I suggest you either move this to just above line 94, or swap lines 94 and 95 and inline this computation.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
```diff
@@ -114,6 +137,10 @@ public SimulationResults getSimulationResults(){
    */
   public SimulationResults getSimulationResultsUntil(Duration endTime){
     //if previous results cover a bigger period, we return do not regenerate
+    if(endTime.longerThan(curTime)){
+      simulateUntil(endTime);
+    }
```
kroffo: Just a naming suggestion, you don't have to go with this if you don't agree, but I think a more fitting name for this method may be `getSimulationResultsUpTo`.

kroffo (approved):
Some minor comments, but overall looks good. Because the issues I commented are so minor, I've gone ahead and given approval, but please do look at them and consider fixing if you agree with them.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/IncrementalSimulationDriver.java
adrienmaillard: I like it! I'll rename this method before merging.

PR #72, opened 2022-03-01
Add name column to scheduling_goal table

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Scheduling goals should have both a short human-readable name and a longer text description. Neither of these is required to be unique, since the unique identifier for scheduling goals is an integer autogenerated by postgres.

## Verification

I opened hasura and saw that the name field was available via GraphQL.

## Documentation

No documentation exists for scheduling goals yet.

## Future work

No next steps.

### DIFF
--- a/scheduler-server/sql/scheduler/tables/scheduling_goal.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_goal.sql
@@ -1,6 +1,7 @@
 create table scheduling_goal (
   id integer generated always as identity,
   revision integer not null default 0,
+  name text not null,
   definition text not null,
 
   model_id integer not null,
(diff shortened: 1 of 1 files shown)


patkenneally (approved):

camargo (approved):

dyst5422 (commented):
Do we have any cases where duplicate names would be problematic? The first one that jumps to mind is returning errors for bad scheduling goal typescript definitions where we use would probably use the goal name as the defining name in the error messages as its user readable. I think that this particular case are perfectly ok and avoidable by process of the users.

mattdailis:
> Do we have any cases where duplicate names would be problematic? The first one that jumps to mind is returning errors for bad scheduling goal typescript definitions where we use would probably use the goal name as the defining name in the error messages as its user readable. I think that this particular case are perfectly ok and avoidable by process of the users.

We're allowing duplicate names for now - to be super precise we can return both the name and the unique numeric id in the error message

EDIT: But if it turns out to be desirable I'm not opposed to adding a uniqueness constraint.

dyst5422:
> > Do we have any cases where duplicate names would be problematic? The first one that jumps to mind is returning errors for bad scheduling goal typescript definitions where we use would probably use the goal name as the defining name in the error messages as its user readable. I think that this particular case are perfectly ok and avoidable by process of the users.
> 
> We're allowing duplicate names for now - to be super precise we can return both the name and the unique numeric id in the error message
> 
> EDIT: But if it turns out to be desirable I'm not opposed to adding a uniqueness constraint.

Yeah, I legitimately am unsure of whether the uniqueness is necessary and in favor of avoiding it until/if we decide its necessary, just wanted to provoke some thought around it.

PR #73, opened 2022-03-02
docs: remove internal docs and deployment variables 

camargo:
## Description

- Update Environment.md with external deployment vars
- Update docker-compose deployment to have CAM auth OFF by default
- Internal docs are now here: https://github.jpl.nasa.gov/Aerie/aerie-internal-docs
- Add warning to change default passwords to deployment README.md
- Formatting and alphabetization

## Documentation

This PR is primarily documentation updates.

### DIFF
--- a/deployment/Environment.md
+++ b/deployment/Environment.md
@@ -11,25 +11,22 @@ This document provides detailed information about environment variables for each
 
 ## Aerie Gateway
 
-| Name                          | Description                                                            | Type     | Default                                        |
-| ----------------------------- | ---------------------------------------------------------------------- | -------- | ---------------------------------------------- |
-| `AUTH_TYPE`                   | Mode of authentication. Set to `none` to fully disable authentication. | `string` | cam                                            |
-| `AUTH_URL`                    | URL of authentication API for the given `AUTH_TYPE`.                   | `string` | https://atb-ocio-12b.jpl.nasa.gov:8443/cam-api |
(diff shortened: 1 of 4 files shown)


pcrosemurgy (approved):

deployment/SSL.md
```diff
@@ -0,0 +1,23 @@
+# SSL
+
+Some Aerie service configurations require connecting to internal JPL services (e.g. authentication via CAM). For this to work you need to add the proper internal JPL certificates to some of the Aerie containers.
```
patkenneally: While this information is really important to capture and communicate to users, should we be placing this information in the public repo/part of the public docs? I'm pondering our other options. Maybe a separate set of docs specific for JPL users which dovetails with the public docs. 🤔

camargo: Yeah I agree. The deployment docs seemed like the most logical place for now.

camargo: We could put them in a folder `deployment/internal` to separate the docs. I can't think of a simple way to do this though without keeping a completely separate internal document set. 
Also related to this, `cam` auth is the default for our deployment, which implies internal deployment. Maybe we switch the default to `none` and do keep an internal document set?

patkenneally (approved):

deployment/SSL.md
patkenneally: We resolved to move specific deployment concerns to a JPL internal docs repo therefore keeping the main project here generic in application.

PR #74, opened 2022-03-03
[EXAMPLE] Mock command expansion endpoints

dyst5422:
## Description

THIS IS NOT TO BE MERGED -> It is the original PR the other PR's were pulled/split from and is not intended to be merged. 

Don't waste your time with code reviews on this one

## Verification

I ran through the workflow for these, but I need to write tests still

## Documentation

I don't know of documentation invalidated, but I definitely want to put some effort into documenting our endpoints

## Future work

Integration with Gateway/Hasura so we have a nice public API

We probably need to change the simulation_id out with dataset_id on the expansion_run table (and cascade those changes through code) so we are pinned to a specific simulation run and know specifically the state that the commands were generated from


command-expansion-server/src/app.ts
```diff
@@ -1,75 +1,293 @@
...
+          activityTypeName: activityInstance.type,
+          commandTypes,
+          activityTypes: await getActivityTypes(graphqlClient, missionModelId, activityInstance.type),
+          activityInstance,
+        })
+        return {
+          activityInstanceId: activityInstance.id,
+          commands: result.commands
+        };
+      } catch(e: any) {
```
pcrosemurgy, line 239: More of a question for my own TS learning, but is catching `any` exception an anti-pattern? Similar to how `catch (final Exception e)` is in Java. My reasoning comes from [this](https://stackoverflow.com/questions/2416316/why-is-the-catchexception-almost-always-a-bad-idea):
>when you catch all exceptions, you may get an exception that cannot deal with and prevent code that is upper in the stack to handle it properly.

dyst5422: Yes and no - here we are JUST wrapping the error with contextual information of where it came from and re-throwing (so the users knows WHICH activity failed to expand). So catching all is a no no, but catching and re-throwing for the purpose of adding additional context as it jumps up the stack is a good thing.

dyst5422: We can definitely specialize errors in the future for better error handling, but in the first cut, I think we're fine saying "Something went wrong while doing this"

pcrosemurgy: That's fair enough. I try to be as specific with exception catching as possible, even when prototyping since it's easy to forget to circle back to these things. This is re-throwing with the intention of *adding* additional context but it's really just re-throwing and *obfuscating* the original context since any exception is now an `InheritedError`.

Either way, not a big deal, but if this `e: any` can be more specific and takes just a minute it may be worthwhile to just get right now rather than later :)

command-expansion-server/src/getCommandTypes.ts
```diff
@@ -0,0 +1,21 @@
...
+    FROM command_dictionary
+    WHERE id = $1;
+  `, [
+    dictionaryId
+  ]);
+
+
+  const [row] = rows;
+  if (rowCount < 0) {
+    throw new ErrorWithStatusCode(`No dictionary with id: ${dictionaryId}`, 404);
```
pcrosemurgy, line 18: Is it possible for `rowCount` to be `< 0`? If I'm understanding its assignment correctly it's a count on the `SELECT` query, right? If so, then `< 0` should not be possible, only `== 0` would be. Good chance I'm just missing something crucial here.

pcrosemurgy: In `getExpansionSet.ts` `rows.length < 1` is used, which is what I would've expected here I think.

command-expansion-server/src/utils/InheritedError.ts
```diff
@@ -0,0 +1,18 @@
+export type ErrorLike = {
+  message: string,
+  name: string,
+  stack?: string | null,
```
pcrosemurgy, line 4: Silly TS question: is the `| null` really needed here? I'd assume that `stack?: <type>` means that `stack` can be `<type>` or `null` already.

In `assertions.ts` I see: `context?: T`, which is the form I would expect for this kind of type.

command-expansion-server/src/app.ts
dyst5422: Yeah, this one especially is tricky because the worker will throw and the error is serialized + deserialized as just a base Error

command-expansion-server/src/getCommandTypes.ts
dyst5422: 100% correct

dyst5422: Feel free to update. I think I've tried to deprecate out the rowcount pattern in place of rows.length as well

command-expansion-server/src/utils/InheritedError.ts
dyst5422: The question mark says its also possibly undefined. So you can omit it, give it a string, or give it null.

command-expansion-server/src/utils/typeguards.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+/** Typeguard resolved settled promises */
+export function isResolved<T>(promise: PromiseSettledResult<T>): promise is PromiseFulfilledResult<T> {
+  return promise.status === 'fulfilled';
+}
+
+/** Typeguard rejected settled promises */
+export function isRejected<T>(promise: PromiseSettledResult<T>): promise is PromiseRejectedResult {
+  return promise.status === 'rejected';
```
pcrosemurgy, line 18: Is there no `promise.isRejected()`? Does JS/TS use a string for `promise.status` instead of an enum?

dyst5422: This is a javascript implementation thing. So no isRejected(), gotta just check status. There _might_ be an enum for it around somewhere, but I haven't checked if they have updated the standard libs with that or not

command-expansion-server/src/utils/InheritedError.ts
pcrosemurgy: Ohh, thanks! Makes sense

command-expansion-server/src/getCommandTypes.ts
dyst5422: I think this was the result of the rebase with Ryan's branch

command-expansion-server/src/utils/typeguards.ts
pcrosemurgy: Got it. That's odd, I'd expect JS to have used an enum here. Maybe an early version of JS used string and now for backwards compat. we're stuck with it.

command-expansion-server/src/getCommandTypes.ts
pcrosemurgy: Cool, I don't see an update pushed yet but feel free to resolve this convo. when you've changed the condition. :)

command-expansion-server/src/utils/typeguards.ts
dyst5422: Well enums dont ACTUALLY exist in JS, they are an addition in TS that is actually just syntactic sugar for an object
```
enum PromiseState {
  Resolved = "resolved",
  Rejected = "rejected"
}
```

transpiles to

```
const PromiseState = {
  Resolved: 'resolved',
  Rejected: 'rejected'
}
```

Or in a simpler typed TS format equivalent to

```
interface PromiseState {
  Resolved: 'resolved',
  Rejected: 'rejected'
}
const PromiseState: PromiseState = {
  Resolved: 'resolved',
  Rejected: 'rejected'
}
```

command-expansion-server/src/getCommandTypes.ts
dyst5422: Any chance you have time to make that update? I need to focus on M2020 today

command-expansion-server/src/utils/typeguards.ts
pcrosemurgy: Ohh wow. Thanks, this is really good to know.

command-expansion-server/src/getCommandTypes.ts
pcrosemurgy: Ohh I misunderstood, yeah I'll push this.

pcrosemurgy: Just pushed bcc4ea311.

dyst5422:
Closing as this prototype has been fully fleshed out

PR #75, opened 2022-03-03
Adjust expansion_run columns

dyst5422:
* **Tickets addressed:** ARIE-1687
* **Review:** By commit  

## Description

We need to swap out simulation_id with dataset_id in the expansion_run
table so that we are locked to a particular run of the simulation for
traceability. We should still be able to attain all of the simulation
information from the dataset through the GraphQL API.

## Verification

Deployed locally with no issues

## Documentation

We should update the diagram for command expansion workflow to properly reflect the table update @goetzrrGit 

## Future work

Ensure future PRs adhere to the new column

### DIFF
--- a/command-expansion-server/sql/commanding/tables/expansion_run.sql
+++ b/command-expansion-server/sql/commanding/tables/expansion_run.sql
@@ -1,7 +1,7 @@
 create table expansion_run (
   id integer generated always as identity,
 
-  simulation_id integer not null,
+  dataset_id integer not null,
   expansion_set_id integer not null,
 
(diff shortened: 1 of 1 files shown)


goetzrrGit (approved):
LGTM

Yes, that diagram needs to be updated. I can do that.

mattdailis (approved):
Looks good to me. I mentioned this to @kroffo this morning, we should get his thoughts on whether this is the best long-term strategy, but that shouldn't hold up this PR, since he's in his class today :+1:

PR #80, opened 2022-03-04
[AERIE-1725]  get command typescript types

dyst5422:
* **Tickets addressed:** AERIE-1725
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This PR will retrieve the command typescript Library used to author the command expansions.  Here is the workflow:

1. Use a Hasura action and provide the `commandDictionaryID`. 
2. The commanding server will retrieve the command typescript from the filesystem. Note: this was generated when the command dictionary was uploaded 
4. The library is base64 encoded and sent back as the response to the Hasura action

## Verification
I tested this locally

1. `docker system prune --all --volumes` nuke everything 
2. `./gradlew clean; ./gradlew assemble; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up` 
3. Follow the step in this PR to upload the command dictionary. Start at step 3: https://github.com/NASA-AMMOS/aerie/pull/70
5. Run the action below. 

```
query commandTypescript {
  getCommandTypeScript(commandDictionaryId: "1") {
    typescript
  }
}
```
6. Verify that you receive `base64` string back from the action.
7. Take the string and convert it with this URL: https://www.base64decode.org/

## Documentation
I will have to add these steps to the User Guide that we will be giving the users

## Future work
Creating Hasura action for the other endpoints in the commanding server and hooking up the rest of the commanding server.

PR #78, opened 2022-03-03
[AERIE-1724]  stub command expansion endpoints

dyst5422:
* **Tickets addressed:** AERIE-1724
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This simply does a first stub out of the endpoints needed for command expansion. Each endpoint returns a 501 not implemented code with an appropriate message

## Verification

Each of these were test manually.

## Documentation

None

## Future work

Add in tests somewhere

PR #81, opened 2022-03-04
[AERIE-1726]  Activity Typescript Types Generation

dyst5422:
* **Tickets addressed:** AERIE-1726
* **Review:** By commit  

## Description

This PR will generate the activity_type's typescript Library used to author the command expansions.  Here is the workflow:

1. Use a Hasura action and provide the `missionModelID` and `Activity_Type's` name. The action will provide this information to the commanding server
2. The commanding server will retrieve the parameters from the activity_type
3. Once the parameters are queried, a typescript library is generated
4. The library is base64 encoded and sent back as the response to the Hasura action

## Verification
I test locally without any issues using `ParmeterTest` activity_type, which has very complicated parameters objects. 

1. `docker system prune --all --volumes` nuke everything 
2. `./gradlew clean; ./gradlew assemble; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up` 
3. Add a mission model jar
4. Run the action which has the activity_type name and mission model 

```
query MyQuery {
  getActivityTypeScript(activityTypeName: "BakeBananaBread", missionModelId: "1") {
    typescript
  }
}
```
5. Verify that you receive `base64` string back from the action.
6. Take the string and convert it with this URL: https://www.base64decode.org/

## Documentation

## Future work
Creating Hasura action for the other endpoints in the commanding server and hooking up the rest of the commanding server.

### DIFF
--- a/command-expansion-server/package.json
+++ b/command-expansion-server/package.json
@@ -21,6 +21,7 @@
     "body-parser": "^1.19.2",
     "express": "^4.17.2",
     "graphql": "^16.3.0",
+    "graphql-request": "^4.0.0",
     "pg": "^8.7.3",
     "reserved-words": "^0.1.2",
     "typescript": "^4.5.5"
--- a/command-expansion-server/package.json
+++ b/command-expansion-server/package.json
@@ -20,7 +20,6 @@
     "@nasa-jpl/aerie-ampcs": "^1.0.0",
     "body-parser": "^1.19.2",
     "express": "^4.17.2",
-    "graphql": "^16.3.0",
     "graphql-request": "^4.0.0",
     "pg": "^8.7.3",
     "reserved-words": "^0.1.2",
--- a/command-expansion-server/src/getActivityTypescript.ts
+++ b/command-expansion-server/src/getActivityTypescript.ts
@@ -1,0 +1,126 @@
+import { GraphQLClient, gql } from "graphql-request";
+import { globalDeclaration, indent, interfaceDeclaration } from "./packages/lib/CodegenHelpers.js";
+import { ErrorWithStatusCode } from "./utils/ErrorWithStatusCode.js";
+
+enum SchemaTypes {
+  Int = "int",
+  Real = "real",
+  Duration = "duration",
+  Boolean = "boolean",
+  String = "string",
+  Series = "series",
+  Struct = "struct",
+  Variant = "variant",
+}
+
+type Schema =
+  | IntSchema
+  | RealSchema
+  | DurationSchema
+  | BooleanSchema
+  | StringSchema
+  | SeriesSchema
+  | StructSchema
+  | VariantSchema;
+
+interface BaseSchema<T extends SchemaTypes | unknown> {
+  type: T;
+}
+
+type StringSchema = BaseSchema<SchemaTypes.String>;
+type BooleanSchema = BaseSchema<SchemaTypes.Boolean>;
+type IntSchema = BaseSchema<SchemaTypes.Int>;
+type RealSchema = BaseSchema<SchemaTypes.Real>;
+type DurationSchema = BaseSchema<SchemaTypes.Duration>;
+
+interface SeriesSchema extends BaseSchema<SchemaTypes.Series> {
+  items: Schema;
+}
+
+interface StructSchema extends BaseSchema<SchemaTypes.Struct> {
+  items: {
+    [key: string]: Schema;
+  };
+}
+
+interface VariantSchema extends BaseSchema<SchemaTypes.Variant> {
+  variants: {
+    key: string;
+    label: string;
+  }[];
+}
+
+interface GraphQLActivityParameter {
+  order: number;
+  schema: Schema;
+}
+
+interface GraphQLActivity {
+  name: string;
+  parameters: GraphQLActivityParameter[];
+  requiredParameters: string[];
+}
+
+export async function getActivityTypescript(
+  graphqlClient: GraphQLClient,
+  missionModelId: number,
+  activityTypeName: string
+): Promise<string> {
+  console.log(`query parameters from ${activityTypeName} with missionModelId: ${missionModelId}`);
+  const response = await graphqlClient.request<{
+    activity_type: GraphQLActivity[];
+  }>(
+    gql`
+      query GetParameters($missionModelId: Int!, $activityTypeName: String!) {
+        activity_type(where: { model_id: { _eq: $missionModelId }, name: { _eq: $activityTypeName } }) {
+          name
+          parameters
+        }
+      }
+    `,
+    { missionModelId, activityTypeName }
+  );
+
+  if (response.activity_type.length === 0) {
+    throw new ErrorWithStatusCode(`Activity type ${activityTypeName} not found`, 404);
+  }
+
+  const activity = response.activity_type[0];
+
+  return generateTypescriptForGraphQLActivitySchema(activity);
+}
+
+function generateTypescriptForGraphQLActivitySchema(activitySchema: GraphQLActivity): string {
+  const propertyDeclarations = Object.entries(activitySchema.parameters)
+    .map(([parameterName, parameterValue]) => `readonly ${parameterName}: ${convertSchemaType(parameterValue.schema)};`)
+    .join("\n");
+  const activityTypeAlias = `type ActivityType = ${activitySchema.name};`;
+
+  return globalDeclaration(`${interfaceDeclaration(activitySchema.name, propertyDeclarations)}\n${activityTypeAlias}`);
+}
+
+function convertSchemaType(schema: Schema): string {
+  const objectDeclaration = (content: string) => `{\n${indent(content)}\n}`;
+  switch (schema.type) {
+    case SchemaTypes.Int:
+    case SchemaTypes.Real:
+    case SchemaTypes.Duration:
+      return "number";
+    case SchemaTypes.Boolean:
+      return "boolean";
+    case SchemaTypes.String:
+      return "string";
+    case SchemaTypes.Series:
+      return `${convertSchemaType(schema.items)}[]`;
+    case SchemaTypes.Struct:
+      return objectDeclaration(
+        Object.entries(schema.items)
+          .map(([key, value]) => `${key}: ${convertSchemaType(value)};`)
+          .join("\n")
+      );
+    case SchemaTypes.Variant:
+      return `(${schema.variants.map((variant) => `'${variant.label}'`).join(" | ")})`;
+    default:
+      return "any";
+  }
+}
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -32,10 +32,6 @@ type Query {
   ): TypeScriptResponse
 }
 
-type TypeScriptResponse {
-  typescript: String!
-}
-
 type Query {
   resourceTypes(
     missionModelId: ID!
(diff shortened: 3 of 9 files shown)


dyst5422:
@goetzrrGit let me know when you think you have this pretty well put together, and I'm going to go through and rebase to cleanup and ensure it builds on our other PRs for merging later.

goetzrrGit:
@dyst5422 This is pretty much done. You can probably `rebase` and `drop `the commits before https://github.com/NASA-AMMOS/aerie/commit/17125226521bc925bad83305bb7fefffdcd706fc. That should clean up this PR and only reflect the new changes. Once the Upload PR is merged It should be rebased into this PR with minor issues.

I remove the other endpoints that are not part of the PR. Also, I didn't include dev environment stuff as I think that can be part of another localized PR. It will just help so the PR isn't so big

dyst5422:
@goetzrrGit Yeah, the intent with these PR's that I put together was to put together a merging plan ahead of time and still providing the niceties of some of the other cleanup and prep work.

So the merge plan was #78 -> [#80, #81, #82, #82] -> The execution PR which builds on all of them.

So I'm going to rebase this on top of #78 and probably cherry pick out lines and pull commits together to tell a cohesive story without all of the development time alterations (unused package removal for example as we switched the api from a form-data to json)

goetzrrGit:
@dyst5422 Sounds like a good idea. Let me know if you need any help?

command-expansion-server/src/getActivityTypescript.ts
```diff
@@ -0,0 +1,131 @@
...
+    [key: string]: Schema,
+  }
+}
+
+interface VariantSchema extends BaseSchema<SchemaTypes.Variant> {
+  variants: {
+    key: string,
+    label: string,
+  }[]
+}
```
goetzrrGit: You are at Typescript God!!!!

Super minor, can you order this to be top-down declaration? It helps with readability.

```
enum SchemaTypes {
  Int = "int",
  Real = "real",
  Duration = "duration",
  Boolean = "boolean",
  String = "string",
  Series = "series",
  Struct = "struct",
  Variant = "variant",
}

interface BaseSchema<T extends SchemaTypes | unknown> {
  type: T;
}
interface SeriesSchema extends BaseSchema<SchemaTypes.Series> {
  items: Schema;
}

interface StructSchema extends BaseSchema<SchemaTypes.Struct> {
  items: {
    [key: string]: Schema;
  };
}

interface VariantSchema extends BaseSchema<SchemaTypes.Variant> {
  variants: {
    key: string;
    label: string;
  }[];
}

type StringSchema = BaseSchema<SchemaTypes.String>;
type BooleanSchema = BaseSchema<SchemaTypes.Boolean>;
type IntSchema = BaseSchema<SchemaTypes.Int>;
type RealSchema = BaseSchema<SchemaTypes.Real>;
type DurationSchema = BaseSchema<SchemaTypes.Duration>;


type Schema =
  | IntSchema
  | RealSchema
  | DurationSchema
  | BooleanSchema
  | StringSchema
  | SeriesSchema
  | StructSchema
  | VariantSchema;
```

goetzrrGit (approved):
You made this way better.

command-expansion-server/src/getActivityTypescript.ts
dyst5422: I think because TS doesn't care much about WHERE the declarations are, many libraries do it this way so that the root or most important types are higher up (seen sooner when looking through the file) and then the needed children come after them.

I'm good either way though, feel free to let me know which you prefer given the above.

dyst5422: See https://github.com/graphql/graphql-js/blob/main/src/language/ast.ts#L139 as a good example of general typing patterns

command-expansion-server/src/getActivityTypescript.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export async function getActivityTypescript(
+  graphqlClient: GraphQLClient,
+  missionModelId: number,
+  activityTypeName: string
+): Promise<string> {
+  console.log(`query parameters from ${activityTypeName} with missionModelId: ${missionModelId}`);
+  const response = await graphqlClient.request<{
+    activity_type: GraphQLActivity[]
+  }>(gql`
```
pcrosemurgy: Maybe this is a little pedantic but the mixing of GraphQL querying and code-gen. logic within the same "translation unit" doesn't feel quite right. I can understand how decoupling this querying from generation may feel a bit overkill right now but in my experience having disciplined decoupling early on can save trouble later.

deployment/hasura/metadata/actions.graphql
```diff
@@ -25,6 +25,17 @@ type Query {
...
 }
 
+type Query {
+  getActivityTypeScript(
+    missionModelId: ID!
+    activityTypeName: String!
+  ): TypeScriptResponse
+}
+
+type TypeScriptResponse {
```
pcrosemurgy: Another slightly pedantic change: would you mind moving `TypeScriptResponse ` to further down in this file?

Looks like this file has this rough ordering:
- Mutations
- Queries
- Types needed for mutations & queries (`*Response` types)
- Scalars

command-expansion-server/package.json
```diff
@@ -21,6 +21,7 @@
     "body-parser": "^1.19.2",
     "express": "^4.17.2",
     "graphql": "^16.3.0",
+    "graphql-request": "^4.0.0",
```
pcrosemurgy, line 23: Is the `graphql` dependency being used? If not then maybe it can be fully replaced with the `graphql-request` dep.?

command-expansion-server/src/getActivityTypescript.ts
pcrosemurgy: One benefit to decoupling would be testability. Currently, in order to test this code gen. you'd need to have a `GraphQLClient`. Instead, if there was an abstract interface that the code generation could take as input then you could write a series of tests fairly easily.

dyst5422: @pcrosemurgy 

I think this is the use case for having **generateTypescriptForGraphQLActivitySchema** below. The exported `getActivityTypescript` is an abstraction over both the query and the generation. Until we have a use case for performing this query without generating types, I don't feel the additional complication of extracting it is worthwhile (and incurs runtime costs). I think we can definitely export the type generation function for testing though

deployment/hasura/metadata/actions.graphql
dyst5422: I'm not disagreeing with you. I think there is an argument to be made for an alternative "grouping" of each Query/Mutation/Subscription operation with its own relevant types and collapsing shared types elsewhere. Not sure which is more productive, but I know I like to see everything together so its easier to grok

command-expansion-server/package.json
goetzrrGit: yeah, `graphql` is used to build schemas which we are not doing in this part of the server. @dyst5422 where you using this library somewhere else?

deployment/hasura/metadata/actions.graphql
goetzrrGit: To be honest I don't have a preference. The other PR has the rough ordering just to be consistent:

[deployment/hasura/metadata/actions.graphql
](https://github.com/NASA-AMMOS/aerie/pull/80/files#diff-1495af69e68d7d8f4bdaf2b9437fb413026dbee6270f8a7676cc95e102543f50R115)

command-expansion-server/src/getActivityTypescript.ts
pcrosemurgy: That's fair enough, all boils down to personal preference about modularity.

>and incurs runtime costs

I'd imagine a call out to another class/function should be very cheap (<10 ms.?). Either way, all good :)

deployment/hasura/metadata/actions.graphql
pcrosemurgy: No strong preference here besides consistency, so feel free to reorder the rest of the actions if you prefer the grouping order.

command-expansion-server/src/getActivityTypescript.ts
```diff
@@ -0,0 +1,131 @@
...
+    query GetParameters {
+      activity_type(where: {
+        model_id: { _eq: ${missionModelId} },
+        name: { _eq: ${activityTypeName} },
+      }) {
+        name
+        parameters
+      }
+    }
+  `);
```
mattdailis: 🔻 Would it be possible to use GraphQL variables instead of string interpolation here? It would be a very minimal change, and it would prevent GraphQL injection issues via the activityTypeName:
```suggestion
  }>(gql`
    query GetParameters($missionModelId: Int!, $activityTypeName: String!) {
      activity_type(where: {
        model_id: { _eq: $missionModelId },
        name: { _eq: $activityTypeName },
      }) {
        name
        parameters
      }
    }
  `, {missionModelId, activityTypeName});
```

mattdailis: Would it be possible to use GraphQL variables instead of string interpolation here? It would be a very minimal change, and it would prevent GraphQL injection issues via the activityTypeName:
```suggestion
  }>(gql`
    query GetParameters($missionModelId: Int!, $activityTypeName: String!) {
      activity_type(where: {
        model_id: { _eq: $missionModelId },
        name: { _eq: $activityTypeName },
      }) {
        name
        parameters
      }
    }
  `, {missionModelId, activityTypeName});
```

command-expansion-server/src/getActivityTypescript.ts
```diff
@@ -0,0 +1,131 @@
...
+    case SchemaTypes.Series:
+      return `${convertSchemaType(schema.items)}[]`;
+    case SchemaTypes.Struct:
+      return objectDeclaration(
+          Object.entries(schema.items)
+          .map(([key, value]) => `${key}: ${convertSchemaType(value)};`)
+          .join('\n')
+      );
+    case SchemaTypes.Variant:
+      return `${schema.variants.map(variant => `'${variant.label}'`).join(" | ")}`;
```
mattdailis: 🔶 It might be a good idea to wrap the enum in parenthesis - the case I'm thinking of is if there's a list of enums, you don't want to end up with:
```
'A' | 'B' | 'C'[]
```

mattdailis, line 127: 🔶 It might be a good idea to wrap the enum in parenthesis - the case I'm thinking of is if there's a list of enums, you don't want to end up with:
```
"A" | "B" | "C"[]
```

mattdailis (changes requested):

command-expansion-server/src/getActivityTypescript.ts
goetzrrGit: Thanks for the suggestion. I made the change and verified it works as expected

command-expansion-server/src/getActivityTypescript.ts
goetzrrGit: good eyes Matt. I tweaked it to include '()' and tested with the `ParameterTest` activity.

mattdailis (approved):

command-expansion-server/package.json
dyst5422: Removed

deployment/hasura/metadata/actions.graphql
dyst5422: Not worth reordering the whole thing. Moved it to follow the other PR

PR #83, opened 2022-03-04
[AERIE-1738] Put `expansion_set`, auto-update `expansion_set_to_rule` join table

dyst5422:
* **Tickets addressed:** AERIE-1738
* **Review:** By commit  

## Description

At this point in command expansion, the user will define a configuration/set. The set is a statement:

- I have a mission model **M**
- I have a command dictionary **D**
- I have `activity_types` **[A...A+1]** defined by **M**
- I have written `expansion logic` **[E...E+1]** using **D**
---- Each `expansion logic` **E** is mapped to a single `activity_type` **A**

`expansion_set`
With **M** and **D**, use **[E...E+1]** for **[A...A+1]**

### **TESTING**
**Upload Command Dictionary**
1. `docker system prune --all --volumes`
2. ./gradlew clean; ./gradlew assemble; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up
3. Upload bananantion.jar
4. Download the command dictionary: https://jira.jpl.nasa.gov/browse/AERIE-1633
5. Copy the XML content to https://www.base64encode.org/ and encode to base64
6. In Hasura use the graphQL to send the command dictionary to the command service
Take the base64 string and add it to the call at <INSERT>

```
mutation MyMutation {
  uploadDictionary(dictionary: "<INSERT>") {
    id
  }
}
```

**Upload a fake expansion logic file**
7. Run the mutation below 

```
mutation Test {
  addCommandExpansionTypeScript(activityTypeName: "BakeBananaBread", expansionLogic: "aGVsbG8gd29ybGQK") {
    id
  }
}

mutation Test {
  addCommandExpansionTypeScript(activityTypeName: "GrowBanana", expansionLogic: "aGVsbG8gd29ybGQK") {
    id
  }
}
```

**Create an Expansion Set**
8. Run the query below 

```
mutation createExpansionSet {
  createExpansionSet(
    commandDictionaryId: 1
    missionModelId: 1
    expansionIds: [1,2]
  ) {
    id
  }
}
```

## Verification
Test locally and both `expansion_set` and `expansion_set_to_rule` are updated

## Documentation
There will be a User Guide for command expansions

## Future work
Expand plan into command blobs.


command-expansion-server/src/app.ts
```diff
@@ -68,8 +75,38 @@ app.get("/expansion/:expansionId(\\d+)", async (req, res) => {
...
+      INSERT INTO expansion_set (command_dict_id, mission_model_id)
+      VALUES (${commandDictionaryId}, ${missionModelId})
+      RETURNING id
+    )
+    INSERT INTO expansion_set_to_rule (set_id, rule_id) VALUES
+      ${expansionIds
+        .map((expansionId: number) => `((SELECT id FROM expansion_set_id), ${expansionId})`)
+        .join(",\n      ")}
+    RETURNING (SELECT id FROM expansion_set_id);
+  `);
```
pcrosemurgy: I marked the CodeQL failure as a false positive here since I believe `db.query()` performs appropriate sanitization.

command-expansion-server/src/app.ts
```diff
@@ -68,8 +75,38 @@ app.get("/expansion/:expansionId(\\d+)", async (req, res) => {
...
+
+  if (rows.length < 1) {
+    throw new Error(`PUT /expansion-set: No expansion set was inserted in the database`);
+  }
+  const id = rows[0].id;
+  console.log(`PUT /expansion-set: Updated expansion set in the database: id=${id}`);
+  res.status(200).json({ expansionSetID: id });
   return;
 });
```
pcrosemurgy, line 116: Should we ignore the rate limiting CodeQL failure? I'm not sure how to accomplish rate limiting with PG or if we want to?

dyst5422: @patkenneally and I discussed it and decided that if we do want to implement rate limiting, it would be on exposed APIS (hasura) not this internal one

command-expansion-server/src/app.ts
dyst5422: I'm all for someone figuring out how to do this appropriately without the string templating, but I just didn't know how to effectively do it

pcrosemurgy: Well I think what you have is okay since `Pool.query()` handles the sanitization. I could be very mistaken though so take my conclusion with a grain of salt 😂

command-expansion-server/src/app.ts
pcrosemurgy: Thanks, in that case I'll dismiss this CodeQL failure

command-expansion-server/src/app.ts
mattdailis: https://node-postgres.com/features/queries#parameterized-query

patkenneally: Let's use parameterized queries as a general rule of thumb.

dyst5422:
> Let's use parameterized queries as a general rule of thumb.

The issue with this one is that we’ll have to template in at least the $num in order to parameterize it appropriately unless you have a sneaky way to do that too

Twisol:
> The issue with this one is that we’ll have to template in at least the $num in order to parameterize it appropriately unless you have a sneaky way to do that too

Templating in a sequence of `$1`, `$2` etc. isn't particularly problematic from a security perspective, since you have control over all data going into that process (and this is a very well-defined, localized kind of expansion). As long as the user data itself isn't getting interpolated, it's okay.

dyst5422:
> > The issue with this one is that we’ll have to template in at least the $num in order to parameterize it appropriately unless you have a sneaky way to do that too
> 
> Templating in a sequence of `$1`, `$2` etc. isn't particularly problematic from a security perspective, since you have control over all data going into that process (and this is a very well-defined, localized kind of expansion). As long as the user data itself isn't getting interpolated, it's okay.

To be fair - technically these apis are all behind hasura which is ensuring the values we use here are Integers, so sql injection isn't really a risk. But I agree that parameterizing and templating in the $num is the more correct form and we should definitely do it

Twisol:
> To be fair - technically these apis are all behind hasura which is ensuring the values we use here are Integers, so sql injection isn't really a risk.

True! But it feels like a latent bug -- if ever these endpoints get exposed for one reason or another, over the evolution of Aerie, we don't want to suddenly have SQL injection vulnerabilities crop up. Local properties are easier to verify than global ones!

patkenneally:
I dwelt on sharing this same thought for the last hour... Assuming that one uncoupled/independent module is secure based on the presence of another nominally uncoupled module seems incorrect. I think that each Aerie module should be secure at it's boundary, regardless of our deployment configuration.

command-expansion-server/src/app.ts
mattdailis: Not necessarily advocating for this approach, but there does seem to be an option that requires no string interpolation at all (even for the $nums):

https://github.com/brianc/node-postgres/issues/957#issuecomment-295583050

dyst5422: @mattdailis thats exactly the kind of incantation we need and should implement.

dyst5422:
@Twisol @patkenneally

I mostly but don't entirely agree. I would think that Hasura is less of a service than an api layer that necessarily is coupling to our services. I would consider each of the services uncoupled from each other, but the API layer I would expect is something we can rely on. We change the form of our service endpoints to suit it after all.

But if that IS the position we want to take, then we need to do input validation on the apis as we can't rely on graphql for it, so let's add an issue for that.

command-expansion-server/src/app.ts
dyst5422: Took me a bit, but I found an SQL ritual to summon the correct rows - pushed it with 7e292c4ab41476f648dc234c2342fcd6e574155b

Twisol: Nice! That looks *clean*.

dyst5422: Clean and very complicated to understand lol - the parameters for the array_fill and array_length functions are...weird

command-expansion-server/src/schemas/expansion-set.ts
```diff
@@ -0,0 +1,26 @@
...
+    },
+    missionModelId: {
+      type: 'integer',
+    },
+    expansionIds: {
+      type: 'array',
+      items: {
+        type: 'integer',
+      },
+      minItems: 1,
```
mattdailis: I'm curious whether it would ever make sense to make an empty expansion set :thinking:

EDIT: Nevermind, this code has been deleted since I started my review last week 😅 . Disregard.

command-expansion-server/src/app.ts
```diff
@@ -1,6 +1,9 @@
+import util from "util";
+
 import express, { Application, Request, Response, NextFunction } from "express";
 import bodyParser from "body-parser";
 import { GraphQLClient } from "graphql-request";
+import Ajv from "ajv";
```
dyst5422: I think we can remove this and the validation call below + schema file as we have the graphql api doing type validation for us. I know @patkenneally and @Twisol had thoughts on that one though that may differ from my own, so I'll let them weigh in.

dyst5422: Removed this. If we want server-side validation, we can add it in the future more comprehensively to all the endpoints.

goetzrrGit (commented):
Verified your last refactor commit looks good with the removal of the AJV library. I don't think I should approve it as I have been contributing to the PR. I will get some other folks on here to move it along.

command-expansion-server/tsconfig.json
```diff
@@ -98,6 +98,6 @@
     // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
     "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
   },
-  "exclude":["commanding_file_store"]
+  "exclude":["commanding_file_store", "build"]
```
pcrosemurgy, line 101: Just so I understand, was this added to exclude generated TS from being imported in this project?

command-expansion-server/src/app.ts
```diff
@@ -83,8 +82,35 @@ app.post('/put-expansion', async (req, res) => {
...
+    commandDictionaryId,
+    missionModelId,
+    expansionIds,
+  ]);
+
+  if (rows.length < 1) {
+    throw new Error(`PUT /expansion-set: No expansion set was inserted in the database`);
+  }
+  const id = rows[0].id;
+  console.log(`PUT /expansion-set: Updated expansion set in the database: id=${id}`);
```
pcrosemurgy: For correctness I'd change this to read:
```
POST /put-expansion-set
```
Same suggestion for a few lines above in `throw new Error(...)`.

command-expansion-server/package-lock.json
```diff
@@ -19,6 +18,7 @@
         "typescript": "^4.5.5"
       },
       "devDependencies": {
+        "@microsoft/tsdoc": "^0.13.2",
```
pcrosemurgy, line 21: @goetzrrGit / @dyst5422 is `@microsoft/tsdoc` analogous to `javadoc`? If so, are we making use of it / plan on using it?

pcrosemurgy (approved):
Thanks Ryan & Dylan, nice job!

command-expansion-server/tsconfig.json
dyst5422: Yeah, so we don't typecheck the build directory

mattdailis (approved):

command-expansion-server/package-lock.json
dyst5422: It is and the intent is that we will be generating documentation at some point. I put this in here so that we have an analogous way to generate documentation. Not yet run though

PR #84, opened 2022-03-05
Feature/Parameterize versioning for releases and snapshots

JoelCourtney:
* **Tickets addressed:** None
* **Review:** By commit  

## Description
Added two properties to `gradle.properties`: `version.number` and `version.isRelease`. `build.gradle` will use these two properties to set the version:

- `version.number`: standard version number
- `version.isRelease`: defaults to `false`. If `true`, the version will be `<version.number>`. If `false`, the version will be `<version.number>-SNAPSHOT-<git commit hash>`.

Github actions will automatically set the `version.isRelease` property to `true` when a tag is pushed.

## Verification
I have manually tested the properties on the command line, making sure that they can be overridden and the correct version string is produced. **I haven't** tested whether the github workflow correctly detects releases and sets the property. Please suggest a way if you think of one.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
+  IS_RELEASE: ${{ startsWith(github.ref, 'refs/tags/v') }}
 
 jobs:
   publish:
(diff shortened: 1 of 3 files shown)


camargo (approved):
Awesome

patkenneally (approved):

PR #86, opened 2022-03-07
Avoid deserializing spawned activities

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR replaces the `spawn(String, Map<String, SerializedValue>)` interface for spawning tasks with a better-typed method, `spawn(DirectiveTypeId<Input, Output>, Input, Task<Output>)`, that shifts the responsibility for when to serialize and deserialize directives from the model to the engine. **In particular**, spawned activities were previously forced by this interface to be serialized on one side and then deserialized on the other; now, though a serialization is still necessary for output production, no deserialization is necessary.

The key insight is that, where the model currently refers to directive types by `String` names, it ought to refer to them by better-typed handles like the existing `Query<Event, State>` for referencing managed cells. We introduce a new such handle, `DirectiveTypeId<Input, Output>`, so that we can provide an `Input` rather than a `Map<String, SerializedValue>`. Moreover, we separate the roles of "information about a task" from "information specifying a task", for which the serialized information played double duty: the `Input` is solely "information about a task", reported to planners, and we provide a `Task<Output>` that solely specifies the task to be executed.

In order for the model to receive these IDs, we replace the one-shot `getTaskTypes()`, which provides a load of directive types to the driver all at once, with an interaction-based `buildRegistry(DirectiveTypeRegistrar<Model>)` method, which allows the model to register directive types one by one, and store the received IDs locally for future use. The task types are communicated to the driver through the provided registrar, while the return value is an opaque value (capturing the saved IDs) to be passed to `instantiate()` when a model is actually constructed.

@adrienmaillard, commit da4e921cc removes some driver benchmarking code that relied on the old `spawn(String, Map)` method. If you still need this code, I'm happy to work with you to find an alternative solution for your needs. I removed it on the assumption that its point has been well-made -- an incremental driver has quadratic wins over one that must reset on every plan change.

## Verification
The compiler itself should validate the majority of changes, since most changes are structural, rather than logical. I ran `SimulateMapSchedule` (foomissionmodel) and `Main` (banananation) and saw no obvious issues; these ought to catch any potential silliness with the construction of `Directive`s in `SimulationEngine`.

## Documentation
* The generated `ActivityTypes` class has changed significantly; any users depending on its contents will need to go through some extra steps. Specifically, `DirectiveTypeRegistry.extract(ActivityTypes::register).taskSpecTypes` will provide the `Map` that was once a field of `ActivityTypes`.
* The `MissionModelFactory` interface has changed as described above.

## Future work
* Move the remaining serialization off the critical path (the core simulation loop) and into results production. Ideally, this will  remove a source of workload heterogeneity that might put pressure on the JVM. Mission models may already be somewhat heterogeneous, but serialization uses an almost completely disjoint set of classes; taking serialization off the critical path may, for instance, allow more of the code cache to be dedicated to JITed model logic.
* Unify the new `spawn` method with the remaining `spawn(Task<Return>)` method. Every task produces a span with `Input` start attributes and `Return` end attributes; the types of these attributes, taken together, define a signature for that task's resulting span. Lambdas are currently tasks with no (`VoidEnum`) start attributes and no end attributes. And, there is no reason a lambda could not have associated start and end attributes in the future.
  * Some care must be taken so that lambdas' spans can still be suppressed in the output, as they are now. For instance, signatures might carry a "hidden" flag that may be set to suppress spans with that signature by default.
* Remove the `MissionModel` parameter from `SimulationEngine#performJobs`, and hence also from all of the internal methods it passes it to. This `MissionModel` is ultimately threaded through to the `EngineScheduler` (which the active task communicates with to inform the engine of its doings), which only uses it to construct a `Directive` that gets stored to the `taskDirective` component table. We should be able to store the `DirectiveTypeId` and `Input` instead, and let the client of the `taskDirective` table -- the results processing logic -- look up the needed information from the `MissionModel` *it* still receives.
  * This is an important step to separating the results processing logic from the simulation engine, as it will make the engine proper completely independent of the `MissionModel` at hand, only concerned with the behaviors of the tasks it's been asked to orchestrate.
* Generalize the `DirectiveTypeRegistry` to also accept resource definitions. Currently, the set of resources exported by a model might vary with the model configuration provided; but this is over-expressive, and both prevents us from caching schematic information from resources and prevents planners from relying on the presence of resources across configuration changes. To my understanding, existing end-users don't use this functionality and didn't ask for it; it was just an accident of implementation. The `buildRegistry` method exists at the exact right time in the `MissionModelFactory`'s lifecycle to receive registered resources before the configuration has been provided.
  * Granted, it may be better to add a new method and a new registry, but the point stands -- the approach in this PR paves the way.
* This PR introduces *another* stack-`Scoped` contextual variable, this time to make the stored `DirectiveTypeId`s available to the `spawn` methods for each activity type. I consider it a strong smell to have *two* context globals around, and I'd like to find some way to unify them. As a side-benefit, we should no longer need to thread an `Executor` through the `ThreadedTask`, `ReplayingTask`, and their respective `ReactionContext`s.
  * This might be necessary, anyway, to unify lambda spawning with directive spawning.


Twisol:
This PR will conflict in interesting ways with #66. They're independent PRs, but they touch similar regions of code, so care must be taken when rebasing one once the other lands.

Twisol:
FYI, I'm chasing an issue in this PR regarding spawning activities from top-level test methods right now.

Twisol:
I've fixed the issue; this branch should be ready for review again.

The new commit isn't well-integrated into the others (in particular, there are test failures before it), so I'll re-integrate it later on, but I figured I should leave it as it is so folks have the opportunity to judge my solution.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+
+public record DirectiveTypeRegistry<Registry, Model>(
+    Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes,
+    Registry registry)
+{
+  public static <Registry, Model>
+  DirectiveTypeRegistry<Registry, Model> extract(final MissionModelFactory<Registry, Model> factory) {
```
mattdailis: FYI The commit called "Use buildRegistry instead of getTaskSpecTypes" does not compile because there's an extra generic in the `MissionModelFactory` parameter.

EDIT: Correction, it's the commit before that, called "Implement DirectiveTypeRegistrar and accumulate its interactions" that doesn't build

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -452,6 +512,8 @@ public JavaFile generateActivityActions(final MissionModelRecord missionModel) {
   public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
     final var typeName = missionModel.getTypesName();
 
+    // Unfortunately, JavaPoet 1.13.0 doesn't support generating Java 16 record definitions,
```
mattdailis, line 464: Not sure if this would be worth our time, but we could make a wrapper method that would generate record syntax for us (and consider upstreaming it to JavaPoet 😉 )

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -461,49 +523,120 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
-                    .builder(
+            .addSuperinterface(
+                ParameterizedTypeName.get(
+                    ClassName.get(Scoping.class),
+                    missionModel.getTypesName(),
+                    ClassName.get(missionModel.topLevelModel)))
+            .addFields(
+                missionModel.activityTypes
+                    .stream()
+                    .map(activityType -> FieldSpec
```
mattdailis, line 483: I see now why you were, in all seriousness, looking up the maximum number of fields allowed in a Java class 😆

examples/config-with-defaults/src/test/java/gov/nasa/jpl/aerie/configwithdefaults/ConfigurationTest.java
```diff
@@ -18,18 +18,18 @@
...
   private final Mission model;
 
-  public ConfigurationTest(final MerlinTestContext<Mission> ctx)
+  public ConfigurationTest(final MerlinTestContext<ActivityTypes, Mission> ctx)
   throws ConfigurationType.UnconstructableConfigurationException
   {
     // Rely on config. defaults by instantiating config. with empty argument map
     final var config = new ConfigurationMapper().instantiate(Map.of());
 
     this.model = new Mission(ctx.registrar(), config);
```
mattdailis, line 31: We now have two separate registrars - one for directives, and one for resources/topics. I wonder if it would make sense to either rename the old registrar to make it clearly a separate concept, or toy with the idea of merging the two.

INSTA-EDIT: Thinking about it for 2 seconds I'm realizing that they're called as part of two different workflows - directives are registered by the framework, automatically, while resources and topics are registered by the users. Not sure which argument that supports though...

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+  public static final class DirectiveTypeId<Input, Output>
+      implements gov.nasa.jpl.aerie.merlin.protocol.driver.DirectiveTypeId<Input, Output>
+  {
+    public final String name;
+
+    private DirectiveTypeId(final String name) {
+      this.name = Objects.requireNonNull(name);
+    }
+  }
+}
```
mattdailis, line 67: My understanding is that we’re using a class instead of a record here because we want to inherit `hashCode` and `equals` from `Object` rather than have them generated by the Java record syntactic sugar. Is that an accurate read on this?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+    private final Map<String, TaskSpecType<Model, ?, ?>> taskSpecTypes = new HashMap<>();
+
+    @Override
+    public <Input, Output> DirectiveTypeId<Input, Output> registerDirectiveType(
+        final String name,
+        final TaskSpecType<Model, Input, Output> taskSpecType)
+    {
+      // TODO: Throw an exception if this name already has already been registered.
+      this.taskSpecTypes.put(name, taskSpecType);
+      return new DirectiveTypeId<>(name);
```
mattdailis, line 54: What you had here initially - `return new DirectiveTypeId<>() {};` blew my mind - an anonymous subclass of an interface with no methods, used solely for its object equality 🤯

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+    final var registry = scope.apply(builder);
+    return new DirectiveTypeRegistry<>(builder.taskSpecTypes, registry);
+  }
+
+  public <Input, Output>
+  TaskSpecType<Model, Input, Output> lookup(final DirectiveTypeId<Input, Output> id) {
+    final TaskSpecType<Model, ?, ?> rawDirectiveType = this.taskSpecTypes.get(id.name);
+
+    @SuppressWarnings("unchecked")
+    final var directiveType = (TaskSpecType<Model, Input, Output>) rawDirectiveType;
```
mattdailis, line 39: Figure it's worth a `// SAFETY:` comment here, or is it obvious from context?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -43,10 +37,16 @@ SimulationResults simulate(
       }
 
       // Schedule the control task.
```
mattdailis: Probably want to update this comment. I noticed you updated the comment in `simulateTask`, but I think that comment would make more sense here instead.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
```diff
@@ -1,15 +1,13 @@
 package gov.nasa.jpl.aerie.merlin.protocol.driver;
 
 import gov.nasa.jpl.aerie.merlin.protocol.model.Task;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
-
-import java.util.Map;
 
 public interface Scheduler {
```
mattdailis, line 5: General thought - I wonder if `Scheduler` is the best word for the interface between the model and the simulation engine. It is certainly responsible for requesting that child tasks be scheduled, but `get` and `emit` don't seem to have to do with scheduling, at least not from the caller's perspective. (Full disclosure, I am motivated to think about this because I spend time thinking about the other thing called a "Scheduler" in the aerie system)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -43,10 +37,16 @@ SimulationResults simulate(
...
-      {
-        final var control = engine.initiateTask(elapsedTime, controlTask);
-        engine.scheduleTask(control, elapsedTime);
+      final var taskToPlannedDirective = new HashMap<String, ActivityInstanceId>();
+      for (final var entry : schedule.entrySet()) {
+        final var directiveId = entry.getKey();
+        final var startOffset = entry.getValue().getLeft();
+        final var directive = entry.getValue().getRight();
+
+        final var taskId = engine.initiateTaskFromInput(missionModel, directive);
```
mattdailis, line 46: I'm curious about the distinction between `initiate` and `instantiate` (and perhaps `initialize`). Care to elaborate? In particular, `initiateTaskFromInput` calls `instantiateDirective`, which makes me wonder if those might mean the same thing.

My own musing: I suppose "initiating" a Task is the domain operation, while "instantiating" a directive quite literally means constructing the Java object representing this directive... but then following that train of thought, I'm curious what the difference between "Start" and "Initiate" would be for a task (since `initiateTaskFromInput` puts a task in the `NotStarted` state - so it has been "initiated", but not "started" 🤔 )

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/RootModel.java
```diff
@@ -6,7 +6,7 @@
 /**
  * The top-level mission model managed by the Merlin framework.
  */
-public record RootModel<Model>(Model model, ExecutorService executor) implements AutoCloseable {
+public record RootModel<Registry, Model>(Model model, Registry registry, ExecutorService executor) implements AutoCloseable {
   public RootModel {}
```
mattdailis, line 10: This was here before this PR, but now I'm curious what the purpose of an empty compact constructor is. Should we delete it?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Scoping.java
```diff
@@ -0,0 +1,21 @@
...
+ * </p>
+ */
+public interface Scoping<Registry, Model> {
+  Undo contextualizeModel(final RootModel<Registry, Model> model);
+
+  // Undos must not throw exceptions, and should be usable in try-with-resources without an Exception catch-all block.
+  interface Undo extends AutoCloseable {
+    void close();
+  }
+}
```
mattdailis, line 21: Hmm... could you help me understand why we want to use Scoped (and Scoping) for the `model` field of `ActivityTypes`? In general, my understanding of the `Scoped` pattern is that we're setting a global variable to a certain value for the duration of a method call. That way the whole call tree below that sees that value of the global variable. It's structured as a linked list of Undo operations, since it's possible for children to set a new value for their children.

(Writing that out helped it click a lot better for me, wow)

What I still would like clarity on is why we want it for mission models - is it valid for tasks to spawn other tasks based on a different mission model...? (Not even sure I'm asking a well-formed question to begin with)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
Twisol: I figured it wouldn't be worth our time, but I *did* go hunting to see if there were any issues open on the subject. [There is one](https://github.com/square/javapoet/issues/829), and some of the branches referencing that issue seem to have taken a shot at this themselves. Nothing's been upstreamed yet though, and I didn't bother testing the capability in any of the forks.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
Twisol: Ah, you're right! (But one commit back, as discussed privately.)

Looks like the next commit compiles because the fix for this got added one commit too late. I've relocated that change to the correct commit.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
Twisol: Right??

examples/config-with-defaults/src/test/java/gov/nasa/jpl/aerie/configwithdefaults/ConfigurationTest.java
Twisol: > toy with the idea of merging the two

As you say, they're used at two separate, and independently important, parts of the model instantiation protocol, so I don't think they can be merged. There's probably some renaming that could be done. As noted in the future work section, resources should *also* be registered via the new (or a similarly-timed) registrar; that would make the old registrar not register much of anything, and hence be in dire need of a new name. But until then, some form of registration is happening at two separate times. 🤷‍♂️

Twisol: Also, changing when resources are registered will be a real breaking change for users who depend on being able to register things with reference to a live, constructed model. It would be much more than a refactor!

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
Twisol: Actually, just because we only want the `DirectiveTypeRegistry` to be able to construct these Ids. If you can conjure an Id up out of nowhere, then you have no guarantee that it will correspond to a directive type in a registry -- nor that the Input and Output types you construct it with will line up with a directive type if one by the given name *does* exist.

Twisol: It's not a foolproof prevention mechanism -- you could just create a registry, register stuff to it to get some IDs, then *query those IDs at a different registry*, but it sets a significantly higher bar.

Twisol: Maybe it would be best to make these IDs non-static, so you can query the directive through the ID itself. That might be safer. Hmmmm.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
Twisol: Haha. I have done that, before! In this case, I was just mocking out what needed to be added.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/DirectiveTypeRegistry.java
Twisol: Yes, good point. I'll even add a `PRECONDITION:` note to the method itself, to call out that the method must only be invoked with IDs acquired from the same registry. (To maybe be obviated by the discussion in an earlier review comment.)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
Twisol: Good catch. Done!

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
Twisol: I originally named this in reference to [process schedulers in operating systems](https://en.wikipedia.org/wiki/Scheduling_(computing)#Process_scheduler), but I totally see the collision. It's more of a set of exposed syscalls than the scheduler itself.

I'm certainly open to hearing alternative names 🤔

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
Twisol: Yes, you're poking at a wart that's been bothering me for some time. If you can wait for my next PR, your questions here should be either answered or obviated 🙏 I'm trying to extract the responsibility for instantiating tasks from the simulation engine altogether, so that the Engine can be much more tightly focused on the *scheduling and execution* of tasks. Where a task comes from shouldn't matter to it.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/RootModel.java
Twisol: 🤔 Based on my understanding, we should be able to remove it. I'd rather leave it in for now, simply because RootModel itself might become a generated class as I try to unify the two `Scoped` variables in a mission model. We'll see...

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Scoping.java
Twisol: Different threads have different thread-locals (which is how Scoped is implemented). We have to make sure we run each activity with its model in-Scope, or else you'll get ugly exceptions when attempting to spawn child activities.

More generally, I don't want to leave any behind visible side-effects; it's too easy to accidentally rely on the Scoped variable still being set when it shouldn't be. We just want a dynamically-scoped variable, and this is how we maintain the stack discipline for that scoping.

Twisol: The most important reason we're using a `Scoped` cell to store the root model is because, if you're in some method during simulation and you `spawn` some other activity, you need the ID of its activity type that you got when you registered it during model initialization. Those IDs are kept in the RootModel, and we don't want to thread them down through every method call from the activity top-level down to everywhere that needs them -- that's a terrible breach of modularity.

Twisol:
@mattdailis, I think I've addressed all of your comments? 😃

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
mattdailis: I do like the operating system analogy... No great names come to mind, I'll keep this in the back of my mind though. Interestingly enough none of the four methods on `Scheduler` actually cause the task to yield - that's done by returning a status from `step`, so the cooperative multitasking is done by a different mechanism 🤔 On my current train of thought, the Scheduler seems to be less about multitasking and more about querying and perturbing the state of the simulation...

This is a discussion to be had separately from this PR, I just wanted to note down the ideas that passed through my head as I read through these commits 😃

mattdailis (approved):
Looks great to me, I'm excited at the possibilities this starts to open up

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Scheduler.java
Twisol: I definitely agree; the analogy isn't perfect, if it ever was. I think it's because `spawn` is part of this interface? 🤔

But yes, it's a good discussion to have 😄

Twisol:
> This PR will conflict in interesting ways with #66. They're independent PRs, but they touch similar regions of code, so care must be taken when rebasing one once the other lands.

With #66 merged, I've rebased this branch and taken care of the conflicts noted above.

Twisol:
Bump! Anything I can do to help move things along? 😄

patkenneally (approved):
I've recorded the changes this requires of mission model developers/tests and will include them in the upcoming release's notes. I have an update for Aerie Lander ready to accommodate this PR.

Twisol:
I've rebased and pushed the branch, checked that all commits build, and -- for good measure -- double-checked that the output of `SimulateMapSchedule` is consistent with `develop`.

Twisol:
> I have an update for Aerie Lander ready to accommodate this PR.

Would you prefer to coordinate merging these updates, or should I go ahead and merge this now?

Twisol:
Rebased on top of latest `develop`; congrats on releasing 0.11 😄

PR #89, opened 2022-03-07
[REFACTOR] Open debug ports on java servers in dev docker compose

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This PR sets the `JAVA_OPTS` environment variables for both the merlin server and the scheduler server.

Both servers will be listening for incoming debugger connections. `aerie_merlin` is listening at port `5005`, and `aerie_scheduler` is listening at port `5006`.

To connect IntelliJ to these ports, go to "Edit Configurations", hit the plus button at the top left, and select "Remote JVM Debug". I recommend adding one configuration for merlin (set the port to 5005) and a separate configuration for the scheduler (set the port to 5006). You can actually run both at the same time, as I just discovered!

## Verification

I followed the above steps and connected to both the scheduler and merlin. I placed a breakpoint in the simulation code before running a simulation to make sure the `aerie_merlin` connection worked. I placed a breakpoint in the scheduler-server code and ran scheduling to make sure that the `aerie_scheduler` connection worked.

## Documentation

Some dev documentation is probably in order here.

## Future work

None off the top of my head.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,8 +57,9 @@ services:
       MERLIN_LOCAL_STORE: /usr/src/app/merlin_file_store
       MERLIN_LOGGING: "true"
       MERLIN_PORT: 27183
+      JAVA_OPTS: "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
     image: aerie_merlin
-    ports: ["27183:27183"]
+    ports: ["27183:27183", "5005:5005"]
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -81,8 +82,9 @@ services:
       SCHEDULER_OUTPUT_MODE: UpdateInputPlanWithNewActivities
       SCHEDULER_PORT: 27193
       SCHEDULER_RULES_JAR: /usr/src/app/merlin_file_store/scheduler_rules.jar
+      JAVA_OPTS: "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005"
     image: aerie_scheduler
-    ports: ["27193:27193"]
+    ports: ["27193:27193", "5006:5005"]
```
pcrosemurgy, line 87: Is this supposed to be `5006:5005`? I would've expected `5006:5006` I think.

mattdailis: Yeah - it just needs to match the port in the JAVA_OPTS line above it. I can set the JAVA_OPTS to use 5006 and update this line accordingly - that would probably be less confusing.

patkenneally (approved):
Thanks for adding this!

docker-compose.yml
pcrosemurgy: Ohh! Got it, thanks. No worries this is all good :)

pcrosemurgy (approved):

PR #82, opened 2022-03-04
[AERIE-1727]  get put expansion

dyst5422:
* **Tickets addressed:** AERIE-1727
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)

PR #92, opened 2022-03-08
[AERIE-1727] Add `put-expansion` endpoint

pcrosemurgy:
* **Tickets addressed:** AERIE-1727
* **Review:** By commit  

## Description
Adds `put-expansion` endpoint and Hasura action.
Initially a `get-expansion` endpoint & action was implemented but it was decided that this is unneeded right now and the UI/clients can simply look inside the `expansion_rules.expansion_logic` column and deserialize the text as base64.

From Dylan's https://github.com/NASA-AMMOS/aerie/pull/82.

## Verification

1. Start with a clean slate: `docker system prune --all --volumes`
1. Start the Docker stack: `./gradlew clean; ./gradlew assemble; docker compose -f docker-compose.yml build; docker compose -f docker-compose.yml up`
1. Run `echo 'hello world' | base64 | pbcopy` to encode a message in base 64 and copy to your clipboard.
1. Run
```graphql
mutation Test {
  addCommandExpansionTypeScript(activityTypeName: "Test", expansionLogic: "aGVsbG8gd29ybGQK") {
    expansionId
  }
}
```
→
```json
{
  "data": {
    "addCommandExpansionTypeScript": {
      "expansionId": 34
    }
  }
}
```
To retrieve the stored expansion logic inspect the `expansion_rules` table and find `"aGVsbG8gd29ybGQK"` under the `expansion_logic` column for id `34`.

## Documentation

## Future work

Implement the `expand` endpoint (AERIE-1741 & AERIE-1729) to actually evaluate the expansion logic.

### DIFF
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -8,11 +8,12 @@ import * as ampcs from "@nasa-jpl/aerie-ampcs";
 import { processDictionary } from "./packages/lib/CommandTypeCodegen.js";
 import { getActivityTypescript } from "./getActivityTypescript.js";
 import { getCommandTypescriptTypes } from "./getCommandTypescriptTypes.js";
+import { ErrorWithStatusCode } from "./utils/ErrorWithStatusCode.js";
 
 const PORT: number = parseInt(getEnv().PORT, 10) ?? 3000;
 
 const app: Application = express();
-app.use(bodyParser.json({ limit: "25mb" }));
+app.use(bodyParser.json({ limit: '25mb' }));
 
 DbExpansion.init();
 const db = DbExpansion.getDb();
@@ -36,12 +37,12 @@ app.post("/put-dictionary", async (req, res) => {
   console.log(`command-lib generated - path: ${commandDictionaryPath}`);
 
   const sqlExpression = `
-      insert into command_dictionary (command_types, mission, version)
-      values ($1, $2, $3)
-      on conflict (mission, version) do update
-      set command_types = $1
-      returning id;
-    `;
+    insert into command_dictionary (command_types, mission, version)
+    values ($1, $2, $3)
+    on conflict (mission, version) do update
+    set command_types = $1
+    returning id;
+  `;
 
   const { rows } = await db.query(sqlExpression, [
     commandDictionaryPath,
@@ -49,28 +50,41 @@ app.post("/put-dictionary", async (req, res) => {
     parsedDictionary.header.version,
   ]);
 
-  if (rows.length < 0) {
+  if (rows.length < 1) {
     console.error(`POST /dictionary: No command dictionary was updated in the database`);
-    res.status(400).json({ message: `POST /dictionary: No command dictionary was updated in the database` });
-  } else {
-    const id = rows[0].id;
-    res.status(200).json({ id });
+    res.status(500).send(`POST /dictionary: No command dictionary was updated in the database`);
+    return;
   }
+  const id = rows[0].id;
+  res.status(200).json({ id });
   return;
 });
 
-app.put("/expansion/:activityTypeName", async (req, res) => {
-  res.status(501).send("PUT /expansion: Not implemented");
+app.post('/put-expansion', async (req, res) => {
+  const activityTypeName = req.body.input.activityTypeName as string;
+  const expansionLogicBase64 = req.body.input.expansionLogic as string;
+
+  const { rows } = await db.query(`
+    INSERT INTO expansion_rules (activity_type, expansion_logic)
+    VALUES ($1, $2)
+    RETURNING id;
+  `, [
+    activityTypeName,
+    expansionLogicBase64,
+  ]);
+
+  if (rows.length < 1) {
+    throw new Error(`POST /put-expansion: No expansion was updated in the database`);
+  }
+
+  const id = rows[0].id;
+  console.log(`POST /put-expansion: Updated expansion in the database: id=${id}`);
+  res.status(200).json({ id });
   return;
 });
 
-app.get("/expansion/:expansionId(\\d+)", async (req, res) => {
-  res.status(501).send("GET /expansion: Not implemented");
-  return;
-});
-
-app.put("/expansion-set", async (req, res) => {
-  res.status(501).send("PUT /expansion-set: Not implemented");
+app.put('/expansion-set', async (req, res) => {
+  res.status(501).send('PUT /expansion-set: Not implemented');
   return;
 });
 
@@ -82,7 +96,6 @@ app.post("/get-command-typescript", async (req, res) => {
   res.status(200).json({
     typescript: commandTypescriptBase64,
   });
-
   return;
 });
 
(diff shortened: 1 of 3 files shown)


command-expansion-server/src/app.ts
```diff
@@ -35,46 +36,87 @@ app.post("/put-dictionary", async (req, res) => {
...
+    throw new ErrorWithStatusCode(`POST /get-expansion: No expansion with id=${expansionId}`, 404);
+  }
+
+  const expansionLogicTypescriptBase64 = Buffer.from(rows[0].expansion_logic).toString('base64');
+  res.status(200).json({
+    typescript: expansionLogicTypescriptBase64,
+  });
+  return;
+});
+
```
pcrosemurgy, line 84: Should we ignore the rate limiting CodeQL failure? I'm not sure how to accomplish rate limiting with PG or if we want to?

command-expansion-server/src/app.ts
```diff
@@ -36,41 +37,77 @@ app.post("/put-dictionary", async (req, res) => {
...
+  const { rows } = await db.query(`
+    SELECT expansion_logic
+    FROM expansion_rules
+    WHERE id = $1;
+  `, [
+    expansionId,
+  ]);
+
+  if (!rows.length) {
```
dyst5422: I try not to use this pattern of `!<anything other than boolean>`.  It's difficult to grok and prone to all kind of bad errors from type coercion. 

I prefer the `if (rows.length < 1)` we've used elsewhere, but there are plenty of other incantations that are fine.

command-expansion-server/src/app.ts
```diff
@@ -36,41 +37,77 @@ app.post("/put-dictionary", async (req, res) => {
...
+  const { rows } = await db.query(`
+    INSERT INTO expansion_rules (activity_type, expansion_logic)
+    VALUES ($1, $2)
+    RETURNING id;
+  `, [
+    activityTypeName,
+    expansionLogicBase64,
+  ]);
+
+  if (!rows.length) {
```
dyst5422: Same comment as above.

dyst5422 (changes requested):

command-expansion-server/src/app.ts
pcrosemurgy: For sure, just pushed this change!

command-expansion-server/src/app.ts
pcrosemurgy: Changed this as well :)

command-expansion-server/src/app.ts
```diff
@@ -36,41 +37,77 @@ app.post("/put-dictionary", async (req, res) => {
...
+    expansionLogicBase64,
+  ]);
+
+  if (rows.length < 1) {
+    throw new Error(`POST /put-expansion: No expansion was updated in the database`);
+  }
+
+  const expansionId = rows[0].id;
+  console.log(`POST /put-expansion: Updated expansion in the database: id=${expansionId}`);
+  res.status(200).json({ expansionId });
```
dyst5422: Do we want the return to be `expansionId` or just `id`?

No strong preference, but I figured I'd pose the question

dyst5422 (approved):

command-expansion-server/src/app.ts
pcrosemurgy: I opted for `expansionId` just to be super clear, can be swayed to `id` though.

It helped me a bit to see the same name (when greping/highlighting) as this block:
```gql
type AddCommandExpansionResponse {
  expansionId: Int!
}
```

dyst5422: I'm inclined to say let's go with id for three reasons. 

1. We have precedent for it here https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/hasura/metadata/actions.graphql#L110-L112
2. The context of WHAT kind of id is being returned is implicated by the type its being returned in (AddCommandExpansionResponse)
3. If we are going to allow for Hasura to grab more information about them from the endpoint (ie return not just the id but the expansion as a complete type { id, typescriptCodeBase64 }) then id is more appropriate (many programs key off of typename + id as a caching key). Its a very common graphql pattern to return the full object on mutations to allow for an insert & pulling other relevant associated information in a single request ("traversing the graph")

pcrosemurgy: For sure, just pushed this change!

command-expansion-server/src/app.ts
```diff
@@ -36,41 +37,77 @@ app.post("/put-dictionary", async (req, res) => {
...
+  ]);
+
+  if (rows.length < 1) {
+    throw new ErrorWithStatusCode(`POST /get-expansion: No expansion with id=${id}`, 404);
+  }
+
+  const expansionLogicTypescriptBase64 = rows[0].expansion_logic;
+  res.status(200).json({
+    typescript: expansionLogicTypescriptBase64,
+  });
```
goetzrrGit: Stupid question, but can you just use a Hasura query to get the expansionLogic base64 string? No action needed

```
query MyQuery3 {
  expansion_rules(where: {id: {_eq: 1}}) {
    expansion_logic
  }
}
```

or 

```
query MyQuery3 {
  expansion_rules(where: {activity_type: {_eq: "BakeBanana"}}) {
    expansion_logic
  }
}
```

Is this endpoint needed?

pcrosemurgy: Nah that's a good point.

I guess I was originally thinking that we wanted this `get-expansion` endpoint to serve as a wrapper in case any deserialization escaping logic changed for the `expansion_logic` TypeScript string. But if we're going forward with base64 then I think we can remove this endpoint and just expect the UI/clients to do their own deserialization.

dyst5422: Yeah, I started thinking about that as well when we went down the base64 path.

I think we can remove the get endpoint here.

command-expansion-server/src/app.ts
```diff
@@ -8,11 +8,12 @@ import * as ampcs from "@nasa-jpl/aerie-ampcs";
 import { processDictionary } from "./packages/lib/CommandTypeCodegen.js";
 import { getActivityTypescript } from "./getActivityTypescript.js";
 import { getCommandTypescriptTypes } from "./getCommandTypescriptTypes.js";
+import { ErrorWithStatusCode } from "./utils/ErrorWithStatusCode.js";
 
 const PORT: number = parseInt(getEnv().PORT, 10) ?? 3000;
 
 const app: Application = express();
-app.use(bodyParser.json({ limit: "25mb" }));
+app.use(bodyParser.json({ limit: '25mb' }));
```
mattdailis, line 16: Some spurious quote changes from double to single quotes seem to be cluttering the diff for this PR. Perhaps you guys have different settings on your autoformatters?

pcrosemurgy: In this case, this change was carried over from a commit by @dyst5422 / @jdylanstewart. I think we're all onboard with using single quotes when possible for string literals that aren't expected to have interpolation, so I'm not expecting much code churn between single <-> double quotes.

goetzrrGit: I am using Prettier in vscode which @camargo recommended to me for Aerie-UI. It adds the double quote and formats upon saving 

I followed this document to set it up. 
https://www.digitalocean.com/community/tutorials/how-to-format-code-with-prettier-in-visual-studio-code

Might be worth keeping checking out

dyst5422: I was planning on adding in a .prettierrc file to configure prettier for single quotes etc, just haven't wanted to clutter up our useful PRs with it.

command-expansion-server/src/app.ts
pcrosemurgy: Done! Just pushed this

kroffo (commented):
Just going to comment instead of approve since this is my first time really looking at typescript, but looks good to me.

mattdailis (approved):

PR #93, opened 2022-03-09
[AERIE-1733] Set merlin log level and output file in docker-compose

JoelCourtney:
* **Tickets addressed:** AERIE-1733
* **Review:** By commit  

## Description

Merlin server uses Javalin, which is configured to use slf4j for logging. I've added the ability to configure the logging level (default INFO) and output file (default is stderr, but can be set to a file).

Level options (decreasing verbosity):
- TRACE (javalin doesn't output trace logs, so this is meaningless)
- DEBUG
- INFO (dev default)
- WARN (deploy default)
- ERROR
- OFF

Javalin instantiates its logger statically in life-before-main, so the log level and file (which are java system properties) cannot be set by merlin at runtime. `JAVA_OPTS` is the only env variable that can set system properties, and I can't interpolate another variable into it. So the configuration has to go directly in `JAVA_OPTS`, which is terrible, but it works.

Configuration is in the `AERIE_MERLIN: environment: JAVA_OPTS` variable in `docker-compose.yml`. I don't know where I should document that.

## Verification
I've been able to set the log level and verify that merlin's log output matches. When I set the output file, the stderr output disappears, but I don't know much about docker and have no idea where to find that file. As far as I know that output vanished into the aether.

## Documentation
I don't know where to document this, please advise.

## Future work
I'll have to do the same thing to the scheduler server.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
@@ -5,7 +5,6 @@ import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import gov.nasa.jpl.aerie.merlin.server.config.AppConfiguration;
 import gov.nasa.jpl.aerie.merlin.server.config.InMemoryStore;
-import gov.nasa.jpl.aerie.merlin.server.config.JavalinLoggingState;
 import gov.nasa.jpl.aerie.merlin.server.config.PostgresStore;
 import gov.nasa.jpl.aerie.merlin.server.config.Store;
 import gov.nasa.jpl.aerie.merlin.server.http.LocalAppExceptionBindings;
@@ -29,11 +28,11 @@ import gov.nasa.jpl.aerie.merlin.server.services.SynchronousSimulationAgent;
 import gov.nasa.jpl.aerie.merlin.server.services.ThreadedSimulationAgent;
 import gov.nasa.jpl.aerie.merlin.server.services.UnexpectedSubtypeError;
 import io.javalin.Javalin;
+import org.slf4j.LoggerFactory;
+
 import java.nio.file.Path;
-import java.util.logging.Logger;
 
 public final class AerieAppDriver {
-  private static final Logger log = Logger.getLogger(AerieAppDriver.class.getName());
 
   public static void main(final String[] args) {
     // Fetch application configuration properties.
@@ -53,7 +52,7 @@ public final class AerieAppDriver {
     // Configure an HTTP server.
     final var javalin = Javalin.create(config -> {
       config.showJavalinBanner = false;
-      if (configuration.javalinLogging().isEnabled()) config.enableDevLogging();
+      if (configuration.enableJavalinDevLogging()) config.enableDevLogging();
       config.enableCorsForAllOrigins();
       config.registerPlugin(merlinBindings);
       config.registerPlugin(new LocalAppExceptionBindings());
@@ -105,9 +104,10 @@ public final class AerieAppDriver {
   }
 
   private static AppConfiguration loadConfiguration() {
+    final var logger = LoggerFactory.getLogger(AerieAppDriver.class);
     return new AppConfiguration(
         Integer.parseInt(getEnv("MERLIN_PORT","27183")),
-        Boolean.parseBoolean(getEnv("MERLIN_LOGGING","true")) ? JavalinLoggingState.Enabled : JavalinLoggingState.Disabled,
+        logger.isDebugEnabled(),
         Path.of(getEnv("MERLIN_LOCAL_STORE","/usr/src/app/merlin_file_store")),
         new PostgresStore(getEnv("MERLIN_DB_TYPE","postgres"),
                           getEnv("MERLIN_DB_USER","aerie"),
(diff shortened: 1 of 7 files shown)


dyst5422 (approved):
I'm not Java savvy enough to really have a good basis for my review, but it looks reasonable to me

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
```diff
@@ -53,7 +50,7 @@ public static void main(final String[] args) {
     // Configure an HTTP server.
     final var javalin = Javalin.create(config -> {
       config.showJavalinBanner = false;
-      if (configuration.javalinLogging().isEnabled()) config.enableDevLogging();
+      config.enableDevLogging();
```
mattdailis: The only thing that made me raise my eyebrows is the word "Dev" here, do you figure this is still appropriate for production?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -27,7 +29,7 @@
  * connected mission model repository.
  */
 public final class LocalMissionModelService implements MissionModelService {
-  private static final Logger log = Logger.getLogger(LocalMissionModelService.class.getName());
+  private static final Logger log = LoggerFactory.getLogger(LocalMissionModelService.class);
```
pcrosemurgy, line 32: Thanks for making this change! Do we also want to update these instances as well?
```
develop:aerie$ ag Logger.getLogger
merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
36:  private static final Logger log = Logger.getLogger(AerieAppDriver.class.getName());

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
30:  private static final Logger log = Logger.getLogger(LocalMissionModelService.class.getName());

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
43:  private static final Logger log = Logger.getLogger(SchedulerBindings.class.getName());
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
JoelCourtney: My understanding is that javalin allows you to set custom logging handlers for request events and websocket events, and `enableDevLogging` just sets them to a default. Without that, either javalin would produce no logs or I'd have to make custom handlers (and I don't think that would be useful).

This way, I use `enableDevLogging` to enable all output, and then filter out stuff below the logging level in slf4j.

pcrosemurgy: I'm assuming that this just enables debug-level logs within Javalin (see https://javalin.io/documentation#configuration):
> enable dev logging (extensive debug logging meant for development)

The actual level setting in the docker-compose file is the real gatekeeper here, so this just allows debug logs if/when you want to enable them. 

@JoelCourtney If you want you can probably wrap this in a conditional to check if the `slf4j` log level is debug or not. I think a `System.getProperty("org.slf4j.simpleLogger.defaultLogLevel").equalsIgnoreCase("debug")` may work?

pcrosemurgy: Since `System.getProperty()` can return a `null` you'd actually want:
```java
Optional.ofNullable(System.getProperty("org.slf4j.simpleLogger.defaultLogLevel")).ifPresent(p -> {
    if (p.equalsIgnoreCase("debug")) config.enableDevLogging();
}
```

pcrosemurgy: Or a simple `logger.isDebugEnabled() || logger.isTraceEnabled()`, sorry to complicate things above.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
JoelCourtney: Looks like you ran that on develop. I removed the instance in AerieAppDriver because it was never used, and I'll get to the instances in scheduler in AERIE-1734. Thanks!

pcrosemurgy:
Nice work Joel – would you mind also removing this row:
```
deployment/Environment.md
56:| `MERLIN_LOGGING`     | Whether or not you want Javalin to log server information | `string` | true                           |
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
pcrosemurgy: Ohh gotcha, perfect!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
JoelCourtney: `logger.isDebugEnabled()` is sufficient. Done. Thanks!

JoelCourtney:
Updated `deployment/Environment.md` and disabled dev logging for levels below INFO.

pcrosemurgy (approved):

PR #94, opened 2022-03-09
[HOTFIX] Boolean enums should be type `boolean`

goetzrrGit:
* **Review:** By commit  

## Description
Converting the command_dictionary boolean enums to typescript yields "TRUE | FALSE" enum, which isn't usable in the authoring part of command expansion. This code change allows you to do the following below

ex.

BAKE_BANANA_BREAD(activity.glutenFree)
BAKE_BANANA_BREAD(!activity.glutenFree)

### DIFF
--- a/command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
+++ b/command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
@@ -174,7 +174,16 @@ ${doc}
 function mapArgumentType(argument: ampcs.FswCommandArgument, enumMap: ampcs.EnumMap): string {
   switch (argument.arg_type) {
     case "enum":
-      return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+      // boolean enum shouldn't be "TRUE | FALSE" but of `boolean` type
+      if (
+        enumMap[argument.enum_name].values.length === 2 &&
+        enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === "true") &&
+        enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === "false")
+      ) {
+        return "boolean";
+      } else {
+        return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+      }
     case "boolean":
       return "boolean";
     case "float":


command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -174,7 +174,12 @@ ${doc}
 function mapArgumentType(argument: ampcs.FswCommandArgument, enumMap: ampcs.EnumMap): string {
   switch (argument.arg_type) {
     case "enum":
-      return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+      // boolean enum shouldn't be "TRUE | FALSE" but of `boolean` type
+      if (argument.enum_name.toLocaleLowerCase().includes("boolean")) {
```
pcrosemurgy: The `includes()` here worries me that a false-positive could occur.

dyst5422: Agreed. We shouldn't do nominal casting here - thats a footgun waiting to go off.

We should explicitly check that the enum has exactly two variants of 'false' and 'true' ie checking that is structurally a boolean.

goetzrrGit: Got it. let me change that.

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -174,7 +174,17 @@ ${doc}
 function mapArgumentType(argument: ampcs.FswCommandArgument, enumMap: ampcs.EnumMap): string {
   switch (argument.arg_type) {
     case "enum":
-      return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+      // boolean enum shouldn't be "TRUE | FALSE" but of `boolean` type
+      if (
+        enumMap[argument.enum_name].values.length === 2 &&
+        enumMap[argument.enum_name].values.filter(({ symbol, numeric }) => {
+          return symbol.toLocaleLowerCase() !== "true" && symbol.toLocaleLowerCase() !== "false";
+        }).length === 0
```
pcrosemurgy: The double negative here (with the `!==` and the "array is empty" check) confuses me at first glance but it makes sense.

pcrosemurgy (approved):

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
```diff
@@ -174,7 +174,17 @@ ${doc}
 function mapArgumentType(argument: ampcs.FswCommandArgument, enumMap: ampcs.EnumMap): string {
   switch (argument.arg_type) {
     case "enum":
-      return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+      // boolean enum shouldn't be "TRUE | FALSE" but of `boolean` type
+      if (
+        enumMap[argument.enum_name].values.length === 2 &&
+        enumMap[argument.enum_name].values.filter(({ symbol, numeric }) => {
```
dyst5422: 100% stylistic and nit-picky 

For readability, can we consider switching this to a 
```
enumMap[argument.enum_name].values.length === 2
&& enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === 'true')
&& enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === 'false')
```

So it reads as "Must have 2 variants AND one must be true AND one must be false"?

Approved for merge regardless

dyst5422 (dismissed):

command-expansion-server/src/packages/lib/CommandTypeCodegen.ts
dyst5422: I just re-read this as "Must have 2 variants and there must be none that are not true and are not false", which makes perfect sense

So I'm going to retract my previous comment.

mattdailis: An enum with the values "true" and "TrUe" would match the code as written I think.

So I just want to point out that there is a semantic difference between Dylan's suggestion and the code as is... does it matter? 🤷

mattdailis: Is there a particular reason we want this match to be case insensitive?

goetzrrGit: If the enum is of length 2 with 'TrUe' as the value, then the most likely intent of the user is probably 'true'.

mattdailis: These values aren't coming from users though, right? They're coming from merlin server:

https://github.com/NASA-AMMOS/aerie/blob/ad11d0997b86a04f27efacc1212197f6fe3252ef/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/SerializedValueJsonParser.java#L67-L69

Now I suppose you could consider the developers of merlin-server "users" since it's from a different component... I'm not sure what the best way to go is. I'm not going to request changes, but I want to call out that there is an edge case that would cause wrong behavior: namely an enum that has "true" in it twice, with different capitalizations.

dyst5422: Good point on the double true matching though Matt. I think we want the incantation of `"Must have 2 variants AND one must be true AND one must be false"`

dyst5422: Re. the capitalization, yeah as Ryan said, it's pretty unambiguous that they intend a boolean if they have 'TrUe' and 'fALse', despite them acting like a sea sponge

goetzrrGit: @mattdailis the values are coming from the user's command_dictionary file. Good point on the possibility of True,True as the enum

goetzrrGit: I can make the change to `"Must have 2 variants AND one must be true AND one must be false"`

dyst5422 (approved):

mattdailis (approved):

PR #96, opened 2022-03-09
build: remove env vars from docker-compose

camargo:
- UI scheduler env vars are obsolete


mattdailis (approved):

JoelCourtney (approved):
I grepped for other uses and found none.

PR #97, opened 2022-03-10
[AERIE-1734] Use configurable logging in scheduler

JoelCourtney:
* **Tickets addressed:** AERIE-1734
* **Review:** By commit  

## Description
Same general approach as for merlin server, but with a lot more bookkeeping.

- Deleted CSTNU from the dependencies, freeing up sfl4j
- Made log level filtering and output file configuration possible with java system properties in docker-compose.yml.
- Changed all instances of `System.out.println(...)` or `e.printStackTrace()` to use a slf4j logger.

If you (the reviewer) are feeling up to it, it might be worth checking that I got the intended log levels right when converting from `System.out.println`. They're all in the "Use slf4j for all scheduler logging" commit. There's a lot of them though, and none of them are likely off by more than one level, so that's up to you.

## Verification
I tested output from Javalin and verified that it filtered properly. I was not able to trigger any of the existing error or warning logs in the scheduler package, so instead I put a warning in the constructor of `gov.nasa.jpl.aerie.scheduler.ActivityType` and verified that it popped out when I ran the test suite. (That warning is gone now of course.)

## Documentation
I've updated `deployment/Environment.md` to reflect this.


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/config/AppConfiguration.java
```diff
@@ -17,7 +17,7 @@
 //TODO: remove backdoor access to directly mounted merlinFileStore (eg via merlin endpoint for downloading mission jars)
 public record AppConfiguration(
     int httpPort,
-    JavalinLoggingState javalinLogging,
+    boolean enableJavalinDevLogging,
```
mattdailis, line 20: 🔶  I think an enum would be more descriptive than a boolean here - would it be possible to revive JavalinLoggingState?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TimeRangeExpression.java
```diff
@@ -90,6 +90,7 @@ public Windows computeRange(Plan plan, Windows domain) {
       }
     }
     //out.println(name + " -> " + inter);
+    // if you uncomment the above line please use slf4j instead of println
```
mattdailis: 🔻 Rather than use our codebase as a message board, let's go ahead and delete commented out code 🤣 

For your context: the scheduler is one of the newest parts of the aerie system, and it's in the (painful) process of transforming from a prototype to a product. Hence the unused libraries, commented out code, etc.

mattdailis (changes requested):
Overall looks great to me - I have one small request for discussion, and one request to delete a comment.

I'm less concerned about the exact logging levels at the moment - I'm sure we will adjust those as we learn better what the failure modes of the scheduler are and how we want the system to handle them.

Since I haven't really worked with a logger before, I am curious to learn when it's appropriate to use `log.error`, and when it's appropriate to throw an exception. (Or should we do both?)

Key:

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TimeRangeExpression.java
JoelCourtney: Whoops, yeah I realized that was dumb halfway through and removed most of them, must have forgot this one.

mattdailis: Awesome, thanks 👍

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/config/AppConfiguration.java
JoelCourtney: I'd rather not. In general I'm on board for descriptive enums even when they have only two variants, but this one in particular is a truthy-or-falsy enum. The information it carries is conceptually a boolean, so it might as well just be a boolean.

I think if `null` was a meaningful third value, that would be a different story, but it gets passed through `Objects.requireNonNull`.

mattdailis: Okay, I won't put up too much of a fight on this one 🙂  I'll make my case about enums for future reference.

The nice thing about enums is the values by themselves hold meaning (`JavalinLoggingState.ENABLED` says more than `true` does). Booleans hold no meaning without the name of the function parameter. For example, consider the relative readability of these two lines of code:

```java
new AppConfiguration(JavalinLoggingState.Enabled);
```

```java
new AppConfiguration(true);
```
In the second case, you have to go to the definition of `AppConfiguration` to have any clue what's going on.

The other nice thing is that an enum has room to grow - if we want to add more options, we wouldn't need to change the type of the variables storing it.

Last point, regarding `null`: Our stance on null is that we try to avoid it as much as is practical, preferring to use things like `Optional`, enums, and sealed interfaces instead.

I'll get off my soapbox now and give you an approval 😅

mattdailis (approved):

kroffo (approved):
Looks good to me!

JoelCourtney:
I updated a bit of AppConfiguration's javadoc; that's what the force push was for.

> Since I haven't really worked with a logger before, I am curious to learn when it's appropriate to use `log.error`, and when it's appropriate to throw an exception. (Or should we do both?)

My understanding is that exceptions are for when you want to break the usual control flow and delegate error handling to some other code. If you can gracefully handle an error in the same place it is detected/created then there's no need to throw/catch.

Twisol:
> Since I haven't really worked with a logger before, I am curious to learn when it's appropriate to use `log.error`, and when it's appropriate to throw an exception. (Or should we do both?)

FWIW, I'm a fan of [this fellow's](https://blog.ploeh.dk/2020/03/23/repeatable-execution/) thoughts on logging.

> The key is repeatability. Just like you should be able to [reproduce builds](https://en.wikipedia.org/wiki/Reproducible_builds) and [repeat deployments](https://en.wikipedia.org/wiki/Continuous_delivery), you should also be able to reproduce execution. If you can replay what happened when a problem manifested itself, you can troubleshoot it. You need to log just enough data to enable you to repeat execution.

He takes this to a further extreme than I would -- I'd include logging at module boundaries, so you have enough information to reproduce execution independently in each module -- but I think there's a good kernel of philosophy there.

PR #100, opened 2022-03-10
[HOTFIX] Remove duration parameter from createAllPlanActivities

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Previously the duration was encoded as an additional parameter on activity instances. This commit deletes the code that does that.

The issue that @camargo found was that running on the scheduler on a plan _twice_ would cause a null pointer exception. The flow was as follows:
1. First run of the scheduler places new activities in the plan. These activities have an extraneous duration parameter
2. Second run of the scheduler attempts to load these activities from the plan. When building the activity instance objects, it tries to look up the duration parameter, and gets null back because it doesn't exist in the activity type's schema.

## Verification

@adrienmaillard and I joined forces in a huddle and were able to reproduce @camargo 's error. After making the change, we followed the same steps and did not see the error anymore.

## Documentation

No documentation needs to be updated (yet), since the behavior of the scheduler is still in flux.

## Future work

- @adrienmaillard and I discussed that it would be nice if the scheduler could load all of the activity durations from the existing plan based on a previously run simulation
- I proposed some refactoring to split ActivityInstance into PlannedActivityInstance and SimulatedActivityInstance
- AerieController lives in the `scheduler`, but seems to be more of a `scheduler-server` concern

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
@@ -365,9 +365,6 @@ public record GraphQLMerlinService(URI merlinGraphqlURI) implements MerlinServic
 
     for (final var act : orderedActivities) {
       requestSB.append(actPre.formatted(planId.id(), act.getType().getName(), act.getStartTime().toString()));
-      if (act.getDuration() != null) {
-        requestSB.append(argFormat.formatted("duration", getGraphQLValueString(act.getDuration())));
-      }
       for (final var arg : act.getArguments().entrySet()) {
         final var name = arg.getKey();
         var value = getGraphQLValueString(arg.getValue());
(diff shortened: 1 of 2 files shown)


camargo (approved):
Tested this from the UI and the error seems to be gone!

JoelCourtney:
Not directly related to this PR, but while digging through `ActivityInstance` to see how the duration is used, it was very confusing to see its member variables buried underneath a bunch of methods, some of them being autogenerated boilerplate. It looks somewhat intentional too, because a few other classes do the same thing. Is there a reason behind that?

adrienmaillard:
> Not directly related to this PR, but while digging through `ActivityInstance` to see how the duration is used, it was very confusing to see its member variables buried underneath a bunch of methods, some of them being autogenerated boilerplate. It looks somewhat intentional too, because a few other classes do the same thing. Is there a reason behind that?

@JoelCourtney it was the original dev's style (you can ask Steve S.). It is being refactored little by little.

adrienmaillard (approved):

PR #102, opened 2022-03-10
[AERIE-1710] Strengthen merlin-server Postgres Boundary

kroffo:
* **Tickets addressed:** AERIE-1710
* **Review:** By commit  

## Description
This branch removes all merlin-server models from the Postgres actions used by the PostgresResultsCellRepository, replacing with postgres-specific models. This establishes a much more well-defined boundary between the merlin-server and postgres by moving the boundary uniformly to the Postgres repository classes.

There are quite a few commits, but they are pretty much all quite small, so don't be intimidated by the number of commits! There's just a lot of files that needed refactoring, so it didn't make sense to make few large commits with a bunch of unrelated refactorings.

I want to specifically call out that I did rename the state column of simulation_dataset to status. This is a more accurate label for this column, as the state is the status combined with relevant data such as a failure reason or simulation results. This takes place in the penultimate commit, so pay special attention there.

## Verification
I built the system along the way whenever I made a commit that touched more than two spots, and of course did a full test at the end. I run with `aerie-sh -c; docker volume prune -f; aerie-sh -br` which wipes the database too, despite the fact that the database is untouched by these commits. I didn't get to actually run anything because for some reason my UI won't load activity types even on develop, but I'm out of time for the day so I'll test tomorrow, or you can now! We absolutely should not merge this without testing things out!

## Documentation
No documentation updates required. All changes are internal refactoring changes.

## Future work
Though the boundary between the server and postgres actions has been strengthened, we still will want to refactor the postgres actions and models at some point so that they aren't just thrown into one massive package. A problem for another day though, as the access-modifiers for all the postgres actions will need to change for that.


merlin-server/sql/merlin/tables/simulation_dataset.sql
```diff
@@ -14,7 +14,7 @@ create table simulation_dataset (
   dataset_revision integer null,
 
   -- Simulation state
-  state text not null default 'incomplete',
+  status text not null default 'incomplete',
```
mattdailis: While we're here I think it would be helpful to make `status` an enum.

I also want to note that this is a merge conflict with #107 that we'll want to test carefully

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityAction.java
```diff
@@ -1,40 +1,27 @@
...
       a.type,
-      coalesce(arguments.arguments, '[]'::json) as arguments
-    from plan as p
-    left join activity as a
-      on p.id = a.plan_id
-    left join json_activity_arguments as arguments
-      on a.id = arguments.activity_id
-    where p.id = ?
-      and a.id = ?
+      ceil(extract(epoch from a.start_offset) * 1000*1000) as start_offset_in_micros,
```
mattdailis, line 20: Wrapping my head around this - so is this doing a conversion from timestampz to microseconds?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
```diff
@@ -438,11 +438,13 @@ public void commit() throws NoSuchPlanException {
...
-              this.planId,
+              this.planId.id(),
               this.name.orElse(null),
               this.startTime.orElse(null),
               this.endTime.orElse(null));
         }
+      } catch (final FailedUpdateException ex) {
+        throw new NoSuchPlanException(this.planId);
       } catch (final SQLException ex) {
         throw new DatabaseException("Failed to update a plan", ex);
```
mattdailis, line 449: I'm a little confused on the distinction between `FailedUpdateException`, `SQLException`, `DatabaseException`, and `NoSuchPlanException`.

In particular - the mapping of `FailedUpdateException -> NoSuchPlanException` in the first catch block, and `SQLException -> DatabaseException` with the text "Failed to update a plan" seems particularly confusing to me.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanAction.java
```diff
@@ -10,39 +10,17 @@
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.Map;
+import java.util.Optional;
 
 /*package-local*/ final class GetPlanAction implements AutoCloseable {
   private static final @Language("SQL") String sql = """
-    with
```
mattdailis, line 14: I'm curious about the tradeoff of making actions one-to-one with tables. One one hand, it seems like a very clean separation of database implementation details from model types. On the other hand, it seems like there are situations where we would now have to call multiple actions (and thus take multiple trips to the database) where before we could use SQL to join multiple tables together.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityAction.java
kroffo: Yes, the postgres `extract` function is converting the stored start_offset to a number of seconds from the Unix Epoch (Jan 1, 1970, midnight). The multiplication converts this from seconds to microseconds, and ceil ensures the result is a whole number. This was copy/pasted from the `GetActivitiesAction`.

[Documentation](https://www.postgresql.org/docs/8.1/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT) for `extract`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
kroffo: The `FailedUpdateException` tells you that the update was attempted, and SQL executed without issue, but the operation did not produce the expected result. I'm unsure if this can actually happen, but the code is written to handle all possible results returned from the PreparedStatement in the action. Basically the SQL returns the number of rows affected by the update, and if that number is not 1 then the update failed to perform what was expected (updating the one relevant row)

The `SQLException` tells you something went wrong during the SQL execution, i.e. the actual database operation itself caused the exception directly.

I suppose the message provided to the `DatabaseException` here makes this confusing, but hopefully that makes it more clear. Let me know what you think!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanAction.java
kroffo: Though most actions do only interact with one table, this is not the case for all. Examine the `LookupSimulationDatasetAction` class, which reads from 5 tables in a single query! That said, I do believe I've seen some areas where we perform multiple operations where we could have performed one. Much of this is an artifact of my inexperience when I was first working on Postgres actions. My rope bridge, if you will.

merlin-server/sql/merlin/tables/simulation_dataset.sql
kroffo: Great idea Matt, I'll implement that now!

kroffo:
Latest force push just rebases of develop without changes.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
mattdailis: That explanation does help :+1: I think I'm okay with this as-is then. Resolving this comment now.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanAction.java
mattdailis: I see - so it's less about actions being one-to-one with tables, and more about actions dealing with primitive data-types and simple records, rather than manipulating our higher-level domain model objects. Thanks for the explanation :+1:

mattdailis (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanAction.java
kroffo: Exactly. The actions provide a record of the state of the database that's useful to the `PostgresResultsCellRepository`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivitiesAction.java
```diff
@@ -1,54 +1,28 @@
...
-    from plan as p
-    left join json_activities as activities
-      on activities.plan_id = p.id
-    where p.id = ?
+      a.id,
+      a.type,
+      ceil(extract(epoch from a.start_offset) * 1000*1000) as start_offset_in_micros,
+      a.arguments
+    from activity as a
+    where a.plan_id = ?
```
goetzrrGit, line 25: Search with very precise time is scary especially at the microsecond level.

In the past, on MSL we had issues with SPICE when converting JAVA `date` to `LMST`  back to `date`. This locked us into the precision of second on the Planning UI.

But if you are not doing a conversion and just storing the java `date` then it should be fine.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -422,6 +422,7 @@ private static void postSimulatedActivities(
   public static final class PostgresResultsCell implements ResultsProtocol.OwnerRole {
     private final DataSource dataSource;
     private final SimulationRecord simulation;
+    private final PlanId planId;
```
goetzrrGit, line 443: Just curious why you wouldn't replace `PlanId` with `long`?

It seems like you are using a lot of `PlanID.id()` which is just the `long` value.

goetzrrGit (commented):
Looks good just had some questions.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivitiesAction.java
kroffo: Well, in this PR I'm just changing the structure of the SQL query, but this line was already present in the original SQL. This change isn't changing the way the start_offset is queried itself.

mattdailis: My understanding is that the query is based on `a.plan_id` - so the precise time is an output, rather than an input to the query. If we were querying on exact time, @goetzrrGit , I agree that would be potentially problematic.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
kroffo: The PostgresResultsCellRepository serves as the boundary between the postgres models and the server models. Internally, we keep a `PlanId` instead of a `long` because we can easily get a long out of it when calling actions while simply passing the value in when a `PlanId` is required. However, if we used a `long` then every time we need a `PlanId` we would need to construct a new one from the `long`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivitiesAction.java
goetzrrGit: Got it thanks for the clarification.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
goetzrrGit: Thank you for the clarification. That makes sense.

goetzrrGit (approved):

PR #103, opened 2022-03-10
refactor: add more Hasura metadata for scheduling

camargo:
## Description

- add goal relationship to analyses
- add goal analysis relationship to request and satisfying activities

## Example

The following query gets all spec goals for a given spec id. Then for each goal it gets the last two analyses based on the scheduling request spec revision. It also demonstrates how to get a count of satisfying activities in an analysis using an aggregate query.

```gql
query SchedulingSpecGoals($specification_id: Int!) {
  scheduling_specification_goals(where: { specification_id: { _eq: $specification_id } }) {
    goal {
      name
      analyses(order_by: { request: { specification_revision: desc } }, limit: 2) {
        request {
          specification_revision
        }
        satisfied
        satisfying_activities {
          activity_id
        }
        satisfying_activities_aggregate {
          aggregate {
            count
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/deployment/hasura/metadata/databases/AerieScheduler/tables/public_scheduling_goal.yaml
+++ b/deployment/hasura/metadata/databases/AerieScheduler/tables/public_scheduling_goal.yaml
@@ -1,3 +1,11 @@
 table:
   name: scheduling_goal
   schema: public
+array_relationships:
+- name: analyses
+  using:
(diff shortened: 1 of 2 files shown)


pcrosemurgy (approved):
🔥

mattdailis (approved):

PR #104, opened 2022-03-11
[FEAT] Expansion Hasura Relationships

dyst5422:
Build out data relationships for hasura to be able to traverse the data
graph for expansion related data

* **Tickets addressed:** NA
* **Review:** By commit  

## Description

Building out the data relationships in command expansion in hasura.

Interesting one is the many-to-many expansion_set to expansion_rule which required building out views to do the proper joins

## Verification

I ran the metadata and tested with some queries. Some of the items don't have data or ways to insert it yet, so those are yet to be tested.

If @mattdailis merges in his api tests package, I'd love to throw some tests of these relationships in there

## Documentation

None

## Future work

There are a couple TODO items about to activity instances and how they need to be related to merlin data, but there are some open questions there re. our data model.

### DIFF
--- a/command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
+++ b/command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
@@ -1,5 +1,6 @@
 create table activity_instance_commands (
   activity_instance_id integer not null,
+  -- TODO: Change this to be the actual simulated instance id (span_id?)
   commands jsonb not null,
   expansion_run_id integer not null,
 
--- a/command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
+++ b/command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
@@ -1,6 +1,5 @@
 create table activity_instance_commands (
   activity_instance_id integer not null,
-  -- TODO: Change this to be the actual simulated instance id (span_id?)
   commands jsonb not null,
   expansion_run_id integer not null,
 
--- a/command-expansion-server/sql/commanding/tables/expansion_run.sql
+++ b/command-expansion-server/sql/commanding/tables/expansion_run.sql
@@ -1,7 +1,7 @@
 create table expansion_run (
   id integer generated always as identity,
 
-  dataset_id integer not null,
+  simulation_dataset_id integer not null,
   expansion_set_id integer not null,
 
   constraint expansion_run_primary_key
@@ -14,7 +14,7 @@ comment on table expansion_run is e''
   'The configuration for an expansion run for a plan.';
 comment on column expansion_run.id is e''
   'The synthetic identifier for this expansion run.';
-comment on column expansion_run.dataset_id is e''
-  'The dataset id used to generate this expansion run.';
+comment on column expansion_run.simulation_dataset_id is e''
+  'The simulation dataset id used to generate this expansion run.';
 comment on column expansion_run.expansion_set_id is e''
   'The command dictionary, mission model, and expansion set id.';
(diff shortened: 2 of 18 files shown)


goetzrrGit (commented):
I was playing around with these new configurations and was able to get `expansion logic ` from an `expansion set` in one trip.

```
query MyQuery {
  expansion_set(where: {expansion_rules: {activity_type: {_eq: "BakeBananaBread"}}}) {
    expansion_rules {
      expansion_logic
    }
  }
}
```

I see relationships added to the original tables are the new `*_view.yaml` tables addressing this issue:

> adding a constraint to our expansion_set_to_rule table that there can only be 1 expansion_rule with a given activity_type in the expansion_set

dyst5422:
> I see relationships added to the original tables are the new `*_view.yaml` tables addressing this issue:
> 
> > adding a constraint to our expansion_set_to_rule table that there can only be 1 expansion_rule with a given activity_type in the expansion_set

Yeah, this constraint is one I think someone with a brighter mind might take a minute to try to figure out. The views themselves don't solve this - they just allow us to directly do the through-joins.

goetzrrGit:
> Yeah, this constraint is one I think someone with a brighter mind might take a minute to try to figure out. The views themselves don't solve this - they just allow us to directly do the through-joins.

Let's push that to v.2 and maybe ask @kroffo for help to see if it is possible via the schema. We can also try to add some logic in the server itself to check for that constraint.

dyst5422:
> > Yeah, this constraint is one I think someone with a brighter mind might take a minute to try to figure out. The views themselves don't solve this - they just allow us to directly do the through-joins.
> 
> Let's push that to v.2 and maybe ask @kroffo for help to see if it is possible via the schema. We can also try to add some logic in the server itself to check for that constraint.

Agreed.

I'm hoping @kroffo or @patkenneally can come up with a clever way to enforce this constraint in the db so we can ensure our data can't get in an invalid state, but worst case we can enforce in the server as you said and hope nobody tries to poke the database directly during development.

command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
```diff
@@ -1,5 +1,6 @@
 create table activity_instance_commands (
   activity_instance_id integer not null,
+  -- TODO: Change this to be the actual simulated instance id (span_id?)
```
mattdailis: Is this something we've made a decision on?

Side note: personally I'm okay with the column name being something that makes sense in the command expansion context, and have it refer to the `span` table's id.

dyst5422: In either case, we need it to point to the activity instance

command-expansion-server/sql/commanding/tables/expansion_run.sql
```diff
@@ -1,7 +1,7 @@
 create table expansion_run (
   id integer generated always as identity,
 
-  dataset_id integer not null,
+  simulation_dataset_id integer not null,
```
pcrosemurgy, line 4: 👍

command-expansion-server/sql/commanding/tables/activity_instance_commands.sql
goetzrrGit: Lets keep `activity_instance_id` for now as `span` might be renamed. Also, the user will be going through the Hasura API and they will understand `activity_instance_id`  better than `span_id` when they do their query.

goetzrrGit (approved):

pcrosemurgy (approved):

command-expansion-server/sql/commanding/tables/expansion_run.sql
goetzrrGit: 👍

PR #106, opened 2022-03-14
[AERIE-1741/29/28/54] TypeScript instantiations of simulated activities

pcrosemurgy:
* **Tickets addressed:** [AERIE-1741](https://jira.jpl.nasa.gov/browse/AERIE-1741) and [AERIE-1729](https://jira.jpl.nasa.gov/browse/AERIE-1729)
* **Review:** By file  

## Description

Review by file is suggested since there are refactoring commits interspersed.

This PR adds the `/expand-all-activity-instances` command expansion endpoint and associated `expandAllActivities` Hasura Action.

### Sub Tasks
- [x] Expansion set dataloader
- [x] Pull in actual expansion logic with expansion set dataloader
- [x] Worker pooling with piscina
- [x] Store results in DB

## Verification

<details>
<summary>Expand</summary>

### Success

1. Follow https://github.com/NASA-AMMOS/aerie/wiki/Command-Expansion-Guide to upload a command dict.
2. Create a banananation mission model, create a plan, add `BakeBananaBread` to the timeline, populate all required arguments, and run the simulation. This will create a simulation dataset needed for the expansion execution later.
3. Follow the steps described in the link above to upload an expansion for `BakeBananaBread`:
```graphql
mutation uploadExpansionLogic {
  addCommandExpansionTypeScript(
    activityTypeName: "BakeBananaBread"
    expansionLogic: "ZXhwb3J0IGRlZmF1bHQgZnVuY3Rpb24gQmFrZUJhbmFuYUJyZWFkRXhwYW5zaW9uTG9naWMoCiAgcHJvcHM6IHsKICAgIGFjdGl2aXR5OiBBY3Rpdml0eVR5cGU7CiAgfSwKICBjb250ZXh0OiBDb250ZXh0Cik6IEV4cGFuc2lvblJldHVybiB7CiAgcmV0dXJuIFsKICAgIFBSRUhFQVRfT1ZFTihwcm9wcy5hY3Rpdml0eS50ZW1wZXJhdHVyZSksCiAgICBQUkVQQVJFX0xPQUYocHJvcHMuYWN0aXZpdHkudGJTdWdhciwgcHJvcHMuYWN0aXZpdHkuZ2x1dGVuRnJlZSksCiAgICBCQUtFX0JSRUFELAogIF07Cn0="
  ) {
    id
  }
}
```
This expansion logic contains the base64-encoded TypeScript source:
```typescript
export default function BakeBananaBreadExpansionLogic(
  props: {
    activity: ActivityType;
  },
  context: Context
): ExpansionReturn {
  return [
    PREHEAT_OVEN(props.activity.temperature),
    PREPARE_LOAF(props.activity.tbSugar, props.activity.glutenFree),
    BAKE_BREAD,
  ];
}
```
5. Create an expansion set, referring to the newly created expansion rule, command dict., and mission model:
```gql
mutation createExpansionSet {
  createExpansionSet(
    commandDictionaryId: 1
    missionModelId: 1
    expansionIds: 1
  ) {
    id
  }
}
```
6. Lastly, run all expansions for the simulation dataset and expansion set:
```gql
mutation ExpandAll {
  expandAllActivities(expansionSetId: 1, simulationDatasetId: 1) {
    id
    expandedActivityInstances {
      id
      commands
      errors
    }
  }
}
```
This results in:
```json
{
  "data": {
    "expandAllActivities": {
      "id": 1,
      "expandedActivityInstances": [
        {
          "id": 1,
          "commands": [
            {
              "arguments": [
                350
              ],
              "stem": "PREHEAT_OVEN"
            },
            {
              "arguments": [
                4,
                true
              ],
              "stem": "PREPARE_LOAF"
            },
            {
              "arguments": [],
              "stem": "BAKE_BREAD"
            }
          ],
          "errors": []
        }
      ]
    }
  }
}
```

### Expansion Failure

When testing with an expansion rule that has a compilation error (`PRiEHEAT_OVEN` instead of `PREHEAT_OVEN`) the `expandAllActivities` mutation should return:
```json
{
  "data": {
    "expandAllActivities": {
      "id": 4,
      "expandedActivityInstances": [
        {
          "id": 1,
          "commands": null,
          "errors": [
            {
              "sourceContext": " 6|   return [\n>7|     PRiEHEAT_OVEN(props.activityInstance.attributes.arguments.temperature),\n        ~~~~~~~~~~~~~\n 8|     PREPARE_LOAF(",
              "location": {
                "line": 6,
                "column": 4
              },
              "message": "TypeError: Cannot find name 'PRiEHEAT_OVEN'. Did you mean 'PREHEAT_OVEN'?",
              "stack": "at BakeBananaBreadExpansionLogic(6:4)"
            }
          ]
        }
      ]
    }
  }
}
```

### Activity Type Uniqueness Failure

If an expansion set contains expansions that refer to the same activity type then an error should be thrown:
```json
{
  "errors": [
    {
      "extensions": {
        "internal": {
          "error": "expecting 2xx or 4xx status code, but found 500",
          "response": {
            "status": 500,
            "body": {
              "message": "duplicate key value violates unique constraint \"expansion_set_to_rule_set_id_activity_type_key\""
            },
```
For example, given expansion rules with IDs `1` and `2`, the following expansion set creation should result in the error above:
```gql
mutation createExpansionSet {
  createExpansionSet(
    commandDictionaryId: 1
    missionModelId: 1
    expansionIds: [1, 2]
  ) {
    id
  }
}
```

### Success & Failure

In a situation where an expansion set contains 2 expansions with 2 different associated activity types it is expected that each activity expansion may fail or pass in isolation. To test this we'll add another expansion rule, this time for the `BiteBanana` activity:
```typescript
export default function BiteBananaBreadExpansionLogic(
  props: {
    activityInstance: ActivityType;
  }
): ExpansionReturn {
  return [
    EAT_BANANA
  ];
}
```
This activity must also be placed on the sim. timeline and a new simulation run is needed. Note that this will result in a new sim. dataset id (in this case ID `2`).

A new expansion set containing both a successful expansion and an erroneous expansion is created with:
```gql
mutation createExpansionSet {
  createExpansionSet(
    commandDictionaryId: 1
    missionModelId: 1
    expansionIds: [1, 3]
  ) {
    id
  }
}
```
Where expansion rule IDs `1` and `3` respectively refer to both an erroneous expansion and a successful expansion.

Finally, using the new simulation dataset (with ID `2`):
```graphql
mutation D {
  expandAllActivities(expansionSetId: 4, simulationDatasetId: 2) {
    id
    expandedActivityInstances {
      id
      commands
      errors
    }
  }
}
```
Returns 2 expansion results, one with errors and one with commands, as expected:
```json
{
  "data": {
    "expandAllActivities": {
      "id": 7,
      "expandedActivityInstances": [
        {
          "id": 2,
          "commands": null,
          "errors": [
            {
              "sourceContext": " 6|   return [\n>7|     PRiEHEAT_OVEN(props.activityInstance.attributes.arguments.temperature),\n        ~~~~~~~~~~~~~\n 8|     PREPARE_LOAF(",
              "location": {
                "line": 6,
                "column": 4
              },
              "message": "TypeError: Cannot find name 'PRiEHEAT_OVEN'. Did you mean 'PREHEAT_OVEN'?",
              "stack": "at BakeBananaBreadExpansionLogic(6:4)"
            }
          ]
        },
        {
          "id": 3,
          "commands": [
            {
              "arguments": [],
              "stem": "EAT_BANANA"
            }
          ],
          "errors": []
        }
      ]
    }
  }
}
```

</details>

## Documentation

See https://github.com/NASA-AMMOS/aerie/wiki/Command-Expansion-Guide (thanks to @goetzrrGit).

## Future work

- User experience improvements (open to concrete suggestions here).

- Running
```gql
query activityTypescript {
  getActivityTypeScript(
    activityTypeName: "BakeBananaBread"
    missionModelId: "2"
  ) {
    typescript
  }
}
```
With a non-existent model ID results in an error complaining about the lack of `BakeBananaBread` from mission model ID 1.
- Possibly returning whole error object: `res.status(err.status ?? err.statusCode ?? 500).send({message: err});`
- Make IDs consistently typed within Hasura actions (see https://github.com/NASA-AMMOS/aerie/pull/106#discussion_r839936884).


command-expansion-server/src/lib/batchLoaders/expansionSetBatchLoader.ts
```diff
@@ -1,35 +1,75 @@
...
+  `, {
+    simulationDatasetIds: keys.map(key => key.expansionSetId),
+  });
 
+  return Promise.all(keys.map(async ({ expansionSetId }) => {
+    const expansionSet = expansion_set.find(table => table.id === expansionSetId);
+    if (expansionSet === undefined) {
+      return new ErrorWithStatusCode(`No simulation_dataset with id: ${expansionSetId}`, 404);
+    }
+    return expansionSet;
```
dyst5422: We should pull the typescript types for the command dictionary from the filesystem in here as well so we have those for expansion

command-expansion-server/src/app.ts
```diff
@@ -27,6 +27,8 @@ app.use(bodyParser.json({ limit: '25mb' }));
 DbExpansion.init();
 const db = DbExpansion.getDb();
 
+const piscina = new Piscina({ filename: path.resolve(__dirname, 'worker.js') });
```
dyst5422: __dirname doesn't exist when using ESM. I believe the proper incantation for this use case is
```
new URL('worker.js', import.meta.url).pathname
```

command-expansion-server/src/lib/batchLoaders/expansionSetBatchLoader.ts
pcrosemurgy: Latest commits should resolve this.

command-expansion-server/src/app.ts
pcrosemurgy: Latest commits should resolve this.

deployment/hasura/metadata/actions.graphql
```diff
@@ -144,6 +151,17 @@ type TypeScriptResponse {
...
 
+type ExpandAllActivitiesResponse {
+  id: Int!
+  expandedActivityInstances: [ExpandedActivityInstance!]!
+}
+
+type ExpandedActivityInstance {
+  id: Int!
+  commands: [Command!]
+  errors: [UserCodeError!]
```
pcrosemurgy: @dyst5422 The backend always responds with a non-null list of `errors`, so I was thinking we might as well propagate that through to the DB and this type?

I could see an argument for either though. Having `commands` and `errors` both nullable is "symmetric".

dyst5422: I'm in favor of having the errors be non-nullable since I think regardless of whether it successfully expanded or not, there is the potential for errors and we should explicitly capture that.

pcrosemurgy: Aren't all errors basically TS `UserCodeError`s? If so I don't think it'd be possible to have a successful expansion and a user code error (for a single activity instance expansion).

dyst5422: For now that's true, but may not be in the future. Just working on various services in the past has taught me to just assume that you are going to have the potential for returning errors alongside data.

dyst5422: For example - a user may write a very complicated expansion that handles internal errors and wants to return those errors alongside commands. We could easily add a context function for creating these errors for return alongside the commands.

pcrosemurgy: Got it. Maybe you're thinking some day there may be a different class of errors, not just compilation errors.

I'll make the change mentioned in the 1st comment here :)

command-expansion-server/src/lib/codegen/ActivityTypescriptCodegen.ts
```diff
@@ -0,0 +1,55 @@
+import { globalDeclaration, indent, interfaceDeclaration } from './CodegenHelpers.js';
+import {GraphQLActivitySchema, Schema, SchemaTypes} from '../batchLoaders/activitySchemaBatchLoader.js';
+
+const commonProperties =
+`readonly duration: Temporal.Duration;
+readonly startOffset: Temporal.Duration;`;
+
+export function generateTypescriptForGraphQLActivitySchema(activitySchema: GraphQLActivitySchema): string {
```
mattdailis, line 8: This function seems to be defined in two places (both in `lib/codegen/ActivityTypescriptCodegen.ts` and in `lib/getActivityTypescript.ts`). Can we delete and/or rename one of them?

command-expansion-server/src/polyfills.ts
```diff
@@ -0,0 +1,5 @@
+import { Temporal as TemporalPolyfill, toTemporalInstant } from '@js-temporal/polyfill';
+
+globalThis.Temporal = TemporalPolyfill;
```
mattdailis, line 3: I'm not sure I understand the need to set a global variable here. This is setting a global variable in the command-expansion-server, not in the user code.

command-expansion-server/jestGlobalSetup.js
```diff
@@ -0,0 +1,68 @@
+import { config } from 'dotenv';
+config();
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import fetch from 'node-fetch';
+import { Writable } from 'stream';
+import readline from 'node:readline/promises';
```
mattdailis, line 8: Note: This import breaks on node 16.4.0, but works on 17.8.0. [This line](https://github.com/NASA-AMMOS/aerie/blob/78cb0e4efe12f9e4f989087bbc6030498a606a61/command-expansion-server/build.gradle#L6) seems to imply that we're using `16.4.0`.

Although...even after installing node 17 I'm still unable to get the tests to pass. Maybe one of you could give me a quick tutorial.

command-expansion-server/src/polyfills.ts
dyst5422: Temporal is the upcoming standard that isn't in ecmascript yet, so we have to poly fill it. This is placing it into the global namespace so that it exists globally as it will when it gets in.

command-expansion-server/jestGlobalSetup.js
dyst5422: Testing node version requirement is higher than deployment version, yes.

command-expansion-server/src/lib/codegen/ActivityTypescriptCodegen.ts
pcrosemurgy: Good catch. One of them was named incorrectly and simply generated TS for `GraphQLActivity` types, I've renamed that function `generateTypescriptForGraphQLActivity`.

command-expansion-server/src/app.ts
```diff
@@ -109,52 +143,138 @@ app.post('/put-expansion-set', async (req, res) => {
...
+    id,
+    expandedActivityInstances,
+  });
   return;
 });
 
+// General error handler
 app.use((err: any, req: Request, res: Response, _next: NextFunction) => {
   logger.error(err);
   res.status(err.status ?? err.statusCode ?? 500).send(err.message);
```
mattdailis: According to the [hasura docs](https://hasura.io/docs/latest/graphql/core/actions/action-handlers/#returning-an-error-response), error responses are expected to be a JSON object with a `message` field (and an optional `extensions` field). I think making this change will make errors show up nicely in hasura. (Currently, they do show up, but they're wrapped in hasura's own "message field was missing" error)

```suggestion
  res.status(err.status ?? err.statusCode ?? 500).send({message: err.message});
```

deployment/hasura/metadata/actions.graphql
```diff
@@ -18,13 +18,20 @@ type Mutation {
 }
 
 type Mutation{
-  createExpansionSet (
-    commandDictionaryId: Int!,
-    missionModelId: Int!,
+  createExpansionSet(
+    commandDictionaryId: Int!
```
mattdailis, line 22: We're a little inconsistent across this file regarding the type of ids. Some of them are `Int!` and some of them are `ID!` (aka string).

mattdailis (approved):
I watched Paul test this yesterday, and it worked pretty well :+1:

command-expansion-server/src/app.ts
dyst5422: I would even say that just returning the entire error object would be good for debug-ability for the operators - we don't want them poking us every time they run into an error. I know that's generally a no-no security-wise, but we aren't exactly dealing with malicious actors here.

pcrosemurgy: Good catch @mattdailis! I'll make that change.

As for returning the whole error object, I'm not too opposed to that – but to move things along for this PR maybe we can table that decision for a follow-up?

deployment/hasura/metadata/actions.graphql
pcrosemurgy: Yeah, not sure which one we should be using. Seems to be split fairly evenly between `Int!` and `ID!` 😂
```
feature/aerie-1728-set-rule-uniq-activity-type:aerie$ ag 'id: id' deployment/hasura/metadata/actions.graphql
30:    missionModelId: ID!
37:    missionModelId: ID!
45:    missionModelId: ID!
52:  commandDictionaryId: ID!): TypeScriptResponse
57:    missionModelId: ID!
76:    missionModelId: ID!
83:    missionModelId: ID!
132:  id: ID!
feature/aerie-1728-set-rule-uniq-activity-type:aerie$ ag 'id: int' deployment/hasura/metadata/actions.graphql
3:    planId: Int!
22:    commandDictionaryId: Int!,
23:    missionModelId: Int!,
63:    specificationId: Int!
69:    planId: Int!
123:  datasetId: Int!
136:  id: Int!
140:  id: Int!
```

pcrosemurgy: May be a good idea to have a follow-up PR to make all IDs consistently typed?

dyst5422: Typing as an ID when it is an Int forces an inferred typecast to string that can cause issues. I think we should move away from ID since all of our ids are Int

command-expansion-server/src/app.ts
patkenneally: Create a ticket for it and we can pick it up in the next cycle.

pcrosemurgy: Just made https://jira.jpl.nasa.gov/browse/AERIE-1764

pcrosemurgy: @mattdailis just pushed this change :)

deployment/hasura/metadata/actions.graphql
pcrosemurgy: @patkenneally I created https://jira.jpl.nasa.gov/browse/AERIE-1765 to tackle this. Should be a quick ticket.

command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -20,18 +20,20 @@ function generateTypescriptCode(dictionary: ampcs.CommandDictionary): {
...
 
   // language=TypeScript
   const values = `
-${typescriptFswCommands.map((fswCommand) => fswCommand.value).join("\n")}
+${typescriptFswCommands.map((fswCommand) => fswCommand.value).join('\n')}
 export const Commands = {${dictionary.fswCommands
     .map((fswCommand) => `\t\t${fswCommand.stem}: ${fswCommand.stem},\n`)
-    .join("")}};
+    .join('')}};
+const globalThis = (0,eval)("this");
```
dyst5422: @mattdailis and I discovered this is unnecessary, so we can remove this line.

for reference, here is where this was concocted from https://stackoverflow.com/questions/9107240/1-evalthis-vs-evalthis-in-javascript/9107367#9107367 but proves unnecessary

command-expansion-server/src/app.ts
```diff
@@ -17,15 +31,35 @@ app.use(bodyParser.json({ limit: '25mb' }));
 
 DbExpansion.init();
 const db = DbExpansion.getDb();
-const graphqlClient = new GraphQLClient(getEnv().MERLIN_GRAPHQL_URL);
 
-const logger = getLogger("app");
+const piscina = new Piscina({ filename: `${path.resolve()}/build/worker.js` });
```
dyst5422: I think we should replace this with the URL incantation we use elsewhere.

```
new URL('worker.js', import.meta.url).pathname
```

command-expansion-server/src/app.ts
```diff
@@ -17,15 +31,35 @@ app.use(bodyParser.json({ limit: '25mb' }));
...
+
+  const context: Context = {
+    commandTypescriptDataLoader: new DataLoader(commandDictionaryTypescriptBatchLoader({graphqlClient})),
+    activitySchemaDataLoader: new DataLoader(activitySchemaBatchLoader({graphqlClient})),
+    simulatedActivityInstanceDataLoader: new DataLoader(simulatedActivityInstanceBatchLoader({graphqlClient})),
+    expansionSetDataLoader: new DataLoader(expansionSetBatchLoader({graphqlClient})),
+  };
+
+  res.locals.context = context;
+  return _next();
```
dyst5422: Minor cleanup - remove the leading underscore here and in the function parameters -> the leading underscore typically indicates and unused parameter (which this is not)

command-expansion-server/src/app.ts
```diff
@@ -1,14 +1,28 @@
...
+import { processDictionary } from './lib/codegen/CommandTypeCodegen.js';
+import { generateTypescriptForGraphQLActivitySchema } from './lib/codegen/ActivityTypescriptCodegen.js';
+import { InferredDataloader } from './lib/batchLoaders/index.js';
+import { commandDictionaryTypescriptBatchLoader } from './lib/batchLoaders/commandDictionaryTypescriptBatchLoader.js';
+import { activitySchemaBatchLoader } from './lib/batchLoaders/activitySchemaBatchLoader.js';
+import { simulatedActivityInstanceBatchLoader } from './lib/batchLoaders/simulatedActivityInstanceBatchLoader.js';
+import { expansionSetBatchLoader } from './lib/batchLoaders/expansionSetBatchLoader.js';
+import path from 'path';
+import Piscina from 'piscina';
+import executeExpansion from './worker.js';
```
dyst5422: We can change this to `import type executeExpansion from './worker.js';` since we are only using it as a type. That way, the import gets erased for runtime.

command-expansion-server/src/app.ts
```diff
@@ -109,52 +143,138 @@ app.post('/put-expansion-set', async (req, res) => {
...
+  });
   return;
 });
 
+// General error handler
 app.use((err: any, req: Request, res: Response, _next: NextFunction) => {
   logger.error(err);
-  res.status(err.status ?? err.statusCode ?? 500).send(err.message);
+  res.status(err.status ?? err.statusCode ?? 500).send({message: err.message});
+  return _next();
```
dyst5422: Can we remove the leading underscore in `_next` so that it isn't confused as being an unused parameter?

command-expansion-server/src/app.ts
```diff
@@ -17,15 +31,35 @@ app.use(bodyParser.json({ limit: '25mb' }));
...
+  activitySchemaDataLoader: InferredDataloader<typeof activitySchemaBatchLoader>,
+  simulatedActivityInstanceDataLoader: InferredDataloader<typeof simulatedActivityInstanceBatchLoader>,
+  expansionSetDataLoader: InferredDataloader<typeof expansionSetBatchLoader>,
+};
+
+app.use(async(req: Request, res: Response, _next: NextFunction) => {
+  const graphqlClient = new GraphQLClient(getEnv().MERLIN_GRAPHQL_URL);
+
+  const context: Context = {
+    commandTypescriptDataLoader: new DataLoader(commandDictionaryTypescriptBatchLoader({graphqlClient})),
```
dyst5422: I forgot about this, we need to add a cache-key function to each of these.

The line 48 should be 
```
    commandTypescriptDataLoader: new DataLoader(commandDictionaryTypescriptBatchLoader({graphqlClient}), {cacheKeyFn: objectCacheKeyFunction}),

```
and the rest (lines 49-51) should similarly add that second parameter to set the cacheKeyFn.

WHAT THIS DOES:
Adjusts how the cache-key is created and checked in the dataloader for in-request caching. The object cache-key function just concatenates the object parameters together as the cache key.

command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -173,50 +175,46 @@ ${doc}
...
-        enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === "true") &&
-        enumMap[argument.enum_name].values.some(({ symbol, numeric }) => symbol.toLocaleLowerCase() === "false")
+        enumMap[argument.enum_name].values.some(({ symbol }) => symbol.toLocaleLowerCase() === 'true') &&
+        enumMap[argument.enum_name].values.some(({ symbol }) => symbol.toLocaleLowerCase() === 'false')
       ) {
-        return "boolean";
+        return 'boolean';
       } else {
-        return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(" | ");
+        return enumMap[argument.enum_name].values.map((value) => `'${value.symbol}'`).join(' | ');
```
dyst5422: We should wrap the output in parentheses for this one.

command-expansion-server/src/lib/mapGraphQLActivityInstance.ts
```diff
@@ -0,0 +1,79 @@
...
+    case SchemaTypes.Real:
+      return value;
+    case SchemaTypes.Duration:
+      return Temporal.Duration.from(parse(value).toISOString());
+    case SchemaTypes.Boolean:
+      return value;
+    case SchemaTypes.String:
+      return value;
+    case SchemaTypes.Series:
+      schema.items;
```
dyst5422: 48 seems like a line we can remove.

command-expansion-server/src/lib/mapGraphQLActivityInstance.ts
```diff
@@ -0,0 +1,79 @@
...
+      return value.map((value: any) => convertType(value, schema.items));
+    case SchemaTypes.Struct:
+      const struct: { [attributeName: string]: any } = {};
+      for (const [attributeKey, attributeSchema] of Object.entries(schema.items)) {
+        struct[attributeKey] = convertType(value[attributeKey], attributeSchema);
+      }
+      return struct;
+    case SchemaTypes.Variant:
+      if (schema.variants.length === 1 && schema.variants[0].key === 'VOID') {
```
dyst5422, line 54: Do we need to handle implicit boolean types like we have for command arguments?

ie if variants is exactly 'true' and 'false'

command-expansion-server/test/batchLoaders/simulatedActivityInstanceBatchLoader.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  simulationDatasetId = await executeSimulation(graphqlClient, planId);
+});
+
+afterAll(async () => {
+  await removeActivity(graphqlClient, activityId);
+  await removePlan(graphqlClient, planId);
+  await removeMissionModel(graphqlClient, missionModelId);
+});
+
+it('[XFAIL] should load simulated activity instance', async () => {
```
dyst5422: Can we get this test fixed up? I can take a stab at it

deployment/hasura/metadata/actions.yaml
```diff
@@ -15,6 +15,10 @@ actions:
     definition:
       kind: synchronous
       handler: http://aerie_commanding:3000/put-expansion-set
+  - name: expandAllActivities
+    definition:
+      kind: synchronous
```
dyst5422, line 20: What does this `synchronous` do?

deployment/hasura/metadata/actions.yaml
```diff
@@ -94,3 +100,5 @@ custom_types:
     - name: ActivityArguments
     - name: ProfileSet
     - name: SchedulingFailureReason
+    - name: Command
+    - name: UserCodeError
```
dyst5422: I think we can probably have a more useful type than a blind scalar here.

```
type UserCodeError {
  message: String!
  stack: String!
  sourceContext: String!
  location: CodeLocation
}

type CodeLocation {
  line: Int!
  column: Int!
}
```

dyst5422 (changes requested):

deployment/hasura/metadata/actions.yaml
pcrosemurgy: From https://hasura.io/docs/latest/graphql/core/actions/index/#kind
>Synchronous: Sync actions return a response to the client after receiving a response from the handler.

From memory, each of our `mutation`s need this.

command-expansion-server/test/batchLoaders/simulatedActivityInstanceBatchLoader.spec.ts
pcrosemurgy: I spent some time trying to get it working but had no luck. I can give it another shot though!

command-expansion-server/src/app.ts
mattdailis: 👍

deployment/hasura/metadata/actions.graphql
mattdailis: :+1:

deployment/hasura/metadata/actions.graphql
```diff
@@ -34,29 +34,29 @@ type Mutation{
 
 type Query {
   getModelEffectiveArguments(
-    missionModelId: ID!
+    missionModelId: Int!
```
mattdailis: `getModelEffectiveArguments` and `getActivityEffectiveArguments` are handled in `merlin-server`. I don't think we should change them as part of this PR.

Just a brief survey of some of the files in `merlin-server` that will need to change:
* HasuraAction
* HasuraParsers
* MissionModelService interface
* LocalMissionModelService
* StubMissionModelService
* MissionModelRepository interface
* PostgresMissionModelRepository
* LocalMissionModelRepository

I think we absolutely should change them, but I think that is well out of scope for this PR.

pcrosemurgy: Yeah, I'm with you Matt. Appreciate the fix, Dylan, but it's probably best to keep this PRs scope in-check

mattdailis: When we have API tests for more of our actions, we will hopefully get notified automatically about issues like this:
Screen Shot 2022-03-31 at 5 54 44 PM

pcrosemurgy: Latest push reverts the `ID` -> `Int` change for `missionModelId`

deployment/hasura/metadata/actions.graphql
pcrosemurgy: I'm curious when `ID` is appropriate in GraphQL? Maybe in situations where the ID does not correspond to a DB (PG) numeric ID? So maybe a UUID would be an appropriate use case?

command-expansion-server/sql/commanding/tables/expansion_set_to_rule.sql
```diff
@@ -1,22 +1,27 @@
...
   constraint expansion_set_to_rule_primary_key
   primary key (set_id,rule_id),
 
   foreign key (set_id)
   references expansion_set (id),
 
-  foreign key (rule_id)
-  references expansion_rule (id)
+  foreign key (rule_id, activity_type)
+  references expansion_rule (id, activity_type),
```
mattdailis: I'm glad this sort of thing worked :+1: though you would think that postgres could figure out that if `id` is unique, (`(id, activity_type)`) must also be unique... 🤷

mattdailis (approved):

deployment/hasura/metadata/actions.graphql
Twisol: In my experience, `ID` is usually the appropriate type. Firstly, clients ought to tread IDs as opaque data anyway, which `Int` doesn't get across; and secondly, our IDs might be single ints now, but that could easily change in the future.

Lastly, `ID` appropriately identifies entities of any type; `Int` might only be applicable to some entities (i.e. those whose ids are specifically integers). GraphQL [encourages](https://graphql.org/learn/global-object-identification/) being able to look up an entity by ID even when you don't know what kind of entity it is; this is called "global object identification".

> Consistent object access enables simple caching and object lookups
>
> To provide options for GraphQL clients to elegantly handle caching and data refetching, GraphQL servers need to expose object identifiers in a standardized way.

dyst5422: The problem is that they deserialize as strings and we need to access the database using ints. If we were using opaque UUID's I would agree with you.

command-expansion-server/src/lib/mapGraphQLActivityInstance.ts
dyst5422: Out of scope for this release

command-expansion-server/sql/commanding/tables/expansion_set_to_rule.sql
dyst5422: Exactly my thoughts.

command-expansion-server/sql/commanding/tables/expansion_rule.sql
```diff
@@ -7,7 +7,7 @@ create table expansion_rule (
   constraint expansion_rule_primary_key
   primary key (id),
 
-  UNIQUE (id, activity_type)
+  UNIQUE (id, activity_type) -- Needed to expose this tuple as a valid foreign key
```
dyst5422: I think we can probably name the constraint here with

```
CONSTRAINT one_activity_type_per_expansion_set UNIQUE (id, activity_type)
```

For a little more clarity when it is violated

dyst5422: Just took care of this and adding a comment on it

deployment/hasura/metadata/actions.graphql
Twisol: If the problem is only that GraphQL delivers strings, why is pushing the ID representation out to clients preferable over using `parseInt(id, 10)` within the server?

> If we were using opaque UUID's I would agree with you.

My argument is that the API shouldn't change depending on what our internal ID representation is. I can appreciate that `ID` requires the backend to explicitly parse the ID to a number, but this doesn't address any of the motivations I gave. The [merlin-server](https://github.com/NASA-AMMOS/aerie/blob/78cb0e4efe12f9e4f989087bbc6030498a606a61/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresMissionModelRepository.java#L199-L207) currently does the same conversion from string to numeric for Postgres.

dyst5422:
🚢 it

dyst5422 (approved):

PR #107, opened 2022-03-16
[AERIE-1384] Multiple simulation workers

patkenneally:
* **Tickets addressed:** AERIE-1384
* **Review:** By commit  

## Description
This PR introduces a multi-tenancy capability for Aerie simulation. Prior to this Aerie was able to execute a single simulation run at a time, with the merlin-server queuing simulation requests. Aerie is a multi-user application and needs to be able to execute multiple concurrent simulation requests. Additionally, it would be advantageous if users are able to deploy Aerie across a number of hosts, with the majority of containers on modest hardware and the merlin-worker containers on high performance hardware tuned for simulation execution speed. Threading concurrent simulation requests would have been a viable first step on the path if not for the SPICE library's thread safety issues. Given these considerations, we've chosen a multiprocessing focused solution.

A small merlin-worker application is added. The worker subsumes the responsibility of running a simulation and writing the results to the merlin postgres database. The merlin-server processes a simulation request, creates the initial dataset tables and associations. The simulation_dataset entry is initially set with a status "pending". This status indicates that the simulation run is queued waiting for a worker to take the job. When a worker takes the job the status is updated (by the worker) to "incomplete".   

The postgres notification feature is used to NOTIFY workers LISTEN(ing) for newly created simulation jobs. The added database trigger calls a postgres function to emit a notification with a payload containing the simulation revision data, plan id, simulation id, and dataset id. Upon receiving a notification a worker will attempt to claim the simulation job by changing it's simulation_dataset status field to "incomplete". A simulation job is considered taken if this SQL operation is successful. Then the successful worker continues with executing the simulation. If the SQL operation is unsuccessful (e.g. another worker was successful in claiming a job earlier) then the worker handles the failure to claim the job and moves on to attempt claiming other jobs. We can be assured that there will be no race conditions due to postgres' ACID transaction processing. 

Of particular note is the relocation of profile, event, and span partition table creation to the worker. It was discovered that a DB deadlock is induced by making multiple concurrent simulation requests. Creating these partition tables at the moment they are needed, and in a separate transactions prevents the recurrence of the observed deadlock.  

Users configure a fixed number of workers by adding additional worker service entries in their docker compose. The containers are set to automatically restart so that workers return even in the case of process ending failures.

## Verification
A test script which created a single Banananation mission model, and six different plans for the mission model. Six concurrent simulation requests were made (one for each plan). A successful simulation run was observed and simulation results returned for each request. Further manual testing was done to edit a single plan and re-simulate. 

The ability to run these kinds of integrated tests via the is something that Aerie needs to setup and start running as a portion of our devops processes. There are a number of tickets capturing this in the backlog. They will be picked up in the coming sprints.

I'm happy to share my test script and data with anyone who wants/needs to exercise this PR in the same way.

## Documentation
- Deployment documentation
- Release 
	- documentation
	- products

## Future work
None


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
```diff
@@ -6,6 +6,9 @@
   private ResultsProtocol() {}
 
   public /*sealed*/ interface State {
```
mattdailis: 🔵 We can uncomment `sealed`, since we're on Java 17 now

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
```diff
@@ -47,7 +47,7 @@ public SimulationDatasetRecord apply(
     if (!results.next()) throw new FailedInsertException("simulation_dataset");
     final var datasetId = results.getLong(1);
     final var state = new SimulationStateRecord(
-        results.getString(2),
+        SimulationStateRecord.SimulationStatus.fromString(results.getString(2)),
```
mattdailis: Just making a note that this will need to be carefully rebased onto #102 , since this change was made independently in both PRs.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ClaimSimulationAction.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public ClaimSimulationAction(final Connection connection) throws SQLException {
+    this.statement = connection.prepareStatement(sql);
+  }
+
+  public void apply(final long datasetId) throws SQLException {
+    this.statement.setLong(1, datasetId);
+
+    final var count = this.statement.executeUpdate();
+    if (count != 1) throw new SQLException("Failed to claim simulation.");
```
mattdailis: 🔵 Idle thought - since this is expected, rather than exceptional behavior, I wonder if it would make sense to return an `Optional<Long>` or something to indicate that the claim action failed. I have no strong reason to go either way.

If we do decide that an exception is the preferred mechanism to communicate a failed claim, my next thought is whether we want a specific exception for this scenario 🤔

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public ListenSimulationCapability(
+      final DataSource dataSource,
+      final BlockingQueue<PostgresSimulationNotificationPayload> notificationQueue) {
+    this.dataSource = dataSource;
+    this.notificationQueue = notificationQueue;
+  }
+
+  public void registerListener() throws SQLException {
+    PGConnection connection = this.dataSource.getConnection().unwrap(PGConnection.class);
```
mattdailis, line 33: 🔶 I'm intrigued by this unwrap - I wonder if it's possible to avoid it. In particular it seems like `DataSource` could be more specific... I'll let you know if I find an alternate way of expressing this.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
```diff
@@ -0,0 +1,60 @@
...
+    public void notification(int processId, String channelName, String payload){
+      System.out.println("Received PSQL Notification: " + processId + ", " + channelName + ", " + payload);
+      final var jsonValue = Json.createReader(new StringReader(payload)).readValue();
+      final var notificationPayload = postgresSimulationNotificationP
+          .parse(jsonValue)
+          .getSuccessOrThrow();
+      try {
+        notificationQueue.put(notificationPayload);
+      } catch (InterruptedException e) {
+        e.printStackTrace();
```
mattdailis: 🔻 Flagging this catch block - we should either rethrow some kind of an exception, or write a comment explaining why we're suppressing it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
```diff
@@ -45,7 +45,8 @@ public static void main(final String[] args) {
     final var simulationAgent = ThreadedSimulationAgent.spawn(
         "simulation-agent",
         new SynchronousSimulationAgent(planController, missionModelController));
-    final var simulationController = new CachedSimulationService(stores.results(), simulationAgent);
+    final CachedSimulationService simulationController = new CachedSimulationService(stores.results(), simulationAgent);
```
mattdailis: 🔶 This change doesn't seem to do anything besides make the variable type explicit - we can probably revert it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateEventPartitionAction.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.intellij.lang.annotations.Language;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/*package-local*/ final class CreateEventPartitionAction implements AutoCloseable {
+  private final Connection connection;
+
+  public CreateEventPartitionAction(final Connection connection) throws SQLException {
```
mattdailis, line 11: 🔵 Just for context - I remember the original intent of doing these in triggers was to encapsulate partition management in the database itself. That said, deadlocks are no fun, and I'm certainly more comfortable with Java code than PL/pgSQL, so I'm okay with handling this in Java

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -70,6 +70,32 @@ public PostgresResultsCellRepository(final DataSource dataSource) {
...
+
+      return Optional.of(new PostgresResultsCell(
+          this.dataSource,
+          simulation,
+          datasetId,
+          planStart));
+    } catch (final NoSuchPlanException ex) {
+      throw new Error(String.format("Cannot process simulation for nonexistent plan %s.%n", planId), ex);
+    } catch(final SQLException | DatabaseException ex) {
+      return Optional.empty();
```
mattdailis: 🔶 I'm wondering if we're suppressing any potentially useful exceptions here. We only expect `SQLException` if we fail to claim, right? A `DatabaseException` would indicate a real error that we shouldn't suppress?

EDIT: I just read further down the file and saw that we're converting `SQLExceptions` to `DatabaseExceptions`, in which case I'm still confused why we're catching both in this catch block.

Not requesting any specific change at the moment, just expressing my confusion 👍

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -170,6 +196,31 @@ private static SimulationDatasetRecord createSimulationDataset(
...
+      final var createSimulationDatasetPartitionsAction = new CreateProfileSegmentPartitionAction(connection);
+      final var createSpanPartitionAction = new CreateSpanPartitionAction(connection);
+      final var createEventPartitionAction = new CreateEventPartitionAction(connection);
+
+      createSimulationDatasetPartitionsAction.apply(datasetId);
+      createSpanPartitionAction.apply(datasetId);
+      createEventPartitionAction.apply(datasetId);
+      transactionContext.commit();
+    } catch (final SQLException ex) {
+      throw new DatabaseException(String.format("Failed to create partitions for simulation dataset id %s", datasetId), ex);
```
mattdailis, line 247: 🔶 Continuing from previous comments - it seems like "failed to create partitions" will be lumped together with "failed to claim a dataset" since both throw indistinguishable `DatabaseException`s.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
```diff
@@ -62,10 +62,25 @@ public static void main(String[] args) throws Exception {
...
 
-      System.out.println(String.format("Received simulation notification for plan Id: %s and data set Id: %s",
-                                       planId,
-                                       datasetId));
+      final Optional<ResultsProtocol.OwnerRole> owner = stores.results().claim(planId, datasetId);
+      if (owner.isEmpty()) continue;
 
+      final var reader = stores.results().lookup(planId);
+      // There is no simulation for this plan
+      if (reader.isEmpty()) continue;
```
mattdailis: 🔵 Is it okay for there to be no simulation for a plan that does have an entry in `simulation_dataset`? We may be protected from this by some foreign key relationships 🤔 

By "is it okay" I'm wondering if we need to throw some kind of exception... at the same time, it probably doesn't make sense for a random worker to crash... I'm not sure what the best answer is here.

docker-compose.yml
```diff
@@ -108,6 +108,49 @@ services:
...
+    container_name: aerie_merlin_worker_2
+    depends_on: ["postgres"]
+    environment:
+      MERLIN_WORKER_DB: "aerie_merlin"
+      MERLIN_WORKER_DB_PASSWORD: "aerie"
+      MERLIN_WORKER_DB_PORT: 5432
+      MERLIN_WORKER_DB_SERVER: "postgres"
+      MERLIN_WORKER_DB_USER: "aerie"
+      MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      MERLIN_WORKER_PORT: 27186
```
mattdailis: 🔵 I'm curious about this port - what is it used for? Should we expose it?

mattdailis (commented):
Generally looks really good, and I'm super stoked for this new functionality. Most of my comments are me trying to understand our error handling story.

Key:

docker-compose.yml
patkenneally: Good catch. A rogue copy and paste.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateEventPartitionAction.java
patkenneally: Is your implication here that we could be triggering the partition creation on the change of status from "pending" to "incomplete" and thus it could remain a triggered function within our schema definitions?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
patkenneally: Thanks for that catch. Failed to fully clean up after an experiment I backed out of :D

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
patkenneally: Agreed. I expect 102 in sooner so will rebase.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateEventPartitionAction.java
mattdailis: I hadn't thought it that far through - but yes, now that you mention it,  I think that would work. I haven't worked with partitions before, so I'm not really sure what to advise here. Maybe we can consult @kroffo or @Twisol or Tung.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
mattdailis: From a cursory look at the types, it seems like your approach is the expected approach 👍  I'll resolve this comment now.

patkenneally: Bummer. I was hoping you would find something I hadn't found :) . I agree, it feels dirty/wrong.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
patkenneally: Given the discussion with @kroffo and @mattdailis I've adjusted both the `ClaimSimulationAction` and the `PostgresResultsCellrepository` which uses it. The action now only returns an SQLException when an unexpected number of result rows are returned by the db update. An `UnclaimableSimulationException` has been added, although no action is taken because of it so I'm unsure if this is the best remedy here. Still pondering...

mattdailis: I think this solution adequately addresses @kroffo 's concerns about using the `SQLException` to mean two different things :+1:

I would maybe request a javadoc comment on `claimSimulationDataset` and/or `ClaimSimulationAction.apply` explaining what the `UnclaimableSimulationException` means.

mattdailis: Some thoughts that went through my head were that if this method were returning some data that the caller needed, we could return an Optional and feel reasonably certain that the caller will use that information.

In this case, the whole point of `claim` is its side-effect - the caller just needs to check whether it succeeded or not. Declaring the checked exception as part of the method signature forces the caller to handle the failure case.

We could potentially contort the data structures to make `claim` return some data that the caller needed (e.g. caller only has `datasetId`, in order to get the rest of the information they call `claim`. Thus if `claim` returns an `Optional.empty`, the caller cannot erroneously proceed and try to run the simulation), but I'm not sure I like that either.

If our external advisory board of @Twisol or @kroffo (or anyone else who has opinions on when to use exceptions vs when to return a success/failure value) wants to weigh in, feel free :+1:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
patkenneally: Done

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
patkenneally: Have resolved this by the resetting the interruption. Reading in "[Java Concurrency In Practice](https://jcip.net/)", they recommend resetting the interrupt when you are in a context that prevent you from taking action. We have no ability to take action within this listener callback.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
patkenneally: I've removed this block of code as it was not needed. It was left over from earlier prototyping efforts.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
patkenneally: Good idea to add some Javadocs. I did so.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateEventPartitionAction.java
Twisol: If you just want the logic centralized in the database, you can always [`CALL`](https://www.postgresql.org/docs/14/sql-call.html) a stored procedure that does the partitioning.

Whether or not this stored procedure is invoked automatically on some change in the database (as a trigger), or manually by the Java client, is an orthogonal concern. 😄

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
Twisol: I haven't looked at the code (GitHub is being very unhelpful; it says "these commits no longer exist"), but not being able to claim a simulation just means the worker should go back to the top and wait for another simulation. It probably couldn't claim because another worker got there first.

Based on what Matt described just above, I would argue for an `Optional<>` based approach, where an object permitting write access to the simulation is provided only if the claim was successful. I'm not sure that an exception is really suitable for a case that is not only expected, but will likely happen quite often (N-1 times simultaneously for each simulation, where N is the number of idle workers 😆)

mattdailis: At the ResultsCellRepository interface level, that's exactly how it works (I think): 

https://github.com/NASA-AMMOS/aerie/blob/11751aaa624f97fa1fbab5f56a91ee694f12d0b8/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/ResultsCellRepository.java#L11

The implementation of the `PostgresResultCellRepository#claim` converts an `UnclaimableSimulationException` into an Optional:

https://github.com/NASA-AMMOS/aerie/blob/11751aaa624f97fa1fbab5f56a91ee694f12d0b8/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java#L90-L115

I think I was musing about the `ClaimSimulationAction`, which is a `void` method that throws an `UnclaimableSimulationException`:

https://github.com/NASA-AMMOS/aerie/blob/11751aaa624f97fa1fbab5f56a91ee694f12d0b8/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ClaimSimulationAction.java#L24

I guess I'm unsure whether it's appropriate for the `UnclaimableSimulationException` to exist. My current stance is that since its local to the repository, it's fine to throw an exception - but I don't feel like I have a good sense of when it makes sense to use an exception, and when a "failure" should be part of the return type.

mattdailis (approved):

adrienmaillard (approved):

PR #109, opened 2022-03-17
[AERIE-1749] command expansion logging

JoelCourtney:
* **Tickets addressed:** AERIE-1749
* **Review:** By commit  

## Description
Converts all uses of `console.log` and `console.error` to use Winston loggers. The first commit creates the logger factory, and the second commit factors everything to use it.

The goal is to make the logging format match the SLF4J format used by the merlin and scheduler servers, so each file that uses a logger is required to get its own instance and configure it with the name of the file. The format is:

```
LEVEL filename - message
```

The logger also automatically prints stack traces when it receives an error object instead of a string message.

The logger can be configured with the `LOG_LEVEL` and `LOG_FILE` environment variables. `LOG_FILE` has a special value of `console` (which is also the default) which sends output to (you guessed it) the console. `LOG_LEVEL` accepts the levels recognized by Winston, which are:

```ts
const levels = {
  error: 0,
  warn: 1, // (eventually) deployment default
  info: 2, // winston default if unset
  http: 3,
  verbose: 4,
  debug: 5, // dev default
  silly: 6
};
```

But aerie-commanding so far only uses `info` and `error`.

I copied logger.ts from aerie-gateway because I figured finding a way to not repeat myself would be way more trouble than its worth here.

## Verification
- With `LOG_FILE: console` and `LOG_LEVEL: debug`, I ran `docker compose build && docker compose up` - Info log messages were printed with the correct format.
- Then set `LOG_LEVEL: error` and rerun - info logs are silenced.
- Then temporarily change one of the info messages to error - log is printed again.
- Reset `LOG_LEVEL` and set `LOG_FILE: log.txt` - log is printed to `log.txt`.
- Reset everything and put `logger.error(new SyntaxError("hello world"))` somewhere in the startup code - out pops a stack trace.

While doing this it is helpful to set the log levels in other containers to `error`.

## Documentation
Documentation of the environment variables has not been updated because aerie-commanding does not have a section in `deployment/Environment.md`.

## Future Work

When aerie-commanding is released, `deployment/Environment.md` should include the `LOG_FILE` and `LOG_LEVEL` variables, and they should be added to `deployment/docker-compose.yml`. (default to `console` and `warn` respectively.)

### DIFF
--- a/command-expansion-server/src/packages/db/db.ts
+++ b/command-expansion-server/src/packages/db/db.ts
@@ -1,6 +1,7 @@
 import type { Pool, PoolConfig } from "pg";
 import pg from "pg";
 import { getEnv } from "../../env.js";
+import getLogger from "../../utils/logger.js";
 
 const { Pool: DbPool } = pg;
 
@@ -11,6 +12,8 @@ const {
   POSTGRES_PORT: port,
   POSTGRES_USER: user,
 } = getEnv();
+
+const logger = getLogger("packages/db/db");
 
 export class DbExpansion {
   private static pool: Pool;
@@ -29,12 +32,12 @@ export class DbExpansion {
         user,
       };
 
-      console.log(`Postgres Config:`);
-      console.log(config);
+      logger.info(`Postgres Config:`);
+      logger.info(config);
 
       DbExpansion.pool = new DbPool(config);
     } catch (error) {
-      console.log(error);
+      logger.error(error);
     }
   }
 }
--- a/command-expansion-server/package-lock.json
+++ b/command-expansion-server/package-lock.json
@@ -15,7 +15,8 @@
         "graphql-request": "^4.0.0",
         "pg": "^8.7.3",
         "reserved-words": "^0.1.2",
-        "typescript": "^4.5.5"
+        "typescript": "^4.5.5",
+        "winston": "^3.6.0"
       },
       "devDependencies": {
         "@microsoft/tsdoc": "^0.13.2",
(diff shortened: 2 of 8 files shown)


camargo (approved):

mattdailis:
> I manually verified that message and errors are logged in the correct format, and that the level and file can be set using the environment variables.

Would you mind fleshing out the `Verification` section a little bit? I think some of @goetzrrGit 's PRs are a good example of this, where he gives enough information for the reviewer to be able to reproduce the steps themselves if they so desire (e.g. #70)

JoelCourtney:
@mattdailis done

mattdailis:
> @mattdailis done

Thank you :+1: Reproducing now.

command-expansion-server/src/packages/db/db.ts
```diff
@@ -29,12 +32,12 @@ export class DbExpansion {
         user,
       };
 
-      console.log(`Postgres Config:`);
-      console.log(config);
+      logger.info(`Postgres Config:`);
+      logger.info(config);
```
mattdailis, line 36: CodeQL seems to be mad at this line for leaking sensitive information, but when I look at it in `docker-compose logs -f aerie_commanding`, it seems to be printing `[Object object]`:

```
aerie_commanding    | INFO packages/db/db - Postgres Config:
aerie_commanding    | INFO packages/db/db - [object Object]
```

So... I don't think we're leaking sensitive information, but I don't think we're logging anything useful either.

mattdailis (approved):
Looks good to me :+1:

I left a comment about how the config gets logged as `[Object object]`. I don't know what the desirable behavior is here, and I'm not sure whether figuring that out should be part of this PR, but I just want to bring attention to that.

command-expansion-server/src/packages/db/db.ts
JoelCourtney: Thanks, I was just about to ask about that. Looks like CodeQL was giving that alert on the old `console.log` too.

Git says @goetzrrGit wrote that code. Ryan, is there any reason I can't delete it?

JoelCourtney: Well, this can be done later in a little hotfix if needed. Resolving this because its not strictly part of this PR.

PR #110, opened 2022-03-18
[HOTFIX] Check activity DurationType before setting its duration parameter

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

When combining #71  with #54 , we didn't re-implement the checks for an activity's DurationType. Thus, we were attempting to instantiate activities with an extraneous duration parameter.

The new simulation driver loops until the activity completes. If an activity failed to be instantiated, rather than throw an exception, it will be placed in the `IllegalSource` state. Thus, it will never reach the `Terminated` state, so the simulation driver will loop forever. This is the behavior that @camargo observed.

With @adrienmaillard 's help, we identified this issue and implemented the fix - prior to adding a duration parameter, the driver will check that an activity's `DurationType` is `Controllable`.

In addition to implementing the fix, we added a method called `initiateTaskOrFail` to the simulation engine, whose job it is to propagate the `UnconstructableTaskException` to the caller.

## Verification

We reproduced Chris's issue by running the api-tests:
```
git checkout prototype/api-tests settings.gradle api-tests
./gradlew :api-tests:run
```
We observed that the tests were timing out.

We then made our fix, and re-ran the tests, and saw that they did not time out.

We also ran the tests without the fix, but with our additional exception logging, and saw in the logs that the errors were reported correctly.

## Documentation

No scheduler docs... yet

## Future work

- @adrienmaillard and I discussed that the scheduler lacks a unified error handling strategy. We should decide on a strategy, standardize it across the scheduler-server and the scheduler, and figure out how to display those errors to users. Adrien proposed some sort of console pane in the UI - we would need to do some UX design on this topic.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -72,21 +72,27 @@ public final class SimulationEngine implements AutoCloseable {
   /** Construct a task defined by the behavior of a model given a type and arguments. */
   public <Model>
   TaskId initiateTaskFromInput(final MissionModel<Model> model, final SerializedActivity input) {
-    final var task = TaskId.generate();
-
-    final Directive<Model, ?, ?> directive;
     try {
-      directive = model.instantiateDirective(input);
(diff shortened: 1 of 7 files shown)


adrienmaillard (approved):

camargo (approved):
Tested this against the front-end and it is no longer hanging

PR #111, opened 2022-03-18
[AERIE-1557] Separate Harusa action parsers from Merlin Parsers

JoelCourtney:
* **Tickets addressed:** AERIE-1557
* **Review:** By commit  

## Description
Separates Hasura action parsers from `MerlinParsers.java` in merlin-server into `HasuraParsers.java`. This has no behavioral change.

## Verification
Since there is no behavioral change, only moving things into another class, running `./gradlew build` should be sufficient. (It passes.)

## Documentation
There does not appear to be any documentation of this, so none needs to be changed.

## Future work
None


mattdailis (approved):
Looks good to me :+1:

kroffo (approved):
Since the Hasura Actions were moved, I did spin up the system to verify that operations were working as expected. I uploaded a model through the UI then created a plan and added an external dataset through the Hasura console as a test, and everything worked as expected.

PR #112, opened 2022-03-21
[AERIE-1676] Look up ActivityTypes from Problem when building Goals

mattdailis:
* **Tickets addressed:** AERIE-1676
* **Review:** By commit  

## Description

If the scheduler is run on a plan that already satisfies its goals, it should not add any new activities to that plan. One of the ways the scheduler determines that an activity satisfies its goal is by checking its activity type. This activity type check uses object equality - so the corresponding `ActivityType` field needs to be identical in order for an `ActivityExpression` to match an `ActivityInstance`.

This PR updates the GoalBuilder to call `problem.getActivityType` with the name of the activity type, rather than construct a `new ActivityType()` itself.

In order to make this change, the GoalBuilder had to be moved later in the workflow, after the `Problem` object is constructed, but before its `setGoals` method is called.

After making this change, I would get `IllegalStateExceptions` because `Evaluation#canAssociateMoreToCreatorOf` assumes that every activity instance was created by a scheduling goal. This assumption is not, in general, true, because there are already activities in the plan before the scheduler is run. I updated `Evaluation#canAssociateMoreToCreatorOf` to return `true` if an activity was not created by any goal.

## Verification

1. Fresh `./gradlew assemble` and `docker-compose build` and `docker-compose up -d`
2. Make an empty plan
3. Add the following scheduling goal:
```typescript
export default function myGoal() {
        activityTemplate: ActivityTemplates.BakeBananaBread('My Goal', { temperature: 325.0, tbSugar: 2, glutenFree: false }),
        interval: 12 * 60 * 60 * 1000 * 1000 // 1 day in microseconds
    })
}
```
4. Run the scheduling goal, observe that it places activities in the plan
5. Run the scheduling goal again, observe that it does NOT place any more activities in the plan
6. Delete a few of the activities it placed from the plan
7. Run the scheduler again and observe that it re-creates the activities you deleted :tada:

https://user-images.githubusercontent.com/1189602/159300691-2a412554-b13f-4f22-acff-6a9b2dce7226.mov

## Documentation

In the soon-to-be-written scheduling documentation, it may be worth describing how the scheduler identifies activities in the plan that already satisfy its goals.

## Future work

- Implement more Goal types and test with more activity types

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
@@ -156,22 +156,26 @@ public class Evaluation {
     return Objects.hash(goalEvals);
   }
 
-  boolean canAssociateMoreToCreatorOf(ActivityInstance instance){
-    Goal creator = getGoalCreator(instance);
-    if(creator instanceof ActivityExistentialGoal activityExistentialCreator) {
+  boolean canAssociateMoreToCreatorOf(final ActivityInstance instance){
+    final var creator = getGoalCreator(instance)
+        .orElseThrow(() -> new IllegalStateException("No goal is referenced as creator for activity " + instance));
+    if (creator instanceof ActivityExistentialGoal activityExistentialCreator) {
       //we can piggyback
       return activityExistentialCreator.childCustody == ChildCustody.Jointly;
     }
     return true;
   }
 
-  Goal getGoalCreator(ActivityInstance instance){
-    for(var goalEval : goalEvals.entrySet()){
+  /**
+   * If an activity instance was already in the plan prior to this run of the scheduler, this method will return Optional.empty()
+   */
+  Optional<Goal> getGoalCreator(final ActivityInstance instance){
+    for(final var goalEval : goalEvals.entrySet()){
       if(goalEval.getValue().getInsertedActivities().contains(instance)){
-        return goalEval.getKey();
+        return Optional.of(goalEval.getKey());
       }
     }
-    throw new IllegalStateException("No goal is referenced as creator for activity " + instance) ;
+    return Optional.empty();
   }
 
 }
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
@@ -157,13 +157,12 @@ public class Evaluation {
   }
 
   boolean canAssociateMoreToCreatorOf(final ActivityInstance instance){
-    final var creator = getGoalCreator(instance)
-        .orElseThrow(() -> new IllegalStateException("No goal is referenced as creator for activity " + instance));
-    if (creator instanceof ActivityExistentialGoal activityExistentialCreator) {
-      //we can piggyback
-      return activityExistentialCreator.childCustody == ChildCustody.Jointly;
-    }
-    return true;
+    final var creator$ = getGoalCreator(instance);
+    if (creator$.isEmpty()) return false;
+    final var creator = creator$.get();
+
+    if (!(creator instanceof ActivityExistentialGoal activityExistentialCreator)) return true;
+    return activityExistentialCreator.childCustody == ChildCustody.Jointly; // we can piggyback
   }
 
   /**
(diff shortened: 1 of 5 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
```diff
@@ -156,22 +156,25 @@ public int hashCode() {
...
+    if (!(creator instanceof ActivityExistentialGoal activityExistentialCreator)) return true;
+    return activityExistentialCreator.childCustody == ChildCustody.Jointly; // we can piggyback
   }
 
-  Goal getGoalCreator(ActivityInstance instance){
-    for(var goalEval : goalEvals.entrySet()){
+  /**
+   * If an activity instance was already in the plan prior to this run of the scheduler, this method will return Optional.empty()
+   */
+  Optional<Goal> getGoalCreator(final ActivityInstance instance){
```
adrienmaillard, line 171: Good catch on the missing use case. What you implemented here is a default behavior saying we can piggyback any user-inserted activity. Which seems fine for now but I think this illustrates a missing piece in the server. 
 
When the scheduler gets the plan, it should fetch who created the activity : a goal or a user. And it should get the custody property associated to the activity. If created with a goal, we get the custody from the goal definition. If created by the user, we should get it from the user somehow. Then, all of this is passed to a construction of the Evaluation in the solver.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
```diff
@@ -98,44 +98,44 @@ protected RecurrenceGoal fill(RecurrenceGoal goal) {
...
    */
-  public java.util.Collection<Conflict> getConflicts(Plan plan) {
+  @Override
+  public java.util.Collection<Conflict> getConflicts(final Plan plan) {
     final var conflicts = new java.util.LinkedList<Conflict>();
 
     //collect all matching target acts ordered by start time
     //REVIEW: could collapse with prior template start time query too?
-    final var satisfyingActSearch = new ActivityExpression.Builder()
+    final var matchingActivities = new java.util.LinkedList<>(plan.find(new ActivityExpression.Builder()
```
adrienmaillard: Have you encountered issues with this goal to start a refactoring? In PR #113, its behavior is also refactored.

adrienmaillard (approved):
This looks good to me and makes the scheduling functionality much more presentable to users 👍

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
mattdailis: Nope, no issues, I can drop the refactoring commit to reduce conflicts with your PR 👍

kroffo (approved):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
mattdailis: That's a good point... I think I need to better understand this "Custody" concept and some of the motivations behind it. In my mind, the user should be able to manually create activities to satisfy any goal they wish... but perhaps I'm under-thinking things here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
mattdailis: I dropped that commit, so this PR no longer touches `RecurrenceGoal.java` :+1:

pcrosemurgy:
>`interval: 12 * 60 * 60 * 1000 * 1000 // 1 day in microseconds`

Not sure if this really matters but shouldn't 1 day in microsec. be `86400 * 1e6`?

pcrosemurgy:
This has probably been brought up before but `scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/` could really use some package structuring. Right now there are 113 Java source files without any hierarchy.

mattdailis:
> > `interval: 12 * 60 * 60 * 1000 * 1000 // 1 day in microseconds`
> 
> Not sure if this really matters but shouldn't 1 day in microsec. be `86400 * 1e6`?

Good catch - it originally was a day, and then when I was testing I changed it to 12 hours, but forgot to update the comment.

> This has probably been brought up before but `scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/` could really use some package structuring. Right now there are 113 Java source files without any hierarchy.

@adrienmaillard has brought this up before, yes. Definitely. 100%.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Evaluation.java
adrienmaillard: Seems like it is outside the scope of this PR.

PR #113, opened 2022-03-22
[AERIE-1731] Schedule activities with uncontrollable durations

adrienmaillard:
* **Tickets addressed:** AERIE-1648
* **Review:** By commit  

## Description

Commits 1 2 and 3 fix some minor issues.

Commit 4 adds several root-finding algorithms.  One interface does not fit them all, unfortunately. One concern I have is the conversion between `Duration` and `Double`. 

Commit 5 adds a configurable timing error feature to `ActivityCreationTemplate`. It is necessary for the root-finding algorithms to work/to converge in a reasonable time to allow for a timing error. The user can set an error and if he does not, the default error has been set to 0.5sec around the desired value (1sec in total).  

Commits 6 fixes the behaviour of `RecurrenceGoal` that was not creating the appropriate number of conflicts.

Commit 7 completely changes and simplify the solving dynamics of the `CardinalityGoal`. Instead of finely trying to instantiate templates and activities beforehand, it assumes almost nothing about durations or constraints and lets the solver handle it. To schedule activities for a certain duration, it produces only ONE conflict at a time if there are not enough activities in the current plan to satisfy the duration constraint.  The solver will then insert one activity with some unpredictable duration (depending on duration type or state or goal constraints...) and the goal will be reevaluated (normal behaviour). It will then see whether there is now enough activities in the plan and otherwise create another template conflict. And so on and so on until enough duration has been created. For occurrence constraints, it is simple, we create the appropriate number of template conflicts as before. 

Commit 8 makes `CoexistenceGoal` return template conflicts instead of instance conflicts. It allows to let the scheduler handle activity creation (with root-finding). It also let the solver handle state, and goal constraints as it should be as a goal does not have access to the complete problem definition. Instantiating activities as late as possible is good practice here to improve satisfiability as the solver has a better overview of the problem.  

Commit 9 changes the signature of activity instantiation to `Optional` as it will soon be able to fail and make that clear. 

Commit 10 adds a planning horizon constraint to the `STN` as planning horizon is now known to activity creation.

Commit 11 make activity creation use root-finding algorithms and simulation to instantiate activities with uncontrollable duration. It uses the secant method as it does not have bracketing requirements and worked with the provided examples. Finding a bracket a priori might not be a trivial problem in our case and require multiple simulations before even starting root-finding per see.

Commit 12 adds two tests. One of the new activities contains a parameter-dependent non-linear duration model coming from Gavin and is inspired by solar panel articulation.  The second is derived from the first and adds a time-dependent component, one of the parameters is now a command to achieve and the corresponding move depends on the time. This time-dependency is modelled directly inside the activity model. In practice, this second activity usually requires several iterations of root-finding before converging. 

Commit 13 adds asserts to aerielander tests for better failure detection when using them. Took advantage of the need to handle time points/durations to replace time scheduler-specific time libraries with standard java Instant/Duration libraries. 

Commit 14 replaced all built-in `assert` into junit5 `assertEquals` `assertTrue`(...) in all tests.

## Verification

All tests are passing and some have been added. 

## Documentation

There is no documentation for the scheduler. Reviewers should especially pay attention to root-finding algorithms and uses.

## Future work

- Change which root-finding algorithm we use?
- Root finding algorithm is quite simple right now and nothing is done if it fails. We could at least do a simple meta-heuristic : (1) use secant method (as it is fast). During computation it saves a `History` with (xi, f(xi)) explored tuples. (2) if (1) fails, and `History` contains two values x1 x2 such that f(x1) and f(x2) are of opposite signs, use Bisection or Brent. We are then sure to converge.


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+      if (fx_m * fx_p > 0) {
+        throw new WrongBracketingException("f(x0)=" + fx_m + " and f(x1)=" + fx_p + " are of the same sign");
+      }
+      for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
+        final double xg = (bracket1 + bracket2) / 2.;
+        final var fxg = ff.valueAt(xg);
+        if (fxg >= confidenceYLow && fxg <= confidenceYHigh) {
+          return new RootFindingResult<>(xg, fxg,null);
+        } else {
+          if (fxg > 0 && bracket1 > 0 || fxg < 0 && bracket1 < 0) {
```
mattdailis: It seems to me that the sign of `bracket1` is not what we care about - we care about the value of `f(bracket1)`, right?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
```diff
@@ -242,21 +240,85 @@ private ActivityInstance createInstanceForReal(final String name, final Window w
...
-      act.setDuration(solved.duration().start);
-    } else {
-      final var computedDur = this.parametricDur.compute(Window.between(earliestStart, earliestStart));
-      if (solved.duration().contains(computedDur)) {
-        act.setDuration(computedDur);
+
+    //the domain of user/scheduling temporal constraints have been reduced with the STN,
+    //now it is time to find an assignment compatible
+    //CASE 1: activity has an uncontrollable duration
+    if(Objects.equals(this.type.getDurationType(), DurationType.uncontrollable())){
```
mattdailis: Object equality _might_ work (because records are magic?), but for sealed interfaces, we prefer `instanceof`

```suggestion
    if (this.type.getDurationType() instanceof DurationType.Uncontrollable) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+    /**
+     * Solves f(x) = y for x in [dxLow, dxHigh] with y in [yLow, yHigh].
+     * the sign of f(bracket1) must be different from the sign of f(bracket2)
+     */
+    RootFindingResult<T> findRoot(
+        Function<T> f,
+        T bracket1,
+        T bracket2,
+        T y,
+        T confidenceYLow,
```
mattdailis: Per @patkenneally 's suggestion, maybe we can rename `confidence` to `tolerance`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+
+      double fx_m = ff.valueAt(bracket1);
+      double fx_p = ff.valueAt(bracket2);
+      if (fx_m * fx_p > 0) {
+        throw new WrongBracketingException("f(x0)=" + fx_m + " and f(x1)=" + fx_p + " are of the same sign");
+      }
+      for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
+        final double xg = (bracket1 + bracket2) / 2.;
+        final var fxg = ff.valueAt(xg);
+        if (fxg >= confidenceYLow && fxg <= confidenceYHigh) {
```
mattdailis: Should this be minus `confidenceYLow`?:
```suggestion
        if (fxg >= -confidenceYLow && fxg <= confidenceYHigh) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+public class EquationSolvingAlgorithms {
+
+  private static final Logger logger = LoggerFactory.getLogger(EquationSolvingAlgorithms.class);
+
+  public record History<T> (LinkedHashMap<T, T> history){}
+  public record RootFindingResult<T>(T x, T fx, History<T> history){}
+
+  public interface BracketingAlgorithm<T>{
+    /**
+     * Solves f(x) = y for x in [dxLow, dxHigh] with y in [yLow, yHigh].
```
mattdailis: This comment seems to imply that `[yLow, yHigh]` is an interval that contains `y`, but the implementation seems to suggest that the interval is really `[y - yLow, y + yHigh]`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+        Double dxLow,
+        Double dxHigh,
+        int maxNbIterations)
+    throws ExceededMaxIterationException, WrongBracketingException, NoSolutionException
+    {
+      if (bracket1 < dxLow || bracket2 > dxHigh) {
+        throw new WrongBracketingException("brackets are out of prescribed root domain");
+      }
+      if (Math.abs(confidenceYLow) - Math.abs(confidenceYHigh) < 1E-6) {
+        throw new NoSolutionException("To use Brent's method, you must provide equal lower and upper confidence bounds.");
```
mattdailis: Judging from the text of the exception, I would expect the subtraction to be inside the `abs` call, and the comparison to be "greater than" instead of less than.

```suggestion
      if (Math.abs(confidenceYLow - confidenceYHigh) > 1E-6) {
        throw new NoSolutionException("To use Brent's method, you must provide equal lower and upper confidence bounds.");
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+      var bracket2Long = bracket2.in(Duration.MICROSECONDS);
+      var fx_m = ff.valueAt(bracket1);
+      var fx_p = ff.valueAt(bracket2);
+      if (fx_m.in(Duration.MICROSECONDS) * fx_p.in(Duration.MICROSECONDS) > 0) {
+        throw new WrongBracketingException("f(bracket1) and f(bracket2) are of the same sign");
+      }
+      for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
+        long xg = Math.round(((float) bracket1Long + bracket2Long) / 2);
+        final var fxg = ff.valueAt(Duration.of(xg, Duration.MICROSECONDS));
+        if (fxg.noShorterThan(confidenceYLow) && fxg.noLongerThan(confidenceYHigh)) {
```
mattdailis: Similar question as above - is `yLow` expected to be a positive, or a negative number? If it's positive, I would expect this to be something like this:
```suggestion
        if (fxg.noShorterThan(Duration.negate(confidenceYLow)) && fxg.noLongerThan(confidenceYHigh)) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,417 @@
...
+      final var history = new LinkedHashMap<Duration, Duration>();
+      if (x0.isEqualTo(x1)) {
+        //particular : only one start possible, derivative would be 0
+        final var val = f.valueAt(x0);
+        if (val.noShorterThan(intervalLow) && val.noLongerThan(intervalHigh)) {
+          return new RootFindingResult<>(x0, val,null);
+        }
+        throw new NoSolutionException();
+      }
+      double xn_m1_L = x0.in(Duration.MICROSECONDS);
```
mattdailis: I think I understand that `xn_m1` means "x<sub>n-1</sub>", but I'm not sure what the `_L` means... "Long"? But it's saved into a variable of type "double" 😕

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: You're right, we now compute the value of the bracket at each iteration and compare to it.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityCreationTemplate.java
adrienmaillard: Done.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Yes, it has been done.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Good catch. I also harmonized how theses bounds are treated in the class.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Yes, it has been updated.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Absolutely. It has been changed.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: This was unsure and lead to this error. They should be positive.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: It is stored in a double value to keep better precision when doing operations on it. I am open to suggestions.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,460 @@
...
+  public record History<T> (LinkedHashMap<T, T> history){}
+  public record RootFindingResult<T>(T x, T fx, History<T> history){}
+
+  public interface BracketingAlgorithm<T>{
+    /**
+     * Solves f(x) in [y - toleranceYLow, y + toleranceYHigh] for x in [dxLow, dxHigh].
+     * the sign of f(bracket1) must be different from the sign of f(bracket2)
+     *
+     * dxLow < dxHigh
+     * toleranceYLow > 0, toleranceYHigh > 0, toleranceYLow < toleranceYHigh
```
mattdailis:
```suggestion
     * toleranceYLow > 0, toleranceYHigh > 0
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRulesTest.java
```diff
@@ -51,11 +52,49 @@ public void schedule(){
   public void firstRule() {
     smallProblem.setGoals(new ArrayList<>(rules.getFirstRuleGoals().values()));
     schedule();
+    var time = planningHorizon.getStartAerie().plus(Duration.MINUTE);
+    assert(TestUtility.activityStartingAtTime(plan, time, rules.getActivityType("SSAMonitoring")));
+    assert(TestUtility.activityStartingAtTime(plan, time, rules.getActivityType("HP3TemP")));
+    for(var t = Duration.HOUR; t.shorterThan(planningHorizon.getEndAerie()); t = t.plus(Duration.HOUR)){
+      assert(TestUtility.activityStartingAtTime(plan, time, rules.getActivityType("HP3TemP")));
```
mattdailis: I think we want to use JUnit's `assertTrue` rather than the built-in Java assert. That way we get nice test failure messages. You can get that by adding this import statement:

```
import static org.junit.jupiter.api.Assertions.*;
```

and then changing all the `assert`s in this file to `assertTrue`

mattdailis (changes requested):

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRulesTest.java
Twisol: (but also, `assert` does nothing by default unless you run the JVM with the `-ea` flag)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
Twisol: What *does* the `_L` mean, though? 👀 This is a rather cryptic variable name!

adrienmaillard: _L indeed means "Long" even though it is stored in a double... I can change it, what do you think would be better names for these variables?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: done

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/MerlInsightRulesTest.java
adrienmaillard: I have replaced all `assert` by `assertTrue` or `assertEquals` (and so on) in all the scheduler tests.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: After discussing with @mattdailis, I have replaced these `_L` by `_double` when they are double and by `_long` when they are long. `xn` have been replaced by `x_n` and `xn_m1` have been replaced by `x_nminus1`.

mattdailis (approved):
Looks good to me, I'm sure we'll learn more about the performance characteristics over time and adjust how we select which algorithms work best

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -0,0 +1,460 @@
+package gov.nasa.jpl.aerie.scheduler;
```
goetzrrGit, line 1: These are some fancy algorithms. I am going to pass on trying to understand them.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
```diff
@@ -173,31 +154,31 @@ protected RecurrenceGoal() { }
...
   {
     final var conflicts = new java.util.LinkedList<MissingActivityConflict>();
-
-    var goalConstraint = this.stateConstraints;
-
     //determine how much flexibility there is in creating activities
     final var recurrenceFlexibility = recurrenceInterval.end.minus(
         recurrenceInterval.start);
 
+    if(end.minus(start).noLongerThan(recurrenceInterval.end)){
```
goetzrrGit: `noLongerThan` is kinda weird. Would this work better `notLongerThan`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
```diff
@@ -173,31 +154,31 @@ protected RecurrenceGoal() { }
...
-      if(goalConstraint != null) {
-        interval.intersectWith(goalConstraint.findWindows(plan, new Windows(Window.between(start, end))));
-      }
+      var interval =  new Windows(Window.between(intervalT.minus(recurrenceInterval.end), Duration.min(intervalT, end)));
       conflicts.add(new MissingActivityTemplateConflict(
           this, new Windows(
           interval),this.getActTemplate()));
+      if(intervalT.compareTo(end) > 0){
+        break;
+      }
```
goetzrrGit: Maybe you could wrap this in a do-while loop instead of a for loop:

```
    var intervalT = start.plus(recurrenceInterval.end);
    do{
      //REVIEW: technically, could create activity at extremely short
      //        intervals (ie minT=0) and still satisfy the goal as currently
      //        framed, which is exactly what the current solver will do since
      //        it chooses the minimum. but it looks ugly. so for now passing
      //        a limited flexibility for creation
      final var minT = intervalT.minus(recurrenceFlexibility);
      var interval =  new Windows(Window.between(intervalT.minus(recurrenceInterval.end), Duration.min(intervalT, end)));
      conflicts.add(new MissingActivityTemplateConflict(
          this, new Windows(
          interval),this.getActTemplate()));
      intervalT = intervalT.plus(recurrenceInterval.end)
   
    } while(intervalT.compareTo(end) <= 0);
```

Also, minT isn't being used in the loop

` final var minT = intervalT.minus(recurrenceFlexibility);`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TimeUtility.java
```diff
@@ -0,0 +1,118 @@
+package gov.nasa.jpl.aerie.scheduler;
```
goetzrrGit, line 1: A more general question do we not have a Time calculation class that does time conversion and math?

goetzrrGit (commented):
I am not too familiar with the scheduling side of things, but the code looks fine with one minor tweak.

I am not sure about the algorithm part as I don't know of a way to verify them.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
adrienmaillard: You want to change the `Duration` method names everywhere?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TimeUtility.java
adrienmaillard: I do not think so. I think we would benefit from having a central place for these things. Is there a module that all other already depend on or could depend on? Also, do you think there would be something wrong with these robust-to-overflow operations moving inside Duration?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
Twisol: The methods were named in the pattern "x is shorter than y" (strictly less than) and "x is no longer than y" (less-than-or-equal). While "x is not longer than y" technically says the same thing, it doesn't sound right to my ear.

Framed differently, saying "I'm no taller than you" has different emphasis than saying "I'm not taller than you".

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TimeUtility.java
Twisol: `Duration` is part of the `merlin.protocol` package, which means it's part of the data-carrying interface between the simulation driver and a model. I'd rather not upstream scheduler utilities into a package that everyone else is forced to depend on.

Twisol: Arguably, the `Duration` in `merlin.protocol` should be significantly reduced in scope to begin with. If it's spread far beyond its purpose as a data carrier between driver and model, then it's too attractive as a coupling point, and we should try to minimize that potential.

adrienmaillard: I see. Note that I was only thinking about pulling the "bounded" operations to the `Duration` class, not centralize conversions between "string" time and `Duration`. Another class would do that.

Twisol: Yes, fair.

It's not expected that modelers will need to worry about the end of time (so to speak); if a simulation spans 600,000 years, an `ArithmeticException` is likely just a symptom of a deeper problem. Saturating operations would unfortunately only hide the problem; it's not the right semantics for model use-cases. `Duration` does not even provide an absolute frame of reference, as almost all uses between the model and driver are relative (e.g. `delay(2, SECONDS)`.

I suspect that, if the scheduler has legitimate need for saturating operations on durations, it's confronting time with a meaningfully different attitude than is meant to be served by the `merlin.protocol` `Duration`.

mattdailis: (Correct me if I'm wrong, Adrien) - I think we're most likely to run into overflow issues where we denote "inifinity" by "Duration.MAX_VALUE". For example, in `Window`:
https://github.com/NASA-AMMOS/aerie/blob/45d42efc91dbe676fd9aad9e620315a637c1cfe9/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Window.java#L125

Using all of the operations in `Window`, this poses no issue. However, the scheduler adds additional operations to shift windows forward and backward in time - which would result in an overflow.

I bet we could model our way around this without running into ArithmeticExceptions, or needing to change `Duration` directly.

(If there's another use case I'm missing here, let me know @adrienmaillard )

Twisol: That's a good point, and I think I committed the original sin there. I agree that if you're translating windows around, this representation of "forever" no longer cuts it.

adrienmaillard: You may be right @mattdailis. The overflow especially happens with loose temporal constraints such as endsafter() or endsbefore(), startsafter(), startsbefore(). They were modelled as a minimal operation on a time point where you add/subtract infinity and get the resulting interval without really modelling the constraint explicitly. We could revise that and do differently but I'd say this is out of scope (and there is already a lot of out-of-scope stuff in this PR).

Twisol: Yes, FWIW I'm not pushing back on anything in this PR (not that I've looked!). Just against upstreaming stuff into Duration.

mattdailis: Totally agree on the scope :+1: I've captured this potential design improvement in a JIRA ticket: [Model "forever" better in the scheduler](https://jira.jpl.nasa.gov/browse/AERIE-1780)

Copying the content below for the benefit of those who don't have access to JIRA (which is most people in this thread... 🙄 ):

> As of writing, the scheduler models "forever" as a window from Duration.MIN_DURATION to Duration.MAX_DURATION. This worked well for operations like intersections (defined on constraints) but it causes an arithmetic overflow for the additional operations added by the scheduler (which might look to create windows before and after another window).
>
> The goal of this ticket is to design and implement a solution that does not use MAX_VALUE and MIN_VALUE to denote infinite durations. This solution should not need to touch any classes in merlin.protocol, since this is purely a scheduling concern.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
adrienmaillard: I've done what you suggested.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TaskNetwork.java
```diff
@@ -107,7 +107,9 @@ public void addEnveloppe(String nameAct, String envName, double t1, double t2){
     stn.addTimepoint(etenvTpName);
 
     stn.addDurCst(startHorizon, stenvTpName, t1-stHorizon, t1-stHorizon);
-    stn.addDurCst(endHorizon, etenvTpName, t2-stHorizon, t2-stHorizon);
+    stn.addDurCst(startHorizon, etenvTpName, t2-stHorizon, t2-stHorizon);
```
patkenneally: :large_blue_circle: What issue was fixed here? The commit message doesn't give me any indication on what fix I should be seeing here. This might be a candidate for an added explanation paragraph below the initial commit message.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/STNTest.java
```diff
@@ -81,5 +81,18 @@ public void testEnveloppe(){
...
+    var stn = new TaskNetwork(2,10);
+    stn.addAct("Act");
+    stn.addStartInterval("Act", 2,15);
+    stn.addDurationInterval("Act", 1, 20);
+    stn.addEnveloppe("Act", "win", 3,7);
+    var success = stn.propagate();
+    assert(success);
+    var d = stn.getAllData("Act");
+    System.out.println(d);
+    //var expD1 = new TaskNetwork.TNActData(Pair.of(3.,3.), Pair.of(4.,4.), Pair.of(1.,1.));
```
patkenneally: :large_orange_diamond: Is this commented out code important? If not can it be deleted?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/RecurrenceGoal.java
adrienmaillard: Actually, I rolled back on these changes because the number of conflicts produced is different. Let's take an example with `start` = 0, `end` = 2h10 and a `recurrenceInterval` of 1 hour. With the current impl, here is what happen for each iteration : 
(1) `intervalT` = 0 + 1h = 1h, we insert a conflict at 1h
(2) `intervalT` = 1h + 1h = 2h, we insert a conflict at 2h
(3) `intervalT` = 2h+1h = 3h, `intervalT` is now after the end but nonetheless we need to create a conflict for the period [2h, 2h10]. The for loop does not check if we are beyond `end`. We create the conflict and we ensure that the temporal context of the conflict does not go beyond `end` (with the `Duration.min` in line 175). And then we break. 

Here are now what happens with your proposed update: 
`intervalT` = 0 + 1h = 1h
(1) we insert a conflict at 1h, then increment `intervalT` to 2h
(2) we insert a conflict at 2h, then increment `intervalT` to 3h, then we exit the loop

We would exit the loop when we increment `intervalT` to 3h before creating the appropriate conflict for the last period.

goetzrrGit (approved):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/ActivityExpression.java
```diff
@@ -275,6 +290,16 @@ B basedOn(@NotNull ActivityInstance existingAct) {
     public abstract @NotNull
     AT build();
 
+    private Window extendUpToAbsoluteError(Window window, Duration absoluteError){
```
patkenneally: :large_orange_diamond: `final` on parameters

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/CardinalityGoal.java
```diff
@@ -3,22 +3,21 @@
...
-import java.util.TreeMap;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 /**
  * describes the desired coexistence of an activity with another
  */
 public class CardinalityGoal extends ActivityTemplateGoal {
+
+  private static final Logger logger = LoggerFactory.getLogger(ActivityCreationTemplate.class);
```
patkenneally: :large_orange_diamond: Is this logger using the correct class or is this a copy/paste error?

patkenneally (approved):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/TaskNetwork.java
adrienmaillard: The `addDurCst` method is creating a duration constraint between timepoints. `startHorizon` and `endHorizon` are timepoints representing the planning horizon. The method here is setting an "envelope" constraint for an activity, meaning that the activity must start and finish between given `t1`, `t2` bounds. But in a STN, absolute timepoints do not exist ("march 26") and must be defined with regard to a virtual origin which is what these two constraints do for the timepoints of this "envelope" (setting an given duration between the start of the horizon and the start/end of the envelope) in which the activity will then be later constrained (below in this method). 
In this context, the constraint was simply wrongly defined.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/STNTest.java
adrienmaillard: Will uncomment and make this test work.

PR #114, opened 2022-03-22
[HOTFIX] Merlin-Server Refactor Issues Fix

kroffo:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Two issues made it through my PR for AERIE-1710 that @camargo found (we really should have a full test with the UI for all branches before merging, sorry my own testing was not diligent enough to catch this!).

1. The SetSimulationStateAction now references the status column rather than the previously named state column. The ResultsProtocol.State reference was also lifted out of this action to the PostgresResultsCellRepository.

2. The UpdateSimulatedActivityParentsAction contained a bug causing it to try to set the parent ID only for activities with no parents! This condition has been flipped so that in fact only activities that actually have a parent will have their parent ID set.

## Verification
I tested that simulation failed first, made the changes, and saw the issues disappear. @camargo will also verify before we merge this.

## Documentation
N/A

## Future work
Let's hope none!


mattdailis (approved):

patkenneally (approved):
Nice. Funnily, I also found these issues just this morning while rebasing on develop.

camargo (approved):
Works on my end! Thanks for the quick fix

PR #120, opened 2022-03-28
[AERIE-1694] Provide Scheduling DSL Typescript via API

mattdailis:
* **Tickets addressed:** AERIE-1694
* **Review:** By commit  

## Description

The task of providing the Typescript to the API involved a few things:
1. Making sure that the Typescript we generate includes `declare global` type declarations, which are what allows the intellisense in the code editor to make those types available without requiring users to add `import` statements to their goal definitions.
2. Adding plumbing to connect the API to the `TypescriptCodeGenerationService`.

The `Use vm.Module in scheduling-dsl-compiler` commit modifies the `executor.ts` to compile the user code and the library code separately, then link them together. This is a step in the direction of using [aerie-ts-code-runner](https://github.com/NASA-AMMOS/aerie-ts-user-code-runner), though I'm not quite ready to make the jump to that in this PR.

## Verification

1. Bring up a fresh aerie via `docker-compose up -d`
2. Upload a mission model
3. Call the `schedulingDslTypescript` Hasura action and print out the result (I did this in a node script)
4. Copy the outputted typescript into a string, and add it as an `extraLib` to Monaco (screenshot below):
Screen Shot 2022-03-28 at 3 22 17 PM

## Documentation

I have a ticket (AERIE-1755) to start documenting the Scheduling DSL - this functionality will need to be documented as part of that effort.

## Future work
- Properly detect and escalate typescript type checking errors, as well as javascript runtime errors. Using [aerie-ts-code-runner](https://github.com/NASA-AMMOS/aerie-ts-user-code-runner) may be the best bet here


camargo (approved):
Works well on the front-end! LGTM

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
```diff
@@ -24,7 +24,7 @@ public SchedulingDSLCompilationService(final TypescriptCodeGenerationService typ
     this.typescriptCodeGenerationService = typescriptCodeGenerationService;
     final var schedulingDslCompilerRoot = System.getenv("SCHEDULING_DSL_COMPILER_ROOT");
     final var schedulingDslCompilerCommand = System.getenv("SCHEDULING_DSL_COMPILER_COMMAND");
-    this.nodeProcess = new ProcessBuilder("node", schedulingDslCompilerCommand)
+    this.nodeProcess = new ProcessBuilder("node", "--experimental-vm-modules", schedulingDslCompilerCommand)
```
camargo: Do we know the absolute `node` path to use here?

mattdailis: It's different depending on whether you're running unit tests, or running in Docker 🤔 In Docker it's wherever nvm installs it to. In unit testing, it's in the gradle directory.

I could potentially make a `NODE_PATH` environment variable, but I'm not sure if that's better from a security perspective.

Open to ideas for how to do this better

camargo: hmm yeah `NODE_PATH` is decent but you're right it potentially moves the issue to another env var. I think for now `NODE_PATH` is a good compromise because `PATH` is system-wide and `NODE_PATH` is set for the application specifically.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -41,15 +41,24 @@ void testCodeGen() {
...
-                 + "    name: string\n"
                  + "    args: {\n"
                  + "      duration: Duration\n"
                  + "      fancy: { subfield1: string, subfield2: { subsubfield1: Double, }[], }\n"
                  + "      peelDirection: (\"fromTip\" | \"fromStem\")\n"
                  + "    }): PeelBanana {\n"
-                 + "      return { name, activityType: 'PeelBanana', args };\n"
+                 + "      return { activityType: 'PeelBanana', args };\n"
                  + "    },\n"
                  + "}\n"
```
dyst5422, line 51: Semicolon missing here to close out the const assignment

scheduler-server/scheduling-dsl-compiler/src/executor.ts
```diff
@@ -1,15 +1,56 @@
...
+}
+
+export default async function executeSourceCode(opts: { source: string, filename: string, dslTS: string, generatedCode: string }) {
+  const compilerOptions = {
+    module: ts.ModuleKind.ES2022,
       target: ts.ScriptTarget.ESNext,
       sourceMap: true,
-    }
+  }
+  const transpiledSource = ts.transpileModule(opts.source, {
```
dyst5422, line 52: This transpiling doesn't catch/produce type errors. It simply does type annotation erasure. Probably would be good to pair program bringing in https://github.com/NASA-AMMOS/aerie-ts-user-code-runner as it handles all of that

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -41,15 +41,24 @@ void testCodeGen() {
...
+                 + "declare global {\n"
+                 + "  var ActivityTemplates: {\n"
+                 + "    PeelBanana: (\n"
+                 + "      args: {\n"
+                 + "        duration: Duration\n"
```
dyst5422, line 56: Duration isn't a JavaScript entity, so this will surely fail at runtime/typecheck time. The actual implementations for the activity template generators aren't in scope at runtime either. We solved this on the command expansion side of things by putting those on the global object in the DSL file.

dyst5422 (changes requested):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
mattdailis: I've gone the `NODE_PATH` route :+1:

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
mattdailis: I've added a `Duration` class to the DSL, so instead of `24 * 60 * 60 * 1000 * 1000`, the syntax will now be `Duration.ofHours(24)`.

scheduler-server/scheduling-dsl-compiler/src/executor.ts
mattdailis: As Dylan and I discussed in the huddle, producing useful type errors will be important to a good user experience. That said, it is out of the scope of this PR, which is focused on providing types to the UI prior to execution.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationService.java
mattdailis: Alas, I cannot for the life of me figure out what the absolute path to node is in Github Actions

mattdailis: Alright - a compromise I've reached is that in production, `NODE_PATH` will be set to `"/root/.nvm/versions/node/v${NODE_VERSION}/bin/node"`. In the `./gradlew test` environment, it will look in the `.gradle` directory to try to find the version of node that gradle installed. If it can't find that, it will set `NODE_PATH` to `"node"`, and fall back to whatever node is on the path.

I think this follows the spirit of the rule, since in production, it will always use an absolute path. I think it's okay to bend the rule in the unit testing environment (though I am still confused why gradle in the github action doesn't install node in the same location as when I run it on my machine 😕 )

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
mattdailis: As @dyst5422 pointed out, there's a new feature coming down the pipeline in typescript that adds a `Duration` class to the standard library. It is worth considering using that for the scheduling dsl. However, I've decided that I want to give it some careful thought, and make it a separate PR. I've dropped the commit that changes the type of Duration, so it is now just a type alias for `number`.

dyst5422:
I still have big concerns with NOT having typechecking being part of the initial feature - having scheduling rules stop working once we bring that in is going to be not good for the users.

mattdailis:
> I still have big concerns with NOT having typechecking being part of the initial feature - having scheduling rules stop working once we bring that in is going to be not good for the users.

I agree with you - but that's not in scope for this PR. Let's make a ticket for that now.

mattdailis:
I created [AERIE-1767](https://jira.jpl.nasa.gov/browse/AERIE-1767) and assigned it to myself :+1: We can prioritize getting it in before the release if we get concurrence from @patkenneally

dyst5422 (approved):

PR #121, opened 2022-03-28
[AERIE-1709]  Avoid regenerating activities after scheduling

adrienmaillard:
* **Tickets addressed:** AERIE-1709
* **Review:** By commit  

## Description

This ticket has been created to address the following behavior of the scheduler server: when finished with a scheduling request, instead of sending updates to merlin, the plan was deleted and then regenerated from scratch. This updated the ids of activities even though their properties were not affected. 

The approach taken has been to discriminate between activities to add, activities to delete, and activities to update.
For that, we need to compare the initial plan with the new plan. To do that, we introduce two new classes representing simple merlin activities and merlin plans.  

When fetching the initial plan before scheduling, we save a version of the merlin distant plan for later comparison. We also save a mapping from merlin activity instance ids to scheduler activity instance id. 

Then after scheduling, we loop through activities from the new plan:  
If an activity with the same id is not present in the old plan, it is a new activity, we insert it in the distant merlin plan. 
If it is present, we check if it has changed by comparing their two merlin representations and if they are not the same, we update the merlin instance with the one from the plan (the most "recent"). 
And if an activity is present in the old plan but absent from the new plan, it means that the activity was deleted during scheduling, we then delete it from the distant merlin plan. 

The first three commits are performing these changes. 

The next four commits are removing some useless code and organizing the scheduler into packages. 

## Verification

For changes not related with the ticket: all tests are passing.  

For changes related with the ticket, I performed tests from the UI with the bananation mission model. When hitting several times the scheduler, even if new activities are inserted in the plan, the ids of activities already present do not change. 
 
ids

## Documentation

Please check the validity of the update/delete/add model.

## Future work

None regarding this issue.


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Plan.java
```diff
@@ -58,7 +55,14 @@
    *
    * @return map of all activities in the plan by type
    */
-  java.util.Map<String, java.util.List<ActivityInstance>> getActivitiesByType();
+  java.util.Map<ActivityType, java.util.List<ActivityInstance>> getActivitiesByType();
```
patkenneally: 🔻 In keeping with the rest of the Aerie code base, we can add these name spaces as explicit imports and remove them here and anywhere else added in this PR.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/MerlinPlan.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class MerlinPlan {
+
+  private final Map<ActivityInstanceId, MerlinActivityInstance> activities;
+
+  public MerlinPlan(){
+    activities = new HashMap<>();
+  }
+
+  public void addActivity(ActivityInstanceId id, MerlinActivityInstance activity){
```
patkenneally: :large_orange_diamond: Should these parameters be `final`?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/MerlinPlan.java
```diff
@@ -0,0 +1,28 @@
...
+    if(activities.containsKey(id)){
+      throw new IllegalArgumentException("Activity with same id is already in the plan");
+    }
+    activities.put(id, activity);
+  }
+
+  public Optional<MerlinActivityInstance> getActivityById(ActivityInstanceId id){
```
patkenneally: 🔶 Should this parameter be final?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -185,17 +191,52 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
...
+    final var response = postRequest(request).orElseThrow(() -> new NoSuchPlanException(planMetadata.planId()));
+    final var jsonplan = response.getJsonObject("data").getJsonObject("plan_by_pk");
+    var activities = jsonplan.getJsonArray("activities");
+    final var plan = new PlanInMemory();
+    for (int i = 0; i < activities.size(); i++) {
+      final var jsonActivity = activities.getJsonObject(i);
+      String type = activities.getJsonObject(i).getString("type");
+
+      if(!activityTypes.containsKey(type)){
+        throw new IllegalArgumentException("Activity type found in JSON object after request to merlin server has not been found in types extracted from mission model. Probable inconsistency between mission model used by scheduler server and merlin server.");
```
patkenneally: :large_orange_diamond: Let's use multi-line string here e.g. 
`"some string text"`
`+ "more string"`
`+ "more"`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/MerlinActivityInstance.java
```diff
@@ -0,0 +1,20 @@
+package gov.nasa.jpl.aerie.scheduler.server.models;
```
patkenneally, line 1: :large_orange_diamond: this file has a number of extra new lines that can be removed and each parameter in the constructor prototype can take a new line.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -44,7 +47,7 @@ public PlanMetadata getPlanMetadata(final PlanId planId) {
       }
 
       @Override
-      public Plan getPlanActivities(final PlanMetadata planMetadata, final Problem mission)
+      public Triple<Plan, MerlinPlan, Map<SchedulingActivityInstanceId, ActivityInstanceId>> getPlanActivities(final PlanMetadata planMetadata, final Problem mission)
```
JoelCourtney: (not a review) What's the difference between a `Plan` and a `MerlinPlan`?

adrienmaillard: A `Plan` is the scheduler's representation of a plan, with partially instanciated activity parameters, activity durations and so on.  A `MerlinPlan` is limited to the representation of a plan in the merlin database.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -189,17 +195,52 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
 
   /**
    * {@inheritDoc}
+   * @return
    */
   @Override
-  public Plan getPlanActivities(final PlanMetadata planMetadata, final Problem problem)
-  throws IOException, NoSuchPlanException
+  public Triple<Plan, MerlinPlan, Map<SchedulingActivityInstanceId, ActivityInstanceId>> getPlanActivities(final PlanMetadata planMetadata, final Problem problem)
```
mattdailis: Just in terms of organization of the code - it seems a bit out of place for `GraphQLMerlinService` to know about `Problem`. (I know this was the case before this PR, just noticing this now).

So I guess I would expect MerlinService to return a `MerlinPlan`, and then `loadInitialPlan` can convert a `MerlinPlan` into a `Plan`.

adrienmaillard: Sounds cleaner, I am going to do that.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -286,16 +308,86 @@ public void createSimulationForPlan(final PlanId planId)
...
+          updateActivity(planId, schedulerActIntoMerlinAct, activityInstanceId);
+        }
+        ids.put(activity, activityInstanceId);
+      } else {
+        //act was not present in initial plan, create new activity
+        toAdd.add(activity);
+      }
+    }
+    final var actsFromNewPlan = plan.getActivitiesById();
+    for (final var idActInInitialPlan : idsFromInitialPlan.keySet()) {
```
mattdailis: IntelliJ is telling me that we can use `entrySet()` instead of `keySet()`, so we won't need to call `.get()` below.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -212,10 +217,26 @@ private Solver createScheduler(final PlanMetadata planMetadata, final Problem pr
    * @throws ResultsProtocolFailure when the requested plan cannot be loaded, or the target plan revision has
    *     changed, or aerie could not be reached
    */
-  private Plan loadInitialPlan(final PlanMetadata planMetadata, final Problem problem) {
+  private Triple<Plan, MerlinPlan, Map<SchedulingActivityInstanceId, ActivityInstanceId>> loadInitialPlan(final PlanMetadata planMetadata, final Problem problem) {
```
mattdailis: I know we only use this `Triple` in one place, but I think it would help readability (on the caller's side) to define a record:

```
record PlanComponents(Plan schedulerPlan, MerlinPlan merlinPlan, Map<SchedulingActivityInstanceId, ActivityInstanceId> idMap) {}
```
It can go right below this function, next to the `SchedulerMissionModel` record, which serves a similar purpose.

This way, the caller code can refer to `loadedPlanComponents.idMap()` and  `loadedPlanComponents.merlinPlan()` instead of `loadedPlanComponents.getRight()` and  `loadedPlanComponents.getMiddle()`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Plan.java
```diff
@@ -58,7 +55,14 @@ public interface Plan {
...
+
+  /**
+   * fetches activities in the plan by type
```
patkenneally: 🔻 These Java docs seem like they were copy and pasted but not edited for their new context.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -189,17 +191,37 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
...
+    MerlinPlan merlinPlan = new MerlinPlan();
+    final var request =
+        "query { plan_by_pk(id:%d) { activities { id start_offset type arguments } duration start_time }} ".formatted(
+            planMetadata.planId().id());
+    final var response = postRequest(request).orElseThrow(() -> new NoSuchPlanException(planMetadata.planId()));
+    final var jsonplan = response.getJsonObject("data").getJsonObject("plan_by_pk");
+    var activities = jsonplan.getJsonArray("activities");
+    for (int i = 0; i < activities.size(); i++) {
+      final var jsonActivity = activities.getJsonObject(i);
+      String type = activities.getJsonObject(i).getString("type");
```
patkenneally: :large_orange_diamond: `final var` for these?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -189,17 +191,37 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
...
+      String start = jsonActivity.getString("start_offset");
+      var arguments = jsonActivity.getJsonObject("arguments");
```
patkenneally: :large_orange_diamond: `final` ?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -189,17 +191,37 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
...
+      final var deserializedArguments = BasicParsers
+          .mapP(new SerializedValueJsonParser()).parse(arguments)
```
patkenneally: :large_blue_circle: this reads easier with `.parse` dropped onto a new line.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -189,17 +191,37 @@ public PlanMetadata getPlanMetadata(final PlanId planId)
...
+          .getSuccessOrThrow((reason) -> new InvalidJsonException(new InvalidEntityException(List.of(reason))));
+      final var merlinActivity = new MerlinActivityInstance(type, Duration.of(parseGraphQLInterval(start).getDuration().toNanos() / 1000, Duration.MICROSECONDS), deserializedArguments);
+      final var actPK = new ActivityInstanceId(jsonActivity.getJsonNumber("id").longValue());
+      merlinPlan.addActivity(actPK, merlinActivity);
+    }
+    return merlinPlan;
   }
 
+
```
patkenneally: :large_blue_circle: Some extra new lines here.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -286,16 +308,86 @@ public void createSimulationForPlan(final PlanId planId)
...
 
+  public void deleteActivity(final ActivityInstanceId id)
+  throws MerlinServiceException, IOException, NoSuchPlanException
+  {
+    final var request = "delete_activity_by_pk( id : %d ){ id }".formatted(id);
+    final var response = postRequest(request).orElseThrow(() -> new NoSuchPlanException(null));
+    try {
+      response.getJsonObject("data").getJsonObject("delete_activity_by_pk").getJsonNumber("id").longValueExact();
+    } catch (ClassCastException | ArithmeticException e) {
+      throw new NoSuchPlanException(null);
```
patkenneally: :large_orange_diamond: I'm unclear about why this and above are throwing a `NoSuchPlanException`, and with a null reason. Can you help me understand your reasoning?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -286,16 +308,86 @@ public void createSimulationForPlan(final PlanId planId)
...
+
+    final var response = postRequest(updateReq).orElseThrow(() -> new NoSuchPlanException(planId));
+    try {
+      response.getJsonObject("data").getJsonObject("update_activity_by_pk").getJsonNumber("id").longValueExact();
+    } catch (ClassCastException | ArithmeticException e) {
+      throw new NoSuchPlanException(planId);
+    }
+  }
+
+
```
patkenneally: :large_orange_diamond: Extra new line

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -286,16 +308,86 @@ public void createSimulationForPlan(final PlanId planId)
...
+      var value = getGraphQLValueString(arg.getValue());
+      argumentsSb.append(argFormat.formatted(name, value));
+    }
+    final var updateReq = """
+        mutation {
+          update_activity_by_pk( pk_columns : { id : %d, plan_id : %d }, _set: {start_offset : \"%s\", arguments : { %s }}) {
+           affected_rows
+          }
+        }
+        """.formatted(id, planId, activity.startTimestamp().toString(), argumentsSb.toString());
```
patkenneally: :small_red_triangle_down: Should these be `id.id()` and `planId.id()` ? Also for the variable name `id` it seems like it would read better as something like `instanceId` or similar.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -286,16 +308,86 @@ public void createSimulationForPlan(final PlanId planId)
...
+
+    final var response = postRequest(updateReq).orElseThrow(() -> new NoSuchPlanException(planId));
+    try {
+      response.getJsonObject("data").getJsonObject("update_activity_by_pk").getJsonNumber("id").longValueExact();
+    } catch (ClassCastException | ArithmeticException e) {
+      throw new NoSuchPlanException(planId);
```
patkenneally, line 385: :small_red_triangle_down: The same question as above; should this truly be a `NoSuchPlanException` ? Or is something more akin to some parsing exception appropriate?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -120,7 +126,7 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
       //store the solution plan back into merlin (and reconfirm no intervening mods!)
       //TODO: make revision confirmation atomic part of plan mutation (plan might have been modified during scheduling!)
       ensurePlanRevisionMatch(specification, getMerlinPlanRev(specification.planId()));
-      final var instancesToIds = storeFinalPlan(planMetadata, solutionPlan);
+      var instancesToIds = storeFinalPlan(planMetadata, loadedPlanComponents.getRight(), loadedPlanComponents.getMiddle(), solutionPlan);
```
patkenneally: :large_orange_diamond: We lost the `final` here. I'd also be partial to dropping the method parameters each to a new line so this line doesn't run so far off to the right.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -212,10 +215,26 @@ private Solver createScheduler(final PlanMetadata planMetadata, final Problem pr
...
+      final var merlinPlan =  merlinService.getPlanActivities(planMetadata, problem);
+      final Map<SchedulingActivityInstanceId, ActivityInstanceId> schedulingIdToMerlinId = new HashMap<>();
+      final var plan = new PlanInMemory();
+      final var activityTypes = problem.getActivityTypes().stream().collect(Collectors.toMap(ActivityType::getName, at -> at));
+      for(final var elem : merlinPlan.getActivitiesById().entrySet()){
+        final var activity = elem.getValue();
+        if(!activityTypes.containsKey(activity.type())){
+          throw new IllegalArgumentException("Activity type found in JSON object after request to merlin server has not been found in types extracted from mission model. Probable inconsistency between mission model used by scheduler server and merlin server.");
+        }
+        var schedulerActType = activityTypes.get(activity.type());
```
patkenneally: :large_orange_diamond: `final` ?

patkenneally (changes requested):

Mostly formatting changes :)

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
Twisol:
```java
"""
These work too, and were added a
few versions of Java
ago.
"""
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: In general, this exception as well as `NoSuchMissionModelException` are thrown because we **assume** this is the reason it fails. In this case, it does not make a lot of sense so I am going to replace it with a new exception `NoSuchActivityInstanceException`.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: See my above comment about the other case. In this case, there are more reasons why it may fail (plan is absent or activity is absent) so we pick one. Failure at the class cast or arithmetic level only means that something is wrong with the request. In other words, we assume the json unpacking is right on our side and they are proxy exception for the real issue.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/Plan.java
adrienmaillard: Did this in `Plan` and `PlanInMemory`.

adrienmaillard:
I've just force-pushed changes to address all of the above comments.

patkenneally (approved):

mattdailis (approved):
Looks good to me, thanks for making those changes 💯

PR #122, opened 2022-03-31
test: add e2e-test project

camargo:
* **Review:** By commit  

## Description

- Runs all tests in the repo against a complete, fresh instance of Aerie (including db, unit, and e2e tests)!
- Include Playwright and add some starter e2e API tests
- Create testing workflow to execute tests in GH Actions
- Report all test results as GH Action artifacts

I removed the PR workflow because it's redundant when this gets merged. The new test workflow should be the main gating PR workflow.

See example output from the new test workflow here: https://github.com/NASA-AMMOS/aerie/actions/runs/2073963676

## Verification

Run `./gradlew test`

Edit: You need all Aerie containers running for the e2e tests to execute properly.

## Documentation

See README.md


e2e-tests/.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
pcrosemurgy: I haven't encountered an `.editorconfig` file before, is it kind of like an `.idea` analogue? If so then maybe it should be git-ignored.

e2e-tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+test-results
```
pcrosemurgy, line 2: Minor note: I usually prefer a top-level `.gitignore`. From looking through our other `.gitignores` they seem to have quite a bit of duplication between each other and compared to the top-level one (see `merlin-server/.gitignore` which contains a lot of items that are in the main `.gitignore` anyways).

e2e-tests/.editorconfig
camargo: It should not be ignored. It's to ensure when editing code in this folder that your editor is using the proper text format settings: https://editorconfig.org/

e2e-tests/src/utilities/gql.ts
```diff
@@ -0,0 +1,22 @@
...
+  DELETE_MISSION_MODEL: `#graphql
+    mutation DeleteModel($id: Int!) {
+      delete_mission_model_by_pk(id: $id) {
+        id
+      }
+    }
+  `,
+};
+
+export default gql;
```
pcrosemurgy, line 22: This is awesome. In case anyone else is wondering this `const gql` can eventually be populated with the contents of https://github.com/NASA-AMMOS/aerie-ui/blob/develop/src/utilities/gql.ts, almost verbatim. :)

e2e-tests/.gitignore
camargo: I actually typically prefer the opposite for monorepos. The idea is each package is standalone and ignores what it needs.

pcrosemurgy: That's fair :)

e2e-tests/.editorconfig
pcrosemurgy: Oh cool, haven't seen that standard before!

pcrosemurgy (approved):
Thanks a lot Chris!

e2e-tests/.editorconfig
camargo: I will say it's not strictly required because I added Prettier to the project. But it's nice if you have Editorconfig in your editor already.

mattdailis:
Looks awesome! I'm curious what we want it to do when we run `./gradlew test` locally.

I just checked out this branch and ran `./gradlew test` to see what would happen. The `db-tests` run fine, then the `e2e-tests` hit some errors (probably because I haven't deployed the correct docker containers). Then they open up an "html report" and pause:

Screen Shot 2022-03-31 at 6 23 43 PM

When I press Ctrl+C, it exits out to the shell, instead of continuing with the tests in the other subprojects.

e2e-tests/.editorconfig
Twisol: There's also a `.editorconfig` already in the top-level repo; it was set up for the Java projects. I broadly agree with your comment [over here](https://github.com/NASA-AMMOS/aerie/pull/122#discussion_r840097019) regarding monorepos, but maybe it makes sense to have a single consistent `.editorconfig`? It's more about the human developers interacting with Aerie projects than the specifics of any one project.

If you agree, maybe it would be good to merge this in with the top-level `.editorconfig`? Alternatively, we could duplicate the current `.editorconfig` into all Java projects, but this seems like a recipe for accidentally having subtly different code styles throughout the codebase.

mattdailis (approved):
The separate `test` and `e2eTest` gradle tasks seem like a good strategy. Approving :+1:

PR #125, opened 2022-04-04
test: add ui health check

camargo:
## Description

Adds an e2e test that checks the UI server health

### DIFF
--- a/e2e-tests/docker-compose-test.yml
+++ b/e2e-tests/docker-compose-test.yml
@@ -99,6 +99,19 @@ services:
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store
+  aerie_ui:
+    container_name: aerie_ui
+    depends_on: ["postgres"]
+    environment:
+      AUTH_TYPE: none
+      GATEWAY_CLIENT_URL: http://localhost:9000
+      GATEWAY_SERVER_URL: http://aerie_gateway:9000
+      HASURA_CLIENT_URL: http://localhost:8080/v1/graphql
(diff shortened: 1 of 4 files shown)


dyst5422 (approved):
This isn't directly related to this PR, but why did you make the mission model type global instead of simply importing it where necessary?

camargo:
@dyst5422 Just for convenience. Since the types are not really part of a specific library it just reduces the boilerplate needed to use what's in the `types` directory.

dyst5422:
> @dyst5422 Just for convenience. Since the types are not really part of a specific library it just reduces the boilerplate needed to use what's in the `types` directory.

I don't REALLY care, but just as devils advocate... I'd say that polluting the global namespace for type checking is nearly as bad as at runtime and should only really be used for extending built-ins/globalThis.

camargo:
@dyst5422 Yeah I used to think this way. I think app-specific types are safe for global use across an application and you don't gain much extra safety from using them in modules since they are compiled away anyway. Any collisions can be addressed with a type namespace, and there is less overall syntax to use them. You avoid really long import statements of different types, or just using * to import all the types into a variable anyway.

mattdailis (approved):

dyst5422:
> @dyst5422 Yeah I used to think this way. I think app-specific types are safe for global use across an application and you don't gain much extra safety from using them in modules since they are compiled away anyway. Any collisions can be addressed with a type namespace, and there is less overall syntax to use them. You avoid really long import statements of different types, or just using * to import all the types into a variable anyway.

I'll let you convince me as we go 😃

PR #130, opened 2022-04-07
[HOTFIX] Gave aerie user permission to command expansion db

goetzrrGit:
## Description
I noticed that we didn't give the aerie user permission to access the command expansion database. Right now that isn't a problem but we should probably set up the permissions so we don't run into any issues in the future.

## Verification
Nuke everything and verified the command expansion database was created with all its tables.

### DIFF
--- a/deployment/postgres-init-db/init-aerie.sh
+++ b/deployment/postgres-init-db/init-aerie.sh
@@ -29,7 +29,7 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-E
 
   \echo 'Initializing aerie_commanding database...'
   CREATE DATABASE aerie_commanding;
-  GRANT ALL PRIVILEGES ON DATABASE aerie_ui TO aerie;
+  GRANT ALL PRIVILEGES ON DATABASE aerie_commanding TO aerie;
   \echo 'Done!'
 
 EOSQL


patkenneally (approved):

deployment/postgres-init-db/init-aerie.sh
```diff
@@ -29,7 +29,7 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-E
 
   \echo 'Initializing aerie_commanding database...'
   CREATE DATABASE aerie_commanding;
-  GRANT ALL PRIVILEGES ON DATABASE aerie_ui TO aerie;
+  GRANT ALL PRIVILEGES ON DATABASE aerie_commanding TO aerie;
```
mattdailis, line 32: I am curious why this wasn't an issue before - do we not use the `aerie` user when our services interact with the database?

mattdailis (approved):
Looks good to me - I left one comment with a question

deployment/postgres-init-db/init-aerie.sh
goetzrrGit: I am not sure as we are using `aerie` user in the docker file for the command expansion container.

mattdailis: Feel free to merge - we can dig into our database users and how they work offline.

PR #131, opened 2022-04-11
[AERIE-1782] Scheduler integration tests

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This PR is intended to set up follow-up PRs that grow the scheduling DSL. It establishes tests that:
- Exercise code paths that turn the Typescript goal definition into a scheduler Goal object
- Actually run the scheduler
- Make assertions on the post-conditions of a scheduling run
- Do not require a deployment of aerie (no database, no hasura)

The nice thing about testing at this level is that we can express scheduling as a simple function:

```
(Plan x List[Typescript Goal definitions]) => (Plan x Analysis)
```

Tests written in this way become a specification of the semantics of our scheduling DSL. Future PRs that grow the DSL will be expected to update these tests to demonstrate the new semantics.

~This PR will need to be rebased onto #121 , since there are some MerlinService interface changes there.~ Done.

## Verification

Running `./gradlew :scheduler-server:test` should run these tests, and they should pass.

## Documentation

Some developer documentation might be helpful - perhaps a short checklist of steps to follow when growing the scheduling DSL.

## Future work

- The MerlinService seems a bit too broad, maybe we can split it up into a set of smaller interfaces
- These tests only work with banananation at the moment. It would be good to figure out how to make them work with aerie-lander
- #121 does not re-write unchanged activities, so these tests will need to be updated to take that into account
- It should be possible to make assertions about the properties of the simulation (perhaps by running another simulation after the scheduling run completes)

### DIFF
--- a/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
+++ b/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
@@ -1,0 +1,207 @@
+package gov.nasa.jpl.aerie.scheduler.server.services;
+
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
+import gov.nasa.jpl.aerie.scheduler.server.config.PlanOutputMode;
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalRecord;
+import gov.nasa.jpl.aerie.scheduler.server.models.PlanId;
+import gov.nasa.jpl.aerie.scheduler.server.models.Specification;
+import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
+import gov.nasa.jpl.aerie.scheduler.server.models.Timestamp;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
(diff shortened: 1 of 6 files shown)


scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
```diff
@@ -0,0 +1,371 @@
...
+    agent.schedule(new ScheduleRequest(new SpecificationId(1L), $ -> RevisionData.MatchResult.success()), writer);
+    assertEquals(1, writer.results.size());
+    final var result = writer.results.get(0);
+    assert result instanceof MockResultsProtocolWriter.Result.Success;
+    return new SchedulingRunResults(((MockResultsProtocolWriter.Result.Success) result).results(), this.merlinService.updatedPlan);
+  }
+
+  record SchedulingRunResults(ScheduleResults scheduleResults, Collection<PlannedActivityInstance> updatedPlan) {}
+
+  class MockResultsProtocolWriter implements ResultsProtocol.WriterRole {
```
JoelCourtney: Can you separate the mock classes into their own files like is done in merlin server? It'll make this file a little faster to understand.

JoelCourtney (approved):
Looks good to me

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
mattdailis: Good idea - I went ahead and split them into separate files.

mattdailis:
Rebased onto develop :+1: It required fewer changes than I expected, because `updatePlanActivities` in MerlinService receives both the new plan, and the initial plan.

adrienmaillard (approved):
I like it. One more way of testing the scheduler.

PR #132, opened 2022-04-11
[AERIE-1756] Initial barebones coexistence goal

mattdailis:
* **Tickets addressed:** AERIE-1756
* **Review:** By commit  

## Description

This PR adds an _extremely_ simplistic version of the Coexistence goal. This version basically allows the user to specify "clean-up" activities. For example, this goal makes sure that every time a banana is picked, it is then immediately peeled:

```typescript
  forEach: "PickBanana",
  activityTemplate: ActivityTemplates.PeelBanana({ peelDirection: 'fromStem' })
})
```

The version in this PR hard-codes that the new activity must start at the end time of the old activity.

## Verification

I've added a unit test to `SchedulingDSLCompilationServiceTests.java` as well as an integration test to `SchedulingIntegrationTests.java`. You can run these with `./gradlew scheduler-server:test`.

## Documentation

I've updated the [Scheduling Guide](https://github.com/NASA-AMMOS/aerie/wiki/Scheduling-Guide) with the [coexistence goal](https://github.com/NASA-AMMOS/aerie/wiki/Scheduling-Guide#coexistence-goal-coming-soon).

## Future work

- Flesh out the forEach clause of the CoexistenceGoal to include:
  - Richer ActivityExpressions
  - Resource expressions
  - A way to specify a temporal relationship to the identified windows
- Enforce that the clean-up activity starts causally after the existing activity. The current implementation may have some subtle concurrency issues, since the end time of the existing activity overlaps with the start time of the new activity.

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
@@ -5,6 +5,7 @@ export interface ActivityTemplate {
 
 export enum NodeKind {
   ActivityRecurrenceGoal = 'ActivityRecurrenceGoal',
+  ActivityCoexistenceGoal = 'ActivityCoexistenceGoal',
   GoalAnd = 'GoalAnd',
   GoalOr = 'GoalOr'
 }
@@ -18,14 +19,24 @@ export enum NodeKind {
  */
 export type Goal =
   | ActivityRecurrenceGoal
+  | ActivityCoexistenceGoal
   ;
-// TODO coexistence goal
 // TODO cardinality goal
 
 export interface ActivityRecurrenceGoal {
   kind: NodeKind.ActivityRecurrenceGoal,
   activityTemplate: ActivityTemplate,
   interval: number,
+}
+
+export interface ActivityExpression {
+  type: string
+}
+
+export interface ActivityCoexistenceGoal {
+  kind: NodeKind.ActivityCoexistenceGoal,
+  activityTemplate: ActivityTemplate,
+  forEach: ActivityExpression
 }
 
 export type GoalSpecifier =
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
@@ -3,6 +3,12 @@ export interface ActivityTemplate {
   args: {[key: string]: any},
 }
 
+export enum NodeKind {
+  ActivityRecurrenceGoal = 'ActivityRecurrenceGoal',
+  GoalAnd = 'GoalAnd',
+  GoalOr = 'GoalOr'
+}
+
 /**
  * Goal
  *
@@ -17,7 +23,7 @@ export type Goal =
 // TODO cardinality goal
 
 export interface ActivityRecurrenceGoal {
-  kind: 'ActivityRecurrenceGoal',
+  kind: NodeKind.ActivityRecurrenceGoal,
   activityTemplate: ActivityTemplate,
   interval: number,
 }
@@ -39,12 +45,12 @@ export type GoalComposition =
   ;
 
 export interface GoalAnd {
-  kind: 'GoalAnd',
+  kind: NodeKind.GoalAnd,
   goals: GoalSpecifier[],
 }
 
 export interface GoalOr {
-  kind: 'GoalOr',
+  kind: NodeKind.GoalOr,
   goals: GoalSpecifier[],
 }
 
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -1,6 +1,9 @@
 import * as AST from './scheduler-ast.js';
+import { ActivityType } from "./mission-model-generated-code";
 
 interface ActivityRecurrenceGoal extends Goal {}
+interface ActivityCoexistenceGoal extends Goal {}
+
 export class Goal {
   private readonly goalSpecifier: AST.GoalSpecifier;
 
@@ -43,6 +46,15 @@ export class Goal {
       interval: opts.interval,
     });
   }
+  public static CoexistenceGoal(opts: { activityTemplate: ActivityTemplate, forEach: ActivityType }): ActivityCoexistenceGoal {
+    return Goal.new({
+      kind: AST.NodeKind.ActivityCoexistenceGoal,
+      activityTemplate: opts.activityTemplate,
+      forEach: {
+        type: opts.forEach
+      },
+    });
+  }
 }
 
 declare global {
@@ -52,6 +64,8 @@ declare global {
     public or(...others: Goal[]): Goal
 
     public static ActivityRecurrenceGoal(opts: { activityTemplate: ActivityTemplate, interval: Duration }): ActivityRecurrenceGoal
+
+    public static CoexistenceGoal(opts: { activityTemplate: ActivityTemplate, forEach: ActivityType }): ActivityCoexistenceGoal
   }
   type Duration = number;
   type Double = number;
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -1,4 +1,4 @@
-import type * as AST from './scheduler-ast.js';
+import * as AST from './scheduler-ast.js';
 
 interface ActivityRecurrenceGoal extends Goal {}
 export class Goal {
@@ -18,7 +18,7 @@ export class Goal {
 
   public and(...others: Goal[]): Goal {
     return Goal.new({
-      kind: 'GoalAnd',
+      kind: AST.NodeKind.GoalAnd,
       goals: [
         this.goalSpecifier,
         ...others.map(other => other.goalSpecifier),
@@ -28,7 +28,7 @@ export class Goal {
 
   public or(...others: Goal[]): Goal {
     return Goal.new({
-      kind: 'GoalOr',
+      kind: AST.NodeKind.GoalOr,
       goals: [
         this.goalSpecifier,
         ...others.map(other => other.goalSpecifier),
@@ -38,7 +38,7 @@ export class Goal {
 
   public static ActivityRecurrenceGoal(opts: { activityTemplate: ActivityTemplate, interval: Duration }): ActivityRecurrenceGoal {
     return Goal.new({
-      kind: 'ActivityRecurrenceGoal',
+      kind: AST.NodeKind.ActivityRecurrenceGoal,
       activityTemplate: opts.activityTemplate,
       interval: opts.interval,
     });
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
@@ -5,6 +5,8 @@ import gov.nasa.jpl.aerie.merlin.driver.MissionModel;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModelLoader;
 import gov.nasa.jpl.aerie.merlin.protocol.model.SchedulerModel;
 import gov.nasa.jpl.aerie.merlin.protocol.model.SchedulerPlugin;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.constraints.scheduling.GlobalConstraint;
 import gov.nasa.jpl.aerie.scheduler.goals.Goal;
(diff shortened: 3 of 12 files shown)


scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -46,6 +48,15 @@ export class Goal {
       interval: opts.interval,
     });
   }
+  public static CoexistenceGoal(opts: { activityTemplate: ActivityTemplate, forEach: string }): ActivityCoexistenceGoal {
```
dyst5422: Lets codegen out the activity names into their own map like ActivityTemplates so we enforce that the activity type exists.

Strings in the eDSL are categorically bad UX UNLESS its truly is a freeform string

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
```diff
@@ -23,6 +23,16 @@ export interface ActivityRecurrenceGoal {
   interval: number,
 }
 
+export interface ActivityExpression {
+  type: string
```
dyst5422, line 33: We should codegen out the activity types so its enforced that we have correct values here

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -46,6 +48,15 @@ export class Goal {
...
+    return Goal.new({
+      kind: 'ActivityCoexistenceGoal',
```
dyst5422: I think we should have an enum of these node kinds at this point.

```
enum NodeKind {
  ActivityCoexistenceGoal: 'ActivityCoexistenceGoal'
  ActivityRecurrenceGoal: 'ActivityRecurrenceGoal',
  ...etc
}
```

and then `kind: NodeKind.ActivityCoexistenceGoal`

dyst5422 (changes requested):

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
mattdailis: Completely agree, thanks for the quick review.

I've updated the type of `forEach` to be `ActivityType`, which is codegen'd to be a union of all the names of the activities:

```typescript
type ActivityType =
  | "PeelBanana"
  | "BiteBanana";

export const ActivityTypes = {
  PeelBanana: "PeelBanana",
  BiteBanana: "BiteBanana"
}

declare global {
  var ActivityTypes: {
    PeelBanana: ActivityType,
    BiteBanana: ActivityType
  }
}
```

Does that seem like a viable solution? I think this means that the exact string will still be accepted (users are not forced to use `ActivityTypes.`). Not sure if that's a good or a bad thing.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
mattdailis: We aren't generating any part of the AST currently, just the DSL. My understanding is that the AST is static, and cannot depend on a particular mission model - in which case this field would remain a string.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
mattdailis: We certainly can do this for aesthetics (and I suppose to give us a nice list of all kinds) - but it won't give us any additional compile time guarantees, since `Goal.new` takes a `GoalSpecifier`, and all of the variants of GoalSpecifier specify that their `kind` field needs to be a particular string.

I'd be happy to do this - but I'd like to hear some more reasoning behind your "should" :+1:

dyst5422: Yeah, its more of an organizational thing -> easier to ensure in the future and gives some nice comparison behavior if we even need to manipulate the AST. As you stated too, having a single list for all of the node kinds makes it easier to verify the Java-Typescript interface as well

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
dyst5422: Good point, so I suppose its just in the fluent dsl where we need to be more specific and make it a type that is supplied by the model specific piece

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
dyst5422: I think we can make an it an enum to ease user UX. A long list of strings is going to be bad for UX as they won't be able to look through it all, versus ActivityTypes.<autocomplete> is nice to work with


BTW ...
```
export const ActivityTypes =
  PeelBanana: "PeelBanana"
  BiteBanana: "BiteBanana";
```
isn't valid typescript. We'd need it to be an enum or an object.

mattdailis: Good catch - I wrote that part by hand instead of copying from IntelliJ, oops. Just went back and edited the snippet.

Currently mulling over what the empty interface would look like, and what we would gain from a tagged enum vs a regular enum.

dyst5422: Sorry the term is String Enum. String enums have strings as values instead of numbers -> so what is actually deserialized has the proper string names instead of needing to map back from numerics to strings.

I don't think the empty interface approach makes sense (this would be similar to what we do for commands

```
interface PREHEAT_OVEN extends Command {}
```
where its merely an extension of Command that retains a type-level name vs the type aliases that `type PREHEAT_OVER` would do.

I definitely think string enums are the way here

mattdailis: Thanks for the explanation - I've updated ActivityType to be a String enum. (I think it's kind of neat that `ActivityTypes` just becomes an alias for `ActivityType`).

I made the change in a new commit for now - I'll go squash it with the previous commit next time I rebase.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
mattdailis: Turns out that importing from the AST file is broken at runtime 😅  I'm getting an `Unable to resolve dependency "./scheduler-ast.js"` error. My guess is that because we were only using the types from the AST, and never values, the import statement was skipped at runtime. The NodeKind enum introduces a runtime dependency on the AST file, and the import fails.

I'll postpone adding a `NodeKind` to the AST until [[AERIE-1767]: Perform Scheduling DSL Type checking on backend](https://jira.jpl.nasa.gov/browse/AERIE-1767) is ready, since your work will allow imports to actually work at runtime :+1:

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -49,6 +52,7 @@ void testCodeGen() {
                  + "      return { activityType: 'PeelBanana', args };\n"
                  + "    },\n"
                  + "}\n"
+                 + "export const ActivityTypes = ActivityType;\n"
```
dyst5422: In fact, we don't even need to export a const here, we can just export then enum directly

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -58,6 +62,9 @@ void testCodeGen() {
                  + "        peelDirection: (\"fromTip\" | \"fromStem\")\n"
                  + "      }) => PeelBanana\n"
                  + "  }\n"
+                 + "  var ActivityTypes: {\n"
```
dyst5422: We can define the enum here too using the same syntax I believe instead of patching it is as an object

`enum ActivityType {
...
}`

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
mattdailis: True - I guess I was trying to mirror the pattern we used for `ActivityTemplates`, where the plural is used for the name of the map, and the singular is used for the name of the type. Not sure how important that consistency is though.

dyst5422: Fair point! I agree

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
dyst5422: You're comment about plurality in the naming is coming to mind...

mattdailis: Yeah, I'm wondering if there's a clean way to consolidate the `const` declaration with the `var` declaration for the same identifier... drawing blanks at the moment.

mattdailis: Oh wait, would `var ActivityTypes: ActivityType` work? I'll go give that a try.

mattdailis: Ah, the incantation `var ActivityTypes: typeof ActivityType` seems to work :+1:

mattdailis: Now I'm wondering if we could do something similar with `ActivityTemplates`:

```typescript
const ActivityTemplatesAlias = ActivityTemplates;
declare global {
  var ActivityTemplates: typeof ActivityTemplatesAlias
}
```

> Note: the alias is needed because simply writing `var ActivityTemplates: typeof ActivityTemplates` gives the error: `'ActivityTemplates' is referenced directly or indirectly in its own type annotation`.

Pro: "single source of truth"
Con: I could conceive of us wanting to export the types without the implementations (e.g. to reduce file size). Perhaps having them separate now would make that easier to do in the future.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
dyst5422: Now that its merged, is this something we want to add in  a commit to this PR?

scheduler-server/scheduling-dsl-compiler/tsconfig.json
```diff
@@ -95,6 +95,9 @@
 
     /* Completeness */
     // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
-    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+    "skipLibCheck": true,                                 /* Skip type checking all .d.ts files. */
   },
+  "exclude": [
+    "./src/libs/scheduler-edsl-fluent-api.ts"  // This file is used at runtime, does not need to be compiled.
```
dyst5422: This may not need to be compiled, but we do need it for the type assertion in main.ts with the typechecking now happening so we need it included. For the command expansion service, we have an additional build step to copy the ts files used at runtime to the build folder. I can add that here too.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -46,21 +46,36 @@ public static String generateTypescriptTypesFromMissionModel(final MissionModelT
...
     result.add("};");
     result.add("/** End Codegen */");
     return joinLines(result);
   }
 
+  private static String generateActivityTypeEnum(final ArrayList<ActivityTypeCode> activityTypeCodes) {
+    final var result = new ArrayList<String>();
+    result.add("enum ActivityType {");
+    for (final var activityTypeCode : activityTypeCodes) {
+      result.add(indent("%s = \"%s\"".formatted(activityTypeCode.activityTypeName(), activityTypeCode.activityTypeName())));
```
dyst5422: Fairly certain we need commas at the end of the lines here.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationServiceTest.java
dyst5422: I doubt we will reach the point of needing to reduce file size on this, and if we want to be able to split out the definitions from the implementations (which we very well may way to do) we'd need to split it into different files (or at least be able to generate a separate declarations file) so I don't think we should worry about it in this PR.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -243,6 +245,15 @@ private PlanComponents loadInitialPlan(final PlanMetadata planMetadata, final Pr
         act.setArguments(activity.arguments());
```
dyst5422, line 245: We might want to have conversions in both directions for these two classes so we aren't doing this ad-hoc all over the code base. Not necessary for this PR.

dyst5422 (approved):

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
mattdailis: I added `NodeKind` in the latest commit :+1:

mattdailis: Hmm, this seems to be causing some import errors, looking into it now

mattdailis:
In the latest force push, I rebased onto develop, and re-ordered the commits to move all of the refactoring commits before the functionality-changing commits.

Here is the current grouping of commits:
#### Update libraries (there is a bugfix in 0.2.6 that we need)
- Bump aerie-ts-user-code-runner to 0.2.6

#### Pure Refactoring
- Add NodeKind enum
- Use sum parser for goal types

#### Bug fix
- Set ActivityInstance duration if it is controllable

#### New functionality
- Add another sample activity to TypescriptCodeGenerationServiceTest
- Generate ActivityType enum and ActivityTypes global
- Add basic CoexistenceGoal to EDSL
- Add parser for Coexistence goal
- Instantiate CoexistenceGoal from goal definition
- Add test for Coexistence Goal

adrienmaillard (approved):

PR #134, opened 2022-04-11
[AERIE-1767] Typechecking Scheduling Goal EDSL

dyst5422:
Adds in the aerie-ts-user-code-runner as a dependency and uses it for
typechecking and execution of user code

* **Tickets addressed:** AERIE-1767
* **Review:** By file  

## Description
We introduced a different paradigm for returning information from the typescript executor. We took the approach of Rust and had an error case, which is handled and represents errors in the typechecking or execution and is represented as a result with errors in Java, and panics, which indicates an issue somewhere in our code that needs to be fixed and which propagates a throw in Java. Then most of the related file updates are to manage this new interfacing.

Much of the work in this tickets was running into various unhandled type error cases scoped to the execution handle in the code runner that needed to be addressed.

There are a couple of changes to our AST/EDSL files and codegen that are aimed at ensuring the files can execute in the code runner environment (adding in an import for the ActivityTemplate, adding in the global assignment of globally available values, and a couple of syntactic fixes (insert commas)).

## Verification

Tests were modified and run to ensure behavior, but a lot of the error handling is in the code executor itself, which had a handful of test cases added

## Documentation
No documentation should need to be added

## Future work
We need to have a way to record errors during type-checking/execution in the database for retrieval later, so adding a field for that in the DB and finding a suitable workflow for it is necessary future work.

As a side note, I think the more flexing of the DSLs we can get by users early on, the better off we are going to be because there are undoubtedly going to be additional unhandled cases we discover and need to address.


dyst5422:
Rebased

dyst5422:
First toe dip into the Java side of things, so feel free to tear it apart!

.gitignore
```diff
@@ -3,6 +3,7 @@
 
 ## Cache ##
 .DS_Store
+*.tsbuildinfo
```
mattdailis, line 6: Where is the `.tsbuildinfo` file coming from? I remember I explicitly disabled incremental compilation in the `scheduling-dsl-compiler` tsconfig file:

https://github.com/NASA-AMMOS/aerie/blob/5ba6ced4ff9bac5db0db5915006a8f63ad575d9f/scheduler-server/scheduling-dsl-compiler/tsconfig.json#L6

scheduler-server/scheduling-dsl-compiler/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "author": "",
   "dependencies": {
+    "@nasa-jpl/aerie-ts-user-code-runner": "^0.2.5",
```
mattdailis, line 13: I think we might be better off setting an exact version number here and making PRs to bump the version number as necessary. I'm not sure we can guarantee backwards compatibility at this point. What do you think?

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -1,38 +1,61 @@
...
+import { UserCodeRunner } from '@nasa-jpl/aerie-ts-user-code-runner';
+import type { Goal } from './libs/scheduler-edsl-fluent-api.js'
+
+const codeRunner = new UserCodeRunner();
+const schedulerEDSL = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-edsl-fluent-api.ts`, 'utf8');
+const schedulerAST = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-ast.ts`, 'utf8');
 
 process.on('uncaughtException', (err) => {
-  console.error("uncaughtException")
+  console.error('uncaughtException');
```
mattdailis, line 11: Is there a prettier/editorconfig file you're using for this formatting? Is there something we should add to the developer docs to make sure we have the same one? I'd like to avoid diffs that change quotes from one kind to another.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SpecificationLoadException.java
```diff
@@ -1,9 +1,14 @@
...
 
+import java.util.List;
+
 public final class SpecificationLoadException extends Exception {
-  public SpecificationLoadException(final SpecificationId specificationId, final String reason) {
-    super("Failed to load specification with id `" + specificationId.id() + "` reason: " + reason);
+  public List<SchedulingCompilationError.UserCodeError> errors;
+  public SpecificationLoadException(final SpecificationId specificationId, final List<SchedulingCompilationError.UserCodeError> errors) {
+    super("Failed to load specification with id `" + specificationId.id());
+    this.errors = errors;
```
mattdailis, line 12: I paired with Dylan on this part, so I'd like to get a third opinion here (maybe @pcrosemurgy ?) - is storing a field on an exception class kosher? It's not used at all in the message - but it is used in the catch block here:

https://github.com/NASA-AMMOS/aerie/blob/5ba6ced4ff9bac5db0db5915006a8f63ad575d9f/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java#L139-L141

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -91,8 +92,18 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
       ensurePlanRevisionMatch(specification, planMetadata.planRev());
       //create scheduler problem seeded with initial plan
       final var schedulerMissionModel = loadMissionModel(planMetadata);
-      final var planningHorizon = new PlanningHorizon(Time.fromInstant(specification.horizonStartTimestamp().toInstant()), Time.fromInstant(specification.horizonEndTimestamp().toInstant())) ;
-      final var problem = new Problem(schedulerMissionModel.missionModel(), planningHorizon, new SimulationFacade(planningHorizon, schedulerMissionModel.missionModel()), schedulerMissionModel.schedulerModel());
+      final var planningHorizon = new PlanningHorizon(
```
mattdailis, line 97: I'm not a fan of unrelated formatting changes cluttering the diff of a PR - that said, I do think you have improved the formatting here.

@patkenneally what's our policy on things like this?

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -31,8 +31,8 @@ class SchedulingDSLCompilationServiceTests {
   private static final PlanId PLAN_ID = new PlanId(1L);
   SchedulingDSLCompilationService schedulingDSLCompilationService;
 
-  @BeforeAll
-  void setUp() throws SchedulingDSLCompilationService.SchedulingDSLCompilationException, IOException {
+  @BeforeEach
```
mattdailis: I'm curious about the motivation for `@BeforeEach` vs `@BeforeAll`

If we do go with `@BeforeEach` - I think we can remove the `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` annotation on this class.

For context: JUnit by default creates a new instance of the test class for each test. The lifecycle annotation forces it to re-use the same instance across all tests - which is necessary if your `@BeforeAll` method sets any fields on that instance.

.gitignore
dyst5422: If one wanted to turn on incremental build for faster dev cycles, we wouldn't want it in git, so its more cautionary than anything

scheduler-server/scheduling-dsl-compiler/package.json
dyst5422: I have been following the pre-1.0 release semver -> no interface breaking changes without a 0.3 bump. And npm follows the pre-1.0 semantics of only allowing fix version bumping. So I think this follows the nature of semver sufficiently to allow us to use this. Thoughts?

scheduler-server/scheduling-dsl-compiler/src/main.ts
dyst5422: We should just add a config file for it. I can do that in this PR if you'd like.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
dyst5422: I was seeing issues in testing where multiple tests running in parallel were hitting the same executor node instance - which we have explicitly not supported at this point so they would hang waiting for a response that would never come. So this was to ensure that each had their own instance for the test to avoid collisions. I'm open to other solutions, but thought parallelizing the node process out of scope for this PR.

scheduler-server/scheduling-dsl-compiler/package.json
mattdailis: I see - I did not understand the difference between `>` and `^`. This should be fine as-is then :+1: We'll just need to be careful with version numbers on the user-code-runner side.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
```diff
@@ -89,21 +91,23 @@ private static GoalCompilationResult compileGoalDefinition(
...
-    } catch (SchedulingDSLCompilationService.SchedulingDSLCompilationException e) {
-      return new GoalCompilationResult.Failure(e.getMessage());
+    final var goalCompilationResult = schedulingDSLCompilationService.compileSchedulingGoalDSL(
+        planId,
+        pgGoal.definition(),
+        pgGoal.name());
+
+    if (goalCompilationResult instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Error g) {
+      return new GoalCompilationResult.Failure(g.errors());
+    } else if (goalCompilationResult instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success g) {
```
goetzrrGit, line 101: I was thinking the below is really wordy. Thoughts?
 `SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success` 

to

`SchedulingDSLCompilationService.results.Success`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SpecificationLoadException.java
goetzrrGit: I think storing a field is kosher as there is a reason you are creating a custom exception class, to begin with. The default doesn't have what you need.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
goetzrrGit: I was thinking we can set up a standard formating config file for java and javascript and have a PR just for formatting changes.

ex.

I know in javascript you can 'format on save' and open all the javascript files and save them which will format them. Then make a PR with those changes. It will be a simple approve because no logic was changed just the format.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
dyst5422: I'm hesitant about having it as just `Result` because I think that has some very specific connotations w.r.t. shape and behavior. IE a standard `Result` type is more of the style `Result<SuccessType, ErrorType>` with associated methods to act on it. I'm entirely open to changing it though if there is consensus otherwise (I might be polluted by my affection for the Rust Result type here)

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
mattdailis: I see - if that's the case I think your approach is fine. A little bit of searching on the web seems to suggest that JUnit doesn't run tests in parallel by default though, so I'd like to understand:
1. are our tests being run in parallel?
2. if not, can we understand the behavior you observed in your testing?

I have a hunch that there might be an actual bug that we're sidestepping by re-starting the subprocess between tests. I'll try to reproduce on my end as well

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
goetzrrGit: That makes sense that others might see the `result` type being `Result<SuccessType, ErrorType>`

Maybe `SchedulingDSLCompilationService.CompilationResult.Success` but this is in the weeds and was just an idea I had. If no one and any strong feelings you can ignore this. :)

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -1,38 +1,61 @@
...
+        [
+          ts.createSourceFile('scheduler-ast.ts', schedulerAST, ts.ScriptTarget.ESNext),
+          ts.createSourceFile('scheduler-edsl-fluent-api.ts', schedulerEDSL, ts.ScriptTarget.ESNext),
+          ts.createSourceFile('mission-model-generated-code.ts', missionModelGeneratedCode, ts.ScriptTarget.ESNext),
+        ],
+    );
+
+    if (result.isErr()) {
+      process.stdout.write('error\n' + JSON.stringify(result.unwrapErr().map(err => err.toJSON())) + '\n');
+      return;
```
mattdailis, line 37: Aha. Here's the bug. We need to re-register the `data` callback here. Otherwise, once we've returned an error, we're no longer listening to standard input.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
mattdailis: See comment here: https://github.com/NASA-AMMOS/aerie/pull/134/files#r848808391

scheduler-server/scheduling-dsl-compiler/src/main.ts
dyst5422: Good catch! And debugging!

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
dyst5422: Bad assumption on my part that most test runners run in parallel by default - spoiled by JS tooling I guess

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
mattdailis: To paraphrase Pat's statement from standup: 
> to the extent possible, formatting changes should not be mixed with functional changes in the same commit.

So it's okay to make this change, but it should be in its own commit.

My own thoughts here: changes fall into one of two (ish) categories:
1. Changes that affect the behavior of the program
2. Refactoring / Reformatting

From a reviewer's perspective, reviewing a PR is much more difficult when the two kinds of changes are mixed together. Keeping them separate allows the reviewer to focus on one of these things at a time:
1. Checking that the new behavior of the program is desirable
2. Checking that the changes did not change the behavior of the program

Expecting a reviewer to do a mix of the two at the same time puts a lot of cognitive load on the reviewer, and results in a more difficult, and lower quality review process.

Now, this PR is far from egregious in that sense - the changes are fairly limited and not too hard to review (but since I was around for the writing of a lot of these changes, my opinion is skewed)

mattdailis: @goetzrrGit I do like the idea of setting up an autoformatter for Java, but I'd like to get buy-in from the team at large first. I do not have strong opinions about what the correct format is, but I'm sure other people do.

patkenneally: @mattdailis As you've said above. 

For this PR and generally for Aerie's repositories; my preference here is that a quick interactive rebase be done separating out the formatting changes from the functional ones. Also, from the initial look through (I'm coming back for a proper review after my next meeting), there are multiple stages in the development of this work which might be better communicated in separate commits. Breaking out the changes into perhaps 3 (maybe 4) stages which walk the reviewer through each change to the final functionality, would greatly help reviewers. 

I want to remind us all of the standard and reasons for the way we build up PRs. The following are linked from our contributing.md. 
[How to write a good commit message](https://chris.beams.io/posts/git-commit/)
[Telling Stories Through Your Commits](https://blog.mocoso.co.uk/talks/2015/01/12/telling-stories-through-your-commits/)
[How to Make Your Code Reviewer Fall in Love with You](https://mtlynch.io/code-review-love/)

dyst5422: Broken up into a small refactor commit, a commit for adding the common code execution library, and a commit for handling user code errors differently than exceptions (panics).

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
dyst5422: Addressed this with the break up of the commits. Thanks @mattdailis

dyst5422:
rebased against develop

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
dyst5422: You're making me feel that we need a Result utility type for Java. But I know that java is very throw - encouraging, so don't feel I have sufficient background to make that suggestion. 

Personally, I've very throw-averse (probably from too much time with Rust)

Twisol: > You're making me feel that we need a Result utility type for Java.

We kinda sorta have one in `gov.nasa.jpl.aerie.merlin.framework.Result`, but it was developed to meet model use-cases; I'd argue against just having other components import the modeling `framework`. Anyway, it could use some touching-up since it was created before we had `sealed` interface -- but we don't yet have pattern-matching in `switch` to seal the deal.

Twisol: > Personally, I've very throw-averse (probably from too much time with Rust)

Personally, I agree with you, but Java makes it pretty painful to avoid exceptions. I tend to categorize exceptions as one of three kinds: caller-callee contract exceptions, component integration exceptions, and we-didn't-code-for-this-scenario exceptions.

Contract exceptions occur when the caller didn't respect some requirements that the callee needs for correct operation, e.g. indices out of bounds and whatnot. These ought to be checked exceptions, which as far as the calling convention is concerned makes them like additional variants of a returned sum type (that you happen to match on using `try-catch`). Sometimes it can make sense to prevent this kind of exception from capturing call stack information (since that's somewhat performance-sensitive); you can do this by passing the appropriate `bool` flags to the exception's superclass in the constructor.

Integration exceptions occur when code can't abide by its contract due to reasons outside its control, i.e. because some other component it has been integrated with (such as via dependency injection) has failed. These especially happen when the callee's detailed contract cannot be communicated to the caller, e.g. because it's hidden behind an abstract interface that suppresses some of those details. Since the caller can't be expected to abide by requirements it can't be aware of, responsibility falls to the agent that integrated these components. (This is a very Erlang supervision-like perspective on failure.)

"We-didn't-code-for-this-scenario exceptions" are as they say: some design assumptions have been violated, so the program simply can't have a strategy for handling this in the first place. These are like top-level integration exceptions in that the developer is at fault for wiring together these components, but it's a little bit broader of a category. It's most like Rust's `panic`.

mattdailis (approved):
Looks good to me, thanks for splitting out those commits :+1:

patkenneally (approved):
Thanks for splitting this out. It really helps guide my attention to what we are reviewing at each step. Additionally, because I'm still building my TypeScript muscle :)

dyst5422:
> Thanks for splitting this out. It really helps guide my attention to what we are reviewing at each step. Additionally, because I'm still building my TypeScript muscle :) 

Still refining my commit breakup skills. Took me walking away to figure out where the logical break should be. Going to rebase and merge in a few here.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
patkenneally: This is exactly the write up that I went hunting for because I thought @Twisol had written it before. Copied and saved this now for prosperity. :)

Twisol: I did write it somewhere, a long time ago in another PR, but good luck finding it ^_^;

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -128,13 +137,13 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
                                                 loadedPlanComponents.merlinPlan(),
                                                 solutionPlan);
       //collect results and notify subscribers of success
-      final var results = collectResults(solutionPlan,instancesToIds, goals);
+      final var results = collectResults(solutionPlan, instancesToIds, goals);
       writer.succeedWith(results);
-    } catch (final ResultsProtocolFailure | NoSuchSpecificationException | SpecificationLoadException e) {
+    } catch (final SpecificationLoadException e) {
+      //unwrap failure message from any anticipated exceptions and forward to subscribers
+      writer.failWith(e.getMessage() + SchedulingCompilationError.schedulingErrorJsonP.unparse(e.errors).toString());
```
pcrosemurgy, line 144: @dyst5422 Sorry for jumping on this post-merge, but I think that instead of concatenating 2 strings (without any delimiter between them) and serializing 90% of the exception as JSON we should maybe just go the extra 10% and make a serializer for `SpecificationLoadException` as whole.

pcrosemurgy: Maybe this is just a very temp. solution? If so we should probably add a `TODO` to note that the data model is not fleshed out enough for us to worry about serializing this exception correctly yet.

pcrosemurgy: When rebasing with this PRs changes in my other branch I made this change:
https://github.com/NASA-AMMOS/aerie/pull/140/files#diff-e8610735ea15be17f6b2f0c58f2204cb97c8bcdad4946a9e5ccbb6884650fb77R144-R148
Let me know what you think :)

PR #136, opened 2022-04-12
Setup default prettier format and formatted Command Expansion Server

goetzrrGit:
## Description
I updated the prettier format configuration and formatted all of the Command Expansion Server. This should help with PR where formatting adds clutter to the PR review. On Vscode I am using 

Name: Prettier - Code formatter
Id: esbenp.prettier-vscode
Description: Code formatter using prettier
Version: 9.5.0
Publisher: Prettier
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

## Verification
No code was changed just the formatting.


goetzrrGit:
Going to fix a weird merge issue as there should only be 1 commit not 5.

command-expansion-server/.prettierrc
```diff
@@ -1,9 +1,11 @@
...
-  "parser": "typescript",
+  "trailingComma": "all",
   "singleQuote": true,
-  "semi": true,
   "arrowParens": "avoid",
+  "semi": true,
   "printWidth": 120,
-  "trailingComma": "all"
+  "tabWidth": 2,
+  "useTabs": true,
```
dyst5422: [Why Would You Say Something So Controversial Yet So Brave?](https://c.tenor.com/EKt2Luq-ivIAAAAC/why-would-you-say-something-so-controversial-yet-so-brave.gif)

I am a fan of tabs due to its improved behavior in all environments (someone can set their own indent width), but good luck on this one in general!

camargo: *gets out the popcorn*

goetzrrGit: 'spaces vs tabs' !!!

pcrosemurgy: ### Poll
❤️ for tabs
🚀 for spaces

patkenneally: Not to stifle a hearty discussion, but the majority of Aerie repos use spaces. Let's use spaces. :)

dyst5422: I don't really care, but the (I think good) argument for tabs is that you can have people with 20/20 vision working on the same project as someone with shite vision as they can configure indentation in their editor to their liking.

I've used both and really dont care, we're all pressing tab to indent anyways.

goetzrrGit: For someone with one bad eye, I do agree with Dylan, but the idea here is to have settings that the developer shouldn't change unless they really have to. That way we don't have formatting noise in our PR. I can make a poll in aerie slack and just use those results for the setting

goetzrrGit: We got spaces now :)

command-expansion-server/.prettierrc
```diff
@@ -1,9 +1,11 @@
...
-  "parser": "typescript",
+  "trailingComma": "all",
   "singleQuote": true,
-  "semi": true,
   "arrowParens": "avoid",
+  "semi": true,
   "printWidth": 120,
-  "trailingComma": "all"
+  "tabWidth": 2,
+  "insertSpaces": true,
```
camargo: Is `insertSpaces` a Prettier option? I can't find it [documented](https://prettier.io/docs/en/options.html). The default behavior of Prettier is `spaces` so I think it can be excluded.

goetzrrGit: Oh, I must have found an older document as it was `editor.insertSpaces` but yeah the document you linked doesn't have it. I removed it.

dyst5422 (approved):

mattdailis (approved):

camargo (approved):

PR #137, opened 2022-04-13
[AERIE-1783] Add commanding container to the deployment docker compose

patkenneally:
* **Tickets addressed:** AERIE-1783
* **Review:** By commit  

## Description
Added commanding container to deployment docker compose.

## Verification
Ran `docker compose up` and uploaded a command dictionary and received a dict ID of 1 and no errors. Used upload query as defined in the test file https://github.com/NASA-AMMOS/aerie/blob/fca42c28f23488d97506dd798070d879ae284f3b/command-expansion-server/test/utils/CommandDictionary.ts#L3

## Documentation
None required and no new documentation needed.

## Future work
None

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -90,13 +90,13 @@ services:
       SCHEDULER_DB_USER: "aerie"
       SCHEDULER_LOCAL_STORE: /usr/src/app/scheduler_file_store
       SCHEDULER_OUTPUT_MODE: UpdateInputPlanWithNewActivities
-      SCHEDULER_PORT: 27193
+      SCHEDULER_PORT: 27185
       SCHEDULER_RULES_JAR: /usr/src/app/merlin_file_store/scheduler_rules.jar
       JAVA_OPTS: >
         -Dorg.slf4j.simpleLogger.defaultLogLevel=WARN
         -Dorg.slf4j.simpleLogger.logFile=System.err
     image: "${REPOSITORY_DOCKER_URL}/aerie-scheduler:${DOCKER_TAG}"
-    ports: ["27193:27193"]
+    ports: ["27185:27185"]
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -1,5 +1,24 @@
 version: "3.7"
 services:
+  aerie_commanding:
+    container_name: aerie_commanding
+    depends_on: ["postgres"]
+    environment:
+      LOG_FILE: console
+      LOG_LEVEL: warn
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      COMMANDING_SERVER_PORT: 27184
+      COMMANDING_DB: aerie_commanding
+      COMMANDING_DB_PASSWORD: aerie
+      COMMANDING_DB_PORT: 5432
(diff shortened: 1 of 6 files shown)


mattdailis (approved):
Compared with development docker-compose.yml. The changes seem reasonable to me
- [x] log level reduced from debug to warn
- [x] image points at docker repository instead of local filesystem
- [x] don't mount `./command-expansion-server` directly into the container

deployment/docker-compose.yml
```diff
@@ -1,5 +1,24 @@
 version: "3.7"
 services:
+  aerie_commanding:
+    container_name: aerie_commanding
+    depends_on: ["postgres"]
+    environment:
+      LOG_FILE: console
+      LOG_LEVEL: warn
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      COMMANDING_SERVER_PORT: 3000
```
camargo: In a near-future commit we should change the port to match more closely with the other servers ports

camargo (approved):

dyst5422 (approved):

deployment/docker-compose.yml
patkenneally: Is there anything that would prohibit us doing that as part of this PR? If not, I'll go ahead and make the change.

mattdailis:
Should we keep the e2e-test file in sync with the port number changes? https://github.com/NASA-AMMOS/aerie/blob/916518159601466a3dddd837023a8f9a9f075b65/e2e-tests/docker-compose-test.yml#L16

I imagine it doesn't really matter what the port numbers are, as long as they're wired together correctly, but I also think it's worth keeping them consistent.

patkenneally:
I've updated the commanding port to be 27184 and the scheduler port to be 27185. The merlin server port is 27183 so this blocks things together. This seems easy.

camargo:
@mattdailis Yes I think we should keep it consistent across the code base. 
Even the hardcoded default fallbacks should be to our default specified ports [here](https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/app.ts#L25) and [here](https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/env.ts#L18).

mattdailis:
> @mattdailis Yes I think we should keep it consistent across the code base. Even the hardcoded default fallbacks should be to our default specified ports [here](https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/app.ts#L25) and [here](https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/env.ts#L18).

Here's one more spot where it's hard-coded: the SchedulerAppDriver https://github.com/NASA-AMMOS/aerie/blob/565d33c0b615e925229b97f198fd9fc183730bfd/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/SchedulerAppDriver.java#L150

mattdailis (approved):

camargo:
We should update the port docs in the Wiki as well: https://github.com/NASA-AMMOS/aerie/wiki/Product-Guide#tcp-port-requirements

camargo (approved):

patkenneally:
hahaha on it. Doing deploy docs as well.

deployment/docker-compose.yml
patkenneally: Change made.

PR #139, opened 2022-04-13
[DEVOPS] Add docker config. validation step to `test` workflow

pcrosemurgy:
* **Tickets addressed:** None
* **Review:** By commit  

## Description
Adds docker config. validation step to `test` GH Action.

## Verification
Tested with an intentionally erroneous config.:
```
ERROR: The Compose file '././docker-compose.yml' is invalid because:
services.aerie_merlin.build.args contains an invalid type, it should be an object, or an array
```

Also tested with a temp. commit to this PR: https://github.com/NASA-AMMOS/aerie/runs/6012834575?check_suite_focus=true

## Documentation
None.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,14 @@ jobs:
         run: ./gradlew classes
       - name: Assemble
         run: ./gradlew assemble
+      - name: Validate Docker Configurations
+        # `docker-compose` and `docker compose` have different "leniencies"
+        run: |
+          find . -name docker-compose\*.yml | while read -r file; do \
+            echo "$file"; \
+            docker-compose -f "$file" config -q && \
+            docker compose -f "$file" config -q; \
+          done
       - name: Start Services
         run: |
           docker compose -f ./e2e-tests/docker-compose-test.yml up -d --build


.github/workflows/test.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         run: ./gradlew classes
       - name: Assemble
         run: ./gradlew assemble
+      - name: Validate Docker Configurations
+        run: find . -name docker-compose\*.yml | while read -r file; do docker-compose -f "$file" config -q; done
```
pcrosemurgy: Interestingly, `find . -name docker-compose\*.yml -exec docker compose -f {} config -q \;` just returned:
```
services.aerie_merlin.build.args must be a mapping
```
Instead of:
```
ERROR: The Compose file '././docker-compose.yml' is invalid because:
services.aerie_merlin.build.args contains an invalid type, it should be an object, or an array
```
When the following config. error was (intentionally) introduced:
```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 78ccc3a35..4f3ef7405 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,7 +51,7 @@ services:
       context: ./merlin-server
       dockerfile: Dockerfile
       args:
-        IMAGE_INCLUDE_JDK: "true"
+        IMAGE_INCLUDE_JDK "true"
     container_name: aerie_merlin
     depends_on: ["postgres"]
     environment:
```

Either way, the `find ... | while read ...` pattern is more than good enough here. I am curious why the `-exec` form differs though, in my mind it shouldn't behave differently than the `find ... | while read ...` form I have.

pcrosemurgy: lol – it's a `docker-compose` vs `docker compose` thing 😂 my bad.

pcrosemurgy: I really prefer the `docker-compose` output since it actually indicates which file is invalid. I'll remedy this by adding an `echo` prior to `docker compose config`.

camargo:
Does this fail the test workflow when it finds an error in a compose file?

pcrosemurgy:
@camargo I'll test this out in this PR, good idea. I tested this with a `... && echo passed` locally and the echo statement was not hit so a non-zero exit status *should* propagate to the test workflow.

.github/workflows/test.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         run: ./gradlew classes
       - name: Assemble
         run: ./gradlew assemble
+      - name: Validate Docker Configurations
+        run: find . -name docker-compose\*.yml | while read -r file; do echo "$file"; docker compose -f "$file" config -q; done
```
JoelCourtney: Could you use `docker-compose ... && docker compose ...` please? Obviously we have one case where V2 is more permissive than V1, but for all we know there could be other cases where V1 is more permissive, and I think we should test for both.

JoelCourtney (approved):
Thanks!

.github/workflows/test.yml
pcrosemurgy: Hmm, is one form really more permissive than the other? The only difference I see is in the error message string being reported, not an absence of error reporting from either tool.

pcrosemurgy:
@camargo https://github.com/NASA-AMMOS/aerie/runs/6012834575?check_suite_focus=true The error propagates :) I'll drop the erroneous commit now.

.github/workflows/test.yml
JoelCourtney: The original bug was that V2 allows `IMAGE_INCLUDE_JDK: true` (without quotes around `true`), while V1 does not

pcrosemurgy: Ohh, that's right, thanks! I'll add that.

camargo (approved):

PR #123, opened 2022-03-31
[AERIE-1744] Scheduler goal priority update SQL triggers

JoelCourtney:
* **Tickets addressed:** AERIE-1744
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description
This PR implements a set of SQL triggers for inserting, updating, and deleting rows from the `scheduling_specification_goals` and `scheduling_template_goals` tables, and slightly modifies the constraints for those tables as well. It also has some non-behavioral refactors for the `db-tests` files.

1. fixes a small unrelated typo
2. Implements the triggers, adds non-negative-priority constraint, sets (spec/template_id, priority) constraint to deferred.
3. Refactors database management out of MerlinDatabaseTests.java into a separate file so SchedulerDatabaseTests.java can use it.
4. Small quality-of-life refactor, combines all the table-clearing functions in MerlinDatabaseTests.java into a single `clearTable(String tableName)` function.
5. Implements the database tests for commit 2.

## Verification

The tests for these triggers are the content of the last commit of the PR.

## Future work

PR #141, opened 2022-04-13
[AERIE-1785] Provide default priorities if not provided

JoelCourtney:
* **Tickets addressed:** AERIE-1785
* **Review:** By commit  

## Description
Adds functionality to the SQL triggers from #123 to allow inserts to not provide a priority, in which case the trigger will select the next consecutive priority for the relevant specification or template. 

## Verification
I added a test to db-tests for this.

### DIFF
--- a/scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
@@ -1,7 +1,11 @@
 create table scheduling_specification_goals (
   specification_id integer not null,
   goal_id integer not null,
-  priority integer not null constraint non_negative_specification_goal_priority check (priority >= 0),
+  priority integer
+    not null
+    default null -- Nulls are detected and replaced with the next
+                 -- available priority by the insert trigger
+    constraint non_negative_specification_goal_priority check (priority >= 0),
 
   constraint scheduling_specification_goals_primary_key
     primary key (specification_id, goal_id),
@@ -31,7 +35,12 @@ comment on column scheduling_specification_goals.priority is e''
 
 create or replace function insert_scheduling_specification_goal_func()
   returns trigger as $$begin
-    if NEW.priority > (
+    if NEW.priority IS NULL then
+      NEW.priority = (
+        select coalesce(max(priority), -1) from scheduling_specification_goals
+        where specification_id = NEW.specification_id
+      ) + 1;
+    elseif NEW.priority > (
       select coalesce(max(priority), -1) from scheduling_specification_goals
         where specification_id = new.specification_id
     ) + 1 then
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/SchedulerDatabaseTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/SchedulerDatabaseTests.java
@@ -273,6 +273,50 @@ class SchedulerDatabaseTests {
           new int[]{0, 1, 2}
       );
     }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"specification", "template"})
+    void shouldGeneratePriorityWhenNull(String tableStem) throws SQLException {
+      connection.createStatement().executeUpdate("""
+          insert into scheduling_%s_goals(%s_id, goal_id)
+          values (%d, %d);
+        """.formatted(
+          tableStem, tableStem,
+          specAndTemplateIds.get(tableStem)[0],
+          goalIds[0]
+      ));
+      checkPriorities(
+          tableStem, 0,
+          new int[]{0},
+          new int[]{0}
+      );
+      connection.createStatement().executeUpdate("""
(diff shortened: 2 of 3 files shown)


JoelCourtney:
Currently not sure if this will pass tests; for some reason I can't get the e2eTests to run locally like I did before. Using this PR as testing.

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
```diff
@@ -1,7 +1,10 @@
 create table scheduling_specification_goals (
   specification_id integer not null,
   goal_id integer not null,
-  priority integer not null constraint non_negative_specification_goal_priority check (priority >= 0),
+  priority integer
+    not null
```
mattdailis, line 5: Hmm, how does `not null default null` work? Seems counter intuitive

goetzrrGit: +1

goetzrrGit:
I started with a fresh docker but I was unable to run the test? I could be doing something wrong here.

Execution failed for task ':db-tests:test'.
> No tests found for given includes: [gov.nasa.jpl.aerie.database.SchedulerDatabaseTests](filter.includeTestsMatching)

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
JoelCourtney: As long as the null is replaced with a number by the trigger before its inserted, thats fine. I considered making the default -1 (which would also violate a constraint if not changed), but I figured if something went wrong in the trigger and it wasn't updated, it would be easier to debug if it was null.

JoelCourtney:
@goetzrrGit What command are you using? For some other reason I wasn't able to test this locally, so I just relied on the PR workflow. Also @camargo will probably know about this better, if this is an e2e-test thing.

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
JoelCourtney: I would have liked to do a autoincrement-related solution, but we have multiple lists interleaved in the tables (one list per spec or template), so autoincrement or SERIAL wouldn't work.

mattdailis:
> I started with a fresh docker but I was unable to run the test? I could be doing something wrong here.
> 
> Execution failed for task ':db-tests:test'.
> 
> > No tests found for given includes: [gov.nasa.jpl.aerie.database.SchedulerDatabaseTests](filter.includeTestsMatching)

Yeah, I believe we renamed that one to be `:db-tests:e2eTest` instead of test. The thought process there was that e2eTest expects a deployment of aerie to be available.

goetzrrGit:
> @goetzrrGit What command are you using? For some other reason I wasn't able to test this locally, so I just relied on the PR workflux. Also @camargo will probably know about this better, if this is an e2e-test thing.

I tried to run it in IntelliJ but if this a GitHub action thing then you can ignore my attempts to run the test locally.

mattdailis:
> > @goetzrrGit What command are you using? For some other reason I wasn't able to test this locally, so I just relied on the PR workflux. Also @camargo will probably know about this better, if this is an e2e-test thing.
> 
> I tried to run it in IntelliJ but if this a GitHub action thing then you can ignore my attempts to run the test locally.

@goetzrrGit You should be able to run `./gradlew :db-tests:e2eTest`. I hadn't considered the IntelliJ user case, might be worth looking into how to configure that in the future

goetzrrGit:
> > > @goetzrrGit What command are you using? For some other reason I wasn't able to test this locally, so I just relied on the PR workflux. Also @camargo will probably know about this better, if this is an e2e-test thing.
> > 
> > 
> > I tried to run it in IntelliJ but if this a GitHub action thing then you can ignore my attempts to run the test locally.
> 
> @goetzrrGit You should be able to run `./gradlew :db-tests:e2eTest`. I hadn't considered the IntelliJ user case, might be worth looking into how to configure that in the future

ohhh, got it, thanks @mattdailis

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
JoelCourtney: I've added a comment explaining this, or I don't mind doing it a different way if you want.

mattdailis: Thanks for the comment :+1: I think this is fine - the timing of the constraints and the triggers is a bit nuanced, but the comment helps elucidate that.

So - my current understanding is that the order of events is:
1. INSERT statement comes in, and it's missing the priority field
2. That priority field is set to null by the "default null"
3. The trigger replaces the null with a number
4. The NOT NULL constraint is checked - if we did our job right, this should never fail

mattdailis (approved):

goetzrrGit (approved):
Thanks for the explanation about the not `null` and setting `null` and all the tests passed on my end.

PR #143, opened 2022-04-14
docs: add ui ORIGIN env var

camargo:
## Description

- Specifies where UI is served from
- Update environment doc with description

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -109,6 +109,7 @@ services:
       GATEWAY_SERVER_URL: http://aerie_gateway:9000
       HASURA_CLIENT_URL: http://localhost:8080/v1/graphql
       HASURA_SERVER_URL: http://hasura:8080/v1/graphql
+      ORIGIN: http://localhost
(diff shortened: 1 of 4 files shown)


patkenneally (approved):

pcrosemurgy (approved):

deployment/docker-compose.yml
```diff
@@ -109,6 +109,7 @@ services:
       GATEWAY_SERVER_URL: http://aerie_gateway:9000
       HASURA_CLIENT_URL: http://localhost:8080/v1/graphql
       HASURA_SERVER_URL: http://hasura:8080/v1/graphql
+      ORIGIN: http://localhost
```
mattdailis, line 112: Is the expectation that users would edit this value to match their url?

mattdailis (approved):

deployment/docker-compose.yml
camargo: Yes exactly. So for example if they deploy the UI server behind an HTTPS proxy you can update the ORIGIN to be https.

PR #144, opened 2022-04-15
[AERIE-1798]  Fix JSON parse errors in scheduling dsl compiler

mattdailis:
* **Tickets addressed:** AERIE-1798
* **Review:** By commit  

## Description

While writing e2e tests for the scheduler, @adrienmaillard observed a periodic test failure. We did not observe a pattern at the time - but in hindsight, I think every 12 (or so) successful test runs, there would be two test failures. Both would be "panics" of the scheduling dsl compiler, as it failed to parse JSON. The first would be an "Unexpected end of JSON input", and the second would be "Illegal character in JSON" (paraphrasing).

What I believe is happening is that the buffer allocated for standard input on the node subprocess is filling up mid-message, and calling our message handler twice, each time with half a message. This happens periodically because our banananation messages are much smaller than the buffer. (Attempted illustration below - note that Message 3 is split across the two buffers).

```mermaid
gantt
Buffer 1 :done,    des1, 2014-01-06,2014-01-11
Buffer 2 :done,    des1, 2014-01-11,2014-01-16
Message 1   :         des3, 2014-01-06, 2d
Message 2   :         des4, after des3, 2d
Message 3   :         des5, after des4, 2d
```

I would expect this to happen much more often with a larger mission model, since the size of the messages scales with the number of activity types in the mission model. (The generated typescript library has a function for every activity type).

Why does our handler get called twice? I _think_ it's because the `process.stdin.on('data', ...)` triggers the callback every time the buffer is written to. If the buffer fills up, it calls our callback, and continues writing from the beginning of a new buffer.

Why do we get two test failures in a row? Since our handler is called twice, it writes two messages to standard output. The Java code only reads one of those messages every time. So, the first time, it sees the failure that resulted from the handler being called on the first half of the message. The next time the tests are run, the error from the handler being called on the second half of the message is still sitting in standard output, patiently waiting to be read. **This means that all subsequent runs of the tests will be reading the result of the previous test** 😱 (that didn't dawn on me until writing this up just now)

The fix is to call our callback every time we get a complete message, to avoid calling it with partial messages. Our protocol demarcates the end of messages with a newline. The fix in this PR is to use the `readline` module from the node standard library. This abstracts away the buffer, and allows us to register a callback on the `'line'` event.

## Verification

Once I had a hunch that there was a buffering issue, I decided to try to force the error by passing the scheduling dsl compiler a very large message that would not fit into the buffer. If my mental model of the problem was correct, this would cause our handler to be called on a partial message. Indeed, it failed with an `Unexpected end of JSON` error. 

Screen Shot 2022-04-15 at 7 31 52 AM

(Screenshot taken from this run: https://github.com/NASA-AMMOS/aerie/runs/6039227681#r20)

After making the `readline` change described above, this test no longer fails.

## Documentation

I think this PR write-up as well as the git history serves as documentation of this fix.

## Future work

We need to avoid making the same mistake on the constraints dsl compiler (cc: @JoelCourtney )

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/main.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/main.ts
@@ -1,22 +1,45 @@
 import fs from 'fs';
 import ts from 'typescript';
 import { UserCodeRunner } from '@nasa-jpl/aerie-ts-user-code-runner';
-import type { Goal } from './libs/scheduler-edsl-fluent-api.js'
+import type { Goal } from './libs/scheduler-edsl-fluent-api.js';
 
 const codeRunner = new UserCodeRunner();
 const schedulerEDSL = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-edsl-fluent-api.ts`, 'utf8');
 const schedulerAST = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-ast.ts`, 'utf8');
 
+function registerInputCallback(callback: (data: string) => void) {
+  process.stdin.once('data', data => callback(data.toString()));
+}
+
+function writeOutputOneLine(data: string) {
+  process.stdout.write(data + '\n');
+}
+
+function writeSuccess(data: string) {
+  writeOutputOneLine('success');
+  writeOutputOneLine(data);
+}
+
+function writeError(data: string) {
+  writeOutputOneLine('error');
+  writeOutputOneLine(data);
+}
+
+function writePanic(data: string) {
+  writeOutputOneLine('panic');
+  writeOutputOneLine(data);
+}
+
 process.on('uncaughtException', (err) => {
   console.error('uncaughtException');
   console.error((err && err.stack) ? err.stack : err);
-  process.stdout.write('panic\n' + err.stack ?? err.message);
+  writePanic(err.stack ?? err.message);
   process.exit(1);
 });
 
-async function handleRequest(data: Buffer) {
+async function handleRequest(data: string) {
   try {
-    const { goalCode, missionModelGeneratedCode } = JSON.parse(data.toString()) as { goalCode: string, missionModelGeneratedCode: string };
+    const { goalCode, missionModelGeneratedCode } = JSON.parse(data) as { goalCode: string, missionModelGeneratedCode: string };
 
     const result = await codeRunner.executeUserCode<[], Goal>(
         goalCode,
@@ -32,8 +55,8 @@ async function handleRequest(data: Buffer) {
     );
 
     if (result.isErr()) {
-      process.stdout.write('error\n' + JSON.stringify(result.unwrapErr().map(err => err.toJSON())) + '\n');
-      process.stdin.once('data', handleRequest);
+      writeError(JSON.stringify(result.unwrapErr().map(err => err.toJSON())));
+      registerInputCallback(handleRequest);
       return;
     }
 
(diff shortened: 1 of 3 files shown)


dyst5422:
This was a pretty big oversight on our part the first time we did this. Seems like we may need to patch 0.11 with this fix.

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -1,22 +1,48 @@
...
+import type { Goal } from './libs/scheduler-edsl-fluent-api.js';
 
 const codeRunner = new UserCodeRunner();
 const schedulerEDSL = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-edsl-fluent-api.ts`, 'utf8');
 const schedulerAST = fs.readFileSync(`${process.env.SCHEDULING_DSL_COMPILER_ROOT}/src/libs/scheduler-ast.ts`, 'utf8');
 
+const input = readline.createInterface(process.stdin);
+
+function registerInputCallback(callback: (data: string) => void) {
+  input.once('line', callback);
```
dyst5422, line 14: We might as well inject the handler here and have no parameters on this function

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -45,18 +71,18 @@ async function handleRequest(data: Buffer) {
...
-    process.stdout.write('panic\n' + JSON.stringify(error.stack ?? error.message) + '\n');
+    writePanic(JSON.stringify(error.stack ?? error.message));
   }
-  process.stdin.once('data', handleRequest);
+  registerInputCallback(handleRequest);
 }
 
-process.stdin.once('data', data => {
+registerInputCallback(data => {
   if (data.toString().trim() === 'ping') {
```
dyst5422: data is already string according to line 13

dyst5422 (approved):

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -241,4 +241,44 @@ export default function myGoal() {
...
+                  })
+                }
+            """ + " ".repeat(9001), "goalfile");
+    final var expectedGoalDefinition = new SchedulingDSL.GoalSpecifier.GoalDefinition(
+        SchedulingDSL.GoalKinds.ActivityRecurrenceGoal,
+        new SchedulingDSL.ActivityTemplate(
+            "PeelBanana",
+            Map.ofEntries(
+                Map.entry("peelDirection", SerializedValue.of("fromStem")),
+                Map.entry("fancy", SerializedValue.of(Map.ofEntries(
```
adrienmaillard, line 269: Can you explain how specifying a `fancy` field (and subfields) that do not exist in the `PeelBanana` activity of the bananation mission model does not make compilation fail?

dyst5422: We create a mock mission model which is imported and used here https://github.com/NASA-AMMOS/aerie/pull/144/files/0ca7153d893c52396ff60091e246a0fcac2627d9#diff-9ddb02eebf6befa5cd70d068f1189dcf09f46ccd8dd7e9329da070278105016eR106 which does have these additional fields.

mattdailis: Good question, and thanks @dyst5422 for the response. I'm planning on changing the names of the activities in the mock mission model so that it doesn't pretend to be banananation and cause confusion :+1:

mattdailis: Here's the change: https://github.com/NASA-AMMOS/aerie/compare/refactor/rename-mock-activities

I can add that to this PR or open a new PR, as you guys prefer.

adrienmaillard: Seems I missed this. But there is little chance I would be the only one confused so 👍 to renaming.

adrienmaillard (approved):

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
adrienmaillard: I missed this. But as I will probably not be the only one confused, 👍 to the renaming. You can put it in this PR or another, I don't mind.

mattdailis: I've included this change as a commit in this PR.

PR #145, opened 2022-04-15
[FIX] Fix special case issue in RecurrenceGoal and hasura request.

adrienmaillard:
* **Review:** By commit  

## Description

When testing scheduling, I have noticed these two issues: 
- the delete hasura request was not correctly written in `GraphQLMerlinService`.
- the `RecurrenceGoal` was creating a conflict even if the temporal domain of this conflict is a singleton.

## Verification

Verified the same way these were found, by exercising scheduling from the UI directly. 

## Future work

One scheduling e2e test is coming in another PR.

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
@@ -112,7 +112,7 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
       final var strideDur = actStartT.minus(prevStartT);
       if (strideDur.compareTo(recurrenceInterval.end) > 0) {
         //fill conflicts for all the missing activities in that long span
-        conflicts.addAll(makeRecurrenceConflicts(prevStartT, actStartT, plan));
+        conflicts.addAll(makeRecurrenceConflicts(prevStartT, actStartT));
 
       } else {
         /*TODO: right now, we associate with all the activities that are satisfying but we should aim for the minimum
@@ -128,7 +128,7 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
 
     //fill in conflicts for all missing activities in the last span up to the
     //goal's own end time (also handles case of no matching acts at all)
-    conflicts.addAll(makeRecurrenceConflicts(prevStartT, lastStartT, plan));
+    conflicts.addAll(makeRecurrenceConflicts(prevStartT, lastStartT));
 
     return conflicts;
   }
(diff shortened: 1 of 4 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -178,10 +178,12 @@ private java.util.Collection<MissingActivityConflict> makeRecurrenceConflicts(
       //        framed, which is exactly what the current solver will do since
       //        it chooses the minimum. but it looks ugly. so for now passing
       //        a limited flexibility for creation
-      var interval =  new Windows(Window.between(intervalT.minus(recurrenceInterval.end), Duration.min(intervalT, end)));
-      conflicts.add(new MissingActivityTemplateConflict(
-          this, new Windows(
-          interval),this.getActTemplate()));
+      var interval = Window.between(intervalT.minus(recurrenceInterval.end), Duration.min(intervalT, end));
+      if(!interval.isSingleton()) {
```
mattdailis: I'm curious why this is a problem - if an activity has zero duration, it can fit into a singleton interval, right?

adrienmaillard: This has more to do with the semantics of the goal. This is a corner case happening when the goal planning horizon is  a multiple of the recurrence interval. Ex: 8h horizon for a 1hr recurrence. We go through each hour and find that we need one activity for each. At some point, we have `intervalT` = 9, we do i`ntervalT.minus(1)` = 8, we get `intervalT`=8. For the other part of the resulting interval, we take the minimum between end of planning horizon (to avoid setting the domain of the conflict outside of the overall horizon), here 8, and `intervalT` = 8. Which makes an empty interval. But actually, as the interval is empty, it means we do not need to schedule anything there, we already satisfied the goal of 1 act every hour. 
If the end of the scheduling horizon was at 8h05, we would need an activity between 8h and 8h05, and that's effectively what it would schedule.

Twisol: I'm a little confused myself. The endpoints of a `Window` can be either inclusive or exclusive; if you're treating a singleton as though it were empty, it suggests maybe the endpoints should be exclusive rather than inclusive, so that the window *is* actually empty.

adrienmaillard: After discussing with @mattdailis, I have reverted the change and only replaced the break statement from `if(intervalT.compareTo(end) > 0)` by `if(intervalT.compareTo(end) >= 0)`, making `end` an exclusive bound. 
This procedure is called twice (above), once in the case where there is too large of a gap between 2 activities, then we are sure we do not want to include the `end` as it corresponds to the start of the second activity. For the other call, `end` corresponds to the end of the `temporalContext` which is the scheduling horizon of the goal.

mattdailis (approved):

JoelCourtney (approved):
LGTM

PR #146, opened 2022-04-15
[AERIE-1775] Type check on expansion set creation

dyst5422:
* **Tickets addressed:** AERIE-1775
* **Review:** By commit  

## Description
We started by fixing up the tests for 0.11.0.

Then we added in an expansion batch loader so we can load the expansions for typechecking when the request to create an expansion set comes in.

Last, we added a typecheck function to our worker so we can typecheck the expansions in parallel on our workers.

## Verification
I ran the existing tests and tested the create expansion set action manually though the hasura api. I am working on some automated api tests, but they aren't part of this PR.

## Documentation
No known documentation needs update.

## Future work
No known future work directly related to this


command-expansion-server/test/utils/MissionModel.ts
```diff
@@ -1,13 +1,12 @@
...
+import fetch, { FormData, fileFrom } from 'node-fetch';
 import { gql, GraphQLClient } from 'graphql-request';
 import { randomUUID } from 'crypto';
 
 export async function uploadMissionModel(graphqlClient: GraphQLClient): Promise<number> {
   /*
    * Insert a mission model
    */
-  const bananaNationMissionModelUrl = new URL('../inputs/banananation-0.10.0-SNAPSHOT.jar', import.meta.url);
+  const bananaNationMissionModelUrl = new URL('../inputs/banananation-0.11.0-SNAPSHOT.jar', import.meta.url);
```
mattdailis, line 9: 🔵 I think there may be a way to pull the "latest" banananation from the `examples` directory. Not necessary for this PR though.

command-expansion-server/test/batchLoaders/expansionBatchLoader.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+it('should load expansion data', async () => {
+  const expansions = await expansionBatchLoader({graphqlClient})([
+    { expansionId },
+  ]);
+  if (expansions[0] instanceof Error) {
+    throw expansions[0];
+  }
+  expect(expansions[0].activityType).toBe('PeelBanana');
+  expect(expansions[0].expansionLogic).toBeString();
```
mattdailis, line 25: 🔵 Just checking my understanding - is this `toBeString()` call the reason we've added the `jest-extended` dependency?

command-expansion-server/src/app.ts
```diff
@@ -118,10 +123,51 @@ app.post('/put-expansion', async (req, res, next) => {
...
+
+    return result.errors;
+  }));
+
+  const errors = unwrapPromiseSettledResults(typecheckErrorPromises).reduce((accum, item) => {
+    if (item instanceof Error) {
+      accum.push(item);
+    } else {
+      accum.concat(item);
+    }
```
mattdailis, line 166: 🔵 Under what circumstances does the type of `item` vary between `Error` and `Error[]`?

mattdailis (approved):
Looks reasonable to me - I had a few questions as I was reading, but nothing that should block merge.

> Key:
>  🔶 Requests that should not block merge, but should at least be discussed 

command-expansion-server/test/utils/MissionModel.ts
dyst5422: Yeah, I thought about it and didn't want to have to deal with that at this time.

command-expansion-server/test/batchLoaders/expansionBatchLoader.spec.ts
dyst5422: Yes. The additional matchers are generally handy.

command-expansion-server/src/app.ts
dyst5422: It's an error if it: fails to load the activity schema or there is some kind of panic-style failure. Otherwise, it is an array of diagnostic style errors. I think it would be good for use to pick and use a more specific/different nomenclature to differentiate between panic errors and diagnostic errors (user errors).

mattdailis: Gotcha - thanks for clarifying :+1:

command-expansion-server/test/utils/MissionModel.ts
goetzrrGit: If you want you can use this script I wrote as an example if you want to pull the latest snapshot down from Github. You can probably hook it up to the `npm run test`

https://github.jpl.nasa.gov/Aerie/aerie-ui-test/blob/develop/scripts/pull_models.sh

goetzrrGit (approved):
I ran through the test and got the `[XFAIL] should load expansion set data` as expected.

I added a comment linking to a script to fetch the latest `mission_model` from GitHub if you wanted to incorporate that. It is not needed for this PR so you can merge it when it is ready.

PR #149, opened 2022-04-21
[AERIE-1790] Added Relative, Absolute, and Epoch timing.

goetzrrGit:
* **Tickets addressed:** AERIE-1790
* **Review:** By commit  

## Description

I added these timing options to the seqJson generated during expansion runs:

“ABSOLUTE”,
“COMMAND_RELATIVE”,
“EPOCH_RELATIVE”,
“COMMAND_COMPLETE”

The time values use [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) and follow the stage 3 proposal for [Temporal](https://tc39.es/proposal-temporal/docs/)

I am using this spec file here as a reference: [spec](https://github.jpl.nasa.gov/Aerie/aerie-seq-schema/blob/develop/schema.json)

Here is an example of expansion logic:

```
    BAKE_BREAD.relativeTiming(Temporal.Duration.from({ minutes: 5 })),
    BAKE_BREAD.absoluteTiming(Temporal.Instant.from("2022-04-20T20:17:13Z")),
    BAKE_BREAD.epochTiming("starPoch", Temporal.Duration.from({ minutes: 15 })),
```
Absolute time is not a duration but rather a [Temporal.Instant](https://tc39.es/proposal-temporal/docs/instant.html) as you will need to provide an exact time of when you will like to run the command.

You will notice the `epochTiming` will require an epoch name. From what I gathered, epoch names are global so the user should provide a valid name. I am storing the epochName in the `epoch : {} `field in the generated seqJson.

Here is the output from the expansion run

```
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "PT5M",
                    "type": "COMMAND_RELATIVE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "2022-04-20T20:17:13Z",
                    "type": "ABSOLUTE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "PT15M",
                    "type": "EPOCH_RELATIVE"
                },
                "type": "command",
                "epoch": "Nova",
                "metadata": {}
            }
        ],
        "metadata": {}
    }
```

## Verification
I ran through the whole expansion process to generate the above seqJson. 

## Documentation
none

## Future work
Continue to add missing fields in the seqJson generated by sequence expansions.

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -2,33 +2,99 @@
 export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
   public readonly stem: string;
   public readonly arguments: A;
+  public readonly absoluteTime: Temporal.Instant | null = null;
+  public readonly epochTime: { epochName: string; time: Temporal.Duration } | null = null;
+  public readonly relativeTime: Temporal.Duration | null = null;
 
-  private constructor(opts: { stem: string; arguments: A }) {
+  private constructor(opts: CommandOptions<A>) {
     this.stem = opts.stem;
     this.arguments = opts.arguments;
+    if ('absoluteTime' in opts) {
+      this.absoluteTime = opts.absoluteTime;
+    } else if ('epochTime' in opts) {
+      this.epochTime = opts.epochTime;
+    } else if ('relativeTime' in opts) {
+      this.relativeTime = opts.relativeTime;
+    }
   }
 
-  public static new<A extends any[] | { [argName: string]: any }>(opts: { stem: string; arguments: A }): Command<A> {
-    return new Command({
-      stem: opts.stem,
-      arguments: opts.arguments,
-    });
+  public static new<A extends any[] | { [argName: string]: any }>(opts: CommandOptions<A>): Command<A> {
+    if ('absoluteTime' in opts) {
+      return new Command<A>({
+        ...opts,
+        absoluteTime: opts.absoluteTime,
+      });
+    } else if ('epochTime' in opts) {
+      return new Command<A>({
+        ...opts,
+        epochTime: opts.epochTime,
+      });
+    } else if ('relativeTime' in opts) {
+      return new Command<A>({
+        ...opts,
+        relativeTime: opts.relativeTime,
+      });
+    } else {
+      return new Command<A>(opts);
+    }
   }
 
   public toSeqJson() {
     return {
       id: 'command',
       metadata: {},
+
       steps: [
         {
           stem: this.stem,
-          time: { type: 'COMPLETE' },
+          time: {
+            tag: this.absoluteTime
+              ? this.absoluteTime.toString()
+              : this.relativeTime
+              ? this.relativeTime.toString()
+              : this.epochTime
+              ? this.epochTime.time.toString()
+              : '',
+            type: this.absoluteTime
+              ? 'ABSOLUTE'
+              : this.relativeTime
+              ? 'COMMAND_RELATIVE'
+              : this.epochTime
+              ? 'EPOCH_RELATIVE'
+              : 'COMMAND_COMPLETE',
+          },
           type: 'command',
           metadata: {},
+          // include epoch field if we have an epoch time
+          ...(this.epochTime !== null ? { epoch: this.epochTime?.epochName } : {}),
           args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
         },
       ],
     };
+  }
+
+  public absoluteTiming(absoluteTime: Temporal.Instant): Command<A> {
+    return Command.new({
+      stem: this.stem,
+      arguments: this.arguments,
+      absoluteTime: absoluteTime,
+    });
+  }
+
+  public epochTiming(epochName: string, epochTime: Temporal.Duration): Command<A> {
+    return Command.new({
+      stem: this.stem,
+      arguments: this.arguments,
+      epochTime: { epochName, time: epochTime },
+    });
+  }
+
+  public relativeTiming(relativeTime: Temporal.Duration): Command<A> {
+    return Command.new({
+      stem: this.stem,
+      arguments: this.arguments,
+      relativeTime: relativeTime,
+    });
   }
 }
 
@@ -36,12 +102,21 @@ declare global {
   export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
     public readonly stem: string;
     public readonly arguments: A;
+    public readonly absoluteTime: Temporal.Instant | null;
+    public readonly epochTime: { epochName: string; time: Temporal.Duration } | null;
+    public readonly relativeTime: Temporal.Duration | null;
 
-    private constructor(opts: { stem: string; arguments: A });
+    private constructor(opts: CommandOptions<A>);
 
-    public static new<A extends any[] | { [argName: string]: any }>(opts: { stem: string; arguments: A }): Command<A>;
+    public static new<A extends any[] | { [argName: string]: any }>(opts: CommandOptions<A>): Command<A>;
 
     public toSeqJson(): any;
+
+    public absoluteTiming(absoluteTime: Temporal.Instant): Command<A>;
+
+    public epochTiming(epochName: string, epochTime: Temporal.Duration): Command<A>;
+
+    public relativeTiming(relativeTime: Temporal.Duration): Command<A>;
   }
   type Context = {};
   type ArgType = boolean | string | number;
@@ -62,5 +137,18 @@ declare global {
   type F<BitLength extends 32 | 64> = number;
   type F32 = F<32>;
   type F64 = F<64>;
+  type CommandOptions<A> = { stem: string; arguments: A } & (
+    | {
+        absoluteTime: Temporal.Instant;
+      }
+    | {
+        epochTime: { epochName: string; time: Temporal.Duration };
+      }
+    | {
+        relativeTime: Temporal.Duration;
+      }
+    // CommandComplete
+    | {}
+  );
 }
 /** END Preface */


command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,122 @@
...
+      relativeTime: relativeTime,
+    });
+  }
+
+  public epochTiming(epochName: string, epochTime: Temporal.Duration): Command<A> {
+    return new Command({
+      stem: this.stem,
+      arguments: this.arguments,
+      epochTime: { epochName: epochName, time: epochTime },
+    });
```
goetzrrGit: In case you are wondering why I am creating a new command object instead of doing the below:
```
public absoluteTiming(absoluteTime: Temporal.Duration): Command<A> {
    this.absoluteTime = absoluteTime;
    return this;
  }
```

If we are using a command multiple times you will override the values or store multiple time values in the single command.


In this example:
```
// My instrument will prepare 2 bread loafs

PREPARE_BREAD
BAKE_BREAD.absoluteTiming(Temporal.Duration.from({ minutes: 5 })),
....
....
PREPARE_BREAD
BAKE_BREAD.absoluteTiming(Temporal.Duration.from({ minutes: 15 })),
```

Both BAKE_BREAD will generate an absolute time of 15 minutes, not 5 and 15 
```

    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "PT15M",
                    "type": "ABSOLUTE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
  {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "PT15M",
                    "type": "ABSOLUTE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
```

dyst5422: Ah yes, this is a limitation of how we are skipping parens for commands without arguments by making it a single exported singleton. This could be sufficient cause to reverse that decision and require the parens.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,122 @@
 export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
   public readonly stem: string;
   public readonly arguments: A;
+  public readonly absoluteTime?: Temporal.Duration;
```
dyst5422: Let's make these `Temporal.Duration | null` so we're explicit about there being _nothing_ in there. undefined is kind of weak in that bad indexing or undefined variables can produce undefined

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,122 @@
...
+          },
           type: 'command',
           metadata: {},
           args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
         },
       ],
     };
+    // include epoch field if we have an epoch time
+    if (this.epochTime) {
+      seqJson['epoch'] = this.epochTime.epochName;
```
dyst5422: we can put this in above with a ternary like so
```
{
       id: 'command',
       metadata: {},
       ...(this.epochTime !== null ? { epoch: this.epochTime.epochName } : {})
       steps: [
         {
```

I think it's a little more clear that we have a conditional attribute that way

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,122 @@
...
+  private constructor(opts: {
+    stem: string;
+    arguments: A;
+    absoluteTime?: Temporal.Duration;
+    relativeTime?: Temporal.Duration;
+    epochTime?: { epochName: string; time: Temporal.Duration };
+  }) {
     this.stem = opts.stem;
     this.arguments = opts.arguments;
+    this.absoluteTime = opts.absoluteTime;
```
dyst5422: We need to prevent multiple of these from being set on the same command.

dyst5422 (changes requested):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: Or we can put it in and strip undefined attributes prior to returning.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I am not sure I understand. The command will expand to the latest time value you specify

```
PEEL_BANANA.absoluteTiming(Temporal.Duration.from({ minutes: 10 })).absoluteTiming(Temporal.Duration.from({ minutes: 50 })),

BAKE_BREAD.absoluteTiming(Temporal.Duration.from({ minutes: 10 })).epochTiming("Saturn", Temporal.Duration.from({ minutes: 15 }))
```

PEEL_BANANA - 50 minutes absolute
BAKE_BREAD - 15 minutes epoch

dyst5422: Yes, but we aren't enforcing in the type system that only one of these values is populated at once. We can do this by typing the input type as
```
{
  stem: string;
  arguments: A;
} & (
  {
    absoluteTime: Temporal.Duration;
    epochTime: null;
    relativeTime: null;
  }
  | {
    absoluteTime: null;
    epochTime: { epochName: string; time: Temporal.Duration };
    relativeTime: null;
  }
  | {
    absoluteTime: null;
    epochTime: null;
    relativeTime: Temporal.Duration;
  }
)
```

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,118 @@
 export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
   public readonly stem: string;
   public readonly arguments: A;
+  public readonly absoluteTime?: Temporal.Duration | null;
+  public readonly epochTime?: { epochName: string; time: Temporal.Duration } | null;
```
dyst5422: Lets make these non-optional.

dyst5422 (changes requested):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: Basically keeping to the rule of "Make invalid states unrepresentable"

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: They are now non-optional.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I did some reading up on this as I was a little confusing when looking at the example you provided. I have never seen intersection types before. I got it implemented now. 

https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types

The change required me to update [CommandTypeCodegen.ts](https://github.com/NASA-AMMOS/aerie/pull/149/commits/09058f7ed1a2fc7d28df2ff2a97c8c37e286b08b#diff-664406c819e249960325a4124e84cd5c9f675976870b5a8e5c7a92ad6cf9ea73)

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,154 @@
...
-      stem: opts.stem,
-      arguments: opts.arguments,
-    });
+  public static new<A extends any[] | { [argName: string]: any }>(opts: {
+    stem: string;
+    arguments: A;
+    absoluteTime: Temporal.Duration;
+    epochTime: { epochName: string; time: Temporal.Duration } | null;
+    relativeTime: Temporal.Duration | null;
+  }): Command<A> {
```
dyst5422: This isn't doing what we want. The short circuit operator (`||`) only short circuits on falsy values (`undefined`, `null`, `false`). Here, we are using an `object`, so it never short circuits.

I think we want to make the signature something like
```opts: {
  stem: string;
  arguments: A;
} & (
  {
    absoluteTime: Temporal.Duration;
  }
  | {
    epochTime: { epochName: string; time: Temporal.Duration };
  }
  | {
    relativeTime: Temporal.Duration;
  }
  | {} // CommandComplete
)
```

So we only have to the value of concern, and then do an if statement for construction (could be ternary, but probably clearer as an if series)

```
if (opts.absoluteTime !== undefined) {
  return new Command({
    ...opts,
    epochTime: null,
    relativeTime: null,
  });
} else if (opts.epochTime !== undefined) {
  return new Command({
    ...opts,
    absoluteTime: null,
    relativeTime: null,
  });
} else if (opts.absoluteTime !== undefined) {
  return new Command({
    ...opts,
    absoluteTime: null,
    epochTime: null,
  });
} else {
  return new Command({
    ...opts,
    absoluteTime: null,
    epochTime: null,
    relativeTime: null,
  });
}
```
Might need to play around with it a little for the type system as it may complain with an error about `never` type. In that case, use a type guard (https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) to narrow the type appropriately.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,154 @@
...
+
       steps: [
         {
           stem: this.stem,
-          time: { type: 'COMPLETE' },
+          time: {
+            tag: this.absoluteTime
+              ? this.absoluteTime.toString()
+              : this.relativeTime
+              ? this.relativeTime.toString()
```
dyst5422, line 54: We should double check that the string output of duration complies with the seqJson spec.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,154 @@
...
+          // include epoch field if we have an epoch time
+          ...(this.epochTime !== null ? { epoch: this.epochTime?.epochName } : {}),
           args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
         },
       ],
     };
   }
+
+  public absoluteTiming(absoluteTime: Temporal.Duration): Command<A> {
+    return new Command({
```
dyst5422: We should use Command.new(). It's a pattern that allows us to have the constructor private and simple, and delegate any logic. This becomes important when we need to have some kind of async functionality in the construction (time conversion via a service for example) because constructors can't be async, but Command.new can be. Other use cases involve construction that is more complex than simple instantiation (needing to instantiate child object first).

If we do so, we can omit the null values here too.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,154 @@
...
+      arguments: this.arguments,
+      absoluteTime: null,
+      epochTime: null,
+      relativeTime: relativeTime,
+    });
+  }
 }
 
 declare global {
   export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
```
dyst5422, line 102: We need to update the declaration here to match the class above. (the new static function is off)

command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -64,8 +64,11 @@ function generateFswCommandCode(
     const value = `
 ${doc}
 export const ${fswCommandName}: ${fswCommandName} = Command.new({
-\tstem: '${fswCommand.stem}',
-\targuments: [],
+  stem: '${fswCommand.stem}',
+  arguments: [],
+  absoluteTime: null,
```
dyst5422: We can omit the null in these if we make the change suggested above.

dyst5422 (changes requested):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Yep from our slack thread, `Durations` will be using [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations)

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I think I type guard the different time values correctly as I am able to expand my commands. 

```
 if ('absoluteTime' in opts) {
      return new Command<A>({
        ...opts,
        absoluteTime: opts.absoluteTime,
      });
    }
```

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Very interesting!!!

command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: I reverted back this file.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,136 @@
...
+  public readonly epochTime: { epochName: string; time: Temporal.Duration } | null;
+  public readonly relativeTime: Temporal.Duration | null;
 
-  private constructor(opts: { stem: string; arguments: A }) {
+  private constructor(opts: { stem: string; arguments: A } & timingType) {
     this.stem = opts.stem;
     this.arguments = opts.arguments;
+    if ('absoluteTime' in opts) {
+      this.absoluteTime = opts.absoluteTime;
+      this.epochTime = null;
```
dyst5422: 100% a stylistic preference: We can have the properties initialize to `null` by doing like `public readonly absoluteTime: Temporal.Instant | null = null;` and then we don't have to assign here.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,136 @@
...
-  public static new<A extends any[] | { [argName: string]: any }>(opts: { stem: string; arguments: A }): Command<A> {
-    return new Command({
-      stem: opts.stem,
-      arguments: opts.arguments,
-    });
+  public static new<A extends any[] | { [argName: string]: any }>(
+    opts: {
+      stem: string;
+      arguments: A;
+    } & timingType,
```
dyst5422: Type names should conventionally use capitalized camelcase `TimingType` to distinguish it from variables.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,136 @@
...
+          time: {
+            tag: this.absoluteTime
+              ? this.absoluteTime.toString()
+              : this.relativeTime
+              ? this.relativeTime.toString()
+              : this.epochTime
+              ? this.epochTime.time.toString()
+              : '',
+            type: this.absoluteTime
+              ? 'ABSOLUTE'
```
dyst5422, line 59: Stylistic: we _might_ want to make these into a string enum

```
enum TimeType {
  ABSOLUTE = 'ABSOLUTE',
 ...
}
```

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -62,5 +152,17 @@ declare global {
   type F<BitLength extends 32 | 64> = number;
   type F32 = F<32>;
   type F64 = F<64>;
+  type timingType =
```
dyst5422: `TimingType` to distinguish types + interfaces from variables

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I agree, less code in the file too :)

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Thanks for the suggestion. I made the change

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I updated the to TimingType

dyst5422 (approved):

dyst5422:
Looks very solid to me!

dyst5422:
Given our discussion with Mike about time formats, I think we'll have to add to the formatters for the time serialization. So we may want to hold this PR until we get full concurrence on that.

goetzrrGit:
> Given our discussion with Mike about time formats, I think we'll have to add to the formatters for the time serialization. So we may want to hold this PR until we get full concurrence on that.

I was thinking that too or creating a new ticket to add a time formatted to this.

dyst5422:
Since we have other work to go on the seqJson support, I think we should merge and handle that in a different PR

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -2,46 +2,126 @@
 export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
   public readonly stem: string;
   public readonly arguments: A;
+  public readonly absoluteTime: Temporal.Instant | null = null;
+  public readonly epochTime: { epochName: string; time: Temporal.Duration } | null = null;
+  public readonly relativeTime: Temporal.Duration | null = null;
 
-  private constructor(opts: { stem: string; arguments: A }) {
+  private constructor(opts: { stem: string; arguments: A } & TimingType) {
```
camargo: Can we extract this type? It seems to be duplicated inline in this file

goetzrrGit: Hi Chris, are you talking about` { stem: string; arguments: A }` `A` is a generic value so I am not sure how to pull that part out as a type?

dyst5422:
```ts
type CommandConstructorArgs<A> = {
  stem: string;
  arguments: A;
} & ({
} | {
} | {
} | {})
```

camargo: Something like:
```ts
type CommandOptions<A> = {
  arguments: A;
  stem: string;
} & TimingType;
```

camargo: Jinx!

goetzrrGit: ohhh got it. Thanks!!!

goetzrrGit: Should be good to go. I ran through expansion without any errors.

camargo (approved):
LGTM

dyst5422:
🚢

PR #150, opened 2022-04-21
[AERIE-1787] Constraint Typescript compilation on Merlin server

JoelCourtney:
* **Tickets addressed:** AERIE-1787
* **Review:** By commit  

## Description
The general idea was to copy and reword/rework the existing scheduler-server TS compilation into merlin-server, replacing scheduling goals with constraints. I ended up creating a `DummyConstraint` that does nothing and holds an integer for testing purposes, and it should be replaced the first time we add a real constraint to the eDSL.

During the process I also found some things that could be optimized in the code that I was copying, so I made those changes both in merlin and scheduler.

1. Optimize docker layers & installed packages
    - Adds the `--no-install-recommends` flag to `apt install` in docker files. This is best practice for using apt in docker, but in this case the behavior is unchanged because neither curl or openjdk-17-... have recommended packages.
    - Combines the node-installation commands into one RUN layer. This should knock off a couple seconds to the `docker compose build` time, but its hard to tell because openjdk installation takes much longer than node installation.
2. Install Node in merlin-server
    - Copy the node installation part of the dockerfile.
3. Use same Node version in Docker and Gradle
    - Docker installed node v16.14.0 in the scheduler-server container, but the node plugin in Gradle used v16.4.0. I set both to 16.14.0
4. Rename typescript-compiler -> scheduler-typescript-compiler
5. Copy & rename scheduler-dsl-compiler -> constraints-dsl-compiler
    - Copies the typescript files and gradle/docker build infrastructure into merlin.
    - No actual constraints are in the AST, only the `Or` and `And` nodes.
6. Add dummy constraint
    - Adds a useless constraint to the AST. It does nothing and holds a number. This is just to have something to test against; it should be replaced by the first actual constraint.
7. Add constraint compilation service
    - Copies the compilation service into the Java side. The typescript code generation generates nothing. This includes a parser and definition for the DummyConstraint.
8. Add constraints DSL tests
    - Uses the same tests as scheduler-server, but tests against the DummyConstraint.

## Verification
This is the content of the last commit. Run with `./gradlew test`.

## Documentation
None? I don't know of any documentation for the scheduling eDSL, so I'm not sure where/how to document this.

## Future work
When a real constraint is added to the eDSL, the dummy constraint should be removed and the tests should be changed to test the real constraint.

### DIFF
--- a/merlin-server/build.gradle
+++ b/merlin-server/build.gradle
@@ -75,6 +75,7 @@ dependencies {
   implementation 'org.slf4j:slf4j-simple:1.7.26'
   implementation 'org.glassfish:javax.json:1.1.4'
   implementation 'org.apache.bcel:bcel:6.5.0'
+  implementation 'org.json:json:20171018'
 
   implementation 'com.impossibl.pgjdbc-ng:pgjdbc-ng:0.8.9'
   implementation 'com.zaxxer:HikariCP:5.0.0'
--- a/merlin-server/build.gradle
+++ b/merlin-server/build.gradle
@@ -1,6 +1,7 @@
 plugins {
   id 'java'
   id 'application'
+  id 'com.github.node-gradle.node' version '3.1.1'
 }
 
 java {
@@ -9,10 +10,51 @@ java {
   }
 }
 
+node {
+  nodeProjectDir = file("${project.projectDir.toPath().resolve("constraints-dsl-compiler")}")
+  version = '16.14.0'
+  npmVersion = '8.5.0'
+  download = true
+}
+
+task assembleConstraintsDSLCompiler(type: NpmTask) {
+  dependsOn processResources
+  dependsOn npmInstall
+  args = ['run', 'build']
+  inputs.files('package.json', 'package-lock.json', 'tsconfig.json')
+  inputs.dir('src')
+}
+
+task generateASTJsonSchema(type: NpmTask) {
+  dependsOn processResources
+  dependsOn npmInstall
+  args = ['run', 'generate-ast-schema']
+  inputs.files('package.json', 'package-lock.json', 'tsconfig.json')
+  inputs.dir('src')
+}
+
+assemble {
+  dependsOn assembleConstraintsDSLCompiler
+  dependsOn generateASTJsonSchema
+}
+
 test {
+  dependsOn(assembleConstraintsDSLCompiler)
   useJUnitPlatform {
     includeEngines 'jqwik', 'junit-jupiter'
   }
+
+  // Add node bin directory to PATH, helps CI/CD services without node installed
+  projectDir.toPath().resolve('.gradle/nodejs').toFile().listFiles().each {
+    environment 'NODE_PATH', "$it/bin/node"
+  }
+
+  if (System.getenv("NODE_PATH") == null) {
+    environment 'NODE_PATH', "node"
+  }
+
+  environment "CONSTRAINTS_DSL_COMPILER_ROOT", projectDir.toPath().resolve('constraints-dsl-compiler')
+  environment "CONSTRAINTS_DSL_COMPILER_COMMAND", './build/main.js'
 }
 
 application {
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsDSL.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsDSL.java
@@ -1,0 +1,69 @@
+package gov.nasa.jpl.aerie.merlin.server.models;
+
+import gov.nasa.jpl.aerie.json.Iso;
+import gov.nasa.jpl.aerie.json.JsonParser;
+import gov.nasa.jpl.aerie.json.ProductParsers;
+import gov.nasa.jpl.aerie.json.Unit;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.util.List;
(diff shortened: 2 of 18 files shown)


scheduler-server/scheduling-dsl-compiler/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "typescript-compiler",
+  "name": "scheduler-typescript-compiler",
```
mattdailis: 🔵 Good idea to change the name - it may even be a good idea to make it match the folder, `scheduling-dsl-compiler`

merlin-server/build.gradle
```diff
@@ -9,10 +10,51 @@ java {
...
   }
+
+  // Add node bin directory to PATH, helps CI/CD services without node installed
+  projectDir.toPath().resolve('.gradle/nodejs').toFile().listFiles().each {
+    environment 'NODE_PATH', "$it/bin/node"
+  }
+
+  if (System.getenv("NODE_PATH") == null) {
+    environment 'NODE_PATH', "node"
+  }
```
mattdailis, line 54: 🔵 (Not an issue for this PR, just jotting down some notes)

I have some reason to suspect this code isn't _quite_ doing what we want it to. I assumed that the `environment` call on line 49 would mutate the current environment in a way that would be visible to `System.getenv` - but I think that may not be true.

JoelCourtney:
@mattdailis I removed `synchronized`.

scheduler-server/build.gradle
```diff
@@ -12,7 +12,7 @@ java {
 
 node {
   nodeProjectDir = file("${project.projectDir.toPath().resolve("scheduling-dsl-compiler")}")
-  version = '16.4.0'
+  version = '16.14.0'
```
mattdailis, line 15: Good catch!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsDSL.java
```diff
@@ -0,0 +1,69 @@
...
+    return productP
+        .field("kind", literalP("ConstraintOr"))
+        .field("constraints", listP(constraintSpecifierP))
+        .map(Iso.of(untuple(($, constraints) -> new ConstraintSpecifier.ConstraintOr(constraints)),
+                    $ -> tuple(Unit.UNIT, $.constraints())));
+  }
+
+
+  private static final JsonParser<ConstraintSpecifier> constraintSpecifierP =
+      recursiveP(self -> chooseP(constraintAndF(self), constraintOrF(self), dummyConstraintDefinitionP));
```
mattdailis, line 49: 🔵 Note for the future: on the scheduling side of things, I updated this to use `sumP` instead of chooseP.

I think it will be worth looking into the `ConstraintParsers` and seeing if we can leverage those here. (Again, not for this PR, but rather for when we're actually defining the new Constraints language)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
```diff
@@ -0,0 +1,111 @@
...
+  public void close() {
+    this.nodeProcess.destroy();
+  }
+
+  /**
+   * NOTE: This method is not re-entrant (assumes only one call to this method is running at any given time)
+   */
+  public ConstraintsDSLCompilationResult compileConstraintsDSL(final PlanId planId, final String constraintTypescript, final String constraintName)
+  {
+    final var missionModelGeneratedCode = JSONObject.quote(this.typescriptCodeGenerationService.generateTypescriptTypesForPlan(planId));
```
mattdailis, line 53: 🔵 Another note for the future: this `JSONObject.quote` calls are the only reason we need the `org.json` dependency. We could refactor this to use `javax.json.JsonObjectBuilder` and remove the dependency.

mattdailis (approved):
Looks great to me! I'd be happy to merge as-is.

My comments are mostly notes for future work - none of them are blockers on this PR.

> Key:

dyst5422 (approved):

scheduler-server/scheduling-dsl-compiler/package.json
JoelCourtney: Good idea. I've squashed a couple commits into it that rename them to "scheduling-dsl-compiler" and "constraints-dsl-compiler". This is a non-functional change that doesn't effect any other code.

JoelCourtney:
The force-push is both to rebase onto develop and to rename the node packages to "scheduling-dsl-compiler" and "constraints-dsl-compiler".

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
JoelCourtney: Dropped [AERIE-1815](https://jira.jpl.nasa.gov/browse/AERIE-1815) into the backlog

merlin-server/build.gradle
JoelCourtney: Resolving this conversation so I can merge the PR. We can still discuss this though. The docs [here](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:environment) seem to say that this sets temporary environment variables that are cleared after the process ends.

PR #152, opened 2022-04-26
[AERIE-1814] Optimize computations of constraints

adrienmaillard:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

@mattdailis and I have noticed that bounds were partially present in constraints but not used. Scheduling often computes resource constraints on very restricted domains but current constraints are computed over the whole simulation results. 

I have added a `bounds` parameter to the `evaluate` method of the `Expression` interface. The previous prototype is still there as a default. This allows for a few optimizations that have been added in the operations of `LinearProfile` and `DiscreteProfile`.

I added a benchmark in which we want to compute a constraint on a very reduced interval (1/1000 of the duration of the profile), the version with time bounding is approx. 100x faster than the one without.  

All of the previously (and applicable) tests were duplicated and modified to test these optimizations.  

## Verification

WIth new tests. 

## Documentation

Look closely to the optimizations in case I missed a corner case. Or a better optimization !

## Future work

None.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
@@ -58,7 +58,8 @@ public final class LinearProfile implements Profile<LinearProfile> {
               p.valueAt(intersection.start) + o.valueAt(intersection.start),
               p.rate + o.rate
           );
-        }
+        },
+        Window.FOREVER
     );
 
     return new LinearProfile(profilePieces);
@@ -84,12 +85,15 @@ public final class LinearProfile implements Profile<LinearProfile> {
     );
   }
 
+  private static boolean profileOutsideBounds(final LinearProfilePiece piece, final Window bounds){
+    return piece.window.isStrictlyBefore(bounds) || piece.window.isStrictlyAfter(bounds);
+  }
+
   private Windows getWindowsSatisfying(final LinearProfile other, final Window bounds, final BiFunction<LinearProfilePiece, LinearProfilePiece, Windows> condition) {
     final var windows = new Windows();
-    for (final var satisfying : processIntersections(this, other, condition)) {
+    for (final var satisfying : processIntersections(this, other, condition, bounds)) {
       windows.addAll(satisfying);
     }
-
     return Windows.intersection(
         windows,
         new Windows(bounds)
@@ -105,14 +109,17 @@ public final class LinearProfile implements Profile<LinearProfile> {
 
       final var iter = this.profilePieces.iterator();
       var prev = iter.next();
-      changePoints.add(prev.changePoints());
-
+      if(!profileOutsideBounds(prev, bounds)) {
+        changePoints.add(prev.changePoints());
+      }
       while (iter.hasNext()) {
         final var curr = iter.next();
-        changePoints.add(curr.changePoints());
+        if(!profileOutsideBounds(curr, bounds)) {
+          changePoints.add(curr.changePoints());
+        }
 
         if (Window.meets(prev.window, curr.window)) {
-          if (prev.finalValue() != curr.initialValue) changePoints.add(Window.at(prev.window.end));
+          if (prev.finalValue() != curr.initialValue && !profileOutsideBounds(prev, bounds)) changePoints.add(Window.at(prev.window.end));
         } else {
           throw new Error("Unexpected gap in profile pieces not allowed");
         }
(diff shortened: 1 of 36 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ViolationsOf.java
```diff
@@ -18,10 +19,10 @@ public ViolationsOf(Expression<Windows> expression) {
   }
 
   @Override
-  public List<Violation> evaluate(SimulationResults results, Map<String, ActivityInstance> environment) {
-    final var bounds = new Windows(results.bounds);
+  public List<Violation> evaluate(SimulationResults results, final Window bounds, Map<String, ActivityInstance> environment) {
+    final var boundsW = new Windows(bounds);
     final var satisfiedWindows = this.expression.evaluate(results, environment);
```
mattdailis: 🔻 Looks like we might have missed one call to `evaluate`.

```suggestion
    final var satisfiedWindows = this.expression.evaluate(results, bounds, environment);
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -58,7 +58,8 @@ public LinearProfile plus(final LinearProfile other) {
               p.valueAt(intersection.start) + o.valueAt(intersection.start),
               p.rate + o.rate
           );
-        }
+        },
+        Window.FOREVER
```
mattdailis, line 62: 🔻 This change was made in the first commit, but I think it belongs in the second commit.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/Expression.java
```diff
@@ -2,16 +2,20 @@
...
 import gov.nasa.jpl.aerie.constraints.model.ActivityInstance;
 import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Window;
 
 import java.util.Map;
 import java.util.Set;
 
 public interface Expression<T> {
-  T evaluate(final SimulationResults results, final Map<String, ActivityInstance> environment);
+  T evaluate(final SimulationResults results, final Window bounds, final Map<String, ActivityInstance> environment);
```
mattdailis, line 11: 🔶 I do think this is a good approach - one alternative that comes to mind is to have the caller set the `SimulationResults.bounds` field before calling `evaluate`, but I think your way is better.

Tagging @kroffo and @Twisol in case they have any reactions to this.

mattdailis (changes requested):
Looks good to me overall, just one missed call to the wrong `evaluate` method.

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/model/DiscreteProfileTimeDomainTest.java
```diff
@@ -0,0 +1,103 @@
...
+    ));
+
+    final var other = new DiscreteProfile(List.of(
+        new DiscreteProfilePiece(Window.between(0, Inclusive, 5, Exclusive, SECONDS), SerializedValue.of(true)),
+        new DiscreteProfilePiece(Window.between(5, Inclusive, 10, Exclusive, SECONDS), SerializedValue.of(true)),
+        new DiscreteProfilePiece(Window.between(10, Inclusive, 15, Exclusive, SECONDS), SerializedValue.of(false)),
+        new DiscreteProfilePiece(Window.between(15, Inclusive, 20, Inclusive, SECONDS), SerializedValue.of(false))
+    ));
+
+    final var result = profile.equalTo(other, Window.between(6, 14, SECONDS));
```
mattdailis, line 31: I'm a big fan of these test cases :+1:

adrienmaillard:
I have adressed all comments made by @mattdailis in the latest push.

mattdailis (approved):
Looks good to me

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/model/DiscreteProfileTimeDomainTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.List;
+
+import static gov.nasa.jpl.aerie.constraints.Assertions.assertEquivalent;
+import static gov.nasa.jpl.aerie.constraints.time.Window.Inclusivity.Exclusive;
+import static gov.nasa.jpl.aerie.constraints.time.Window.Inclusivity.Inclusive;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECONDS;
+
+public class DiscreteProfileTimeDomainTest {
+
```
patkenneally: Extra new line

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/model/DiscreteProfileTimeDomainTest.java
```diff
@@ -0,0 +1,103 @@
+package gov.nasa.jpl.aerie.constraints.model;
+
+import gov.nasa.jpl.aerie.constraints.model.DiscreteProfile;
```
patkenneally: Inellij seems to suggest that these to imports can be removed.

patkenneally (approved):

PR #157, opened 2022-04-28
[AERIE-1811] Scheduling Language Resource Constraints

mattdailis:
* **Tickets addressed:** AERIE-1811
* **Review:** By commit  

## Description

~Opening a DRAFT PR to allow people to provide feedback before I've crossed all my "i"s and dotted my "t"s.~

This PR expands the CoexistenceGoal to be able to place activities in windows defined by state constraints.

Constraints supported:
- [x] during activity type
- [x] greater than
- [x] less than
- [x] equal to
- [x] not equal to
- [x] between
- [x] composite AND
- [x] composite OR
- [x] Transition from/to values

TODO prior to merging:
- [x] Split "constraints" DSL out from scheduling DSL into a separate TS file

## Verification

Testing is done via the SchedulingIntegrationTests.

## Documentation

When this PR is ready to be merged, we definitely need to document all of these.

## Future work

- Type check which resources can be used in numeric comparisons


deployment/hasura/metadata/actions.graphql
```diff
@@ -149,9 +149,15 @@ type SchedulingResponse {
 type SchedulingDslTypescriptResponse {
   status: SchedulingDslTypescriptResponseStatus!
   typescript: String
+  typescriptFiles: [TypescriptFile]
```
dyst5422, line 154: Very Smart!

scheduler-server/scheduling-dsl-compiler/src/libs/mission-model-generated-code.ts
```diff
@@ -9,3 +9,8 @@ export enum ActivityType {
   // This indicates to the compiler that we are using a string enum so we can assign it to string for our AST
   _ = '_',
 }
+
+export type Resource =
```
dyst5422: I think this would be a prime candidate for an enum so that they can have a good autocompletion for resources `Resource.someResource`

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -46,33 +46,72 @@ export class Goal {
...
+  private static new(windowSpecifier: AST.WindowsExpression) {
+    return new WindowSet(windowSpecifier);
+  }
+  public __serialize() {
+    return this.windowSpecifier;
+  }
+
+  public static gt(resource: Resource, value: Double): WindowSet {
+    return WindowSet.new({
+      kind: 'WindowsExpressionGreaterThan',
```
dyst5422: I think the use of the kind name in a second place here is a good motivator for making the node kinds an enum. That way we have a symbol to refactor vs having to change strings all over the place

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -65,6 +65,18 @@ public static String generateTypescriptTypesFromMissionModel(final MissionModelT
     return joinLines(result);
   }
 
+  private static String generateResourceTypes(final Collection<MissionModelService.ResourceType> resourceTypes) {
+    if (resourceTypes.isEmpty()) {
+      return "export type Resource = null;";
+    }
+    final var result = new ArrayList<String>();
+    result.add("export type Resource =");
```
dyst5422: See enum comment above

dyst5422 (changes requested):

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
mattdailis: That's a good idea - playing around with it now, it seems like intellisense works better with the enum than it did with the union type, probably because of the additional context:

Screen Shot 2022-05-04 at 12 40 01 PM

dyst5422: Might make sense to have a super intuitive mapping to a name that doesn't require the brackets and quotes - I think its not particularly intuitive to have to do the string indexing. Also not sure if that is really a viable option the way it was for command stems.

mattdailis: I would think the most intuitive form of a resource name is the exact string (any mangling we do would make it less intuitive in my opinion... but maybe that's a Basak/users question).

dyst5422: I agree that the most intuitive form of the name is the exact string, but that requires a less intuitive string indexing. I think the question is which is the lesser evil.

scheduler-server/scheduling-dsl-compiler/src/libs/mission-model-generated-code.ts
mattdailis: Resource is now a string enum 👍

patkenneally (approved):

scheduler-server/scheduling-dsl-compiler/src/libs/windows-edsl-fluent-api.ts
```diff
@@ -0,0 +1,107 @@
...
+  class WindowSet {
+    public readonly __astnode: AST.WindowsExpression
+    public and(...others: WindowSet[]): WindowSet
+    public or(...others: WindowSet[]): WindowSet
+    public static gt(resource: Resource, value: Double): WindowSet
+    public static lt(resource: Resource, value: Double): WindowSet
+    public static eq(resource: Resource, value: Double): WindowSet
+    public static neq(resource: Resource, value: Double): WindowSet
+    public static between(resource: Resource, lowerBound: Double, upperBound: Double): WindowSet
+    public static transition(resource: Resource, from: any, to: any): WindowSet
```
dyst5422: the any types here are a bit concerning. Is there anyway we can make this a type that we replace with generated code at runtime? Like we do with ActivityTypes and Resources

mattdailis: Latest commit attempts to address this problem using function overloading.

I don't want to delve too deep into it in this PR, since @JoelCourtney's work parallels this. If we decide this is a good way to do things, hopefully we can reuse some of these ideas on the Constraints side.

mattdailis: After a discussion with @dyst5422 and @adrienmaillard , I've updated this to use [Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) instead of function overloading. Let me know what you think!

I'll squash the last two commits before merging this PR, since one of them uses function overloading, and the other uses conditional types.

I'll echo the sentiment in my previous comment that this code will need to move out of the scheduler and into the Constraints eDSL during the next sprint.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -123,7 +123,7 @@ public void setName(String name) {
   protected String name = "TRE_" + Math.abs(new Random().nextInt());
   protected List<TimeRangeExpression> timeRangeExpressions;
   protected List<Object> filtersAndTransformers;
-  protected List<StateConstraintExpression> stateExpr;
+  public List<StateConstraintExpression> stateExpr;
```
adrienmaillard: Why not an accessor?

adrienmaillard (approved):
Nothing blocking for me.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
mattdailis: Updated to use an accessor :+1:

https://github.com/NASA-AMMOS/aerie/blob/4fffd68931d5a1160ae6abcefeae068de2258711/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java#L130-L132

dyst5422 (approved):
Matt and I had a quick Slack chat about how it might make sense to replace the and/or infix operators with Goal.Any([]) and Goal.All([]) collection operators, but I'll leave it up to him whether to implement that in this PR or not

mattdailis:
Last force push renames a number of things to be more similar to the constraints verbage:
- `gt` became `greaterThan`
- `lt` became `lessThan`
- `eq` became `equalTo`
- `neq` become `notEqualTo`
- `and` became `all`, and was converted to a static method
- `or` became `any` and was converted to a static method

I'll do one last survey to make sure things like `>=` and `<=` are expressible, then I'll merge.

EDIT: I suppose for the time being, it'll need to be:
```typescript
WindowSet.any(
  WindowSet.lessThan(...),
  WindowSet.equalTo(...)
)
```

Another note: discrete vs linear resources are not properly differentiated in this PR. That will come via the constraints language in an upcoming PR.

PR #158, opened 2022-04-29
[FIX] Stricter/more correct type checking

dyst5422:
* **Tickets addressed:** NA
* **Review:** By commit  

## Description

There is an issue that was identified while working on the constraint dsl. The typescript type system is structural type system and not a nominal type system. A strange side effect of this is that almost anything can be assigned to `{}` since everything in JS is an object. Anything sufficiently implements the interface `{}`.

Skip to our classes used for the EDSL - if they don't have public methods on them, then they are effectively the empty type `{}`. A user could return a number and the type checker would be fine with it.

So we thought about it from the structural interface standpoint and asked ourselves "What does something the user returns need to have to be valid". The answer is some way to generate an Abstract Syntax Tree node. In order to type check that, it needs to be publicly visible. So our previous attempts to hide the ast node in the background of the EDSL classes is actually counterproductive. We made the ast node `public readonly` and renamed it to `__astNode` to make clear that it is an internal property only and isn't part of the user api. We could then remove the serialize method as it was unnecessary.

## Verification

Pre-existing tests are all passing

## Documentation

No documentation need for this change

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/main.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/main.ts
@@ -63,13 +63,9 @@ async function handleRequest(data: string) {
       return;
     }
 
-    // We're doing a string index in order to access the private method __serialize of Goal
-    // The previous strategy of exporting a symbol and using it to get at a public method
-    // didn't work because the symbol in the vm is different than the symbol we were using
-    // here to deserialize the Goal due to different evaluation contexts
-    const stringified = JSON.stringify(result.unwrap()['__serialize']());
+    const stringified = JSON.stringify(result.unwrap().__astNode);
     if (stringified === undefined) {
-      throw Error(JSON.stringify(result.unwrap()) + ' was not JSON serializable');
+      throw Error(JSON.stringify(result.unwrap().__astNode) + ' was not JSON serializable');
     }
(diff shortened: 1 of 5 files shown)


JoelCourtney (approved):

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -67,7 +67,7 @@ async function handleRequest(data: string) {
     // The previous strategy of exporting a symbol and using it to get at a public method
     // didn't work because the symbol in the vm is different than the symbol we were using
     // here to deserialize the Goal due to different evaluation contexts
```
JoelCourtney: This comment probably needs to be updated. (specifically the first line, which github won't let me highlight.)

pcrosemurgy (approved):

mattdailis (changes requested):
Looks good to me - I'm a bit fuzzy on how classes work in terms of structural subtyping. What I'm inferring from this change is that if a field is public, it is considered part of the type, and if it's private, it isn't considered part of the type.

The only thing I'd request for this PR is that we add a test that would have failed before this change.

mattdailis (approved):
Thanks for adding the test! Looks good to me.

PR #159, opened 2022-04-29
[AERIE-1816] New constraints pipeline

JoelCourtney:
* **Tickets addressed:** AERIE-1816
* **Review:** By commit  

## Description

Splits the constraints pipeline into two pipelines so we can keep the old one enabled while we develop the new one.

Commits:

1. Remove unused goalName/constraintName arguments
2. Rework Constraints eDSL to match Expression tree
    - Until the old pipeline is in the trash, I think its better to use the same constraints Expressions for evaluation. Once the old pipeline is gone we can worry about changing the expression tree to something better suited to the eDSL if we want.
    - This commit deserializes the JSON coming back from the constraints compilation service directly into constraints Expressions.
    - Adds a `True` WindowsExpression just for the sake of having a simple terminal node in the AST to test with. I don't think it has any practical use and we'll probably delete it later.
3. Split the constraint evaluation pipeline
    - adds a `MERLIN_USE_NEW_CONSTRAINT_PIPELINE` environment variable (defaults to `false`) to decide which pipeline to use.
4. Fix constraints eDSL typechecking
    - This is actually @dyst5422 's contribution. Fixes the eDSL's ability to typecheck, which I broke by splitting the AST into two types of nodes. Thanks!
5. Move ConstraintsDSL class next to ConstraintParsers
    - Just for consistency.
6. Document constraints AST
7. Create constraints DSL parser tests

squashed:

8. Add constraints compilation shutdown hook
    - A piece of missed code that I forgot to copy from `scheduler-server`. Thanks Matt!
9. Prefix windows expression tags with "WindowsExpression"
    - (see discussion with Dylan](https://github.com/NASA-AMMOS/aerie/pull/159#discussion_r862257399)

## Verification

- the `ConstraintsDSLCompilationService` tests have been updated to match the new AST.
- Commit 7 adds tests for the JSON parsers.

## Documentation
I've added a bit of in-code docs in the constraints AST.

## Future work
- bring the eDSL up to feature parity with the old constraints pipeline
- remove the pipeline switch
- (probably) remove the `True` Expression node.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintsDSL.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintsDSL.java
@@ -1,4 +1,4 @@
-package gov.nasa.jpl.aerie.merlin.server.models;
+package gov.nasa.jpl.aerie.constraints.json;
 
 import gov.nasa.jpl.aerie.constraints.model.Violation;
 import gov.nasa.jpl.aerie.constraints.time.Windows;
@@ -14,7 +14,6 @@ import gov.nasa.jpl.aerie.json.Unit;
 import java.util.List;
 
 import static gov.nasa.jpl.aerie.json.BasicParsers.chooseP;
-import static gov.nasa.jpl.aerie.json.BasicParsers.intP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.listP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.literalP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.productP;
@@ -42,7 +41,7 @@ public class ConstraintsDSL {
                     $ -> tuple(Unit.UNIT, $.expressions)));
   }
 
-  private static JsonParser<True> trueP =
+  private static final JsonParser<True> trueP =
       productP
           .field("kind", literalP("WindowsExpressionTrue"))
           .map(
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
@@ -1,38 +1,33 @@
 export enum NodeKind {
-  DummyConstraint = 'DummyConstraint',
-  ConstraintAnd = 'ConstraintAnd',
-  ConstraintOr = 'ConstraintOr'
+  ViolationsOf = 'ViolationsOf',
+  WindowsExpressionAnd = 'WindowsExpressionAnd',
+  WindowsExpressionOr = 'WindowsExpressionOr',
+  WindowsExpressionTrue = 'WindowsExpressionTrue'
 }
 
-export type Constraint = | DummyConstraint;
+export type Constraint = | ViolationsOf;
 
-export interface DummyConstraint {
-  kind: NodeKind.DummyConstraint,
-  someNumber: number,
+export interface ViolationsOf {
+  kind: NodeKind.ViolationsOf,
+  expression: WindowsExpression,
 }
 
-export type ConstraintSpecifier =
-    | Constraint
-    | ConstraintComposition
-    ;
+export type WindowsExpression =
+  | And
+  | Or
+  | True;
 
-/**
- * Constraint Composition
- *
- * Compose constraints together
- */
-export type ConstraintComposition =
-    | ConstraintAnd
-    | ConstraintOr
-    ;
-
-export interface ConstraintAnd {
-  kind: NodeKind.ConstraintAnd,
-  constraints: ConstraintSpecifier[],
+export interface And {
+  kind: NodeKind.WindowsExpressionAnd,
+  expressions: WindowsExpression[],
 }
 
-export interface ConstraintOr {
-  kind: NodeKind.ConstraintOr,
-  constraints: ConstraintSpecifier[],
+export interface Or {
+  kind: NodeKind.WindowsExpressionOr,
+  expressions: WindowsExpression[],
+}
+
+export interface True {
+  kind: NodeKind.WindowsExpressionTrue
 }
 
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
@@ -5,28 +5,52 @@ export enum NodeKind {
   WindowsExpressionTrue = 'WindowsExpressionTrue'
 }
 
+/**
+ * Top-level constraints that directly produce violations
+ * when evaluated.
+ */
 export type Constraint = | ViolationsOf;
 
+/**
+ * Generates violations whenever its expression generates a window.
+ */
 export interface ViolationsOf {
   kind: NodeKind.ViolationsOf,
   expression: WindowsExpression,
 }
 
+/**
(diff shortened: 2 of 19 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintsDSL.java
```diff
@@ -0,0 +1,68 @@
...
+  private static JsonParser<And> andF(final JsonParser<Expression<Windows>> windowsExpressionP) {
+    return productP
+        .field("kind", literalP("And"))
+        .field("expressions", listP(windowsExpressionP))
+        .map(Iso.of(untuple(($, expressions) -> new And(expressions)),
+                    // This parser is only used for deserialization, so we don't care about the reverse operation.
+                    $ -> tuple(Unit.UNIT, $.expressions)));
+  }
+
+  private static JsonParser<Or> orF(final JsonParser<Expression<Windows>> windowsExpressionP) {
```
pcrosemurgy, line 35: Hopping on this prematurely so no rush in responding.
I think this is supposed to be `orP` (same with `andF` above being a typo for `andP`).

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -1,38 +1,57 @@
...
+
+/**
+ * Generates a violation when all of its sub-expressions generate windows.
  */
-export type ConstraintComposition =
-    | ConstraintAnd
-    | ConstraintOr
-    ;
+export interface And {
+  kind: NodeKind.And,
```
dyst5422: I think we're going to have `And`/`Or` etc AST nodes on the Goal side of things too, so are the two different `And`/`Or` nodes going to be in the same trees such that we would want to further specify here with a kind like `"WindowExpressionAnd"`. I'm inclined to say "It doesn't hurt". I think contextual differences for nodes with the same tags become very confusing (see query vs aggregation contexts for mongodb's syntax).

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
```diff
@@ -39,14 +42,29 @@ public static void main(final String[] args) {
...
 
+    final var typescriptCodeGenerationService = new TypescriptCodeGenerationService();
+
+    final ConstraintsDSLCompilationService constraintsDSLCompilationService;
+    try {
+      constraintsDSLCompilationService = new ConstraintsDSLCompilationService(typescriptCodeGenerationService);
+    } catch (IOException e) {
+      throw new Error("Failed to start ConstraintsDSLCompilationService", e);
+    }
+
```
mattdailis, line 53: To help the node process terminate cleanly, the scheduler-server has a shutdown hook that calls the `close` method on the compilation service:

https://github.com/NASA-AMMOS/aerie/blob/b8d8fe322ae688ef89ea63faf2e8c670a0c5af55/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/SchedulerAppDriver.java#L65

```suggestion
Runtime.getRuntime().addShutdownHook(new Thread(constraintsDSLCompilationService::close)); 
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintsDSL.java
mattdailis: This nomenclature seems to go back to a PR by @kroffo [[AERIE-561] Deserialize constraint definitions](https://github.jpl.nasa.gov/Aerie/aerie/pull/732)

My _guess_ is the `F` ending stands for "Function"? These are all static methods that take other parsers and combine them in various ways.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
JoelCourtney: added

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: So are you recommending changing the type names too, or just the string in the `kind` field? Personally I'd find it pretty annoying to have to type `WindowsExpression.WindowsExpressionAnd(...)`, but I'm not sure that's what you're suggesting.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintsDSL.java
JoelCourtney: Yeah, that was my assumption too. P is for parser objects, F is for functions that generate parser objects. I'm not a huge fan (I'd rather just write out `Parser`) but that's the convention.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
dyst5422: Yes, just the strings. I think the eDSL should just maintain the shorthand `.and` and `.or`.

You can look at the scheduling eDSL for examples where the and/or functions are not static, but attached to the instances.

https://github.com/NASA-AMMOS/aerie/blob/develop/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts#L22-L40

JoelCourtney: done

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -1,38 +1,57 @@
...
+  WindowsExpressionOr = 'WindowsExpressionOr',
+  WindowsExpressionTrue = 'WindowsExpressionTrue'
 }
 
-export type Constraint = | DummyConstraint;
+/**
+ * Top-level constraints that directly produce violations
+ * when evaluated.
+ */
+export type Constraint = | ViolationsOf;
```
mattdailis, line 12: Based on a conversation in the huddle last week, I think we established that constraints are allowed to be bare windows expressions.

```suggestion
export type Constraint =
  | ViolationsOf
  | WindowsExpression;
```

I'm not sure what implications this has for serialization, since WindowsExpressions don't have `__astnode`.

mattdailis (approved):
Approving on the basis of the feature flag/pipeline - of course there's more work to go in the constraints language itself.

patkenneally (approved):
Nice work!

It would be preferential for the last two commits to be rebased so that they are incorporated into the commit where their code initially enters the code base in this PR. Of course, disregard this comment if you had already planned to do so.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: Agreed. I haven't yet worked out how to do this, but that will be part of the next PR.

JoelCourtney:
> Nice work!
> 
> It would be preferential for the last two commits to be rebased so that they are incorporated into the commit where their code initially enters the code base in this PR. Of course, disregard this comment if you had already planned to do so.

Thanks! Squashed.

PR #160, opened 2022-05-02
[AERIE-1827] eDSL output the correct time format in seqJSON

goetzrrGit:
* **Tickets addressed:** AERIE-1827
* **Review:** By commit  

## Description
Talking with Michael L Schaffer and the seqJson team. NASA has a document discussing the time and duration used by every mission:

https://pds.nasa.gov/datastandards/pds3/standards/sr/Chapter07.pdf

Currently, we are storing time and duration as ISO 8601. We need to convert these times to the following formats used by Mike's team

@dyst5422 and @camargo had a chat with Mike and these were the agreed time formats below. 
```
yyyy-doyThh:mm:ss.sss
hh:mm:ss.sss
```

I am using this spec file here as a reference: [spec](https://github.jpl.nasa.gov/Aerie/aerie-seq-schema/blob/develop/schema.json)

Here is an example of expansion logic:

```
    BAKE_BREAD.relativeTiming(Temporal.Duration.from({ minutes: 5 })),
    BAKE_BREAD.absoluteTiming(Temporal.Instant.from("2022-04-20T20:17:13Z")),
    BAKE_BREAD.epochTiming(Temporal.Duration.from({ minutes: 15 })),
```

Here is the output from the expansion run:

```
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "00:05:00.000",
                    "type": "COMMAND_RELATIVE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "2022-110T20:17:13.000",
                    "type": "ABSOLUTE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    },
    {
        "id": "command",
        "steps": [
            {
                "args": [],
                "stem": "BAKE_BREAD",
                "time": {
                    "tag": "00:15:00.000",
                    "type": "EPOCH_RELATIVE"
                },
                "type": "command",
                "metadata": {}
            }
        ],
        "metadata": {}
    }
```

## Verification
I ran through the whole expansion process to generate the above seqJson.

## Documentation
None that I can think of. 

## Future work
Continue to add missing fields in the seqJson generated by sequence expansions.

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -39,17 +39,41 @@ export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}>
     }
   }
 
+  // return  yyyy-doyThh:mm:ss.sss
+  private instantToDoy(time: Temporal.Instant): string {
+    const utcZonedDate = time.toZonedDateTimeISO('UTC');
+    return `${utcZonedDate.year}-${this.format(utcZonedDate.dayOfYear, 3)}T${this.format(
+      utcZonedDate.hour,
+      2,
+    )}:${this.format(utcZonedDate.minute, 2)}:${this.format(utcZonedDate.second, 2)}.${this.format(
+      utcZonedDate.millisecond,
+      3,
+    )}`;
+  }
+
+  // return hh:mm:ss.sss
+  private durationToHms(time: Temporal.Duration): string {
+    return `${this.format(time.hours, 2)}:${this.format(time.minutes, 2)}:${this.format(time.seconds, 2)}.${this.format(
+      time.milliseconds,
+      3,
+    )}`;
+  }
+
+  private format(number: number, size: number): string {
+    return number.toString().padStart(size, '0');
+  }
+
   public toSeqJson() {
     return {
       args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
       stem: this.stem,
       time: {
         tag: this.absoluteTime
-          ? this.absoluteTime.toString()
+          ? this.instantToDoy(this.absoluteTime) // yyyy-doyThh:mm:ss.sss
           : this.relativeTime
-          ? this.relativeTime.toString()
+          ? this.durationToHms(this.relativeTime) // hh:mm:ss.sss
           : this.epochTime
-          ? this.epochTime.time.toString()
+          ? this.durationToHms(this.epochTime.time) // hh:mm:ss.sss
           : '',
         type: this.absoluteTime
           ? 'ABSOLUTE'


dyst5422:
At some point (not in this PR necessarily) we should add some helper template strings so that timing can be specified in a more user friendly way.

See https://gist.github.com/dyst5422/31f71f64df3a1a85a0405672f95ee8aa as an example

This will allow for time specification like

```ts
A`2022-001T00:00:00`.BAKE_BREAD,
R`00:40:00         `.PREHEAT_OVEN(5),
C`                 `.PREHEAT_OVEN(5),
E`00:30:00         `.BAKE_BREAD,
```

goetzrrGit:
> At some point (not in this PR necessarily) we should add some helper template strings so that timing can be specified in a more user friendly way.
> 
> See https://gist.github.com/dyst5422/31f71f64df3a1a85a0405672f95ee8aa as an example
> 
> This will allow for time specification like
> 
> ```ts
> A`2022-001T00:00:00`.BAKE_BREAD,
> R`00:40:00         `.PREHEAT_OVEN(5),
> C`                 `.PREHEAT_OVEN(5),
> E`00:30:00         `.BAKE_BREAD,
> ```

Agreed, I have a ticket for v0.13.0 to address this and other changes we might want.
https://jira.jpl.nasa.gov/browse/AERIE-1824

dyst5422:
@goetzrrGit on the example above, the HMS format doesn't match the HMS format we have stated above. It looks like it uses a colon to separate the seconds and subseconds rather than a period. Is that intentional?

goetzrrGit:
> @goetzrrGit on the example above, the HMS format doesn't match the HMS format we have stated above. It looks like it uses a colon to separate the seconds and subseconds rather than a period. Is that intentional?

Definitely not intentional. Good catch!!!

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -39,17 +39,42 @@ export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}>
     }
   }
 
+  private buildSeqTime(time: Temporal.Instant | Temporal.Duration): string {
```
mattdailis: 🔵 I'm wondering about the pros/cons of having one `buildSeqTime` function, as opposed to splitting it into two functions, one for absolute time, and one for duration.

At the call site, we already know which one it is, and can call the correct function.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -39,17 +39,42 @@ export class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}>
...
+    // return hh:mm:ss.sss
+    else {
+      return `${this.format(time.hours, 2)}:${this.format(time.minutes, 2)}:${this.format(
+        time.seconds,
+        2,
+      )}.${this.format(time.milliseconds, 3)}`;
+    }
+  }
+
+  private format(number: number, size: number = 2): string {
```
mattdailis: 🔵 I'm curious about the default size - why `2`? I would vote for not having a default size.

mattdailis (approved):
Approving, because none of my questions need to block merge

> Key:

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I think a pro was combining the time-building logic into one method that can be used by different time objects. We can expand this method to accept another time format in the future. It is also compact. A con (isn't really that bad) is to have two or more time-building methods so slightly more code with method declaration.  

If the time conversion logic was really complicated then I would have split it into two methods but I felt it wasn't too bad to read.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Good point, I will remove it cause I am overriding the default anyways when calling this method.

mattdailis (approved):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: Now that matt mentions it, I agree with him. I think this generic time building function is pushing context about the time format down into a function that shouldn't know about that context.

The place its called (toSeqJson) knows what formats it needs and should request those formats rather than delegating that knowledge elsewhere.

I'd recommend having two functions - durationToHms and instantToDoy

dyst5422 (changes requested):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Sounds good, pretty easy change on my end, I retested and pushed the changes to the PR

dyst5422 (approved):

goetzrrGit:
Thanks for the review @mattdailis and @dyst5422

PR #163, opened 2022-05-03
[FIX][AERIE-1829] Satisfying activities missing 

adrienmaillard:
* **Tickets addressed:** AERIE-1829
* **Review:** By commit  

## Description

Activities that cannot be associated to their creator goal (which happens if you run the scheduler several times in the same plan as there is no tracking what activity were created by which goal in previous runs) were considered to have Solely custody. 

If you ran the scheduler twice without activity created, you would not see the satisfying activities associated to the goal. 

We now allow goal-activity association even if the activity does not have an explicit creator goal defined.   

## Verification

With a test specifically designed to reproduce the issue in `SimulationFacadeTest`.

## Documentation

None.

## Future work

If goal/activity custody is going to be a feature, there will be a need to track from run to run what goals created which activity. And think about use cases where for example a user creates a goal, schedules activities and then deletes the goal. What happens to these activities and their custody?

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
@@ -58,7 +58,11 @@ public class TimeRangeExpression {
                 Window.Inclusivity.Exclusive)).build();
         final var anchorActs = plan.find(anchorActSearch);
         for (var anchorAct : anchorActs) {
-          actTw.add(Window.between(anchorAct.getStartTime(), Window.Inclusivity.Inclusive, anchorAct.getEndTime(), Window.Inclusivity.Exclusive));
+          var endInclusivity = Window.Inclusivity.Exclusive;
+          if(anchorAct.getDuration().isZero()){
+            endInclusivity = Window.Inclusivity.Inclusive;
+          }
+          actTw.add(Window.between(anchorAct.getStartTime(), Window.Inclusivity.Inclusive, anchorAct.getEndTime(), endInclusivity));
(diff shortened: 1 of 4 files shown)


mattdailis (approved):

camargo:
When trying to run this recurrence goal twice on an empty plan:

```ts
export default (): Goal => Goal.ActivityRecurrenceGoal({
  activityTemplate: ActivityTemplates.BakeBananaBread({ temperature: 325.0, tbSugar: 2, glutenFree: false }),
  interval: 12 * 60 * 60 * 1000 * 1000 // 1/2 day in microseconds
})
```

I'm now seeing this error in the scheduler:

```json
{"status":"failed","reason":"Cannot compute end time: activity instance does not have a duration yet"}
```

Can anyone reproduce on their end?

adrienmaillard:
> When trying to run this recurrence goal twice on an empty plan:
> 
> ```ts
> export default (): Goal => Goal.ActivityRecurrenceGoal({
>   activityTemplate: ActivityTemplates.BakeBananaBread({ temperature: 325.0, tbSugar: 2, glutenFree: false }),
>   interval: 12 * 60 * 60 * 1000 * 1000 // 1/2 day in microseconds
> })
> ```
> 
> I'm now seeing this error in the scheduler:
> 
> ```json
> {"status":"failed","reason":"Cannot compute end time: activity instance does not have a duration yet"}
> ```
> 
> Can anyone reproduce on their end?

Activities of the initial plan were simulated but their durations were not pulled back from the simulation which caused this error. It has been fixed by the second commit of this PR.

mattdailis (approved):

camargo (approved):
LGTM. Tested and works on front-end.

PR #164, opened 2022-05-04
[AERIE-1794] Accumulate & persist unfinished activities

pcrosemurgy:
* **Tickets addressed:** AERIE-1794
* **Review:** By file  

## Description

Persists unfinished activities without our Postgres DB.

Suggesting to review by file due to the nature of the commit in this PR. Changing `SimulationResults` to contain `UnfinishedActivity` propagates to many places, so instead of having non-compilable intermediate commits I decided to keep this as one commit.

## Verification

Activities with `null` durations now denote unfinished activities:
Screen Shot 2022-05-04 at 1 44 39 PM

Inserting `GrowBanana` at the end of a plan with a duration that surpasses the plan end results in:
```json
{
  "data": {
    "simulate": {
      "results": {
        "unfinishedActivities": {
          "1": {
            "parent": null,
            "arguments": {
              "quantity": 0,
              "growingDuration": 10000000000
            },
            "children": [],
            "type": "GrowBanana",
            "startTimestamp": "2022-001T22:25:50.239000"
          }
        },
        "start": "2022-001T00:00:00.000000",
...
```
Note that this PR probably didn't need to even add `unfinishedActivities` to the sim. response JSON since it will be removed by https://jira.jpl.nasa.gov/browse/AERIE-1796 (see future work section). However, no harm done here – it's kind of nice to be able to see unfinished activities right in the HTTP response while testing this PR.

## Documentation

https://github.com/NASA-AMMOS/aerie/wiki/Simulation-Results#unfinished-activities currently contains up-to-date information regarding unfinished activities.

## Future work

- Carve out simulation HTTP response JSON (see https://jira.jpl.nasa.gov/browse/AERIE-1796).

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
@@ -1,9 +1,9 @@
 package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
 
 import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
-import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
 import gov.nasa.jpl.aerie.merlin.driver.SimulatedActivity;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
+import gov.nasa.jpl.aerie.merlin.driver.UnfinishedActivity;
 import gov.nasa.jpl.aerie.merlin.driver.timeline.EventGraph;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
@@ -307,19 +307,15 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
     final var simulationStart = startTimestamp.toInstant();
 
     final var profiles = ProfileRepository.getProfiles(connection, simulationDatasetRecord.datasetId(), simulationWindow);
-    final var activities = getSimulatedActivities(connection, simulationDatasetRecord.datasetId(), startTimestamp);
-
-    // TODO: Currently we don't store unfinished activities, but when we do we'll have to update this
-    final Map<ActivityInstanceId, SerializedActivity> unfinishedActivities = Map.of();
-
+    final var activities = getActivities(connection, simulationDatasetRecord.datasetId(), startTimestamp);
     final var topics = getSimulationTopics(connection, simulationDatasetRecord.datasetId());
     final var events = getSimulationEvents(connection, simulationDatasetRecord.datasetId(), startTimestamp);
 
     return new SimulationResults(
         profiles.realProfiles(),
         profiles.discreteProfiles(),
-        activities,
-        unfinishedActivities,
+        activities.getLeft(),
+        activities.getRight(),
         simulationStart,
         topics,
         events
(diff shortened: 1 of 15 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostActivitiesAction.java
```diff
@@ -17,38 +17,42 @@
...
+        final var actEnd = act.start().plus(duration.dividedBy(Duration.MICROSECOND), ChronoUnit.MICROS);
+        return new Timestamp(actEnd);
+      });
 
       statement.setLong(1, datasetId);
       setTimestamp(statement, 2, startTimestamp);
       setTimestamp(statement, 3, simulationStart);
-      setTimestamp(statement, 4, endTimestamp);
+
+      if (endTimestamp.isPresent()) setTimestamp(statement, 4, endTimestamp.get());
```
mattdailis: I'm a bit unclear on what happens when endTimestamp is empty - I would think that not providing a value for slot number 4 would cause some sort of error...?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -351,33 +348,69 @@ private static Map<ActivityInstanceId, SimulatedActivity> getSimulatedActivities
...
+    }
+  }
+
+  private static Map<ActivityInstanceId, UnfinishedActivity> getUnfinishedActivities(
+      final Connection connection,
+      final long datasetId,
+      final Timestamp startTime
+  ) throws SQLException {
+    try (final var getActivitiesAction = new GetActivitiesRecord(connection)) {
+      final var activityRecords = getActivitiesAction.get(datasetId, startTime);
```
mattdailis: I'm noticing that we're calling the same action twice in a row. (not necessarily a bad thing, but it makes me pause) 

I wonder if it would make sense to either
1. Make separate actions for getting finished and unfinished activities (and filter for null durations in the sql itself)
2. Or, leave the action as is, but only call it once, and share the response where needed.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ActivityRecord.java
```diff
@@ -8,13 +8,13 @@
 import java.util.Map;
 import java.util.Optional;
 
-/*package-local*/ record SimulatedActivityRecord(
+/*package-local*/ record ActivityRecord(
```
mattdailis: I see the desire to distinguish between finished and unfinished activities - but to me `ActivityRecord` is too ambiguous (we need to distinguish between activities in the plan vs activities that have been simulated)

I would propose that we leave the term `SimulatedActivities` as it was, but where necessary draw the distinction between completed and unfinished activities.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostActivitiesAction.java
pcrosemurgy: When `endTimestamp` is empty the DB's `span.duration` value will be null :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
pcrosemurgy: Good catch! This would result in a performance hit for very large plans.

I'll go for option 2 because I'd like to embed as little domain-logic into our SQL actions as possible.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ActivityRecord.java
pcrosemurgy: My reasoning here was that an unfinished activity is not necessarily a simulated activity (`SimulatedActivityRecord`). The record just needs to have a fairly 1-1 correspondence with what's inside the DB, which can be a simulated or an unfinished activity depending on the presence of a `duration`.

Let me know how you feel about that. I'm open to keeping this `SimulatedActivityRecord` if that's preferred :)

mattdailis: Hmm, I'm having trouble wrapping my head around `an unfinished activity is not necessarily a simulated activity`, maybe we have slightly different definitions of those two terms in mind. I think I'm thinking of "unfinished" as "partially simulated". How does that agree/differ from your mental model?

pcrosemurgy: Yeah, to me "partially simulated" is not "simulated", in the same sense that something "partially complete" would not be called "complete".

pcrosemurgy: This naming mostly just keeps with the convention within `SimulationResults` where `simulatedActivities` and `unfinishedActivities` are the names used.

mattdailis: I see - thanks for helping me understand - now our mental models are aligned, if not our nomenclature :+1:

I'm happy to agree to disagree on this one - I'll just write down my thought process

The concepts that I'm wrapping my brain around are:
1. Activity types
2. Activity instances in the plan
3. The result of simulating an activity instance that has not completed
4. The result of simulating an activity instance that has completed

The postgres action representations of these concepts in this PR are:
1. `String` (also stored in the `activity_type` table w/ValueSchema)
2. `ActivityInstanceRecord`
3. `ActivityRecord` where `duration` and `computedAttributes` are both `Optional.empty()`
4. `ActivityRecord` where `duration` and `computedAttributes` are both `Optional.of(...)`

So - I guess I feel like 3 and 4 should somehow signify that they're the result of running a simulation, as opposed to 1 and 2 🤔 Out of context I would think that `ActivityRecord` refers to `ActivityInstanceRecord`...

Perhaps `SimulatedOrPartiallySimulatedActivityRecord` 😅

patkenneally: My 2 cents; I think we do need to indicating that the record's usage context is downstream from a simulation (whether completed or uncompleted). I'm happy to find a new name, but in the instance that we can't, I think staying with `SimulatedActivityRecord` should be our fallback position.

adrienmaillard (approved):
LGTM

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ActivityRecord.java
pcrosemurgy: >So - I guess I feel like 3 and 4 should somehow signify that they're the result of running a simulation, as opposed to 1 and 2

Ahh I see your angle here. I avoided this since I wanted to keep as much domain logic out of our "get activities action" as possible (such as determining whether an activity w/ a `null` duration is unfinished or not). If you're comfortable with that logic residing within the PG action then I can make that change :)

No super strong opinion from me, just trying to stick with convention, which seems to avoid embedding domain logic and exposing `*Record` DB data with a 1-1 mapping with the associated DB table. This is why I opted for embedding this duration `null` check logic in the PG results cell repo.

>My 2 cents; I think we do need to indicating that the record's usage context is downstream from a simulation (whether completed or uncompleted).

In this PR the simulated and unfinished activities are stored respectively in `SimulationResults.simulatedActivities` and `SimulationResults.unfinishedActivities`, so this information is available downstream from a simulation.

pcrosemurgy: @mattdailis Also note that 2 and 3 above are for different use cases (one is grabbing activity instances from a plan, the other is grabbing activities recorded within a sim.'s `span`).

mattdailis: > @mattdailis Also note that 2 and 3 above are for different use cases (one is grabbing activity instances from a plan, the other is grabbing activities recorded within a sim.'s span).

☝️ yes, exactly - I think my point is I'd like to tag the record classes with the use case they correspond to. Here's where my confusion comes from - when I open up the `postgres` package, I see `ActivityInstanceRecord` and `ActivityRecord`, and I pause - what's the difference between an Activity Instance and an Activity?

Screen Shot 2022-05-11 at 3 33 44 PM

mattdailis: Conceivably, we could have the Records match the table names, opting for `SpanRecord`

pcrosemurgy: Big thanks to @mattdailis for taking the time to help nail this down. We've settled on returning `SpanRecord`s from the DB. If/when "span" gets renamed this can also be renamed, but that's a separate topic.

Latest push attempts to make a better a distinction between planned activities (in `ActivityInstanceRecord`) and spans (simulated/unfinished activities, in `SpanRecord`).

mattdailis (approved):
Great work, Paul

PR #166, opened 2022-05-04
[AERIE-1823]  sequence generation flow

dyst5422:
* **Tickets addressed:** AERIE-1823
* **Review:** By commit  

## Description

Get the specified sequence + simulated activity ids + commands from the latest expansion run for each simulated activity (filtered on simulation dataset)
Get start time for each simulated activity and join with simulated activity command data
Call the seqbuilder on the sorted simulated activity to generate the sequence
Return the sequence as seqJson with sequence#toSeqJson

## Verification

Ran through testing locally.

## Documentation

[Create Sequence](https://github.com/NASA-AMMOS/aerie/wiki/Create-Sequence)
[Link Simulated Activity to Sequence](https://github.com/NASA-AMMOS/aerie/wiki/Link-Simulated-Activity-to-Sequence)
[Retrieve SeqJson Serialization of Sequence](https://github.com/NASA-AMMOS/aerie/wiki/Retrieve-SeqJson-Serialization-of-Sequence)

## Future work

Before merging:
- [x] test simple concat
- [x] test with non-existent expansion
- [x] test with expansions throwing errors


dyst5422:
Apparently can't convert to draft, so applied the don't merge tag instead

command-expansion-server/src/app.ts
```diff
@@ -333,6 +339,58 @@ app.post('/expand-all-activity-instances', async (req, res, next) => {
...
+  const seqId = req.body.input.seqId as string;
+  const simulationDatasetId = req.body.input.simulationDatasetId as number;
+
+  const { rows } = await db.query<{
+    metadata: Record<string, unknown>;
+    commands: ReturnType<Command['toSeqJson']>,
+    activity_instance_id: number,
+    errors: string,
+  }>(`
+    select distinct on (activity_instance_commands.expansion_run_id) sequence.metadata, activity_instance_commands.commands, activity_instance_commands.activity_instance_id, activity_instance_commands.errors
```
camargo: Not necessarily this PR but the formatting for these queries is inconsistent from endpoint to endpoint in this file which might make longterm maintenance harder. Can we switch to use a consistent format? (caps vs lowercase, indentation, etc.)

dyst5422: I was thinking about that the whole time I was working on this and had to resist the temptation for refactoring scope creep. When we get the rest of the PR finished up, I'll do a refactoring pass. + commit

goetzrrGit: Are we using the expansion errors for anything? I can see adding the generated error message into the seqjson, but maybe that is a to-do for later as I am not sure where we would stick those errors in the seqJSON besides a `description` field.

command-expansion-server/src/app.ts
```diff
@@ -333,6 +339,58 @@ app.post('/expand-all-activity-instances', async (req, res, next) => {
...
+  }
+  const seqMetadata = rows[0].metadata;
+  const sortedActivityInstances = (activityInstances as Exclude<typeof activityInstances[number], Error>[])
+      .map(ai => ai[0])
+      .sort((a, b) => Temporal.Duration.from(parseInterval(a.start_offset).toISOString()).nanoseconds - Temporal.Duration.from(parseInterval(b.start_offset).toISOString()).nanoseconds)
+
+  const sortedActivityInstancesWithCommands: (SimulatedActivityInstance<{[p: string]: unknown}, {[p: string]: unknown}> & { commands: ReturnType<Command['toSeqJson']> })[] = sortedActivityInstances.map(ai => ({ ...ai, commands: assertDefined(rows.find(row => row.activity_instance_id === ai.id)).commands }));
+
+  // TODO: Come up with a way to make this configurable and load that configured seq builder if present
+  const seqBuilder = defaultSeqBuilder;
```
goetzrrGit, line 509: Good idea to make this configurable

goetzrrGit (commented):
Amazing work Dylan!!! I still need to test it though. I just had some generic questions.

command-expansion-server/src/app.ts
dyst5422: I was just talking to matt about this. Can we put them into the seqJson simply under a non-existent command stem:
```
{
  stem: "$$ERROR$$",
  args: ["Error message here"],
  time: {
    tag: <activity start time>,
    type: "ABSOLUTE"
  },
  metadata: {}
}
```

command-expansion-server/src/app.ts
dyst5422: Discussed with Pat and Matt and while this is useful, until there is a request for doing more complicated things when building the sequence, we won't be implementing this. And in fact, this might force that conversation of whether we should be doing sequence building at all and encourage them to just build a couple hundred line python script to pull from the api and do the building themselves.

command-expansion-server/src/app.ts
```diff
@@ -54,7 +62,10 @@ app.use(async (req: Request, res: Response, next: NextFunction) => {
     activitySchemaDataLoader: new DataLoader(activitySchemaBatchLoader({ graphqlClient }), {
       cacheKeyFn: objectCacheKeyFunction,
     }),
-    simulatedActivityInstanceDataLoader: new DataLoader(simulatedActivityInstanceBatchLoader({ graphqlClient }), {
+    simulatedActivitiesDataLoader: new DataLoader(simulatedActivitiesBatchLoader({ graphqlClient }), {
+      cacheKeyFn: objectCacheKeyFunction,
+    }),
+    simulatedActivityInstanceBySimulatedActivityIdDataLoader: new DataLoader(simulatedActivityInstanceBySimulatedActivityIdBatchLoader({ graphqlClient }), {
```
patkenneally: :large_blue_circle: What are your thoughts on reformatting these long lines so they are initially visible without a wide monitor or an IDE auto-wrapping? 

Legend:

command-expansion-server/src/app.ts
```diff
@@ -353,6 +364,117 @@ app.post('/expand-all-activity-instances', async (req, res, next) => {
   return next();
 });
 
+export interface SeqBuilder {
```
patkenneally, line 396: :large_blue_circle: Is this commit both fixing simulated activity batch loaders updated in the last commit and adding the SeqBuilder capability? I want to make sure I'm understanding the ideas/intentions introduced in this commit given the commit message.

command-expansion-server/src/app.ts
```diff
@@ -14,7 +14,8 @@ import { commandDictionaryTypescriptBatchLoader } from './lib/batchLoaders/comma
 import { activitySchemaBatchLoader } from './lib/batchLoaders/activitySchemaBatchLoader.js';
```
patkenneally, line 14: :large_blue_circle: Seems like this commit has the message that should go with some of the code that enters the code base in [ff6fdf](https://github.com/NASA-AMMOS/aerie/pull/166/commits/ff6fdf9141556131dcc0b6081fb93b360dc171c5)

patkenneally (approved):

command-expansion-server/test/sequence-generation.spec.ts
```diff
@@ -0,0 +1,242 @@
+import { gql, GraphQLClient } from 'graphql-request';
```
mattdailis, line 1: This test seems like it would do well as part of our `e2eTest` suite that runs on every PR.

mattdailis (approved):

command-expansion-server/src/app.ts
dyst5422: Definitely a rebase error. I think all of this should be in 8128bba6d51d9e9b362a5b649c8bc99fefd241d8.

dyst5422:
@mattdailis & @patkenneally sorry for the churn, but the easiest way to resolve the various rebasing issue was to just rewrite the whole commit history with a reset and cherry picking for a more cohesive commit story.

mattdailis (approved):
> @mattdailis & @patkenneally sorry for the churn, but the easiest way to resolve the various rebasing issue was to just rewrite the whole commit history with a reset and cherry picking for a more cohesive commit story.

No worries - I too got confused by the commits while reviewing, so I think the re-write was a good idea :+1:

command-expansion-server/src/app.ts
dyst5422: Did a full auto-format with prettier which also addressed this

dyst5422:
rebase on develop + fix up tests

command-expansion-server/test/sequence-generation.spec.ts
dyst5422: Would love to add this in. Not sure how to and won't hold up this PR for it.

PR #167, opened 2022-05-04
[AERIE-1813] Replace scheduler state constraints by merlin constraints

adrienmaillard:
* **Tickets addressed:** AERIE-1813
* **Review:** By commit  

## Description

There are 7 commits. I apologize in advance for the third one (3ecbf84f) that is the main one and contains all of the important changes of this PR. State constraints were so central in the scheduler I have not found it to be possible to make more incremental changes. 

Basically we replaced `StateConstraintExpression` with `Expression<Windows>` everywhere in the scheduler. Here are the consequences:
1. `SimulationResults` are needed pretty much everywhere we evaluate scheduler-specific expressions such as `TimeExpression`, `TimeRangeExpression`, or `DurationExpression`... so it had to be added to all of these interfaces. 
2. Before, for each resource used in a scheduler construct, we created actual objects (`SimResource`, implementation of `ExternalState`) containing resource values derived from simulation results. And these objects would be linked to a `SimulationFacade`. Every time an expression would be evaluated for some given domain, the resource would check whether it contained enough data to evaluate the expression and if not would ask the simulation to simulate until it has enough data. This was kind of a dynamic but hidden process, isolated from the solver. Now, the data flow is more explicit, more top/down, as the origin of the simulation results used to evaluate an expression is explicitly mentioned in the methods. This resulted in changes in the `SimulationFacade` (mostly deletions as we do not need the interface with SimResource anymore).
3. Two features that I have not found in the constraints package have been relocated from `ExternalState`: (a) computing the value of a resource at some time `t` has now been relocated into `StateQueryParam`. (b) computing the concatenated timeline representing the intervals during which a discrete resource is constant (as a Windows made of totally disjoint intervals) has been relocated to `TimeRangeExpression`. 
4. Almost of the tests are now linked with a mission model. 

## Verification

Many tests have been updated.  

## Documentation

No documentation has been invalidated. 

## Future work

- performance testing now that the simulation flow has been modified. Make sure we simulate adequately whenever we make changes to the plan while not simulating too many times.


mattdailis:
This one might be worth a walkthrough/guided discussion. I'm also curious about its implications for #157

mattdailis:
I tried rebasing this PR onto #157 locally - here's what I came up with for the `GoalBuilder`. Note the `NotImplementedException` - if we use your `WindowsOf` technique from #171 we could do a proper conversion here. (Although we should think carefully about whether all "ActivityExpressions" are supported in the constraints language... I don't think we can specify arguments yet).

```java
private static TimeRangeExpression timeRangeExpressionOfConstraintExpression(
    final SchedulingDSL.ConstraintExpression constraintExpression,
    final Function<String, ActivityType> lookupActivityType) {
  if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.ActivityExpression c) {
    return new TimeRangeExpression.Builder()
        .from(ActivityExpression.ofType(lookupActivityType.apply(c.expression().type())))
        .build();
  } else {
    return new TimeRangeExpression.Builder()
        .from(expressionOfConstraintExpression(constraintExpression))
        .build();
  }
}

private static Expression<Windows> expressionOfConstraintExpression(
    final SchedulingDSL.ConstraintExpression constraintExpression) {
  if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.ActivityExpression c) {
    throw new NotImplementedException("Nested ActivityExpressions are not yet supported");
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.GreaterThan c) {
    return new GreaterThan(new RealResource(c.resource().name()), new RealValue(c.value()));
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.LessThan c) {
    return new LessThan(new RealResource(c.resource().name()), new RealValue(c.value()));
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.EqualLinear c) {
    return new Equal<>(new RealResource(c.resource().name()), new RealValue(c.value()));
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.NotEqualLinear c) {
    return new NotEqual<>(new RealResource(c.resource().name()), new RealValue(c.value()));
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.Between c) {
    return new And(new GreaterThan(new RealResource(c.resource().name()), new RealValue(c.lowerBound())),
                   new LessThan(new RealResource(c.resource().name()), new RealValue(c.upperBound())));
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.Transition c) {
    return new Transition(new DiscreteResource(c.resource().name()), c.from(), c.to());
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.And c) {
    return new And(c.expressions()
                    .stream()
                    .map(GoalBuilder::expressionOfConstraintExpression)
                    .toList());
  } else if (constraintExpression instanceof SchedulingDSL.ConstraintExpression.Or c) {
    return new Or(c.expressions()
                   .stream()
                   .map(GoalBuilder::expressionOfConstraintExpression)
                   .toList());
  } else {
    throw new UnexpectedSubtypeError(SchedulingDSL.ConstraintExpression.class, constraintExpression);
  }
}

```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/resources/StateQueryParam.java
```diff
@@ -1,28 +1,61 @@
...
+      //TODO: improve the profile data structure to allow fast time-keyed query
+      //TODO: improve Window to allow querying containment in interval directly
+      //for now we look for the last matching profile segment, if any (last to get latest if any overlaps)
+      final var profile = simulationResults.realProfiles.get(this.nameResource);
+      final Predicate<LinearProfilePiece> containsQueryTimeP = (piece)
+          -> !Window.intersect(piece.window, Window.at(queryT)).isEmpty();
+      final var matchPiece = Lists.reverse(profile.profilePieces).stream()
+                                  .filter(containsQueryTimeP).findFirst();
+      final var dblVal = matchPiece.map(piece -> piece.valueAt(queryT)).orElse(null);
+      return SerializedValue.of(dblVal);
```
mattdailis: Rather than throw a null pointer exception (which is what will happen if we call `SerializedValue.of(null)`), we can throw an Error:

```suggestion
      final var piece = Lists.reverse(profile.profilePieces)
                             .stream()
                             .filter(containsQueryTimeP)
                             .findFirst()
                             .orElseThrow(() -> new Error(
                                 "Linear profile for %s not have a segment at the desired time %s"
                                     .formatted(nameResource, queryT)));
    return SerializedValue.of(piece.valueAt(queryT));
```

The other cases below should also probably throw errors - alternatively, this method could return an `Optional`, if it's not truly an error situation.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/GlobalConstraints.java
```diff
@@ -1,13 +1,19 @@
...
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityExpression;
+import gov.nasa.jpl.aerie.scheduler.model.ActivityType;
 
 import java.util.List;
 
 public class GlobalConstraints {
 
-  public static NAryMutexConstraint atMostOneOf(List<ActivityExpression> types) {
-    return NAryMutexConstraint.buildMutexConstraint(types);
+  public static NAryMutexConstraint atMostOneOf(ActivityExpression... types) {
```
mattdailis, line 10: I support this change, but I don't think it's relevant to the giant commit. (Not requesting a change necessarily, just commenting as I read through)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/filters/FilterEverViolated.java
```diff
@@ -1,23 +1,24 @@
...
-  public FilterEverViolated(StateConstraintExpression expr) {
+  public FilterEverViolated(final Expression<Windows> expr) {
     this.expr = expr;
   }
 
   @Override
-  public boolean shouldKeep(Plan plan, Window range) {
-    return !(expr.findWindows(plan, new Windows(range)).equals(new Windows(range)));
+  public boolean shouldKeep(final SimulationResults simulationResults, final Plan plan, final Window range) {
+    return !(expr.evaluate(simulationResults, range, null).equals(new Windows(range)));
```
mattdailis: The "empty" case for environment is an empty map, not `null`.
```suggestion
    return !(expr.evaluate(simulationResults, range, Map.of()).equals(new Windows(range)));
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -30,91 +30,78 @@ public class TimeRangeExpression {
...
-      if (inter.isEmpty()) {
-        break;
-      }
     }
 
     for (var expr : stateExpr) {
-      Windows windowsState = expr.findWindows(plan, domain);
+      if(inter.isEmpty()) break;
+      final var domainOfInter = Window.between(inter.minTimePoint().get(), inter.maxTimePoint().get());
+      Windows windowsState = expr.evaluate(simulationResults, domainOfInter, null);
```
mattdailis: The "empty" case of environment is an empty Map
```suggestion
      Windows windowsState = expr.evaluate(simulationResults, domainOfInter, Map.of());
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/NAryMutexConstraint.java
```diff
@@ -59,39 +58,13 @@ private Windows findWindows(Plan plan, Windows windows, ActivityType actToBeSche
...
-        }
-        return cState;*/
-    return null;
+  @Override
+  public ConstraintState isEnforced(
+      final Plan plan,
+      final Windows windows,
+      final SimulationResults simulationResults)
+  {
+    throw new IllegalArgumentException("Not implemented");
```
mattdailis: I don't mind either way, but I just want to point out that there exists a `NotImplementedException` for this purpose.

mattdailis (commented):
I'm still working through this PR, but in the interest of getting you some feedback sooner, here are the comments I have so far

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -123,19 +110,19 @@ public void setName(String name) {
...
-    return List.copyOf(stateExpr);
+  public static TimeRangeExpression constantValuesOf(String nameDiscreteProfile) {
+    return new Builder().ofEachValue(nameDiscreteProfile).build();
   }
 
-  public static TimeRangeExpression constantValuesOf(ExternalState sce) {
-    return new Builder().ofEachValue(sce).build();
+  public List<Expression<Windows>> getStateConstraints() {
+    return List.copyOf(stateExpr);
   }
```
mattdailis: We can remove `getStateConstraints` now - we only needed it because the `GoalBuilder` needed to build an "Or" state constraint - but that's done via `Expression<Windows>` now.

```suggestion
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -572,28 +571,26 @@ private void satisfyGoalGeneral(Goal goal) {
...
+      resourceConstraints.add(goalConstraints);
     }
     if (missing instanceof final MissingActivityInstanceConflict missingInstance) {
       final var act = missingInstance.getInstance();
-      actType = act.getType();
-      StateConstraintExpression c = act.getType().getStateConstraints();
-      if (c != null) stateConstraints.add(c);
+      final var c = act.getType().getStateConstraints();
+      if (c != null) resourceConstraints.add(c);
     } else if (goal instanceof ActivityTemplateGoal) {
```
mattdailis: Not a blocker - just a style suggestion - instead of casting to `((ActivityTemplateGoal) goal)` below, we can define a new `g` variable like so:

```suggestion
    } else if (goal instanceof ActivityTemplateGoal g) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -299,21 +235,27 @@ private DiscreteProfile convertToConstraintModelDiscreteProfile(
...
-      handleSimulationResults(results);
+      //simulation results from the last simulation, as converted for use by the constraint evaluation engine
+      lastSimConstraintResults = convertToConstraintModelResults(results);
+      lastSimDriverResults = results;
     }
   }
+
+  public Map<String, gov.nasa.jpl.aerie.constraints.model.ActivityInstance> getLatestEnvironment(){
+    return this.latestEnvironment;
+  }
```
mattdailis: 🔻 I don't think this environment means the same thing as what the constraints mean by environment.

My understanding is the environment is defined by `ForEachActivity` expressions, where the strings are the aliases. I think we can delete this method, and then in `PrioritySolver`, we can pass `Map.of()` (I'll leave a comment there as well)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -665,14 +662,18 @@ private Windows narrowByStateConstraints(
...
+    //make sure the simulation results cover the domain
+    simulationFacade.computeSimulationResultsUntil(totalDomain.end);
 
     //iteratively narrow the windows from each constraint
+    //REVIEW: could be some optimization in constraint ordering (smallest domain first to fail fast)
     for (final var constraint : constraints) {
-      ret = constraint.findWindows(plan, ret);
-      assert ret != null;
+      //REVIEW: loop through windows more efficient than enveloppe(windows) ?
+      final var validity = constraint.evaluate(simulationFacade.getLatestConstraintSimulationResults(), totalDomain, simulationFacade.getLatestEnvironment());
```
mattdailis: 🔻Since the `environment` is a mapping of activity aliases to activity instances (where the activity aliases come from `ForEachActivity`), the only valid environment to pass at the top level is `Map.of()`.

```suggestion
      final var validity = constraint.evaluate(simulationFacade.getLatestConstraintSimulationResults(), totalDomain, Map.of());
```

Perhaps this is worth adding a new default `evaluate` method to the `Expression` interface, which takes only the first two arguments, and supplies `Map.of()` itself.

mattdailis (changes requested):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: You are right. I misunderstood this interface and thought it was expected that we provide the environment. In a similar way that we may provide a `Plan` for some scheduler expressions.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationFacadeTest.java
```diff
@@ -188,55 +172,181 @@ public void associationToExistingSatisfyingActivity(){
...
     var expected = new Windows(Window.betweenClosedOpen(t1, t2));
     assertThat(actual).isEqualTo(expected);
   }
 
   @Test
   public void whenValueEqualDoubleOnSimplePlan() throws SimulationFacade.SimulationException {
     facade.simulateActivities(makeTestPlanP0B1().getActivities());
-    var actual = getFruitRes().whenValueEqual(SerializedValue.of(3.00), entireHorizon);
+    facade.computeSimulationResultsUntil(tEnd);
+    var actual = new Equal(getFruitRes(), new RealValue(3.0)).evaluate(facade.getLatestConstraintSimulationResults());
```
mattdailis: Very minor point - since `Equal` is generic over the profile type, we should always include `<>` when instantiating it:
```suggestion
    var actual = new Equal<>(getFruitRes(), new RealValue(3.0)).evaluate(facade.getLatestConstraintSimulationResults());
```

(EDIT: Joel made the same comment in his review)

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestFilters.java
```diff
@@ -1,122 +1,104 @@
...
+            "smallState1", smallState1(horizon),
+            "smallState2", smallState2(horizon)
+        )
+    );
+
+    final var ste = new Equal(new DiscreteResource("smallState1"), new DiscreteValue(SerializedValue.of(true)));
+    final var ste2 = new And(
+        new Equal(new DiscreteResource("smallState1"), new DiscreteValue(SerializedValue.of(true))),
+        new Equal(new DiscreteResource("smallState2"), new DiscreteValue(SerializedValue.of(true)))
+    );
```
mattdailis:
```suggestion
    final var ste = new Equal<>(new DiscreteResource("smallState1"), new DiscreteValue(SerializedValue.of(true)));
    final var ste2 = new And(
        new Equal<>(new DiscreteResource("smallState1"), new DiscreteValue(SerializedValue.of(true))),
        new Equal<>(new DiscreteResource("smallState2"), new DiscreteValue(SerializedValue.of(true)))
    );
```

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestOrderingConstraints.java
```diff
@@ -74,15 +74,15 @@ public void testCardinalityConstraint() {
...
+    ConstraintState cs = cc.isEnforced(plan, new Windows(Window.between(t0, h.toDur(TestUtility.timeFromEpochSeconds(10)))), null);
     assertTrue(cs.isViolation);
     CardinalityConstraint cc2 =
         new CardinalityConstraint.Builder().atMost(0).type(type1).inInterval(Window.between(
             t0,
                     t10)).build();
 
     //constraint is valid in interval (8,10)
-    ConstraintState cs2 = cc2.isEnforced(plan, new Windows(Window.between(t8, t10)));
+    ConstraintState cs2 = cc2.isEnforced(plan, new Windows(Window.between(t8, t10)), null);
```
mattdailis: It's a little unfortunate that some constraints require sim results, and others don't, but we want to give them a uniform interface. 🤷

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -30,91 +30,78 @@ public class TimeRangeExpression {
...
-  public Windows computeRange(Plan plan, Windows domain) {
+  public Windows computeRange(final SimulationResults simulationResults, final Plan plan, final Windows domain) {
 
     Windows inter = new Windows(domain);
-    //particularly important for combining with constant states as they are by definition adjacent values
-    //TODO:remove
-    //inter.doNotMergeAdjacent();
 
-    if (constantWin != null) {
+    if (constantWin != null && !inter.isEmpty()) {
```
mattdailis, line 39: I wonder if it would make sense to add `if (inter.isEmpty()) return inter;` after every call to `inter.intersectWith` - I _think_ that's equivalent to what you have, and makes it clear that as soon as `inter` is empty, we don't need to do any more work.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -30,91 +30,78 @@ public class TimeRangeExpression {
...
+      container.stream().reduce((a, b) -> {
+        timeline.add(Window.window(a.start, Window.Inclusivity.Exclusive, b.start, Window.Inclusivity.Exclusive));
+        return b;
+      });
+      inter.intersectWith(timeline);
     }
+
     for (var filterOrtransform : filtersAndTransformers) {
+      if(inter.isEmpty()) break;
       if (filterOrtransform instanceof TimeWindowsFilter) {
```
mattdailis: Here's another place where we could use "pattern matching instanceof" instead of casting:
```suggestion
      if (filterOrtransform instanceof TimeWindowsFilter f) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/resources/StateQueryParam.java
```diff
@@ -1,28 +1,61 @@
...
 
+import java.util.function.Predicate;
+
 /**
  * Class allowing to define state query expression for instantiation of parameters
  */
 public class StateQueryParam implements VariableArgumentComputer {
 
-  public final QueriableState state;
+  public final String nameResource;
```
JoelCourtney: [minor] could you name this `resourceName` instead?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/NAryMutexConstraint.java
```diff
@@ -22,29 +24,26 @@ public class NAryMutexConstraint extends GlobalConstraintWithIntrospection {
 
   Set<ActivityExpression> activityExpressions;
 
-  public static NAryMutexConstraint buildMutexConstraint(List<ActivityExpression> actTypes) {
-    NAryMutexConstraint mc = new NAryMutexConstraint();
-    mc.activityExpressions = new HashSet<>(actTypes);
-    return mc;
+  public NAryMutexConstraint(ActivityExpression... activityExpressions){
+    this.activityExpressions = new HashSet<>(Arrays.stream(activityExpressions).toList());
```
JoelCourtney: [minor] should be able to use `Arrays.asList(activityExpressions)` here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/transformers/TransformerEnveloppe.java
```diff
@@ -13,20 +14,20 @@ public class TransformerEnveloppe implements TimeWindowsTransformer {
   private final List<TimeRangeExpression> insideExprs;
 
 
-  public TransformerEnveloppe(List<TimeRangeExpression> insideExprs) {
+  public TransformerEnveloppe(final List<TimeRangeExpression> insideExprs) {
```
JoelCourtney: [minor] Could you fix the typo to `TransformerEnvelope` please?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/transformers/TransformWithReset.java
```diff
@@ -20,21 +21,21 @@ public TransformWithReset(TimeRangeExpression reset, TimeWindowsTransformer filt
...
-      var resetPeriods = resetExpr.computeRange(plan, Windows.forever());
+      var resetPeriods = resetExpr.computeRange(simulationResults, plan, Windows.forever());
 
       for (var window : resetPeriods) {
         // get windows to filter that are completely contained in reset period
         Windows cur = windowsToTransform.subsetContained(window);
         if (!cur.isEmpty()) {
           //apply filter and union result
-          ret.addAll(transform.transformWindows(plan, cur));
+          ret.addAll(transform.transformWindows(plan, cur, null));
```
JoelCourtney: It looks like the `TransformerEnvelope.transformWindows(...)` method uses the `simulationResults` argument, so I think it should be passed into `transformWindows` instead of using `null`.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestEnveloppes.java
```diff
@@ -43,10 +43,10 @@ public void testEnveloppes() {
         .thenTransform(enveloppe)
         .name("encounter_envelopper_TRE")
```
JoelCourtney: [minor] Same typo here, and in file name

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -77,13 +77,13 @@ protected ActivityExpression() { }
     Map<String, SerializedValue> arguments = new HashMap<>();
     Map<String, VariableArgumentComputer> variableArguments = new HashMap<>();
 
-    public B withArgument(String argument, QueriableState state, TimeExpression timeToQuery) {
-      variableArguments.put(argument, new StateQueryParam(state, timeToQuery));
+    public B withArgument(String argument, String nameState, TimeExpression timeToQuery) {
```
JoelCourtney: [minor] same as the `nameResource` comment, calling it `nameState` sounds like a variable representing "the state of the name" rather than "name of the state"

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -205,14 +192,13 @@ public Builder name(String name) {
     /**
      * relative instant use case
      *
-     * @param state x
+     * @param nameDiscreteProfile name of the discrete profile
```
JoelCourtney, line 194: [minor] same comment about `nameThing` vs `thingName`

JoelCourtney (changes requested):
Thanks! I only have two actual requests, the rest is just nitpicking.

1. I'm getting a bunch of unchecked call warnings when compiling the scheduler tests. You can make them go away by using `new Equal<>(...)` instead of `new Equal(...)`. I don't really know why that helps, because it doesn't add any information for type checking, but it does. Looks like the warnings are only emitted on a clean build.
2. See `TransformerEnvelope.transformWindows(...)` comment below

for comments below, [minor] = not a merge blocker, not a big deal if you don't address it

mattdailis (commented):
Last batch of comments from me - aside from the `transformWindows` issue that Joel called out, I don't see any additional blockers

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestOrderingConstraints.java
adrienmaillard: Thought about the same thing yesterday about `Expression` and its context map.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: I have added the default method.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/transformers/TransformWithReset.java
adrienmaillard: Good catch !

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
adrienmaillard: Did that.

adrienmaillard:
Thank you for the comments, I addressed them all in the latest push.

JoelCourtney (approved):
Awesome!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/resources/StateQueryParam.java
```diff
@@ -1,28 +1,65 @@
...
+      //TODO: improve Window to allow querying containment in interval directly
+      //for now we look for the last matching profile segment, if any (last to get latest if any overlaps)
+      final var profile = simulationResults.discreteProfiles.get(this.resourceName);
+      final Predicate<DiscreteProfilePiece> containsQueryTimeP = (piece)
+          -> !Window.intersect(piece.window, Window.at(queryT)).isEmpty();
+      final var matchPiece = Lists.reverse(profile.profilePieces).stream()
+                                  .filter(containsQueryTimeP).findFirst();
+      return matchPiece.map(piece -> piece.value).orElse(null);
+    } else {
+      return null;
```
mattdailis: I think in these two cases as well we'll want to throw some sort of error or exception. Case 1: There was no segment at the desired time, and Case 2: the requested resource does not exist.

mattdailis (approved):
Looks good to me - I've left one comment on error cases in StateQueryParam, but that isn't strictly speaking in scope for this PR, so I'll leave it up to you whether you want to address it or not.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestOrderingConstraints.java
mattdailis: Let's update this test to actually use simulation results, rather than use null. I did it on the `feature/AERIE-1813--replace-state-constraints--update-test-ordering-constraints`, feel free to merge it into your branch or write it yourself if you prefer:

https://github.com/NASA-AMMOS/aerie/commit/7684e08cce01be2959d5aba063e8289f8550f7a4

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/BinaryMutexConstraint.java
```diff
@@ -63,7 +64,7 @@ private Windows findWindows(Plan plan, Windows windows, ActivityType actToBeSche
 
   //Non-incremental checking
   //TODO: does not help finding where to put acts
-  public ConstraintState isEnforced(Plan plan, Windows windows) {
+  public ConstraintState isEnforced(Plan plan, Windows windows, SimulationResults simulationResults) {
```
mattdailis, line 68:
```suggestion
  @Override
  public ConstraintState isEnforced(Plan plan, Windows windows, SimulationResults simulationResults) {
```

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/CardinalityConstraint.java
```diff
@@ -111,7 +112,7 @@ private CardinalityConstraint() {
 
 
   //Non-incremental checking
-  public ConstraintState isEnforced(Plan plan, Windows windows) {
+  public ConstraintState isEnforced(Plan plan, Windows windows, SimulationResults simulationResults) {
```
mattdailis, line 116:
```suggestion
  @Override
  public ConstraintState isEnforced(Plan plan, Windows windows, SimulationResults simulationResults) {
```

mattdailis (commented):
Let's use `@Override` annotations everywhere that we're overriding anything, to make it obvious (and to have the compiler tell us if we make a mistake)

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestOrderingConstraints.java
adrienmaillard: Great thanks.

mattdailis (approved):

PR #168, opened 2022-05-06
[HOTFIX] Close Simulation Engine in Scheduler

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

The scheduler's IncrementalSimulationDriver creates a new SimlationEngine whenever it restarts the simulation from the beginning. This is fine, but it neglects to call the `close` method on the old SimulationEngine object. This results in a resource leak, as the number of threads in use continues to grow without bound.

The fix is simple - adding one line to close the simulation engine before starting a new one.

## Verification

I added a test to the `IncrementalSimulationTest` called `testThreadsReleased`. It re-initializes a simulation twenty thousand times, each time simulating one activity. Running this test before adding the fix resulted in a `Failed to start thread` error after about eight thousand iterations (which spawned sixteen thousand threads).

Screen Shot 2022-05-05 at 5 36 50 PM

After making the fix, the test ran to completion. Instead of taking 50 seconds, it took less than 4.

## Documentation

No documentation is relevant to this change.

## Future work

None at the moment.

### DIFF
--- a/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
+++ b/scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
@@ -1,11 +1,11 @@
-package gov.nasa.jpl.aerie.scheduler;
+package gov.nasa.jpl.aerie.scheduler.simulation;
 
 import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
 import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
 import gov.nasa.jpl.aerie.merlin.protocol.model.TaskSpecType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.MissingArgumentsException;
-import gov.nasa.jpl.aerie.scheduler.simulation.IncrementalSimulationDriver;
+import gov.nasa.jpl.aerie.scheduler.SimulationUtility;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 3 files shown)


adrienmaillard (approved):

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
```diff
@@ -53,6 +54,23 @@ public void durationTest(){
...
+    final var activity = new TestSimulatedActivity(
+        Duration.of(0, SECONDS),
+        new SerializedActivity("BasicActivity", Map.of()),
+        new ActivityInstanceId(1));
+    final var fooMissionModel = SimulationUtility.getFooMissionModel();
+    final var executor = (ThreadPoolExecutor) fooMissionModel.getModel().executor();
+    incrementalSimulationDriver = new IncrementalSimulationDriver(fooMissionModel);
+    incrementalSimulationDriver.initSimulation();
+    for (var i = 0; i < 20000; i++) {
+      incrementalSimulationDriver.initSimulation();
```
Twisol: Can probably remove the first call to `initSimulation`?

Twisol (approved):

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
mattdailis: Sure thing :+1: I had added it when I was trying to figure out exactly how many threads should be used. In the end I decided it didn't matter whether the test uses 2 or 3 threads, just that the number was kept in control.

PR #169, opened 2022-05-08
Generate `return` for activities with return values

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
It looks like output attributes are not supported at the code generation level -- when I change `BakeBananaBreatActivity` to return an `int`, the code generator produces code that fails to compile. (We need to return a `Task<Integer>`, but the generated code does not actually return the value produced by the activity effect model.)

This PR adds an extra case to code generation, checking whether the effect model returns non-`void`, and if so, generating the necessary `return` statement.

The first commit is just a whitespace change.

## Verification
I changed `BakeBananaBreadActivity` to return an `int`, and observed a compile-time failure before the change and a successful compilation afterwards.

## Documentation
Existing documentation should be unchanged. (If anything, the documentation would have been inaccurate *before* this PR.)

### DIFF
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
@@ -15,8 +15,9 @@ public record BakeBananaBreadActivity(double temperature, int tbSugar, boolean g
   }
 
   @EffectModel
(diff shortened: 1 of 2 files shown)


mattdailis (approved):
Good catch :+1: Thanks for making the fix.

The lesson I'm learning from this is that it's important to have examples for all of our code generating facilities - otherwise we have no feedback when we accidentally break features.

patkenneally (approved):

PR #170, opened 2022-05-09
Factor directive handling out of SimulationEngine

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description

This PR continues the engine refactoring work begun in #86, now removing the dependence of the engine on the `MissionModel` and `Directive` types. We introduce a series of internal topics on the event graph, recording the association of a task with an input directive (if any), a task's input attributes, and a task's output attributes. This information is extracted during results processing, rather than routed through the simulation engine itself.

* The two roles of the `Query` type -- to query cells and to emit events -- have been factored into distinct `Query` and `Topic` types.
  * This allows us to define topics without an associated accumulator (cell).
* The `Event` type now tracks **provenance**, i.e. which task emitted it.
  * This allows us to associate data with a task *without* knowing its ID explicitly, which is exactly what we need to track directive associations and input/output attributes.
* The `Task#step` method now returns (as part of its `TaskStatus`) a `Task` object representing the next step of the `Task`, rather than expecting that the same `Task` be stepped multiple times.
  * This allows us to easily inject an `emit` action into each top-level task recording which directive ID the task is associated with, without needing to manually idempotize that action for subsequent calls to `step`.
* Each generated activity types now has a pair of typed topics for its start and input attributes; every activity instance now bookends the effect model by emitting events on these topics.

As usual for my PRs, I've collected a series of minor refactorings at the start of the branch, setting up for the more involved work at the end of the branch. There are two changes of note in the front half:

* The flatten and unflatten operations on event graphs have been rewritten, and the associated test suite simplified to focus on these operations as an isolated unit. Graph flattening is now written in terms of an EffectTrait implementation, allowing the complex custom traversal logic to be removed. Graph unflattening is now written in terms of an explicit intermediate [trie](https://en.wikipedia.org/wiki/Trie). Both of these changes, ideally, make the algorithms easier to understand.
* `protocol.types.ValueSchema.Unit` and `framework.VoidEnum` have been unified into a single `protocol.types.Unit`. The inline documentation previously attached to `VoidEnum` has been carried forward.

## Verification
Visual confirmation on the output of the `SimulateMapSchedule` integrated test.

## Documentation
* The `framework.VoidEnum` type has been removed. Users should use the `protocol.types.Unit` type instead, which is equivalent except for naming.
* Custom implementations of `ActivityMapper` must now provide two new methods, `getInputTopic` and `getOutputTopic`.

## Future work
* Remove `Scheduler#spawn(DirectiveTypeId<Input, Output>, Input, Task<Output>)`; the simulation engine no longer needs the first two parameters, since we the same information is emitted into the event log by the spawned task itself.
* Allow cells to subscribe to more than one topic. This will allow certain "fan-out" patterns (used by modelers to effect correlated changes on multiple cells with one action) to be internalized into Merlin itself.
* Control creation of `Topic`s by allocating them during model initialization (i.e. via the `Initializer`), to support efficient masking of events by topic without relying on object identity or hash lookups.
* Add more internal topics to further decouple results computation from simulation, with the end goal that simulation and results computation are pipelined by the production and consumption of top-level commits on the event log.
  * Task parent/child relationships are a good candidate for this.
  * Resources behave like topics in some ways but not others; tracking resources between simulation and results computation is still an open problem.
* Persist event provenance data to the database, and display events in the UI.
  * Ideally, UI support for displaying the input and output attributes of an activity would be generalized to display any event emitted by an activity, together with an indicator within the activity's span for when the event occurred.
* The results computation code currently special-cases on the assumed names of the activity type topics to recognize these events specially, but this specialization should occur further downstream.
  * Ideally, the UI would make all events from a task accessible in the same way, including the input and output attribute events, so there wouldn't be any specialization anywhere.


Twisol:
I can't really address this test failure in this PR because it's occurring in code in another repository. (In fact, the Lander JAR was built against an existing version of Merlin, but the tests are running it against the branch's version.)

May I recommend that these particular integration-type tests be moved to a nightly run or non-required action (or both!), so they don't gate PRs?

---

image

Twisol:
I pulled the Lander repository, built a new JAR, and committed it as part of this PR. (This approach won't work if/when we pull Lander from GitHub Packages instead of vendoring a built JAR.) 

Let's see if it works now.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/timeline/Event.java
```diff
@@ -30,12 +33,16 @@ public Topic<?> topic() {
...
   @Override
   public String toString() {
     return "<@%s, %s>".formatted(System.identityHashCode(this.inner.topic), this.inner.event);
   }
 
-  private record GenericEvent<EventType>(Topic<EventType> topic, EventType event) {
+  private record GenericEvent<EventType>(Topic<EventType> topic, EventType event, TaskId provenance) {
     private GenericEvent {
       Objects.requireNonNull(topic);
       Objects.requireNonNull(event);
```
mattdailis, line 49: 🔵 Do we need to add a third null check here? Or perhaps get rid of all the null checks? Or is there a difference between topic, event, and provenance in terms of whether we care about its nullity?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -820,7 +819,7 @@ public Optional<JavaFile> generateActivityMapper(final MissionModelRecord missio
                                         "activity",
                                         effectModel.methodName(),
                                         "model",
-                                        VoidEnum.class,
+                                        Unit.class,
```
mattdailis, line 928: 🔻 Commit "Unify VoidEnum with Unit" doesn't build because it emits `Unit.VOID` instead of `Unit.UNIT`.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/ActivityTypeRecord.java
```diff
@@ -13,4 +15,12 @@ public record ActivityTypeRecord(
     MapperRecord mapper,
     ExportDefaultsStyle defaultsStyle,
     Optional<EffectModelRecord> effectModel
-  ) implements ExportTypeRecord { }
+  ) implements ExportTypeRecord {
+  public TypeName getOutputTypeName() {
```
mattdailis, line 19: 🔵 I like this `getOutputTypeName` method - I think there are a few other places that would benefit from it. (Perhaps it's worth splitting into a refactor commit?)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -390,13 +354,51 @@ public boolean isTaskComplete(final TaskId task) {
...
       final MissionModel<?> missionModel) {
+    final Map<String, ActivityInstanceId> taskToPlannedDirective = new HashMap<>();
+
+    for (final var point : timeline) {
+      if (!(point instanceof TemporalEventSource.TimePoint.Commit p)) continue;
+      if (!p.topics().contains(activityTopic)) continue;
+
+      p.events()
+       // Filter events that match the given topic.
+       .substitute($ -> ($.topic() == activityTopic) ? EventGraph.atom($) : EventGraph.empty())
```
mattdailis: 🔵 This `substitute` call is redundant with the `extract` call below - but I think it adds clarity, so I'm fine to leave it as is :+1: 

EDIT: Looks like this was refactored in a later commit.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -390,13 +354,51 @@ public boolean isTaskComplete(final TaskId task) {
...
+
+             @Override
+             public Consumer<Map<String, ActivityInstanceId>> concurrently(
+                 final Consumer<Map<String, ActivityInstanceId>> left,
+                 final Consumer<Map<String, ActivityInstanceId>> right)
+             {
+               return $ -> { left.accept($); right.accept($); };
+             }
+           },
+           ev -> $ -> ev.extract(activityTopic).ifPresent(id -> $.put(ev.provenance().id(), id)))
```
mattdailis: 🔵 Some thoughts here - I renamed `id` with `directiveId` when reading this code because I got worried for a minute that our Map was backwards.

An invariant here is that no task will ever emit two different directive IDs to activityTopic. That seems reasonable, but I figured I'd call it out anyhow.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -463,26 +515,28 @@ public SimulationResults computeResults(
...
+        final var inputAttributes = taskInfo.input().get(task.id());
+        final var outputAttributes = taskInfo.output().get(task.id());
+
         simulatedActivities.put(activityId, new SimulatedActivity(
-            directive.getType(),
-            directive.getArguments(),
+            inputAttributes.getTypeName(),
+            inputAttributes.getArguments(),
             startTime.plus(e.startOffset().in(Duration.MICROSECONDS), ChronoUnit.MICROS),
             e.joinOffset().minus(e.startOffset()),
```
mattdailis, line 516: 🔶 We could conceivably extract the start and end time of an activity from its input/output events, and remove that information from ExecutionState... but I'm not sure how well that aligns with the goal of pushing the specialization downstream.

EDIT: Perhaps, rather than specialize to input/output events, we could define the start and end as the first and last event emitted by this activity 🤔 That does depend on an activity always emitting events in its first and last instant though...

mattdailis (changes requested):
This was a good read. Only real change requested is that there's a stretch of commits in the middle that generate code that doesn't build - other than that, I have no concerns. Thanks for taking the time to make these improvements!

> Key:

mattdailis:
Just want to call out that this PR and #164 both touch `computeResults` - I doubt the rebase would be too bad in either direction, but just want to call that out. (CC @pcrosemurgy )

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/ActivityTypeRecord.java
Twisol: Yes, I agree. Unfortunately, the whole annotation processor is getting quite hairy, and I'd like to take more drastic action against it. I figured for this PR it would be better to avoid making too much of my own mess, and then come back later to improve the whole thing. Does that seem fair to you?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
Twisol: Oops, nice catch.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/ActivityTypeRecord.java
mattdailis: Sounds good :+1:

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/timeline/Event.java
Twisol: Thanks; I've added a null check for `provenance`.

Honestly, I go back and forth on how useful these are. I'd like to say "just don't use null in the first place!", but experience suggests that `null` will happen whether we intend it or not, so having these checks helps during development...

pcrosemurgy:
> Just want to call out that this PR and #164 both touch `computeResults` - I doubt the rebase would be too bad in either direction, but just want to call that out. (CC @pcrosemurgy )

All good :) I'm going to try to get #164 in today but don't mind dealing with merge conflicts, so no worries if this gets in first.

mattdailis (approved):
Confirmed that all commits build now :+1:

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: Hmm, you're right -- I've reshaped the commits around here to better isolate the changes.

```
* 437cca12a Extract task info from event graph
* c48da3625 Associate tasks to activities by emitted metadata
```

I also realized that my wrapper tasks for emitting directive IDs wouldn't correctly `release` the underlying task, so I [changed that](https://github.com/NASA-AMMOS/aerie/compare/fd372429db79a0bbabd84419ad5419b90ec445ed..65eb4cd593d7d89a4e9722f016ac67816f4345e2).

Twisol:
Based on some out-of-band discussions with @mattdailis, I've removed `AutoCloseable` from the `Task` interface (it's dangerously misleading now that a `Task` cedes ownership of its transient system resources to the next continuation) and improved some of the documentation around `Task#release`.

[Diff here.](https://github.com/NASA-AMMOS/aerie/compare/65eb4cd593d7d89a4e9722f016ac67816f4345e2..6fd8e93c39fdbcb7b45ea2651549dbd8dedf0bed)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: Yes, and honestly (see Future Work), I'd like to see this handling of directive associations and input/output attributes pushed out past the Merlin API boundary altogether. A lot of the results code is basically an upstreamed analysis of the simulation, which ought to be done elsewhere.

Twisol: > 🔵 Some thoughts here - I renamed `id` with `directiveId` when reading this code because I got worried for a minute that our Map was backwards.

Good call; I've renamed the variable.
```diff
-          ev.extract(this.activityTopic).ifPresent(id -> taskInfo.taskToPlannedDirective.put(ev.provenance().id(), id));
+          ev.extract(this.activityTopic)
+            .ifPresent(directiveId -> taskInfo.taskToPlannedDirective.put(ev.provenance().id(), directiveId));
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol: Yes, I thought about that too. I'm worried that it's more of a *re-interpretation* of what it means for a task to have a duration, rather than a mere reflowing of information.

We have other things that don't fit into this paradigm (e.g. resource profiles), so I'm happy to punt on this here and deal with all of the misfits more holistically.

Twisol: (As a specific case, if the simulation is still in progress, as it will be during pipelined results handling, how do we know *which event* is the last event? Maybe the task just hasn't emitted anything for a while!)

Twisol:
I think I've addressed all feedback so far. (Thanks @mattdailis!)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -383,20 +332,107 @@ public boolean isTaskComplete(final TaskId task) {
...
       final TemporalEventSource timeline,
-      final MissionModel<?> missionModel) {
+      final MissionModel<?> missionModel
+  ) {
+    // Collect per-task information from the event graph.
+    final var taskInfo = new TaskInfo();
+
+    for (final var point : timeline) {
+      if (!(point instanceof TemporalEventSource.TimePoint.Commit p)) continue;
+      if (!p.topics().contains(activityTopic)) continue;
```
mattdailis: Another subtle point that I'll call out - this statement assumes that input and output attributes are only emitted during Commits that also include events on `activityTopic`. I'm having some trouble proving to myself whether or not this assumption is true.

`emitAndThen` emits to `activityTopic`, and then yields - I _think_ this means that any subsequent steps won't be in the same Commit. Certainly I wouldn't expect the output attributes to be in that Commit.

Twisol: > this statement assumes that input and output attributes are only emitted during Commits that also include events on `activityTopic`

Ah, I goofed -- originally this loop only handled `activityTopic` events, and I wanted to avoid processing commits that were definitely not of interest. This guard should simply be removed now.

For activities spawned by directive (i.e. they were in the input plan), technically the start attributes event would be in the same commit as the directive association event. This would not be trued for child activities (which aren't associated with any directive), and would anyway be almost never true for end attributes.

> `emitAndThen` emits to `activityTopic`, and then yields

No, it doesn't yield -- it emits and then immediately runs the remainder of the task (i.e. the task being wrapped).

Twisol: > This guard should simply be removed now.

I've [removed this guard](https://github.com/NASA-AMMOS/aerie/compare/4313cf2a9a5a4d496e752bbf7074dfe84e99c8b4..70d8eb3d553c3ec1e0d5e439d780f6b1c3745401) from the commit and re-pushed.

mattdailis: Looks great, thanks!

patkenneally (approved):

mattdailis:
Rebased onto develop. The least trivial conflict was in `computeResults`, but it was expected, per comments above. (and it was still pretty trivial)

Twisol:
Thanks for the rebase!

I realized before I fell asleep last night (what weird timing) that my change adding two topics to generated mappers will be a backwards-incompatible change for any users who defined their own activity mappers. I don't think anybody is doing that, since there was a similar break that went unnoticed for much longer on `develop`, but it's worth adding to the release notes.

Twisol:
I'm tweaking the commit where I introduce these topics to add `getInputTopic` and `getOutputTopic` methods to the `ActivityMapper` interface. Then we can just document that this interface has changed, and the compiler will help users identify where changes need to be made to their custom mappers.

Hold...

Twisol:
> I'm tweaking the commit where I introduce these topics to add `getInputTopic` and `getOutputTopic` methods to the `ActivityMapper` interface. Then we can just document that this interface has changed, and the compiler will help users identify where changes need to be made to their custom mappers.
> 
> Hold...

Done! I've updated the PR summary to note this user-facing change.

PR #171, opened 2022-05-11
[AERIE-1812] Add cardinality goal (with implied mutex) to the scheduling eDSL

adrienmaillard:
* **Tickets addressed:** AERIE-1812
* **Review:** By commit  

## Description

A first version of the CardinalityGoal available in the backend is implemented in the scheduling eDSL. 

Here below are the two flavors. The first is the "duration" one where the user can specify a lower bound for the sum of the durations of activities present in the plan: 
```javascript
export default function myGoal() {
    return Goal.CardinalityGoal({
        activityTemplate: ActivityTemplates.GrowBanana({
            quantity: 1,
            growingDuration: 1000000,
        }),
        inPeriod: { start: 0, end: 10000000 },
        specification: { duration: 10 * 1000000 }
    })
}
```
The second one is about the minimum number of occurrences that should be present in the plan:
```javascript
export default function myGoal() {
    return Goal.CardinalityGoal({
        activityTemplate: ActivityTemplates.GrowBanana({
            quantity: 1,
            growingDuration: 1000000,
        }),
        inPeriod: { start: 0, end: 10000000 },
        specification: { occurrence: 10 }
    })
}
```

Finally, both specifications can be put together and solved by the same goal: 
```javascript
export default function myGoal() {
    return Goal.CardinalityGoal({
        activityTemplate: ActivityTemplates.GrowBanana({
            quantity: 1,
            growingDuration: 1000000,
        }),
        inPeriod: { start: 0, end: 10000000 },
        specification: { occurrence: 10, duration: 10 * 1000000 }
    })
}
```

The cardinality goal requires to insert a certain number of occurrences/duration of an activity type. Without an explicit noOverlap/mutex constraint, the scheduler can/would insert activities of the same type concurrently. And we assume this is not the behavior the user expects. For now we assume the user implies a self-mutex on all activity types of the mission model. 

The creation of `WindowsOf` expression to translate `Violation` objs returned by `ForEachActivity` expressions into `Windows` is a byproduct of the early prototype. As it is useful, we keep it. 

During testing, a few issues have been encountered and fixed:
- `Evaluation::removeAssociation` was producing a ConcurrentModification exception.
- When converting simulation `SimulationResults`  to constraints `SimulationResults` in `SimulationFacade`, instantaneous activities disappeared from because a closed-open interval (used to represent durative activities) cannot represent a single instant in a `Windows`. 
- In the case of an activity type with controllable duration, the duration parameter was incorrectly handled in the scheduler server in both "ways" (1) when a plan/scheduling specification is fetched (2) when the scheduling results are saved.
-  When a `CardinalityGoal` of the "duration" flavor is specified and that its activity type produce instantaneous activities, it is possible that the cardinality goal is unsatisfiable. But in the case of uncontrollable activities, this behavior can be detected only after simulation. A diagnostic mechanism has been added to the `CardinalityGoal`. If the scheduler keeps adding zero-duration activities to satisfy a duration goal without progress for a certain number of iterations, an `UnsatisfiableGoalConflict` is returned. 

## Verification

Tests of the two types of `CardinalityGoal` have been added to `SchedulingIntegrationTests`.
I have also tested the goal within the UI. 
## Documentation

None. 

## Future work

- Implement global constraints and let the user add them explicitly. 
- Implement complete support of CardinalityGoal functionalities.
- Reflect seriously about what instantaneous activities mean for the scheduler

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -3,6 +3,7 @@ import * as WindowsEDSL from './windows-edsl-fluent-api'
 
 interface ActivityRecurrenceGoal extends Goal {}
 interface ActivityCoexistenceGoal extends Goal {}
+interface ActivityCardinalityGoal extends Goal {}
 
 export class Goal {
   public readonly __astNode: AST.GoalSpecifier;
@@ -49,6 +50,14 @@ export class Goal {
       forEach: opts.forEach.__astnode,
     });
   }
+  public static CardinalityGoal(opts: { activityTemplate: ActivityTemplate, specification: AST.CardinalityGoalArguments, inPeriod: ClosedOpenInterval }): ActivityCardinalityGoal {
+    return Goal.new({
+      kind: AST.NodeKind.ActivityCardinalityGoal,
+      activityTemplate: opts.activityTemplate,
+      specification: opts.specification,
+      inPeriod : opts.inPeriod
+    });
+  }
 }
 
 declare global {
@@ -61,12 +70,15 @@ declare global {
     public static ActivityRecurrenceGoal(opts: { activityTemplate: ActivityTemplate, interval: Duration }): ActivityRecurrenceGoal
 
     public static CoexistenceGoal(opts: { activityTemplate: ActivityTemplate, forEach: WindowsEDSL.WindowSet }): ActivityCoexistenceGoal
+
+    public static CardinalityGoal(opts: { activityTemplate: ActivityTemplate, specification: AST.CardinalityGoalArguments, inPeriod: ClosedOpenInterval }): ActivityCardinalityGoal
   }
   type Duration = number;
   type Double = number;
   type Integer = number;
 }
 
+export interface ClosedOpenInterval extends AST.ClosedOpenInterval {}
 export interface ActivityTemplate extends AST.ActivityTemplate {}
 
 // Make Goal available on the global object
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
@@ -12,6 +12,8 @@ import gov.nasa.jpl.aerie.constraints.tree.Or;
 import gov.nasa.jpl.aerie.constraints.tree.RealResource;
 import gov.nasa.jpl.aerie.constraints.tree.RealValue;
 import gov.nasa.jpl.aerie.constraints.tree.Transition;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.DurationValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.constraints.TimeRangeExpression;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityCreationTemplate;
@@ -135,8 +137,16 @@ public class GoalBuilder {
   private static ActivityCreationTemplate makeActivityTemplate(
       final SchedulingDSL.ActivityTemplate activityTemplate,
       final Function<String, ActivityType> lookupActivityType) {
-    var builder = new ActivityCreationTemplate.Builder()
-        .ofType(lookupActivityType.apply(activityTemplate.activityType()));
+    var builder = new ActivityCreationTemplate.Builder();
+    final var type = lookupActivityType.apply(activityTemplate.activityType());
+    if(type.getDurationType() instanceof DurationType.Controllable durationType){
+      //detect duration parameter
+      if(activityTemplate.arguments().containsKey(durationType.parameterName())){
+        builder.duration(new DurationValueMapper().deserializeValue(activityTemplate.arguments().get(durationType.parameterName())).getSuccessOrThrow());
+        activityTemplate.arguments().remove(durationType.parameterName());
+      }
+    }
+    builder = builder.ofType(type);
     for (final var argument : activityTemplate.arguments().entrySet()) {
       builder = builder.withArgument(argument.getKey(), argument.getValue());
     }
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
@@ -12,13 +12,23 @@ import gov.nasa.jpl.aerie.constraints.tree.Or;
 import gov.nasa.jpl.aerie.constraints.tree.RealResource;
 import gov.nasa.jpl.aerie.constraints.tree.RealValue;
 import gov.nasa.jpl.aerie.constraints.tree.Transition;
+import gov.nasa.jpl.aerie.constraints.time.Window;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+import gov.nasa.jpl.aerie.constraints.tree.During;
+import gov.nasa.jpl.aerie.constraints.tree.ForEachActivity;
+import gov.nasa.jpl.aerie.constraints.tree.ViolationsOf;
+import gov.nasa.jpl.aerie.constraints.tree.WindowsOf;
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.DurationValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
+import gov.nasa.jpl.aerie.scheduler.Range;
+import gov.nasa.jpl.aerie.scheduler.constraints.TimeRangeExpression;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.constraints.TimeRangeExpression;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityCreationTemplate;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityExpression;
 import gov.nasa.jpl.aerie.scheduler.constraints.timeexpressions.TimeAnchor;
+import gov.nasa.jpl.aerie.scheduler.goals.CardinalityGoal;
 import gov.nasa.jpl.aerie.scheduler.goals.CoexistenceGoal;
 import gov.nasa.jpl.aerie.scheduler.goals.CompositeAndGoal;
 import gov.nasa.jpl.aerie.scheduler.goals.Goal;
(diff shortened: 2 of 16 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/BinaryMutexConstraint.java
```diff
@@ -41,11 +41,11 @@ public Windows findWindows(Plan plan, Windows windows, Conflict conflict) {
 
 
   private Windows findWindows(Plan plan, Windows windows, ActivityType actToBeScheduled) {
-
+    Windows validWindows = new Windows(windows);
     if (!(actToBeScheduled.equals(actType) || actToBeScheduled.equals(otherActType))) {
-      throw new IllegalArgumentException("Activity type must be one of the mutexed types");
+      //not concerned by this constraint
+      return validWindows;
```
goetzrrGit, line 48: A little confused by this. Why are you initializing a new Window object from an existing Window object?
 
 Why not `return windows`

goetzrrGit (commented):
Ran through the code and didn't spot any issues. I would prefer someone who has worked on the  Schedular stamp an `approval`

dyst5422:
`inPeriod` and `duration` don't specify units. I would encourage use of a proper Temporal.Duration for durations or allow numeric inputs with a unit specified in the field name. ie `inPeriod: { startMs: 0, endMs: 100000 }`

We're really relying heavily on the type hinting to document the intricacies or the API for the users, so having specificity here is crucial.

dyst5422 (approved):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/Evaluation.java
```diff
@@ -91,11 +91,11 @@ public Optional<Integer> getNbConflictsDetected() {
      * @param createdByThisGoal IN a boolean stating whether the instance has been created by this goal or not
      */
     public void associate(java.util.Collection<ActivityInstance> acts, boolean createdByThisGoal) {
-      acts.forEach((a)->this.acts.put(a, createdByThisGoal));
+      acts.forEach(a ->this.acts.put(a, createdByThisGoal));
     }
 
     public void removeAssociation(java.util.Collection<ActivityInstance> acts){
-      acts.forEach((a)->this.acts.remove(a));
+      this.acts.entrySet().removeIf(act -> acts.contains(act.getKey()));
```
mattdailis, line 98: I learned something new - I thought the `EntrySet` was an immutable copy of the Map - in fact, it is quite the opposite, it is a live view into the Map that allows you to mutate the Map as though it were a Set.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -193,9 +193,13 @@ private gov.nasa.jpl.aerie.constraints.model.ActivityInstance convertToConstrain
   {
     final var startT = Duration.of(startTime.until(driverActivity.start(), ChronoUnit.MICROS), MICROSECONDS);
     final var endT = startT.plus(driverActivity.duration());
+    var activityWindow = Window.betweenClosedOpen(startT, endT);
+    if(startT.isEqualTo(endT)){
+      activityWindow = Window.between(startT, endT);
+    }
```
mattdailis: If we need to choose one between `final` and `var`, we should choose `final` 😝 

```suggestion
    final Window activityWindow;
    if (startT.isEqualTo(endT)) {
      activityWindow = Window.between(startT, endT);
    } else {
      activityWindow = Window.betweenClosedOpen(startT, endT);
    }
```

EDIT: Of course we could have both `final` and `var` if we make it a one-liner:

```suggestion
    final var activityWindow = startT.isEqualTo(endT)
      ? Window.between(startT, endT)
      : Window.betweenClosedOpen(startT, endT);
```

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -1,8 +1,11 @@
 import * as AST from './scheduler-ast.js';
+import { ActivityType } from "./mission-model-generated-code";
+import {CardinalityGoalArguments} from "./scheduler-ast.js";
```
mattdailis: I think `ActivityType` is unused, so we can delete that import.

The `CardinalityGoalArguments` import can also be deleted - instead, we should access it via `AST.CardinalityGoalArguments`.

```suggestion
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
```diff
@@ -135,8 +161,16 @@ private static Expression<Windows> expressionOfConstraintExpression(
...
       final SchedulingDSL.ActivityTemplate activityTemplate,
       final Function<String, ActivityType> lookupActivityType) {
-    var builder = new ActivityCreationTemplate.Builder()
-        .ofType(lookupActivityType.apply(activityTemplate.activityType()));
+    var builder = new ActivityCreationTemplate.Builder();
+    final var type = lookupActivityType.apply(activityTemplate.activityType());
+    if(type.getDurationType() instanceof DurationType.Controllable durationType){
+      //detect duration parameter
+      if(activityTemplate.arguments().containsKey(durationType.parameterName())){
+        builder.duration(new DurationValueMapper().deserializeValue(activityTemplate.arguments().get(durationType.parameterName())).getSuccessOrThrow());
```
mattdailis, line 169: This is the best we can do for now - however, I think we'll run into issues with this soon, because users can (and do) use different value mappers than what we provide. (In particular, Clipper overrides the `DurationValueMapper` to return more human-readable units, instead of microseconds as a long)

I've captured this in a ticket:
> AERIE-1863: Do not assume Durations are serialized as integers
> 
> In order to interpret the expected duration of a ControllableDuration activity, the scheduler must be able to deserialize a serialized Duration argument into a Duration java object.
> 
> Mission models are allowed to override the default value mappers we provide them, so it is not safe for us to assume anything about the serialized representation of Durations.
> 
> One solution is to add methods to the SchedulerModel that will serialize and deserialize Durations for us. The correct value mapper will be available at annotation processing time, so the generated SchedulerModel should have access to it.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -120,19 +137,31 @@ protected CardinalityGoal fill(CardinalityGoal goal) {
...
+        throw new IllegalArgumentException("Goal is incorrectly parametrized: minimum duration x minimum occurence > maximum duration");
       }
       return goal;
     }
 
+    public boolean isUnsatisfiableDurationType(){
+      return (thereExists.getDurationRange() != null &&
+              thereExists.getDurationRange().isSingleton() &&
+              thereExists.getDurationRange().start.isZero() &&
+              (thereExists.getType().getDurationType() instanceof DurationType.Controllable) &&
```
mattdailis: I'm curious why the duration type is needed for `isUnsatisfiableDuraitonType`? I think I just don't quite understand this method.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -210,6 +244,30 @@ public Collection<Conflict> getConflicts(Plan plan, final SimulationResults simu
...
+    if(this.durationRange != null){
+      final var inserted = plan.getEvaluation().forGoal(this).getInsertedActivities();
+      for(var insertedActivity : inserted){
+        //new activity inserted to satisfy this goal
+        if(!insertedSoFar.containsKey(insertedActivity.getId())){
+          //record it
+          insertedSoFar.put(insertedActivity.getId(), insertedActivity);
+          //if duration is zero, contribute to current no progress counter
+          if(insertedActivity.getDuration().isZero()) {
+            this.stepsWithoutProgress++;
```
mattdailis: I'm not sure I agree with this definition of progress - I think it's okay if we added one zero duration activity, as long as we also added another activity with a non-zero duration, right?

I would think that progress is if the total duration has increased since the last round. (and lack of progress is if it has not)

Also - if we are trying to reach a certain number of occurrences, I think it's okay to place zero-duration activities. It's only if we have one of these cases that placing zero duration activities becomes a problem:
1. We have already satisfied the number of occurrences
2. Or we only had a durationRange goal, not an occurrences goal

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/BinaryMutexConstraint.java
mattdailis: `Windows` is mutable - `new Windows(windows)` is a copy constructor that creates a fresh copy that can be mutated independently of the source.

mattdailis: `Windows` is mutable - `new Windows(windows)` is a copy constructor that creates a fresh copy that can be mutated independently of the source.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
adrienmaillard: You are right, it is not required, I'll remove it.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
adrienmaillard: I have rewritten this to follow your idea.

adrienmaillard:
@dyst5422 I have started a branch for replacing number durations by Temporal.Duration but I think I need some help with the ts compilation/dependencies and stuff. I'll poke people.

dyst5422:
> @dyst5422 I have started a branch for replacing number durations by Temporal.Duration but I think I need some help with the ts compilation/dependencies and stuff. I'll poke people.

Yeah a separate PR is definitely the way to go on this. If you point me to the branch, I can pick up that task and get it PR'd

adrienmaillard:
@dyst5422 you could start off the prototype/AERIE-XXX--Temporal.Duration_in_scheduling_edsl branch

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -45,6 +49,19 @@ public class CardinalityGoal extends ActivityTemplateGoal {
    */
   protected TimeRangeExpression expr;
 
+  // Following members are related to diagnosis of the scheduler being stuck with inserting 0-duration activities
+  /**
+   * Activities inserted so far to satisfy this goal
+   */
+  private Map<SchedulingActivityInstanceId, ActivityInstance> insertedSoFar = new HashMap<>();
```
mattdailis: Since we never retrieve the values in this Map, I think we only need a Set:

```suggestion
  private Set<SchedulingActivityInstanceId> insertedSoFar = new HashSet<>();
```

mattdailis (approved):
Thanks for making those updates, looks good to me :+1:

PR #172, opened 2022-05-12
[AERIE-1838] Adds a simulated_activity view

dyst5422:
This implements a view that abstracts out the details of the simulation domain language into the domain language of activity planning so that users (and we) can interact within the domain language of the domain with which we are working.

* **Tickets addressed:** AERIE-1838
* **Review:** By commit  

## Description

I decided that reinventing the wheel of our other hasura types was #bad, so I went the route of just grouping together the necessary ids and allowing hasura to do the rest of the necessary joins to pull the actual data.

## Verification

I evaluated both by inspection of the SQL and by deploying it locally and ensuring the view was correctly deployed. Additionally, I tested it out with some local data.

Definitely interested in what other people think is a good test method for a change like this.

## Documentation

Would an updated table diagram showing the relationships between the view and the tables be useful here?

### DIFF
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
@@ -1,0 +1,50 @@
+table:
+  name: simulated_activity
+  schema: public
+object_relationships:
+- name: dataset
+  using:
+    manual_configuration:
+      remote_table:
+        name: dataset
+        schema: public
+      insertion_order: null
+      column_mapping:
+        dataset_id: id
+- name: parent_simulated_activity
+  using:
+    manual_configuration:
+      remote_table:
+        name: simulated_activity
+        schema: public
+      insertion_order: null
+      column_mapping:
+        parent_simulated_activity_id: simulated_activity_id
+- name: plan
+  using:
+    manual_configuration:
+      remote_table:
+        name: plan
+        schema: public
+      insertion_order: null
+      column_mapping:
+        plan_id: id
+- name: activity
+  using:
+    manual_configuration:
+      remote_table:
+        name: activity
+        schema: public
+      insertion_order: null
+      column_mapping:
+        activity_id: id
+- name: mission_model
+  using:
+    manual_configuration:
+      remote_table:
+        name: mission_model
+        schema: public
+      insertion_order: null
+      column_mapping:
+        model_id: id
+
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
@@ -2,49 +2,19 @@ table:
   name: simulated_activity
   schema: public
 object_relationships:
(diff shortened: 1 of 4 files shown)


dyst5422:
@patkenneally specifically looking for your input on this change. I know we had the conversation without you.

~~I feel this is a bit of a bandaid over some issues in our data model as I think there should be a column in the span field which enforces the foreign key constraint that is being related through the jsonb `directiveId` field.~~

~~I also think that using views to imbue our more domain agnostic simulation artifacts (span being the prime example) with a domain specific model is a good approach if we want to retain the domain agnostic data representation.~~

Reduced scope to merely abstracting out the simulation domain language to the activity planning domain language

dyst5422:
~~As I'm thinking about this more, it seems like it would be a better use of time to actually patch the deficiencies in our data model than bandaid over them.~~

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: simulated_activity
+  schema: public
+object_relationships:
+- name: dataset
+  using:
+    manual_configuration:
+      remote_table:
+        name: dataset
+        schema: public
+      insertion_order: null
```
mattdailis: Just curious what insertion_order means, and why it should be null - is that basically signifying that we should never insert into a view?

mattdailis (approved):
I think you've made the view option quite clean - the use of hasura to control the relationships seems like a good idea to me :+1:

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
dyst5422: I've honestly no idea, I pulled it in from pattern matching other metadata similarly

Twisol:
> I feel this is a bit of a bandaid over some issues in our data model as I think there should be a column in the span field which enforces the foreign key constraint that is being related through the jsonb `directiveId` field.

Data organized by the `dataset` table (including `span`s) may also be uploaded from non-Merlin sources, which may not have a concept of "input directive" in the first place. The `simulation_dataset` join table describes the relationship a given dataset has with an input `simulation` description, if any.

It might make sense to add a new `directive_span` table, conceptually associated with the `simulation_dataset`, that assigns to each directive the span that it produces under simulation; but we would probably want to set up yet another per-dataset partition in doing so.

Putting `directiveId` into the span attributes is the same as what a third-party dataset would need to do to make extra information available to a planner via the UI. I think a case can easily be made that the originating directive is a sufficiently special concept to Merlin that we should model it explicitly, but neither is it a concept native to all datasets.

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulated_activity_view.yaml
dyst5422: removed these because no idea what they were about

goetzrrGit (approved):
I verified I can see the new view in Hasura. This will be a good bandaid until we have a bigger meeting to discuss the change in our data models.

PR #173, opened 2022-05-12
[AERIE-1852] Command expansion authoring columns

dyst5422:
* **Tickets addressed:** AERIE-1852
* **Review:** By commit  

## Description

Added authoring informational columns to the expansion_rule table for the UI to use to default the type checking libs.

Additionally, performs type checking if the authoring information is provided and returns type errors purely as a programmatic API usage convenience, but does not block insertion on that type checking.

## Verification

Did a run through the expansion submission process locally. This is fully backwards compatible.

## Documentation

Might want to adjust the command authoring documentation, but this could also be considered an "advanced" feature that only the UI needs to really know about formally.

### DIFF
--- a/command-expansion-server/sql/commanding/tables/expansion_rule.sql
+++ b/command-expansion-server/sql/commanding/tables/expansion_rule.sql
@@ -4,11 +4,18 @@ create table expansion_rule (
   activity_type text not null,
   expansion_logic text not null,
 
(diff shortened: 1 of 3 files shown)


camargo (approved):
LGTM. Thanks for the quick update!

mattdailis (approved):

PR #174, opened 2022-05-12
[AERIE-1819] Build Constraints eDSL to parity with JSON implementation

JoelCourtney:
* **Tickets addressed:** AERIE-1819
* **Review:** By commit  

## Description

1. Implement basic Constraints eDSL
    - This attempts to replicate the existing JSON spec as ergonomically as possible.
    - The most important part is the bottom of `constraints-edsl-fluent-api.ts`, which declares the API that the constraints author will interact with. Everything else in the commit is the implementation and/or glue code for that interface.
2. Remove need for redundant IfThen backend representation
    - The TS API is capable of hiding a lot of behavior that we don't need to implement on the backend. Whenever a node in the expression tree can be represented purely as the composition of other nodes, it should be converted to those nodes by the API.
    - For example `IfThen(C, E)` is equivalent to `Or(Not(C), E)`, so the API performs the conversion immediately.
3. Remove outdated comment
    - Should have been removed in my last PR.
4. Automatic type conversion for literal Real and Discrete Values
    - While writing tests I found it tedious to write `Real.Value(n)` over and over. So I allowed the API to take literals directly in many functions, and wrap them in `Real.Value()` or `Discrete.Value()` if needed.
    - So `resource.lessThan(Real.Value(5))` can now be `resource.lessThan(5)`
5. Add tests for Constraints API
    - Adds tests to `ConstraintsDSLCompilationServiceTests.java` for all nodes of the API.

## Verification

Tests for all of the JSON parsers are included in the first commit.

Tests for all of the constraints API is the content of the last commit.

## Documentation
Please advise. A tutorial in the wiki is probably in order?

## Future work
- Type checking for Activity Types and parameters
- Delete the `IfThen` class when the old constraints pipeline is removed.
- Change the `And`/`Or` expression nodes to `All`/`Any` when the old constraints pipeline is removed.


merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -1,87 +1,324 @@
 import * as AST from './constraints-ast.js';
 
-interface ViolationsOf extends Constraint {}
-
 export class Constraint {
   /** Internal AST node */
   public readonly __astNode: AST.Constraint;
 
-  private constructor(astNode: AST.Constraint) {
+  public constructor(astNode: AST.Constraint) {
```
dyst5422, line 7: I left a similar comment for Ryan at some point.

The pattern of having a private constructor and a public new function is a bit of an interesting one.

 A constructor MUST be synchronous and must return the class type. A static new function can be asynchronous and does not need to return the class instance. This is useful for cases where you may need an async call as part of the initialization (time conversion via a service is the straightforward example). Or you may want to have a fallible object creation and return a Result type.

As a result of running into these issues requiring extensive refactoring that is less obvious a number of times, I've gotten into the habit of defaulting to the static new pattern.

I'm not saying this pattern is a necessary part of this or any codebase, just some food for thought on the matter.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -1,87 +1,324 @@
...
   public readonly __astNode: AST.WindowsExpression;
 
-  private constructor(expression: AST.WindowsExpression) {
+  public constructor(expression: AST.WindowsExpression) {
     this.__astNode = expression;
   }
 
-  private static new(expression: AST.WindowsExpression): WindowsExpression {
-    return new WindowsExpression(expression);
+  public static During(alias: string): Windows {
```
dyst5422, line 37: What are the aliases in these static cases referencing? Or rather, are the aliases required to exist a priori and can we enforce that in the type system?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -1,87 +1,324 @@
...
     });
   }
+  public not(): Windows {
+    return new Windows({
+      kind: AST.NodeKind.WindowsExpressionNot,
+      expression: this.__astNode
+    })
+  }
+
+  public violations(nominal: boolean = true): Constraint {
```
dyst5422: this nominal parameter appears unused

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -1,87 +1,324 @@
...
+  }
+  public notEqual(other: Real): Windows {
+    return new Windows({
+      kind: AST.NodeKind.ExpressionNotEqual,
+      left: this.__astNode,
+      right: other.__astNode
+    })
+  }
+
+  public changed(): Windows {
```
dyst5422, line 214: Does this produce windows of zero duration?

dyst5422 (commented):
Killer job!

The way I'm reading this and the tests, it's valid to return a Window since the underlying ast node is a part of the constraint node union. I'm curious why the type checker doesn't complain when you return a window though as we have the output type as `Constraint` and the types are mostly unrelated.

Additional note, if we can add doc comments for the methods (just above them), it'll show those in the editing ui for great UX.
/** doc comment */

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -235,14 +256,20 @@ export class Discrete {
     })
   }
 
-  public equal(other: Discrete): Windows {
+  public equal(other: any): Windows {
```
dyst5422, line 258: Matt, Adrien, and I came up with a clever solution for constraining these types in a huddle this morning. I'm sure Matt will be working with you on getting that put in here soon enough.

JoelCourtney: In this particular instance, the type is intentionally not constrained. It allows you to use `discreteResource.equal("A string")` instead of `discreteResource.equal(Discrete.Value("A string"))`, because I discovered that constantly wrapping all literals in `Discrete.Value` or `Real.Value` got very tedious, very fast. In the implementation for the function, it checks if the input is already `instanceof Discrete`, and if not will wrap it in `Discrete.Value`.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: The aliases are created by `ForEachActivity`. example:

```typescript
export default () => {
    return Constraint.ForEachActivity(
        "activity type",
        "activity alias",
        Windows.During("activity alias").not()
     )
}
```

I don't think it can be enforced.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Removed. Was an idea to allow the user to make `false` a violation, but I decided it wasn't worth since `.not()` is available.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Yes. The evaluation boils down to calling `Window.at()` when the value changes:
```java
public static Window at(final Duration point) {
    return new Window(point, Inclusive, point, Inclusive);
}
```

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
dyst5422: I think we actually can. I think the necessary parts are here where we demonstrate the constraining of a parameter based on another parameter and specifically for string literals.

[TypescriptPlayground](https://www.typescriptlang.org/play?#code/LAKALgngDgpgBAZTAJwJYDsDmAZVYbICGANgDwAqAfHALxzlwwAe+6AJgM5wcoaZwB+bryyMWMdlwZCeaUQC56cRbL4BuUJpAAzAK7oAxmFQB7dN1QBbKMRgBRJoWu2KY1p2FzMAGjgBxNwkPJC9cfCIyKkoACm1UZB5FEL4wghIKSl8OGAMzNkU-AEpFADcTVDY4AG8AXy0OKxt7R2cYaIByACYAZnbfLt7C0AbWhycmjp6+uHbugBZ2oZBQXPQeOBKSWhmpuABqGYWNZZARprHW6M3ifsGtA2JCDi4AIRMmV2Z3LlUsairQHA4AB9YEAIyI7EUDDo+jYMDi6BglSecEI6AgxyBq1kuiMJmQVxIuhg0MK1TqIEpoD0hmMZjgYPeyIuTVIgKUXyCPxEPg5AS5kkQvNSEQyoBiHLiCTASRFeDSkUyHOyq3ycDeHz81AAPnB0LpLGCCN5QMUNuVKrUtEymCyWhMBtMkQB3DXvSaDJa2+3jWye50wN2ayaLQpAA)

Which enforces consistency if its a string literal, but doesn't block constructed strings like `const a = 'a' + 'b';`

JoelCourtney: **y i k e s**

Ok I'll try that out. That'll be part of the upcoming typechecking PR. Thanks!

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: The reason for the change from `Discrete` to `Any` is to avoid wrapping values in `Discrete.Value(...)` when its redundant. If we could have a `Discrete<T>` type, then I would do the following:

```typescript
public equal(other: Discrete<T> | T): Windows {
    if (other instanceof T) {
        other = Discrete.Value(other);
    }
    // carry on
}
```

The problem is determining the type `T` when the left-hand operand is created from `Discrete.Resource` or `Discrete.Parameter`. I understand how I'd type-check the resource names, like this:

```typescript
public enum Resource {
    MyResource = "My Resource",
    Resource2 = "Resource Two"
}

// ... in the Discrete class

public static Resource(resource: Resource): Discrete {
    // do the thing
}
```

How would I transform that function into returning `Discrete<T>`?

JoelCourtney:
> The way I'm reading this and the tests, it's valid to return a Window since the underlying ast node is a part of the constraint node union. I'm curious why the type checker doesn't complain when you return a window though as we have the output type as Constraint and the types are mostly unrelated.

Its because `AST.WindowsExpression` is a valid production of the `AST.Constraint` type in `constraints-ast.ts`. You return a `Constraint`, which contains an instance of `AST.Constraint`, which is allowed to be an instance of `AST.WindowsExpression`.

Then the Java JSON parsers recognize that the JSON node which was expected to be a Constraint is actually a WindowsExpression, and wraps the expression in `ViolationsOf`. Personally I'm not attached to this pattern, and I've already made a `window.violations()` function if you want to make the conversion explicit. It would be trivial to turn off the implicit way.

I had a discussion with Basak, Adrien, and Chris about this today. Basak brought up that long-term we might want to use constraints for other purposes like determining a window where some condition is true without making violations. With that in mind it might be nice to require the user to end the constraint in `window.violations()`, so that later we could add the option for `window.show()` or something like that.

Thoughts? If no one objects I think that's what I'll go with.

JoelCourtney:
Also I'd like to change `window.and(...others)` and `window.or(...others)` to `Windows.All(...windows)` and `Windows.Any(...others)`. Its minor but `and` and `or` are mixing the notions of windows and booleans unnecessarily. (Thanks @dyst5422 for pointing that out when we were talking about `IfThen`.)

edit: done. Will squash that commit when merging. Leaving it separate for now just to be clear.

JoelCourtney:
> Additional note, if we can add doc comments for the methods (just above them), it'll show those in the editing ui for great UX.
> /** doc comment */

Done

dyst5422 (approved):
Killer work Joel!

mattdailis (approved):
Looks great to me :+1:

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: marking resolved, we can talk about this as part of the next PR.

PR #175, opened 2022-05-16
[AERIE-1838] Add activity_id to span

dyst5422:
This is an alternative approach to the simulated_activity_instance view
that creates foreign key relationship column on span with activity

* **Tickets addressed:** AERIE-1838
* **Review:** By commit  

## Description

Adds a foreign key constraint to the span column with the activity column

## Verification

Did local testing

### DIFF
--- a/merlin-server/sql/merlin/tables/span.sql
+++ b/merlin-server/sql/merlin/tables/span.sql
@@ -3,6 +3,7 @@ create table span (
 
   dataset_id integer not null,
   parent_id integer null,
+  activity_id integer null,
 
   start_offset interval not null,
   duration interval null,
@@ -20,6 +21,11 @@ create table span (
     foreign key (dataset_id, parent_id)
     references span
     on update cascade
+    on delete cascade,
+  constraint span_has_activity
+    foreign key (activity_id)
+    references activity
+    on update cascade
     on delete cascade
 )
 partition by list (dataset_id);
@@ -34,6 +40,8 @@ comment on column span.dataset_id is e''
   'The dataset this span is part of.';
 comment on column span.parent_id is e''
   'The span this span refines.';
+comment on column span.activity_id is e''
+  'The activity this span represents in the simulation.';
 comment on column span.start_offset is e''
   'The offset from the dataset start at which this span begins.';
 comment on column span.duration is e''
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_span.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_span.yaml
@@ -12,6 +12,12 @@ object_relationships:
       remote_table:
         name: span
         schema: public
+- name: activity
+  using:
+    foreign_key_constraint_on: activity_id
+- name: dataset
+  using:
+    foreign_key_constraint_on: dataset_id
 array_relationships:
 - name: spans
   using:
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
@@ -19,8 +19,8 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatemen
 
 /*package-local*/ final class PostSpansAction implements AutoCloseable {
   private static final @Language("SQL") String sql = """
-      insert into span (dataset_id, start_offset, duration, type, attributes)
-      values (?, ?::timestamptz - ?::timestamptz, ?::timestamptz - ?::timestamptz, ?, ?)
+      insert into span (dataset_id, start_offset, duration, type, activity_id, attributes)
+      values (?, ?::timestamptz - ?::timestamptz, ?::timestamptz - ?::timestamptz, ?, ?, ?)
     """;
 
   private final PreparedStatement statement;
@@ -52,7 +52,12 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatemen
 
       setTimestamp(statement, 5, startTimestamp);
       statement.setString(6, act.type());
-      statement.setString(7, buildAttributes(act.attributes().directiveId(), act.attributes().arguments(), act.attributes().computedAttributes()));
+      if (act.attributes().directiveId().isPresent()){
+        statement.setLong(7, act.attributes().directiveId().get());
+      } else {
+        statement.setNull(7, java.sql.Types.BIGINT);
+      }
+      statement.setString(8, buildAttributes(act.attributes().directiveId(), act.attributes().arguments(), act.attributes().computedAttributes()));
       statement.addBatch();
     }
 


dyst5422:
@mattdailis @JoelCourtney @adrienmaillard I am once again asking for your PR review

Bernie-I-Am-Once-Again-Asking-For-Your-Support


Need this or the alternative PR to unblock sequence workflow

merlin-server/sql/merlin/tables/span.sql
```diff
@@ -20,6 +21,11 @@ create table span (
     foreign key (dataset_id, parent_id)
     references span
     on update cascade
+    on delete cascade,
+  constraint span_has_activity
```
mattdailis, line 25: The reason I don't think we want to couple the span and activity tables is that they have potentially different lifecycles. If I delete an activity from a plan, that doesn't necessarily mean I want to delete the span from the simulation results.

dyst5422: Fair point, so remove the cascading then, yeah?

dyst5422: Ooh. If we remove the activity a span references, is there a way to tell PG to convert it from the activity id to null?

dyst5422: Answer: yes. `on delete set null`

dyst5422: Alternatively, we could use a join table for this

mattdailis: I do think that's do-able (something like `on delete set null`), however I'm still wondering about the wisdom of doing that.

We have multiple independent actors here - someone editing the plan, and the merlin worker(s) running simulations. If we force plan updates to also update simulation results in a transaction, I wonder if we'll run into any database locking issues 🤔 I need to read up a bit to say for sure.

mattdailis: > Alternatively, we could use a join table for this

This _might_ provide the decoupling we need - we'd need to think through how and when the join table gets updated

patkenneally: I'm pondering this PR generally also. A knee jerk reaction is that a join table might be better here.

dyst5422: I think one of the issues we're generally bumping against here is immutable data referencing mutable data.

We're treating simulation output and downstream items as immutable (commands and sequences) but things prior to simulation as mutable (plans, simulations).

I don't think we've really established the proper way to interface the two.

Furthermore, I'm unsure of whether we've consciously made the decision about whether mutable or immutable is the proper model for many parts of the system.

In my mind, we've kind of got this hybrid approach because of how we are representing a plan in the UI (and thus in great part the mental model we adopt) as both a `plan` and as a simulation result (displaying activities and spans equivalently). I'd be interested in how a discussion of the approach there as it relates to our data model might result in conscious decisions on the data mutability model.

tungxbui: What is the strategy here for changes to a plan and have simulation updates at the same time? Sounds like this is your key problem to all this questions. Ideally you don't want to have plan changes while a simulation is occurring. From the users' point of view, that would be inconvenient to have to wait for a simulation to finish before make any changes. One of the ways to deal with this is to not really delete a plan but rather mark it deleted so that the simulation data still have something to hang onto. Some form of versioning approach. Do we decide on how this be done? That may leads you to the right solution here.

dyst5422: @tungxbui I think you're suggesting we move the immutable approach all the way up to the plan level and create a new object for each change. This would definitely introduce data storage concerns but would resolve much of our referential integrity problems.

dyst5422: I'm going to pose a question pointed in the alternative direction. Do we really need immutable data downstream? Do we need to maintain multiple simulation outputs for a plan? Do we need to retain multiple expansion outputs for a given plan?

Or is there a valid case for use saying "there is a plan, a single simulation output for it that is invalidated and cleared on plan edit, a single command output for it that is cleared on the above clearing" such that we cascade mutation downstream.

I genuinely don't have any idea which approach fits our use case better. I know I concern about data traceability from a spacecraft ops perspective, but that might be something we push down to the missions to maintain if they need it.

tungxbui: @dyst5422, that is just one way but I think we have to come up with a reasonable way to deal with this and then handle consistently from there on. Otherwise we will be back to this problem again. The data can be clean up after when the simulation is done and it detects that the plan has now obsolete. That is another way to keep the data storage down. But otherwise we might end up with data integrity problem when you have multiple actors change various part of the plan at the same time. It will be hard to keep the data consistent as one does not know what the other does.

tungxbui: Hmmm, I don't even aware that we retain multiple expansion outputs for a plan. That is a good question. This sounds like even more difficult to keep everything straight. I don't have answer to that. Do we have a requirement for this or just something we think it maybe a good idea?

dyst5422: Right now, I think its a result of maintaining multiple simulation results, which directly drive the command generation

tungxbui: @dyst5422, I do like this idea:  "we move the immutable approach all the way up to the plan level." I think it will keep thing consistent. Also, that will keep the history of the changes as you suggest that missions may want to have this. The clean up can be done as part of operation process. As of right now, you can't really tell which change comes first, which comes later or even what exactly is the change as part of it can be overwritten by later changes.

dyst5422:
Another way to rephrase this problem is as a cache invalidation issue:

We have few mutable source resources - the plan, simulation, expansions sets/rules, external datasets. Everything downstream from those is the cached result of some processing - simulation results, expanded commands, etc. Right now, we're not invalidating that cache at all, just inserting additional results into the cache.

I think of a similar case that we are actually treating as a process output is sequence SeqJson output where we generate it on request.

Twisol:
I replied to the other PR (#172), but I think [my comments there](https://github.com/NASA-AMMOS/aerie/pull/172#issuecomment-1129259225) are more relevant to the discussion happening here.

> > I feel this is a bit of a bandaid over some issues in our data model as I think there should be a column in the span field which enforces the foreign key constraint that is being related through the jsonb `directiveId` field.
> 
> Data organized by the `dataset` table (including `span`s) may also be uploaded from non-Merlin sources, which may not have a concept of "input directive" in the first place. The `simulation_dataset` join table describes the relationship a given dataset has with an input `simulation` description, if any.
> 
> It might make sense to add a new `directive_span` table, conceptually associated with the `simulation_dataset`, that assigns to each directive the span that it produces under simulation; but we would probably want to set up yet another per-dataset partition in doing so.
> 
> Putting `directiveId` into the span attributes is the same as what a third-party dataset would need to do to make extra information available to a planner via the UI. I think a case can easily be made that the originating directive is a sufficiently special concept to Merlin that we should model it explicitly, but neither is it a concept native to all datasets.

merlin-server/sql/merlin/tables/span.sql
Twisol: > In my mind, we've kind of got this hybrid approach because of how we are representing a plan in the UI (and thus in great part the mental model we adopt) as both a `plan` and as a simulation result (displaying activities and spans equivalently). I'd be interested in how a discussion of the approach there as it relates to our data model might result in conscious decisions on the data mutability model.

I rather deeply agree with you here. One of the things that's bothered me for a while is that, in the process of generating simulation results, we assign a `directiveId` to every span, even those produced by child activities (which thus don't have an associated *input* directive). We have to be careful to assign fresh IDs to children, but reuse the input IDs for root spans. IIRC (@camargo?), this makes it easier for the UI to look up the directives associated with a span, whilst still identifying each span uniquely. I would much rather give spans and directive independent ID spaces, and explicitly join between them. (Much as in my `directive_span` suggestion elsewhere in this PR discussion.)

In my mind, directives are point-like, and spans are durative; directives are specifications of intent, and spans are reports of effect. Previous planning tools conflate the two, and have been successful for it; but I'm not convinced our core data model needs to follow suit, even if our visualizations do.

> **Tung:** One of the ways to deal with this is to not really delete a plan but rather mark it deleted so that the simulation data still have something to hang onto. Some form of versioning approach

> **Dylan:** I'm going to pose a question pointed in the alternative direction. Do we really need immutable data downstream? Do we need to maintain multiple simulation outputs for a plan? Do we need to retain multiple expansion outputs for a given plan?

Plan versioning has been on our radar since... forever, really. I'm a big fan of an approach based on immutability, but to date we've justified the current approach as "a simulation is a snapshot of the effects of some prior version of the plan". You can see that in the tables involved, we maintain `revision` columns; so if you see a `simulation_dataset` referencing a revision of a `plan` that has a newer revision number, you know that the simulation is out of date, and that the input plan has been, per se, "lost".

I would be quite interested in a *snapshot*-based approach, where plan HEAD is always mutable, but resimulation requires a snapshot to be taken (whether automatically or manually). Planners could always go back and view a saved revision of a plan, and even delete it (also deleting its cached simulation results). Allowing the HEAD revision to be mutable also avoids a lot of the data retention issues, since it effectively coalesces any changes to a plan in between simulations.

(I would *not* worry about branching history for plan revisions at all at this point. For now, if someone wants to edit a saved plan snapshot, they can clone a new plan from that revision and begin working from its HEAD.)

Twisol: One big reason to save multiple simulations is to do a differential analysis between different options. You could do that with two separate plans, but that seems somewhat unnatural from a planner's perspective. In particular, saving simulation results as a planner refines their plan makes it much easier to validate that your changes have, in fact, improved the risk and reward profile. I believe you got this pretty much for free with APGen, since simulation results are/can be just output to the filesystem normally (and then uploaded for simultaneous viewing in, say, Raven). In such a context, it would be weird to invalidate old runs just because you prepared new inputs.

Twisol:
> but we would probably want to set up yet another per-dataset partition in doing so.

Maybe we wouldn't need to do that; after all, the `activity` table isn't partitioned. I think that's because we estimated that the lifetime corpus of plans for a mission is probable too small to be worth setting up per-plan partitions. Maybe temporal partitioning will make sense eventually? At any rate, it's probably not worth the time or effort to partition a `directive_span` table either, since its size will be proportional to `activity`.

merlin-server/sql/merlin/tables/span.sql
tungxbui: @Twisol, you have many good points there. It does confirm that we do need an approach in some shape or form to allow different versions of this data set to be retained and reviewed. A more complete versioning, it sounds like.

dyst5422:
Closing for now. The long term architectural design we use to address the connection between simulation outputs and plan or between different simulation result sets is going to be a big conversation.

For now, we're going with the view solution to the immediate needs of the project

merlin-server/sql/merlin/tables/span.sql
dyst5422: I agree generally with the immutable data approach, but felt it was worth asking and having the conversation about the needs specific to this project and whether the needs to do such things as differential analysis actual fell within the required feature set or was simple an aspirational behavior we'd like to see in planning.

I'm also a fan of the snapshot approach and Basak and I discussed it earlier today. We still however haven't addressed one of the usability issues associated with downstream usage of simulation outputs - transferring user injected information from one simulation result set to another. Prime case: associating simulated activities (spans in our parlance) with sequences and not having that be destroyed with subsequent re-sims.

PR #176, opened 2022-05-17
[AERIE-1843] Add constraintsDslTypescript hasura action

JoelCourtney:
* **Tickets addressed:** AERIE-1843
* **Review:** By commit  

## Description
Adds a hasura action to get generated typescript files for the constraints dsl.

Action name is `constraintsDslTypescript`, and takes a mission model id int (or string, see discussion with Chris below) as input. Example query:

```graphql
query MyQuery {
  constraintsDslTypescript(missionModelId: 1) {
    reason
    status
    typescriptFiles {
      content
      filePath
    }
  }
}
```

Currently it doesn't use the missionModelId argument, but it will be used so I included it here.

## Verification
Try it out in the hasura console!

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -76,7 +76,13 @@ type Query {
 type Query {
   schedulingDslTypescript(
     missionModelId: Int!
-  ): SchedulingDslTypescriptResponse
+  ): DslTypescriptResponse
+}
+
+type Query {
+  constraintsDslTypescript(
+    missionModelId: ID!
+  ): DslTypescriptResponse
 }
 
 type Query {
@@ -112,11 +118,6 @@ enum SchedulingStatus {
   incomplete
 }
 
-enum SchedulingDslTypescriptResponseStatus {
-  success
-  failure
-}
-
 type ResourceType {
   name: String!
   schema: ResourceSchema!
@@ -148,11 +149,16 @@ type SchedulingResponse {
   reason: SchedulingFailureReason
 }
 
-type SchedulingDslTypescriptResponse {
-  status: SchedulingDslTypescriptResponseStatus!
+type DslTypescriptResponse {
+  status: DslTypescriptResponseStatus!
   typescript: String
   typescriptFiles: [TypescriptFile]
   reason: String
+}
+
+enum DslTypescriptResponseStatus {
+  success
+  failure
 }
 
 type TypescriptFile {
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -151,7 +151,6 @@ type SchedulingResponse {
 
 type DslTypescriptResponse {
   status: DslTypescriptResponseStatus!
-  typescript: String
   typescriptFiles: [TypescriptFile]
   reason: String
 }
@@ -162,8 +161,8 @@ enum DslTypescriptResponseStatus {
 }
 
 type TypescriptFile {
-  filename: String,
-  contents: String
+  filePath: String,
+  content: String
 }
 
 type CommandDictionaryResponse {
--- a/deployment/hasura/metadata/actions.yaml
+++ b/deployment/hasura/metadata/actions.yaml
@@ -47,6 +47,10 @@ actions:
     definition:
       kind: ""
       handler: http://aerie_scheduler:27185/schedulingDslTypescript
+  - name: constraintsDslTypescript
+    definition:
+      kind: ""
+      handler: http://aerie_merlin:27183/constraintsDslTypescript
   - name: simulate
     definition:
       kind: ""
@@ -101,7 +105,7 @@ custom_types:
     - name: EffectiveArgumentsResponse
     - name: AddExternalDatasetResponse
     - name: SchedulingResponse
-    - name: SchedulingDslTypescriptResponse
+    - name: DslTypescriptResponse
     - name: CommandDictionaryResponse
     - name: AddCommandExpansionResponse
     - name: TypeScriptResponse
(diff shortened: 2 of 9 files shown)


deployment/hasura/metadata/actions.graphql
```diff
@@ -148,13 +149,18 @@ type SchedulingResponse {
   reason: SchedulingFailureReason
 }
 
-type SchedulingDslTypescriptResponse {
-  status: SchedulingDslTypescriptResponseStatus!
+type DslTypescriptResponse {
+  status: DslTypescriptResponseStatus!
   typescript: String
```
mattdailis: Since there's no UI code calling this action yet, let's go ahead and skip the `typescript` field, and just return `typescriptFiles`.

JoelCourtney: That's the same struct returned by the scheduling code action, so it would break the UI code for scheduling DSL.

mattdailis: Ah, I missed that, okay :+1: we can drop them both in one fell swoop once the UI is ready

deployment/hasura/metadata/actions.graphql
```diff
@@ -148,13 +149,18 @@ type SchedulingResponse {
...
   reason: String
 }
 
+enum DslTypescriptResponseStatus {
+  success
+  failure
+}
+
 type TypescriptFile {
   filename: String,
```
camargo: What would it take to change `contents` to `content`, and `filename` to `filePath`? This would allows us to pass a `TypescriptFile[]` directly to Monaco :)

Twisol: As a curious observer -- is there a way (maybe interface subtyping?) to formalize the alignment you want with Monaco, so that the intent to use this data with Monaco is clear to both us humans and the compiler? (Or, once we're passing things to Monaco, will the structural type system reliably give us an error regardless?)

EDIT: I just realized this is part of the GraphQL model for the action, so it wouldn't make sense / wouldn't be possible to upstream any particular Monaco signature into this model. My question reduces to "will TypeScript catch us out if the field names change in the future?"

deployment/hasura/metadata/actions.graphql
```diff
@@ -76,7 +76,13 @@ type Query {
 type Query {
   schedulingDslTypescript(
     missionModelId: Int!
-  ): SchedulingDslTypescriptResponse
+  ): DslTypescriptResponse
+}
+
+type Query {
+  constraintsDslTypescript(
+    missionModelId: ID!
```
camargo, line 84: Is there a reason `ID!` is used here instead of `Int!`? We should probably be consistent across the different queries.

deployment/hasura/metadata/actions.graphql
camargo: Yeah another option I'm thinking is to just alias the query on the client. But I'm all for making the type alignment more formal.

deployment/hasura/metadata/actions.graphql
JoelCourtney: Yep, I re-used the `hasuraMissionModelActionP` parser in merlin, which expects to deserialize the argument as a string. The existing action that already uses this parser is defined with `ID!` instead of `Int!`.

The query can accept an int as well though, so mechanically the UI wouldn't have to do anything different.

I can also change it to use `Int` because that's what it should be, but there's quite a few of them. Probably best to open a small follow-up PR later this week.

deployment/hasura/metadata/actions.graphql
JoelCourtney: Yeah I'll make that change, but you'll have to update the scheduling Dsl request too. They return the same struct.

dyst5422: There is a great package @graphql-codegen to manage the typescript graphql typing interface. @camargo likely something you might want to use long term to avoid issues

camargo: Yup good recommendation. Have used it on projects before and it's nice.

JoelCourtney: Oh, right the UI doesn't use this field yet. @camargo Since you're making that change, can I go ahead and remove the `DslTypescriptResponse.typescript` field while we're at it?

camargo: Yes no problem on my end

JoelCourtney: done

deployment/hasura/metadata/actions.graphql
JoelCourtney: Created MAPSS-1855 for this

camargo (approved):
LGTM! Tested against the front-end branch and it works well.

mattdailis (approved):

pcrosemurgy:
Hey @JoelCourtney, silly question: do we need to worry about escaping within `content`? On the command expansion server we perform base64 decoding/encoding. From memory this is mainly just to make client-side use a bit easier since escaping logic isn't really needed with this encoding.

Tagging @goetzrrGit since he may also be able to answer this. :)

mattdailis:
> Hey @JoelCourtney, silly question: do we need to worry about escaping within `content`? On the command expansion server we perform base64 decoding/encoding. From memory this is mainly just to make client-side use a bit easier since escaping logic isn't really needed with this encoding.
> 
> Tagging @goetzrrGit since he may also be able to answer this. :)

My understanding is JSON stringify handles all of that, and the base64 is an additional layer of encoding that we've added on top 😅 My personal opinion is that we have no need to add the extra layer of encoding - it turned out that the issues we had with command dictionaries were the result of double-parsing a string (nothing to do with escaping characters). It would be good for us to reach a consensus on this so that our api can be more uniform.

pcrosemurgy:
Thanks, that makes a lot of sense. Sounds like our use of base64 in the command expansion server may be a bit paranoid. Good to know that isn't needed here at least.

dyst5422:
Thanks @mattdailis  I will make a JIRA that I can PR this week to remove our base64 encoding on the commanding side of things.

PR #183, opened 2022-05-25
[AERIE-1842] Type-checking for Constraints eDSL

JoelCourtney:
* **Tickets addressed:** AERIE-1842
* **Review:** By commit  

## Description
Adds type-checking for the following values in the constraints DSL:
- Activity type names
- Resource names
    - All resources, including Reals, are considered valid Discrete resources. Real resource names only accept the subset of resources that have a value schema exactly equal to `ValueSchema.REAL`.
- Activity aliases generated by `ForEachActivity` - kind of
    - Instead of enforcing that a specific string is declared and used to reference an alias, the user instead now provides a function of type `(instance: ActivityInstance<A extends ActivityTypeName>) => Constraint`. Essentially, the alias is no longer a string but a variable. (Thanks @mattdailis for the idea!)
    - Since the expression tree in the backend still expects the old string-based alias approach, `ForEachActivity` calls `.toString()` on the function handle and searches through to find the argument name. The "alias" string is set to the argument name and stored inside the `ActivityInstance` class.
    - Activity parameters are accessed through `instance.parameters().myParameter`. The `During`, `StartOf`, and `EndOf` functions are now `instance.during()`, `instance.start()` and `instance.end()`.
    - Example usage:
```typescript
// Forbids the usage of an activity type.
export default (): Constraint => Constraint.ForEachActivity(
    "activityType",
    (instance) => instance.during().not()
);
```
- Schemas for discrete profiles
    - The schemas of all discrete profiles are checked, including those from `Discrete.Resource(...)` and `instance.parameters().myDiscreteParameter`. This is done by adding a `__schemaInstance: Schema` private field to the `Discrete<Schema>` class, and populating it with an unrelated dummy default value.

## Verification
Many tests have been added and updated in `ConstraintsDSLCompilationServiceTests`.

## Documentation
The in-code docs for the constraints API have been updated.

## Future work
- In my opinion we are ready to start expanding the DSL with new features and remove the JSON version.
- Wiki tutorial
- Video demo

## Checklist

checkmark = committed

- [x] "Lets go ahead and make ActivityInstance#parameters be a getter"
- [x] "Can we make the ActivityType an enum instead of a union of strings"
- [x] "Who the heck invited this `__schemaInstance` field and how do we make it go away"
- [x] Autoformat
- [x] Bring back `once` in main.ts
- [x] squash a few commits

### DIFF
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -1,4 +1,5 @@
 import * as AST from './constraints-ast.js';
+import './mission-model-generated-code.js';
 
 export class Constraint {
   /** Internal AST node */
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -9,7 +9,7 @@ export class Constraint {
     this.__astNode = astNode;
   }
 
-  public static ForbiddenActivityOverlap(activityType1: string, activityType2: string): Constraint {
+  public static ForbiddenActivityOverlap(activityType1: ActivityTypeName, activityType2: ActivityTypeName): Constraint {
     return new Constraint({
       kind: AST.NodeKind.ForbiddenActivityOverlap,
       activityType1,
@@ -17,7 +17,7 @@ export class Constraint {
     })
   }
 
-  public static ForEachActivity(activityType: string, alias: string, constraint: Constraint): Constraint {
+  public static ForEachActivity(activityType: ActivityTypeName, alias: string, constraint: Constraint): Constraint {
     return new Constraint({
       kind: AST.NodeKind.ForEachActivity,
       activityType,
@@ -296,7 +296,7 @@ declare global {
      * @param activityType2
      * @constructor
      */
-    public static ForbiddenActivityOverlap(activityType1: string, activityType2: string): Constraint
+    public static ForbiddenActivityOverlap(activityType1: ActivityTypeName, activityType2: ActivityTypeName): Constraint
 
     /**
      * Check a constraint for each instance of an activity type.
@@ -306,7 +306,7 @@ declare global {
      * @param constraint constraint to apply
      * @constructor
      */
-    public static ForEachActivity(activityType: string, alias: string, constraint: Constraint): Constraint
+    public static ForEachActivity(activityType: ActivityTypeName, alias: string, constraint: Constraint): Constraint
   }
 
   export class Windows {
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -273,7 +273,7 @@ declare global {
      * Check a constraint for each instance of an activity type.
      *
      * @param activityType activity type to check
-     * @param expression
+     * @param expression function of an activity instance that returns a Constraint
      * @constructor
      */
     public static ForEachActivity<A extends Gen.ActivityTypeName>(activityType: A, expression: (instance: Gen.ActivityInstance<A>) => Constraint): Constraint
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -15,19 +15,22 @@ export class Constraint {
     return new Constraint({
       kind: AST.NodeKind.ForbiddenActivityOverlap,
       activityType1,
-      activityType2
-    })
+      activityType2,
+    });
   }
 
-  public static ForEachActivity<A extends Gen.ActivityType>(activityType: A, expression: (instance: Gen.ActivityInstance<A>) => Constraint): Constraint {
-    let alias = "activity alias " + Constraint.__numGeneratedAliases;
+  public static ForEachActivity<A extends Gen.ActivityType>(
+    activityType: A,
+    expression: (instance: Gen.ActivityInstance<A>) => Constraint,
+  ): Constraint {
+    let alias = 'activity alias ' + Constraint.__numGeneratedAliases;
     Constraint.__numGeneratedAliases += 1;
     return new Constraint({
       kind: AST.NodeKind.ForEachActivity,
       activityType,
       alias,
-      expression: expression(new Gen.ActivityInstance(activityType, alias)).__astNode
-    })
+      expression: expression(new Gen.ActivityInstance(activityType, alias)).__astNode,
+    });
   }
 }
 
@@ -42,17 +45,13 @@ export class Windows {
   public static All(...windows: Windows[]): Windows {
     return new Windows({
       kind: AST.NodeKind.WindowsExpressionAll,
-      expressions: [
-        ...windows.map(other => other.__astNode),
-      ],
+      expressions: [...windows.map(other => other.__astNode)],
     });
   }
   public static Any(...windows: Windows[]): Windows {
     return new Windows({
       kind: AST.NodeKind.WindowsExpressionAny,
-      expressions: [
-        ...windows.map(other => other.__astNode),
-      ],
+      expressions: [...windows.map(other => other.__astNode)],
     });
   }
 
@@ -62,25 +61,25 @@ export class Windows {
       expressions: [
         {
           kind: AST.NodeKind.WindowsExpressionNot,
-          expression: condition.__astNode
+          expression: condition.__astNode,
         },
-        this.__astNode
-      ]
-    })
+        this.__astNode,
+      ],
+    });
   }
 
   public not(): Windows {
     return new Windows({
       kind: AST.NodeKind.WindowsExpressionNot,
-      expression: this.__astNode
-    })
+      expression: this.__astNode,
+    });
   }
 
   public violations(): Constraint {
     return new Constraint({
       kind: AST.NodeKind.ViolationsOf,
-      expression: this.__astNode
-    })
+      expression: this.__astNode,
+    });
   }
 }
 
@@ -94,28 +93,28 @@ export class Real {
   public static Resource(name: Gen.RealResourceName): Real {
     return new Real({
       kind: AST.NodeKind.RealProfileResource,
-      name
-    })
+      name,
+    });
   }
   public static Value(value: number): Real {
     return new Real({
       kind: AST.NodeKind.RealProfileValue,
-      value
-    })
+      value,
+    });
   }
 
   public rate(): Real {
     return new Real({
       kind: AST.NodeKind.RealProfileRate,
-      profile: this.__astNode
-    })
+      profile: this.__astNode,
+    });
   }
   public times(multiplier: number): Real {
     return new Real({
       kind: AST.NodeKind.RealProfileTimes,
       multiplier,
-      profile: this.__astNode
-    })
+      profile: this.__astNode,
+    });
   }
   public plus(other: Real | number): Real {
     if (!(other instanceof Real)) {
@@ -124,8 +123,8 @@ export class Real {
     return new Real({
       kind: AST.NodeKind.RealProfilePlus,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
 
   public lessThan(other: Real | number): Windows {
@@ -135,8 +134,8 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.RealProfileLessThan,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
   public lessThanOrEqual(other: Real | number): Windows {
     if (!(other instanceof Real)) {
@@ -145,8 +144,8 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.RealProfileLessThanOrEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
   public greaterThan(other: Real | number): Windows {
     if (!(other instanceof Real)) {
@@ -155,8 +154,8 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.RealProfileGreaterThan,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
   public greaterThanOrEqual(other: Real | number): Windows {
     if (!(other instanceof Real)) {
@@ -165,8 +164,8 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.RealProfileGreaterThanOrEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
 
   public equal(other: Real | number): Windows {
@@ -176,8 +175,8 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.ExpressionEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
   public notEqual(other: Real | number): Windows {
     if (!(other instanceof Real)) {
@@ -186,20 +185,20 @@ export class Real {
     return new Windows({
       kind: AST.NodeKind.ExpressionNotEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
 
   public changed(): Windows {
     return new Windows({
       kind: AST.NodeKind.ProfileChanged,
-      expression: this.__astNode
-    })
+      expression: this.__astNode,
+    });
   }
 }
 
 export class Discrete<Schema> {
-  public readonly __astNode: AST.DiscreteProfileExpression
+  public readonly __astNode: AST.DiscreteProfileExpression;
 
   public constructor(profile: AST.DiscreteProfileExpression) {
     this.__astNode = profile;
@@ -208,14 +207,14 @@ export class Discrete<Schema> {
   public static Resource<R extends Gen.ResourceName>(name: R): Discrete<Gen.DiscreteResourceSchema<R>> {
     return new Discrete({
       kind: AST.NodeKind.DiscreteProfileResource,
-      name
-    })
+      name,
+    });
   }
   public static Value<Schema>(value: Schema): Discrete<Schema> {
     return new Discrete({
       kind: AST.NodeKind.DiscreteProfileValue,
-      value
-    })
+      value,
+    });
   }
 
   public transition(from: Schema, to: Schema): Windows {
@@ -223,8 +222,8 @@ export class Discrete<Schema> {
       kind: AST.NodeKind.DiscreteProfileTransition,
       profile: this.__astNode,
       from,
-      to
-    })
+      to,
+    });
   }
 
   public equal(other: Schema | Discrete<Schema>): Windows {
@@ -234,8 +233,8 @@ export class Discrete<Schema> {
     return new Windows({
       kind: AST.NodeKind.ExpressionEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
   public notEqual(other: Schema | Discrete<Schema>): Windows {
     if (!(other instanceof Discrete)) {
@@ -244,15 +243,15 @@ export class Discrete<Schema> {
     return new Windows({
       kind: AST.NodeKind.ExpressionNotEqual,
       left: this.__astNode,
-      right: other.__astNode
-    })
+      right: other.__astNode,
+    });
   }
 
   public changed(): Windows {
     return new Windows({
       kind: AST.NodeKind.ProfileChanged,
-      expression: this.__astNode
-    })
+      expression: this.__astNode,
+    });
   }
 }
 
@@ -267,7 +266,10 @@ declare global {
      * @param activityType2
      * @constructor
      */
-    public static ForbiddenActivityOverlap(activityType1: Gen.ActivityType, activityType2: Gen.ActivityType): Constraint
+    public static ForbiddenActivityOverlap(
+      activityType1: Gen.ActivityType,
+      activityType2: Gen.ActivityType,
+    ): Constraint;
 
     /**
      * Check a constraint for each instance of an activity type.
@@ -276,7 +278,10 @@ declare global {
      * @param expression function of an activity instance that returns a Constraint
      * @constructor
      */
-    public static ForEachActivity<A extends Gen.ActivityType>(activityType: A, expression: (instance: Gen.ActivityInstance<A>) => Constraint): Constraint
+    public static ForEachActivity<A extends Gen.ActivityType>(
+      activityType: A,
+      expression: (instance: Gen.ActivityInstance<A>) => Constraint,
+    ): Constraint;
   }
 
   export class Windows {
@@ -290,7 +295,7 @@ declare global {
      *
      * @param windows any number of windows expressions
      */
-    public static All(...windows: Windows[]): Windows
+    public static All(...windows: Windows[]): Windows;
 
     /**
      * Produce a window when any argument produces a window.
@@ -299,19 +304,19 @@ declare global {
      *
      * @param windows one or more windows expressions
      */
-    public static Any(...windows: Windows[]): Windows
+    public static Any(...windows: Windows[]): Windows;
 
     /**
      * Only check this expression of the condition argument produces a window.
      *
      * @param condition
      */
-    public if(condition: Windows): Windows
+    public if(condition: Windows): Windows;
 
     /**
      * Negate all the windows produced this.
      */
-    public not(): Windows
+    public not(): Windows;
 
     /**
      * Produce a constraint violation whenever this does NOT produce a window.
@@ -319,7 +324,7 @@ declare global {
      * Essentially, express the condition you want to be satisfied, then use
      * this method to produce a violation whenever it is NOT satisfied.
      */
-    public violations(): Constraint
+    public violations(): Constraint;
   }
 
   export class Real {
@@ -331,77 +336,77 @@ declare global {
      * @param name
      * @constructor
      */
-    public static Resource(name: Gen.RealResourceName): Real
+    public static Resource(name: Gen.RealResourceName): Real;
 
     /**
      * Create a constant real profile for all time.
      * @param value
      * @constructor
      */
-    public static Value(value: number): Real
+    public static Value(value: number): Real;
 
     /**
      * Create a real profile from this profile's derivative.
      */
-    public rate(): Real
+    public rate(): Real;
 
     /**
      * Create a real profile by multiplying this profile by a constant
      * @param multiplier
      */
-    public times(multiplier: number): Real
+    public times(multiplier: number): Real;
 
     /**
      * Create a real profile by adding this and another real profile together.
      * @param other
      */
-    public plus(other: Real | number): Real
+    public plus(other: Real | number): Real;
 
     /**
      * Produce a window whenever this profile is less than another real profile.
      * @param other
      */
-    public lessThan(other: Real | number): Windows
+    public lessThan(other: Real | number): Windows;
 
     /**
      * Produce a window whenever this profile is less than or equal to another real profile.
      * @param other
      */
-    public lessThanOrEqual(other: Real | number): Windows
+    public lessThanOrEqual(other: Real | number): Windows;
 
     /**
      * Produce a window whenever this profile is greater than to another real profile.
      * @param other
      */
-    public greaterThan(other: Real | number): Windows
+    public greaterThan(other: Real | number): Windows;
 
     /**
      * Produce a window whenever this profile is greater than or equal to another real profile.
      * @param other
      */
-    public greaterThanOrEqual(other: Real | number): Windows
+    public greaterThanOrEqual(other: Real | number): Windows;
 
     /**
      * Produce a window whenever this profile is equal to another real profile.
      * @param other
      */
-    public equal(other: Real | number): Windows
+    public equal(other: Real | number): Windows;
 
     /**
      * Produce a window whenever this profile is not equal to another real profile.
      * @param other
      */
-    public notEqual(other: Real | number): Windows
+    public notEqual(other: Real | number): Windows;
 
     /**
      * Produce an instantaneous window whenever this profile changes.
      */
-    public changed(): Windows
+    public changed(): Windows;
   }
 
   export class Discrete<Schema> {
     /** Internal AST Node */
-    public readonly __astNode: AST.DiscreteProfileExpression
+    public readonly __astNode: AST.DiscreteProfileExpression;
 
     /**
      * Internal instance of the Schema type, for type checking.
@@ -421,14 +426,14 @@ declare global {
      * @param name
      * @constructor
      */
-    public static Resource<R extends Gen.ResourceName>(name: R): Discrete<Gen.DiscreteResourceSchema<R>>
+    public static Resource<R extends Gen.ResourceName>(name: R): Discrete<Gen.DiscreteResourceSchema<R>>;
 
     /**
      * Create a constant discrete profile for all time.
      * @param value
      * @constructor
      */
-    public static Value<Schema>(value: Schema): Discrete<Schema>
+    public static Value<Schema>(value: Schema): Discrete<Schema>;
 
     /**
      * Produce an instantaneous window whenever this profile makes a specific transition.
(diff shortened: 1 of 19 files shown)


merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -1,17 +1,82 @@
...
+            import { Discrete, Real, Windows } from './constraints-edsl-fluent-api.js';
+            export type ActivityTypeName =
+              | "activity"
+              ;
+            export type ResourceName = "mode" | "state of charge";
+            export type DiscreteResourceSchema<R extends ResourceName> =
+              R extends "mode" ? ( | "Option1" | "Option2") :
+              R extends "state of charge" ? number :
+              void;
+            export function discreteResourceSchemaDummyValue<R extends ResourceName>(resource: R): DiscreteResourceSchema<R> {
```
dyst5422: I'm confused why this is here. I know its used in the constructor for Discrete, but I'm struggling to understand the purpose there too. It seems unused.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -1,17 +1,82 @@
...
+              return (
+                resource === "mode" ? "Option1" :
+                resource === "state of charge" ? 1.0 :
+                undefined
+              ) as DiscreteResourceSchema<R>;
+            }
+            export type RealResourceName = "state of charge";
+            export type ActivityParameters<A extends ActivityTypeName> =
+              A extends "activity" ? {Param: Discrete<string>, AnotherParam: Real, } :
+              void;
```
dyst5422: I don't think `void` should be used here, but rather `undefined`. https://stackoverflow.com/a/58885486

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -1,17 +1,82 @@
...
+            import * as AST from './constraints-ast.js';
+            import { Discrete, Real, Windows } from './constraints-edsl-fluent-api.js';
+            export type ActivityTypeName =
+              | "activity"
+              ;
+            export type ResourceName = "mode" | "state of charge";
+            export type DiscreteResourceSchema<R extends ResourceName> =
+              R extends "mode" ? ( | "Option1" | "Option2") :
+              R extends "state of charge" ? number :
+              void;
```
dyst5422: I don't think void should be used here, but rather undefined. https://stackoverflow.com/a/58885486

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -8,20 +9,24 @@ export class Constraint {
...
+    let expressionString = expression.toString();
+    let startIndex = expressionString.indexOf("(") + 1;
+    let endIndex = expressionString.indexOf(")");
+    let alias = expressionString.substring(startIndex, endIndex);
     return new Constraint({
       kind: AST.NodeKind.ForEachActivity,
       activityType,
       alias,
-      expression: constraint.__astNode
+      expression: expression(new Gen.ActivityInstance(activityType, alias)).__astNode
```
dyst5422: What does a call of this (Constraint.ForEachActivity) look like?

PS There are vulnerabilities in this expressionString implementation. Example: its perfectly valid to wrap a callback function in parens, so for example

`[1,2,3].map(((((num) => num + 1))))` would typecheck fine.

Or you can omit the parens for an arrow function

`[1,2,3].map(num => num + 1)` would typecheck fine.

dyst5422 (dismissed):

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: I thought `void` would act like a `never` type. Turns out there's an explicit `never` type. I'll use that instead. Thanks!

dyst5422: We should probably throw at the end of the implementations then instead of returning undefined.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
dyst5422: 🤦 Didn't see the example in the descriptions. I don't think we need to do the string parsing of the function - we can just autogenerate an alias.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: You highlighted multiple lines, I assume you are talking about `discreteResourceSchemaDummyValue`? That is so in the `Discrete.Resource` function it can get an instance of the discrete schema to assign to `this.__schemaInstance`. The `DummyValue` function has to be generated code because the dummy value's type depends on the mission model's resource types.

It is only for type checking, and the value is technically unused. But if it isn't stored in `this.__schemaInstance`, then `Discrete<number>` can be assigned to `Discrete<string>` or any other `Discrete` type erroneously.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Is there any possibility that the Java constraint evaluation could throw an error that needs to include the alias name? If so then the user wouldn't know which alias it refers to. If not then I'm down to autogenerate.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
dyst5422: Ah, ok. Hmm... I'm not seeing that behavior. For example, if I remove the dummy value on Discrete, then this results in a type check error
```
Discrete.Value(0).equal(Discrete.Value('d'));
```

Do you have a minimal code example of what you are seeing w.r.t. the need for this?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
dyst5422: Hmmm... interesting thought. I would think we would want to have a more contextualized error w.r.t. the plan in particular - indicating an activity by id directly.

JoelCourtney: I've made that change as an extra commit that I'll either squash or revert later.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: IIRC this one failed to produce an error when I removed the field:
```typescript
export default () => {
  return Constraint.ForEachActivity(
    "activity",
    (alias) => Discrete.Resource("mode").equal(alias.parameters().Param)
  )
}
```

This is using the `StubMissionModel` mock class, where `mode` is either `"Option1"` or `"Option2"`, and the `Param` parameter of `activity` is any `string`.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: Done. I added a `throw` if it comes out undefined, but I can't write a test for it because ideally that's unreachable.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: done

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: That's the very last test case in `ConstraintsDSLCompilationServiceTests`. If you remove the field it will fail.

dyst5422:
Can we make the ActivityType an enum instead of a union of strings. I think the autocompletion of

```
Constraint.ForEachActivity(
  ActivityTypes.activity,
  (alias) => Discrete.Resource("mode").equal(alias.parameters().Param)
)
```

is better than
```
Constraint.ForEachActivity(
  'activity',
  (alias) => Discrete.Resource("mode").equal(alias.parameters().Param)
)
```


@mattdailis you did this in your revision of the windows expressions. Thoughts?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
```diff
@@ -33,4 +171,158 @@ private static String joinLines(final Iterable<String> result) {
...
+  private static String valueSchemaToTypescriptProfile(final ValueSchema valueSchema) {
+    var basicTypescript = valueSchemaToTypescript(valueSchema);
+    if (basicTypescript.equals("number")) {
+      return "Real";
+    } else {
+      return "Discrete<" + basicTypescript + ">";
+    }
+  }
+
+  private static String valueSchemaToTypescriptDefault(final ValueSchema valueSchema) {
```
pcrosemurgy: Are these default literals intended to be replaced by the user within the autogenerated code? Only asking because I find the defaults to be a little unexpected. For example:
- `"string"` instead of `""` as the string default, and
- `/path` instead of something like `/dev/null` as the path default.

JoelCourtney: No, the values of the defaults don't matter at all. They are never used or changed, and are discarded at the end :). In any other situation these values would be dead weight, but the defaults are assigned to the `ActivityInstance.__schemaInstance` field, because the data needs to exist in order to be typechecked.

JoelCourtney:
> Can we make the ActivityType an enum instead of a union of strings.

I like that too, my only concern is that the activity type comes from the `@ActivityType("...")` annotation (*not* the activity's Java class name), which can use any string. (such as a string with spaces.) I'm not totally against trying to sanitize the names, but that just seems like an extra place for things to fail.

mattdailis:
> > Can we make the ActivityType an enum instead of a union of strings.
> 
> I like that too, my only concern is that the activity type comes from the `@ActivityType("...")` annotation (_not_ the activity's Java class name), which can use any string. (such as a string with spaces.) I'm not totally against trying to sanitize the names, but that just seems like an extra place for things to fail.

We ran across that Resources as well, on the scheduler. In practice, (read "Clipper"), the ActivityType always matches the java class name, but you're right, it doesn't _have_ to.

Resources, on the other hand, tend to be named as paths `/like/this`, so they are definitely not valid identifiers.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
dyst5422: I just removed the dummy value, and all tests pass.

dyst5422:
Lets go ahead and make ActivityInstance#parameters be a getter so its used as 

`instance.parameters.Param` instead of `instance.parameters().Param`

JoelCourtney:
> Lets go ahead and make ActivityInstance#parameters be a getter

I don't understand, it already is.

dyst5422:
> I like that too, my only concern is that the activity type comes from the `@ActivityType("...")` annotation (_not_ the activity's Java class name), which can use any string. (such as a string with spaces.) I'm not totally against trying to sanitize the names, but that just seems like an extra place for things to fail.

Yeah. I think it's worth it for activity types. I think its not necessary to include the sanitation in this PR since our only customer is already doing so fine.

dyst5422:
> > Lets go ahead and make ActivityInstance#parameters be a getter
> 
> I don't understand, it already is.

`public get parameters() {`

makes it so you don't have to use it as a function call

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
pcrosemurgy: Ohh interesting. I trust you know what you're doing here but it seems odd to have to generate literals to run type checking. In my mind type semantics exist for this purpose in the first place. I'm probably just confusing myself though, if you have time for a short walkthrough of this PR I'd be interested in joining!

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -219,34 +198,29 @@ export class Real {
   }
 }
 
-export class Discrete {
+export class Discrete<Schema> {
   public readonly __astNode: AST.DiscreteProfileExpression
+  private readonly __schemaInstance: Schema;
```
mattdailis: If we never query it, I suspect we might not need to store it 🤔

merlin-server/constraints-dsl-compiler/src/main.ts
```diff
@@ -43,16 +48,12 @@ async function handleRequest(data: Buffer) {
...
-  if (data.toString().trim() === 'ping') {
-    // Enable testing the health of the service by sending 'ping' and expecting 'pong' in return.
-    process.stdout.write('pong\n');
-    process.stdin.once('data', handleRequest);
-  }
+let lineReader = readline.createInterface({
+  input: process.stdin
 });
 
+lineReader.on('line', handleRequest);
```
mattdailis: The reason we used `once` on the scheduler side, was to make sure that the node service is "listening" for requests only when it has completed the previous request. If we get multiple requests in rapid succession, it _might_ be possible that we'd respond to them out of order...? @dyst5422 is that accurate?

mattdailis, line 59: The reason we used `once` on the scheduler side, was to make sure that the node service is "listening" for requests only when it has completed the previous request. If we get multiple requests in rapid succession, it _might_ be possible that we'd respond to them out of order...? @dyst5422 is that accurate?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
```diff
@@ -50,6 +51,8 @@ public final class StubMissionModelService implements MissionModelService {
       NONEXISTENT_ACTIVITY_TYPE,
       Map.of());
 
+  public static final Map<String, ValueSchema> RESOURCES;
```
mattdailis, line 54: It would be more idiomatic to use the `Map.of` factory method here.

```suggestion
  public static final Map<String, ValueSchema> RESOURCES = Map.of(
    "mode", ValueSchema.ofVariant(List.of(
        new ValueSchema.Variant("Option1", "Option1"),
        new ValueSchema.Variant("Option2", "Option2")
    )),
    "state of charge", ValueSchema.REAL
  );
```

mattdailis, line 54: It would be more idiomatic to use the `Map.of` factory method here.

```suggestion
  public static final Map<String, ValueSchema> RESOURCES = Map.of(
    "mode", ValueSchema.ofVariant(List.of(
        new ValueSchema.Variant("Option1", "Option1"),
        new ValueSchema.Variant("Option2", "Option2")
    )),
    "state of charge", ValueSchema.REAL
  );
```

mattdailis (commented):
Haven't quite finished going through all the code, but I'll leave what comments I have so far now.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
dyst5422: Yeah. I think as long as it has a type parameter, we don't need to actually supply a value and we get all the type checking benefits.

dyst5422:
Before we merge, we should get a final code auto-formatting commit in of the TS files.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: You might have not re-run the tests because they were cached. I just tried again. If you remove the field it fails, even if you keep the extra constructor argument.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Dylan, you taught me this with the whole `__astNode` problem. The type parameter doesn't exist in the emitted Javascript, and the only way the type can be checked is if the data exists. What you've just described is a nominal type system. Matt, if you remove that field and re-run the tests (make sure to invalidate the cache by making a small change to the tests file), the last test will fail.

JoelCourtney: Ok, good news, most of this was unnecessary. Apparently you can remove everything related to `__schemaInstance` and default dummy values *except* for the declared `__schemaInstance` field in `declare global`, and all tests will pass, even though nothing is ever assigned to that field. I still don't really understand why, but I'm definitely not mad that it works :). Thanks @dyst5422 for the help!

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
JoelCourtney: resolving, see [this comment](https://github.com/NASA-AMMOS/aerie/pull/183#discussion_r882873669)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
JoelCourtney: There's definitely something funky going on here that I don't understand, but see [this comment](https://github.com/NASA-AMMOS/aerie/pull/183#discussion_r882873669) for the current situation.

mattdailis:
> change instance.during() to be a more accurate name now that it is a method instead of function. (something like instance.window() or instance.active())

Perhaps `inProgress()`? (Just brainstorming)

JoelCourtney:
> Perhaps `inProgress()`? (Just brainstorming)

Anything that makes this example read nicely would be great:

```typescript
    return Constraint.ForEachActivity(
        ActivityType.BiteBanana,
        instance => instance.during().not()
    )
}
```

(this forbids usage of BiteBanana)

merlin-server/constraints-dsl-compiler/src/main.ts
JoelCourtney: `once` is back now

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
JoelCourtney: done

JoelCourtney: done

mattdailis (approved):

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
JoelCourtney: Actually, after implementing the change, the code gen test fails 50% of the time. Apparently `Map.of` does not maintain a deterministic order. HashMap does (it doesn't keep insertion order, it keeps hash order), so it worked by coincidence. reverting...

mattdailis: Ah, dang - why does the order matter? Is it because we're doing string matching on generated code?

JoelCourtney:
Changes since Matt's approving review:

- Install prettier in package.json, create `reformat` script.
- Run `reformat`
- Change `instance.during()` to `instance.window()`. I'm trying to avoid using Boolean-suggestive words, which is why I didn't go with `inProgress` or `active`
    - On that note, `window.not()` is probably due to be reworded as well, but I'll leave that for another time. As part of the PR for removing the old constraints pipeline, I'm planning to rename the Java expression tree nodes to match the API (like `Windows.All`, `Windows.Any`). That would be a good time.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
JoelCourtney: exactly. sometimes the resources are inserted into the typescript code in a different order, which makes the CodeGen test complain, because it's just checking for exact string equality.

JoelCourtney: As part of reverting, I changed it to a `LinkedHashMap`, which preserves insertion order.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
```diff
@@ -16,10 +16,15 @@ public String generateTypescriptTypesFromMissionModel(String missionModelId)
   {
     final var activityTypes = this.missionModelService.getActivityTypes(missionModelId);
 
-
-
     final var result = new ArrayList<String>();
     result.add("/** Start Codegen */");
+
```
pranav-super, line 27: Before this change, was this method just a placeholder?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -220,34 +220,36 @@ export class Real {
   }
 }
 
-export class Discrete {
+export class Discrete<Schema> {
```
pranav-super, line 200: What does a Schema achieve here?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -17,12 +19,14 @@ export class Constraint {
     })
   }
 
-  public static ForEachActivity(activityType: Gen.ActivityTypeName, alias: string, constraint: Constraint): Constraint {
+  public static ForEachActivity<A extends Gen.ActivityTypeName>(activityType: A, expression: (instance: Gen.ActivityInstance<A>) => Constraint): Constraint {
```
pranav-super: What does this new interface achieve over the old one? Why are local ActivityInstances useful?

pranav-super (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -161,6 +161,8 @@ private void getSimulationResults(final Context ctx) {
       ctx.status(400).result(ResponseSerializers.serializeInvalidJsonException(ex).toString());
     } catch (final NoSuchPlanException ex) {
       ctx.status(404).result(ResponseSerializers.serializeNoSuchPlanException(ex).toString());
+    } catch (final MissionModelService.NoSuchMissionModelException ex) {
+      ctx.status(404).result(ResponseSerializers.serializeNoSuchMissionModelException(ex).toString());
```
pcrosemurgy, line 165: Thanks for adding this! Since this PR introduces `serializeNoSuchMissionModelException()` would you mind tacking this on to every `ctx.status(404)` related to a `MissionModelService.NoSuchMissionModelException` in this class?

This should keep things nice & consistent.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
```diff
@@ -1,27 +1,160 @@
...
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
-import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 
-import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 
 public class TypescriptCodeGenerationService {
```
pcrosemurgy: A couple nit-picky changes I'd make:
- Making this class `final`
- `public TypescriptCodeGenerationService(MissionModelService missionModelService)` -> `public TypescriptCodeGenerationService(final MissionModelService missionModelService)`
- `public String generateTypescriptTypesFromMissionModel(String missionModelId)` -> `public String generateTypescriptTypesFromMissionModel(final String missionModelId)`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
pcrosemurgy: I think so :)

Before this change the method was:
```java
  public static String generateTypescriptTypesFromMissionModel() {
    final var result = new ArrayList<String>();
    result.add("/** Start Codegen */");
    result.add("/** End Codegen */");
    return joinLines(result);
  }

File: merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts (hunk 10 above)
Comment (dyst5422) [reply merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts:200]: The type parameter here constrains an instance of Discrete to a specific type - ie `Discrete<number>` not assignable to `Discrete<string>`

File: merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts (hunk 11 above)
Comment (dyst5422) [reply merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts]: No management of aliases, which were very difficult to type check. Plus I think its more intuitive when making reference to activity properties

**Comment by dyst5422** [comment]
> Changes since Matt's approving review:
> 
> * Install prettier in package.json, create `reformat` script.
> * Run `reformat`
> * Change `instance.during()` to `instance.window()`. I'm trying to avoid using Boolean-suggestive words, which is why I didn't go with `inProgress` or `active`
>   
>   * On that note, `window.not()` is probably due to be reworded as well, but I'll leave that for another time. As part of the PR for removing the old constraints pipeline, I'm planning to rename the Java expression tree nodes to match the API (like `Windows.All`, `Windows.Any`). That would be a good time.

Perhaps `invert()` in place of `not()`? Since what we're doing is essentially inverting the window from `--------[----]--------` to `[--------]----[--------]`

File: merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts (hunk 10 above)
Comment (JoelCourtney) [reply merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts:200]: Yes, that is correct. This way `Discrete.Value(5).equal("not a number")` is not a valid comparison.

File: merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java (hunk 9 above)
Comment (JoelCourtney) [reply merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java:27]: Yup, I copied a lot of boilerplate from the scheduling eDSL compiler, including this, in a previous PR. This PR fills in the code gen.

File: merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts (hunk 11 above)
Comment (JoelCourtney) [reply merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts]: The previous approaches had the user refer to activity instance aliases by a string name. For example, to forbid usage of an activity, you'd do this:

```typescript
export default(): Constraint => Constraint.ForEachActivity(
    "MyActivity",
    "activity instance alias",
    Windows.During("activity instance alias").not()
)
```

If you happened to mis-type the alias (i.e. `Windows.During("some other alias")`), there would be no immediate error and it would fail in the backend. After all of the changes in this PR, you would now write it as this:

```typescript
export default(): Constraint => Constraint.ForEachActivity(
    ActivityType.MyActivity,
    instance => instance.window().not()
)
```

This hands the problem of validating activity alias names to the typescript compiler. It also makes it easier to know the schema of the activity instance parameters. E.g. if the activity has a `someNumber` parameter, we can now access it as `instance.parameters.someNumber`, which will carry enough info about the parameter's type to do compile-time validation.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
```diff
@@ -33,4 +166,98 @@ private static String joinLines(final Iterable<String> result) {
...
+    }
+  }
+
+  private static boolean valueSchemaIsReal(final ValueSchema valueSchema) {
+    return valueSchema.match(new ValueSchema.DefaultVisitor<>() {
+      @Override
+      protected Boolean onDefault() {
+        return false;
+      }
+
```
mattdailis, line 256: Any objections to making this `valueSchemaIsNumeric`, and adding an `onInt() { return true; }` case? I can't think of a reason that would bite us - it should be okay to use integers for Real profiles 🤔 

Context: for the scheduler tests, I've been using the `/plant` resource for all of my numeric comparisons. I ran into this error:
```
Argument of type '"/plant"' is not assignable to parameter of type 'RealResourceName'
```

EDIT: Sorry, could've picked a better line of code - the method is a couple lines up, called `valueSchemaIsReal`.

JoelCourtney: Yeah, that'll totally work. Jonathan already thought of this, and the backend seems already equipped to handle discrete ints as reals. one sec

JoelCourtney: done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
JoelCourtney: done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
JoelCourtney: done

pcrosemurgy (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationService.java
mattdailis: I have one more case I'd like to add to `valueSchemaIsNumeric` (I don't need to hold up this PR, I can add it when I'm rebasing as well 👍)

I finally understood my confusion about `Real` vs `Linear` - `Linear` resources are serialized as `{initial: number, rate: number}`, and the constraints backend will handle all the interpolation. On the frontend, I think we can treat them as numeric. Here's my proposal for how `valueSchemaIsNumeric` would look:

```java
private static boolean valueSchemaIsNumeric(final ValueSchema valueSchema) {
    return valueSchema.equals(ValueSchema.INT)
        || valueSchema.equals(ValueSchema.REAL)
        || valueSchema.equals(LINEAR_RESOURCE_SCHEMA);
  }
```

where `LINEAR_RESOURCE_SCHEMA` is defined as so:

```java
private static final ValueSchema LINEAR_RESOURCE_SCHEMA = ValueSchema.ofStruct(Map.of(
      "initial", ValueSchema.REAL,
      "rate", ValueSchema.REAL));
```

I've changed `/fruit` to be a linear resource in my branch and added a test with this scheduling goal to the scheduler integration tests:
```typescript
export default (): Goal => {
 return Goal.CoexistenceGoal({
   activityTemplate: ActivityTemplates.PeelBanana({peelDirection: "fromStem"}),
   forEach: Windows.All(
     Real.Resource("/fruit").lessThan(4.0),
     Real.Resource("/fruit").greaterThan(2.0)
   )
 })
```

PR #184, opened 2022-05-26
[AERIE-1862] Change nvm install location to /usr/src

mattdailis:
* **Tickets addressed:** AERIE-1862
* **Review:** By file  

## Description

In the `scheduler-server` and `merlin-server` Dockerfiles, the `NVM_DIR` variable was set to `/root/.nvm`, which mean that `nvm` and `node` would be installed in subfolders of the `/root` folder. The `/root` folder is locked down so that non-root users can't access its contents. This prevents our containers from running with non-root users.

This PR changes the `NVM_DIR` to `/usr/src/.nvm` in both the `scheduler-server` and the `merlin-server`.

The original choice of `/root/.nvm` came from this stack overflow answer: https://stackoverflow.com/a/57546198

## Verification

I'm hoping that the tests run by Github actions on this PR are sufficient to verify this hasn't broken anything. I've also verified that the `SchedulingDslCompiler` doesn't crash after setting a non-root user via the docker-compose file like so:

```
user: "12345:10100"
```

## Documentation

If we decide to support non-root users as a use case, we could mention it in our deployment docs.

## Future work

I'm not sure if we intend to support "non-root users" as a feature - if we do, it may be worth thinking about how we avoid regressions. We could set up our docker compose file to run all of our services as non-root users 🤷

### DIFF
--- a/merlin-server/Dockerfile
+++ b/merlin-server/Dockerfile
@@ -7,9 +7,10 @@ RUN cd /usr/src/app && tar --strip-components 1 -xf server.tar -C extracted
 FROM eclipse-temurin:18-focal
 
 ENV NODE_VERSION=16.14.0
-ENV NVM_DIR=/root/.nvm
-ENV PATH="/root/.nvm/versions/node/v${NODE_VERSION}/bin/:${PATH}"
+ENV NVM_DIR=/usr/src/.nvm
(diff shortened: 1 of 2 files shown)


JoelCourtney (approved):

goetzrrGit (approved):
Thumbs up!

camargo (approved):

PR #185, opened 2022-05-26
[AERIE-1847]  Add duration filter to constraints eDSL

adrienmaillard:
* **Tickets addressed:** AERIE-1847
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This is building on top of #179.

I have reimplemented the filters in the constraints library. 

## Verification

Several tests have been written/modified. 

## Documentation

Documentation about these two constraints has been added to the [Constraints](https://github.com/NASA-AMMOS/aerie/wiki/Constraints) page.

## Future work

All of the other filters/transformers.

PR #186, opened 2022-05-26
[AERIE-1848] Add expand/clamp transformers to constraints eDSL

adrienmaillard:
* **Tickets addressed:** AERIE-1848
* **Review:** By commit  

## Description

Sibling PR to #185.
I have reimplemented the transformers in the constraints library.

## Verification

New tests.

## Documentation

Documentation about these operators has been added to the constraints page. 

## Future work

Other filters and transformers.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -10,6 +10,7 @@ import gov.nasa.jpl.aerie.json.BasicParsers;
 import gov.nasa.jpl.aerie.json.Iso;
 import gov.nasa.jpl.aerie.json.JsonParser;
 import gov.nasa.jpl.aerie.json.Unit;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 
 import java.util.List;
 
@@ -18,6 +19,7 @@ import static gov.nasa.jpl.aerie.json.BasicParsers.chooseP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.doubleP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.listP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.literalP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.longP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.productP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.recursiveP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
@@ -152,6 +154,22 @@ public final class ConstraintParsers {
               untuple((kind, alias) -> new StartOf(alias)),
               $ -> tuple(Unit.UNIT, $.activityAlias)));
 
+  static final JsonParser<Duration> durationP =
+      longP
+          . map(Iso.of(
+              microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
+              duration -> duration.in(Duration.MICROSECONDS)));
+
+  static JsonParser<ShiftBy> shiftByF(JsonParser<Expression<Windows>> windowsExpressionP) {
+    return productP
+        .field("kind", literalP("WindowsExpressionShiftBy"))
+        .field("windowExpression", windowsExpressionP)
+        .field("fromStart", durationP)
+        .field("fromEnd", durationP)
+        .map(Iso.of(
+            untuple((kind, windowsExpression, fromStart, fromEnd) -> new ShiftBy(windowsExpression, fromStart, fromEnd)),
+            $ -> tuple(Unit.UNIT, $.windows, $.fromStart, $.fromEnd)));
+  }
   static final JsonParser<EndOf> endOfP =
       productP
           .field("kind", literalP("WindowsExpressionEndOf"))
@@ -279,7 +297,8 @@ public final class ConstraintParsers {
           notEqualF(discreteProfileExprP),
           allF(selfP),
           anyF(selfP),
-          invertF(selfP)));
+          invertF(selfP),
+          shiftByF(selfP)));
 
   static final JsonParser<ViolationsOf> violationsOfP =
       productP
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ShiftBy.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ShiftBy.java
@@ -1,0 +1,60 @@
+package gov.nasa.jpl.aerie.constraints.tree;
+
+import gov.nasa.jpl.aerie.constraints.model.ActivityInstance;
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Window;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public final class ShiftBy implements Expression<Windows> {
+  public final Expression<Windows> windows;
+  public final Duration fromStart;
+  public final Duration fromEnd;
+
+  public ShiftBy(final Expression<Windows> left, final Duration fromStart, final Duration fromEnd) {
+    this.windows = left;
+    this.fromStart = fromStart;
+    this.fromEnd = fromEnd;
+  }
+
+  @Override
+  public Windows evaluate(final SimulationResults results, final Window bounds, final Map<String, ActivityInstance> environment) {
+    final var windows = this.windows.evaluate(results, bounds, environment);
+    return windows.shiftBy(this.fromStart, this.fromEnd);
+  }
+
+  @Override
+  public void extractResources(final Set<String> names) {
+    this.windows.extractResources(names);
+  }
+
+  @Override
+  public String prettyPrint(final String prefix) {
+    return String.format(
+        "\n%s(shift %s by %s %s)",
+        prefix,
+        this.windows.prettyPrint(prefix + "  "),
+        this.fromStart.toString(),
+        this.fromEnd.toString()
+    );
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof ShiftBy)) return false;
+    final var o = (ShiftBy)obj;
+
+    return Objects.equals(this.windows, o.windows) &&
+           Objects.equals(this.fromStart, o.fromStart) &&
+           Objects.equals(this.fromEnd, o.fromEnd);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(this.windows, this.fromStart, this.fromEnd);
+  }
+}
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
@@ -12,6 +12,7 @@ export enum NodeKind {
   WindowsExpressionActivityWindow = 'WindowsExpressionActivityWindow',
   WindowsExpressionStartOf = 'WindowsExpressionStartOf',
   WindowsExpressionEndOf = 'WindowsExpressionEndOf',
+  WindowsExpressionShiftBy = 'WindowsExpressionShiftBy',
   ExpressionEqual = 'ExpressionEqual',
   ExpressionNotEqual = 'ExpressionNotEqual',
   RealProfileLessThan = 'RealProfileLessThan',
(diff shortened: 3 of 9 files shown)


JoelCourtney:
I have a concern that "clamp" is suggestive of a different operation, and that what you've implemented here is actually "shrink". Usually "clamp" means ensuring an input lies in a range. If the input is outside of the range, the closest edge of the range is returned. If the input is inside the range, it is returned unchanged. I think it could cause confusion if we call this operation "clamp" instead of "shrink"

Twisol:
> Usually "clamp" means ensuring an input lies in a range. If the input is outside of the range, the closest edge of the range is returned. If the input is inside the range, it is returned unchanged.

For instance, [the `clamp` function](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/clamp.xhtml) provided by OpenGL, which is where I learned the terminology.

Twisol:
It seems like `contractBy` and `expandBy` do the same operation, i.e. offsetting the start and end of a window by some amount. They only appear to differ in the assumed sign of the applied offsets. Would it be desirable and/or possible to keep the AST limited to the one fundamental operation (of applying offsets to endpoints)?

adrienmaillard:
@Twisol we could do that. How would call this operation ? `OffsetEndpointsBy` ?

mattdailis:
> @Twisol we could do that. How would call this operation ? `OffsetEndpointsBy` ?

I'm okay with that name, unless someone suggests a better one :+1:

It occurs to me that this can be used to implement the window "shift" operator as well.

JoelCourtney:
I'll vote for just calling it `windows.shift(leftBoundShift, rightBoundShift)`.

Twisol:
I can't seem to come up with a better name for that particular operation. We're already not far off a full convolution operation between Windows though -- maybe we just implement that, call it "convolveWith", and implement all the other common forms in terms of it once and for all?

My internet is out (so I'm on my phone), but let me try to write some pseudocode for how "convolveWith" works on Windows. (This isn't the fully abstract definition of convolution, and it's been months since I worked out the correspondence, but the abstract definition specializes to this one on functions R -> {0, 1}.)

```python
def convolve(windows1, windows2):
  convolution = Windows()
  for x in windows1:
    for y in windows2:
      convolution.add(Window(x.start + y.start, x.end + y.end))
  return convolution
```

Then we can implement  `w.offsetEndpointsBy(left, right)` as `convolve(w, Windows(Window(left, right)))`.

Twisol:
I just worked out the correspondence with convolution for myself, and I think it doesn't admit shrinking windows. So you can safely ignore me --  I'm fine with Joel's name.

adrienmaillard:
`Shift` would be the name of the constraint tree node and `shift` the name of the method in windows. Are we in agreement ? If yes, I am also going to mirror this in the constraint edsl (and replace both shrink/expand nodes by just a shift node).

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ShiftBy.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public void extractResources(final Set<String> names) {
+    this.windows.extractResources(names);
+  }
+
+  @Override
+  public String prettyPrint(final String prefix) {
+    return String.format(
+        "\n%s(clamp %s by %s %s)",
```
patkenneally: :small_red_triangle_down: Should this be "shiftBy" now that we've changed the method name?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ShiftBy.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public final class ShiftBy implements Expression<Windows> {
+  public final Expression<Windows> windows;
+  public final Duration fromStart;
+  public final Duration fromEnd;
+
+  public ShiftBy(final Expression<Windows> left, final Duration clampFromStart, final Duration clampFromEnd) {
```
patkenneally: :small_red_triangle_down: Should these parameters be renamed to `shiftFromStart` and `shiftFromEnd` ?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -152,6 +154,22 @@ static JsonParser<Rate> rateF(final JsonParser<Expression<LinearProfile>> linear
               untuple((kind, alias) -> new StartOf(alias)),
               $ -> tuple(Unit.UNIT, $.activityAlias)));
 
+  private static final JsonParser<Duration> durationP =
```
patkenneally: :large_orange_diamond: I think private is not needed here?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -152,6 +154,22 @@ static JsonParser<Rate> rateF(final JsonParser<Expression<LinearProfile>> linear
...
+      longP
+          . map(Iso.of(
+              microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
+              duration -> duration.in(Duration.MICROSECONDS)));
+
+  static JsonParser<ShiftBy> shiftByP(JsonParser<Expression<Windows>> windowsExpressionP) {
```
patkenneally: :large_orange_diamond: I think this method should be named `shiftByF` in keeping with the naming convention.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -68,6 +70,13 @@ export interface WindowsExpressionInvert {
   expression: WindowsExpression;
 }
 
+export interface WindowsExpressionShiftBy {
+  kind: NodeKind.WindowsExpressionShiftBy,
+  windowExpression: WindowsExpression,
+  fromStart: number,
```
patkenneally: :large_orange_diamond: There is a `type Duration = number` on line 482. Should we be using that type here?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -277,6 +278,31 @@ export default() => {
     );
   }
 
+  @Test
+  void testClampBy() {
```
patkenneally: :large_orange_diamond: Should this be `testShiftBy` ?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -277,6 +278,31 @@ export default() => {
...
+    checkSuccessfulCompilation(
+        """
+            export default() => {
+              return Real.Resource("state of charge").rate().equal(Real.Value(4.0)).shiftBy(1000, -200)
+            }
+        """,
+        new ViolationsOf(new ShiftBy(new Equal<>(new Rate(new RealResource("state of charge")), new RealValue(4.0)), Duration.of(1000, Duration.MICROSECOND), Duration.of(-200, Duration.MICROSECOND)))
```
patkenneally: :large_orange_diamond: This is a loooooong line, I suggest we add some breaks/formatting for easier parsing.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -277,6 +278,31 @@ export default() => {
...
+
+  @Test
+  void testExpandBy() {
+    checkSuccessfulCompilation(
+        """
+            export default() => {
+              return Real.Resource("state of charge").rate().equal(Real.Value(4.0)).shiftBy(-1000, 200)
+            }
+        """,
+        new ViolationsOf(new ShiftBy(new Equal<>(new Rate(new RealResource("state of charge")), new RealValue(4.0)), Duration.of(-1000, Duration.MICROSECOND), Duration.of(200, Duration.MICROSECOND)))
```
patkenneally: 🔶 Same here on formatting.

patkenneally (changes requested):
Nicely done.

Legend:

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
JoelCourtney: I general I prefer to restrict visibility as much as possible, much like `const`/`final`-correctness. If it can be private, I think it should be.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: Seconded: I added that line, it is currently not used. I only added it because that same line existed in `scheduling-ast.ts` too as a placeholder. When we replace `Duration` with a more appropriate type, we'll have to change this to `Duration` anyway.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -83,6 +83,15 @@ export class Windows {
       expression: this.__astNode,
     });
   }
+
+  public shiftBy(fromStart: number, fromEnd: number) : Windows {
```
JoelCourtney: Same comment as Pat's about `number` vs `Duration`.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -277,6 +278,31 @@ export default() => {
     );
   }
 
+  @Test
+  void testClampBy() {
+    checkSuccessfulCompilation(
+        """
+            export default() => {
+              return Real.Resource("state of charge").rate().equal(Real.Value(4.0)).shiftBy(1000, -200)
```
JoelCourtney, line 286: (not blocker) Fun fact, you can omit `Real.Value` in the right-hand-side of all operations in the API :) So `.equal(4)` will work too.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -277,6 +278,31 @@ export default() => {
...
+            }
+        """,
+        new ViolationsOf(new ShiftBy(new Equal<>(new Rate(new RealResource("state of charge")), new RealValue(4.0)), Duration.of(1000, Duration.MICROSECOND), Duration.of(-200, Duration.MICROSECOND)))
+    );
+  }
+
+  @Test
+  void testExpandBy() {
```
JoelCourtney: This test seems redundant since you've merged them into one operation.

JoelCourtney (changes requested):
Love it! Only one real change request that Pat didn't mention, the rest are minor. (see comments) I'm having some trouble with github's comment notifications, so feel free to message me on slack when you're done and I'll get you a quick approval.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ShiftBy.java
adrienmaillard: I've set them to `fromStart` and `fromEnd`.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
adrienmaillard: I've renamed it `testShift`. And following Joel's comment, I've removed the other very similar test.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
adrienmaillard: Nice feature !

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
adrienmaillard: Removed it.

adrienmaillard:
I've integrated all of your changes. I've also updated the documentation on the Constraints page.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
adrienmaillard: It is in the constraints edsl, not the ast. But I have added it to the AST and used it in this node.

JoelCourtney (approved):

patkenneally (approved):

PR #187, opened 2022-05-26
[AERIE-2033] Replace long representation of duration with Temporal.Duration in scheduling eDSL

adrienmaillard:
* **Tickets addressed:** AERIE-2033
* **Review:** By commit  

## Description

- Copy the Temporal library from the expansion server. 
- Replace durations as number by `Temporal.Duration` in the scheduling eDSL

## Verification

Relevant tests have been modified. 

## Documentation

I will update the wiki pages.

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/polyfills.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/polyfills.ts
@@ -1,0 +1,3 @@
+import { Temporal } from '@js-temporal/polyfill';
+
+globalThis.Temporal = Temporal as typeof globalThis.Temporal;
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
@@ -1,3 +1,4 @@
+/// <reference path="./TemporalPolyfillTypes.ts"; />
 import type * as WindowsExpressions from "./constraints-ast.js";
 import "./constraints-edsl-fluent-api.js";
 
@@ -7,13 +8,13 @@ export interface ActivityTemplate {
 }
 
 export interface ClosedOpenInterval {
-  start: number
-  end: number
+  start: Temporal.Duration
+  end: Temporal.Duration
 }
 
 export type CardinalityGoalArguments =
-  |  { duration: number, occurrence: number }
-  |  { duration: number }
+  |  { duration: Temporal.Duration, occurrence: number }
+  |  { duration: Temporal.Duration }
   |  { occurrence: number }
 
 export enum NodeKind {
@@ -42,7 +43,7 @@ export type Goal =
 export interface ActivityRecurrenceGoal {
   kind: NodeKind.ActivityRecurrenceGoal,
   activityTemplate: ActivityTemplate,
-  interval: number,
+  interval: Temporal.Duration,
 }
 
 export interface ActivityCardinalityGoal {
@@ -82,7 +83,7 @@ export interface TimeExpressionRelativeFixed {
   fixed: boolean  // true means op(anchor, operand) is exactly the time, false means it's a range between op(anchor, operand) and anchor
   operation: {
     operator: TimeExpressionOperator,
-    operand: Duration
+    operand: Temporal.Duration
   }
 }
 
@@ -104,14 +105,14 @@ export enum TimingConstraintOperator {
 export interface ActivityTimingConstraintSingleton {
   windowProperty: WindowProperty;
   operator: TimingConstraintOperator;
-  operand: Duration
+  operand: Temporal.Duration
   singleton: true
 }
 
 export interface ActivityTimingConstraintRange {
   windowProperty: WindowProperty;
   operator: TimingConstraintOperator;
-  operand: Duration
+  operand: Temporal.Duration
   singleton: false
 }
 
--- a/scheduler-server/scheduling-dsl-compiler/src/main.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/main.ts
@@ -1,6 +1,8 @@
+import './libs/polyfills.js'
 import fs from 'fs';
 import ts from 'typescript';
-import {UserCodeRunner} from '@nasa-jpl/aerie-ts-user-code-runner';
+import { UserCodeRunner } from '@nasa-jpl/aerie-ts-user-code-runner';
+import vm from "node:vm";
 import * as readline from 'readline';
 
 const codeRunner = new UserCodeRunner();
@@ -19,6 +21,10 @@ const windowsEDSL = fs.readFileSync(
 const windowsAST = fs.readFileSync(
   `${process.env["SCHEDULING_DSL_COMPILER_ROOT"]}/src/libs/constraints/constraints-ast.ts`,
   "utf8"
+);
+const temporalPolyfillTypes = fs.readFileSync(
+    `${process.env["SCHEDULING_DSL_COMPILER_ROOT"]}/src/libs/TemporalPolyfillTypes.ts`,
+    'utf8',
 );
 const tsConfig = JSON.parse(fs.readFileSync(new URL('../tsconfig.json', import.meta.url).pathname, 'utf-8'));
 const { options } = ts.parseJsonConfigFileContent(tsConfig, ts.sys, '');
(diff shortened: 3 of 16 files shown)


adrienmaillard:
All tests are now working but (hey @dyst5422), this error pops up when trying to assemble:
```
/Users/maillard/Documents/Aerie/aerie-public/aerie/scheduler-server/scheduling-dsl-compiler/src/libs/polyfills.ts (5,16): Property 'toTemporalInstant' does not exist on type 'Date'.
src/libs/scheduler-ast.ts (10,10): Cannot find namespace 'Temporal'.
src/libs/scheduler-ast.ts (11,8): Cannot find namespace 'Temporal'.
src/libs/scheduler-ast.ts (15,18): Cannot find namespace 'Temporal'.
src/libs/scheduler-ast.ts (16,18): Cannot find namespace 'Temporal'.
src/libs/scheduler-ast.ts (44,13): Cannot find namespace 'Temporal'.
```

mattdailis:
Re-opening since I think we got pretty close with this PR, just need to push it over the finish line.

This may take some re-basing.

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -21,10 +22,18 @@ const windowsAST = fs.readFileSync(
   `${process.env["SCHEDULING_DSL_COMPILER_ROOT"]}/src/libs/constraints/constraints-ast.ts`,
   "utf8"
 );
+const temporalPolyfillTypes = fs.readFileSync(
+    `${process.env["SCHEDULING_DSL_COMPILER_ROOT"]}/src/libs/TemporalPolyfillTypes.ts`,
+    'utf8',
+);
+const polyfills = fs.readFileSync(
```
mattdailis: 🔶 This `polyfills` variable seems unused - is it necessary?

dyst5422: It should not be necessary if we are supplying the Temporal implementation via context as we do on line 89. The polyfills.ts looks to simply put Temporal on the globalThis and polyfill in Date.toTemporalInstant - which shouldn't be necessary as Date is generally not a great API - it was designed after the Java Date api that was subsequently deprecated and we shouldn't be passing any Date's to our stuff, so it shouldn't be necessary.

dyst5422: Where we DO need it is actually imported into our process (`import 'polyfills.ts';` as the first line) so that we can correctly serialize/deserialize the Temporal types for transfer to/from the Java process.

dyst5422: Command expansion has the correct (hopefully) working incantations for this, so you can reference what we have going on there https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/worker.ts

dyst5422: I think we'll likely need to borrow the Temporal serialization/deserialization code from the command expansion as well (or at least check that it is serializing with JSON.stringify() to a format we support) so that we can serialize everything correctly (something we DONT have to do on command expansion because we never have temporal objects left over once we have the seqJSON - which does the serialization within the toSeqJson call) https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/utils/temporalSerializers.ts

dyst5422: It looks like they serialize/deserialize into a format that isn't necessarily what we want on the Java side of things, but we could also just roll with it and write a parser on the Java side to handle that format.

scheduler-server/build.gradle
```diff
@@ -94,6 +94,7 @@ dependencies {
   implementation project(':parsing-utilities')
   implementation project(':constraints')
   implementation project(':scheduler')
+  implementation project(':contrib')
```
mattdailis: 🔻 My understanding is that `contrib` is to be depended on by mission models only - we should identify why this dependency is necessary and make it unnecessary. (This came to light especially with the recent SPICE adventures)

scheduler-server/scheduling-dsl-compiler/src/libs/polyfills.ts
```diff
@@ -0,0 +1,5 @@
+import { Temporal, toTemporalInstant } from '@js-temporal/polyfill';
+
+globalThis.Temporal = Temporal as typeof globalThis.Temporal;
+
+Date.prototype.toTemporalInstant = toTemporalInstant as typeof Date.prototype.toTemporalInstant;
```
mattdailis: 🔵 As Dylan pointed out, I think we can scrap the `toTemporalInstant` part - we don't need to interoperate with the JS Date type.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
```diff
@@ -82,7 +83,7 @@ export interface TimeExpressionRelativeFixed {
   fixed: boolean  // true means op(anchor, operand) is exactly the time, false means it's a range between op(anchor, operand) and anchor
   operation: {
     operator: TimeExpressionOperator,
-    operand: Duration
+    operand: Temporal.Duration
```
mattdailis, line 86: 🔵 I am wondering about the pros/cons of aliasing `Temporal.Duration` to just be `Duration` - on one hand, it'll be part of the language eventually, so this would be in line with using Typescript "directly". On the other hand, I don't _think we're planning to support other aspects of Temporal 🤔

scheduler-server/scheduling-dsl-compiler/src/main.ts
mattdailis: Thanks for weighing in, Dylan :+1: It looks like this PR changes the serialization format to be an ISO8601 string, and uses `java.time.Duration.parse` to parse it. I'm not sure how it _becomes_ an ISO8601 string, maybe that's the default toJSON behavior...?

mattdailis (changes requested):
I tested this locally and it worked! A few changes to request.

One that I didn't mention in my comments is that we'll need to add the TemporalPolyfillTypes to the response from `GenerateSchedulingLibAction.java` - in my local test, the editor in the browser underlined `Temporal` in red, saying it didn't know what it was.

> Key:
>  🔶 Requests that should not block merge, but should at least be discussed 

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
dyst5422: I think we should stick to established standards where they exist. In this case, the way to use a Duration in JS once the browsers implement is going to be `Temporal.Duration`. Introducing some other behavior is likely to become confusing.

scheduler-server/scheduling-dsl-compiler/src/main.ts
dyst5422: Seems like the standard json stringification behavior

scheduler-server/build.gradle
adrienmaillard: Using `StringValueMapper` at L.137 of `GoalBuilder`. Will see how to replace it.

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
```diff
@@ -1,3 +1,4 @@
+import "./TemporalPolyfillTypes.js";
```
dyst5422: This import shouldn't be necessary.

I assume it's because the file itself complains about type errors without it - which is fine since we aren't running it in the environment that typescript thinks we're running it in where we are actually supplying the global Temporal.

We can replace this with a triple-slash directive (a special annotation typescript understands for local type inclusion, but does not translate into runtime code). https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-

Also probably not worth our effort if it doesn't immediately work.

scheduler-server/build.gradle
adrienmaillard: Replaced 
```
       builder.duration(Duration.fromISO8601String(
            new StringValueMapper().deserializeValue(activityTemplate.arguments().get(durationType.parameterName())).getSuccessOrThrow()));
``` 
by
```
builder.duration(Duration.fromISO8601String(activityTemplate.arguments().get(durationType.parameterName()).asString().get()));
```

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-ast.ts
adrienmaillard: This `/// <reference path="./TemporalPolyfillTypes.js"; />`? File complains. Stops complaining when replacing `.js` with `.ts`.

dyst5422: That makes sense to me. Typescript is looking for TS files, not JS files, and the whole extension proxying thing likely isn't build into the references yes. The whole extension proxying is terrible and caused by the terrible migration of module systems in Nodejs.

dyst5422 (approved):

scheduler-server/build.gradle
mattdailis: Awesome :+1:

mattdailis (approved):
Confirmed that code hinting works in the UI. Looks good to me!

PR #188, opened 2022-05-26
[AERIE-1858] Facilitate recommended pattern for computed attributes

mattdailis:
* **Tickets addressed:** AERIE-1858
* **Review:** By commit  

## Description

The goal of this PR is to demonstrate a recommended pattern for declaring computed attributes. We think that the most idiomatic way to declare computed attributes is with a `record`, like so:

https://github.com/NASA-AMMOS/aerie/blob/6a5a5b4fa805772f2db76b1d884085eef8374cc9/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java#L29-L39

Records are a low-boilerplate mechanism for expressing sets of key-value pairs, where the keys are fixed (as opposed to `Maps`, which describe sets key-value pairs where the keys are dynamic).

Prior to this PR, it is still definitely possible to use `records` for computed attributes. However, every new record type would need a corresponding `ValueMapper` to be added to the value mapper factory.

In the interest of making our "recommended practice" as easy as possible for users, this PR adds an `@AutoValueMapper` annotation. This annotation can only be applied to records, and it will generate an `AutoValueMappers.java` file that contains a method that looks like this:

```java
@Generated("gov.nasa.jpl.aerie.merlin.processor.MissionModelProcessor")
public final class AutoValueMappers {
  public static ValueMapper<BiteBananaActivity.ComputedAttributes> gov_nasa_jpl_aerie_banananation_activities_BiteBananaActivity_ComputedAttributes(
      final ValueMapper<Boolean> biteSizeWasBigValueMapper,
      final ValueMapper<Flag> newFlagValueMapper) {
    return new RecordValueMapper<>(
      gov.nasa.jpl.aerie.banananation.activities.BiteBananaActivity.ComputedAttributes.class,
      List.of(
        new RecordValueMapper.Component<>(
          "biteSizeWasBig",
          gov.nasa.jpl.aerie.banananation.activities.BiteBananaActivity.ComputedAttributes::biteSizeWasBig,
        biteSizeWasBigValueMapper),
        new RecordValueMapper.Component<>(
          "newFlag",
          gov.nasa.jpl.aerie.banananation.activities.BiteBananaActivity.ComputedAttributes::newFlag,
        newFlagValueMapper)
    ));
  }
}
```

It will simultaneously create a `TypeRule` that would have been the result of parsing that factory method, had that method been defined prior to annotation processing.

## Verification

I have added unit tests for `RecordValueMapper`. I've also done some visual spot checking. What I would still like to do prior to merging this PR is some end-to-end tests, just as a sanity check that the serialization/deserialization doesn't hit any snags
- [x] manual e2e test where AutoValueMapper is used for computed attributes

<details><summary>Value Schema</summary>

```
{
    "type": "struct",
    "items": {
        "newFlag": {
            "type": "variant",
            "variants": [
                {
                    "key": "A",
                    "label": "A"
                },
                {
                    "key": "B",
                    "label": "B"
                }
            ]
        },
        "biteSizeWasBig": {
            "type": "boolean"
        }
    }
}
```

</details>

~- [ ] manual e2e test where AutoValueMapper is used as a parameter~

## Documentation

- [x] The [existing documentation](https://github.com/NASA-AMMOS/aerie/wiki/Activities#computed-attributes) already recommends using a `Map`. We should just update it to recommend a `record` instead.

- [x] Perhaps an extra section on `@AutoValueMapper` should be added to our ValueMapper docs.

## Future work

- Test in the UI
- Some of this AutoValueMapper stuff may be reworked if/when we refactor the annotation processing code


Twisol:
@mattdailis, I haven't had a chance to look too deeply at this yet, but I have a quick question. If a user opts not to use `@AutoValueMapper`, they can still provide their own custom mapper, correct? I think they would just add the value mapper to a family in one of the package's `@MissionModel.WithMappers` annotations, right?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
```diff
@@ -80,22 +82,48 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
-            missionModelGen.generateSchedulerModel(missionModelRecord),
-            missionModelGen.generateActivityActions(missionModelRecord),
-            missionModelGen.generateActivityTypes(missionModelRecord)
+            missionModelGen.generateMissionModelFactory(missionModelRecord$),
+            missionModelGen.generateSchedulerModel(missionModelRecord$),
+            missionModelGen.generateActivityActions(missionModelRecord$),
+            missionModelGen.generateActivityTypes(missionModelRecord$)
         ));
+
+        final var autoValueMapperRequests = roundEnv.getElementsAnnotatedWith(AutoValueMapper.class).stream().map(Element::asType).toList();
```
Twisol: Please add `AutoValueMapper.class` to the list of annotation roots in `getSupportedAnnotationTypes()`. (Any annotation we look up via `roundEnv.getElementsAnnotatedWith` should be added there, or else the compiler might skip us on rounds where no other annotation of interest is available.)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
```diff
@@ -80,22 +82,48 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
+        {
+          final var recordType = elementUtils.getTypeElement("java.lang.Record").asType();
+          for (final var request : autoValueMapperRequests) {
+            if (!typeUtils.isSubtype(request, recordType)) {
```
Twisol: Rather than using subtype analysis, it's better to ask the source Element if it is of kind `ElementKind.RECORD`. This check is both cheaper and clearer, and it avoids accessing elements that might not be accessible from the root element (which is generally important for incremental compilation support in e.g. Gradle.)
```suggestion
            if (request.getKind() == ElementKind.RECORD) {
```
You'll need to defer the `Element::asType` transformation you currently have on these elements until later, because `getKind()` is a method on `Element`, not on `TypeMirror`. (Generally, I've found it's more convenient to work with source `Element`s up until the point you really need to convert it to a prototypical `TypeMirror`.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
```diff
@@ -80,22 +82,48 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
+              messager.printError("AutoValueMapper is only allowed on record types. %s is not a record".formatted(request.toString()));
```
Twisol: Where possible, interpolate type names from class references. If the type ever gets renamed, we want to make sure it gets renamed in this string, too.
```suggestion
              messager.printError("@%s is only allowed on record types. %s is not a record".formatted(AutoValueMapper.class.getSimpleName(), request.toString()));
```
(It's also nice for users if we explicitly prefix annotation type names with `@`.)

Twisol: Whenever possible, try to use the overloads of `printError` that take an `Element` as an extra argument. This associates the error message to the location of that element, making it easier for a user to jump to and fix the issue. (There are also even more precise overloads of `printMessage` that focus on an annotation, or even an attribute of an annotation.)

(As with a sibling comment, this change requires deferring the use of `Element::asType` to later, so that `request` has the right type.)
```suggestion
              messager.printError("AutoValueMapper is only allowed on record types. %s is not a record".formatted(request.toString()), request);
```

Twisol: I don't think `toString` is guaranteed to produce the output we want across any particular implementation of the annotation processing API. (I believe Eclipse wraps and re-implements these types, for instance, and I have no idea what its behavior is.)

Prefer `request.getSimpleName()`, which, once again, is a method on `Element` and not `TypeMirror`. 😜 
```suggestion
              messager.printError("AutoValueMapper is only allowed on record types. %s is not a record".formatted(request.getSimpleName()));
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
```diff
@@ -80,22 +82,48 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
+            }
+          }
+        }
+        final var autoValueMappers = missionModelGen.generateAutoValueMappers(
+            missionModelRecord$,
+            autoValueMapperRequests);
+        generatedFiles.add(autoValueMappers.getLeft());
```
Twisol: (**TL;DR**: We should only process `@AutoValueMapper` annotations observed on the return types of activity effect models, not all such annotations. If we want to extend `@AutoValueMapper` to non-activity-input/output cases, we can deploy another scoping mechanism (or lean on an existing one). While `@AutoValueMapper` should not be "owned" by activity type processing, it's not a wholly independent process, either.)

---

Looking ahead to a better-behaved processor, this approach of generating a single mapper family for all `@AutoValueMapper`s observed in the first round (and thus ignoring / failing on later rounds) probably isn't tenable in the long term. I don't think we need to change that behavior but we *should* think about what kind of annotation design will allow us to do better in the future with a minimum of churn for customers.

As I understand it, the value of `@AutoValueMapper` comes in two parts:
* The generation of a `ValueMapper` implementation for each annotated type.
* The generation of a mapper family with a factory method for each `ValueMapper` generated above.

The first part is easy to do incrementally, since each mapper implementation is independent of the others.

The second is not so easy, because we can only generate the mapper family once we know about all of the mappers it needs to reference. If we just go off of the list returned by `roundEnv.getElementsAnnotatedWith()` in one round, then we will miss any `@AutoValueMapper`-annotated elements generated in later rounds. There are basically three ways to handle this kind of aggregate output:

1. Don't output the mapper family until the final round. This lets us accumulate all `@AutoValueMapper` requests until the last possible moment, but other annotation processors won't be able to process the resulting mapper family. This may be an acceptable solution for some kinds of output that aren't meant to be processed further.
2. Limit the scope of `@AutoValueMapper`s considered, so that we can enumerate a full list early in the process and thus know exactly when there are no further mappers to consider. This is exactly what `@WithActivityType` does for activity types; we provide a courtesy check that all `@ActivityType`-annotated elements are claimed by a `@WithActivityType`, but morally, we're only processing activity types referenced from the `package-info.java` annotations. We can do this explicitly (e.g. `@WithAutoMapperFor(...)`) or implicitly ("all and only `@AutoValueMapper`s on return values from an activity type's effect model), but we must ultimately circumscribe some closed family of elements that we actually process.
3. For each `@AutoValueMapper`, generate a mapper family with one factory method, rather than accumulating all of them into a single family. We still need some way to know when we have a full set of `TypeRule`s, which implies that a user would have to add an `@WithMappers` for every single output type.

You might think we could get away with (1) here, because we don't actually need to process the generated mapper family since we're generating its `TypeRule`s ahead of time. Unfortunately, we have a similar problem, in that we don't want to resolve mappers for types until we have all of the `TypeRule`s available. If we started resolving without all of the rules, we might end up using an implementation we wouldn't have. As a very concrete example, we currently have a catch-all rule for any enum. If an activity returns an enum annotated with `@AutoValueMapper` (ignoring that it's currently built only for records, I hope we see that this is a splinter of the full story of use), then we might accidentally use the catch-all rather than the specific enum mapper. What a hard problem to debug!

I also think (3) is a non-starter, because it basically requires a user to add both a `@WithMappers` and a `@WithActivityType` for each activity type. This seems redundant, and also easy to forget about when developing an activity type (e.g. changing its return from `void` to an actual output).

On the other hand, I think (2) is pretty tenable. `@AutoValueMapper` would not be processed in the absence of other information (because we run into the open family issue again), but rather describes an inert desire *in the scope of* some other process. We can generate an `ActivityAttributeMappers` class covering all *and only* the value mappers for the processed activity types, which is a closed family because the activity types are themselves a closed family (enumerated by `@WithActivityTypes`.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +655,90 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+                  List.of(new TypePattern.ClassPattern((ClassName) ClassName.get(component.getRight()).box(), List.of()))))
+              .toList(),
+          typeName,
+          methodName);
+      typeRules.add(typeRule);
+
+      final var methodBuilder = MethodSpec
+          .methodBuilder(methodName)
+          .addModifiers(Modifier.PUBLIC)
+          .addModifiers(Modifier.STATIC)
```
Twisol: As implied by plural `Modifiers`, you can combine these calls.
```suggestion
          .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +655,90 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+          .returns(ParameterizedTypeName.get(ClassName.get(ValueMapper.class), TypeName.get(record)))
+          .addParameters(
+              recordComponents
+                  .stream()
+                  .map(recordComponent -> ParameterSpec
+                      .builder(
+                          ParameterizedTypeName.get(
+                              ClassName.get(ValueMapper.class),
+                              ClassName.get(recordComponent.getRight()).box()),
+                          recordComponent.getLeft() + "ValueMapper",
```
Twisol: If we have `record A(int a, int b, int c)`, I believe this will produce a method with parameters `(ValueMapper<Integer>, ValueMapper<Integer>, ValueMapper<Integer>)`. D'you think it would be possible to deduplicate parameters so we only need one of any given type?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: Change made :+1:

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
mattdailis: Done :+1:

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
mattdailis: Change made :+1: (Though I'm still fuzzy on `Element` vs `TypeElement`, but both of them have a `getKind()` method, so 🤷 )

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
mattdailis: Good point, change made :+1:

mattdailis: Done. I do really like how the warning prints out with the exact code location ❤️

mattdailis: Good point! Thanks. Change made.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: I've reworked the generator to add one parameter per unique type, and to track which record components need which type of value mapper.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
mattdailis: I've updated the code to only generate value mappers for records that are the return type of an effect model.

pcrosemurgy:
@mattdailis Sorry if I'm just out of the loop but what's the state of this PR? Is it in kind of a do-not-merge place right now?

mattdailis:
> @mattdailis Sorry if I'm just out of the loop but what's the state of this PR? Is it in kind of a do-not-merge place right now?

It's available for review! I probably need to rebase/resolve merge conflicts, I can do that tomorrow morning :+1:

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +658,100 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+                AnnotationSpec
+                    .builder(javax.annotation.processing.Generated.class)
+                    .addMember("value", "$S", MissionModelProcessor.class.getCanonicalName())
+                    .build())
+            .addModifiers(Modifier.PUBLIC, Modifier.FINAL);
+
+    record ComponentMapperNamePair(String componentName, String mapperName) {}
+    for (final var record : recordTypes) {
+      if (record.getKind() != ElementKind.RECORD) {
+        messager().printError("@%s is only allowed on records".formatted(AutoValueMapper.class.getSimpleName()), record);
```
pcrosemurgy: Good edge-case to check for! This looks like something that should likely be in the mission model parser logic instead of the code generator though.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +658,100 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+                      .build())
+                  .toList());
+
+      final var iter = componentToMapperName.iterator();
+      final var codeBlockBuilder = CodeBlock.builder();
+      codeBlockBuilder.add("return new $T<>(\n", RecordValueMapper.class);
+      codeBlockBuilder.add("  $L.class,\n", record.toString());
+      codeBlockBuilder.add("  $T.of(\n", List.class);
+
+      while (iter.hasNext()) {
```
pcrosemurgy: I'd replace this with:
```java
for (final var iter = componentToMapperName.iterator(); iter.hasNext();)
```

This whole block is cool with me but you could also do some variety of a `componentToMapperName.stream().map(...).reduce(...)` to inject the `","` text between each generated block. I think keeping this imperative actually makes the most sense for readability though.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/MissionModelRecord.java
```diff
@@ -13,20 +14,23 @@ public final class MissionModelRecord {
...
-      final List<ActivityTypeRecord> activityTypes)
+      final List<ActivityTypeRecord> activityTypes,
+      final Collection<TypeElement> autoValueMapperRequests)
   {
     this.$package = Objects.requireNonNull($package);
     this.topLevelModel = Objects.requireNonNull(topLevelModel);
     this.modelConfigurationType = Objects.requireNonNull(modelConfigurationType);
     this.typeRules = Objects.requireNonNull(typeRules);
     this.activityTypes = Objects.requireNonNull(activityTypes);
+    this.autoValueMapperRequests = autoValueMapperRequests;
```
pcrosemurgy: For consistency I'd either add `Objects.requireNonNull(...)` here or refactor this class to be a `record` (without any `Objects.requireNonNull(...)` checks if you like).

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +658,100 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+                              ClassName.get(mapperRequest.getKey()).box()),
+                          mapperRequest.getValue(),
+                          Modifier.FINAL)
+                      .build())
+                  .toList());
+
+      final var iter = componentToMapperName.iterator();
+      final var codeBlockBuilder = CodeBlock.builder();
+      codeBlockBuilder.add("return new $T<>(\n", RecordValueMapper.class);
+      codeBlockBuilder.add("  $L.class,\n", record.toString());
```
Twisol: Instead of using `toString` (which is not always guaranteed to produce a meaningful name on `TypeElement`s), we should use JavaPoet's static `ClassName.get(TypeElement)` method. We could also use `TypeElement`'s own `getQualifiedName()`, which returns a `Name` (a kind of `CharSequence`, yes Java is weird), but `ClassName` works better with codegen -- which brings me to the next point:

As a rule of thumb, we should use `$T` whenever the literal refers to a type. For instance, JavaPoet can automatically import types used with `$T` placeholders, or it can use the fully-qualified class name if the short name is already used by another class in scope. It just gives JavaPoet more to work with.

```suggestion
      codeBlockBuilder.add("  $T.class,\n", ClassName.get(record));
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -646,6 +658,100 @@ public JavaFile generateActivityTypes(final MissionModelRecord missionModel) {
...
+      codeBlockBuilder.add("  $T.of(\n", List.class);
+
+      while (iter.hasNext()) {
+        final var recordComponent = iter.next();
+        codeBlockBuilder.add("    new $T<>(\n", RecordValueMapper.Component.class);
+        codeBlockBuilder.add("      $S,\n", recordComponent.componentName());
+        codeBlockBuilder.add("      $L::$L,\n", record.toString(), recordComponent.componentName());
```
Twisol:
```suggestion
        codeBlockBuilder.add("      $T::$L,\n", ClassName.get(record), recordComponent.componentName());
```

pcrosemurgy:
@mattdailis let me know if you want/need to pair at all to get this PR through 👌 The conflicts with `develop` are likely from changes I've pushed through since this PR was opened.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: Hmm, I'm not convinced that using an iterator for loop is more readable than a while loop (I think it kind of obfuscates rather than clarifies what's going on)... however! By poking at this you've given me the motivation to go read the javapoet docs a little bit, and it turns out there's a `CodeBlock.join` method, which makes your second suggestion the best option, I think. That loop has been replaced with:

```java
codeBlockBuilder.add(
    CodeBlock.join(componentToMapperName
      .stream()
      .map(recordComponent ->
               CodeBlock
                   .builder()
                   .add("    new $T<>(\n", RecordValueMapper.Component.class)
                   .add("      $S,\n", recordComponent.componentName())
                   .add("      $T::$L,\n", ClassName.get(record), recordComponent.componentName())
                   .add("    $L)", recordComponent.mapperName())
                  .build())
      .toList(), ",\n"));
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: I've moved this to the parser, as requested :+1:

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/MissionModelRecord.java
mattdailis: Done :+1:

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
mattdailis: 📏 👍

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
pcrosemurgy: Ohhh awesome! Thanks, this looks very clean.

Twisol: I had no idea `CodeBlock.join` existed. *Nice.*

pcrosemurgy (approved):
Great work Matt!

Last question/comment I have is about the `@AutoValueMapper`. I'm wondering if this should be called `RecordValueMapper` or something that indicates it's only supported for computed attribute return types.

Maybe the thinking is that this annotation will eventually support auto value mappers for other parts of mission model? If not then I could see it causing some confusion if developers think it's truly an auto value mapper (without context).

Twisol:
> Maybe the thinking is that this annotation will eventually support auto value mappers for other parts of mission model? If not then I could see it causing some confusion if developers think it's truly an auto value mapper (without context).

I can definitely see this being extended to activity types in the future -- I'm imagining this coming in when we extract resource registration out of model initialization. At that point we should probably move activity type registration into the same place that resource registration ends up, which would mean `@ActivityType`'s behavior will reduce almost entirely to `@AutoValueMapper`'s behavior.

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/serialization/mappers/RecordValueMapperTest.java
```diff
@@ -0,0 +1,182 @@
...
+    if (!(SerializedValue.of(Map.of(
+        "aString", SerializedValue.of("one small step for crayon"),
+        "fancy", permutation1)).equals(serializedValue) || SerializedValue.of(Map.of(
+        "aString", SerializedValue.of("one small step for crayon"),
+        "fancy", permutation2)).equals(serializedValue))) {
+      assertEquals(SerializedValue.of(Map.of(
+          "aString", SerializedValue.of("one small step for crayon"),
+          "fancy", permutation1)), serializedValue);
+    } else {
+      //pass
```
adrienmaillard: Why this empty block ?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/AutoValueMapper.java
```diff
@@ -0,0 +1,10 @@
+package gov.nasa.jpl.aerie.merlin.framework.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE)
+public @interface AutoValueMapper {}
```
adrienmaillard: I agree with @pcrosemurgy comment on the name. `AutoValueMapper` sounds like it works for every kind of object. Maybe `RecordAutoValueMapper` ? Or we plan to generalize this auto value mapper ?

adrienmaillard (approved):
It's all great. Thank you for the tour, it definitely helped. I only have two minor non-blocking comments.

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/serialization/mappers/RecordValueMapperTest.java
mattdailis: The problem I ran into was that the Map would be serialized in a non-deterministic order, so this test checks equality with both possible orders. I've updated the comment to explain this.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/annotations/AutoValueMapper.java
mattdailis: I've updated the annotation to be `AutoValueMapper.Record`

PR #189, opened 2022-05-26
[AERIE-1861] Hasura `resourceSamples` action

pcrosemurgy:
* **Tickets addressed:** AERIE-1861
* **Review:** By commit  

## Description
Adds `resourceSamples` Hasura action.

An action to return resource samples is needed as an intermediate step for decoupling sim. results and resource samples/profiles. Clients (UI and GQL users) can incrementally switch over to ditching sim. results in favor of this action and sim. results stored in the DB.

## Verification

Running the following query returns resource samples for a given dataset:
```gql
query {
  resourceSamples(planId: 1) {
    resourceSamples
  }
}
```

Prior to running a simulation (prior to the existence of dataset ID `1`) this returns:
```
{
  "data": {
    "resourceSamples": {
      "resourceSamples": { }
    }
  }
}
```

After running a simulation (after dataset ID `1` exists) this returns:
```
{
  "data": {
    "resourceSamples": {
      "resourceSamples": {
        "/data/line_count": [
          {
            "x": 0,
            "y": 12
          },
...
```

## Documentation
This section will be updated after 1-2 reviews, this way any changes to the response structure can be reflected properly in our Wiki documentation.

## Future work
Carving out sim. results (https://jira.jpl.nasa.gov/browse/AERIE-1796).

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -92,6 +92,12 @@ type Query {
 }
 
 type Query {
+  resourceSamples(
+    planId: Int!
+  ): ResourceSamplesResponse
+}
+
+type Query {
   validateActivityArguments(
     activityTypeName: String!
     missionModelId: ID!
@@ -242,9 +248,15 @@ type CodeLocation {
   column: Int!
 }
 
+type ResourceSamplesResponse {
+  resourceSamples: ResourceSamples!
+}
+
 scalar ResourceSchema
 
 scalar MerlinSimulationResults
+
+scalar ResourceSamples
 
 scalar MerlinSimulationFailureReason
 
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
@@ -12,10 +12,12 @@ import gov.nasa.jpl.aerie.constraints.time.Window;
 import gov.nasa.jpl.aerie.constraints.tree.Expression;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.merlin.server.exceptions.NoSuchPlanException;
 import gov.nasa.jpl.aerie.merlin.server.models.Constraint;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import org.apache.commons.lang3.tuple.Pair;
 
 import javax.json.Json;
 import java.io.StringReader;
@@ -28,7 +30,7 @@ import java.util.Map;
 import java.util.Objects;
 
 public final class GetSimulationResultsAction {
-  public /*sealed*/ interface Response {
+  public sealed interface Response {
     record Pending() implements Response {}
     record Incomplete() implements Response {}
     record Failed(String reason) implements Response {}
@@ -74,6 +76,13 @@ public final class GetSimulationResultsAction {
     } else {
       throw new UnexpectedSubtypeError(ResultsProtocol.State.class, response);
     }
+  }
+
+  public Map<String, List<Pair<Duration, SerializedValue>>> getResourceSamples(final PlanId planId)
+  throws NoSuchPlanException
+  {
+    final var revisionData = this.planService.getPlanRevisionData(planId);
+    return simulationService.getResourceSamples(planId, revisionData);
   }
 
   public Map<String, List<Violation>> getViolations(final PlanId planId, final SimulationResults results)
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/UncachedSimulationService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/UncachedSimulationService.java
@@ -1,11 +1,20 @@
 package gov.nasa.jpl.aerie.merlin.server.services;
 
+import java.util.List;
+import java.util.Map;
+
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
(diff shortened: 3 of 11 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -76,6 +79,11 @@ public Response run(final PlanId planId) throws NoSuchPlanException {
     }
   }
 
+  public Map<String, List<Pair<Duration, SerializedValue>>> getResourceSamples(final SimulationDatasetId datasetId) {
+    if (simulationService instanceof final CachedSimulationService css) return css.getResourceSamples(datasetId);
+    throw new RuntimeException("Unable to retrieve resource samples from uncached simulation service");
```
pcrosemurgy: Maybe this is not too graceful but we basically can only return resource samples for simulation results that are **cached**. So an uncached simulation service should not support returning resource samples.

For some peace of mind I made `SimulationService` a `sealed` class that can be only `CachedSimulationService`/`UncachedSimulationService`, so this "contract" is at least codified.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/UncachedSimulationService.java
```diff
@@ -5,7 +5,10 @@
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 import gov.nasa.jpl.aerie.merlin.server.remotes.InMemoryResultsCellRepository.InMemoryCell;
 
-public record UncachedSimulationService (SimulationAgent action) implements SimulationService {
+public record UncachedSimulationService (
+    SimulationAgent agent
+) implements SimulationService {
```
pcrosemurgy, line 16: Reformatted this to match `CachedSimulationService`'s formatting. Makes it easier to compare the two.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/ResultsCellRepository.java
```diff
@@ -2,15 +2,18 @@
...
 
 public interface ResultsCellRepository {
   ResultsProtocol.OwnerRole allocate(PlanId planId);
 
-  Optional<ResultsProtocol.OwnerRole> claim(PlanId planId, Long datasetId);
+  Optional<ResultsProtocol.OwnerRole> claim(PlanId planId, SimulationDatasetId datasetId);
 
   Optional<ResultsProtocol.ReaderRole> lookup(PlanId planId);
 
+  Optional<ResultsProtocol.ReaderRole> lookup(SimulationDatasetId datasetId);
```
mattdailis: Following up on our discussion in slack, since simulation dataset ids are not the same thing as dataset ids, we should use the more verbose `simulationDatasetId` to be uber-clear.

When a lot of this code was written, the simulation dataset did not have a synthetic key.

pcrosemurgy: Good point! I'll use `simulationDatasetId`.

pcrosemurgy: Done (:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
mattdailis: Hmm... this does make me wonder if we've designed ourselves into a corner a bit. Why is it that the uncached service cannot implement `getResourceSamples`? In theory, I would expect an uncached service's `getResourceSamples` to run a simulation (since it doesn't have cached results), and use those simulation results to produce a response.

pcrosemurgy: >Why is it that the uncached service cannot implement getResourceSamples?

The technical reason here is that `getResourceSamples(SimulationDatasetId simulationDatasetId)` doesn't make sense in an uncached scenario (where a simulation dataset must first exist). The Hasura action `resourceSamples` accepts a `simulationDatasetId` so we have to disallow an uncached sim. service from trying to return resource samples when provided a simulation dataset ID.

We could maybe make another Hasura action for retrieving samples given a plan ID. This action could lazily run a sim. and return samples. However, we should keep in mind that `UncachedSimulationService` is only used in the `DevAppDriver` – not in production. ~~So I'm not sure it deserves a full Hasura action and the additional complexity/bloat that comes with it.~~

**Edit:** A full Hasura action wouldn't be needed but `GetSimulationResultsAction.getResourceSamples()` would probably have to be redesigned to be polymorphic (accept either a `PlanId` or `SimulationDatasetId`) and in each polymorphic method we'd still have to throw an error if the underlying `SimulationService` is cached vs. uncached.

pcrosemurgy: Happy to hop on a call about this. The solution I've settled on feels more like a "best of all bad options".

pcrosemurgy: @mattdailis I talked with @patkenneally and we seem to all be in agreeance that `resourceSamples` should be provided a plan ID. This way the cached vs. uncached backends will both be satisfied. The only catch here is that the cached backend will just need to check if a plan has an associated simulation prior to checking for samples.

pcrosemurgy: I've reworked this PR to have `resourceSamples` accept a plan ID. :)

deployment/hasura/metadata/actions.graphql
```diff
@@ -91,6 +91,12 @@ type Query {
   ): MerlinSimulationResponse
 }
 
+type Query {
+  resourceSamples(
+    planId: Int!
+  ): ResourceSamplesResponse
```
camargo, line 97: Would GraphQL allow us to just return a `ResourceSamples!` here?

pcrosemurgy: Unfortunately not :/ I tried that at first but you must return a `type` type, not a `scalar`. Kind of annoying because I consider both to be "types".

camargo (approved):
Tested locally with the corresponding UI branch. Works great!

mattdailis (approved):

PR #197, opened 2022-06-02
[AERIE-1857]  command expansion timestamps

dyst5422:
* **Tickets addressed:** AERIE-1857
* **Review:** By commit  

## Description

Added created_at columns to all non-join tables in command expansion. Added updated_at to sequence and expansion_rule columns as those are the ones that get updated.

Note that the trigger to update the sequence updated_at when the sequence_to_simulated_activity join table updates is in the sequence_to_simulated_activity.sql file is because of execution order when deploying, the sequence_to_simulated_activity table needs to exist before creating the trigger.

This ordering of sql file execution has a high likelihood of biting us in other places.

## Verification

Added tests for the updated_at columns, but the create_at columns are so simple, as long as they parse via postgres, a test would just be overhead

## Documentation

No documentation needed

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/CommandExpansionDatabaseTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/CommandExpansionDatabaseTests.java
@@ -10,6 +10,7 @@ import java.io.File;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.UUID;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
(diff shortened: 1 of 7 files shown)


db-tests/src/test/java/gov/nasa/jpl/aerie/database/CommandExpansionDatabaseTests.java
```diff
@@ -0,0 +1,147 @@
...
+        final var created_at = insertRes.getTimestamp("created_at");
+        final var updated_at = insertRes.getTimestamp("updated_at");
+
+        assertEquals(created_at, updated_at);
+
+        final var updateRes = statement.executeQuery("""
+        update expansion_rule set expansion_logic = 'updated-logic'
+        where id = %d
+        returning created_at, updated_at
+      """.formatted(id));
```
goetzrrGit: For some reason, my workspace is telling me `formatted` is not found.

goetzrrGit (commented):
I ran through the whole command expansion to Sequence JSON by hand and it looks good. 

I am unable to run the test though as `formatted` doesn't exist. Maybe you wanted `format`?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/CommandExpansionDatabaseTests.java
dyst5422: Might be a misconfiguration on your ide
Screen Shot 2022-06-06 at 8 39 05 AM

goetzrrGit: Probably, I am going to 'Approve' as it is an issue on my end, and I already tested it by hand and even updated the sequence to verify the modified date was changed.

goetzrrGit (approved):

camargo (approved):
LGTM! Will add this to the UI table columns after this is merged.

PR #201, opened 2022-06-06
[AERIE 1871]  simulated activity id metadata in generated sequences

dyst5422:
* **Tickets addressed:** AERIE-1871
* **Review:** By commit  

## Description

Adds the simulated activity id to the command metadata

## Verification

Added the metadata to the expansion test check

## Documentation

No documentation update needed (at some point when things are stable, we'll document the output of expansion more fully)

## Future work

None


command-expansion-server/src/lib/codegen/CodegenHelpers.ts
```diff
@@ -12,3 +12,5 @@ export function indent(text: string): string {
     .map(line => `\t${line}`)
     .join('\n');
 }
+
+export type Mutable<T> = { -readonly [P in keyof T]: T[P] };
```
goetzrrGit, line 16: I have never seen a negative type before?
`-readonly`

goetzrrGit (approved):
Good test!!! Seems pretty straightforward.

command-expansion-server/src/lib/codegen/CodegenHelpers.ts
dyst5422: It's a cute little utility thing. Useful in cases like this where we really want it to be immutable, but we want to knowingly override that here

dyst5422: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers

mattdailis (approved):

command-expansion-server/src/worker.ts
```diff
@@ -81,6 +82,12 @@ export async function executeExpansion(opts: {
         commands = [commands];
       }
       const commandsFlat = commands.flat() as Command[];
+      for (const command of commandsFlat) {
+        (command as Mutable<Command>).metadata  = {
```
pcrosemurgy, line 86: I trust that you're doing the right thing here but I'm not sure I follow what's going on with `Mutable<>`.

Would a more functional map-style approach not work? I'm assuming you want these to be mutable as an optimization?

goetzrrGit: Going to resolve this conversation and merge this PR as I am blocked. Dylan still seems to be out today, but he might see this and respond to your question. I don't have an answer for you :(

goetzrrGit:
I am going to merge this as Dylan is still out and will need this for my PRs. 

Dylan, I hope you don't mind.

PR #205, opened 2022-06-06
[AERIE-1882] Remove old constraints pipeline

JoelCourtney:
* **Tickets addressed:** AERIE-1882
* **Review:** By commit  

## Description
- **Commit 1:** removes the pipeline switch and enables the constraints eDSL compilation by default.
- **Commits 2-5:** removes unneeded or redundant code that was being kept around to keep the old pipeline functional.
- **Commits 6-9:** renames nodes of the expression tree to be less suggestive of boolean operations.
    - Currently a constraint can be envisioned as either a list of set operations that transform profiles into violation windows, or as a function from profile values to booleans. However, once we start adding operations like #186, the function-of-values-to-booleans mindset will break down. So I'm trying to make the API suggest the set operations mindset to avoid confusion.
- **Commit 10:** This fixes an oversight from my last PR. Real profiles technically have a schema of `{ initial: REAL, rate: REAL }` instead of just `REAL`, so they were not being recognized.

## Verification
Tests for the new pipeline are updated. Tests for the old pipeline are removed.

## Documentation
API documentation is updated. Wiki documentation is not yet. I will update the Wiki after this is approved, but before it is merged.


Twisol:
> * Currently a constraint can be envisioned as either a list of set operations that transform profiles into violation windows, or as a function from profile values to booleans. However, once we start adding operations like #186, the function-of-values-to-booleans mindset will break down. So I'm trying to make the API suggest the set operations mindset to avoid confusion.

FWIW, at the time we started designing constraints, the conceptual foundation was that a constraint was a *time-indexed proposition*. Rather than simply being true or false, it takes on boolean valuations *at individual times*. (There was some idea that propositions should be continuous as functions from the real line to the booleans, to avoid certain pathological behavior, but I don't think that's important here.)

More to the point, all logical operators behave point-wise in this sense. `NOT P` is true at a time `t` if `P` is false at time `t`, and vice versa. `P AND Q` is true whenever both `P` and `Q` are true. And so on.

If you think of "expand" as satisfying use-cases like "within 5 minutes before or after P is true", I don't think it's in conflict with the idea of time-indexed boolean propositions. It's not an operator of classical logic, but that's what modal logic is for. And we anticipated `expand` and `contract` at the time (though not by those names; I think I imagined them in terms of a convolution operator.)

JoelCourtney:
Rebased on top of #179

mattdailis:
Looks like the test failures have to do with struct order non-determinism again...

One thing we could do is change this line:
https://github.com/NASA-AMMOS/aerie/blob/ba4febe6f5802220b852b9ed21c83b91665b324f/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java#L211

to:
```java
        for (final var member: values.keySet().stream().sorted().toList()) {
```

mattdailis (approved):

pcrosemurgy (approved):

JoelCourtney:
Updated wiki

PR #206, opened 2022-06-06
[AERIE-1884] Serialize Boolean Arguments when Expanding

goetzrrGit:
* **Tickets addressed:** AERIE-1884
* **Review:** By commit  

## Description
Europa Clipper will need to have their boolean arguments serialized to '0' and '1' in the Sequence JSON.

I added a serialize method in case we need to serialize more arguments in the future

`AVS_SS_READ_CHECK_EMEM_NAND(10, 10, 10, false, 10)`

## Verification
Generated a Sequence JSON from the above command and verified it serialized the boolean value

```
      {
          "type": "command",
          "stem": "AVS_SS_READ_CHECK_EMEM_NAND",
          "args": [
            10,
            10,
            10,
            0,  <------ boolean value 'false'
            10
          ],
          "time": {
            "tag": "00:15:00.000",
            "type": "EPOCH_RELATIVE"
          },
          "metadata": {}
        }
```

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -142,7 +142,10 @@ export class Command<
 
   public toSeqJson(): CommandSeqJson {
     return {
-      args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
+      args:
+        typeof this.arguments == 'object'
+          ? this.serializeArguments(Object.values(this.arguments))
+          : this.serializeArguments(this.arguments),
       stem: this.stem,
       time:
         this.absoluteTime !== null
@@ -155,6 +158,16 @@ export class Command<
       type: 'command',
       metadata: {},
     };
+  }
+
+  private serializeArguments(args: ArgType[]): ArgType[] {
+    return args.map(arg => {
+      if (typeof arg == 'boolean') {
+        // Europa Clipper boolean values are 0 or 1
+        return arg ? 1 : 0;
+      }
+      return arg;
+    });
   }
 
   public static fromSeqJson<A extends ArgType[]>(json: CommandSeqJson<A>): Command<A> {
--- a/command-expansion-server/test/sequence-generation.spec.ts
+++ b/command-expansion-server/test/sequence-generation.spec.ts
@@ -48,7 +48,7 @@ beforeEach(async () => {
     return [
       PREHEAT_OVEN({temperature: 70}),
       BAKE_BREAD,
-      PREPARE_LOAF(50, false),
+      PREPARE_LOAF(50, true),
     ];
   }
   `,
@@ -111,11 +111,11 @@ it('should return sequence seqjson', async () => {
   expect(getSequenceSeqJson.metadata).toEqual({});
   expect(getSequenceSeqJson.steps).toEqual([
     { type: 'command', stem: 'PREHEAT_OVEN', time: { type: 'COMMAND_COMPLETE' }, args: [70], metadata: {} },
-    { type: 'command', stem: 'PREPARE_LOAF', time: { type: 'COMMAND_COMPLETE' }, args: [50, false], metadata: {} },
+    { type: 'command', stem: 'PREPARE_LOAF', time: { type: 'COMMAND_COMPLETE' }, args: [50, 0], metadata: {} },
     { type: 'command', stem: 'BAKE_BREAD', time: { type: 'COMMAND_COMPLETE' }, args: [], metadata: {} },
     { type: 'command', stem: 'PREHEAT_OVEN', time: { type: 'COMMAND_COMPLETE' }, args: [70], metadata: {} },
     { type: 'command', stem: 'BAKE_BREAD', time: { type: 'COMMAND_COMPLETE' }, args: [], metadata: {} },
-    { type: 'command', stem: 'PREPARE_LOAF', time: { type: 'COMMAND_COMPLETE' }, args: [50, false], metadata: {} },
+    { type: 'command', stem: 'PREPARE_LOAF', time: { type: 'COMMAND_COMPLETE' }, args: [50, 1], metadata: {} },
   ]);
 
   // Cleanup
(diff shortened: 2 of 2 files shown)


Twisol:
Pardon my dropping these comments on this PR, but it does seem relevant...

On the topic of command serialization, has serialization for the dynamic invocation commands (e.g. `SEQ_VAR_CMD`) been considered? There's a pretty fair amount of complexity in that. Commands with variable arguments need to be either translated into dynamic invocations, or their variable references need to be preserved in the output (so that something like CTS can do that encoding), and given that booleans are being encoded here, I imagine you're on the road to the former solution.

Different missions also have different needs for encoding. Worse, they're governed by controlled documents (specifically, FG-ICDs). Given that Aerie is an open-source codebase, are there any plans for decoupling Aerie components from the mission-specific encoding information, or for obtaining clearance to release such data as part of Aerie?

(cc/ @patkenneally)

dyst5422:
Was just going to say that the serialization should be derived from the command dictionary definition for the arguments at a minimum and if we need some sort of logic that isn't specified there, we should ideally advocate for its addition, but have injected logic for this that is configured in app.ts if that is not possible. 

@Twisol i agree for the most part, but until we get another mission utilizing Aerie, we won't know the problem/solution space for this completely. So a fully fleshed out multi mission solution will likely have to wait until that point. But we can definitely develop the capacity to make this part of the api agnostic to it.

goetzrrGit:
Sorry guys, I am not following the conversation very well. I might need you to ELI5 :P 

As of now, we don't have a way to configure AERIE to be mission-specific. I know Dylan brought this up a month ago when he put together the Sequence JSON logic. He made the Sequence JSON generation as generic as possible, but the output is tailored to Clipper as they are the customer using this feature at the moment. I think we will need to rethink this when we get a new mission. 


From a "dynamic" invocation command, there is a max number of arguments the commands can dynamically expand to. For the commands I have looked at so far in the Clipper command dictionary, I have seen a max of 10. 

For `SEQ_VAR_CMD` the command dictionary max "dynamic" arguments are 7. We are transcoding the dictionary like this:

```
/**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
	function SEQ_VAR_CMD(
	opcode: U16, 
	): SEQ_VAR_CMD;

	/**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
	function SEQ_VAR_CMD(
	dynamic_arg1: VarString<8, 1024>,
	opcode: U16, 
	): SEQ_VAR_CMD;

	/**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
	function SEQ_VAR_CMD(
	dynamic_arg1: VarString<8, 1024>,
	dynamic_arg2: VarString<8, 1024>,
	opcode: U16, 
	): SEQ_VAR_CMD;
	
     etc....
```

dyst5422:
Perhaps it makes sense to just output them without any special handling to serialization (eg dates, booleans, etc) and have the mission write their own transformer for it to implement their own mission specific serialization scheme? @patkenneally

dyst5422:
> Sorry guys, I am not following the conversation very well. I might need you to ELI5 :P
> 
> As of now, we don't have a way to configure AERIE to be mission-specific. I know Dylan brought this up a month ago when he put together the Sequence JSON logic. He made the Sequence JSON generation as generic as possible, but the output is tailored to Clipper as they are the customer using this feature at the moment. I think we will need to rethink this when we get a new mission.
> 
> From a "dynamic" invocation command, there is a max number of arguments the commands can dynamically expand to. For the commands I have looked at so far in the Clipper command dictionary, I have seen a max of 10.
> 
> For `SEQ_VAR_CMD` the command dictionary max "dynamic" arguments are 7. We are transcoding the dictionary like this:
> 
> ```
> /**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
> 	function SEQ_VAR_CMD(
> 	opcode: U16, 
> 	): SEQ_VAR_CMD;
> 
> 	/**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
> 	function SEQ_VAR_CMD(
> 	dynamic_arg1: VarString<8, 1024>,
> 	opcode: U16, 
> 	): SEQ_VAR_CMD;
> 
> 	/**Executes a dynamic command with arguments that are either specified via the ground or the values of on-board variables.*/
> 	function SEQ_VAR_CMD(
> 	dynamic_arg1: VarString<8, 1024>,
> 	dynamic_arg2: VarString<8, 1024>,
> 	opcode: U16, 
> 	): SEQ_VAR_CMD;
> 	
>      etc....
> ```

Twisol:
No worries, Ryan. I have two concrete concerns:

1. If booleans are being encoded in this output format, then it seems likely that other types of arguments will need to be encoded as well. The encoding process for commands involving variables is the most involved, so I raise it as a (Clipper-specific) example.

    What does the JSON output format for `SEQ_ECHO L00STR` look like? Given that booleans are also being encoded, I expect it to look like a rather horrifying hex string.

    As such, I would strongly argue that the JSON output is *not* the right place to be doing this kind of encoding.

2. Regardless of whether we can build a multimission system at this point in time, concrete encoding formats *are* governed by controlled documents. Command dictionaries and information derived from them are much the same. I don't think it's appropriate for Aerie to include mission-specific encoding logic without authorization for uncontrolled release of such information.

Twisol:
> Perhaps it makes sense to just output them without any special handling to serialization (eg dates, booleans, etc) and have the mission write their own transformer for it to implement their own mission specific serialization scheme? @patkenneally

IMHO, this makes sense to me. There is/was a strong desire in MGSS to keep the logic for command encoding centralized to one product, namely SLINC II / CTS, which used to be owned by MPSA (but was transferred to MPCS shortly before I left). By pushing this concern downstream, missions are free to either (a) feed sequence data into CTS for the so-called "gold standard" encoding, or (b) take ownership of the artifact encoding process themselves.

Clipper is still using CTS last I heard. If they weren't going to use CTS, I wouldn't have had to extend CTS' encoding capabilities for them.

PR #207, opened 2022-06-07
[HOTFIX] Simulation results persistence bugs

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

There are two bugs addressed in this PR:
1. Events from two different transactions (aka "commits") are treated as duplicates and rejected by postgres
2. There are no spans persisted for child activities

The cause of bug 1 is that the event table didn't include `transaction_index` in its primary key - so events that differ in the transaction index, but not in the real time or causal time, would erroneously trigger an error. (This is the bug that @camargo noticed when simulating a plan that included a `parent` activity from Banananation).

I noticed bug 2 when testing to make sure I had fixed bug one 😅 . I expected to see the whole decomposition tree of the `parent` activity, but instead I would only see the top level activity, and no children. I checked the database, and saw that the child activities were not being persisted in the `span` table.

I traced the bug to the `TaskInfo::isActivity` method, which defined activities as "tasks that correspond to directives". This definition includes top level activities, but not child activities. This PR changes that definition to be "tasks that have emitted an event to a registered `input` topic".

## Verification

I've run simulations through the UI to make sure that this fix did the job:

Screen Shot 2022-06-07 at 10 44 49 AM

## Documentation

No documentation should be affected; these are internal bug fixes.

## Future work

- Just noticed the "undefined" start times in the table at the bottom of the screenshot - not sure what's up with those.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -343,7 +343,7 @@ public final class SimulationEngine implements AutoCloseable {
     }
 
     public boolean isActivity(final TaskId id) {
-      return this.taskToPlannedDirective.containsKey(id.id());
+      return this.input.containsKey(id.id());
     }
 
     public record Trait(MissionModel<?> missionModel, Topic<ActivityInstanceId> activityTopic)
--- a/merlin-server/sql/merlin/tables/event.sql
+++ b/merlin-server/sql/merlin/tables/event.sql
@@ -9,7 +9,7 @@ create table event
   topic_index integer not null,
 
   constraint event_natural_key
-    primary key (dataset_id, real_time, causal_time),
+    primary key (dataset_id, real_time, transaction_index, causal_time),
   constraint event_owned_by_topic
     foreign key (dataset_id, topic_index)
       references topic


pcrosemurgy:
>so events that differ in the transaction index, but not in the real time or causal time, would erroneously trigger an error.

Out of curiosity, what was this error (just roughly)?

pcrosemurgy (approved):
Great work Matt!

goetzrrGit (approved):

camargo (approved):
Tested locally and it works great! LGTM

camargo:
@mattdailis `undefined` is because the default UI views need to be updated with the correct property names - will open a quick PR for that now

mattdailis:
> Out of curiosity, what was this error (just roughly)?

Here's the message @camargo posted:

```
2022-06-06 23:27:19.533 UTC [1] LOG:  database system is ready to accept connections
2022-06-06 23:28:34.366 UTC [88] ERROR:  duplicate key value violates unique constraint "event_1_pkey"
2022-06-06 23:28:34.366 UTC [88] DETAIL:  Key (dataset_id, real_time, causal_time)=(1, 132 days 11:48:07.128, .1) already exists.
2022-06-06 23:28:34.366 UTC [88] STATEMENT:  insert into event (dataset_id, real_time, transaction_index, causal_time, topic_index, value)
	  values ($1, $2::timestamptz - $3::timestamptz, $4, $5, $6, $7)
2022-06-06 23:28:34.368 UTC [88] ERROR:  duplicate key value violates unique constraint "event_1_pkey"
2022-06-06 23:28:34.368 UTC [88] DETAIL:  Key (dataset_id, real_time, causal_time)=(1, 189 days 11:48:07.128, .1) already exists.
2022-06-06 23:28:34.368 UTC [88] STATEMENT:  insert into event (dataset_id, real_time, transaction_index, causal_time, topic_index, value)
	  values ($1, $2::timestamptz - $3::timestamptz, $4, $5, $6, $7)
2022-06-06 23:28:34.368 UTC [88] ERROR:  duplicate key value violates unique constraint "event_1_pkey"
2022-06-06 23:28:34.368 UTC [88] DETAIL:  Key (dataset_id, real_time, causal_time)=(1, 189 days 11:48:07.128, .1) already exists.
2022-06-06 23:28:34.368 UTC [88] STATEMENT:  insert into event (dataset_id, real_time, transaction_index, causal_time, topic_index, value)
	  values ($1, $2::timestamptz - $3::timestamptz, $4, $5, $6, $7)
```

I'd be happy to talk through what all these columns mean - the names aren't great, and there's some subtlety from the sim engine involved here. Long story short, the simulation engine runs in "batches", and each batch can result in a "commit". So, the simulation timeline looks something like this:

- Delay (some duration)
- Commit (some events happened)
- Delay (some duration)
- Commit (some events happened)
- Commit (some more events happened)
- Commit (even more events happened)
- Delay (some duration)

etc...

We need to be able to describe the exact time at which an event occurred. What we do is we add up all the delays up to the time of the event. This is stored in the `real_time` column (should probably be renamed to `start_offset` column).

Now, we can have several Commits with no delays in between. So, we number them, starting from zero, and store that number in the `transaction_index` column.

Lastly, within a commit, events are tagged with a string that determines their position within the event graph. These strings are chains of numbers that look something like this: `.1.3.2` (we can go over exactly how that numbering system works separately). This tag is stored in the `causal_time` column.

PR #208, opened 2022-06-07
fix: update start time field for activity table

camargo:
## Description

- Update start time field name for activity table in default UI views


goetzrrGit (approved):
Interesting that it is `camel_case`. Looks good

camargo:
@goetzrrGit Yeah I used to alias to camel but in general I'm trying to make the UI use more of the actual DB column names we query for simplicity (note start_time is not a column name but it's derived from start_offset which is).

mattdailis (approved):

PR #191, opened 2022-05-27
[AERIE-1832] Hasura `validatePlan` action

pcrosemurgy:
* **Tickets addressed:** AERIE-1832
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description
Exposes plan validation endpoint for validating activity instances within a plan prior to simulation. Note that our `SynchronousSimulationAgent` will report the same validation errors (if any), this endpoint just allows clients to run their own validation prior to simulation.

## Verification

<details>
<summary>Expand</summary>

### Reporting Success

With a single `BiteBanana` activity inserted into a plan with ID `1` the following query:
```gql
query {
  validatePlan(planId:1) {
    success
    errors
  }
}
```
results in:
```json
{
  "data": {
    "validatePlan": {
      "success": true
    }
  }
}
```

### Reporting Failures

An activity that will result in an **unconstructable argument** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"300 seconds\", type: \"BiteBanana\", arguments: { biteSize: true } }) {
    id
  }
}
```

An activity that will result in an **extraneous parameter** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"310 seconds\", type: \"BiteBanana\", arguments: { bad: 42 } }) {
    id
  }
}
```

An activity that will result in a **missing arguments** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"320 seconds\", type: \"BakeBananaBread\", arguments: { tbSugar: 42 } }) {
    id
  }
}
```

An activity that will result in a **nonexistent activity type** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"330 seconds\", type: \"BadActivity\", arguments: { } }) {
    id
  }
}
```

To trigger plan validation a simple `validatePlan` call is needed:
```gql
query {
  validatePlan(planId:1) {
    success
    errors
  }
}
```

This results in:
```json
{
  "data": {
    "validatePlan": {
      "success": false,
      "errors": {
        "1": {
          "reason": "gov.nasa.jpl.aerie.merlin.protocol.model.TaskSpecType$UnconstructableTaskSpecException: Unconstructable argument \"biteSize\": Expected real number, got true"
        },
        "2": {
[...]

PR #211, opened 2022-06-08
[AERIE-1878] Add Hasura `constraintViolation` action

skovati:
* **Tickets addressed:** AERIE-1878
* **Review:** By commit  

### No longer a breaking change, backwards compatibility maintained

Instead of constraint violations being included with the `results` of the `simulate` query, they are now accessed through a separate API call `constraintViolations`. Implemented so that querying `constraintViolations` before a simulation has been run will not start the simulation and instead return an empty JSON object. This was done in anticipation of future work that will *only* run the simulation on the `simulate` query.

This was implemented by adding a getter method to the `SimulationService` interface that allow the `getViolations` method in `GetSimulationResultsAction` to be called without passing simulation results, thus allowing us to query for constraint violations irrespective of simulation status. This getter logic was necessary so that while `UncachedSimulationService` will just run the simulation regardless, `CachedSimulationService` will return an empty `Map` if uncached.

## Verification

Tested with the banananation model with one PeelBanana activity, and the following constraint:
```TypeScript
export default (): Constraint =>
    Real.Resource("/fruit").equal(4)
```
which is violated when `peelDirection` is "fromStem"

### Old Behavior
Query of:
```GraphQL
query OldLameQuery {
  simulate(planId: 1) {
    status
    results
    reason
  }
}
```
returns:
```JSON
{
  "data": {
    "simulate": {
      "status": "complete",
      "results": {
        "unfinishedActivities": {},
        "start": "2022-020T12:12:12.000000",
        "activities": {
          "1": {
            "parent": null,
            "arguments": {
              "peelDirection": "fromStem"
            },
            "children": [],
            "computedAttributes": "UNIT",
            "type": "PeelBanana",
            "duration": 0,
            "startTimestamp": "2022-020T15:02:31.354000"
          }
        },
        "constraints": {
          "model/no_mashed": [
            {
              "windows": [
                {
                  "start": 10219354000,
                  "end": 86400000000
                }
              ],
              "associations": {
                "resourceIds": [
                  "/fruit"
                ],
                "activityInstanceIds": []
              }
            }
          ]
        },
        "events": [
          {
            "graph": {
              "suffix": {
  ...
```
i.e. constraint violations are included in the simulation results.

### New Behavior
Query of
```GraphQL
query NewCoolQuery {
  constraintViolations(planId: 1) {
    constraintViolations
  }
}
```
before running the simulation returns:
```JSON
{
  "data": {
    "constraintViolations": {
      "constraintViolations": {}
    }
  }
}
```

and after running the simulation returns:
```JSON
{
  "data": {
    "constraintViolations": {
      "constraintViolations": {
        "model/no_mashed": [
          {
            "windows": [
              {
                "start": 10412050000,
                "end": 86400000000
              }
            ],
            "associations": {
              "resourceIds": [
                "/fruit"
              ],
              "activityInstanceIds": []
            }
          }
        ]
      }
    }
  }
}
```

## Documentation
While not explicitly violated, should update [API docs](https://github.com/NASA-AMMOS/aerie/wiki/Aerie-GraphQL-API-Software-Interface-Specification) and [Sim Results](https://github.com/NASA-AMMOS/aerie/wiki/Simulation-Results) to mention new GraphQL query.

## Future work
Work on separating simulation result API calls contributes to possibly streaming sim results.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
@@ -75,6 +75,9 @@ public final class MerlinBindings implements Plugin {
       path("resourceSamples", () -> {
         post(this::getResourceSamples);
       });
+      path("constraintViolations", () -> {
+        post(this::getConstraintViolations);
+      });
       path("refreshModelParameters", () -> {
         post(this::postRefreshModelParameters);
       });
@@ -189,6 +192,23 @@ public final class MerlinBindings implements Plugin {
       ctx.status(404).result(ResponseSerializers.serializeNoSuchPlanException(ex).toString());
     }
 }
+  private void getConstraintViolations(final Context ctx) {
+    try {
+      final var planId = parseJson(ctx.body(), hasuraPlanActionP).input().planId();
+
+      final var constraintViolations = this.simulationAction.getViolations(planId);
+
+      ctx.result(ResponseSerializers.serializeConstraintViolations(constraintViolations).toString());
+    } catch (final InvalidJsonException ex) {
+      ctx.status(400).result(ResponseSerializers.serializeInvalidJsonException(ex).toString());
+    } catch (final InvalidEntityException ex) {
+      ctx.status(400).result(ResponseSerializers.serializeInvalidEntityException(ex).toString());
+    } catch (final NoSuchPlanException ex) {
+      ctx.status(404).result(ResponseSerializers.serializeNoSuchPlanException(ex).toString());
+    } catch (final MissionModelService.NoSuchMissionModelException ex) {
+      ctx.status(404).result(ResponseSerializers.serializeNoSuchMissionModelException(ex).toString());
+    }
+  }
 
   private void validateActivityArguments(final Context ctx) {
     try {
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
@@ -240,6 +240,15 @@ public final class ResponseSerializers {
         .build();
   }
 
+  public static JsonValue serializeConstraintViolations(final Map<String, List<Violation>> violations) {
+    return Json
+        .createObjectBuilder()
+        .add("constraintViolations", serializeMap(
+            v -> serializeIterable(ResponseSerializers::serializeConstraintViolation, v),
+            violations))
+        .build();
+  }
+
   private static Map<Integer, Pair<String, ValueSchema>> topicsById(List<Triple<Integer, String, ValueSchema>> topics) {
     final Map<Integer, Pair<String, ValueSchema>> topicsById = new HashMap<>();
     for (final var topic : topics) {
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/CachedSimulationService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/CachedSimulationService.java
@@ -1,17 +1,11 @@
 package gov.nasa.jpl.aerie.merlin.server.services;
 
-import java.util.List;
-import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
 
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
-import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 import gov.nasa.jpl.aerie.merlin.server.remotes.ResultsCellRepository;
-import org.apache.commons.lang3.tuple.Pair;
 
 public record CachedSimulationService (
     SimulationAgent agent,
@@ -33,11 +27,11 @@ public record CachedSimulationService (
   }
 
   @Override
-  public <T> Optional<T> get(final PlanId planId, final RevisionData revisionData, final Function<SimulationResults, T> getter) {
+  public Optional<SimulationResults> get(final PlanId planId, final RevisionData revisionData) {
     return this.store.lookup(planId) // Only return results that have already been cached
         .map(ResultsProtocol.ReaderRole::get)
         .map(state -> state instanceof final ResultsProtocol.State.Success s ?
-            getter.apply(s.results()) :
+            s.results() :
             null);
   }
 }
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.tuple.Pair;
 
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -64,7 +65,7 @@ public final class GetSimulationResultsAction {
       return new Response.Failed(r.reason());
     } else if (response instanceof ResultsProtocol.State.Success r) {
       final var results = r.results();
-      final var violations = getViolations(planId, results);
+      final var violations = getViolations(planId);
 
       return new Response.Complete(results, violations);
     } else {
@@ -76,13 +77,17 @@ public final class GetSimulationResultsAction {
   throws NoSuchPlanException
   {
     final var revisionData = this.planService.getPlanRevisionData(planId);
-    return simulationService.get(planId, revisionData, r -> r.resourceSamples).orElseGet(Map::of);
+    return this.simulationService
+        .get(planId, revisionData)
+        .map(r -> r.resourceSamples)
+        .orElseGet(Collections::emptyMap);
   }
 
-  public Map<String, List<Violation>> getViolations(final PlanId planId, final SimulationResults results)
+  public Map<String, List<Violation>> getViolations(final PlanId planId)
   throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException
   {
     final var plan = this.planService.getPlan(planId);
+    final var revisionData = this.planService.getPlanRevisionData(planId);
 
     final var constraintJsons = new HashMap<String, Constraint>();
 
@@ -97,9 +102,13 @@ public final class GetSimulationResultsAction {
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
+    final var results$ = this.simulationService.get(planId, revisionData);
 
     final var activities = new ArrayList<ActivityInstance>();
-    for (final var entry : results.simulatedActivities.entrySet()) {
+    final var simulatedActivities = results$
+        .map(r -> r.simulatedActivities)
+        .orElseGet(Collections::emptyMap);
+    for (final var entry : simulatedActivities.entrySet()) {
       final var id = entry.getKey();
       final var activity = entry.getValue();
 
@@ -113,9 +122,11 @@ public final class GetSimulationResultsAction {
           activity.arguments(),
           Window.between(activityOffset, activityOffset.plus(activity.duration()))));
     }
-
-    final var discreteProfiles = new HashMap<String, DiscreteProfile>(results.discreteProfiles.size());
-    for (final var entry : results.discreteProfiles.entrySet()) {
+    final var _discreteProfiles = results$
+        .map(r -> r.discreteProfiles)
+        .orElseGet(Collections::emptyMap);
+    final var discreteProfiles = new HashMap<String, DiscreteProfile>(_discreteProfiles.size());
+    for (final var entry : _discreteProfiles.entrySet()) {
       final var pieces = new ArrayList<DiscreteProfilePiece>(entry.getValue().getRight().size());
 
       var elapsed = Duration.ZERO;
@@ -132,9 +143,11 @@ public final class GetSimulationResultsAction {
 
       discreteProfiles.put(entry.getKey(), new DiscreteProfile(pieces));
     }
-
+    final var _realProfiles = results$
+        .map(r -> r.realProfiles)
+        .orElseGet(Collections::emptyMap);
     final var realProfiles = new HashMap<String, LinearProfile>();
-    for (final var entry : results.realProfiles.entrySet()) {
+    for (final var entry : _realProfiles.entrySet()) {
       final var pieces = new ArrayList<LinearProfilePiece>(entry.getValue().size());
 
       var elapsed = Duration.ZERO;
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SimulationService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SimulationService.java
@@ -1,7 +1,6 @@
 package gov.nasa.jpl.aerie.merlin.server.services;
 
 import java.util.Optional;
-import java.util.function.Function;
 
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
@@ -9,5 +8,5 @@ import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 
 public interface SimulationService {
   ResultsProtocol.State getSimulationResults(PlanId planId, RevisionData revisionData);
-  <T> Optional<T> get(PlanId planId, RevisionData revisionData, Function<SimulationResults, T> getter);
+  Optional<SimulationResults> get(PlanId planId, RevisionData revisionData);
 }
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -99,6 +99,12 @@ type Query {
 }
 
 type Query {
+  constraintViolations(
+    planId: Int!
+  ): ConstraintViolationsResponse
+}
+
+type Query {
   validateActivityArguments(
     activityTypeName: String!
     missionModelId: ID!
@@ -255,11 +261,17 @@ type ResourceSamplesResponse {
   resourceSamples: ResourceSamples!
 }
 
+type ConstraintViolationsResponse {
+  constraintViolations: ConstraintViolations!
+}
+
 scalar ResourceSchema
 
 scalar MerlinSimulationResults
 
 scalar ResourceSamples
+
+scalar ConstraintViolations
 
 scalar MerlinSimulationFailureReason
 
--- a/deployment/hasura/metadata/actions.yaml
+++ b/deployment/hasura/metadata/actions.yaml
@@ -63,6 +63,10 @@ actions:
     definition:
       kind: ""
       handler: http://aerie_merlin:27183/resourceSamples
+  - name: constraintViolations
+    definition:
+      kind: ""
+      handler: http://aerie_merlin:27183/constraintViolations
   - name: validateActivityArguments
     definition:
       kind: ""
@@ -137,10 +141,12 @@ custom_types:
     - name: UserCodeError
     - name: CodeLocation
     - name: ResourceSamplesResponse
+    - name: ConstraintViolationsResponse
   scalars:
     - name: ResourceSchema
     - name: MerlinSimulationResults
     - name: ResourceSamples
+    - name: ConstraintViolations
     - name: MerlinSimulationFailureReason
     - name: ModelArguments
     - name: ActivityArguments
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/UncachedSimulationService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/UncachedSimulationService.java
@@ -1,18 +1,12 @@
 package gov.nasa.jpl.aerie.merlin.server.services;
 
-import java.util.List;
-import java.util.Map;
 import java.util.Optional;
-import java.util.function.Function;
 
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
-import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.merlin.server.mocks.InMemoryRevisionData;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 import gov.nasa.jpl.aerie.merlin.server.remotes.InMemoryResultsCellRepository.InMemoryCell;
-import org.apache.commons.lang3.tuple.Pair;
 
 public record UncachedSimulationService (
     SimulationAgent agent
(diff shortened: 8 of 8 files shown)


skovati:
Working to refactor with new generic interface method introduced by [this PR](https://github.com/NASA-AMMOS/aerie/pull/191)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SimulationService.java
```diff
@@ -3,13 +3,20 @@
...
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 import org.apache.commons.lang3.tuple.Pair;
 
 public interface SimulationService {
   ResultsProtocol.State getSimulationResults(PlanId planId, RevisionData revisionData);
   Map<String, List<Pair<Duration, SerializedValue>>> getResourceSamples(PlanId planId, RevisionData revisionData);
+  Map<ActivityInstanceId, SimulatedActivity> getSimulatedActivities(PlanId planId, RevisionData revisionData);
+  Map<String, Pair<ValueSchema, List<Pair<Duration, SerializedValue>>>> getDiscreteProfiles(PlanId planId, RevisionData revisionData);
+  Map<String, List<Pair<Duration, RealDynamics>>> getRealProfiles(PlanId planId, RevisionData revisionData);
```
pcrosemurgy: Looks like you may need to rebase with `develop`. Maybe you have already and it's just not pushed :)

pcrosemurgy: Oops, sent too soon. Thanks!

skovati:
Refactored to use new generic getter. Looking for additional feedback!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -132,9 +135,10 @@ public Map<String, List<Violation>> getViolations(final PlanId planId, final Sim
 
       discreteProfiles.put(entry.getKey(), new DiscreteProfile(pieces));
     }
-
+    final var _realProfiles = this.simulationService.get(planId, revisionData,
+        r -> r.realProfiles).orElseGet(Map::of);
     final var realProfiles = new HashMap<String, LinearProfile>();
-    for (final var entry : results.realProfiles.entrySet()) {
+    for (final var entry : _realProfiles.entrySet()) {
```
pcrosemurgy, line 150: Very minor naming convention thing but in other areas of aerie we usually use a `$` prefix (for better or worse) instead of a `_`.

pcrosemurgy (approved):
Great job!

@camargo does the UI make use of constraint violations from the sim. results HTTP response? I did a quick check and I don't think it does.

camargo:
@pcrosemurgy Yes it does (see: https://github.com/NASA-AMMOS/aerie-ui/blob/develop/src/utilities/effects.ts#L988). The UI will need to be updated to use this new action.

pcrosemurgy:
Ohh. In that case @skovati I think we'll need to not remove the following line:
```java
.add("constraints", serializeMap(v -> serializeIterable(ResponseSerializers::serializeConstraintViolation, v), violations))
```
In a follow-up PR we can remove this once the UI has been updated to use this action.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
Twisol: Usually a `suffix$`, as I recall. The `$` is meant to indicate that it's a transparent container of something, so you might see `Optional<Foo> foo$`. It avoids either taking up a valuable name for an unwrapped `Foo foo` or belaboring the point with `optionalFoo`.

The distinction between `realProfiles` and `_realProfiles` here is somwhat different (i.e. one isn't an unwrapped version of the other), so I'm personally not fussed about switching to `$`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -240,6 +239,15 @@ public static JsonValue serializeResourceSamples(final Map<String, List<Pair<Dur
         .build();
   }
 
+  public static JsonValue serializeConstraintViolations(final Map<String, List<Violation>> violations) {
+    return Json
+      .createObjectBuilder()
+      .add("constraintViolations", serializeMap(v ->
+          serializeIterable(ResponseSerializers::serializeConstraintViolation, v),
+          violations))
```
Twisol: It's a little hard to see from the formatting that the `v ->` on the first line only scopes the second line and not the third line. Maybe shift the `v ->` down a line so the whole lambda is in one line?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -97,9 +98,10 @@ public Map<String, List<Violation>> getViolations(final PlanId planId, final Sim
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
-
     final var activities = new ArrayList<ActivityInstance>();
-    for (final var entry : results.simulatedActivities.entrySet()) {
+    final var simulatedActivities = this.simulationService.get(planId, revisionData,
+        r -> r.simulatedActivities).orElseGet(Map::of);
```
Twisol: This formatting splits a logical method call and simultaneously groups half of it with a separate method call. Maybe something like this makes the logic flow more clearly?

```suggestion
    final var simulatedActivities = this.simulationService
        .get(planId, revisionData, r -> r.simulatedActivities)
        .orElseGet(Map::of);
```

(This happens in a few other places in this diff.)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Ah sure thing, I'll push that change!

skovati: Good point @Twisol, I'll stick with the `_` prefix barring any other input

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
skovati: Totally agree, change incoming!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Good call, latter is much clearer

Twisol: I have a personal preference for `Collections::emptyMap` over `Map::of`, since the latter has overloads of multiple arity and you have to recognize that the nullary overload is being used here. (My first question when I see `Map::of` is "of what?", and it takes a half-second to realize there's nothing coming in from the caller.)

No big deal though.

skovati: Good point, I considered using `.orElse(Map.of())` for that same clarity reason, but `Collections::emptyMap` is even better!

Twisol: `Map.of()` is cheap, so `orElse` would have been fine -- but `Collections::emptyMap` exists ^_^

skovati: Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
skovati: Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Switched to `Collections::emptyMap`.

skovati:
@camargo @pcrosemurgy
Branch has been updated to maintain backwards compatibility.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -189,6 +192,23 @@ private void getResourceSamples(final Context ctx) {
       ctx.status(404).result(ResponseSerializers.serializeNoSuchPlanException(ex).toString());
     }
 }
+  private void getConstraintViolations(final Context ctx) {
+    try {
+      final var planId = parseJson(ctx.body(), hasuraPlanActionP).input().planId();
+
+      final var constraintViolations = this.simulationAction.getViolations(planId);
```
mattdailis, line 199: These calls to `getViolations` and `serializeConstraintViolations` show up in an earlier commit than the definitions of those methods do, so there are a couple commits at the beginning of this PR that don't compile.

I tried reversing the order of all four commits in this PR, and they all built nicely, so if that's okay with you, I might request a reversal of the commit order.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -97,9 +99,11 @@ public Map<String, List<Violation>> getViolations(final PlanId planId, final Sim
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
-
     final var activities = new ArrayList<ActivityInstance>();
-    for (final var entry : results.simulatedActivities.entrySet()) {
+    final var simulatedActivities = this.simulationService
+      .get(planId, revisionData, r -> r.simulatedActivities)
```
mattdailis: I'm noticing three calls to `simulationService.get` - any thoughts on the tradeoffs between keeping them separate, vs merging them into one call (and then pulling `simulatedActivities`, `discreteProfiles`, and `realProfiles` out afterwards)?

Twisol: Making one call also goes well with my note on #191 -- you can just pass `$ -> $` (or its better named equivalent, `Function.identity()`) to get an `Optional<SimulationResults>`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
skovati: Good call, I was doing some bespoke rebasing on my branch and didn't realize I ended up with invalid commits. Reversing the order works perfectly.

skovati: Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: I do like the idea of condensing to one function call. My first attempt at this looks like
```java
final var results = this.simulationService
    .get(planId, revisionData, Function.identity());
```
followed by
```java
final var simulatedActivities = results
    .map(r -> r.simulatedActivities)
    .orElseGet(Collections::emptyMap);
```
for each field.

Is there a simpler way of extracting a field from an `Optional` that I'm overlooking?

pcrosemurgy: Sorry for the churn @skovati, sloppy thinking on my part when I added `SimulationService.get()`. If you want you can change it to:
```java
Optional<SimulationResults> get(final PlanId planId, final RevisionData revisionData);
```
(Thanks @Twisol)
Or use the `Function.identity()` approach you have above.

Then in `getViolations()` I would just call that once, with something like: `final var results$ = simulationService.get(planId, revisionData);`.

skovati: A-ok with either option, changing `SimulationService.get()` would require a slight change to your work for [1832](https://github.com/NASA-AMMOS/aerie/pull/191) @pcrosemurgy. Happy to make that change, just let me know your preference.

pcrosemurgy: No strong opinion :) Whatever you see fit is good with me

skovati: I'll stick with the generic `get` function in that case, barring additional input, since otherwise we'd have two methods in `SimulationService` that both return slightly different forms of sim results.

pcrosemurgy: Well, they both accomplish the same thing so you'd only want 1 either way :)
```java
<T> Optional<T> get(final PlanId planId, final RevisionData revisionData, final Function<SimulationResults, T> getter);
```
or
```java
Optional<SimulationResults> get(final PlanId planId, final RevisionData revisionData);
```

So using the first form you can do:
```java
final var simResults$ = simulationService.get(planId, revisionData, Function.identity());
final var simulatedActs = simResults$.map(r -> r.simulatedActivities).orElseGet(Map::of);
```

Using the second form you can return the same data with a slightly simpler call:
```java
final var simResults$ = simulationService.get(planId, revisionData);
final var simulatedActs = simResults$.map(r -> r.simulatedActivities).orElseGet(Map::of);
...
```

**TL;DR** either is really fine, sorry for causing confusion here with my weird `get()` interface.

skovati: er yeah, I was referencing the already existing `getSimulationResults` in `SimulationService`, so changing `get()` to only return `SimulationResults` felt a little redundant (of course it isn't, since the whole point is we need this optional logic for uncached sims) while the current generic implementation offers slightly more functionality

Twisol: > er yeah, I was referencing the already existing `getSimulationResults` in `SimulationService`, so changing `get()` to only return `SimulationResults` felt a little redundant (of course it isn't, since the whole point is we need this optional logic for uncached sims) while the current generic implementation offers slightly more functionality

It sounds like you'll have two methods that return slightly different forms of sim results no matter what your choice here is, no? If they serve different use-cases, we ought to allow them to evolve to serve those use-cases effectively. If we eventually find we can unify them, that's something we can do later. If we eventually find they're impossible to unify, we accept that we have two ways of getting sim results and that there's a fundamental difference between them.

Whether you apply your getter to the return value yourself, or let the method call your getter for you, does not seem to be materially relevant to the fact that there are two ways to get results. My vote would be to remove the getter callback, because it doesn't serve the use-case any better than using Optional's own map method does.

Twisol: More meaningfully, the getter argument entices you into calling get() multiple times with different getters, doing extra work to get the same initial simulation results every time. From a "pit of success" perspective, removing the getter argument avoids this enticement and makes the right thing the default.

pcrosemurgy: >It sounds like you'll have two methods that return slightly different forms of sim results no matter what your choice here is, no? If they serve different use-cases, we ought to allow them to evolve to serve those use-cases effectively. 

Sorry if I'm misunderstanding what you're saying but it sounds like you're saying we have the need for both forms of `get()`? 

Looking at where `SimulationService.get()` is used we only need one form (the `Optional<SimulationResults>` return form being the most ergonomic as you point out).

Twisol: No, I was responding to @skovati's note that get() and getSimulationResults() both get simulation results. They serve different use cases, so avoiding similarity for similarity's sake prevents them from evolving effectively to meet their respective use cases.

pcrosemurgy: Ahh, yeah 100% agree :)

skovati: Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -97,9 +100,13 @@ public Map<String, List<Violation>> getViolations(final PlanId planId, final Sim
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
+    final var _results = this.simulationService.get(planId, revisionData, Function.identity());
```
Twisol: Now *this* is where I'd use `results$` 😉

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Whoops, pushed changes before I saw your input @Twisol, happy to make that change

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Change has been made!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Alright nongeneric version has been pushed!

Twisol: Looks great! I think we've addressed Matt's initial concern here, too.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/CachedSimulationService.java
```diff
@@ -33,11 +27,11 @@ public ResultsProtocol.State getSimulationResults(final PlanId planId, final Rev
   }
 
   @Override
-  public <T> Optional<T> get(final PlanId planId, final RevisionData revisionData, final Function<SimulationResults, T> getter) {
+  public Optional<SimulationResults> get(final PlanId planId, final RevisionData revisionData) {
     return this.store.lookup(planId) // Only return results that have already been cached
         .map(ResultsProtocol.ReaderRole::get)
         .map(state -> state instanceof final ResultsProtocol.State.Success s ?
-            getter.apply(s.results()) :
+            s.results() :
```
pranav-super, line 34: Why did you make this change?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -113,9 +122,11 @@ public Map<String, List<Violation>> getViolations(final PlanId planId, final Sim
           activity.arguments(),
           Window.between(activityOffset, activityOffset.plus(activity.duration()))));
     }
-
-    final var discreteProfiles = new HashMap<String, DiscreteProfile>(results.discreteProfiles.size());
-    for (final var entry : results.discreteProfiles.entrySet()) {
+    final var _discreteProfiles = results$
```
pranav-super, line 125: Can you justify this change?

pranav-super (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/CachedSimulationService.java
skovati: To sum up my conversation above, I moved the getter callback outside of this `get()` function, since to get the full results object out of this getter would require a somewhat awkward
```java
final var results = simulationService
    .get(planId, revisionData, Function.identity())
    .orElseGet(Collections::emptyMap);
```
so now, instead of calling
```java
final var simActs = simulationService
    .get(planId, revisionData, r -> r.simulatedActivities)
    .orElseGet(Collections::emptyMap);
```
you can use
```java
final var simActs = simulationService
    .get(planId, revisionData)
    .map(r -> r.simulatedActivities)
    .orElseGet(Collections::emptyMap);
```
which allows easy extraction of the full `SimulationResults` object if needed.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: Essentially, it was necessary to remove `getViolations()`'s dependence on the `SimulationResults` argument, since we want to call this method without having run the simulation, so that we can return an empty violation map if the sim was uncached but `CachedSimulationService` was being used. Instead, we are getting an `Optional<SimulationResults` object from the new getter in the `SimulationService` interface.

skovati:
rebased onto upstream develop

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
mattdailis: One thought I have is that in maintaining backwards compatibility, we now load simulation results twice (in the case where `getViolations` is called from `GetSimulationResultsAction::run`). If we weren't planning on getting rid of that call, I would make a fuss about it, but as it stands I'm okay with leaving that in here until we remove that call in a follow-up PR :+1:

mattdailis (approved):

mattdailis:
Great work, Luke! 🎉

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
skovati: That is a good point. The backwards compatibility was a bit of an afterthought. I suppose the double loading is somewhat inherent to allowing these queries to be made separately though.

mattdailis: One way to "shim in" the backwards compatibility, without loading the data twice, would be to have two `getViolations`methods, one with one parameter `getViolations(PlanId)`, and the other with two parameters `getViolations(PlanId, SimulationResults)`. The first would load the results and call the second, whereas the "old" workflow would continue more or less unchanged.

If we wanted to keep both, I would push for this approach, but since we're removing one, this would add some unnecessary clutter.

skovati: I agree it would add some unnecessary clutter, especially since the removal of the deprecated API *should* happen within the same release (ie 0.12.1). I'll leave it as is then, barring additional feedback!

skovati:
rebased onto upstream develop

PR #213, opened 2022-06-09
[AERIE 1885] Add enabled/disabled to goal in sceduling specification

patkenneally:
* **Tickets addressed:** AERIE-1885
* **Review:** By commit 

## Description
This PR adds the capability to toggle a scheduling goal within a scheduling specification as either enabled or disabled. A disabled scheduling goal is ignored during scheduling execution. 

## Verification
A test is added `testSingleActivityPlanGoalDisabled()` in `SchedulingIntegrationTests.java`. Also tested through the UI by adding three goals, disabling the second goal directly via the db, observed correct behavior. 

## Documentation
Two sentence have been added to https://github.com/NASA-AMMOS/aerie/wiki/Scheduling-Guide#specifying-the-order-of-goals noting the effect of enabled/disabled goals.

## Future work
There have been discussions around how the ordering/priority of goals is implemented in the UI. These discussions are some ways off resulting in tickets, so for now there are no clear next steps resulting from this PR.

### DIFF
--- a/scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
@@ -6,6 +6,7 @@ create table scheduling_specification_goals (
     default null -- Nulls are detected and replaced with the next
                  -- available priority by the insert trigger
     constraint non_negative_specification_goal_priority check (priority >= 0),
+  enabled boolean default true,
 
   constraint scheduling_specification_goals_primary_key
     primary key (specification_id, goal_id),
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
@@ -16,6 +16,7 @@ import java.util.List;
             s.specification_id,
             s.goal_id,
             s.priority,
+            s.enabled,
             g.name,
             g.definition,
             g.revision
@@ -26,7 +27,8 @@ import java.util.List;
       g.goal_id,
       g.name,
       g.definition,
-      g.revision
+      g.revision,
+      g.enabled
     from goals as g
       where g.specification_id = ?
       order by g.priority asc
@@ -48,7 +50,8 @@ import java.util.List;
       final var revision = resultSet.getLong("revision");
       final var name = resultSet.getString("name");
       final var definition = resultSet.getString("definition");
-      goals.add(new PostgresGoalRecord(id, revision, name, definition));
+      final var enabled = resultSet.getBoolean("enabled");
+      goals.add(new PostgresGoalRecord(id, revision, name, definition, enabled));
     }
 
     return goals;
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/GoalRecord.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/GoalRecord.java
@@ -1,3 +1,3 @@
 package gov.nasa.jpl.aerie.scheduler.server.models;
 
-public final record GoalRecord(GoalId id, SchedulingDSL.GoalSpecifier definition) {}
+public record GoalRecord(GoalId id, SchedulingDSL.GoalSpecifier definition, boolean enabled) {}
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresGoalRecord.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresGoalRecord.java
@@ -4,5 +4,6 @@ public record PostgresGoalRecord(
     long id,
     long revision,
     String name,
-    String definition
+    String definition,
+    boolean enabled
 ) {}
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
@@ -105,7 +105,7 @@ public final class PostgresSpecificationRepository implements SpecificationRepos
     if (goalCompilationResult instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Error g) {
       return new GoalCompilationResult.Failure(g.errors());
     } else if (goalCompilationResult instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success g) {
-      return new GoalCompilationResult.Success(new GoalRecord(new GoalId(pgGoal.id()), g.goalSpecifier()));
+      return new GoalCompilationResult.Success(new GoalRecord(new GoalId(pgGoal.id()), g.goalSpecifier(), pgGoal.enabled()));
     } else {
       throw new Error("Unhandled variant of SchedulingDSLCompilationResult: " + goalCompilationResult);
     }
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java
@@ -51,6 +51,7 @@ public final class PostgresSpecificationRepository implements SpecificationRepos
 
     final var goals = postgresGoalRecords
         .stream()
+        .filter(PostgresGoalRecord::enabled)
         .map((PostgresGoalRecord pgGoal) -> compileGoalDefinition(
             missionModelService,
             planId,
--- a/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
+++ b/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
@@ -43,6 +43,8 @@ public class SchedulingIntegrationTests {
 
   private record MissionModelDescription(String name, Map<String, SerializedValue> config, Path libPath) {}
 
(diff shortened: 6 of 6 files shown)


pcrosemurgy (approved):
This seems sane & innocuous to me!

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
```diff
@@ -3,10 +3,12 @@ create table scheduling_specification_goals (
   goal_id integer not null,
   priority integer
     not null
-    default null -- Nulls are detected and replaced with the next
+    default null, -- Nulls are detected and replaced with the next
                  -- available priority by the insert trigger
-    constraint non_negative_specification_goal_priority check (priority >= 0),
+  enabled boolean default true,
 
+  constraint non_negative_specification_goal_priority
```
mattdailis: 🔵  Is there a semantic difference between attaching a constraint to a column, vs having the constraint at the top level of the table? Or is that merely a matter of taste?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
```diff
@@ -48,7 +50,8 @@ public List<PostgresGoalRecord> get(final long specificationId) throws SQLExcept
       final var revision = resultSet.getLong("revision");
       final var name = resultSet.getString("name");
       final var definition = resultSet.getString("definition");
-      goals.add(new PostgresGoalRecord(id, revision, name, definition));
+      final var enabled = resultSet.getBoolean("enabled");
```
mattdailis, line 53: 🔶 I'd like to chat through the pros and cons of filtering on the `enabled` field in the `SynchronousSchedulerAgent` vs at other stages in the pipeline.

The options, as I see them are:
1. Filter in the `SynchronousSchedulerAgent` (as we have in this PR)
2. Filter in the `PostgresSpecificationRepository` (by adding a `filter` line before the `map` line [here](https://github.com/NASA-AMMOS/aerie/blob/8cdb2e757f489502647f54ee3510594360f57155/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java#L52-L59))
3. Filter in the `GetSpecificationGoalsAction`, by updating the SQL query to only return enabled goals

### Option 1: Filter in `SynchronousSchedulerAgent`
Pros: 
- The data in the database is faithfully represented all the way through the workflow
- The "business logic" i.e. interpretation of the `enabled` field happens in the scheduler itself

Cons:
- Type errors or runtime errors in the scheduling-dsl-compiler will fail the scheduling run, regardless of whether the errors originated in an enabled or a disabled goal (this could also be addressed by separating the compilation workflow from the scheduling run workflow, but currently the two are coupled together)
  - (I put this as a Con because I think it's undesirable, but I acknowledge that this can be up for debate)
- I question, should `enabled` really be the concern of the scheduler? Is there anything the scheduler would ever do with a disabled goal, besides ignore it?

### Option 2: Filter in `PostgresSpecificationRepository`
Pros:
- The `PostgresGoalRecord` is still a faithful representation of the data in the database
- Disabled Goals will not be compiled, and thus cannot cause the scheduling run to fail

Cons:
- The enabled field is "hidden" from the scheduler (but I think this might actually be a "Pro")

### Option 3: Filter in SQL
Pros:
- Only the `GetSpecificationGoalsAction` and the table itself have to change - the rest of the code can be oblivious to the `enabled` field

Cons:
- There is no longer a one-to-one mapping between `PostgresGoalRecord` and a row in the spec goal table (though, arguably, there never was such a mapping, since the information for `PostgresGoalRecord` is the result of a join between two tables).

I'd be curious to hear your thoughts - the reason I bring this up is I initially picked Option 3 when thinking through this design, but I see you've picked Option 1, and now I'm leaning towards Option 2 as a middle ground 😅.

mattdailis (commented):

mattdailis (approved):
None of my concerns should block merge, especially if @camargo is planning to start work on this. Should we choose to refactor, the behavior from his perspective will remain unchanged.

patkenneally:
First off, thank you for your thoughtfulness @mattdailis !

I had some similar thoughts as well. I think the solution lies somewhere between option 1 and option 2. A refactoring maybe in order.

It would seem that PostgresSpecificationRepository#getSpecification should return the full specification, disabled goals and all. Previously, it made sense that the goals were being compiled in `getSpecification` because all goals in the specification were used in a scheduling execution. It also makes sense that the goal's compilation is orchestrated by the PostgresSpecificationRepository as the output type `Specification` only makes sense with the Java representation of goals. With the addition of enabled/disabled, we'd like to not compile disabled goals (as you identified). 

Okay so I started writing thinking the answer lay between options 1 and 2, but now I'm veering towards option 2 as being the better solution. Right now the system has need to retrieve a `Specification` with only enabled goals, therefore not compiling disabled goals. If a future use case arises that needs the full `Specification`, we can refactor.  

Summary 
Option 1: con, Unnecessarily compiles disabled goals
Option 2: pro, Disabled goals not compiled and returns a full `Specification` from the perspective of the scheduler (given our current use cases/needs)
Option 3: con, The `GetSpecificationGoalsAction` should return all goals rather than filtering them out. I think the PostgresSpecificationRepository  should filter them out.

Again, thank you for your thoughtful response.

Twisol:
> Previously, it made sense that the goals were being compiled in `getSpecification` because all goals in the specification were used in a scheduling execution. It also makes sense that the goal's compilation is orchestrated by the PostgresSpecificationRepository as the output type `Specification` only makes sense with the Java representation of goals.

Am I reading correctly that the database driver is *also* responsible for compiling scheduling TypeScript into Java rules? 🤔

scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql
patkenneally: I'll have to go, dig up the particulars, and educate myself about the answer to that question. In any case, I shouldn't have been making such an unrelated change, in the schema no less. Touching the schema like that will soon be an even bigger no-no when we start capturing our schema migrations. 

I've undone the change.

patkenneally:
The PostgresSpecificationRepository currently makes the call to compile the goals, with the `Specification` being returned to the Scheduler. [PostgresSpecificationRepository Line 54](https://github.com/NASA-AMMOS/aerie/blob/6ffca5fa1d4d64d71327f3136b5022e4b9128879/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresSpecificationRepository.java#L54) 

I could see that a refactor could separate this out to a different entity so that neither the `SynchronousSchedulerAgent`, nor the `PostgresSpecificationRepository` would be involved.

mattdailis:
> Am I reading correctly that the database driver is _also_ responsible for compiling scheduling TypeScript into Java rules? 🤔

Yes, and I don't love it either 😅 Originally this decision was made because the Specification was defined in terms of the scheduler's `Goal` type, so in order to create a `Specification` object, you needed to turn the typescript strings into `Goal` objects. I would love to decouple that, but I'm having trouble envisioning exactly how.

mattdailis:
In some ways, the `Specification` doesn't need to be concerned with the contents of the goals - it merely needs to refer to those goals (whether by id, or by reference to a java object, perhaps hidden behind an interface).

We could have the specification store a list of thunks, `() -> Goal`, (or, in what I think is a morally equivalent way, a list of `GoalIds`) which would allow us to defer compilation until after a specification has been created, without the specification needing to know that the goals are represented as typescript.

camargo (approved):

Twisol:
> without the specification needing to know that the goals are represented as typescript.

:thinking: I think the domain model is a little thin here, specifically regarding the relationship between "specification" and "goal", and what I think is an unnamed concept "goal source". Because the scheduler wants to work with concrete goals, and there's nowhere else between the scheduler and the repository to turn goal source into concrete goals, the responsibility for performing this interpretation has fallen to the repository. (Let me know if I've misunderstood the domain here.)

It seems to me that the database should be returning the source-level data, and whoever receives it should then pass the goal source on to some other component to turn the specification into goals, before then passing the goals on to whoever needed them. At a very abstract level, this could be done by dependency injection, giving whoever calls the repository a component that knows how to interpret goal sources.

Put slightly differently, it's not a Postgres abstraction that goals are stored as typescript -- it's a high-level design decision. Individual components may not need to know about this, but that decision *does* need to be located somewhere. Imagine, for the sake of probing the architecture, if we used the filesystem instead of Postgres to store TypeScript source -- would we really want to duplicate that code between multiple database drivers, or couple two drivers to the same post-processing concern? This is really the remit of the caller to interpret the stored data.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationGoalsAction.java
patkenneally: I have made a change which operates as Option 2. I have also taken a note for us to discuss and triage any possible refactoring (as discussed below with @mattdailis and @Twisol ).

Twisol: > * I question, should `enabled` really be the concern of the scheduler?

yes

> Is there anything the scheduler would ever do with a disabled goal, besides ignore it?

It could tell you which goals it skipped because you asked it not to run them. More specifically, it could tell you which goals it could have applied had you not inhibited it.

Twisol: this would be useful specifically because it gives users *actionable feedback* if they're not happy with the results of the current scheduling run.

patkenneally: @Twisol do you have any objections to this PR being merged?  I ask because I want to make sure I'm giving due time and recognition to the discussions being had here, and that we come to a conclusion, even if that's to create a ticket to take up these improvements next.

Twisol: Sorry for the delay (internet trouble), and sorry for the ambiguity. I have no problem with the PR itself; the thing I'm poking at seems to be preexisting.

PR #218, opened 2022-06-13
[AERIE-1766] Add healthcheck endpoints

skovati:
* **Tickets addressed:** AERIE-1766
* **Review:** By commit  

## Description
Relatively simple extension of existing HTTP servers to include a `/health` endpoint. Note that this is only for `GET` HTTP requests, in contrast to the `POST` requests that are commonly used throughout Aerie.

For merlin-worker, an entirely new HTTP server had to be added, hence the addition of new exposed ports for workers in `docker-compose.yml`.

## Verification

Added several e2e tests that hit new `/health` endpoints

`curl -i http://0.0.0.0:2718{3,4,5,7,8}/health`
now returns
```http
HTTP/1.1 200 OK
Date: Tue, 14 Jun 2022 21:54:55 GMT
Content-Type: application/json
Content-Length: 0

HTTP/1.1 200 OK
X-Powered-By: Express
Date: Tue, 14 Jun 2022 21:54:55 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Content-Length: 0

HTTP/1.1 200 OK
Date: Tue, 14 Jun 2022 21:54:55 GMT
Content-Type: application/json
Content-Length: 0

HTTP/1.1 200 OK
Date: Tue, 14 Jun 2022 21:54:55 GMT
Content-Type: text/plain
Content-Length: 0

HTTP/1.1 200 OK
Date: Tue, 14 Jun 2022 21:54:55 GMT
Content-Type: text/plain
Content-Length: 0
```

## Documentation

No invalidated documentation. Can update existing API documentation to make note of new endpoint.

## Future work

Automatically hit these new health checks
*Possibly* add introspective logic to health checks

### DIFF
--- a/merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
+++ b/merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
@@ -16,6 +16,7 @@ import gov.nasa.jpl.aerie.merlin.server.services.LocalPlanService;
 import gov.nasa.jpl.aerie.merlin.server.services.SynchronousSimulationAgent;
 import gov.nasa.jpl.aerie.merlin.server.services.UnexpectedSubtypeError;
 import gov.nasa.jpl.aerie.merlin.worker.postgres.PostgresSimulationNotificationPayload;
+import io.javalin.Javalin;
 
 import java.nio.file.Path;
 import java.util.Optional;
@@ -56,6 +57,9 @@ public final class MerlinWorkerAppDriver {
         new LinkedBlockingQueue<>();
     final var listenAction = new ListenSimulationCapability(hikariDataSource, notificationQueue);
     listenAction.registerListener();
+
+    final var app = Javalin.create().start(8080);
+    app.get("/health", ctx -> ctx.status(200));
 
     while (true) {
       final var notification = notificationQueue.take();
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -100,6 +100,10 @@ app.get('/', (_: Request, res: Response) => {
   res.send('Aerie Command Service');
 });
 
+app.get('/health', (_: Request, res: Response) => {
+  res.status(200).send();
+});
+
 app.post('/put-dictionary', async (req, res, next) => {
   const dictionary = req.body.input.dictionary as string;
   logger.info(`Dictionary received`);
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -131,7 +131,7 @@ services:
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
         -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=INFO
         -Dorg.slf4j.simpleLogger.logFile=System.err
-    ports: ["5007:5005"]
+    ports: ["5007:5005", "27187:8080"]
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store:ro
@@ -153,7 +153,7 @@ services:
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
         -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=INFO
         -Dorg.slf4j.simpleLogger.logFile=System.err
-    ports: ["5008:5005"]
+    ports: ["5008:5005", "27188:8080"]
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store:ro
--- a/e2e-tests/docker-compose-test.yml
+++ b/e2e-tests/docker-compose-test.yml
@@ -129,7 +129,7 @@ services:
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
         -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=INFO
         -Dorg.slf4j.simpleLogger.logFile=System.err
-    ports: ['5007:5005']
+    ports: ['5007:5005', '27187:8080']
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store:ro
@@ -150,7 +150,7 @@ services:
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
         -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=INFO
         -Dorg.slf4j.simpleLogger.logFile=System.err
-    ports: ['5008:5005']
+    ports: ['5008:5005', '27188:8080']
     restart: always
     volumes:
       - aerie_file_store:/usr/src/app/merlin_file_store:ro
--- a/e2e-tests/src/tests/health.test.ts
+++ b/e2e-tests/src/tests/health.test.ts
@@ -16,4 +16,24 @@ test.describe('Health', () => {
     const healthy = await req.healthUI(request);
     expect(healthy).toBeTruthy();
   });
+
+  test('Merlin is healthy', async ({ request }) => {
+    const healthy = await req.healthMerlin(request);
+    expect(healthy).toBeTruthy();
+  });
+
+  test('Commanding is healthy', async ({ request }) => {
+    const healthy = await req.healthCommanding(request);
+    expect(healthy).toBeTruthy();
+  });
+
+  test('Scheduler is healthy', async ({ request }) => {
+    const healthy = await req.healthScheduler(request);
+    expect(healthy).toBeTruthy();
+  });
+
+  test('Worker is healthy', async ({ request }) => {
+    const healthy = await req.healthWorker(request);
+    expect(healthy).toBeTruthy();
+  });
 });
--- a/e2e-tests/src/utilities/requests.ts
+++ b/e2e-tests/src/utilities/requests.ts
@@ -2,7 +2,7 @@ import type { APIRequestContext } from '@playwright/test';
 import { sync as glob } from 'fast-glob';
 import { createReadStream } from 'fs';
 import { basename, resolve } from 'path';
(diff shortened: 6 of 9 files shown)


merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
```diff
@@ -57,6 +58,9 @@ public static void main(String[] args) throws Exception {
     final var listenAction = new ListenSimulationCapability(hikariDataSource, notificationQueue);
     listenAction.registerListener();
 
+    Javalin app = Javalin.create().start(8080);
+    app.get("/health", ctx -> ctx.status(200));
```
pcrosemurgy, line 62: I'm pretty sure Javalin is a fairly lightweight web service but I'm hesitant to add the Javalin dep. to our worker driver unless we need a REST service for more than just `/health`.

pcrosemurgy:
This is a cool idea! I haven't really heard of a [`/health` endpoint](https://www.ibm.com/garage/method/practices/manage/health-check-apis/) before but am not really opposed to this :)

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
patkenneally: This is a fair point. @skovati and myself were pondering ways the workers' status could be tracked. Still pondering :) 

Edit: Another option is some sort of watch dog process. Long term there is a need to be able to kick start stuck workers/workers that haven't been heard from for a while. But I think something like a watchdog might be more than we need at the moment.

pcrosemurgy: A `/health` endpoint seems like a good way to do that then :) I was worried this was added just for symmetry with our other "services", but it sounds like there's a real use/need.

pcrosemurgy:
>`curl -i http://0.0.0.0:2718{3,4,5,7,8}/health`

Love seeing bash brace expansion put to use 👌

pcrosemurgy (approved):

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
pcrosemurgy: Leaving this thread open in case anyone else wants to chime in.
@skovati feel free to click resolve before merging :)

Twisol: I kind of agree with Paul, and I'd like to know a little more about the use case for these health checks. I know what the endpoints are ostensibly for, but how do they support us or our customers? A monitor like [god](http://godrb.com/) (or even `systemd`) can revive dead processes; system admins can use `top` or `ps` (or of course `docker ps` in Docker deployments) to see what's running; and a 200 /health response doesn't tell us if our capabilities are actually available.

Most concretely, I'm pretty sure `Javalin` is multithreaded by default. If the worker locks up in some way, does this solution actually identify the unhealthy state?

mattdailis: Not sure if this use case deserves a whole http server - but one time when health checks would've saved me and Adrien several hours was when we misconfigured the testing docker-compose file, so all requests to the scheduler server would fail with this (super helpful) hasura error:
```
{
    extensions: { internal: [Object], path: '$', code: 'unexpected' },
    message: 'http exception when calling webhook'
}
```
The e2e tests call a handful of health endpoints just as a sanity check before running the tests, so had the scheduler-server had a health endpoint and a corresponding check in the tests, we would have known right away that the server wasn't even listening for those requests.

Now, I imagine there are other ways we could address that use case (perhaps by logging the result of `docker-compose ps` before starting the end to end test suite).

This use case doesn't justify adding a health check to the worker though, since as @Twisol points out, the health of the worker's http server is potentially independent of its health as a service. 

> Fun fact, the Curiosity rover [experienced an anomaly](https://llis.nasa.gov/lesson/11201) near the beginning of its mission that was not caught automatically for this exact same reason. There was a watch dog that would poll the "health" endpoint on the flight software. However, the flight software is multi-threaded, so it replied with the equivalent of a `200 OK` to the watch dog, while its other thread was hung, which was preventing it from shutting down to re-charge. Only a "force shutdown" command from earth saved it from browning out.

skovati: You bring up two good points @Twisol that I think relate to my "Further Work" section. One of which is that the use case for these health checks isn't well defined at the moment, e.g. this could be used for anything from a status page in the UI to detecting unhealthy pods in a Kubernetes cluster. The second point is that these health checks currently lack any sort of logic, they simply return 200 if the HTTP server thread is working. An option I considered for merlin-worker was making a test connection to the database, and returning an HTTP response accordingly.

camargo: Yeah a `/health` endpoint with an automatic `200` returned is the most basic example of a "liveness" check. We should certainly look into more advanced ways to measure health on our servers. This article has a great overview of some of the more comprehensive types of heath checking: https://aws.amazon.com/builders-library/implementing-health-checks/#Ways_to_measure_health

camargo (approved):

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
skovati: Also, I might be mistaken, but it seems that Javalin is already a dependency of merlin-worker
https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-worker/build.gradle#L35

pcrosemurgy: @skovati True! Sorry, I used the term dependency kind of loosely there – Javalin is a build dependency but not imported anywhere in our Java source for `merlin-worker/` (prior to this PR), so there was no run-time dependency on it.

Twisol: > One of which is that the use case for these health checks isn't well defined at the moment

Yeah, this is the meat of my concern -- I have nothing in particular against the technical approach being taken here. I also have nothing against the mere concept of health checks! But we ought to get or concretely forecast some value from a change.

If we don't know what use cases we're trying to satisfy, I'd rather spend the effort defining those use cases. As this PR has demonstrated, actually adding health checks when we need them is not an overwhelming task.

> An option I considered for merlin-worker was making a test connection to the database, and returning an HTTP response accordingly.

That's a viable technical element, but what's still missing is an experimental setup. A health check is a kind of experiment. What hypothesis are we testing? What measurements can we make? And how do we support (or falsify) our hypothesis from those measurements? An ill-founded experiment can lead to misplaced confidence in the software, both from ourselves and from our customers.

(To that end, something to consider: should the health checks instead *fail* by default? Is that a less dangerous state of affairs?)

skovati: Here is an example of slightly more complex health checking I was considering @camargo @Twisol 

```suggestion
    app.get("/health", ctx -> {
      try ( var connection = hikariDataSource.getConnection() ) {
        ctx.status(200);
      } catch (SQLException e) {
        ctx.status(500).result(e.toString()));
      }
    });
```

Twisol: Dwelling on this a little more, I realized I've been treating `/health` itself as the thing that judges whether the service is healthy or not. If we treat it just as a measurement, a source of information for that downstream judgment, I'm a lot happier.

From that standpoint, I think I'd prefer it *always* return 200, with a body like `{"dbAvailable": false}`. This is genuinely useful technical information, independently of the "is it healthy" use case.

(Of course, there are systems that do genuinely treat `/health` as the arbiter of healthiness, and this interpretation diverges from those systems. If we want to treat `/health` as the arbiter of healthiness, my concern about an ill-defined experimental setup remains.)

camargo: Agree. We should return useful health info in the body depending on the check we do.

skovati:
Small formatting changes, rebased on upstream develop

skovati:
Added e2e tests for new health check endpoints. Rebased onto upstream develop.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
patkenneally: To summarize some  of the questions/points raised.
1. We have users asking for dedicated health check endpoints on the services. Currently users are doing less desirable aliveness checks like simply opening TCP connections etc.
2. The health check should (and will in a future PR) return useful information describing service specific health status.
3. The compliment of the experimental setup is currently being done by users in their AWS deployment. Aerie hasn't yet published a recommended AWS deployment guide, but that is coming also. 

@skovati I would suggest removing the draft tag and us continuing with through the PR process with this work.

skovati: Sounds good. I'll merge this simple implementation then, and create a ticket for adding additional functionality to these health checks.

PR #220, opened 2022-06-14
[AERIE-1906] Resolve LGTM errors

pcrosemurgy:
* **Tickets addressed:** [AERIE-1906](https://jira.jpl.nasa.gov/browse/AERIE-1906)
* **Review:** By commit  

## Description
Resolves LGTM errors reported [here](https://lgtm.com/projects/g/NASA-AMMOS/aerie/alerts/?mode=tree&severity=error&ruleFocus=6770060).

## Verification
All tests passing.

## Documentation
None.

## Future work
Resolve LGTM warnings (see https://lgtm.com/projects/g/NASA-AMMOS/aerie/alerts/?mode=tree).


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
```diff
@@ -25,8 +25,8 @@
  * pattern on a class closed to extension allows us to guarantee that no ambiguity occurs.
  */
 public abstract class SerializedValue {
-  static public SerializedValue NULL = SerializedValue.ofNull();
-
+  public static final SerializedValue NULL = SerializedValue.ofNull();
+  private static final int NULL_HASH = new Object().hashCode();
```
pcrosemurgy: Open to alternatives here. This was added since we need a hash code for `NULL` values.

At first I thought `NULL.hashCode()` would work but that results in a stack overflow due to unbound recursion (see `match()`).

Twisol: I think there are two alternatives worth considering. (Personally, I prefer option 2.)

1. Use `Objects.hashCode(x)` rather than `x.hashCode()`. In particular, this lets us write `Objects.hashCode(null)` (which always returns 0, but more importantly, which doesn't throw an exception).

    I'd also recommend using `Objects.hashCode(self.getClass(), ...)`, so that each variant is distinguished both by its type and by the data it carries. (Since `null` returns 0, which is all too easy to get a hash collision with.)

2. Update each variant factory method to use a local record instead of an anonymous class. Records get `hashCode` and `equals` for free, so you don't need to implement them at the top-level at all.

    ```java
    public static SerializedValue of(final long value) {
      record LongValue(long value) extends SerializedValue {
        @Override
        public <T> T match(final Visitor<T> visitor) {
          return visitor.onInt(this.value);
        }
      }

      return new LongValue(value);
    }
    ```

pcrosemurgy: Thanks! That's a great idea. I've never seen a `record` `extends` before – is that only possible with an abstract parent class?

pcrosemurgy: Yeah, not sure if `record` `extends` is legal here but I think you're completely on the right track, will try to work in a solution like this :)

mattdailis: Making SerializedValue a sealed interface and changing `extends` to `implements` in @Twisol 's example should do the trick, I think 🤔

mattdailis: Ah, and perhaps moving the record declaration out of the method...?

EDIT: after typing this into IntelliJ, it seems like inline record definitions are legal. _Wild!_

EDIT 2: Ah, but if we want it to be sealed, we'll still need to move the record declarations to the top level:
Screen Shot 2022-06-15 at 3 52 20 PM

pcrosemurgy: Thanks :) Will play around

Twisol: Whoops, yeah, records can only implement interfaces. (Records already implicitly extend the abstract class [`Record`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Record.html).)

`SerializedValue` was written before we had `sealed`. It's definitely meant to be a `sealed interface`.

Twisol: > EDIT: after typing this into IntelliJ, it seems like inline record definitions are legal. _Wild!_

:>

pcrosemurgy: Super cool – inline `record` simplifies this class a lot. Will push an update in a bit.

pcrosemurgy: Latest push: https://github.com/NASA-AMMOS/aerie/pull/220/commits/38cf6ab4961b5356640fe027ddc015149074ee5d#diff-ef23bb759d0daf96a6450cf32cc5b4fd0ac733ece05f20bfcbcd4cfe3c2fc130

mattdailis: Two thoughts here:
1. (as you and I briefly discussed on slack) The `toString` representation will contain the class names of the records, is that okay?
2. How important is it that `SerializedValue` be sealed? (Noting that the existing abstract class wasn't sealed, so we haven't actually loosened anything)

Twisol: `SerializedValue` has/had a private constructor, so it was as sealed as anything could be before `sealed` existed.

```java
  // Closed type family -- the only legal subclasses are those defined within the body of
  // this class definition.
  private SerializedValue() {}
```

pcrosemurgy: >(as you and I briefly discussed on slack) The toString representation will contain the class names of the records, is that okay?

I think this should be okay – as long as every use of `toString()` is for debugging/dev purposes.

>How important is it that SerializedValue be sealed? (Noting that the existing abstract class wasn't sealed, so we haven't actually loosened anything)

I guess making it sealed would prevent someone from doing something silly like making a custom implementation / anonymous implementation of `SerializedValue`. I'm onboard with making this interface sealed, would just involve yanking out those inline records and adding a `permits` declaration.

Twisol: > How important is it that SerializedValue be sealed?

Since it has a `Visitor` defined, the answer is either "very, because when pattern-matching lands we want to get rid of the visitor and use exhaustive `switch`, which relies on `sealed`", or "not very, because any subclass of `SerializedValue` has to invoke one of the `Visitor` methods, so it will behave like an existing instance anyway."

I'd go with "very".

mattdailis: Sounds good, let's make it sealed then :+1:

> SerializedValue has/had a private constructor, so it was as sealed as anything could be before sealed existed.

TIL

Twisol: > and adding a `permits` declaration.

No need to add a `permits` for classes defined in the same file. This is fine, for instance:

```java
sealed interface A {
  record X() implements A {}
  record Y() implements A {}
}
```

See [JEP 409](https://openjdk.org/jeps/409):
> When the permitted subclasses are small in size and number, it may be convenient to declare them in the same source file as the sealed class. When they are declared in this way, the sealed class may omit the permits clause and the Java compiler will infer the permitted subclasses from the declarations in the source file.

pcrosemurgy: Ohh true! I was thinking the `static` visitor classes defined in the interface would accidentally be picked up as an implied "permitted" class – only applies to non-`static` classes though so that's good :)

pcrosemurgy: Oh wait, I'm confusing myself. All of the classes in the interface are `static`. I'm curious if `abstract class DefaultVisitor<T> implements Visitor<T>` would be picked up as an implied permitted class – will test it out.

Twisol: > I think this should be okay – as long as every use of `toString()` is for debugging/dev purposes.

I think it's generally good practice to consider `toString()` as morally for debugging only. There are some classes that break this pattern (like `StringBuilder`), but given that there is no contract whatsoever for the content of the produced string -- and that the default `toString` outputs both the class and what is effectively the object's address, of all things -- treating it as debug-only is generally safe.

(If you need to produce string output for some other, better constrained purpose, I would define a different method altogether.)

Twisol: > Oh wait, I'm confusing myself. All of the classes in the interface are `static`. I'm curious if `abstract class DefaultVisitor<T> implements Visitor<T>` would be picked up as an implied permitted class – will test it out.

No, because it doesn't `implements SerializedValue`.

pcrosemurgy: d'oh! That's right

Twisol:
By the way, the Javadoc for `SerializedValue` still refers to its ancient previous name, `SerializedParameter`. Could you update that in this PR by any chance? 🙏

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
Twisol: All the same concerns here apply equally to `ValueSchema`, by the by. If we're changing `SerializedValue` to a `sealed interface` to ease implementation of `hashCode` and `equals`, we probably ought to give `ValueSchema` the same treatment.

pcrosemurgy:
> By the way, the Javadoc for `SerializedValue` still refers to its ancient previous name, `SerializedParameter`. Could you update that in this PR by any chance? 🙏

Done :)

pcrosemurgy:
There's an interesting failure I'm looking into now:
```
Method 'gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue.of(double)' must be InterfaceMethodref constant
java.lang.IncompatibleClassChangeError: Method 'gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue.of(double)' must be InterfaceMethodref constant
	at gov.nasa.jpl.aerielander.mappers.seis.ChannelRateValueMapper.serializeValue(ChannelRateValueMapper.java:42)
	at gov.nasa.jpl.aerielander.mappers.seis.ChannelRateValueMapper.serializeValue(ChannelRateValueMapper.java:12)
	at gov.nasa.jpl.aerie.merlin.framework.Registrar$1.serialize(Registrar.java:57)
	at gov.nasa.jpl.aerie.merlin.driver.engine.SimulationEngine.serializeProfile(SimulationEngine.java:616)
	at gov.nasa.jpl.aerie.merlin.driver.engine.SimulationEngine.computeResults(SimulationEngine.java:455)
	at gov.nasa.jpl.aerie.scheduler.simulation.IncrementalSimulationDriver.getSimulationResultsUpTo(IncrementalSimulationDriver.java:152)
	at gov.nasa.jpl.aerie.scheduler.simulation.SimulationFacade.computeSimulationResultsUntil(SimulationFacade.java:249)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.getConflicts(PrioritySolver.java:503)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.satisfyGoalGeneral(PrioritySolver.java:425)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.satisfyGoal(PrioritySolver.java:286)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.solve(PrioritySolver.java:245)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.getNextSolution(PrioritySolver.java:112)
	at gov.nasa.jpl.aerie.scheduler.AerieLanderRulesTest.schedule(AerieLanderRulesTest.java:43)
	at gov.nasa.jpl.aerie.scheduler.AerieLanderRulesTest.firstRule(AerieLanderRulesTest.java:51)
```
Not familiar with `must be InterfaceMethodref constant`, will dig deeper.

Line 42 of `ChannelRateValueMapper.java` contains:
```java
                    if (!map.containsKey("outRate")) {
```
Where `map` is of type `Map<String, SerializedValue>`.

Twisol:
`IncompatibleClassChangeError` is like `NoSuchMethodError` in that it should only ever happen because you compiled code against one version of a class and then loaded it into a JVM with another. Make sure that your model JARs have been recompiled.

pcrosemurgy:
Ahh that's right! This JAR is likely the one that's causing this snag: `scheduler/src/test/resources/gov/nasa/jpl/aerie/scheduler/aerie-lander-0.11.1-SNAPSHOT.jar`.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -170,37 +140,28 @@ public <T> T match(final Visitor<T> visitor) {
...
+    return new VariantSchema(variants);
   }
 
-  public static final ValueSchema REAL = ofReal();
-  public static final ValueSchema INT = ofInt();
-  public static final ValueSchema BOOLEAN = ofBoolean();
-  public static final ValueSchema STRING = ofString();
-  public static final ValueSchema DURATION = ofDuration();
-  public static final ValueSchema PATH = ofPath();
+  ValueSchema REAL = new RealSchema(new Object());
```
pcrosemurgy: Not too happy about the seemingly useless `new Object()` here. It exists to allow each `record` to delegate its hash code, otherwise `ValueSchema.INT.hashCode() == ValueSchema.REAL.hashCode()` when both schema `record`s are defined like:
```java
record SomeTypeSchema() implements ValueSchema {
  ...
}
```
without any members.

pcrosemurgy: But considering that we get a free `equals()` and `hashCode()` out of this I think the tradeoff may be worthwhile.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
pcrosemurgy: Just pushed `ValueSchema` refactoring. Marking this thread as resolved, feel free to re-open if any `SerializedValue` issues come up.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
mattdailis: Do I understand the problem correctly, in that records with the same elements, but different class names, have the same hashCode?

Twisol: It's okay if different `hashCode` implementations sometimes return the same thing. What's important is that `hashCode(x) == hashCode(y)` whenever `x.equals(y)`, which we get free for records.

Especially in the contexts where we use ValueSchema -- that is, approximately never as a key -- we shouldn't be concerned with this collision. The `new Objec()` is a poor trade.

pcrosemurgy: So you're saying `ValueSchema.INT.hashCode() == ValueSchema.REAL.hashCode()` should be acceptable? I don't think `ValueSchema.INT.equals(ValueSchema.REAL)` should yield `true`, so I would expect a unique hash code for these 2 schemas as well, you know?

pcrosemurgy: Or put more clearly:
```
ValueSchema.INT.equals(ValueSchema.REAL)
```
currently is false under this PR.

So I would expect `ValueSchema.INT.hashCode()` to not be the same value as `ValueSchema.REAL.hashCode()`.

Twisol: Hash functions, by definition, take a big space of elements and squish it down to a smaller space of numbers, because numbers are much easier to compare. Hash functions will always have collisions; a system that assumes hashes are unique is definitionally incorrect.

The right angle to take on `hashCode` is performance-based -- how evenly can we spread objects amongst numbers, so that we can filter a set down to a much smaller list of candidates rather than iterating over the entire set?

If you want to ask if two schemas are equal, use `equals()`!

mattdailis: The arrow points the other way - if they're equal, they _must_ have the same hashcode. That's the only rule. The converse doesn't need to hold (i.e. if they're not equal, the hashcodes may or may not be equal, it doesn't matter*)

*it matters for performance, but not correctness. In this case, a) we don't use ValueSchemas as keys very often, and b) even if we did, we'd be looking at number of collisions linear in the number of variants of `ValueSchema` (i.e. on the order of 10), which I doubt would have a measurable effect on performance

Twisol: > I don't think `ValueSchema.INT.equals(ValueSchema.REAL)` should yield `true`,

They don't. The implication goes the other way -- if the hashes are *distinct*, the objects *cannot* be equal. Nothing says that equal hashes implies equal objects.

Per [the documentation for `hashCode`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Object.html#hashCode()):

> It is *not* required that if two objects are unequal according to the [equals](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object)) method, then calling the hashCode method on each of the two objects must produce distinct integer results. However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hash tables.

pcrosemurgy: Thanks @Twisol @mattdailis, was getting myself confused with uniqueness/equality. I'll take the `Object` out of these records!

pcrosemurgy:
Will diagnose the e2e test failures next week.

I'm unable to get the e2e tests to even pass on `develop` with:
```
$ docker-nuke aerie ; gradle clean assemble && docker-compose build && docker-compose up
```
followed by (in a separate shell):
```
$ cd e2e-tests
$ npm install && npm test
```
This results in:
```
  15 failed
    src/tests/health.test.ts:10:3 › Health › Hasura is healthy =====================================
    src/tests/mission-models.test.ts:8:3 › Mission Models › Upload mission model .jar file to the gateway
    src/tests/mission-models.test.ts:15:3 › Mission Models › Create mission model using the jar_id returned from the gateway
    src/tests/mission-models.test.ts:28:3 › Mission Models › Delete mission model ==================
    src/tests/scheduler.test.ts:16:3 › Scheduling › Upload jar and create mission model ============
    src/tests/scheduler.test.ts:34:3 › Scheduling › Create Plan ====================================
    src/tests/scheduler.test.ts:47:3 › Scheduling › Create Simulation ==============================
    src/tests/scheduler.test.ts:58:3 › Scheduling › Create Scheduling goal =========================
    src/tests/scheduler.test.ts:83:3 › Scheduling › Get Plan Revision ==============================
    src/tests/scheduler.test.ts:90:3 › Scheduling › Create Scheduling Specification ================
    src/tests/scheduler.test.ts:104:3 › Scheduling › Create Scheduling Specification Goal ==========
    src/tests/scheduler.test.ts:118:3 › Scheduling › Run scheduling ================================
    src/tests/scheduler.test.ts:135:3 › Scheduling › Get Plan ======================================
    src/tests/scheduler.test.ts:144:3 › Scheduling › Delete plan ===================================
    src/tests/scheduler.test.ts:152:3 › Scheduling › Delete mission model ==========================
```

patkenneally (approved):

pcrosemurgy:
Finally got the e2e tests figured out here :) Also note that #238 has been pulled out of this PR.

mattdailis (approved):
Nice work, Paul

PR #221, opened 2022-06-14
[AERIE-1776]  adjust compilation targets

dyst5422:
* **Tickets addressed:** AERIE-1776
* **Review:** By commit  

## Description

Adjusted the compiler config base to the recommended Node16 values from https://github.com/tsconfig/bases#centralized-recommendations-for-tsconfig-bases
and then went through and adjusted any code to meet the stricter compiler rules.

## Verification

Built scheduling, constraints, and command expansion. Ran command expansion tests and global gradlew test

## Documentation

No documentation invalidated

## Future work

Need to adjust the user code runner to the same compiler base


mattdailis (approved):
Looks fine to me :+1:

command-expansion-server/src/app.ts
```diff
@@ -54,7 +54,7 @@ type Context = {
   expansionDataLoader: InferredDataloader<typeof expansionBatchLoader>;
 };
 
-app.use(async (req: Request, res: Response, next: NextFunction) => {
+app.use(async (_: Request, res: Response, next: NextFunction) => {
```
goetzrrGit, line 57: More of a general question. How come `req` -> `_`

Is it because it isn't being used?

goetzrrGit (approved):
The test passed on my end. Looks good.

command-expansion-server/src/app.ts
skovati: Looks like it, renaming unused variables to `_` gets rid of a linter warning for me. Curious if we should adopt this practice elsewhere

camargo: Yeah this is a widely used pattern in JS apps - especially when using callback based APIs where you do not control what parameters are in the callback.

camargo (approved):

PR #225, opened 2022-06-16
[AERIE-1880] ApplyWhen - scheduling goal windowing

pranav-super:
* **Tickets addressed:** AERIE-1880
* **Review:** By commit  

## Description
This ticket was approached first by defining what it means for this change to be made (i.e. what it means for `ApplyWhen` to be applied on a goal), and then applying it in the backend goal by goal, and finally in the frontend. **NOTE: IN DEVELOPMENT, "ApplyWhen" WAS ALSO REFERRED TO AS "AnalyzeWhen", OWING TO A DISCUSSION THAT LED TO THE CREATION OF THIS TICKET - THEY PRESENTLY MEAN THE SAME THING.**

So, what does `ApplyWhen` mean? The idea is that an `Expression<Windows>` that, when evaluated, yields a `Windows`, can be given to restrict when a goal is active. This could be a strict temporal window (which needs to be wrapped as an expression - see the `WindowsWrapperExpression` class), or a resource window (which when simulated with the SimulationFacade yields a temporal window). **Do not that these windows can be disjoint!** What can be a little vague is what it means for each type of goal to have such a restriction applied:
- For a `CoexistenceGoal`, this means that any activities who match the coexistence criterion (the `forEach: ActivityExpression`) that lie outside of the window aren't considered for scheduling according to this goal. Similarly, if a matching activity is spliced by a set of goal windows, i.e. different windows contain parts of a matching activity, each separate window is treated as a match and an activity will be scheduled for each of those, following the goal's specification.
- For a `RecurrenceGoal`, this means recurrence only happens within this window. Note that recurrence is local, and partial recurrence isn't allowed. That means that every disjoint window has its own recurrence interval, and if a window cuts a recurrence interval short (i.e. the window is 11 seconds long but the recurrence interval is 2, meaning the last interval would be spliced), that interval won't be scheduled.
- Finally, for a `CardinalityGoal`, whatever can fit in the window is scheduled. 

The commits go in the order of tests, backend, then frontend. They start by detailing the creation of test files and supporting classes (such as a new sample `Goal` type under the `FooMissionModel`). 
Then the commits detail updates to the backend, first with clarifying an error message that appeared a few times in testing (it wasn't actually an error, more of a warning about the temporal network solver that places activities), then a refactor to use `Expression<Windows>`, a refactor of the goal types, and finally one of the EDSL itself. The EDSL update and the addition of its new tests are included as one commit. Finally, a boundary issue that was found in `RecurrenceGoal` is resolved in the last commit.

## Verification
The changes were validated through testing. This was primarily achieved by the `TestAnalyzeWhen.java` file, which aims to do extensive bounds checking for each of the goals to ensure that the implementation behaves consistently and predictably. There was some odd behavior encountered, namely with `UncontrollableDurationActivities` - this is documented.

Similarly, for the EDSL, both a parsing test was added (`SchedulingDSLCompilationTests.java`), as well as a test that verified parsing and the results of scheduling (`SchedulingIntegrationTests.java`). Only one of each was written, as one was sufficient to verify parsing, and one was sufficient to verify integration (as actual boundaries of the feature were tested in `TestAnalyzeWhen.java`).

## Documentation
The only possible invalidation would be regarding that of uncontrollable duration activities which were found to be problematic, but those were not documented. The documentation additions coming from this PR will be primarily just explaining how to use `ApplyWhen`, its edge cases, and the fact that `UncontrollableDurationActivities` have been found to be problematic.

## Future work
A new ticket was created to investigate `UncontrollableDurationActivities` - that is a direct product of the testing done while working on [this](https://jira.jpl.nasa.gov/projects/AERIE/issues/AERIE-1887) ticket.
Furthermore, in discussing this feature, a distinction was made between possible features, namely a `PlaceWhen` and `AnalyzeWhen` (of which this ticket handles the latter, despite analysis not being implemented) - a future feature might involve a separate `PlaceWhen` and `AnalyzeWhen`, as well as the creation of an analysis mode. For now, they can safely and reasonably be conflated, as analysis is not yet an existing feature.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsWrapperExpression.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsWrapperExpression.java
@@ -1,0 +1,34 @@
+package gov.nasa.jpl.aerie.constraints.tree;
+
+import gov.nasa.jpl.aerie.constraints.model.ActivityInstance;
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Window;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.Set;
+
+public class WindowsWrapperExpression implements Expression<Windows> {
+  public final Windows windows;
+
+  public WindowsWrapperExpression(final Windows windows) { this.windows = windows; }
+
+  @Override
+  public Windows evaluate(final SimulationResults results, final Window bounds, final Map<String, ActivityInstance> environment) {
+    return windows;
+  }
+
+  public String prettyPrint(final String prefix) {
+    return String.format(
+      "%s(expression-wrapping %s)",
+      prefix,
+      this.windows.toString()
+    );
+  }
+  /** Add the resources referenced by this expression to the given set. **/
+  public void extractResources(Set<String> names) { }
+}
--- a/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -40,6 +40,14 @@ export class Goal {
     });
   }
 
+  public applyWhen(window: WindowsEDSL.Windows): Goal {
+    return Goal.new({
+      kind: AST.NodeKind.ApplyWhen,
+      goal: this.__astNode,
+      window: window.__astNode
+    });
+  }
+
   public static ActivityRecurrenceGoal(opts: { activityTemplate: ActivityTemplate, interval: Duration }): ActivityRecurrenceGoal {
     return Goal.new({
       kind: AST.NodeKind.ActivityRecurrenceGoal,
@@ -161,6 +169,8 @@ declare global {
     public and(...others: Goal[]): Goal
 
     public or(...others: Goal[]): Goal
+
+    public static applyWhen(window: WindowsEDSL.Windows): Goal
 
     public static ActivityRecurrenceGoal(opts: { activityTemplate: ActivityTemplate, interval: Duration }): ActivityRecurrenceGoal
 
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
@@ -1,5 +1,4 @@
 package gov.nasa.jpl.aerie.scheduler.server.models;
-
 import gov.nasa.jpl.aerie.constraints.time.Windows;
 import gov.nasa.jpl.aerie.constraints.tree.Expression;
 import gov.nasa.jpl.aerie.json.Iso;
@@ -12,11 +11,9 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.TimeUtility;
 import gov.nasa.jpl.aerie.scheduler.constraints.timeexpressions.TimeAnchor;
 import org.apache.commons.lang3.tuple.Pair;
-
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-
 import static gov.nasa.jpl.aerie.json.BasicParsers.boolP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.enumP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.intP;
@@ -32,9 +29,7 @@ import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
 import static gov.nasa.jpl.aerie.json.Uncurry.tuple;
 import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
 import static gov.nasa.jpl.aerie.scheduler.server.http.SerializedValueJsonParser.serializedValueP;
-
 public class SchedulingDSL {
-
   private static final JsonParser<ActivityTemplate> activityTemplateP =
       productP
           .field("activityType", stringP)
@@ -43,12 +38,11 @@ public class SchedulingDSL {
               untuple(ActivityTemplate::new),
               $ -> tuple($.activityType(), $.arguments())));
 
-
   private static final JsonParser<Duration> durationP =
       longP
-      . map(Iso.of(
-          microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
-          duration -> duration.in(Duration.MICROSECONDS)));
+          . map(Iso.of(
+              microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
+              duration -> duration.in(Duration.MICROSECONDS)));
 
   private static final JsonParser<ClosedOpenInterval> intervalP =
       productP
@@ -65,6 +59,7 @@ public class SchedulingDSL {
           .map(Iso.of(
               untuple(CardinalitySpecification::new),
               $ -> tuple($.duration(), $.occurrence())));
+
   private static final ProductParsers.JsonObjectParser<GoalSpecifier.RecurrenceGoalDefinition> recurrenceGoalDefinitionP =
       productP
           .field("activityTemplate", activityTemplateP)
@@ -144,6 +139,16 @@ public class SchedulingDSL {
                     GoalSpecifier.GoalOr::goals));
   }
 
+  private static ProductParsers.JsonObjectParser<GoalSpecifier.GoalApplyWhen> goalApplyWhenF(final JsonParser<GoalSpecifier> goalSpecifierP) {
+    return productP
+        .field("goal", goalSpecifierP)
+        .field("window", windowsExpressionP)
+        .map(Iso.of(untuple(GoalSpecifier.GoalApplyWhen::new),
+                    goalDefinition -> tuple(
+                        goalDefinition.goal(),
+                        goalDefinition.windows())));
+  }
+
 
   private static final JsonParser<GoalSpecifier> goalSpecifierP =
       recursiveP(self -> SumParsers.sumP("kind", GoalSpecifier.class, List.of(
@@ -151,13 +156,12 @@ public class SchedulingDSL {
           SumParsers.variant("ActivityCoexistenceGoal", GoalSpecifier.CoexistenceGoalDefinition.class, coexistenceGoalDefinitionP),
           SumParsers.variant("ActivityCardinalityGoal", GoalSpecifier.CardinalityGoalDefinition.class, cardinalityGoalDefinitionP),
           SumParsers.variant("GoalAnd", GoalSpecifier.GoalAnd.class, goalAndF(self)),
-          SumParsers.variant("GoalOr", GoalSpecifier.GoalOr.class, goalOrF(self))
+          SumParsers.variant("GoalOr", GoalSpecifier.GoalOr.class, goalOrF(self)),
+          SumParsers.variant("ApplyWhen", GoalSpecifier.GoalApplyWhen.class, goalApplyWhenF(self))
       )));
 
 
   public static final JsonParser<GoalSpecifier> schedulingJsonP = goalSpecifierP;
-
-
   public sealed interface GoalSpecifier {
     record RecurrenceGoalDefinition(
         ActivityTemplate activityTemplate,
@@ -176,19 +180,19 @@ public class SchedulingDSL {
     ) implements GoalSpecifier {}
     record GoalAnd(List<GoalSpecifier> goals) implements GoalSpecifier {}
     record GoalOr(List<GoalSpecifier> goals) implements GoalSpecifier {}
+    record GoalApplyWhen(
+        GoalSpecifier goal,
+        Expression<Windows> windows
+    ) implements GoalSpecifier {}
   }
 
   public record LinearResource(String name) {}
-
   public record CardinalitySpecification(Optional<Duration> duration, Optional<Integer> occurrence){}
   public record ClosedOpenInterval(Duration start, Duration end){}
   public record ActivityTemplate(String activityType, Map<String, SerializedValue> arguments) {}
-
   public sealed interface ConstraintExpression {
     record ActivityExpression(String type) implements ConstraintExpression {}
-
     record WindowsExpression(Expression<Windows> expression) implements ConstraintExpression {}
   }
-
   public record ActivityTimingConstraint(TimeAnchor windowProperty, TimeUtility.Operator operator, Duration operand, boolean singleton) {}
 }
--- a/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
+++ b/scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
@@ -748,6 +748,58 @@ public class SchedulingIntegrationTests {
     assertEquals("Fyffes", changeProducer.args().get("producer").asString().get());
   }
 
+  @Test
+  void testApplyWhen() {
+    final var growBananaDuration = Duration.of(1, Duration.SECONDS);
+
+    final var results = runScheduler(
+        BANANANATION,
+        List.of(
+            new MockMerlinService.PlannedActivityInstance(
+                "GrowBanana",
+                Map.of(
+                    "quantity", SerializedValue.of(1),
+                    "growingDuration", new DurationValueMapper().serializeValue(growBananaDuration)),
+                Duration.of(1, Duration.SECONDS)),
+            new MockMerlinService.PlannedActivityInstance(
+                "GrowBanana",
+                Map.of(
+                    "quantity", SerializedValue.of(1),
+                    "growingDuration", new DurationValueMapper().serializeValue(growBananaDuration)),
+                Duration.of(2, Duration.SECONDS)),
+            new MockMerlinService.PlannedActivityInstance(
+                "GrowBanana",
+                Map.of(
+                    "quantity", SerializedValue.of(1),
+                    "growingDuration", new DurationValueMapper().serializeValue(growBananaDuration)),
+                Duration.of(3, Duration.SECONDS)),
+            new MockMerlinService.PlannedActivityInstance(
+                "PickBanana",
+                Map.of(
+                    "quantity", SerializedValue.of(100000)
+                ),
+                Duration.of(48, Duration.HOURS))
+        ),
+        List.of(new SchedulingGoal(new GoalId(0L), """
+                  export default () => Goal.ActivityRecurrenceGoal({
+                      activityTemplate: ActivityTemplates.ChangeProducer({producer: "Morpheus"}),
+                      interval: 24 * 60 * 60 * 1000 * 1000
+                    }).applyWhen(Real.Resource("/plant").greaterThan(1.0))""", true)
+        )
+    );
+
+    for (MockMerlinService.PlannedActivityInstance i : results.updatedPlan().stream().toList()) {
+      System.out.println(i.type().toString() + ": " + i.startTime().toString());
+    }
+
+    assertEquals(1, results.scheduleResults.goalResults().size()); //starts an instant before 12:00 (when there is more than 1 plant - the window is about 00:00-24:00 because the plant gets created early, then the window is subdivided, and then a window is picked, seemingly arbitrarily, as this is uncontrollable duration (see line 304 of ActivityCreationTemplate), after which cadence is maintained) then schedules 4, as expected. 7 total then.
+    assertEquals(6, results.updatedPlan().size());
+
+    final var planByActivityType = partitionByActivityType(results.updatedPlan());
+    final var changeProducer = planByActivityType.get("ChangeProducer").iterator().next();
+    assertEquals("Morpheus", changeProducer.args().get("producer").asString().get());
+  }
+
   private static Map<String, Collection<MockMerlinService.PlannedActivityInstance>>
   partitionByActivityType(final Iterable<MockMerlinService.PlannedActivityInstance> activities) {
     final var result = new HashMap<String, Collection<MockMerlinService.PlannedActivityInstance>>();
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
@@ -48,11 +48,7 @@ public class TimeRangeExpression {
       if(minTimepoint.isPresent() && maxTimepoint.isPresent()) {
         final var anchorActSearch = new ActivityExpression.Builder()
             .basedOn(actTemplate)
-            .startsIn(Window.between(
-                Duration.max(Duration.ZERO, minTimepoint.get()),
-                Window.Inclusivity.Inclusive,
-                maxTimepoint.get(),
-                Window.Inclusivity.Exclusive)).build();
+            .startsIn(inter).build(); //check if it exists IN the windows, not just the upper and lower bounds of the window
         final var anchorActs = plan.find(anchorActSearch, simulationResults);
         for (var anchorAct : anchorActs) {
           var endInclusivity = Window.Inclusivity.Exclusive;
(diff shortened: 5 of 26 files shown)


scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
```diff
@@ -725,6 +725,56 @@ export default (): Goal => {
...
+                      window: Real.Resource("/plant").greaterThan(1.0)
+                  })""", true)
+        )
+    );
+
+    System.out.println("---RESULTS---");
+    for(MockMerlinService.PlannedActivityInstance p : results.updatedPlan()) {
+      System.out.println(p.startTime().toString() + ", " + p.type().toString());
+    }
+    System.out.println("-------------");
```
adrienmaillard: Remove the prints.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -48,11 +48,12 @@ public Windows computeRange(final SimulationResults simulationResults, final Pla
       if(minTimepoint.isPresent() && maxTimepoint.isPresent()) {
         final var anchorActSearch = new ActivityExpression.Builder()
             .basedOn(actTemplate)
-            .startsIn(Window.between(
+            /*.startsIn(Window.between(
                 Duration.max(Duration.ZERO, minTimepoint.get()),
                 Window.Inclusivity.Inclusive,
                 maxTimepoint.get(),
-                Window.Inclusivity.Exclusive)).build();
+                Window.Inclusivity.Exclusive)).build();*/
```
adrienmaillard: Remove commented lines.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ActivityTemplateGoal.java
```diff
@@ -125,4 +127,11 @@ protected ActivityTemplateGoal() { }
    */
   protected ActivityCreationTemplate desiredActTemplate;
 
+  /**
+   * checked by getConflicts every time it is invoked to see if the Window(s)
+   * corresponding to when this goal has changed, which is unexpected behavior
+   * that needs to be caught
+   */
+  protected Windows evaluatedTemporalContext;
```
adrienmaillard: What about calling it `initialTemporalContext` or something that conveys the same meaning ?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -178,69 +178,90 @@ public boolean isUnsatisfiable(){
...
-    if (this.durationRange != null && !this.durationRange.contains(total)) {
-      if (total.compareTo(this.durationRange.start) < 0) {
-        durToSchedule = this.durationRange.start.minus(total);
-      } else if (total.compareTo(this.durationRange.end) > 0) {
-        logger.warn("Need to decrease duration of activities from the plan, impossible because scheduler cannot remove activities");
-        return List.of(new UnsatisfiableGoalConflict(this,
-                                                     "Need to decrease duration of activities from the plan, impossible because scheduler cannot remove activities"));
-      }
+    //make sure it hasn't changed
+    if (this.evaluatedTemporalContext != null && !windows.includes(this.evaluatedTemporalContext)) {
```
adrienmaillard: Why use `includes()` and not `equals()` ?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CoexistenceGoal.java
```diff
@@ -157,10 +157,42 @@ protected CoexistenceGoal fill(CoexistenceGoal goal) {
...
-    final var conflicts = new java.util.LinkedList<Conflict>();
+  public java.util.Collection<Conflict> getConflicts(Plan plan, final SimulationResults simulationResults) { //TODO: check if window gets split and if so, notify user?
+
+    //NOTE: temporalContext IS A WINDOWS OVER WHICH THE GOAL APPLIES, USUALLY SOMETHING BROAD LIKE A MISSION PHASE
+    //NOTE: expr IS A WINDOWS OVER WHICH A COEXISTENCEGOAL APPLIES, FOR EXAMPLE THE WINDOWS CORRESPONDING TO 5 SECONDS AFTER EVERY BASICACTIVITY IS SCHEDULED
+    //NOTE: IF temporalContext IS SMALLER THAN expr OR SOMEHOW BISECTS IT, ODDS ARE THIS ISN'T ANTICIPATED USER BEHAVIOR. GENERALLY, ANALYZEWHEN SHOULDN'T BE PROVIDING
+    //        A SMALLER WINDOW, AND HONESTLY DOESN'T MAKE SENSE TO USE ON TOP BUT IS SUPPORTED TO MAKE CODE MORE CONSISTENT. IF ONE NEEDS TO USE ANALYZEWHEN ON TOP
+    //        OF COEXISTENCEGOAL THEY SHOULD PROBABLY REFACTOR THEIR COEXISTENCE GOAL. ONE SUCH USE WOULD BE IF THE COEXISTENCEGOAL WAS SPECIFIED IN TERMS OF
+    //        AN ACTIVITYEXPRESSION AND THEN ANALYZEWHEN WAS A MISSION PHASE, BUT I THINK ITS POSSIBLE TO JUST SPECIFY AN EXPRESSION<WINDOWS> THAT COMBINES THOSE.
```
adrienmaillard: Reformat: change to lowercase and remove "I" statement.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ProceduralCreationGoal.java
```diff
@@ -200,16 +200,16 @@ protected ProceduralCreationGoal() { }
...
    */
-  private Collection<ActivityInstance> getRelevantGeneratedActivities(Plan plan) {
+  private Collection<ActivityInstance> getRelevantGeneratedActivities(Plan plan, SimulationResults simulationResults) {
 
     //run the generator in the plan context
     final var allActs = generator.apply(plan);
 
     //filter out acts that don't have a start time within the goal purview
-    final var goalContext = getTemporalContext();
+    final var goalContext = getTemporalContext(); //can extract Windows by calling evaluate, don't need to pass more to evaluate as it won't be another type of expression<windows>
```
adrienmaillard: Comment is not clear.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ProceduralCreationGoal.java
```diff
@@ -200,16 +200,16 @@ protected ProceduralCreationGoal() { }
...
     final var filteredActs = allActs.stream().filter(
         act -> ((act.getStartTime() != null)
-                && goalContext.contains(act.getStartTime()))
+                && goalContext.evaluate(simulationResults).includesPoint(0,act.getStartTime()))
```
adrienmaillard: You evaluate the expression for each activities, I think this is inefficient. Evaluate once and pass the resulting windows.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -0,0 +1,1377 @@
...
+
+
+    problem.setGoals(List.of(goal));
+
+    final var solver = new PrioritySolver(problem);
+
+    var plan = solver.getNextSolution().orElseThrow();
+    for(ActivityInstance a : plan.getActivitiesByTime()){
+      System.out.println(a.getStartTime());
+    }
```
adrienmaillard: Remove or `logger.debug()`.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -0,0 +1,1377 @@
...
+        .build();
+
+
+    problem.setGoals(List.of(goal));
+
+    final var solver = new PrioritySolver(problem);
+
+    var plan = solver.getNextSolution().orElseThrow();
+    for(ActivityInstance a : plan.getActivitiesByTime()){
+      System.out.println(a.getStartTime());
```
adrienmaillard: Use `logger.debug` or remove. Same for similar prints below.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -0,0 +1,1377 @@
...
+    }
+
+    System.out.println("ahh!");
+
+    //assertTrue(TestUtility.activityStartingAtTime(plan,Duration.of(1, Duration.SECONDS), activityType)); //cutting off mid interval should fail
```
adrienmaillard: Remove commented lines.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -0,0 +1,1377 @@
...
+    problem.setGoals(List.of(goal));
+
+    final var solver = new PrioritySolver(problem);
+
+    var plan = solver.getNextSolution().orElseThrow();
+    for(ActivityInstance a : plan.getActivitiesByTime()){
+      System.out.println(a.getStartTime());
+    }
+
+    System.out.println("ahh!");
```
adrienmaillard: Remove print.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -178,69 +178,90 @@ public boolean isUnsatisfiable(){
...
     final var conflicts = new LinkedList<Conflict>();
-    //at this point, have thrown exception if not satisfiable
-    //compute the missing association conflicts
-    for(final var act : acts){
-      if(!associatedActivitiesToThisGoal.contains(act) && planEvaluation.canAssociateMoreToCreatorOf(act)){
-        //they ALL have to be associated
-        conflicts.add(new MissingAssociationConflict(this, List.of(act)));
+
+    for(Window subWindow: windows) {
+      Windows timeDomain = this.expr.computeRange(simulationResults, plan, new Windows(subWindow));
```
adrienmaillard: Why keep both `expr` and `temporalContext` now that they are the same kind ?

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -162,6 +170,8 @@ declare global {
 
     public or(...others: Goal[]): Goal
 
+    public static applyWhen(opts: { goal: Goal, window: WindowsEDSL.Windows } ): Goal
```
adrienmaillard: Have you tested this directly in the UI ?

adrienmaillard (changes requested):
Thank you for this work, you haven't started with the easiest of tasks. 

I was not there when this ticket was discussed so I still am unsure about the need for this feature. 

It seems to me that this new operator duplicates the goal effects. It is then some kind of composite goal. I would call it `applyForEach` or `forEach` in the edsl. `applyWhen` makes sense to me if it was a property of a goal, not a high level operator applying on goals. Opinions @mattdailis @patkenneally @JoelCourtney ? 

It is a departure from the original meaning of `temporalContext` that meant that we do not check/place activities elsewhere.  
 
There is clearly a difference between how it is presented in the edsl and how it is implemented in the backend. A more symmetric implementation (not that it would be better) would have been a new ForEach goal that would have actually created phantom goals with each small temporal context.

Anyway, I appreciate the thoroughness of your tests and the struggle that it must have been to deal with all the semantics at once.  

You'll see I have a few specific questions and minor remarks.

mattdailis:
There was some confusion about whether ApplyWhen was available to users yet or not - I'm pasting the writeup below, feel free to edit this message, and then we can move it back into the wiki when we're ready:

## Restricting when a goal applied
The Aerie scheduler provides support for restricting _when_ a goal applies with the `Goal.applyWhen()` node. This node allows users to provide a `Window` (which could be a time or a resource-based window) and a goal to apply within that window.

The `Goal.applyWhen()` node, like the goals, takes one argument - an `options` argument, which requires two key-value pairs: a `goal` (which needs to be some declaration of a goal), and a window. What follows is an example that applies a daily recurrence goal only when a given resource is greater than 2. If the resource is less than two, then the goal is no longer applied.

image

#### Notes on boundaries:
* If you are trying to schedule an activity, or a recurrence within a window but that window cuts off either the activity or the recurrence interval (depending on the goal type), it will not be scheduled. For example, if you had a recurrence interval of 3 seconds, scheduling a 2 second activity each recurrence, and had the following window, you'd get the following:
```
RECURRENCE INTERVAL: [++-++-++-]
GOAL WINDOW:         [+++++----]
RESULT:              [++-------]
```
That, is, the second activity won't be scheduled as the goal window cuts off its recurrence interval. 
* Scheduling is _local_, not global. This means for every window that is matched (as it is possible to have disjoint windows, imagine a resource that fluctuates upward and downward but only applying that goal when the resource is over a certain value), the goal is applied individually. So, for that same recurrence interval setup as before, we could have:
```
RECURRENCE INTERVAL: [++-++-++-++-]
GOAL WINDOW:         [+++++--+++--]
RESULT:              [++-----++---] //(the second one is applied independently of the first!)
```
* When mapping out a temporal window to apply a goal over, keep in mind that the ending boundary of the goal is _exclusive_, i.e. if I want to apply a goal in the window of 10-12 seconds, it will apply only on seconds 10 and 11. This is in line with the [fencepost problem](https://icarus.cs.weber.edu/~dab/cs1410/textbook/3.Control/fencepost.html).

pranav-super:
> It is a departure from the original meaning of temporalContext that meant that we do not check/place activities elsewhere.

By this do you mean the entire plan doesn't have activities placed elsewhere, or that activities aren't placed elsewhere with respect to the goal? Because in this case the latter is true.

> to me that this new operator duplicates the goal effects. It is then some kind of composite goal. I would call it applyForEach or forEach in the edsl. applyWhen makes sense to me if it was a property of a goal, not a high level operator applying on goals. 

> There is clearly a difference between how it is presented in the edsl and how it is implemented in the backend. A more symmetric implementation (not that it would be better) would have been a new ForEach goal that would have actually created phantom goals with each small temporal context.

The idea was to implement it as a property (decision made because temporalContext already existed and was referenced in a couple locations, so it made sense to leverage it, plus some sort of temporal restriction like this is a lot easier if implemented as a property of the goal), but to expose it to the edsl as a wrapper (I think this felt somewhat cleaner to represent). That being said, does anyone have strong opinions on implementing it as an optional goal property in the edsl @JoelCourtney @mattdailis @patkenneally?

> There was some confusion about whether ApplyWhen was available to users yet or not - I'm pasting the writeup below, feel free to edit this message, and then we can move it back into the wiki when we're ready:

My apologies

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ActivityTemplateGoal.java
pranav-super: `initiallyEvaluatedTemporalContext` would be the best way to convey the meaning but I'm not sure if there's a more succinct way to say it (unless the fact that its a `Windows` object implies its been evaluated in which case your suggestion is the best name, I'm not sure what's the most readable option if one is unfamiliar with the code).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
pranav-super: That's a good point, `equals()` is more accurate and more readable.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ProceduralCreationGoal.java
pranav-super: The idea here was just to clarify the arguments being passed to `evaluate()`, as there were more arguments previously, and `getTemporalContext()` had a different return type (a Window). This clarification made more sense during the refactor but after everything is a lot more meaningless. It has been removed.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CoexistenceGoal.java
pranav-super: Fixed

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
pranav-super: Fixed

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ActivityTemplateGoal.java
pranav-super: Fixed

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
pranav-super: Fixed

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
pranav-super: Fixed

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/ProceduralCreationGoal.java
pranav-super: Fixed!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
pranav-super: No reason, it has been factored out.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
pranav-super: Fixed

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
pranav-super: Fixed

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
pranav-super: Replaced all

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
pranav-super: Fixed

scheduler-server/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
pranav-super: Yes, I was able to recreate the test from SchedulingIntegrationTests in the UI and it worked!
image
(I did restrict the window a bit in this one ^^)

adrienmaillard:
Thank you for making the changes I have pointed out. I am still a little bit unsatisfied by the edsl implementation as a wrapper rather than a property. I'll wait for others to voice their opinions on this subject before approving.

dyst5422:
I agree with @adrienmaillard. I think it feels weird to not have it be a method on a goal that modifies the context of that goal directly - especially since it doesn't modify the goal directly but returns a new goal. This is important because it allows us to still reuse goals in multiple contexts. IE
```typescript
const goal = Goal.ActivityRecurrenceGoal({
  activityTemplate: ActivityTemplates.BiteBanana({
    biteSize: 1
  }),
  interval: 24 * 60 * 60 * 1000 * 1000 //24 hours in microseconds
});
return goal.applyWhen(Real.Resource (Resources["/sample/resource/l"]).greaterThan(2.0))
.or(goal.applyWhen(Real.Resource (Resources["/sample/resource/l"]).lessThan(1.0)))
```

adrienmaillard (approved):
Thank you for this work.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsWrapperExpression.java
```diff
@@ -0,0 +1,30 @@
...
+  public WindowsWrapperExpression(final Windows windows) { this.windows = windows; }
+
+  @Override
+  public Windows evaluate(final SimulationResults results, final Window bounds, final Map<String, ActivityInstance> environment) {
+    return windows;
+  }
+
+  public String prettyPrint(final String prefix) {
+    return "WindowsWrapper: " + windows.toString();
+  }
```
JoelCourtney: Please use the usual format that the other expressions use. e.g. from the Rate class:

```java
@Override
  public String prettyPrint(final String prefix) {
    return String.format(
        "(rate-of %s)",
        this.profile.prettyPrint(prefix + "  ")
    );
  }
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
```diff
@@ -32,32 +29,26 @@
...
+              microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
+              duration -> duration.in(Duration.MICROSECONDS)));
   private static final JsonParser<ClosedOpenInterval> intervalP =
       productP
           .field("start", durationP)
           .field("end", durationP)
           .map(Iso.of(
               untuple(ClosedOpenInterval::new),
               $ -> tuple($.start(), $.end())));
-
```
JoelCourtney: (not terribly important) lol why did you remove all the newlines in this file?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
```diff
@@ -176,19 +168,19 @@ record CardinalityGoalDefinition(
     ) implements GoalSpecifier {}
     record GoalAnd(List<GoalSpecifier> goals) implements GoalSpecifier {}
     record GoalOr(List<GoalSpecifier> goals) implements GoalSpecifier {}
+    record GoalApplyWhen(
+        GoalSpecifier goal,
+        Expression<Windows> window
```
JoelCourtney: Please call it `windows` to avoid confusion

JoelCourtney (changes requested):
Love it! I have one tiny change for the WindowsWrapperExpression class, but other than that I think its great.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsWrapperExpression.java
JoelCourtney: Huh actually these functions are never used and the example I gave was wrong. Cool! Should be:

```java
@Override
  public String prettyPrint(final String prefix) {
    return String.format(
        "%s(rate-of %s)",
        prefix,
        this.profile.prettyPrint(prefix + "  ")
    );
  }
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
pranav-super: not sure, I guess just for formatting but I don't remember doing that often, might be a rebase quirk

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
pranav-super: Done!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/models/SchedulingDSL.java
pranav-super: Undone!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsWrapperExpression.java
pranav-super: Done!

JoelCourtney (approved):

PR #226, opened 2022-06-16
[AERIE-1912]  temporal object types don't transfer to worker

dyst5422:
* **Tickets addressed:** AERIE-1912
* **Review:** By commit  

## Description

So because Temporal isn't yet into the spec, its not part of the structuredClone algorithm that is used when transferring data between workers (not sure if it ever will be in the structuredClone, but I would expect it to find its way in there when its formally brought into the spec).

The solution here is to serialize the Temporal objects before sending across to the worker, and deserialize on the worker.

I added two functions - one to serialize the Temporal objects in an object tree, and another to deserialize an object tree with Temporal objects.

## Verification

There is the test added in the first commit for regression testing our worker, and a test with the change to actually test the serializer/deserializer.

## Documentation

No documentation update needed

## Future work

Add a test for durations/instants that are part of the attributes of an activity


dyst5422:
Rewrote the commit history to be clearer (there were some parts split between commits)

command-expansion-server/test/sequence-generation.spec.ts
```diff
@@ -1,166 +1,186 @@
 import { gql, GraphQLClient } from 'graphql-request';
```
pranav-super, line 1: Were the changes to this file largely organizational or were there major logical changes too?

command-expansion-server/test/sequence-generation.spec.ts
```diff
@@ -546,3 +546,61 @@ describe('sequence generation', () => {
   }, 10000);
 
 });
+
```
pranav-super, line 553: Does this test factor in the serialization changes implicitly? In that case, why didn't the pre-existing tests for this functionality suffice?

command-expansion-server/test/sequence-generation.spec.ts
dyst5422: These are all organizational. I did all logical changes in a later commit

command-expansion-server/test/sequence-generation.spec.ts
dyst5422: We did not have a test that utilized the temporal objects from the simulated activity, just Temporal objects that were created inline.

The two lines below show in the expansion where we actually exercise the serialization/deserialization of Temporal objects on the simulated activity

https://github.com/NASA-AMMOS/aerie/pull/226/files/810fba7da16c4cd8db39f5d07200fb506259d2c3..a90d35c0167dffd5f918bc00bc99d0d77abc0e7d#diff-8e57f67494408300790faaeea0a418e551a6d8a64425748d1452c6ff5dad82e9R561-R562

pranav-super (approved):
Makes sense!

goetzrrGit (approved):

PR #227, opened 2022-06-16
[AERIE-1874] Add command expansion test into e2e tests

dyst5422:
* **Tickets addressed:** AERIE-1874
* **Review:** By commit  

## Description

Removed auth for command expansion tests and added an e2etest task for that project.

## Verification

Ran e3etests locally

## Documentation

None

### DIFF
--- a/command-expansion-server/jestGlobalSetup.js
+++ b/command-expansion-server/jestGlobalSetup.js
@@ -1,67 +1,8 @@
 import { config } from 'dotenv';
 config();
-import fs from 'fs';
-import os from 'os';
-import path from 'path';
-import fetch from 'node-fetch';
-import { Writable } from 'stream';
-import readline from 'node:readline';
 
-process.env.MERLIN_GATEWAY_URL = 'http://localhost:9000';
-process.env.MERLIN_GRAPHQL_URL = 'http://localhost:8080/v1/graphql';
+process.env.MERLIN_GATEWAY_URL = process.env.MERLIN_GATEWAY_URL ?? 'http://localhost:9000';
+process.env.MERLIN_GRAPHQL_URL = process.env.MERLIN_GRAPHQL_URL ?? 'http://localhost:8080/v1/graphql';
 
 export default async () => {
-  const ssoFilePath = `${os.tmpdir()}/.aerie/sso.txt`;
-  try {
-    const ssoToken = await fs.promises.readFile(ssoFilePath, 'utf8');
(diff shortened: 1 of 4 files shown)


dyst5422:
Looks like we're missing permissions on a directory creation?
https://github.com/NASA-AMMOS/aerie/runs/6927902592?check_suite_focus=true#step:13:23

```
EACCES: permission denied, mkdir '/usr/src/app/commanding_file_store/aerie/'
```

I'm definitely out of my depth with this one, does one of our CI gods want to chime in/point me in the right direction?

dyst5422:
Apparently, it was just an issue of trying to mount the local code directory as we do for our development docker-compose. Everything looks good now though!

mattdailis (approved):

command-expansion-server/jestGlobalSetup.js
```diff
@@ -11,57 +11,59 @@ process.env.MERLIN_GATEWAY_URL = 'http://localhost:9000';
...
-        if (!this.muted) process.stdout.write(chunk, encoding);
-        callback();
-      },
-    });
-    const readlineInterface = readline.createInterface({
-      input: process.stdin,
-      output: mutableStdout,
-      terminal: true,
-    });
+  // Removing this because it appears we don't need it with auth: none, keeping it here in case we need it in the future though
```
patkenneally: Just a thought: My preference would be to use the Git history to retrieve this if it is indeed needed in the future. My thinking here is that as we get further away from the time when we first commented out code, we slowly lose the context of why  we did so. The Git history with commit messages is a good way to document those decisions. I worry about a code base becoming sprinkled with commented code over time.

patkenneally (approved):

command-expansion-server/jestGlobalSetup.js
dyst5422: Fair point. I guess the risk there is that the existence of this code is forgotten when hidden in the git history. Not sure which point should win out in this case.

dyst5422: I removed it - authenticating for tests isn't something I think we generally want, so I can't imagine this code being worth keeping around

patkenneally: Myself not being very familiar with this particular test set up, is there a way to factor out the login code to say a separate file that can be kept along side and selectively used?

patkenneally: > I removed it 

Okay 👍

PR #229, opened 2022-06-22
[AERIE-1865] User can trigger an analyze only scheduling analysis

adrienmaillard:
* **Tickets addressed:** AERIE-1865
* **Review:** By commit  

## Description

An `analysis_only` field was added to the `scheduling_specification` table in the scheduler server. If the boolean is set to true, the scheduler should only run an analysis, otherwise it would proceed to do a normal scheduling run (insert activities to resolve conflicts).

In the solver, an `analysis_only` boolean is used to bypass branches where activities are inserted in the plan to resolve missing activity conflicts. However, association conflicts are still processed.

There is no change in what scheduling returns. Note however that compared with a regular scheduling run, `inserted_activities` will always be empty.

## Verification

One test has been added to check that the solver was indeed processing only association conflicts.

## Documentation

None. 

## Future work

Update the UI

### DIFF
--- a/scheduler-server/sql/scheduler/tables/scheduling_specification.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_specification.sql
@@ -7,7 +7,7 @@ create table scheduling_specification (
   horizon_start timestamptz not null,
   horizon_end timestamptz not null,
   simulation_arguments jsonb not null,
-
+  analysis_only boolean not null,
   constraint scheduling_specification_synthetic_key
     primary key(id)
 );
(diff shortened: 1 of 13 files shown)


scheduler-server/sql/scheduler/tables/scheduling_specification.sql
```diff
@@ -7,7 +7,7 @@ create table scheduling_specification (
   horizon_start timestamptz not null,
   horizon_end timestamptz not null,
   simulation_arguments jsonb not null,
-
+  analysis_only boolean not null,
```
mattdailis, line 10: To start getting some practice, do we want to write a migration script here?

mattdailis (approved):
Looks good to me

pranav-super (approved):

scheduler-server/sql/scheduler/tables/scheduling_specification.sql
mattdailis: No need to hold up this PR on this point :+1: We'll start practicing migrations in a future PR.

JoelCourtney:
@adrienmaillard Why is analysis_only a field in the specification table? Seems like it should be a argument to the schedule hasura action instead, which would avoid the issue of multiple users doing different actions at the same time.

PR #231, opened 2022-06-22
[AERIE-1910] - Scheduling EDSL Parameterless Activity Fix

pranav-super:
* **Tickets addressed:** AERIE-1910
* **Review:** By commit  

## Description
This is a simple PR to address the concern in AERIE-1910 about parameterless activities allowing bogus parameters. So, if there was an activity, say `EmptyActivity`, admitting no parameters (meaning the correct instantiation would be `EmptyActivity({})`), the EDSL could allow `EmptyActivity({bogus: "param"})` to be passed. To avoid that, the fix this PR takes per @mattdailis's recommendation, is just to not accept any parameters. That is, instead of instantiating with an empty dictionary (`EmptyActivity({})`), now, no dictionary is taken (`EmptyActivity()`).

## Verification
Tests were created! A new test was created in the compilation tester, and the integration tester, and it was tested with a deployment of AERIE to confirm that the code editor could suggest this change as well.

## Documentation
A quick update to the **ActivityTemplate** section was done.

## Future work
None, maybe updating the constraints eDSL to also handle this?

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
@@ -86,11 +86,23 @@ public class TypescriptCodeGenerationService {
     final var result = new ArrayList<String>();
     result.add("const ActivityTemplateConstructors = {");
     for (final var activityTypeCode : activityTypeCodes) {
-      result.add(indent("%s: function %sConstructor(args: {".formatted(activityTypeCode.activityTypeName(), activityTypeCode.activityTypeName())));
-      result.add(indent(indent(generateActivityArgumentTypes(activityTypeCode.parameterTypes()))));
-      result.add(indent("}): %s {".formatted(activityTypeCode.activityTypeName())));
-      result.add(indent(indent("return { activityType: ActivityType.%s, args };".formatted(activityTypeCode.activityTypeName()))));
-      result.add(indent("},"));
+      if(activityTypeCode.parameterTypes().isEmpty()) {
+        result.add(indent("%s: function %sConstructor(): %s {".formatted(
+            activityTypeCode.activityTypeName(),
+            activityTypeCode.activityTypeName(),
+            activityTypeCode.activityTypeName())));
+        result.add(indent(indent("return { activityType: ActivityType.%s, args: {} };".formatted(activityTypeCode.activityTypeName()))));
+        result.add(indent("},"));
+      }
+      else {
+        result.add(indent("%s: function %sConstructor(args: {".formatted(
+            activityTypeCode.activityTypeName(),
+            activityTypeCode.activityTypeName())));
+        result.add(indent(indent(generateActivityArgumentTypes(activityTypeCode.parameterTypes()))));
+        result.add(indent("}): %s {".formatted(activityTypeCode.activityTypeName())));
+        result.add(indent(indent("return { activityType: ActivityType.%s, args };".formatted(activityTypeCode.activityTypeName()))));
+        result.add(indent("},"));
+      }
     }
     result.add("};");
     return joinLines(result);
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BananaNapActivity.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BananaNapActivity.java
@@ -1,0 +1,20 @@
+package gov.nasa.jpl.aerie.banananation.activities;
+
+import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+
+/**
+ * Nap time [banana style]!!!!
+ * This activity has no effect :)
+ *
+ * @subsystem fruit
+ * @contact Jane Doe
+ */
+@ActivityType("BananaNap")
+public final class BananaNapActivity {
+  @EffectModel
+  public void run(final Mission mission) {
+
+  }
+}
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/package-info.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/package-info.java
@@ -16,10 +16,12 @@
 @WithActivityType(DecomposingActivity.ChildActivity.class)
 @WithActivityType(DecomposingActivity.GrandchildActivity.class)
 @WithActivityType(BakeBananaBreadActivity.class)
+@WithActivityType(BananaNapActivity.class)
 
 package gov.nasa.jpl.aerie.banananation;
 
 import gov.nasa.jpl.aerie.banananation.activities.BakeBananaBreadActivity;
+import gov.nasa.jpl.aerie.banananation.activities.BananaNapActivity;
 import gov.nasa.jpl.aerie.banananation.activities.BiteBananaActivity;
 import gov.nasa.jpl.aerie.banananation.activities.ChangeProducerActivity;
 import gov.nasa.jpl.aerie.banananation.activities.DecomposingActivity;
@@ -28,6 +30,7 @@ import gov.nasa.jpl.aerie.banananation.activities.LineCountBananaActivity;
 import gov.nasa.jpl.aerie.banananation.activities.ParameterTestActivity;
 import gov.nasa.jpl.aerie.banananation.activities.PeelBananaActivity;
 import gov.nasa.jpl.aerie.banananation.activities.PickBananaActivity;
+import gov.nasa.jpl.aerie.banananation.activities.BananaNapActivity;
 import gov.nasa.jpl.aerie.banananation.activities.ThrowBananaActivity;
 import gov.nasa.jpl.aerie.contrib.serialization.rulesets.BasicValueMappers;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.MissionModel;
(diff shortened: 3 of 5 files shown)


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -86,11 +86,35 @@ private static String generateActivityTemplateConstructors(final Iterable<Activi
     final var result = new ArrayList<String>();
     result.add("const ActivityTemplateConstructors = {");
     for (final var activityTypeCode : activityTypeCodes) {
-      result.add(indent("%s: function %sConstructor(args: {".formatted(activityTypeCode.activityTypeName(), activityTypeCode.activityTypeName())));
-      result.add(indent(indent(generateActivityArgumentTypes(activityTypeCode.parameterTypes()))));
-      result.add(indent("}): %s {".formatted(activityTypeCode.activityTypeName())));
-      result.add(indent(indent("return { activityType: ActivityType.%s, args };".formatted(activityTypeCode.activityTypeName()))));
-      result.add(indent("},"));
+      if(activityTypeCode.parameterTypes().size() == 0) {
```
mattdailis: Prefer `.isEmpty()` to `.size() == 0`
```suggestion
      if(activityTypeCode.parameterTypes().isEmpty()) {
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -86,11 +86,35 @@ private static String generateActivityTemplateConstructors(final Iterable<Activi
...
+        /*
+        Relax: function RelaxConstructor(): Relax {
```
mattdailis: This comment looks like it was helpful when writing this code, but I'm worried it will mostly be confusing in the future. For one, it's already out of date (the returned dictionary is missing the `args` key) and it will only get more out of date as we continue to make changes to this code. I suggest removing these comments, but making sure that the `TypescriptCodeGenerationTest` has examples of activities with and without parameters.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
```diff
@@ -697,6 +698,31 @@ export default (): Goal => {
     assertEquals(Map.of("peelDirection", SerializedValue.of("fromStem")), peelBanana.args());
   }
 
+  @Test
+  void testEmptyActivity() {
```
mattdailis: What is this test testing that isn't already covered by the new unit tests? I think we can remove this one.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
```diff
@@ -0,0 +1,27 @@
...
+ * @subsystem fruit
+ * @contact Jane Doe
+ */
+@ActivityType("Relax")
+public final class RelaxActivity {
+  private static final double RELAX_DURATION = 1.0;
+
+  @EffectModel
+  public void run(final Mission mission) {
+    //delay(Duration.of(RELAX_DURATION, Duration.HOURS));
```
mattdailis: I think we can delete this line, as well as the `RELAX_DURATION` field.

mattdailis (changes requested):
Generally looks solid to me. Just a few minor code cleanup requests.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
```diff
@@ -0,0 +1,27 @@
+package gov.nasa.jpl.aerie.banananation.activities;
+
+import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Parameter;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Validation;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.util.List;
```
skovati:
```suggestion
```
Looks like these imports are unused.

skovati (changes requested):
Looks pretty good otherwise!

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
pranav-super: I guess just that none of them test an empty activity as BananaNation doesn't have any. I thought it would be good to prove that this update doesn't break anything in the integration tests but I don't disagree that it's extraneous. I'll go ahead and remove it!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pranav-super: Sounds good. I wasn't sure about how to use `delay()` for this model so I included it as a placeholder but I can get rid of it. It doesn't seem like any of the activities in BananaNation have a duration, which I thought might be useful to test but I'm unsure how to do that.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pranav-super: Facts

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
```diff
@@ -0,0 +1,19 @@
...
+
+import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+
+/**
+ * Relax? I needed something with no parameters this was the best one I could come up with
+ *
+ * @subsystem fruit
+ * @contact Jane Doe
```
pcrosemurgy: Pedantic since these `@` javadoc lines are harmless but I don't see a compelling reason to check them into VCS

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+@ActivityType("Relax")
+public final class RelaxActivity {
+  @EffectModel
+  public void run(final Mission mission) {
+    return;
```
pcrosemurgy: Can safely remove this `return`

pcrosemurgy (approved):

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pcrosemurgy: Also, this could be called `EffectlessActivity` if you're looking for another name

skovati: I agree the name could be more descriptive, preferably something banana / monke themed...

pranav-super: So should I just remove the @'s? It seems like all the other activities have them so I added them for the sake of uniformity. I can rename the activity to something better too.

pcrosemurgy: Ahh true! Didn't realize the others do. Nice, in that case I think what you have is the right thing to do. :)

Twisol: FYI, the *reason* those nonstandard Javadoc tags are present is because we originally (and, I think, still?) wanted to extract that information out at annotation processing time, so it could be flowed out to planners via the UI.

That hasn't happened yet, so these tags don't do anything. Their presence is a good reminder that we still have yet to implement that functionality... but only, of course, if we knew it was a thing to begin with 😁

pranav-super: How's BananaNapActivity

mattdailis (approved):

skovati (approved):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
pranav-super: Fixed!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
pranav-super: I removed it!

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
pranav-super: Removed!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pranav-super: Removed!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pranav-super: Fixed!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/RelaxActivity.java
pranav-super: Removed!

PR #232, opened 2022-06-22
[AERIE-1911] - Range check recurrence interval

pranav-super:
* **Tickets addressed:** AERIE-1911
* **Review:** By commit  

## Description
The focus of the PR is to range check recurrence intervals, that is, not allow negative recurrence intervals. To do this, we catch the error within the builder of the `RecurrenceGoal` (specifically in `fill()`). Also, since the RecurrenceGoal stores its interval as a window (with the start representing a minimum recurrence interval, the end representing a maximum, a feature not yet supported), that was replaced with a record `MinMaxAllowableRecurrenceInterval` just to make it more explicit.

## Verification
A new test was added to the existing recurrence goal tests - TestRecurrenceGoal.
Similarly, another test was added to the integration tests, to verify the entire process worked.

## Documentation
Just an update to the recurrence goal section.

## Future work
This might play weird with the AnalyzeWhen ([AERIE-1880]) PR, just because of the `MinMaxAllowableRecurrenceInterval` fix.

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
@@ -20,6 +20,14 @@ import java.util.List;
 public class RecurrenceGoal extends ActivityTemplateGoal {
 
   /**
+   * A tuple-like type to store two allowable durations for the recurrence interval - a minimum and a maximum,
+   * which will eventually be supported by the scheduler. Default behavior for now should be to set min and max equal.
+   * @param min the minimum allowable duration to use (for the recurrence interval)
+   * @param max the maximum allowable duration to use (for the recurrence interval)
+   */
+  public record MinMaxAllowableRecurrenceInterval(Duration min, Duration max) {}
+
+  /**
    * the builder can construct goals piecemeal via a series of method calls
    */
   public static class Builder extends ActivityTemplateGoal.Builder<Builder> {
@@ -37,11 +45,11 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
      * @return this builder, ready for additional specification
      */
     public Builder repeatingEvery(Duration interval) {
-      this.every = Window.at(interval);
+      this.every = new MinMaxAllowableRecurrenceInterval(interval, interval);
       return getThis();
     }
 
-    protected Window every;
+    protected MinMaxAllowableRecurrenceInterval every;
 
     /**
      * {@inheritDoc}
@@ -69,8 +77,13 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
       super.fill(goal);
 
       if (every == null) {
-        throw new IllegalArgumentException(
-            "creating recurrence goal requires non-null \"every\" duration interval");
+        throw new IllegalArgumentException("creating recurrence goal requires non-null \"every\" duration interval");
+      }
+      if (every.min.isNegative()) {
+        throw new IllegalArgumentException("Duration passed to RecurrenceGoal as the goal's minimum recurrence interval cannot be negative!");
+      }
+      else if (every.max.isNegative()) {
+        throw new IllegalArgumentException("Duration passed to RecurrenceGoal as the goal's maximum recurrence interval cannot be negative!");
       }
       goal.recurrenceInterval = every;
 
@@ -111,7 +124,7 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
       //check if the inter-activity gap is too large
       //REVIEW: should do any check based on min gap duration?
       final var strideDur = actStartT.minus(prevStartT);
-      if (strideDur.compareTo(recurrenceInterval.end) > 0) {
+      if (strideDur.compareTo(recurrenceInterval.max) > 0) {
         //fill conflicts for all the missing activities in that long span
         conflicts.addAll(makeRecurrenceConflicts(prevStartT, actStartT));
 
@@ -143,10 +156,11 @@ public class RecurrenceGoal extends ActivityTemplateGoal {
 
   /**
    * the allowable range of durations over which a target activity must repeat
+   * Of type: MinMaxAllowableRecurrenceInterval
    *
    * REVIEW: need to work out semantics of min on recurrenceInterval
    */
-  protected Window recurrenceInterval;
+  protected MinMaxAllowableRecurrenceInterval recurrenceInterval;
 
   /**
    * creates conflicts for missing activities in the provided span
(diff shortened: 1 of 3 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -72,6 +80,9 @@ protected RecurrenceGoal fill(RecurrenceGoal goal) {
         throw new IllegalArgumentException(
             "creating recurrence goal requires non-null \"every\" duration interval");
       }
+      if (every.min.isNegative() || every.max.isNegative()) {
+        throw new IllegalArgumentException("Duration passed to RecurrenceGoal as the goal's recurrence interval cannot be negative!");
```
mattdailis: One question on my mind is does/should this fail the whole scheduling run, or only the current goal?

pranav-super: This would fail the entire run, I'm not sure if it should just fail the goal. I think that makes sense but it might be a good idea to fail entirely just so that no time is wasted running scheduling prior to the goals actually being properly specified. What are your thoughts?

dyst5422: This is literally the conversation we're having here https://jpl.slack.com/archives/C0163E42UBF/p1656012405258839

pranav-super: Oh neat, then this seems like its the correct behavior per [this message](https://jpl.slack.com/archives/C0163E42UBF/p1656016362871939?thread_ts=1656012405.258839&cid=C0163E42UBF).

patkenneally: :small_red_triangle_down: @pranav-super can you add a paraphrasing of the idea in the message to which you are referring? Aerie is a public open-source project and external viewers cannot see a private Slack conversation.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
```diff
@@ -48,4 +49,32 @@ public void testRecurrence() {
     assertTrue(TestUtility.activityStartingAtTime(plan,Duration.of(16, Duration.SECONDS), activityType));
   }
 
+  @Test
+  public void testRecurrenceNegative() {
+    var planningHorizon = new PlanningHorizon(TestUtility.timeFromEpochSeconds(0),TestUtility.timeFromEpochSeconds(20));
+    final var fooMissionModel = SimulationUtility.getFooMissionModel();
+    Problem problem = new Problem(fooMissionModel, planningHorizon, new SimulationFacade(
```
patkenneally: :large_orange_diamond: There's some small formatting tweaks to be done in this method. E.g. You can align the args to `Problem` one each on a new line. Drop the second `Duration` arg in `.forAllTimeIn` to a new line to reduce the line length. Use a text block for the message in your assertion.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -19,6 +19,14 @@
  */
 public class RecurrenceGoal extends ActivityTemplateGoal {
 
+  /**
+   * A tuple-like type to store two allowable durations for the recurrence interval - a minimum and a maximum,
+   * which will eventually be supported by the scheduler. Default behavior for now should be to set min and max equal.
+   * @param min the minimum allowable duration to use (for the recurrence interval)
+   * @param max the maximum allowable duration to use (for the recurrence interval)
+   */
+  public record MinMaxAllowableRecurrenceInterval(Duration min, Duration max) {};
```
patkenneally: :large_orange_diamond: Semi colon can be removed.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -19,6 +19,14 @@
  */
 public class RecurrenceGoal extends ActivityTemplateGoal {
 
+  /**
```
patkenneally, line 22: Thanks for adding Javadocs 👍

patkenneally (changes requested):

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/TestRecurrenceGoal.java
Twisol: I suspect all of the variables on the highlighted lines can be made `final var`, too, as well as the `RecurrenceGoal` some lines down.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
pranav-super: Context: one of the customers of the AERIE project was having a conversation with the developers about failure behavior. They said, 
> For runs that include rules that don't compile, I don't think it should even start. 
> (Different than rules that fail to find any way to satisfy themselves in the plan)
> Identifying the failed compile rules will help so they can be chased down of course

Furthermore, it appears this was existing behavior and this is the correct way to handle issues per the `failure_reason` column, which gets populated when the scheduler bails with an exception, see [here](https://github.com/NASA-AMMOS/aerie/blob/8bb7be8ed6ae2ebb7c1841b8a1a58910119b623f/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java), on line 149.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -69,8 +77,10 @@ protected RecurrenceGoal fill(RecurrenceGoal goal) {
       super.fill(goal);
 
       if (every == null) {
-        throw new IllegalArgumentException(
-            "creating recurrence goal requires non-null \"every\" duration interval");
+        throw new IllegalArgumentException("creating recurrence goal requires non-null \"every\" duration interval");
+      }
+      if (every.min.isNegative() || every.max.isNegative()) {
+        throw new IllegalArgumentException("Duration passed to RecurrenceGoal as the goal's recurrence interval cannot be negative!");
```
JoelCourtney: Long term it would be nice for this error to specify which of the bounds are negative.

JoelCourtney (approved):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
pranav-super: Done!

patkenneally (approved):

PR #233, opened 2022-06-23
[HOTFIX] banananation cmd dict model congruency

dyst5422:
* **Review:** By commit  

## Description

The Peel Banana command expected an enum of 'fromStem' | 'fromBase' |
'fromCenter' while the model produced a string with a custom validator.

This changes the model to an enum of 'fromStem' | 'fromTip' and makes
the command match that.

Additional copy changes to the banananation command dictionary (spelling, copy/paste errors, name)

## Verification

Inspection & used the new command dictionary and model in an expansion

## Documentation

None

## Future work

None

### DIFF
--- a/command-expansion-server/cdict/command_banananation.xml
+++ b/command-expansion-server/cdict/command_banananation.xml
@@ -18,12 +18,28 @@
         <enum_table name="string">
             <values>
                 <enum symbol="fromStem" numeric="0" />
-                <enum symbol="fromBase" numeric="1" />
-                <enum symbol="fromCenter" numeric="2" />
+                <enum symbol="fromTip" numeric="1" />
             </values>
         </enum_table>
     </enum_definitions>
     <command_definitions>
+        <fsw_command opcode="0xFFFF" stem="ECHO" class="FSW">
+          <arguments>
+            <var_string_arg name="echo_string" prefix_bit_length="8" max_bit_length="1024" cts_item_type="dynamic_arg">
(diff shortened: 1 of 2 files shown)


command-expansion-server/cdict/command_banananation.xml
```diff
@@ -70,9 +86,9 @@
                     <range_of_values>
                         <include min="0" max="100" />
                     </range_of_values>
-                    <description>Number of bananans to grow</description>
+                    <description>Number of bananas to grow</description>
                 </unsigned_arg>
-                <unsigned_arg name="duration" bit_length="8" units="none">
+                <unsigned_arg name="durationS" bit_length="8" units="none">
```
skovati:
```suggestion
                <unsigned_arg name="durations" bit_length="8" units="none">
```

skovati (approved):
Not super familiar with command dicts, but looks good to me. Nice catch!

command-expansion-server/cdict/command_banananation.xml
dyst5422: This was intended as duration-in-seconds, not a plural of duration. Not sure how better to annotate that in the naming. I know durationMs is super obvious, but less human friendly than seconds.

skovati: Ahh gotcha, `durationSecs` or something could be clearer at first glance, but I also think it's fine as is then.

camargo (approved):

PR #234, opened 2022-06-23
[FIX] Root-finding issue

adrienmaillard:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

@mattdailis and @pranav-super encountered an issue with root-finding. The case involved a 0-duration uncontrollable activity and a non-singleton start interval (and no end-time constraint).

Although the secant method might diverge or not find a zero, a review of the code has shown room for improvements and necessary fixes. I am gonna comment on the changes directly below.

I have replicated these changes to the `double` version of the algorithm. 

## Verification

A test has been added to illustrate the specific case where no solution was found. Other tests have been modified.  

## Documentation

None. 

## Future work

None.

### DIFF
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/ZeroDurationUncontrollableActivity.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/ZeroDurationUncontrollableActivity.java
@@ -1,0 +1,14 @@
+package gov.nasa.jpl.aerie.foomissionmodel.activities;
+
+import gov.nasa.jpl.aerie.foomissionmodel.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+
+import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.delay;
+
+@ActivityType("ZeroDurationUncontrollableActivity")
+public final class ZeroDurationUncontrollableActivity {
+  @EffectModel
+  public void run(final Mission mission) {
+  }
+}
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
@@ -21,10 +21,10 @@ public class EquationSolvingAlgorithms {
 
   public interface BracketingAlgorithm<T>{
     /**
-     * Solves f(x) in [y - toleranceYLow, y + toleranceYHigh] for x in [dxLow, dxHigh].
+     * Solves f(x) in [y - toleranceYLow, y + toleranceYHigh] for x in [xLow, xHigh].
      * the sign of f(bracket1) must be different from the sign of f(bracket2)
      *
-     * dxLow {@literal <} dxHigh
+     * xLow {@literal <} xHigh
      * toleranceYLow > 0, toleranceYHigh > 0
      */
     RootFindingResult<T> findRoot(
@@ -34,16 +34,16 @@ public class EquationSolvingAlgorithms {
         T y,
         T toleranceYLow,
         T toleranceYHigh,
-        T dxLow,
-        T dxHigh,
+        T xLow,
+        T xHigh,
         int maxNbIterations) throws ExceededMaxIterationException, WrongBracketingException,
                                     NoSolutionException;
   }
   /**
-   * Solves f(x) = y for x in [dxLow, dxHigh] with confidence interval [yLow, yHigh] around y such that we stop when
+   * Solves f(x) = y for x in [xLow, xHigh] with confidence interval [yLow, yHigh] around y such that we stop when
    * value y is in [y-yLow, y+yHigh].
    * x0 and x1 are initial guesses for x, they must be close to the solution to avoid diverging
-   * It is considered that algorithm is diverging when the iterated value of x goes out of [dxLow, dxHigh].
+   * It is considered that algorithm is diverging when the iterated value of x goes out of [xLow, xHigh].
    */
   public interface SecantAlgorithm<T>{
     RootFindingResult<T> findRoot(Function<T> f,
@@ -52,17 +52,17 @@ public class EquationSolvingAlgorithms {
                                   T y,
                                   T toleranceYLow,
                                   T toleranceYHigh,
-                                  T dxLow,
-                                  T dxHigh,
-                                  int maxNbIterations) throws ZeroOrInfiniteDerivativeException, DivergenceException,
+                                  T xLow,
+                                  T xHigh,
+                                  int maxNbIterations) throws ZeroDerivativeException, InfiniteDerivativeException, DivergenceException,
                                                               ExceededMaxIterationException, NoSolutionException;
   }
 
   /**
-   * Solves f(x) in [y - yLow, y + yHigh] for x in [dxLow, dxHigh] with confidence interval [yLow, yHigh] around y such that we stop when
+   * Solves f(x) in [y - yLow, y + yHigh] for x in [xLow, xHigh] with confidence interval [yLow, yHigh] around y such that we stop when
    * value y is in [y - yLow, y + yHigh].
    * x0 is a first guess for x, it must be close to the solution to avoid diverging.
-   * It is considered that algorithm is diverging when the iterated value of x goes out of [dxLow, dxHigh].
+   * It is considered that algorithm is diverging when the iterated value of x goes out of [xLow, xHigh].
    */
   public interface NewtonAlgorithm<T>{
     RootFindingResult<T> findRoot(FunctionWithDerivative<T> f,
@@ -70,9 +70,10 @@ public class EquationSolvingAlgorithms {
                                   T y,
                                   T toleranceYLow,
                                   T toleranceYHigh,
-                                  T dxLow,
-                                  T dxHigh,
-                                  int maxNbIterations) throws DivergenceException, ZeroOrInfiniteDerivativeException,
+                                  T xLow,
+                                  T xHigh,
+                                  int maxNbIterations) throws DivergenceException,
+                                                              ZeroDerivativeException,
                                                               ExceededMaxIterationException;
   }
 
@@ -85,8 +86,12 @@ public class EquationSolvingAlgorithms {
     T derivativeAt(T x);
   }
 
-  public static class ZeroOrInfiniteDerivativeException extends Exception{
-    public ZeroOrInfiniteDerivativeException() {}
+  public static class ZeroDerivativeException extends Exception{
+    public ZeroDerivativeException() {}
+  }
+
+  public static class InfiniteDerivativeException extends Exception{
+    public InfiniteDerivativeException() {}
   }
 
   public static class DivergenceException extends Exception{
@@ -134,12 +139,12 @@ public class EquationSolvingAlgorithms {
         Double y,
         Double toleranceYLow,
         Double toleranceYHigh,
-        Double dxLow,
-        Double dxHigh,
+        Double xLow,
+        Double xHigh,
         int maxNbIterations)
     throws ExceededMaxIterationException, WrongBracketingException, NoSolutionException
     {
-      if (bracket1 < dxLow || bracket2 > dxHigh) {
+      if (bracket1 < xLow || bracket2 > xHigh) {
         throw new WrongBracketingException("brackets are out of prescribed root domain");
       }
       if (Math.abs(toleranceYLow - toleranceYHigh) > 1E-6) {
@@ -172,12 +177,12 @@ public class EquationSolvingAlgorithms {
         Double y,
         Double toleranceYLow,
         Double toleranceYHigh,
-        Double dxLow,
-        Double dxHigh,
+        Double xLow,
+        Double xHigh,
         int maxNbIterations)
     throws WrongBracketingException, ExceededMaxIterationException
     {
-      if (bracket1 < dxLow || bracket1 > dxHigh || bracket2 < dxLow || bracket2 > dxHigh) {
+      if (bracket1 < xLow || bracket1 > xHigh || bracket2 < xLow || bracket2 > xHigh) {
         throw new WrongBracketingException("Brackets are out of prescribed root domain");
       }
       final var ff = new Function<Double>(){
@@ -226,12 +231,12 @@ public class EquationSolvingAlgorithms {
         Duration y,
         Duration toleranceYLow,
         Duration toleranceYHigh,
-        Duration dxLow,
-        Duration dxHigh,
+        Duration xLow,
+        Duration xHigh,
         int maxNbIterations)
     throws WrongBracketingException, ExceededMaxIterationException
     {
-      if (bracket1.shorterThan(dxLow) || bracket1.longerThan(dxHigh) || bracket2.shorterThan(dxLow) || bracket2.longerThan(dxHigh)) {
+      if (bracket1.shorterThan(xLow) || bracket1.longerThan(xHigh) || bracket2.shorterThan(xLow) || bracket2.longerThan(xHigh)) {
         throw new WrongBracketingException("Brackets are out of prescribed root domain");
       }
       final var ff = new Function<Duration>(){
@@ -280,10 +285,10 @@ public class EquationSolvingAlgorithms {
         Double y,
         Double toleranceYLow,
         Double toleranceYHigh,
-        Double dxLow,
-        Double dxHigh,
+        Double xLow,
+        Double xHigh,
         int maxNbIterations)
-    throws ZeroOrInfiniteDerivativeException, ExceededMaxIterationException
+    throws ZeroDerivativeException, InfiniteDerivativeException, ExceededMaxIterationException
     {
       var history = new ArrayList<Pair<Double, Double>>();
 
@@ -306,20 +311,19 @@ public class EquationSolvingAlgorithms {
       var ff_x_nminus1 = ff.valueAt(x_nminus1);
       if (ff_x_nminus1 >= - toleranceYLow &&
           ff_x_nminus1 <= toleranceYHigh
-          && (x_nminus1 >= dxLow && x_nminus1 <= dxHigh)) {
+          && (x_nminus1 >= xLow && x_nminus1 <= xHigh)) {
         return new RootFindingResult<>(x_nminus1, ff_x_nminus1 + y, new History<>(history));
       }
       var ff_x_n = ff.valueAt(x_n);
       if (ff_x_n >= - toleranceYLow &&
           ff_x_n <= toleranceYHigh
-          && (x_n >= dxLow && x_n <= dxHigh)) {
+          && (x_n >= xLow && x_n <= xHigh)) {
         return new RootFindingResult<>(x_n, ff_x_n + y, new History<>(history));
       }
       for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
         final var localDerivative = (ff_x_n - ff_x_nminus1) / (x_n - x_nminus1);
-        if (localDerivative == 0 || Double.isNaN(localDerivative)) {
-          throw new ZeroOrInfiniteDerivativeException();
-        }
+        if (localDerivative == 0) throw new ZeroDerivativeException();
+        if (Double.isNaN(localDerivative)) throw new InfiniteDerivativeException();
         x_nminus1 = x_n;
         ff_x_nminus1 = ff_x_n;
         x_n = x_n - (ff_x_nminus1 / localDerivative);
@@ -327,7 +331,7 @@ public class EquationSolvingAlgorithms {
         history.add(Pair.of(x_n, ff_x_n));
         if (ff_x_n >= - toleranceYLow &&
             ff_x_n <= toleranceYHigh
-        && (x_n >= dxLow && x_n <= dxHigh)) {
+        && (x_n >= xLow && x_n <= xHigh)) {
           return new RootFindingResult<>(x_n, ff_x_n + y, new History<>(history));
         }
       }
@@ -346,10 +350,10 @@ public class EquationSolvingAlgorithms {
         Duration y,
         Duration toleranceYLow,
         Duration toleranceYHigh,
-        Duration dxLow,
-        Duration dxHigh,
+        Duration xLow,
+        Duration xHigh,
         int maxNbIterations)
-    throws ZeroOrInfiniteDerivativeException, DivergenceException, ExceededMaxIterationException, NoSolutionException
+    throws ZeroDerivativeException, InfiniteDerivativeException, DivergenceException, ExceededMaxIterationException, NoSolutionException
     {
       final var history = new ArrayList<Pair<Duration, Duration>>();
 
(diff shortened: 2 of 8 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -354,14 +365,6 @@ public boolean isApproximation() {
...
       };
 
-      if (x0.isEqualTo(x1)) {
-        //particular : only one start possible, derivative would be 0
-        final var val = ff.valueAt(x0);
-        if (val.noShorterThan(Duration.negate(toleranceYLow)) && val.noLongerThan(toleranceYHigh)) {
-          return new RootFindingResult<>(x0, val.plus(y),null);
-        }
-        throw new NoSolutionException();
-      }
```
adrienmaillard, line 364: These lines have been replaced more adequately by the following block.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -371,12 +374,25 @@ public boolean isApproximation() {
...
+      var ff_x_n = ff.valueAt(x_n);
+      if (ff_x_n.noShorterThan(Duration.negate(toleranceYLow)) && ff_x_n.noLongerThan(toleranceYHigh)) {
+        return new RootFindingResult<>(x_n, ff_x_n.plus(y), null);
+      }
+      if (x0.isEqualTo(x1)) {
+        //particular : only one start possible, derivative would be 0
+        throw new NoSolutionException();
+      }
+      history.add(Pair.of(x0, ff_x_nminus1));
+      history.add(Pair.of(x1, ff_x_n));
```
adrienmaillard: We check whether the initial bounds might satisfy the exit criteria. After these checks, we can be sure that if the two bounds are the same, the derivative will be 0, and thus throw an exception.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -385,14 +401,11 @@ public boolean isApproximation() {
         //localDerivative has been computed with what is now xn_m1 and xn_m2
         x_n_double = x_n_double - (ff_x_nminus1.in(Duration.MICROSECONDS) / localDerivative);
         x_n = Duration.of((long) x_n_double, Duration.MICROSECONDS);
-        //outside of domain, diverging
-        if (x_n_double < dxLow_long || x_n_double > dxHigh_long) {
-          throw new DivergenceException("Looking for root out of prescribed domain :[" + dxLow + "," + dxHigh + "]",
-                                        new History<>(history));
-        }
```
adrienmaillard, line 392: This was wrong, the secant method might go outside of the initial domain to finally come back. This is not sufficient to declare there is divergence.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -385,14 +401,11 @@ public boolean isApproximation() {
...
         ff_x_n = ff.valueAt(x_n);
-        history.put(x_n, ff_x_n);
-        if (ff_x_n.noShorterThan(Duration.negate(toleranceYLow)) && ff_x_n.noLongerThan(toleranceYHigh)) {
+        history.add(Pair.of(x_n, ff_x_n));
+        if (ff_x_n.noShorterThan(Duration.negate(toleranceYLow)) &&
+            ff_x_n.noLongerThan(toleranceYHigh) &&
+            (x_n_double >= dxLow_long && x_n_double <= dxHigh_long)){
```
adrienmaillard: However, the final solution needs to be in the given bounds which is why this check is added here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
Twisol: This sounds like it would make a great comment inline with the code. 🤔

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -82,12 +85,11 @@ public void testNonLinear(){
...
-    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT40M11S"), planningHorizon.fromStart("PT41M40S"), problem.getActivityType("SolarPanelNonLinear")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT11M40S"), planningHorizon.fromStart("PT16M40S"), problem.getActivityType("SolarPanelNonLinear")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT28M20S"), planningHorizon.fromStart("PT33M20S"), problem.getActivityType("SolarPanelNonLinear")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT0S"), planningHorizon.fromStart("PT1M29S"), problem.getActivityType("SolarPanelNonLinear")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT16M40S"), planningHorizon.fromStart("PT18M9S"), problem.getActivityType("SolarPanelNonLinear")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT33M20S"), planningHorizon.fromStart("PT34M49S"), problem.getActivityType("SolarPanelNonLinear")));
```
adrienmaillard, line 94: Because the initial bounds are now checked before iteration, the timings of activities have changed.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -132,12 +134,44 @@ public void testTimeDependent(){
...
+    problem.setGoals(List.of(coexistenceTriangle));
+    final var initialPlan = new PlanInMemory();
+    initialPlan.add(solarPanelActivityTrapezoidal);
+    problem.setInitialPlan(initialPlan);
+
+    final var solver = new PrioritySolver(problem);
+    final var plan = solver.getNextSolution().get();
+    solver.printEvaluation();
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT0.000004S"), planningHorizon.fromStart("PT0.000004S"), problem.getActivityType("ZeroDurationUncontrollableActivity")));
+  }
```
adrienmaillard: The bug happened in the situation in which there was no end timing constraint but an interval start constraint in addition to a 0-duration uncontrollable activity.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Agreed. You are so fast, @Twisol :)

Twisol: I have email notifications on and I'm currently a passenger in a vehicle. I've got everything going for me >:)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
pcrosemurgy: This seems like another good comment to leave in-code too :)

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
pcrosemurgy: So prior to this PR `ZeroDurationUncontrollableActivity` would not be placed as the first planned activity?

mattdailis:
FYI, hoping to dust off this PR and give it a review today :+1:

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -132,12 +134,44 @@ public void testTimeDependent(){
...
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT11M57S"), planningHorizon.fromStart("PT16M40S"), problem.getActivityType("SolarPanelNonLinearTimeDependent")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT28M34S"), planningHorizon.fromStart("PT33M20S"), problem.getActivityType("SolarPanelNonLinearTimeDependent")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT33M20S"), planningHorizon.fromStart("PT36M47S"), problem.getActivityType("SolarPanelNonLinearTimeDependent")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT0S"), planningHorizon.fromStart("PT2M21S"), problem.getActivityType("SolarPanelNonLinearTimeDependent")));
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT16M40S"), planningHorizon.fromStart("PT17M18S"), problem.getActivityType("SolarPanelNonLinearTimeDependent")));
   }
 
+  @Test
+  public void testBug(){
+    final var solarPanelActivityTrapezoidal = new ActivityInstance(problem.getActivityType("ControllableDurationActivity"),
```
mattdailis: 🔶 The copy-pasted variable names are making it a bit hard for me to understand the test here.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -132,12 +134,44 @@ public void testTimeDependent(){
...
+
+    problem.setGoals(List.of(coexistenceTriangle));
+    final var initialPlan = new PlanInMemory();
+    initialPlan.add(solarPanelActivityTrapezoidal);
+    problem.setInitialPlan(initialPlan);
+
+    final var solver = new PrioritySolver(problem);
+    final var plan = solver.getNextSolution().get();
+    solver.printEvaluation();
+    assertTrue(TestUtility.containsActivity(plan, planningHorizon.fromStart("PT0.000004S"), planningHorizon.fromStart("PT0.000004S"), problem.getActivityType("ZeroDurationUncontrollableActivity")));
```
mattdailis: 🔵 Can we put these arguments on separate lines? I'm having trouble reading them.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -371,12 +374,25 @@ public boolean isApproximation() {
...
+        throw new NoSolutionException();
+      }
+      history.add(Pair.of(x0, ff_x_nminus1));
+      history.add(Pair.of(x1, ff_x_n));
       for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
         //(f(xn) - f(xn_m1)) / (xn - xn_m1)
         final double localDerivative =
             (float) (ff_x_n.minus(ff_x_nminus1)).in(Duration.MICROSECONDS) / (x_n_double - x_nminus1_double);
-        if (localDerivative == 0) {
+        if (localDerivative == 0 || Double.isNaN(localDerivative)) {
```
mattdailis: 🔶 Should we add a new `InfiniteDerivative` exception, or maybe change the name of the `ZeroDerivative` exception to encompass both zero and NaN?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -301,7 +303,18 @@ public boolean isApproximation() {
...
+      }
+      var ff_x_n = ff.valueAt(x_n);
+      if (ff_x_n >= - toleranceYLow &&
+          ff_x_n <= toleranceYHigh
+          && (x_n >= dxLow && x_n <= dxHigh)) {
+        return new RootFindingResult<>(x_n, ff_x_n + y, new History<>(history));
+      }
       for (int nbIt = 0; nbIt < maxNbIterations; nbIt++) {
         final var localDerivative = (ff_x_n - ff_x_nminus1) / (x_n - x_nminus1);
         if (localDerivative == 0) {
```
mattdailis: 🔶 Is there a reason we do a `NaN` check in the `Duration` version, but not in the `double` version?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
mattdailis: Correct, @pcrosemurgy - the issue that came up was if you have a scheduling goal that just says "Place Activity B any time after the end of Activity A" - this algorithm would throw a `DivergenceException` (which would result in the activity not being placed).

The issue comes up when the start interval is much more constrained than the end interval. The algorithm tries to find a solution where the activity ends within a certain tolerance of the midpoint of the end interval. If the end interval is very large, there may not be any solution that respects the start interval...

(Now that I've typed that out, I think this algorithm will still break if for some reason neither of the first two endpoints is a valid solution 🤔 but maybe that case will never happen? I'll need to ponder...)

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: Updated the names.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: There was none, I added it to the double version.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: I updated its name to `ZeroOrInfiniteDerivativeException`

mattdailis (approved):
Looks good to me 👍 This is an improvement over what we had before for constant-duration activities.

examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/ZeroDurationUncontrollableActivity.java
```diff
@@ -0,0 +1,14 @@
...
+import gov.nasa.jpl.aerie.foomissionmodel.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+
+import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.delay;
+
+@ActivityType("ZeroDurationUncontrollableActivity")
+public final class ZeroDurationUncontrollableActivity {
+  @EffectModel
+  public void run(final Mission mission) {
```
pranav-super, line 12: By not including a duration, does that make the duration both uncontrollable and set it to zero, implicitly?

pranav-super, line 12: By not including a duration, does that make the duration both uncontrollable and set it to zero, implicitly?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -134,12 +136,47 @@ public void testTimeDependent(){
...
+        .withTimingPrecision(Duration.of(1, Duration.MICROSECONDS))
+        .build();
+
+    //this time expression produces an interval [TimeAnchor.END, TimeAnchor.END + 2 Ms]
+    final var intervalStartTimeExpression = new TimeExpressionRelativeFixed(TimeAnchor.END, false);
+    intervalStartTimeExpression.addOperation(TimeUtility.Operator.PLUS, Duration.of(2, Duration.MICROSECONDS));
+
+    final var coexistenceControllable = new CoexistenceGoal.Builder()
+        .thereExistsOne(zeroDurationUncontrollableActivity)
+        .forAllTimeIn(planningHorizon.getHor())
```
pranav-super: This line fails after rebasing as forAllTimeIn only takes a `Expression<Windows>`; the way to fix this is to change:
`.forAllTimeIn(planningHorizon.getHor())` to `.forAllTimeIn(new WindowsWrapperExpression(new Windows(planningHorizon.getHor())))`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -301,22 +303,31 @@ public boolean isApproximation() {
         }
       };
 
-      double ff_x_nminus1 = ff.valueAt(x_nminus1), ff_x_n = ff.valueAt(x_n);
+      var ff_x_nminus1 = ff.valueAt(x_nminus1);
+      if (ff_x_nminus1 >= - toleranceYLow &&
```
pranav-super, line 312: is this just a check to make sure you don't continue with root finding if the input is close enough to a root?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -301,22 +303,31 @@ public boolean isApproximation() {
...
+          ff_x_nminus1 <= toleranceYHigh
+          && (x_nminus1 >= dxLow && x_nminus1 <= dxHigh)) {
```
pranav-super: Why do we compare `x_n` to thresholds for dx? Shouldn't we compare something like `x_n-x_{n-1}`?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -301,22 +303,31 @@ public boolean isApproximation() {
...
-          throw new ZeroDerivative();
+        if (localDerivative == 0 || Double.isNaN(localDerivative)) {
+          throw new ZeroOrInfiniteDerivativeException();
         }
         x_nminus1 = x_n;
         ff_x_nminus1 = ff_x_n;
         x_n = x_n - (ff_x_nminus1 / localDerivative);
-        //outside of domain, diverging
-        if (x_n < dxLow || x_n > dxHigh) {
-          throw new DivergenceException("Looking for root out of prescribed domain :[" + dxLow + "," + dxHigh + "]");
```
pranav-super, line 315: we use x_n like it were a dx here when comparing to the differential thresholds, but use x_n-x_{n+1} like it were not a differential quantity when computing the local derivative above; why should this work?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
pranav-super: Would it be more useful to split the two?

pranav-super (changes requested):
Just some questions. Changes to tests I have in a separate branch right now.

examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/activities/ZeroDurationUncontrollableActivity.java
adrienmaillard: Yes. The run method will run "instantly" with regard to time (it does not wait/call/spawn).

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: Yes.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: `dx` is maybe misleading here, it is not the differential of x but the domain of x.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: same comment as below: `dx` is the domain of `x`, not its differential.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: exactly

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: It will not lead to a different resolution but I am open to that.

pranav-super: I think it should be provided just for clearer debugging, unless this case is rare?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
pranav-super: Oh...well then I retract the majority of my comments here!

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
pranav-super: It might not be a bad idea to rename it, unless I was the only one confused by it?

Twisol: I'm not in the target audience, but I think I would have made the same mistake. `xLow` and `xHigh` maybe?

adrienmaillard: I replaced them with your suggestion @Twisol.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: I have added ZeroDerivativeException and InfiniteDerivativeException.

adrienmaillard:
I have integrated the changes to tests that @pranav-super made (thanks!) on https://github.com/NASA-AMMOS/aerie/tree/origin/FIX--uncontrollable-activity-wo-end-constraint-and-updated-tests and made the requested changes.

pranav-super (approved):
Aside from the `dx` variable name versus something like `xLow` or `xHigh`, looks good!

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: I'll try building a test case with this case in mind. Maybe it'll lead to another PR :D

PR #235, opened 2022-06-27
[AERIE-1916] refactor `computeResults` out of sim engine

skovati:
* **Tickets addressed:** AERIE-1916
* **Review:** By commit  

## Description

The goal of this ticket is to refactor the `computeResults()` method out of `SimulationEngine` and into `SimulationDriver`, so that it can eventually be pulled inside the `while(true)` simulation loop for simulation results streaming.

This refactoring is mainly a *lift and shift*, meaning no significant changes are being made to the method, and instead private fields are being made public to prototype this change. Eventually (and this is the hope of opening a draft PR) we hope to expose the minimal information needed from these `SimulationEngine` fields, with public getters instead of public fields.

## Verification

Manually testing to ensure simulation as a whole hasn't changed.

## Documentation

None

## Future work

Investigate how to chunk simulation results
Move `computeResults()` into the main sim loop

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
 public final class SimulationDriver {
@@ -144,10 +145,7 @@ public final class SimulationDriver {
     final var realProfiles = new HashMap<String, List<Pair<Duration, RealDynamics>>>();
     final var discreteProfiles = new HashMap<String, Pair<ValueSchema, List<Pair<Duration, SerializedValue>>>>();
 
-    for (final var entry : engine.resources.entrySet()) {
-      final var id = entry.getKey();
-      final var state = entry.getValue();
-
+    engine.getResources().forEach((id, state) -> {
       final var name = id.id();
       final var resource = state.resource();
 
@@ -166,7 +164,7 @@ public final class SimulationDriver {
             throw new IllegalArgumentException(
                 "Resource `%s` has unknown type `%s`".formatted(name, resource.getType()));
       }
-    }
+    });
 
     // Give every task corresponding to a child activity an ID that doesn't conflict with any root activity.
     final var taskToPlannedDirective = new HashMap<>(taskInfo.taskToPlannedDirective());
@@ -177,9 +175,17 @@ public final class SimulationDriver {
             .map(ActivityInstanceId::id)
             .collect(Collectors.toSet());
 
+    // Grab all tasks that are activities, so we don't have to check later
+    final var activityTasks =
+        engine
+            .getTasks()
+            .entrySet()
+            .stream()
+            .filter(e -> taskInfo.isActivity(e.getKey()))
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+
     var counter = 1L;
-    for (final var task : engine.tasks.keySet()) {
-      if (!taskInfo.isActivity(task)) continue;
+    for (final var task : activityTasks.keySet()) {
       if (taskToPlannedDirective.containsKey(task.id())) continue;
 
       while (usedActivityInstanceIds.contains(counter)) counter++;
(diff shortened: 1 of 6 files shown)


skovati:
Rebased on upstream develop.

Looking for architecture/design feedback! Refactor doesn't feel particularly ergonomic, mainly that `taskInfo` seems to hold a lot of the same info as the sim engine `tasks` field, but now both are being accessed in `SimulationDriver`

Since this `computeResults` method is going to be called **much** more often once sim streaming is working, it's important for it to be performant. One thought I had was maybe moving the `taskInfo` population into `SimulationDriver`'s `simulate` method, and update that object incrementally when the timeline changes.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -90,8 +111,223 @@ SimulationResults simulate(
...
+    }
+  }
+
+  /** Compute a set of results from the current state of simulation. */
+  // TODO: Move result extraction out of the SimulationEngine.
+  //   The Engine should only need to stream events of interest to a downstream consumer.
+  //   The Engine cannot be cognizant of all downstream needs.
+  // TODO: Whatever mechanism replaces `computeResults` also ought to replace `isTaskComplete`.
+  // TODO: Produce results for all tasks, not just those that have completed.
+  //   Planners need to be aware of failed or unfinished tasks.
```
mattdailis, line 124: Some of these TODO comments are out of date, we should do a survey.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -90,8 +111,223 @@ SimulationResults simulate(
...
+      final SimulationEngine engine,
+      final Instant startTime,
+      final Duration elapsedTime,
+      final Topic<ActivityInstanceId> activityTopic,
+      final TemporalEventSource timeline,
+      final MissionModel<?> missionModel
+  ) {
+
+    // Collect per-task information from the event graph.
+    final var taskInfo = new TaskInfo();
```
mattdailis, line 135: Let's move `TaskInfo` out of the engine as well - I think it should live together with `computeResults`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
```diff
@@ -149,7 +150,7 @@ public SimulationResults getSimulationResultsUpTo(Duration endTime){
     }
 
     if(lastSimResults == null || endTime.longerThan(lastSimResultsEnd)) {
-      lastSimResults = SimulationEngine.computeResults(
+      lastSimResults = SimulationDriver.computeResults(
```
mattdailis, line 153: I'm debating whether or not it's a problem to have one driver depend on the other. I suspect we'll want to split out the results computation into its own class.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -90,8 +111,223 @@ SimulationResults simulate(
...
+        taskToPlannedDirective
+            .values()
+            .stream()
+            .map(ActivityInstanceId::id)
+            .collect(Collectors.toSet());
+
+    // Grab all tasks that are activities, so we don't have to check later
+    final var activityTasks =
+        engine
+            .getTasks()
```
mattdailis, line 181: One refactor that I tried on my end was to add a `getActivityTasks` method to `TaskInfo` (which returns a `Set<TaskId>`), and then have the engine provide a `getTaskState(TaskId)` method, but upon reflection, I don't know if that gains us anything.

mattdailis (commented):
I think the lift and shift went alright, I'm struggling to come up with anything I would consider "cleaner". At first I thought the `ExecutionState` type should be kept internal to the Sim Engine, but upon inspecting its fields, all of them are relevant to results computation 🤷

Twisol:
Sorry, I haven't had a chance to look at this yet -- I saw the request for review. (I need to move out of my apartment this Friday.) I've wanted results computation pulled out of the engine for *a long time*, so I'm happy to see progress on this. 😃

One abstract concern I have is that, since results computation is still coupled to the engine, pulling it out of the engine class might end up spreading out the problem, while also breaking encapsulation of the engine itself. I kept the results code local to the engine because, until I had a better architectural idea for what ought to happen, having a localized mess is better than having a metastasized mess.

I'd be very interested to hear whether there are any semi-solid ideas floating around for what the goal state of the codebase ought to be. @mattdailis and I previously tossed around some ideas; my most recent PRs were centered on pushing more information into the event graph, which is a much more principled way of getting information out incrementally over a simulation.

Some things don't seem to work well that way, and so might need other solutions. Specifically, resources and activity lifecycle information don't "branch" in the same way as the general event graph (and hence downstream consumers shouldn't need to account for such impossibilities).

Twisol:
someone explain to me why github's behavior for cmd+enter is "Close with comment" and not merely "comment"

Twisol:
OMG

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
skovati: Definitely agree, will work on that change!

skovati:
Closing as the team has determined `computeResults` needing to be refactored out of `SimulationEngine` is more of a symptom of a larger problem than the solution. The idea is to move forward with "timeline segment streaming" where event like updates are relayed to the client, which will be prototyped in AERIE-1921. This change will very likely invalidate the need for AERIE-1916 and this accompanying PR.

PR #236, opened 2022-06-28
[AERIE-1922] Add absolute time to activities and resource profiles

dyst5422:
* **Tickets addressed:** AERIE-1922
* **Review:** By commit  

## Description

Because Postgres doesn't allow cross table selects for generated columns, we have to use a view instead that can do the appropriate query through to plan to get the start time. The view adds the plan start time to the start offset of the item to get the start time and adds duration to that to get the end time. In the case of the resource profiles, we have to find the time the next profile takes over to determine the end time of the profile.

## Verification

Did some basic local inspection of DB contents with the new views and added some tests to the command expansion server to use the new absolute times.

## Documentation

None - all changes should be auto-documented in the graphql API and the command expansion types

### DIFF
--- a/merlin-server/sql/merlin/views/resource_profile.sql
+++ b/merlin-server/sql/merlin/views/resource_profile.sql
@@ -1,0 +1,54 @@
+create view resource_profile as
+(
+select profile_segment.dataset_id as dataset_id,
+       profile_segment.profile_id as profile_id,
+       profile_segment.start_offset as start_offset,
+       profile_segment.dynamics as dynamics,
+       plan.start_time + profile_segment.start_offset as start_time,
+       coalesce(
+             plan.start_time + (
+             select p.start_offset
+             from profile_segment p
+             where p.start_offset > profile_segment.start_offset
+               and p.profile_id = profile_segment.profile_id
+               and p.dataset_id = profile_segment.dataset_id
+             order by p.start_offset
+             limit 1
+           ),
+             plan.start_time + plan.duration
+         ) as end_time
+
+from profile_segment
+       join dataset on profile_segment.dataset_id = dataset.id
+       left join plan_dataset pd on dataset.id = pd.dataset_id
+       left join simulation_dataset sd on dataset.id = sd.dataset_id
+       left join simulation s on sd.simulation_id = s.id
+       join plan on plan.id = s.plan_id or plan.id = pd.plan_id
+);
+
+comment on view resource_profile is e''
+  'A piece of a profile associated with a dataset, starting at a particular offset from the dataset basis. '
+  'The profile is governed at any time T by the latest profile whose start_offset is no later than T.'
+  'This view adds in absolute start and end times to the profile segment.';
+
+comment on column resource_profile.dataset_id is e''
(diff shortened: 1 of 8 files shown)


merlin-server/sql/merlin/views/resource_profile.sql
```diff
@@ -0,0 +1,51 @@
...
+             where p.start_offset > profile_segment.start_offset
+             limit 1
+           ),
+             plan.start_time + plan.duration
+         ) as end_time
+
+from profile_segment
+       join dataset on profile_segment.dataset_id = dataset.id
+       left join plan_dataset pd on dataset.id = pd.dataset_id
+       join simulation_dataset sd on dataset.id = sd.dataset_id
```
mattdailis: I'm wondering how we would do this for external profiles as well - though that may be out of scope for this PR

mattdailis (approved):
Looks solid to me

merlin-server/sql/merlin/views/resource_profile.sql
dyst5422: I think it's covered by the OR clause on line 23.

I believe the story for external profiles is that they are associated with a plan rather than a simulation, so the line `join plan on s.plan_id = plan.id OR plan.id = pd.plan_id` covers it with the `plan.id = pd.plan_id`.

I do have a moment of pause as to whether we need to do a `left join` on the `simulation_dataset` and `simulation` though to ensure we get those profiles WITHOUT a `simulation_dataset` and `simulation`.

goetzrrGit (approved):
A manual run has the start time being fetched correctly

merlin-server/sql/merlin/views/resource_profile.sql
mattdailis: Agreed with your logic, as well as your pause :+1:

merlin-server/sql/merlin/views/resource_profile.sql
```diff
@@ -0,0 +1,51 @@
...
+           ),
+             plan.start_time + plan.duration
+         ) as end_time
+
+from profile_segment
+       join dataset on profile_segment.dataset_id = dataset.id
+       left join plan_dataset pd on dataset.id = pd.dataset_id
+       left join simulation_dataset sd on dataset.id = sd.dataset_id
+       left join simulation s on sd.simulation_id = s.id
+       join plan on plan.id = s.plan_id  or plan.id = pd.plan_id
```
mattdailis: rogue space here

mattdailis (commented):
There are two commits with the same name - I think you maybe meant to amend the last commit?

merlin-server/sql/merlin/views/resource_profile.sql
mattdailis: Assuming it went okay in your testing, I'll resolve this conversation. Treating plan_dataset and simulation_dataset the same way using `left join` sounds like a good idea to me.

dyst5422: As I was handling this one, I also noticed that we are missing a clause to filter for the profile id and dataset id on line 12

PR #237, opened 2022-06-29
[AERIE-1913]  Add simulation dataset id to `simulate` query response

skovati:
* **Tickets addressed:** AERIE-1913
* **Review:** By commit  

## Description

This feature was added so that clients of the API could hit the `simulate` Hasura action, and instantly receive back the simulationDatasetId to subscribe with. This removes the need to poll the `simulate` action until it returns `status: complete`.

This was implemented by adding a new `long simulationDatasetId` field to the `ResultsProtocol` interface, and having both `PostgresResultsCellRepository` and `InMemoryResultsCellRepository` populate it. One possible issue with this implementation is that `InMemoryResultsCellRepository` doesn't really have a concept of `datasetId` so it is just getting hardcoded to 0. The justification for this is that `InMemory` is somewhat of a second class citizen only used in testing, so priority was given to the ergonomic use of `PostgresResults` instead.

## Verification

Manually tested changes in UI and GraphQL API to make sure nothing changed except returning simulationDatasetId.

## Documentation

Need to update `simulate` response documentation

## Future work

Further steps towards simulation result streaming
Make use of this change in the UI

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -151,6 +151,7 @@ type ResourceType {
 
 type MerlinSimulationResponse {
   status: MerlinSimulationStatus!
+  simulationDatasetId: Int!
   reason: MerlinSimulationFailureReason
 }
 
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
@@ -290,10 +290,10 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
 
     return Optional.of(
         switch (record.state().status()) {
-          case PENDING -> new ResultsProtocol.State.Pending();
-          case INCOMPLETE -> new ResultsProtocol.State.Incomplete();
-          case FAILED -> new ResultsProtocol.State.Failed(record.state().reason());
-          case SUCCESS -> new ResultsProtocol.State.Success(getSimulationResults(connection, record, planId));
+          case PENDING -> new ResultsProtocol.State.Pending(record.simulationDatasetId());
+          case INCOMPLETE -> new ResultsProtocol.State.Incomplete(record.simulationDatasetId());
+          case FAILED -> new ResultsProtocol.State.Failed(record.simulationDatasetId(), record.state().reason());
+          case SUCCESS -> new ResultsProtocol.State.Success(record.simulationDatasetId(), getSimulationResults(connection, record, planId));
         });
   }
 
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
@@ -7,16 +7,16 @@ public final class ResultsProtocol {
 
   public sealed interface State {
     /** Simulation in enqueued. */
-    record Pending() implements State {}
+    record Pending(long simulationDatasetId) implements State {}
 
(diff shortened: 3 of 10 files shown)


deployment/hasura/metadata/actions.graphql
```diff
@@ -151,6 +151,7 @@ type ResourceType {
 
 type MerlinSimulationResponse {
   status: MerlinSimulationStatus!
+  simulationDatasetId: Int
```
camargo: Can this be `Int!`?

skovati: Good call! I had that in there before I was returning a datasetId in every case

skovati: Change has been made!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -290,10 +290,10 @@ private static Optional<State> getSimulationState(
 
     return Optional.of(
         switch (record.state().status()) {
-          case PENDING -> new ResultsProtocol.State.Pending();
-          case INCOMPLETE -> new ResultsProtocol.State.Incomplete();
-          case FAILED -> new ResultsProtocol.State.Failed(record.state().reason());
-          case SUCCESS -> new ResultsProtocol.State.Success(getSimulationResults(connection, record, planId));
+          case PENDING -> new ResultsProtocol.State.Pending(datasetId);
```
mattdailis: 🔻 There is an unfortunate subtle difference between `simulationDatasetId` and `datasetId`. In this case, I think we want to update `SimulationDatasetRecord` to have a `simulationDatasetId` field, and pull that field out of the `record` variable here.

mattdailis (changes requested):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
skovati: Would this be separate from the `datasetId` field that already exists in the `SimulationDatasetRecord`?

goetzrrGit (approved):
Besides Matt's suggestions, this seems pretty straightforward.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
mattdailis: Correct, and this is why I say it's unfortunate 😓 

`simulation_dataset` is a join table between the `simulation` table and the `dataset` table. `datasetId` refers to the `id` field on the `dataset` table, while `simulationDatasetId` refers to the `id` field on the `simulation_dataset` table.

Being a join table, the `simulation_dataset` refers to the `dataset` table using a column of its own called `dataset_id`. So, the `SimulationDatasetRecord.datasetId` field comes from the `simulation_dataset.dataset_id` column, which is a foreign key that refers to the `id` column of the `dataset` table.

Let me know if I made things even more confusing.

skovati: Ahh gotcha, so I'll add `simulation_dataset.id` to the queries in `{Get,Lookup,Create}SimulationDatasetAction` and add that into the `SimulationDatasetRecord` as `simulationDatasetId`

skovati: Change has been made!

mattdailis (approved):

PR #238, opened 2022-07-07
[AERIE-1906] Resolve `SerializedValue` LGTM errors

pcrosemurgy:
* **Tickets addressed:** AERIE-1906
* **Review:** By commit  

## Description
From #220.

Resolves LGTM `SerializedValue` errors reported [here](https://lgtm.com/projects/g/NASA-AMMOS/aerie/alerts/?mode=tree&severity=error&ruleFocus=6770060).

## Verification
All tests passing.

## Documentation
None.

## Future work
Resolve LGTM warnings (see https://lgtm.com/projects/g/NASA-AMMOS/aerie/alerts/?mode=tree).


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
```diff
@@ -239,7 +248,7 @@ protected Optional<T> onDefault() {
    *
    * @return True if this object represents a null value, and false otherwise.
    */
-  public boolean isNull() {
+  default boolean isNull() {
```
Twisol, line 251: Why are `OptionalVisitor`'s methods changed? It's still an `abstract class`, so we still want its methods explicitly `public` (and superclass methods are already `default` by, er, default).

merlin-sdk/src/test/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValueTest.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public final class SerializedValueTest {
+  @Test
+  public void testHashCode() {
```
Twisol: We shouldn't need to test that Java's own `hashCode` methods abide by its own contract 🤔

merlin-sdk/src/test/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValueTest.java
```diff
@@ -0,0 +1,26 @@
...
+    final var someValue2 = SerializedValue.NULL;
+    final var list1 = List.of(someValue1, SerializedValue.of(42));
+    final var list2 = List.of(someValue2, SerializedValue.of(43));
+    final var map1 = Map.of(someValue1, SerializedValue.of(42));
+    final var map2 = Map.of(someValue2, SerializedValue.of(43));
+    final var map3 = Map.of(someValue2, SerializedValue.of(42));
+
+    assertThat(someValue1.hashCode()).isEqualTo(someValue2.hashCode());
+    assertThat(list1.hashCode()).isNotEqualTo(list2.hashCode());
+    assertThat(map1.hashCode()).isNotEqualTo(map2.hashCode());
```
Twisol: It is legal (and does not harm correctness) for these hash codes to be the same. The relevant contract is "if `list1.equals(list2)` then `list1.hashCode() == list2.hashCode()`". If the two objects are not `equals()`, then their hash codes can be anything.

It would make sense to test a more specific contract if we had control over the implementation, but Java is responsible for the implementation, and Java only guarantees up to the above contract.

Twisol:
Sorry... GitHub notifications never actually tell me a PR is just a draft. 🤦‍♂️

pcrosemurgy:
>Sorry... GitHub notifications never actually tell me a PR is just a draft. 🤦‍♂️

No worries! All solid feedback :) Thanks

merlin-sdk/src/test/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValueTest.java
pcrosemurgy: Just removed this test :)

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
pcrosemurgy:
```diff
-  public static abstract class OptionalVisitor<T> extends DefaultVisitor<Optional<T>> {
+  abstract class OptionalVisitor<T> extends DefaultVisitor<Optional<T>> {
     @Override
     protected Optional<T> onDefault() {
       return Optional.empty();
@@ -239,7 +248,7 @@ public abstract class SerializedValue {
    *
    * @return True if this object represents a null value, and false otherwise.
    */
-  public boolean isNull() {
+  default boolean isNull() {
```

I don't believe any `OptionalVisitor` methods were changed – directly at least.
You may have misread the diff (or I may have 😂) and thought that `default boolean isNull()` was added to `OptionalVisitor`?

Twisol: Ugh, yes, I thought `OptionalVisitor` was a longer class. (I missed the closing brace.)

Twisol: Ugh, yes, I thought `OptionalVisitor` was a longer class. (I missed the closing brace.)

pcrosemurgy:
Opened this PR prematurely. Am investigating the `e2e-tests` failure.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -628,24 +630,66 @@ private Collection<ResourceType> getResourceTypes(final MissionModelId missionMo
...
+   * @param d the duration to serialize
+   * @return a serialization of the object suitable for use as a graphql value
+   */
+  public String serializeForGql(final Duration d) {
+    //TODO: can probably leverage some serializers from aerie
+    //NB: merlin uses durations in microseconds! (inconsistent with start_offset as a HH:MM:SS.sss string)
+    return Long.toString(d.in(Duration.MICROSECOND));
+  }
+
+  public String serializeForGql(final SerializedValue value) {
```
pcrosemurgy, line 655: This method is only as strong as the methods it relies on (the 2 other `serializeForGql()` overloads above).
There's a preexisting to-do to correctly implement string escaping. This PR at least moves us away from using `serializedValue.toString()` to serialize a `SerializedValue` as a GQL-formatted value.

skovati (approved):
LGTM! (heh) Smart move refactoring `SerializedValue` into an interface.

Should we create a separate ticket to capture the string escape work or were you planning on implementing that in this PR?

adrienmaillard (approved):
Looks good, thank you.

pcrosemurgy:
> Should we create a separate ticket to capture the string escape work or were you planning on implementing that in this PR?

Good point! I'll check in with Adrien on this, the preexisting `TODO` notes about string escaping should probably be removed in place of a ticket.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
pcrosemurgy: @patkenneally @adrienmaillard I've created https://jira.jpl.nasa.gov/browse/AERIE-1949 to correctly implement string `"`/`'` escaping as well as GQL-injection sanitization.

dyst5422:
Why are we doing sanitization for GQL instead of using variables? That's one of the great value propositions of the GraphQL standard - built in sanitization.

pcrosemurgy:
Sure! I'll change the wording of the ticket.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
pcrosemurgy: Edit: " " " as well as using GQL **variables** to avoid the need for sanitization in the first place.

dyst5422:
I don't think my comment is scoped to this ticket as the string formatting predates this ticket. Specifically looking at lines like https://github.com/NASA-AMMOS/aerie/blob/88d926a197c58db74d61676a430c73624ce96b2c/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java#L137-L145 where we were already doing string templating instead of using variables

pcrosemurgy:
Yup! But your point 100% still stands, we should use variables. https://jira.jpl.nasa.gov/browse/AERIE-1949 should address this.

PR #241, opened 2022-07-11
[AERIE-1824] - Rework TimeType API

goetzrrGit:
* **Tickets addressed:** AERIE-1824
* **Review:** By commit  

## Description
Currently, the eDSL for command authoring has the v1 structure below. The original way isn't too bad but there was some feedback from Basak to make it easier to write.

**Original way:**

```
<command>.absoluteTiming(Temporal.instance)
<command>.relativeTiming(Temporal.duration)
<command>.epochTiming(Temporal.duration)
<command>
```

**New Way:**

With this PR you have multiple ways to specify Timing Types

```
// Absolute
A`2020-060T03:45:19`.ADD_WATER
A(Temporal.Instant.from("2025-12-24T12:01:59Z")).ADD_WATER
A("2020-060T03:45:19").ADD_WATER

// Relative
R`00:15:00`.EAT_BANANA
R(Temporal.Duration.from({ minutes: 15 }).EAT_BANANA
R(00:15:00)EAT_BANANA

// Epoch
E`00:15:30`.PREPARE_LOAF(1,true)
E(Temporal.Duration.from({ minutes: 15, seconds: 30 }).PREPARE_LOAF(1,true)
E(00:15:30).PREPARE_LOAF(1,true)

C.BAKE_BREAD

```

## Verification
All JEST Test Pass on my end. 

To run the test locally:

* Nuke all docker containers
* In Docker Compose, Change `cam` to `none`
* Navigate to `Command-Expansion-Server`
* Enter `nvm use 16` to use node 16
* Run the test `npm run test`
* All the tests should pass

## Documentation
I will update the documentation once this PR is merged.

## Future work
Add more helper methods to make authoring easier.

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -183,11 +183,11 @@ export class Command<
   public static fromSeqJson<A extends ArgType[]>(json: CommandSeqJson<A>): Command<A> {
     const timeValue =
       json.time.type === TimingTypes.ABSOLUTE
-        ? { absoluteTime: Command.doyToInstant(json.time.tag) }
+        ? { absoluteTime: doyToInstant(json.time.tag) }
         : json.time.type === TimingTypes.COMMAND_RELATIVE
-        ? { relativeTime: Command.hmsToDuration(json.time.tag) }
+        ? { relativeTime: hmsToDuration(json.time.tag) }
         : json.time.type === TimingTypes.EPOCH_RELATIVE
-        ? { epochTime: Command.hmsToDuration(json.time.tag) }
+        ? { epochTime: hmsToDuration(json.time.tag) }
         : {};
 
     return Command.new<A>({
@@ -234,48 +234,6 @@ export class Command<
     return `${YYYY}-${DOY}T${HH}:${MM}:${SS}.${sss}` as DOY_STRING;
   }
 
-  public static doyToInstant(doy: DOY_STRING): Temporal.Instant {
-    const match = doy.match(DOY_REGEX);
-    if (match === null) {
-      throw new Error(`Invalid DOY string: ${doy}`);
-    }
-    const [, year, doyStr, hour, minute, second, millisecond] = match as [
-      unknown,
-      string,
-      string,
-      string,
-      string,
-      string,
-      string | undefined,
-    ];
-
-    //use to convert doy to month and day
-    const doyDate = new Date(parseInt(year, 10), 0, parseInt(doyStr, 10));
-    // convert to UTC Date
-    const utcDoyDate = new Date(
-      Date.UTC(
-        doyDate.getUTCFullYear(),
-        doyDate.getUTCMonth(),
-        doyDate.getUTCDate(),
-        doyDate.getUTCHours(),
-        doyDate.getUTCMinutes(),
-        doyDate.getUTCSeconds(),
-        doyDate.getUTCMilliseconds(),
-      ),
-    );
-
-    return Temporal.ZonedDateTime.from({
-      year: parseInt(year, 10),
-      month: utcDoyDate.getUTCMonth() + 1,
-      day: utcDoyDate.getUTCDate(),
-      hour: parseInt(hour, 10),
-      minute: parseInt(minute, 10),
-      second: parseInt(second, 10),
-      millisecond: parseInt(millisecond ?? '0', 10),
-      timeZone: 'UTC',
-    }).toInstant();
-  }
-
   /** HH:MM:SS.sss */
   private static durationToHms(time: Temporal.Duration): HMS_STRING {
     const HH = this.formatNumber(time.hours, 2);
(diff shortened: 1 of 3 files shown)


command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -31,13 +31,13 @@ export type CommandOptions<
   M extends Record<string, any> = Record<string, any>,
 > = { stem: string; arguments: A; metadata?: M } & (
   | {
-      absoluteTime: Temporal.Instant;
+      absoluteTime: String | Temporal.Instant;
```
dyst5422: String is not something you should generally use see https://dev.to/josiasaurel/the-difference-between-string-and-string-in-typescript-4mf2

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -146,11 +153,32 @@ export class Command<
       stem: this.stem,
       time:
         this.absoluteTime !== null
-          ? { type: TimingTypes.ABSOLUTE, tag: Command.instantToDoy(this.absoluteTime) }
+          ? {
+              type: TimingTypes.ABSOLUTE,
+              tag: Command.instantToDoy(
+                (typeof this.absoluteTime === 'string'
+                  ? Command.doyToInstant(this.absoluteTime as DOY_STRING)
```
dyst5422: We should do this conversion in the time functions themselves (A,R,E)

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -321,6 +349,57 @@ export class Sequence {
...
+function R(duration: String | Temporal.Duration): GeneratedCommands {
+  return commandsWithTimeValue(duration, TimingTypes.COMMAND_RELATIVE);
+}
+
+// @ts-ignore : Used in generated code
+function E(duration: String | Temporal.Duration): GeneratedCommands {
+  return commandsWithTimeValue(duration, TimingTypes.EPOCH_RELATIVE);
+}
+
+function commandsWithTimeValue(
```
dyst5422: We can better type this as
```typescript
function commandsWithTimeValue<T extends TimingTypes>(
  timeValue: T extends TimingTypes.Absolute ? string | Temporal.Instant : string | Temporal.Duration,
  timeType: T
): GeneratedCommands
```

This allows us to enforce that absolute uses strings or instants and relative uses strings or durations

dyst5422 (changes requested):
I think we can make it a bit more terse using tagged template literals https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates
so we can write it like
```typescript
A`2020-060T03:45:19`.ADD_WATER
R`00:15:00`.EAT_BANANA
E`00:15:30`.PREPARE_LOAF(1,true)
```

We would then support supplying a value that isn't a string using overloads so we can properly type check the variadic inputs.

EX:

```typescript

function A(...args: [TemplateStringsArray, ...string[]]): GeneratedCommands;
function A(timeVal: Temporal.Instant): GeneratedCommands;
function A(timeDOYString: string): GeneratedCommands;
function A(...args: [TemplateStringsArray, ...string[]] | [Temporal.Instant] | [string]): GeneratedCommands {
  let timeVal = timeValue;
  // Handle tagged template format
  if (Array.isArray(args[0])) {
    timeVal = Command.doyToInstant(String.raw(args));
  }
  // Handle string inputs
  else if (typeof args[0] === 'string') {
    timeVal = Command.doyToInstant(args[0]);
  }

  ...do rest of stuff now that we've normalized the time value to a Temporal.Instant


}
```

This allows us to use it in all of the following ways with strong checking

```typescript
A`2020-060T03:45:19`.ADD_WATER

let doyString = '2020-060T03:45:19';
A(doyString).ADD_WATER

let temporalInstant = doyToInstant('2020-060T03:45:19');
A(temporalInstant).ADD_WATER
```

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -89,8 +89,31 @@ declare global {
   type F<BitLength extends 32 | 64> = number;
   type F32 = F<32>;
   type F64 = F<64>;
+
+  // @ts-ignore : 'Commands' found in generated code
```
dyst5422, line 93: Love the comments here

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -263,7 +286,7 @@ export class Command<
     return `${HH}:${MM}:${SS}.${sss}` as HMS_STRING;
   }
 
-  private static hmsToDuration(hms: HMS_STRING): Temporal.Duration {
+  public static hmsToDuration(hms: HMS_STRING): Temporal.Duration {
```
dyst5422: I think we may want to separate these conversions into their own non-exported functions so that they don't become part of the public API?

dyst5422 (changes requested):

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: @goetzrrGit ☝️

goetzrrGit: Oh I didn't see this. I will look at this now.

goetzrrGit: I pulled the methods out in the last commit.

dyst5422 (approved):
LGTM

camargo (approved):

PR #242, opened 2022-07-12
[AERIE-1849] Test composite goals

adrienmaillard:
* **Tickets addressed:** AERIE-1849
* **Review:** By commit  

## Description

OR/AND goals are available in the eDSL since an early version but they were dysfunctional/were not tested. 

While working on testing, I stumbled on the criterion required for satisfying these goals. 

How normal goal satisfaction works: the goal generates conflicts that the solver tries to solve. Whenever a goal cannot be completely satisfied because there are conflicts that cannot be addressed (ex: we want 6 activities in some time period but only 5 fit), the solver would remove all activities associated to this goal from the plan and go to the next goal. Either the goal can be satisfied or not. 

But that’s not always the desired behavior. Sometimes, we want a best-effort approach. So we introduced “partial satisfiability”. 
Partial satisfiability is a boolean property of any goal. If it is set to true, and whenever the goal cannot be totally satisfied, activities associated to the goal are NOT removed from the plan. The goal will still appear unsatisfied. 

By default, partial satisfiability is set to true. We do not have yet the possibility of updating this value in the eDSL but I hope we will get it soon.

Because of the presence of partial satisfiability, the logic of composite goal satisfaction had to be updated/made clearer. 

Commit 2, refactor disjunctive goal: it was a mess of to-be-implemented features and it did not inherit correctly from `Goal`
Commit 3: when building a conjunctive goal, the horizon was not set (`forAllTimeIn`), producing an error. This commit fixes this issue.
Commit 4: The solver has access to results of subgoals but these goals are "anonymous" for the rest of the scheduler, they do not appear by themselves to the user. When building scheduling results for a composite and/or goals, this lead to the results of these anonymous goals to be added to the returned results (with a `null` id) which is not the desired behavior. 
Commit 5: The criterion for satisfaction of composite or/and goal has been refactored to be similar with the semantics of `partialSatisfiability` for non-composite goals. 
Commit 6: An accessor is added to `PlanningHorizon`.
Commit 7: tests are added for composite or/and goals. 
Commit 8: An id bug is fixed in `MockMerlinService`
Commit 9: the `inPeriod` operator of the `CardinalityGoal` is purged from the eDSL  

## Verification

New tests have been added. 

## Documentation

- [x] I have to update the documentation with a paragraph on the logic of goal satisfaction

## Future work

Implement an operator in the scheduling eDSL allowing the user to specify whether he wants partial satisfaction for a goal or not.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
@@ -421,10 +421,24 @@ public record SynchronousSchedulerAgent(
     Map<GoalId, ScheduleResults.GoalResult> goalResults = new HashMap<>();
       for (var goalEval : plan.getEvaluation().getGoalEvaluations().entrySet()) {
         var goalId = goalsToIds.get(goalEval.getKey());
-        var goalResult = new ScheduleResults.GoalResult(goalEval.getValue().getInsertedActivities().stream().map(instancesToIds::get).collect(Collectors.toList()),
-                                                        goalEval.getValue().getAssociatedActivities().stream().map(instancesToIds::get).collect(Collectors.toList()),
-                                                        goalEval.getValue().getScore() >=0);
-        goalResults.put(goalId, goalResult);
+        //goal could be anonymous, a subgoal of a composite goal for example, and thus have no meaning for results sent back
+        if(goalId != null) {
+          final var goalResult = new ScheduleResults.GoalResult(
+              goalEval
+                  .getValue()
+                  .getInsertedActivities()
+                  .stream()
+                  .map(instancesToIds::get)
+                  .collect(Collectors.toList()),
+              goalEval
+                  .getValue()
+                  .getAssociatedActivities()
+                  .stream()
+                  .map(instancesToIds::get)
+                  .collect(Collectors.toList()),
+              goalEval.getValue().getScore() >= 0);
+          goalResults.put(goalId, goalResult);
+        }
       }
     return new ScheduleResults(goalResults);
   }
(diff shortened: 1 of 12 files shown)


scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -424,4 +424,88 @@ export default function() {
...
+        )));
+    if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success r) {
+      assertEquals(
+          expectedGoalDefinition,
+          r.goalSpecifier()
+      );
+    } else if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Error r) {
+      fail(r.toString());
+    }  }
+
```
mattdailis: Rogue closing brace.
```suggestion
    }
  }
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -421,10 +421,24 @@ private ScheduleResults collectResults(final Plan plan, final Map<ActivityInstan
     Map<GoalId, ScheduleResults.GoalResult> goalResults = new HashMap<>();
       for (var goalEval : plan.getEvaluation().getGoalEvaluations().entrySet()) {
         var goalId = goalsToIds.get(goalEval.getKey());
-        var goalResult = new ScheduleResults.GoalResult(goalEval.getValue().getInsertedActivities().stream().map(instancesToIds::get).collect(Collectors.toList()),
-                                                        goalEval.getValue().getAssociatedActivities().stream().map(instancesToIds::get).collect(Collectors.toList()),
-                                                        goalEval.getValue().getScore() >=0);
-        goalResults.put(goalId, goalResult);
+        //goal could be anonymous, a subgoal of a composite goal for example, and thus have no meaning for results sent back
+        if(goalId != null) {
+          var goalResult = new ScheduleResults.GoalResult(
```
mattdailis: I think `goalResult` can be final.

mattdailis (commented):
Generally looks good to me. As you say - we should write some documentation explaining composite goals - currently we have no documentation about them at all.

JoelCourtney (approved):

patkenneally (approved):

PR #245, opened 2022-07-14
[AERIE-1934] Add metadata to activity directives

dyst5422:
* **Tickets addressed:** AERIE-1934
* **Review:** By commit  

## Description

The timestamps were pretty much copy-paste of the same from command expansion.
The name was an interesting one because we have to dynamically generate sequences for incrementation. Because of the scoping of the sequences, their names are dynamically generated as well. I have a vague concern about maintaining all the sequences over the lifetime of a mission, but I'm not sure that's founded.

## Verification

I did a local deploy and tested out with the creation of a number of activities of different types across multiple plans.

## Documentation

Documentation added as comments in SQL.


merlin-server/sql/merlin/tables/activity.sql
```diff
@@ -83,3 +95,41 @@ create trigger increment_revision_on_delete_activity_trigger
...
+
+comment on function generate_activity_name() is e''
+  'Generates a name for an activity as the activity type + number from sequence of the type scoped to its plan.';
+
+create trigger generate_name_trigger
+before insert on activity
+for each row execute function generate_activity_name();
+
+comment on trigger generate_name_trigger on activity is e''
+  'Generates a name for an activity as the activity type + number from sequence of the type scoped to its plan.';
```
goetzrrGit: More of an understanding question:

This is kinda confusing me. What is a sequence with respect to an activity?  Is it something with the Schedular?

goetzrrGit (approved):
No real issues just a question.

merlin-server/sql/merlin/tables/activity.sql
dyst5422: Sequence as in a Postgres Sequence ie a numerically incrementing series of integers.

dyst5422: In this case, it means that for a given plan, each activity directive with the same activity type will have the name `<activityType> 1`, `<activityType> 2`, `<activityType> 3`, etc

goetzrrGit: Ahhh, thank you for the clarification.

merlin-server/sql/merlin/tables/activity.sql
```diff
@@ -83,3 +95,41 @@ create trigger increment_revision_on_delete_activity_trigger
...
+  return new;
+end$$;
+
+comment on function activity_set_updated_at() is e''
+  'Sets the last_modified_at field of an activity to the current time.';
+
+create trigger set_timestamp
+  before update on activity
+  for each row
+execute procedure activity_set_updated_at();
```
mattdailis: Just for my own understanding - what's the difference between `execute function` and `execute procedure`, and why do we use one for `generate_activity_name`, and the other for `activity_set_updated_at`?

dyst5422: From https://www.postgresql.org/docs/current/sql-createtrigger.html

`In the syntax of CREATE TRIGGER, the keywords FUNCTION and PROCEDURE are equivalent, but the referenced function must in any case be a function, not a procedure. The use of the keyword PROCEDURE here is historical and deprecated.`

dyst5422: I went ahead and added a commit normalizing our codebase to using `execute function`

mattdailis: Sounds good, thanks!

mattdailis (approved):

PR #249, opened 2022-07-18
[AERIE-1948] Starts and Ends windows (and spans!) operations

JoelCourtney:
* **Tickets addressed:** AERIE-1948
* **Review:** By commit  

## Description
These operations transform windows by replacing them with either their start point or end point. It doesn't care about the inclusivity of the original bounds.

Commits:
1. Implements the operations.
2. Adds the tests.

## Verification
I added tests for eDSL execution, JSON parsing, and evaluation behavior.

## Documentation
eDSL methods have doc comments. Wiki will be updated after this PR is approved, but before it is merged.

# Note
This PR is only the last two commits! It is rebased on top of #260, which will need to be merged first.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
@@ -122,6 +122,16 @@ public class Spans implements IntervalContainer<Spans>, Iterable<Interval> {
   }
 
   @Override
+  public Spans starts() {
+    return this.map($ -> Interval.at($.start));
+  }
+
+  @Override
+  public Spans ends() {
+    return this.map($ -> Interval.at($.end));
+  }
+
+  @Override
   public boolean equals(final Object obj) {
     if (!(obj instanceof final Spans spans)) return false;
 
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -288,6 +288,24 @@ public final class ConstraintParsers {
             $ -> tuple(Unit.UNIT, $.expression));
   }
 
+  static <I extends IntervalContainer<I>> JsonParser<Starts<I>> startsF(final JsonParser<Expression<I>> intervalExpressionP) {
+    return productP
+        .field("kind", literalP("IntervalsExpressionStarts"))
+        .field("expression", intervalExpressionP)
+        .map(
+            untuple((kind, expr) -> new Starts<I>(expr)),
+            $ -> tuple(Unit.UNIT, $.expression));
+  }
+
+  static <I extends IntervalContainer<I>> JsonParser<Ends<I>> endsF(final JsonParser<Expression<I>> intervalsExpressionP) {
+    return productP
+        .field("kind", literalP("IntervalsExpressionEnds"))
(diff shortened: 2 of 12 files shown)


adrienmaillard (approved):
LGTM

JoelCourtney:
This PR now includes Spans! But to do that I've rebased on top of the `feat/AERIE-1953--split-operation` branch, which includes the `IntervalContainer` interface. So, *only the last two commits* are actually part of this PR. Once the split PR is merged, I'll rebase on top of develop.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -64,6 +69,37 @@ public Spans filter(final Predicate<Window> filter) {
...
   }
 
+  @Override
+  public Spans split(final int numberOfSubSpans) {
+    if (numberOfSubSpans == 1) {
+      return new Spans(this);
+    }
+    return this.flatMap(x -> {
+      // Width of each sub-window, rounded down to the microsecond
+      final var width = Duration.divide(Duration.subtract(x.end, x.start), numberOfSubSpans);
```
goetzrrGit: I am a little confused by this. 

`x` - window
`x.end -x.start` - span/width of the window?

Why are you dividing by `numberOfSubSpans` are all the sub-window the same length?

goetzrrGit (approved):
LGTM. I just had one question for my understanding.

I ran through the unit test but couldn't get `ConstraintsDSLCompilationServiceTests.java` working not a blocker though.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: The variable name is a little misleading I guess, but it explains in the comment above. `width` here is the width of the new sub-windows, not the width of the original window. And no, if the number of microseconds in the original window is not a multiple of `N`, then the last sub-window can be up to `N-1` microseconds longer than the rest. So they're not *exactly* the same width, but they're close.

JoelCourtney: And to be clear, this piece of code is for the `split` operation, which is a different PR. This PR is for `starts` and `ends`, only the last two commits. I should have put that in the main description instead of just a comment.

patkenneally (approved):

JoelCourtney:
I redid portions of this PR to accommodate nullable windows. Given the interpretation that gaps represent "unknown", when a true segment borders null, it is more correct to replace that transition point with a gap, instead of `true`. Here is the exact wording for the doc comment on `starts` in the eDSL:

```
    /**
     * Replaces each window with its start point.
     *
     * Since gaps represent "unknown", true segments that come after a gap don't have a known start point.
     * So instead their first known point is unset and the rest is set to false.
     *
     * True segments that explicitly come directly after false and include their start point have all except their
     * start point set to false. If they don't include the start point, then the whole interval is set to false and the
     * start point is set true.
     */
```

This has the potential to be a bit of a "gotcha", but I think this is ok because we will already be requiring the author to use `assignGaps` religiously.

Twisol:
Nice! I think the presence of gaps is already a bit of a gotcha for these operations, since the actual start time is unknown. You could argue, perhaps, that we should throw an exception when an interval borders `null`, since the user might be assuming that such things aren't present. But I think the behavior you've specced out is consistent and reasonable. :+1:

PR #253, opened 2022-07-19
[AERIE-1943]  mission specific activity directive metadata

dyst5422:
* **Tickets addressed:** AERIE-1943
* **Review:** By commit  

## Description

Add mission specific metadata management.

This adds a table and domain type to manage the contents of the metadata. A column is added to the activity table to hold the activity metadata. We specifically DO NOT prevent extra keys, but we DO validate keys for which a schema exists in the metadata schema table. The schema is there specifically to allow our front-end tools to appropriately display the values and allow appropriate editing.

There are 6 different types defined for metadata:
- string: displays as a single line text field
- long_string: displays as a textarea
- boolean: displays as a checkbox
- number: displays as a numeric input
- enum: displays as a single select dropdown
- enum_multiselect: displays as a multi-select dropdown

If the schema is defined with a type not in that list, no validation occurs and the UI should just display a raw json editor. The intent is for the UI to display all fields for which there is a schema regardless of whether that value is stored in the metadata for the activity. Edits to those fields will write the fields back to the activity table. Fields that are NOT listed in the schema table should NOT be displayed.

The schema for each looks like:
-string: `{ type: 'string' }`
-long_string: `{ type: 'long_string' }`
-boolean: `{ type: 'boolean' }`
-number: `{ type: 'number' }`
-enum: `{ type: 'enum', enumerates: [<a series of values, could be numeric or string>] }`
-enum_multiselect: `{ type: 'enum_multiselect', enumerates: [<a series of values, could be numeric or string>] }`

The value format for each looks like:
-string: <string>
-long_string: <string>
-boolean: <boolean>
-number: <number>
-enum: <string|number>
-enum_multiselect: [\<array of strings or numbers\>]

Example metadata entry for an activity:
```
{
  "Science Intent": "Perform a reading of Saturn's magnetic field",
  "Requires Waiver": true,
  "ACTID": 24543
}
```
and the associated entries in the schema table

```
key               |    schema
-----------------------------------------------
"Science Intent"  | '{ "type": "long_string" }'
"Requires Waiver" | '{ "type": "boolean" }'
"ACTID"           | '{ "type": "number" }'
```
## Verification

I did local tests with this and everything seemed to be working great.

## Documentation

We need to document the above formats so that API users know the formats to use.

## Future work

Automated tests for this feature.

### DIFF
--- a/merlin-server/sql/merlin/tables/activity.sql
+++ b/merlin-server/sql/merlin/tables/activity.sql
@@ -10,6 +10,7 @@ create table activity (
   start_offset interval not null,
   type text not null,
   arguments merlin_argument_set not null,
+  metadata merlin_activity_directive_metadata_set default '{}'::jsonb,
 
   constraint activity_synthetic_key
     primary key (id),
@@ -48,6 +49,8 @@ comment on column activity.type is e''
   'The type of the activity, as defined in the mission model associated with the plan.';
 comment on column activity.arguments is e''
   'The set of arguments to this activity, corresponding to the parameters of the associated activity type.';
+comment on column activity.metadata is e''
+  'The metadata associated with this activity.';
 
 
 create function increment_revision_on_insert_activity()
--- a/merlin-server/sql/merlin/tables/activity.sql
+++ b/merlin-server/sql/merlin/tables/activity.sql
@@ -138,3 +138,59 @@ execute function activity_set_updated_at();
 
 comment on trigger set_timestamp on activity is e''
   'Sets the last_modified_at field of an activity to the current time.';
+
+create function check_activity_directive_metadata()
+returns trigger
+security definer
+language plpgsql as $$
+  declare
+    _key text;
+    _value jsonb;
+    _schema jsonb;
+    _type text;
+    _subValue jsonb;
+  begin
+  for _key, _value in
+    select * from jsonb_each(new.metadata::jsonb)
+  loop
+    select schema into _schema from activity_directive_metadata_schema where key = _key;
+    _type := _schema->>'type';
+    if _type = 'string' then
+      if jsonb_typeof(_value) != 'string' then
+        raise exception 'invalid metadata value for key %. Expected: string, Received: %', _key, _value;
+      end if;
+    elsif _type = 'long_string' then
+      if jsonb_typeof(_value) != 'string' then
+        raise exception 'invalid metadata value for key %. Expected: string, Received: %', _key, _value;
+      end if;
+    elsif _type = 'boolean' then
+      if jsonb_typeof(_value) != 'boolean' then
+        raise exception 'invalid metadata value for key %. Expected: boolean, Received: %', _key, _value;
+      end if;
+    elsif _type = 'number' then
+      if jsonb_typeof(_value) != 'number' then
+        raise exception 'invalid metadata value for key %. Expected: number, Received: %', _key, _value;
+      end if;
+    elsif _type = 'enum' then
+      if (_value not in (select * from jsonb_array_elements(_schema->'enumerates'))) then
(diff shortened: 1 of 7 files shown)


merlin-server/sql/merlin/tables/activity_directive_metadata_schema.sql
```diff
@@ -0,0 +1,31 @@
+create table activity_directive_metadata_schema (
+  key text not null primary key,
+  schema jsonb not null,
```
goetzrrGit, line 3: Did you want `schema` to be of Domain `merlin_activity_directive_metadata_set`

goetzrrGit: I think something like this 

```
create table activity_directive_metadata_schema (
  key text not null primary key,
  schema merlin_activity_directive_metadata_set not null,
```

goetzrrGit (changes requested):
Pretty cool that you can embed logic into the SQL to check the different metadata types. 

I read up on `Domain` as I wasn't familiar with the concept. I had a possible change request.

merlin-server/sql/merlin/tables/activity_directive_metadata_schema.sql
dyst5422: No - though perhaps we should apply a similar domain here. The `merlin_activity_directive_metadata_set` is the actual value metadata, whereas this table holds the schema for that metadata

goetzrrGit (approved):

merlin-server/sql/merlin/tables/activity_directive_metadata_schema.sql
```diff
@@ -0,0 +1,31 @@
+create table activity_directive_metadata_schema (
+  key text not null primary key,
```
mattdailis, line 2: If I understand correctly, we want these schema to be defined per mission model - I think we want to add a `model_id` column to this table, and include it in the primary key.

dyst5422: The design here was one schema per deployment, not per mission model. I'll let @patkenneally or @bashbash change their minds on that if they want though.

mattdailis: I see. I do remember that discussion now that you mention it. Resolving this conversation, then.

merlin-server/sql/merlin/tables/activity.sql
```diff
@@ -83,3 +86,59 @@ create trigger increment_revision_on_delete_activity_trigger
 after delete on activity
 for each row
 execute function increment_revision_on_delete_activity();
+
+create function check_activity_directive_metadata()
```
mattdailis, line 142: This is really awesome - I'm impressed both by postgres's json capabilities, and how quickly you've learned to use them.

Just to confirm a scenario in my head: this validation happens on every insert or update.
1. A key in the schema is changed to have a different type
2. An existing activity's start time is updated (e.g. by dragging in the UI) (let's say that this activity has the old type for that particular metadata field)
3. This function will run, and reject the update (essentially "freezing" all updates to that activity until the metadata is fixed)

Now, perhaps the answer is "don't change the schema for existing keys, just make new keys and delete old ones" - I just want to confirm that that's the desired behavior.

dyst5422: I think the appropriate handling of that scenario is to clear out the metadata field in a second submission from the client to update - try-fail-retry, assuming it isn't already setting the value with the "new" schema.

dyst5422: I wonder if we can just do the validation when the specific metadata field changes...

mattdailis (approved):

PR #256, opened 2022-07-19
[AERIE-1954] Add the container scan action to the publish workflow

patkenneally:
**Tickets addressed:** AERIE-1954
* **Review:** By commit  

## Description
Adds a container scanning github action to the publish workflow. 

## Verification
Verification will be having the updated workflow execute the new action.

## Documentation
No additions/changes required

## Future work
None

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,52 +45,108 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
-      # Build aerie-merlin Docker artifacts.
+      # Build, scan, and push aerie-merlin Docker artifacts.
       - name: Extract metadata (tags, labels) for aerie-merlin Docker image
         id: aerieMerlin
         uses: docker/metadata-action@v3
         with:
           images: ${{ env.REGISTRY }}/nasa-ammos/aerie-merlin
-      - name: Build and push aerie-merlin Docker image
+      - name: Build aerie-merlin Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: ./merlin-server
+          load: true
+          tags: ${{ env.REGISTRY }}/nasa-ammos/aerie-merlin:${{ github.sha }}
+      - name: Scan aerie-merlin Docker image
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ${{ env.REGISTRY }}/nasa-ammos/aerie-merlin:${{ github.sha }}
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'CRITICAL'
+      - name: Push aerie-merlin Docker image
         uses: docker/build-push-action@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -51,6 +51,10 @@ jobs:
         uses: docker/metadata-action@v3
         with:
           images: ${{ env.REGISTRY }}/nasa-ammos/aerie-merlin
+      - name: Scan container
+        uses: Azure/container-scan@v0
+        with:
+          image-name: ${{ env.REGISTRY }}/nasa-ammos/aerie-merlin
```
camargo: Is this the correct order? I think we want to scan the image after we build it.
Also since we are using the `docker/build-push-action` which auto-labels the image with the proper tags, I think we might need to specify the tag in the `Azure/container-scan` step.

skovati (commented):
Do we want to break the workflow up so that it's
build -> scan -> push ?

Otherwise, based on Chris's comment it seems we would optimistically push our (possibly vulnerable) images to ghcr, and _then_ scan them. I think the publish workflow failing before push if vulnerabilities above a certain threshold are detected would be slick.

.github/workflows/publish.yml
patkenneally: Yes this makes sense. I wasn't clear what metadata action was doing. I'll switch it around.

skovati:
Removed the draft label as I think this PR is mature enough. I switched to using the underlying [`trivy`](https://github.com/aquasecurity/trivy-action) scanning action instead of the [`Azure/container-scan`](https://github.com/Azure/container-scan) action so we could get more control over trivy options.

This PR breaks the up the build steps, so that we build each docker image, scan it, and then only if the scan detects no critical vulnerabilities, does it finally push the cached image to the container registry.

I considered uploading `trivy`'s scan results to our Github code scanning page, since [trivy offers outputting in `sarif` form](https://github.com/aquasecurity/trivy-action#using-trivy-with-github-code-scanning)

Additionally, I considered creating a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for this build, scan, push combo, which would significantly reduce the copy pasting done here.

camargo (approved):

goetzrrGit (approved):
I see the required actions listed in the right order. When it merges we can see if it works on the develop branch

camargo:
When doing a release this morning we saw an issue with the trivy image scanner. When we tag a release it creates two tags on the image (latest and the version string). This caused the image scanner to fail: https://github.com/NASA-AMMOS/aerie-gateway/runs/7749247283?check_suite_focus=true

```
2022-08-09T15:15:59.724Z	FATAL	multiple targets cannot be specified
```

skovati:
Ah sorry that case slipped through my testing since I didn't do an actual release, just checked that the develop images were working. I've updated the action so it first builds and caches images with their commit SHA, scans the images referencing their SHA, and then publishes to the registry with the correct tags, i.e. either just `develop` or both `latest` and `v0.X.X`

PR #140, opened 2022-04-13
[AERIE-1685] Validate plan prior to simulating 

pcrosemurgy:
* **Tickets addressed:** AERIE-1685
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This PR adds plan activity instance validation to the simulate workflow.

Some previous discussions:
- Failure cases: https://jpl.slack.com/archives/GHK90JHH8/p1649903505320229
- Broad to-do:   https://jpl.slack.com/archives/GJ07FGRPH/p1649786627113889
- Tickets:       https://jpl.slack.com/archives/GJ07FGRPH/p1649990579962809

Four failure modes are now correctly accumulated within the simulation engine:
  1. Unconstructable argument.
    - **Note:** the use of singular "argument" is intentional here, see future work section.
  2. Extraneous parameter.
    - **Note:** the use of singular "argument" is intentional here, see future work section.
  3. Missing arguments.
  4. Nonexistent activity type.

## Verification

<details>
<summary>Expand</summary>

An activity that will result in an **unconstructable argument** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"300 seconds\", type: \"BiteBanana\", arguments: { biteSize: true } }) {
    id
  }
}
```

An activity that will result in an **extraneous parameter** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"310 seconds\", type: \"BiteBanana\", arguments: { bad: 42 } }) {
    id
  }
}
```

An activity that will result in a **missing arguments** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"320 seconds\", type: \"BakeBananaBread\", arguments: { tbSugar: 42 } }) {
    id
  }
}
```

An activity that will result in a **nonexistent activity type** error is inserted with:
```gql
mutation {
  insert_activity_one(object: { plan_id: $plan_id, start_offset: \"330 seconds\", type: \"BadActivity\", arguments: { } }) {
    id
  }
}
```

To trigger this message a simple `simulate` call is needed:
```gql
query {
  simulate(planId:1) {
    reason
    results
    status
  }
}
```

This results in:
```json
{
  "data": {
    "simulate": {
[...]

PR #258, opened 2022-07-19
[AERIE-1826] Accumulate activity/configuration instantiation errors

pcrosemurgy:
* **Tickets addressed:** AERIE-1826
* **Review:** By commit  

## Description

Following #140 this PR properly accumulates instantiation errors for:
- Unconstructable (unserializable) arguments
- Extraneous arguments
- Missing arguments

Previously these failure cases were handled by a single exception and would fail-fast, resulting in a maximum of one argument failure per activity/configuration. With this PR an `InvalidArgumentsException` reports all unconstructable/extraneous/missing arguments that were encountered.

## Verification

<details>
<summary>Expand</summary>

```gql
  getActivityEffectiveArguments
  (
    missionModelId: 1,
    activityTypeName: "BakeBananaBread",
    activityArguments: { tbSugar: "hi", extra: 1 }
  ) {
    success
    errors
    arguments
  }
```
→
```json
{
  "data": {
    "getActivityEffectiveArguments": {
      "success": false,
      "errors": {
        "unconstructableArguments": [
          {
            "name": "tbSugar",
            "failure": "Expected integral number, got StringValue[value=hi]"
          }
        ],
        "missingArguments": [
          "tbSugar",
          "glutenFree"
        ],
        "extraneousArguments": [
          "extra"
        ]
      },
      "arguments": {
        "temperature": 350
      }
    }
  }
}
```

See future work section for serialization/structuring of these exceptions within HTTP/GQL responses.

</details>

## Documentation

Updated `FooActivityMapper` example within `aerie.wiki`'s `Activity-Mappers.md` (https://github.com/NASA-AMMOS/aerie/wiki/Activity-Mappers#generated-activity-mapper-example).

## Future work

Captured in https://jira.jpl.nasa.gov/browse/AERIE-1968:

Work is needed to decide where these exceptions should be serialized/structured. For some endpoints it may be enough to just call `toString()` while others may benefit from a structured response. Ultimately the UI's needs may drive these decisions.

Specifically, the `validatePlan` endpoint returns a `reason` message with that is currently unstructured. This field and the `simulation_dataset.reason` column within our Merlin DB should likely be structured. This will also require looking at all instances of `writer.failWith()`:
```
merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SynchronousSimulationAgent.java
merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```


merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MapperMethodMaker.java
```diff
@@ -175,15 +177,107 @@ public final MethodSpec makeGetValidationFailuresMethod() {
...
+          .addStatement(
+              "$L.throwIfAny()",
+              "unconstructableArgsExBuilder");
+    }
+
+    mb = mb
+        .addStatement(
+            "$L.throwIfAny()",
+            "extraneousArgsExBuilder")
+        .addCode("\n");
```
mattdailis: 🔶 I'm wondering what we gain or lose from separating these three illegal argument exception types (Unconstructable, Extraneous, and Missing). 

One thing we lose is that we only ever throw one of these three, but the same activity could have all three simultaneously.

I'm wondering if collecting all of this information into one exception, and throwing it at the end, would allow us to both provide the maximum amount of feedback to users, and avoid cluttering the MissionModel interface with several related exception types.

pcrosemurgy: Very good point. I'm cool with going this direction! I'll rework things today to just throw one exception.

pcrosemurgy: This change has now been made :)

dyst5422 (approved):
I like the approach here! 👍

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
```diff
@@ -32,57 +31,47 @@ Map<String, ActivityType> getActivityTypes(String missionModelId)
   throws NoSuchMissionModelException;
   // TODO: Provide a finer-scoped validation return type. Mere strings make all validations equally severe.
   List<String> validateActivityArguments(String missionModelId, SerializedActivity activity)
-  throws NoSuchMissionModelException, TaskSpecType.UnconstructableTaskSpecException, MissingArgumentsException;
+      throws NoSuchMissionModelException, InvalidArgumentsException;
```
mattdailis: Rogue indentation

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```diff
@@ -102,8 +102,7 @@ public void shouldNotGetActivityTypeForNonexistentActivityType() {
 
     @Test
     public void shouldInstantiateActivityInstance()
-    throws MissionModelFacade.NoSuchActivityTypeException, MissionModelFacade.MissionModelContractException,
-           TaskSpecType.UnconstructableTaskSpecException, MissingArgumentsException
+        throws MissionModelFacade.NoSuchActivityTypeException, InvalidArgumentsException
```
mattdailis: Rogue indentation

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/UnconstructableArgumentException.java
```diff
@@ -0,0 +1,11 @@
+package gov.nasa.jpl.aerie.merlin.protocol.types;
+
+public final class UnconstructableArgumentException extends Exception {
```
mattdailis, line 3: 🔵 I suspect we could do away with this exception (since we're throwing it then immediately catching it), but I won't insist on that for this PR.

mattdailis (approved):
Looks good to me - a couple formatting comments, and one musing about UnconstructableArgumentException that can be ignored if desired.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
pcrosemurgy: Fixed, thanks!

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
pcrosemurgy: Also fixed :)

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/UnconstructableArgumentException.java
pcrosemurgy: Oh, yeah that try/catch is kind of annoying. It’s needed since the only way to get a failure string is to use `getSuccessOrThrow()`, which only accepts a function that throws an exception.

For an example:
```java
    for (final var entry : arguments.entrySet()) {
      try {
        switch (entry.getKey()) {
          case "x":
            template.x = this.mapper_x.deserializeValue(entry.getValue())
                .getSuccessOrThrow(failure -> new UnconstructableArgumentException("x", failure));
            break;
          case "y":
            template.y = this.mapper_y.deserializeValue(entry.getValue())
                .getSuccessOrThrow(failure -> new UnconstructableArgumentException("y", failure));
            break;
          case "vecs":
            template.vecs = this.mapper_vecs.deserializeValue(entry.getValue())
                .getSuccessOrThrow(failure -> new UnconstructableArgumentException("vecs", failure));
            break;
          default:
            invalidArgsExBuilder.withExtraneousArgument(entry.getKey());
        }
      } catch (final UnconstructableArgumentException e) {
        invalidArgsExBuilder.withUnconstructableArgument(e.parameterName, e.failure);
      }
    }
```

Using the `getSuccessOrThrow()` defined here: https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Result.java?plain=1#L112

The `throws Throws` requires us to throw an exception if we want to cache/save `failure`. We could maybe change/add to `Result` to have a method that allows us to not `throws Throws`. Maybe I'm missing some other simpler option, open to other ideas :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
pcrosemurgy: Marking as resolved to merge PR :)

PR #260, opened 2022-07-21
[AERIE-1953] Interval split operation for Windows and Spans

JoelCourtney:
* **Tickets addressed:** AERIE-1953
* **Review:** By commit  

## Description
Implements the `split(N)` operation, which splits each window into N equally sized sub-windows by removing N-1 points in the middle. Instantaneous point windows are left unchanged.

In practice since windows are defined in microseconds, the sub-windows might not be *exactly* equal, because the last sub-window might be up to N-1 microns longer than the others if the original duration is not a multiple of N. Also, if the original window is M microns long where M < N (however unlikely that might be), only M sub-windows will be produced.

Commits:
1. I added a basic parser for a nullable value. It is essentially `chooseP(your parser, nullP)` with some extra processing that turns it into `Optional<your parser's output>`.
2. Converts the `CodeLocation` record in ConstraintsCompilationError to have nullable fields, because runtime errors don't report code location
    - The TS API throws a runtime error if N < 1
4. Implements split
5. Tests split

## Verification
Plenty of tests, see last commit.

## Documentation
The API method is doc-commented. I'll update wiki after approval but before merge.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
@@ -1,6 +1,5 @@
 package gov.nasa.jpl.aerie.constraints.time;
 
-import gov.nasa.jpl.aerie.constraints.UnsplittableIntervalException;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 
 import java.util.ArrayList;
@@ -13,7 +12,7 @@ import java.util.function.Predicate;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
-import static gov.nasa.jpl.aerie.constraints.time.Window.Inclusivity.Exclusive;
+import gov.nasa.jpl.aerie.constraints.time.Interval.Inclusivity;
 
 /**
  * A collection of intervals that can overlap.
@@ -69,8 +68,18 @@ public class Spans implements IntervalContainer<Spans>, Iterable<Interval> {
     return new Spans(this.intervals.stream().filter(filter).toList());
   }
 
+  /**
+   * Splits each span into sub-spans.
+   *
+   * @param numberOfSubSpans number of sub-spans for each span.
+   * @param internalStartInclusivity Inclusivity for any newly generated span start points (default Inclusive in eDSL).
+   * @param internalEndInclusivity Inclusivity for any newly generated span end points (default Exclusive in eDSL).
+   * @return a new Spans
+   * @throws UnsplittableSpanException if any span contains only one point or contains fewer microseconds than `numberOfSubSpans`.
+   * @throws UnsplittableSpanException if any span contains {@link Duration#MIN_VALUE} or {@link Duration#MAX_VALUE} (representing unbounded intervals)
+   */
   @Override
-  public Spans split(final int numberOfSubSpans) {
+  public Spans split(final int numberOfSubSpans, final Inclusivity internalStartInclusivity, final Inclusivity internalEndInclusivity) {
     if (numberOfSubSpans == 1) {
       return new Spans(this);
     }
@@ -82,20 +91,32 @@ public class Spans implements IntervalContainer<Spans>, Iterable<Interval> {
 
       // We throw an exception if the interval contains fewer microseconds than the requested number of sub-spans.
       if (x.isSingleton()) {
-        throw new UnsplittableIntervalException("Cannot split an instantaneous interval into " + numberOfSubSpans + " pieces.");
+        throw new UnsplittableSpanException("Cannot split an instantaneous span into " + numberOfSubSpans + " pieces.");
       } else if (numberOfMicroSeconds < numberOfSubSpans) {
-        throw new UnsplittableIntervalException("Cannot split an interval only " + numberOfMicroSeconds + " microseconds long into " + numberOfSubSpans + " pieces.");
+        throw new UnsplittableSpanException("Cannot split a span only " + numberOfMicroSeconds + " microseconds long into " + numberOfSubSpans + " pieces.");
+      }
+
+      // Throw an exception if trying to split an "unbounded" interval.
+      // It is unlikely that a user will ever need to split a Windows or Spans that includes +/- infinity.
+      //
+      // If they do, and it is a legitimate use case that we should support, this block should be replaced
+      // with a check that returns the unbounded interval unchanged, because the split points on an unbounded
+      // interval will be outside the finite range of `Duration`.
+      if (x.contains(Duration.MIN_VALUE)) {
+        throw new UnsplittableSpanException("Cannot split an unbounded span. (interval contains MIN_VALUE, which is a stand-in for -infinity.");
+      } else if (x.contains(Duration.MAX_VALUE)) {
+        throw new UnsplittableSpanException("Cannot split an unbounded span. (interval contains MAX_VALUE, which is a stand-in for +infinity.");
       }
 
       var cursor = Duration.add(x.start, width);
-      final List<Window> ret = new ArrayList<>();
-      ret.add(Window.between(x.start, x.startInclusivity, cursor, Exclusive));
+      final List<Interval> ret = new ArrayList<>();
+      ret.add(Interval.between(x.start, x.startInclusivity, cursor, internalEndInclusivity));
       for (int i = 1; i < numberOfSubSpans - 1; i++) {
         final var nextCursor = Duration.add(cursor, width);
-        ret.add(Window.between(cursor, Exclusive, nextCursor, Exclusive));
+        ret.add(Interval.between(cursor, internalStartInclusivity, nextCursor, internalEndInclusivity));
         cursor = nextCursor;
       }
-      ret.add(Window.between(cursor, Exclusive, x.end, x.endInclusivity));
+      ret.add(Interval.between(cursor, internalStartInclusivity, x.end, x.endInclusivity));
       return ret.stream();
     });
   }
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
@@ -1,5 +1,8 @@
 package gov.nasa.jpl.aerie.constraints.time;
 
+import gov.nasa.jpl.aerie.constraints.UnsplittableIntervalException;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
@@ -10,10 +13,12 @@ import java.util.function.Predicate;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import static gov.nasa.jpl.aerie.constraints.time.Window.Inclusivity.Exclusive;
+
 /**
  * A collection of intervals that can overlap.
  */
-public class Spans implements Iterable<Interval> {
+public class Spans implements IntervalContainer<Spans>, Iterable<Interval> {
   private final List<Interval> intervals;
 
   public Spans() {
@@ -62,6 +67,37 @@ public class Spans implements Iterable<Interval> {
 
   public Spans filter(final Predicate<Interval> filter) {
     return new Spans(this.intervals.stream().filter(filter).toList());
+  }
+
+  @Override
+  public Spans split(final int numberOfSubSpans) {
+    if (numberOfSubSpans == 1) {
+      return new Spans(this);
+    }
+    return this.flatMap(x -> {
+      // Width of each sub-window, rounded down to the microsecond
+      final var width = Duration.divide(Duration.subtract(x.end, x.start), numberOfSubSpans);
+
+      final var numberOfMicroSeconds = Duration.subtract(x.end, x.start).in(Duration.MICROSECOND);
+
+      // We throw an exception if the interval contains fewer microseconds than the requested number of sub-spans.
+      if (x.isSingleton()) {
+        throw new UnsplittableIntervalException("Cannot split an instantaneous interval into " + numberOfSubSpans + " pieces.");
+      } else if (numberOfMicroSeconds < numberOfSubSpans) {
+        throw new UnsplittableIntervalException("Cannot split an interval only " + numberOfMicroSeconds + " microseconds long into " + numberOfSubSpans + " pieces.");
+      }
+
+      var cursor = Duration.add(x.start, width);
+      final List<Window> ret = new ArrayList<>();
+      ret.add(Window.between(x.start, x.startInclusivity, cursor, Exclusive));
+      for (int i = 1; i < numberOfSubSpans - 1; i++) {
+        final var nextCursor = Duration.add(cursor, width);
+        ret.add(Window.between(cursor, Exclusive, nextCursor, Exclusive));
+        cursor = nextCursor;
+      }
+      ret.add(Window.between(cursor, Exclusive, x.end, x.endInclusivity));
+      return ret.stream();
+    });
   }
 
   @Override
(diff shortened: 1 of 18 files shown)


JoelCourtney:
Marking as draft because I'm about to open a PR for basic Spans. Will implement Spans.split as well.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsCompilationError.java
```diff
@@ -30,7 +31,7 @@ public class ConstraintsCompilationError {
 
   public static final JsonParser<List<UserCodeError>> constraintsErrorJsonP = listP(userCodeErrorP);
 
-  public record CodeLocation(Integer line, Integer column) {}
+  public record CodeLocation(Optional<Integer> line, Optional<Integer> column) {}
```
dyst5422: Do you know the case under which you found the code location to be missing? Just making sure we're not missing some root problem here

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Windows.java
```diff
@@ -137,6 +138,44 @@ public Windows shiftBy(Duration fromStart, Duration fromEnd){
...
+        // If there are fewer microseconds in the window than we can split, make one window per microsecond.
+        // This is likely only to come up for instantaneous windows, which we should leave unchanged.
+        final var numberOfMicroSeconds = Duration.subtract(x.end, x.start).in(Duration.MICROSECOND);
+        final int adjustedSubWindows;
+
+        if (x.isSingleton()) {
+          ret.add(Window.at(x.start));
+          return;
+        } else if (numberOfMicroSeconds < numberOfSubWindows) {
+          adjustedSubWindows = (int) numberOfMicroSeconds;
```
dyst5422: Are we sure this behavior is something we want to do without informing the users? I can imagine an activity that does some value incrementation that they would want to operate exactly as many times as specified or be notified otherwise.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsCompilationError.java
JoelCourtney: Its during runtime errors. I added a check that the argument is greater than 0, which throws an error in the typescript.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Windows.java
JoelCourtney: @dyst5422 So should it throw an error if the window is too short? My general idea is that if the the window is too short, it is essentially like generating exclusive windows with 0 duration, which will get removed. I can throw an error instead if you think that makes more sense. But do you agree that splitting an instant should just return the instant unchanged?

dyst5422: I have a feeling that the cases where this would be an issue (there aren't enough microseconds to split) are very small and it probably indicates some breakdown elsewhere if we're trying to split 3 microseconds into 4 activities. As such, I'm inclined to say we throw an error and double check with the users.

I haven't thought about the use cases as in depth as you have, but I struggle to identify a case where we'd be splitting an instantaneous activity that isn't similarly a breakdown in process elsewhere. The case of "I have an instantaneous 'event' that I want mapped into 4 different parallel activities" I think should be accomplished by other means/operators.

I prefer defaulting to the more conservative option (erroring in this case) as the safer solution as a general rule of thumb when there is uncertainty in the desired behavior (we apply this generously in SSG on APSS where we create errors anytime something isn't definite for example).

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsCompilationError.java
dyst5422: Based on the test case here, https://github.com/NASA-AMMOS/aerie-ts-user-code-runner/blob/develop/test/UserCodeRunner.spec.ts#L11-L45 I would expect that the code location should be returned. Can you help me recreate the case where you saw this so that I can ensure that we are indeed returning a code location and I can add a regression test to the code runner?

dyst5422: And as a principle, I think that ANY Error coming from the code runner through normal operation should return a location. The exceptions are when we incorrectly configure the supporting files (like what happened recently with the mission model generated code for the scheduler), there is a legitimate bug in the code runner, or there is some truly exceptional nodejs runtime failure

JoelCourtney: If you revert the changes to `ConstraintsCompilationError.java` to remove the `Optional` ints and `nullable` parsers, the `testSplitArgumentError` test will fail. This is the test to make sure that it correctly throws an error if the argument is less than 1.

The test fails because it cannot parse the error, because its coming in with `{line: null, column: null}`. I'm not sure why it doesn't have location, but this is the code that produces the error:

```ts
public split(numberOfSubWindows: number): Windows {
  if (numberOfSubWindows < 1) {
    throw RangeError(".split numberOfSubWindows cannot be less than 1, but was: " + numberOfSubWindows);
  }
  return new Windows({
    kind: AST.NodeKind.IntervalsExpressionSplit,
    intervals: this.__astNode,
    numberOfSubIntervals: numberOfSubWindows
  })
}
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Windows.java
JoelCourtney: I'm ok with this. I agree, having tiny windows like that is going to be at least an extremely rare, if not non-existent use case. For instants, I also bet it will be rare that users accidentally end up with a instant window mixed in with non-instant windows, so I guess its not a big deal. (It is possible though, if they use something like `resource.greaterThanOrEqual(2)` and the resource has a local maximum at exactly 2.)

I'll turn both cases into an error. If it becomes an issue for a user, we can reevaluate later.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ConstraintsCompilationError.java
dyst5422: Ah, I don't have a regression test for errors thrown in additional files. I'll add one and ensure we properly map back the errors to the location in the users code.

Thanks for finding this @JoelCourtney

JoelCourtney: Ok, I'll remove the `Optional`'s then. I'm still going to keep the `nullable` parser implementation though, for when nulls come up again. (I think we'll end up using it for external profiles.)

dyst5422: https://jira.jpl.nasa.gov/browse/AERIE-1987 for reference.

Easy to debug, easy to fix. Should have it PR'd for you within the hour.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Windows.java
JoelCourtney: done

JoelCourtney:
I've marked ready for review now that this PR includes Spans! This change introduces the `IntervalContainer` interface, which will be the unifying interface between `Windows` and `Spans`. Any operations that will exist on both container types (such as `Split` and many others) should be added to this interface. This way the `Split` node doesn't have to be duplicated for `WindowsSplit` and `SpansSplit`, and can just use the interface.

dyst5422 (approved):
LGTM Thanks again for finding the user code runner bug!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -64,6 +69,37 @@ public Spans filter(final Predicate<Window> filter) {
...
+      } else if (numberOfMicroSeconds < numberOfSubSpans) {
+        throw new UnsplittableIntervalException("Cannot split an interval only " + numberOfMicroSeconds + " microseconds long into " + numberOfSubSpans + " pieces.");
+      }
+
+      var cursor = Duration.add(x.start, width);
+      final List<Window> ret = new ArrayList<>();
+      ret.add(Window.between(x.start, x.startInclusivity, cursor, Exclusive));
+      for (int i = 1; i < numberOfSubSpans - 1; i++) {
+        final var nextCursor = Duration.add(cursor, width);
+        ret.add(Window.between(cursor, Exclusive, nextCursor, Exclusive));
```
adrienmaillard: Why are the intervals exclusive at both ends ? should it not be inclusive for start exclusive for end ? Shouldn't the union of all split intervals be equal to the original interval ?

adrienmaillard (commented):
I only have one (quite fundamental) question but otherwise all of this looks good to me.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
adrienmaillard: I know it is part of your specification to "remove" points but I am wondering why. If we talk about what it would mean in terms of scheduling, it would mean we can't schedule activities starting at those points.

JoelCourtney: For Windows, if any of the internal boundaries were inclusive, the Windows class would coalesce them back into the original window, so they have to be exclusive. The original external boundaries keep the same inclusivity though.

For Spans, we could have it that the spans meet at a point, but I want it to have the same general result as Windows so that if the user converts to windows, their split spans don't coalesce back.

JoelCourtney: > If we talk about what it would mean in terms of scheduling, it would mean we can't schedule activities starting at those points.

I'd love it if we could have the sub-windows be left-inclusive and right-exclusive for exactly that reason, but unfortunately `[0, 1)  union  [1, 2)` is equal to `[0, 2)` because they contain the same points. So they are coalesced.

adrienmaillard: I thought one of the motivation for adding `Spans` was that we could allow the correct behavior for splitting. Actually, the `split` operation, whatever the object it is called from, should return a `Spans` just because it inherently produces sets of contiguous intervals. I understand it is more beautiful if we have only one implementation and it is generic, etc, but if we do that, the meaning of the inclusivity bounds is kind of lost.
When the user goes from `Spans` to `Windows`, he knows/should know that contiguous intervals will be coalesced.

JoelCourtney: Ok, I *almost* agree. I think it will be rare in scheduling to need to split windows. But what about constraints and global scheduling conditions, where the final object is Windows? It might be contrived, but I don't think its unreasonable to need it. Example:

> There must be at least one ActivityA every hour (i.e. locked to the hours of the clock, instead of a recurrence-goal-like constraint where we care about spacing)

I'd like to write:

```
dateResource.changes().invert()
    .split(24)
    .instanceCardinality(ActivityType.ActivityA)
    .greaterThan(0)
```

But if split produces Spans that will coalesce when converted to Windows, I'll have to write this:

```
dateResource.changes().invert()
    .split(24)
    .setBoundsInclusivity(Exclusive, Exclusive)
    .windows()
    .instanceCardinality(ActivityType.ActivityA) // this function won't be defined on Spans
    .greaterThan(0)
```

I guess that's not so bad. If you think that's a reasonable pattern, I'm OK making `split` always produce Spans.

adrienmaillard: I thought everything related with defining goals was going to happen from `Spans` and not from `Windows` so I assume the first example you wrote would work. 
Edit: it is a constraint !

JoelCourtney: That first example is for a constraint, sorry that wasn't clear. This kind of use case would usually be covered by a goal, but perhaps they want to place the needed activities by hand. Again, kinda contrived, I know.

JoelCourtney:
Ok! Now that nullable windows is merged, the situation for split is a little different.

I'm thinking since windows gaps mean "unknown" instead of "undefined", it would technically be unknown where to split a window if a `true` segment neighbors a gap. So, I'm thinking split should throw an error if called on windows that have gaps.

Additionally, I'd like to make the inclusivity of any generated internal edges be configurable, with the defaults being `internalStartInclusivity = Inclusive, internalEndInclusivity = Exclusive`. The general patterns would be this:

- If calling split on Spans, just don't worry about it because its straightforward.
- If calling split on Windows:
    1. if your windows has gaps that you want to keep, save a copy of your windows object for step 5
    2. call `assignGaps(true | false)` (TODO) to remove any gaps
    3. call `split(N, startInclusivity?, endInclusivity?)`, which always produces `Spans` just in case the chosen inclusivities would immediately cause coalescing (which would happen by default)
    4. either use your `Spans` object, or call `.windows()` and do step 5
    5. to optionally regain gaps, call `.select(oldCopy)` (TODO)

example code:

```ts
export default(): Constraint {
    let unsplitWindows = /* some external profile expression with gaps */;
    let splitWindows = unsplitWindows
        .assignGaps(false)
        .split(3, Inclusivity.Exclusive, Inclusivity.Exclusive)
        .windows()
        .select(unsplitWindows);
    // do whatever with your shiny new windows object
}
```

At some point if performance becomes a concern it may be helpful to implement a `.cache()` operation to avoid reevaluating things like `unsplitWindows`. The ast node would hash its children and store/lookup results in the `environment` variable in the expression interface. But that's a separate issue.

@adrienmaillard @dyst5422 @Twisol any thoughts?

Twisol:
I think my first concern is what "N equally-sized windows" means if you happen to have a profile like `true[-inf, 0] false[0,3] true[3, inf]` (or heck, `true[-inf, inf]`). More generally, how should this behave on unbounded intervals?

(I know we use MIN_VALUE right now and not a proper -inf, but MIN_VALUE itself is a completely arbitrary value.)

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: > More generally, how should this behave on unbounded intervals?

I suppose as long as we're trying to represent infinite intervals on a finite domain, simply leaving windows that touch `MIN_VALUE` or `MAX_VALUE` unchanged is fairly accurate.

JoelCourtney:
> More generally, how should this behave on unbounded intervals?

I suppose as long as we're trying to represent infinite intervals on a finite domain, simply leaving windows that touch `MIN_VALUE` or `MAX_VALUE` unchanged is fairly accurate.

Twisol:
D'you think there are any valid use-cases for calling `split` when unbounded `true` intervals exist? It seems more likely to be the result of an incorrect windows expression -- maybe we'd prefer to `throw` when faced with such intervals, so that the author can correct their logic.

JoelCourtney:
> D'you think there are any valid use-cases for calling `split` when unbounded `true` intervals exist? It seems more likely to be the result of an incorrect windows expression -- maybe we'd prefer to `throw` when faced with such intervals, so that the author can correct their logic.

I can't think of any, so I'll add that check and throw an error. If such a use case ever comes up, it'll be very easy and I've left a comment explaining what to do.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: I've resolved this with a new design: `split` always produces `Spans`. The user can specify what the inclusivity of new start/endpoints should be, and the default is [inclusive, exclusive). If the user wants windows instead of spans, they can convert it back with `.windows()` easily.

adrienmaillard (approved):
Great work.

PR #261, opened 2022-07-21
[REFACTOR] Delete unused merlin-server database actions

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This PR takes the stance that "The merlin-server should only have database actions that it actually uses". Unused database actions (especially untested ones) tend to rot as the database schema changes (see `CreateActivityArgumentsAction`, which refers to a non-existent `activity_argument` table).

The following workflows are all done via hasura or the gateway, and do not touch the merlin-server (and thus, I argue, do not need database actions):
- All plan mutations (creating/deleting plans/activities)
- Creating mission models
- Deleting mission models

Now, another option to prevent these actions from rotting, is to hook them into some of our e2e tests. I'm open to that route, if people think that the Repositories should have all of the CRUD operations, not just the ones that are used.

I've split the changes into separate commits, so if people want to veto some specific change, it should be easy to drop from this PR.

## Verification

Since this PR contains exclusively code deletions, I think it's sufficient to check that it compiles.

## Documentation

These are internal, undocumented parts of the system.

## Future work

As a follow up, I'd like to organize the remaining actions into packages.


camargo (approved):
`+4 −722` 🥇
Ran the full suite of UI e2e tests against this branch and all pass. LGTM!

skovati (approved):
LGTM since all e2e tests are passing! Love seeing Hasura handle a lot of this work now

PR #263, opened 2022-07-25
[AERIE-1908] Jetty-Hasura idle timeout clash

adrienmaillard:
* **Tickets addressed:** AERIE-1908
* **Review:** By commit  

## Description

Hasura keeps connections open to a default of 180 seconds to re-use them. But the default HTTP idle timeout of jetty is 30 seconds so it will close the connection early and display a stack trace in the logs of `merlin-server` and `scheduler-server`.

Note that the stack trace is not an error but the normal way Jetty is processing idle timeouts. The stack trace is printed as a debug message and the exception is not even thrown but passed.  

What would it take to change the idle timeout of jetty servers in both the `scheduler-server` and `merlin-server` ? I found only one way of doing so.
I could not find a way of configuring HTTP idle timeouts of the inner server of jetty. Websockets idle timeouts can be configured but not HTTP idle timeouts.
 
So, we replace the server that Jetty is creating by default by a jetty server that we can configure beforehand.

Remarks/questions : 
- After doing this, the timeout is effectively 180s, I have not seen other differences in how the scheduler behaves but I am not sure the inner server is totally equivalent.  I have compared the `Server` objects by hand and they **seem** to be the same but I do not see how to confirm.  
- Would there be another way of configuring idle timeouts? 

## Verification

I researched/tested this by using the goal editor in `aerie-ui`. No new tests.

## Documentation

None

## Future work

Fix the following that I keep seeing in logs 
```
[main] DEBUG io.netty.util.internal.PlatformDependent0 - jdk.internal.misc.Unsafe.allocateUninitializedArray(int): unavailable
java.lang.IllegalAccessException: class io.netty.util.internal.PlatformDependent0$6 cannot access class jdk.internal.misc.Unsafe (in module java.base) because module java.base does not export jdk.internal.misc to unnamed module @7f1dd5e7
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:394)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:674)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.netty.util.internal.PlatformDependent0$6.run(PlatformDependent0.java:353)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:318)
	at io.netty.util.internal.PlatformDependent0.<clinit>(PlatformDependent0.java:344)
	at io.netty.util.internal.PlatformDependent.isAndroid(PlatformDependent.java:294)
	at io.netty.util.internal.PlatformDependent.<clinit>(PlatformDependent.java:93)
	at io.netty.channel.epoll.Native.loadNativeLibrary(Native.java:287)
	at io.netty.channel.epoll.Native.<clinit>(Native.java:70)
	at io.netty.channel.epoll.Epoll.<clinit>(Epoll.java:39)
	at com.impossibl.postgres.protocol.v30.ServerConnectionFactory.createInetSocketChannel(ServerConnectionFactory.java:298)
	at com.impossibl.postgres.protocol.v30.ServerConnectionFactory.createChannel(ServerConnectionFactory.java:259)
	at com.impossibl.postgres.protocol.v30.ServerConnectionFactory.connect(ServerConnectionFactory.java:150)
	at com.impossibl.postgres.protocol.v30.ServerConnectionFactory.connect(ServerConnectionFactory.java:143)
	at com.impossibl.postgres.protocol.v30.ServerConnectionFactory.connect(ServerConnectionFactory.java:124)
	at com.impossibl.postgres.system.BasicContext.<init>(BasicContext.java:213)
	at com.impossibl.postgres.jdbc.PGDirectConnection.<init>(PGDirectConnection.java:236)
	at com.impossibl.postgres.jdbc.ConnectionUtil.createConnection(ConnectionUtil.java:317)
	at com.impossibl.postgres.jdbc.AbstractDataSource.createConnection(AbstractDataSource.java:125)
	at com.impossibl.postgres.jdbc.PGDataSource.getConnection(PGDataSource.java:71)
	at com.zaxxer.hikari.pool.PoolBase.newConnection(PoolBase.java:359)
	at com.zaxxer.hikari.pool.PoolBase.newPoolEntry(PoolBase.java:201)
	at com.zaxxer.hikari.pool.HikariPool.createPoolEntry(HikariPool.java:470)
	at com.zaxxer.hikari.pool.HikariPool.checkFailFast(HikariPool.java:561)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:100)
	at com.zaxxer.hikari.HikariDataSource.<init>(HikariDataSource.java:81)
	at gov.nasa.jpl.aerie.scheduler.server.SchedulerAppDriver.loadStores(SchedulerAppDriver.java:142)
	at gov.nasa.jpl.aerie.scheduler.server.SchedulerAppDriver.main(SchedulerAppDriver.java:76)

```

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
@@ -31,6 +31,12 @@ import gov.nasa.jpl.aerie.merlin.server.services.ThreadedSimulationAgent;
 import gov.nasa.jpl.aerie.merlin.server.services.TypescriptCodeGenerationServiceAdapter;
 import gov.nasa.jpl.aerie.merlin.server.services.UnexpectedSubtypeError;
 import io.javalin.Javalin;
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.LowResourceMonitor;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.server.handler.StatisticsHandler;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
@@ -38,7 +44,7 @@ import java.nio.file.Path;
 
 public final class AerieAppDriver {
 
-  public static void main(final String[] args) {
+  public static void main(final String[] args) throws InterruptedException {
     // Fetch application configuration properties.
     final var configuration = loadConfiguration();
     final var stores = loadStores(configuration);
@@ -75,8 +81,16 @@ public final class AerieAppDriver {
         simulationAction,
         generateConstraintsLibAction
     );
-
     // Configure an HTTP server.
+    //default javalin jetty server has a QueuedThreadPool with maxThreads to 250
+    final var server = new Server(new QueuedThreadPool(250));
+    final var connector = new ServerConnector(server);
+    connector.setPort(configuration.httpPort());
+    //set idle timeout to be equal to the idle timeout of hasura
+    connector.setIdleTimeout(180000);
+    server.addBean(new LowResourceMonitor(server));
+    server.insertHandler(new StatisticsHandler());
+    server.setConnectors(new Connector[]{connector});
     final var javalin = Javalin.create(config -> {
       config.showJavalinBanner = false;
       if (configuration.enableJavalinDevLogging()) config.enableDevLogging();
@@ -85,6 +99,7 @@ public final class AerieAppDriver {
       config.registerPlugin(new LocalAppExceptionBindings());
       config.registerPlugin(new MissionModelRepositoryExceptionBindings());
       config.registerPlugin(new MissionModelExceptionBindings());
+      config.server(() -> server);
     });
 
     // Start the HTTP server.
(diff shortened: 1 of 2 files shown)


mattdailis:
I think you're right - creating your own jetty server object does seem like the preferred way to set jetty configurations.

Is the error you mention in your future work section pre-existing, or is it introduced by this PR?

adrienmaillard:
> I think you're right - creating your own jetty server object does seem like the preferred way to set jetty configurations.
> 
> Is the error you mention in your future work section pre-existing, or is it introduced by this PR?

The error is pre-existing.

mattdailis (approved):

patkenneally:
@adrienmaillard what aspects of the Jetty Server in the Scheduler are you thinking are not the same after your change? 

It may be the same as what you already did by hand, but would something like the Jetty dump functionality help confirm that the Servers are the same (expect for your change)? 

[Jetty server config - dump settings](https://www.eclipse.org/jetty/documentation/jetty-11/operations-guide/index.html#og-module-server-config)

patkenneally (approved):
Pending the settings dump inspection.

adrienmaillard:
I have dumped the conf jetty configurations and after some rearranging of the files, here is the diff I am getting 
[Dump diff](https://github.com/NASA-AMMOS/aerie/files/9213403/compare_jetty.html.zip)

~~Left is with the "custom" jetty server that this PR proposes. Right is with the original jetty that javalin is creating.~~ 
Right is with the "custom" jetty server that this PR proposes. Left is with the original jetty that javalin is creating.

I do not see major differences but 
- the presence of an `AttributeContainerMap` in the custom one
- the `sendServerVersion` has a different value 
- one parameter is 200 in one, 250 in the other.

patkenneally:
Interestingly the Javalin FAQ says that it servers with a `QueuedThreadPool` size of 250 yet as you've shown it's only 200.

adrienmaillard:
> Interestingly the Javalin FAQ says that it servers with a `QueuedThreadPool` size of 250 yet as you've shown it's only 200.

Yeah, it's because I've reversed which is which... Left is the jetty javalin and right is the custom one.

adrienmaillard:
We can actually set the `ThreadPool` ourselves with max size 250 if we want. After doing that, this discrepancy is gone. The really weird thing is that the `idleTimeout` appears to be -1 in both cases when we know it is not the case. So there are hidden parameters we do not see in these dumps.

PR #265, opened 2022-07-26
[HOTFIX]: Create Partition for External Profiles

pranav-super:
* **Tickets addressed:** (none)
* **Review:** By commit  

## Description

Previously, parititons weren't being created automatically for external profiles. As a result, addition of external profiles would fail with the message 'no partition of relation "profile_segment"'.

## Verification

This was tested by deploying AERIE and adding external datasets, and verifying no failure messages were sent and that the external profile was correctly added to the database (verified by running `SELECT` queries).

## Documentation

None.

## Future work

None.


mattdailis (approved):

camargo (approved):
Tested locally and verified the error is no longer thrown. Using the `addExternalDataset` action returns successfully.

PR #267, opened 2022-07-26
[REFACTOR] Use Jackson for constraints & scheduling eDSL JSON

JoelCourtney:
* **Tickets addressed:** AERIE-???
* **Review:** By commit  

## Description
[Jackson](https://github.com/FasterXML/jackson) is a very popular JSON serialization & deserialization library for Java. This PR converts the parser systems for the Constraints and Scheduling eDSLs to use Jackson instead of our in-house parsing solution.

### Motivation

- Jackson is _much_ easier to learn for newcomers.
- Jackson does not require us to maintain parallel trees for the parsers. For example, the current constraints system has three parallel trees (TS AST, JSON parsers, and Java expression tree), and the current scheduling system has four (including an intermediate builder representation in between the parsers and the Java tree). Instead, the Java classes  already implicitly define a schema which Jackson uses for parsing.
- Jackson is more flexible and has more features already implemented. For example, I needed to implement a mutually recursive pair of parsers for the `windows.spans()` and `spans.windows()` operations. I don't think this is possible given the current `recursiveP` pattern, which only supports direct recursion. I didn't have to do anything to accommodate this in Jackson.
- By default, the JSON schema will be more directly tied to the POJO schema it corresponds to. This means the value of a discriminant `kind` field for polymorphic types will match the class name, and the field names will match the java class members.

### Commits

These changes aren't very separable, and few of them break the rule that Aerie should build at every step of the way. But I've separated them according to what part of the code they deal with. Here are some comments on design decisions:

- I had to include some empty interfaces like `WindowsExpression extends Expression<Windows>` because Java doesn't allow you to access the `.class` of a parameterized class. (So `mapper.readValue(string, Expression<Windows>.class)` is not valid syntax).
- There are a few ways of achieving polymorphism in Jackson, and I chose different methods for Expressions and Goals.
    - Expression uses the `MINIMAL_CLASS` type, which means Jackson derives the value of the discriminant field from the class's path, and removes the path segments that are common to all subtypes. Since all the expression nodes are in the same java module, this becomes `'.<ClassName>'`
    - For Goal, I instead explicitly specified the discriminant field, because `MINIMAL_CLASS` isn't ideal for implementors nested inside a sealed interface.

## Verification
All relevant tests were changed to match the new system if needed. All tests pass.

## Documentation
Unlike our current parsing system, Jackson is documented, and actually suffers from having _too many_ tutorials written about it. :)

## Future work

- I successfully moved all of scheduler-server onto Jackson, but haven't included that in this PR because I don't want it to get any more crowded. If this PR goes well, I want to move on to that and slowly move toward removing all of the old parsing system.
- I bet the `SerializedValueSerializer` and `SerializedValueDeserializer` classes that I wrote for this PR don't actually need to exist. I'll look into it.


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public final class SerializedValueSerializer extends StdSerializer<SerializedValue> {
+
+  public SerializedValueSerializer() {
+    super((Class<SerializedValue>) null);
```
pcrosemurgy: Instead of the `null` cast I'd just call `this(null)` here instead.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
```diff
@@ -0,0 +1,113 @@
...
+  public void serialize(final SerializedValue value, final JsonGenerator gen, final SerializerProvider provider)
+  throws IOException
+  {
+    value.match(new SerializedValue.Visitor<>() {
+      @Override
+      public Object onNull() {
+        try {
+          gen.writeNull();
+        } catch (IOException e) {
+          e.printStackTrace();
```
pcrosemurgy: Might want to consider throwing an `Error` here.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -148,9 +147,13 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
...
       //unwrap failure message from any anticipated exceptions and forward to subscribers
-      writer.failWith("%s\n%s".formatted(
-          e.toString(),
-          SchedulingCompilationError.schedulingErrorJsonP.unparse(e.errors).toString()));
+      try {
+        writer.failWith("%s\n%s".formatted(
+            e.toString(),
+            mapper.writeValueAsString(e.errors)));
+      } catch (JsonProcessingException ex) {
+        ex.printStackTrace();
```
pcrosemurgy, line 155: Throwing a fatal `Error` here seems appropriate.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
pcrosemurgy: A couple other changes I'd make would be to:
- Explicitly make the visitor generic type `Void`. 
- Wrap the `IOException` handling.

For the `IOException` handling you could make some interface like:
```java
private interface JsonGeneratorRunnable {
  void run() throws IOException;
}
```
And make a method that takes one of those as an argument:
```java
private static void tryGenerate(final JsonGeneratorRunnable r) {
  try {
    r.run();
  } catch (final IOException e) {
    throw new Error(...);
  }
}
```
This is just a rough outline, you could change things around to also report the type that was trying to be generated.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
JoelCourtney: It looks like that's needed because the super constructor is overloaded.

pcrosemurgy: Yup! But when using `this(null)` the other constructor that you added will be hit.
```java
  public SerializedValueSerializer(Class<SerializedValue> c) {
    super(c);
  }
```

pcrosemurgy: Using `this()` in this context will basically perform an implicit cast.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
JoelCourtney: done

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
JoelCourtney: TIL github doesn't have a facepalm emoji :)

JoelCourtney: done. thanks!

pcrosemurgy: All good! It's subtle!

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
JoelCourtney: Oops, sorry I didn't see your second comment here. I ended up returning the caught error outside of the visitor and throwing it there. This way it doesn't require changing the SerializedValue visitor interface to allow errors. Is that ok?

pcrosemurgy: Ohh I see! Was wondering why it was changed to return an `IOException` 😉

Sorry to be a pain here but I'm not sure how I feel about catching and returning the `IOException`. This means that callers can decide if they want to ignore the serialization error or not. Instead of giving callers the choice (which by the "default" path of least resistance they will ignore) I think I prefer always throwing an exception/error.

Currently within Aerie the convention has been to throw a `RuntimeException` (see https://stackoverflow.com/questions/20461688/runtimeexception-error for a short explanation for why `Error` is probably not preferred here). If you search for `getSuccessOrThrow()` methods you'll see that pattern.

So for convention and for promising that all serialization issues are indeed thrown I'd suggest always throwing a `RuntimeException` when an `IOException` is hit.

pcrosemurgy: Basically, this is what I have in mind:
```java
  private interface JsonGeneratorRunnable {
    void run() throws IOException;
  }

  private static void generateOrThrow(final JsonGeneratorRunnable r) {
    try {
      r.run();
    } catch (final IOException e) {
      throw new RuntimeException(e);
    }
  }

  @Override
  public void serialize(final SerializedValue value, final JsonGenerator gen, final SerializerProvider provider)
  throws IOException
  {
    value.match(new SerializedValue.Visitor<Void>() {
      @Override
      public Void onNull() {
        generateOrThrow(gen::writeNull);
        return null;
      }
...
```

JoelCourtney: > This means that callers can decide if they want to ignore the serialization error or not.

In this case no, because the visitor is only available to be used inside this Serializer. This is just a workaround because the visitor interface doesn't allow throwing Exceptions. Outside of the visitor implementation, the returned exception is re-thrown. Since its an anonymous local implementation, no other code has the visibility to use that visitor.

pcrosemurgy: Oh! True.

I'd still avoid returning an `IOException` and instead use a wrapper method like `generateOrThrow()` to simplify things. I guess returning an exception instance isn't the worst thing but it feels like an anti-pattern.

pcrosemurgy: What I'm mostly getting at is that I don't think `serialize()` should `throws IOException`. At least I'm comfortable with it not throwing an `IOException` since Aerie's convention has been to throw an unchecked `RuntimeException`.

JoelCourtney: So is the idea that `RuntimeException` doesn't need to be declared in the method signature in the `Visitor` interface?

JoelCourtney: I guess I just don't understand why we'd need to convert from `IOException` to `RuntimeException`.

pcrosemurgy: Yup! I've updated the comment above, I mistakenly kept in the `final var error = ...`. https://github.com/NASA-AMMOS/aerie/pull/267#discussion_r930491839

pcrosemurgy: > I guess I just don't understand why we'd need to convert from `IOException` to `RuntimeException`.

Ahh I see! A `RuntimeException` is unchecked, whereas an `IOException` is checked. So when throwing a `RuntimeException` you don't need to add it to the `throws` list (nor do callers need to explicitly `catch` it).

JoelCourtney: Ok lol this is too much :) I'm just gonna do it right and allow exceptions from inside the visitor

pcrosemurgy: Marking as resolved – I was thinking this was publicly called but it's just a Jackson-ism. Thanks for your patience Joel!

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/json/SerializedValueSerializer.java
```diff
@@ -0,0 +1,110 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public final class SerializedValueSerializer extends StdSerializer<SerializedValue> {
```
Twisol, line 12: The purpose of the `merlin.protocol` package is to coordinate communication between mission models and systems that use them (like the simulation engine). It's not clear, but I don't think this class is used to that end.

I'm not fond of giving all models a transitive dependency on Jackson; is it possible to move these serialization classes out of `merlin.protocol`?

JoelCourtney: I'm not sure what you mean by transitive dependency.

In Jackson, the Java class you want to serialize or deserialize defines its own schema. If the JSON schema is different from the Java schema, you usually put annotations on the class that tell Jackson what's different. If the schemas are _too_ different for annotations to handle (which is rare and only happened for SerializedValue), you can write a custom serializer and deserializer. But, the (de)serializer is attached to the class using

```java
@JsonSerialize(using = SerializedValueSerializer.class)
@JsonDeserialize(using = SerializedValueDeserializer.class)
public interface SerializedValue {
    ...
```

The JSON schema is attached to the class. When you want to deserialize it, you use `mapper.readValue(inputString, SerializedValue.class)` just like you would do with a class that doesn't have custom a serializer or deserializer.

edit: I realized that I definitely misinterpreted what you said. One sec

JoelCourtney: Ok I figured out what you said. Yes it is possible to remove the `merlin.protocol` dependency on Jackson, but in my opinion its a worse solution, and the reason is actually because of my original comment.

The serializer and deserializer are attached to the SerializedValue class, so that every other package attempting to communicate with SerializedValues can just call `mapper.readValue(input, SerializedValue.class)`, without even knowing that it has a custom serializer and deserializer. If you move the (de)serializer to another package, they can't be attached to the SerializedValue class, and you'll have to associate them at runtime when you create the object mapper. Which would usually look like this:

```java
private static final ObjectMapper mapper = new ObjectMapper();
static {
    SimpleModule module = new SimpleModule();
    module.addSerializer(SerializedValue.class, new SerializedValueSerializer());
    module.addDeserializer(SerializedValue.class, new SerializedValueDeserializer());
    mapper.registerModule(module);
}
```

And you'll have to remember to do that in every package that creates an ObjectMapper. That's a lot more prone to error, so I'd prefer not to do it.

Twisol: I understand the alternatives regarding making Jackson aware of these types. But, we would be in the same situation if we had a third-party library whose types we needed to serialize, and whose source we could not edit.

We can provide a `MerlinJacksonModule` class in the `merlin.driver` package like the below, and if downstream clients want to use Jackson to serialize Merlin types, they can call `mapper.registerModule(MerlinJacksonModule.get())`. If you forget to do this, you'll find out whether you got it wrong as soon as you try to deserialize something or use the serialized data.

```java
public final class MerlinJacksonModule {
  public static SimpleModule get() {
    final var module = new SimpleModule();
    module.addSerializer(SerializedValue.class, new SerializedValueSerializer());
    module.addDeserializer(SerializedValue.class, new SerializedValueDeserializer());
    // and similar for Duration, maybe
    return module;
  }
}

// elsewhere:
mapper.registerModule(MerlinJacksonModule.get());
```

> And you'll have to remember to do that in every package that creates an ObjectMapper. That's a lot more prone to error, so I'd prefer not to do it.

I'd expect to provide top-level methods for parsing entities of interest, like constraints, which would reduce the need for multiple clients to manage their own Jackson parsers. Jackson should be an implementation detail localized to the few sites that need to serialize or deserialize data.

Twisol:
One of the reasons we developed the current JSON parser combinator system is to define (1) parsing, (2) printing, and (3) schema extraction all at once. Does the Jackson approach let us produce JSON Schema documents describing the expected structure of our serializable entities? We've found this really useful in the past, particularly for the front-end. I'm not sure if @camargo still makes use of this capability?

Twisol:
> * For example, I needed to implement a mutually recursive pair of parsers for the `windows.spans()` and `spans.windows()` operations. I don't think this is possible given the current `recursiveP` pattern, which only supports direct recursion. I didn't have to do anything to accommodate this in Jackson.

If I understand this correctly, an example of mutual recursion exists in `ConstraintsParser`, with `windowsExpressionP` and `andF` mutually recursive, or `violationListExpressionP` and `forEachActivityF`. You can use plain Java methods to capture a rule's dependencies, and close the loop with a `recursiveP` at the point in the grammar where you enter the mutual recursion.

```java
JsonParser<Foo> fooF(JsonParser<Bar> barP) {
  return productP.field("wrap", barP);
}

JsonParser<Bar> barF(JsonParser<Foo> fooP) {
  return chooseP(fooP, literalP("end")); 
}

JsonParser<Bar> barP = recursiveP(selfP -> barF(fooF(selfP)));
```

pcrosemurgy:
> Does the Jackson approach let us produce JSON Schema documents describing the expected structure of our serializable entities? We've found this really useful in the past, particularly for the front-end.

Do you have an example of how/where our current JSON parser combinator system produces JSON Schema docs.? (I'm not skeptical that it does, am just interested in seeing it in action ☺️)

Twisol:
> Do you have an example of how/where our current JSON parser combinator system produces JSON Schema docs.?

I don't think it's hooked up in a live system, but the `JsonParser` interface has a method `JsonObject getSchema()`. I remember one time very distinctly where we locally added a `main` method that called it on something, I think maybe the top-level constraint parser, and airdropped the output to @camargo. We discussed wiring up a route for it at some point, but we never got around to it, since Chris was unblocked for the time being.

pcrosemurgy:
That's awesome! Thanks for the context. May be very helpful again in the future.

Twisol:
Sure. Here's an example and a gist with output: https://gist.github.com/Twisol/7a1d170c4c24cb9c20b834a656c88b8a

```java
public final class ConstraintParsers {
  // ...

  public static void main(final String[] args) {
    System.out.println(constraintP.getSchema());
  }
}
```

I took the liberty of postprocessing it with an online prettifier for human consumption.

camargo:
@Twisol Yeah we used that for a bit with the JSON constraints editor. However since we switched to using the TypeScript-based constraints DSL we no longer use the JSON schema on the front-end.

JoelCourtney:
> Does the Jackson approach let us produce JSON Schema documents describing the expected structure of our serializable entities?

The short answer is yes there are a couple modules that do that. But the officially supported one is somewhat lacking and not planned to continue into Jackson 3.0, and the more complete unofficial one is not something I'm comfortable putting a dependency on.

JoelCourtney:
@Twisol the mutual recursion example worked, thanks! I suspected that it would be only finitely recursive, but it checks out.

JoelCourtney:
I'm closing this just because this not critical and the review process will probably take a while, and in the meantime it will be a pain to to keep this up to date with all of the changes I'm making to the constraints API. In the future when the API's are a little more stable, this won't be hard to re-implement, because it wasn't hard to implement in the first place. So for now every time you have to write a parser and are mildly annoyed, think of Jackson... ;)

PR #268, opened 2022-07-27
[AERIE-1875]  Return Dictionary fields when uploading

goetzrrGit:
* **Tickets addressed:** AERIE-1875
* **Review:** By commit  

## Description
When you upload a command dictionary, it will save the version number, mission name, path , etc. in the database. Instead of making a second fetch call to obtain these fields in the UI, I change the `Upload Command Dictionary` action to return these fields as part of the upload process.

I also added a new field in the table with a trigger for `updated_at`. A mission can upload a dictionary with a minor change that doesn't require a version bump.  I figured we should capture this updated time as well. 

## Verification
Running this QGL mutation to upload a command dictionary you will now receive `command_types_typescript_path` `mission` `version` `created_at``updated_at`

```
mutation Upload(
  $commandDictionaryString: String!
) {
  uploadDictionary(dictionary: $commandDictionaryString) {
    id
    command_types_typescript_path
    mission
    version
    created_at
    updated_at
  }
}
```

## Documentation
I will update the Command Expansion wiki with this change. 

## Future work
None

### DIFF
--- a/command-expansion-server/sql/commanding/tables/command_dictionary.sql
+++ b/command-expansion-server/sql/commanding/tables/command_dictionary.sql
@@ -6,6 +6,7 @@ create table command_dictionary (
   version text not null,
 
   created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
 
   constraint command_dictionary_synthetic_key
       primary key (id),
@@ -25,3 +26,17 @@ comment on column command_dictionary.version is e''
   'A human-meaningful version qualifier.';
 comment on constraint command_dictionary_natural_key on command_dictionary is e''
   'There an only be one command dictionary of a given version for a given mission.';
+
+create function command_dictionary_set_updated_at()
+returns trigger
+security definer
+language plpgsql as $$begin
+  new.updated_at = now();
+  return new;
+end$$;
+
+create trigger trigger_command_dictionary_set_updated_at
+  before
+    update on command_dictionary
+  for each row
+  execute function command_dictionary_set_updated_at();
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -122,7 +122,7 @@ app.post('/put-dictionary', async (req, res, next) => {
     values ($1, $2, $3)
     on conflict (mission, version) do update
       set command_types_typescript_path = $1
-    returning id;
+    returning id, command_types_typescript_path, mission, version, created_at, updated_at;
   `;
 
   const { rows } = await db.query(sqlExpression, [
@@ -134,8 +134,8 @@ app.post('/put-dictionary', async (req, res, next) => {
   if (rows.length < 1) {
     throw new Error(`POST /dictionary: No command dictionary was updated in the database`);
   }
-  const id = rows[0].id;
-  res.status(200).json({ id });
+  const [row] = rows;
+  res.status(200).json(row);
   return next();
 });
 
@@ -282,10 +282,12 @@ app.post('/get-command-typescript', async (req, res, next) => {
 
     res.status(200).json({
       status: Status.SUCCESS,
-      typescriptFiles: [{
-        filePath: 'command-types.ts',
-        content: commandTypescript,
-      }],
+      typescriptFiles: [
+        {
+          filePath: 'command-types.ts',
+          content: commandTypescript,
+        },
+      ],
       reason: null,
     });
   } catch (e) {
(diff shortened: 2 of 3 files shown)


command-expansion-server/sql/commanding/tables/command_dictionary.sql
```diff
@@ -25,3 +26,17 @@ comment on column command_dictionary.version is e''
   'A human-meaningful version qualifier.';
 comment on constraint command_dictionary_natural_key on command_dictionary is e''
   'There an only be one command dictionary of a given version for a given mission.';
+
+CREATE FUNCTION trigger_update_at_timestamp()
```
camargo: Can we use lowercase for SQL keywords to stay consistent with the other files?

command-expansion-server/src/app.ts
```diff
@@ -135,7 +135,13 @@ app.post('/put-dictionary', async (req, res, next) => {
     throw new Error(`POST /dictionary: No command dictionary was updated in the database`);
   }
   const id = rows[0].id;
-  res.status(200).json({ id });
+  const command_types_typescript_path = rows[0].command_types_typescript_path;
```
camargo: If you don't want to keep accessing `rows[0]` here you make this a one-liner using destructuring assignment:

```ts
const [{ command_types_typescript_path, created_at, id, mission, updated_at, version }] = rows;
```

camargo: Similarly if the first `row` has all the properties we need you could do:

```ts
const [row] = rows;
res.status(200).json(row);
```

command-expansion-server/sql/commanding/tables/command_dictionary.sql
goetzrrGit: Sure, no problem

command-expansion-server/src/app.ts
goetzrrGit: Thanks for the suggestion. I will make that change.

command-expansion-server/sql/commanding/tables/command_dictionary.sql
```diff
@@ -25,3 +26,17 @@ comment on column command_dictionary.version is e''
   'A human-meaningful version qualifier.';
 comment on constraint command_dictionary_natural_key on command_dictionary is e''
   'There an only be one command dictionary of a given version for a given mission.';
+
+create function trigger_update_at_timestamp()
```
mattdailis: I think Dylan had a good naming scheme for these in his activity metadata PRs - I think this would be `trigger_command_dictionary_set_updated_at`. I think the key there is to include the table name, and the column name. (same goes for the name of the function)

command-expansion-server/src/app.ts
```diff
@@ -282,10 +282,12 @@ app.post('/get-command-typescript', async (req, res, next) => {
 
     res.status(200).json({
       status: Status.SUCCESS,
-      typescriptFiles: [{
-        filePath: 'command-types.ts',
-        content: commandTypescript,
-      }],
+      typescriptFiles: [
```
mattdailis, line 285: Is this just a formatting change?

mattdailis (changes requested):

command-expansion-server/src/app.ts
goetzrrGit: Yes it is, I was lazy and didn't want to split this change into a reformat commit. :P

command-expansion-server/sql/commanding/tables/command_dictionary.sql
goetzrrGit: Good suggestion, makes it clearer on what is happening

command-expansion-server/sql/commanding/tables/command_dictionary.sql
```diff
@@ -25,3 +26,17 @@ comment on column command_dictionary.version is e''
   'A human-meaningful version qualifier.';
 comment on constraint command_dictionary_natural_key on command_dictionary is e''
   'There an only be one command dictionary of a given version for a given mission.';
+
+create function trigger_command_dictionary_set_updated_at()
```
mattdailis: Thanks for making the change! Though I think the trigger prefix should be on the trigger, not on the function.

goetzrrGit: That is what I get for context switch back to AERIE to make the change really quick. Thanks for catching that it should be updated now.

mattdailis (approved):

camargo (approved):
LGTM

PR #269, opened 2022-07-27
build: update dev docker-compose

camargo:
## Description

- Make auth opt-in for the UI and Gateway
- Mirrors what we are doing for the production docker-compose since we are on public GH now


goetzrrGit (approved):
This is what I do locally all the time :P 

LGTM

mattdailis (approved):

PR #274, opened 2022-07-28
[AERIE-1974] Remove nested ternary type checks in constraints eDSL

JoelCourtney:
* **Tickets addressed:** AERIE-1974
* **Review:** By commit  

## Description
Credit where its due, this is actually mostly @dyst5422 's wizardry, I just messed with the code generation.

Large mission models like Clipper would encounter a stack-overflow-like error when typechecking constraints because of the massive nested ternary expressions involved. This PR refactors that to avoid those types. As an added benefit, the `ActivityInstance` class now doesn't need to be generated.

## Verification
The code gen test was the only one that need to be updated.

## Documentation
None needed

## Future work
None

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
@@ -30,12 +30,16 @@ public final class TypescriptCodeGenerationService {
     }
     result.add("}");
 
-    result.add("export type ResourceName = " + String.join(" | ", resources.keySet().stream().map($ -> "\"" + $ + "\"").toList()) + ";");
-    result.add("export type DiscreteResourceSchema<R extends ResourceName> =");
+    result.add("export type Resource = {");
     for (var resource: resources.keySet()) {
-      result.add(indent("R extends \"" + resource + "\" ? " + valueSchemaToTypescript(resources.get(resource)) + " :"));
+      result.add(indent("\"" + resource + "\": " + valueSchemaToTypescript(resources.get(resource)) + ","));
     }
-    result.add(indent("never;"));
+    result.add("};");
+
+    result.add("export type ResourceName = " + String.join(
+        " | ",
+        resources.keySet().stream().map(key -> ("\"" + key + "\"")).toList()
+    ) + ";");
 
     result.add("export type RealResourceName = " + String.join(
         " | ",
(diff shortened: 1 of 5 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
```diff
@@ -30,12 +30,16 @@ public static String generateTypescriptTypes(
...
 
-    result.add("export type ResourceName = " + String.join(" | ", resources.keySet().stream().map($ -> "\"" + $ + "\"").toList()) + ";");
-    result.add("export type DiscreteResourceSchema<R extends ResourceName> =");
+    result.add("export type Resource = {");
     for (var resource: resources.keySet()) {
-      result.add(indent("R extends \"" + resource + "\" ? " + valueSchemaToTypescript(resources.get(resource)) + " :"));
+      result.add(indent("\"" + resource + "\": " + valueSchemaToTypescript(resources.get(resource)) + ","));
     }
-    result.add(indent("never;"));
+    result.add("}");
```
dyst5422: lets add a semicolon here - I know they aren't strictly necessary, but it's best practice because of some weird stuff that _can_ happen without them.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
```diff
@@ -30,12 +30,16 @@ public static String generateTypescriptTypes(
...
+
+    result.add("export type ResourceName = " + String.join(
```
dyst5422, line 39: I don't remember the answer to this. Did we see an explicit benefit from enumerating them vs using a `keyof Resource` invocation?

dyst5422 (approved):

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
JoelCourtney: Yes, now the error message if you mistype the name is 'not assignable to `ResourceName`' instead of 'not assignable to `keyof Resource`', which is less likely to cause confusion

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java
JoelCourtney: done

mattdailis (approved):

mattdailis:
Just tested this with the big mission model, and it seems to be working 🤞 I say ship it :shipit:

PR #277, opened 2022-08-04
[AERIE-1497] Raise negative plan duration PG exception

pcrosemurgy:
* **Tickets addressed:** AERIE-1497
* **Review:** By commit  

## Description

Changes our plan duration PG constraint check to a conditional trigger with a custom exception.
Previously negative durations were reported (through Hasura) as:
```json
            "error": {
                "path": "$.selectionSet.insert_plan_one.args.object",
                "error": "Check constraint violation. new row for relation \"plan\" violates check constraint \"plan_duration_is_nonnegative\"",
                "code": "permission-error"
            },
```
Now they are reported as:
```json
            "error": {
                "internal": {
                    "statement": "WITH \"plan__mutation_result_alias\" AS (INSERT INTO \"public\".\"plan\" ( \"model_id\", \"name\", \"start_time\", \"duration\" ) VALUES (('1')::integer, ('Bad Plan')::text, ('2022-216T00:00:00')::timestamptz, ('-86400 seconds 0 milliseconds')::interval)  RETURNING * , ('true')::boolean AS \"check__constraint\"), \"plan__all_columns_alias\" AS (SELECT  \"id\" , \"revision\" , \"name\" , \"model_id\" , \"duration\" , \"start_time\"  FROM \"plan__mutation_result_alias\"      ) SELECT  (SELECT  coalesce((json_agg(\"root\" )->0), 'null' ) AS \"root\" FROM  (SELECT  row_to_json((SELECT  \"_1_e\"  FROM  (SELECT  \"_0_root.base\".\"id\" AS \"id\", \"_0_root.base\".\"revision\" AS \"revision\"       ) AS \"_1_e\"      ) ) AS \"root\" FROM  (SELECT  *  FROM \"plan__all_columns_alias\" WHERE ('true')     ) AS \"_0_root.base\"      ) AS \"_2_root\"      ) , (SELECT  coalesce(bool_and(\"check__constraint\" ), 'true' )  FROM \"plan__mutation_result_alias\"      )        ",
                    "prepared": false,
                    "error": {
                        "exec_status": "FatalError",
                        "hint": null,
                        "message": "invalid negative duration: -24:00:00",
                        "status_code": "P0001",
                        "description": null
                    },
                    "arguments": []
                },
                "path": "$.selectionSet.insert_plan_one.args.object",
                "error": "database query error",
                "code": "unexpected"
            },
```

`error.internal.error.message` should always be there when `error.code == "unexpected"`. From testing locally this internal error object is always present even if `HASURA_GRAPHQL_DEV_MODE` is set to false.

## Verification

Tested locally. See above section for example responses.

## Documentation

None.

## Future work

- The UI should check for `error.code == "unexpected"` and grab `error.internal.error.message` if true.
  - In fact, this UI change should be merged within the same release as this PR to ensure that our UI properly reports errors and doesn't keep grabbing `error.error` (which will now contain `"database query error"` for a negative duration as shown above). I'm happy to make a PR for the UI to support this. :)
- I believe the general rule of thumb for deciding between `constraint <constraint_name> check (<bool_expression>)` vs. a trigger that conditionally `raise`s is if we expect the constraint violation error to be seen by an end user then a more readable PG exception should be raised. We may want to change other instances of PG `check` to use a trigger with a friendlier error message.

### DIFF
--- a/merlin-server/sql/merlin/tables/plan.sql
+++ b/merlin-server/sql/merlin/tables/plan.sql
@@ -17,9 +17,7 @@ create table plan (
     foreign key (model_id)
     references mission_model
     on update cascade
-    on delete set null,
-  constraint plan_duration_is_nonnegative
-    check (duration >= '0')
+    on delete set null
 );
 
 create index plan_model_id_index on plan (model_id);
@@ -61,3 +59,16 @@ after update on plan
 for each row
 when (pg_trigger_depth() < 1)
 execute function increment_revision_on_update_plan();
+
+create function raise_duration_is_negative()
+returns trigger
+security definer
+language plpgsql as $$begin
+  raise exception 'invalid plan duration, expected nonnegative duration but found: %', new.duration;
+end$$;
+
+create trigger check_plan_duration_is_nonnegative_trigger
+before insert or update on plan
+for each row
+when (new.duration < '0')
+execute function raise_duration_is_negative();


merlin-server/sql/merlin/tables/plan.sql
```diff
@@ -61,3 +59,16 @@ after update on plan
 for each row
 when (pg_trigger_depth() < 1)
 execute function increment_revision_on_update_plan();
+
+create function raise_duration_is_negative()
+returns trigger
+security definer
+language plpgsql as $$begin
+  raise exception 'invalid negative duration: %', new.duration;
```
mattdailis: Would the word "plan" in here somewhere help avoid ambiguity, or do you figure it's always obvious what duration is being referred to?

mattdailis (approved):

merlin-server/sql/merlin/tables/plan.sql
pcrosemurgy: For sure! Just pushed a fix to change this to:
```sql
  raise exception 'invalid plan duration, expected nonnegative duration but found: %', new.duration;
```

patkenneally (approved):

merlin-server/sql/merlin/tables/plan.sql
pcrosemurgy: Resolving but please let me know if you'd like it reworded at all, happy to make a hotfix PR!

PR #278, opened 2022-08-04
[AERIE-1935] Expose plan start time to mission model

pcrosemurgy:
* **Tickets addressed:** AERIE-1935
* **Review:** By commit  

## Description

Mission modelers currently don't have access to a plan's start time.
With this PR a plan start time `Instant` parameter can optionally be declared in the top-level mission model ("`Mission.java`") constructor.

If a top-level mission model defines a constructor with `Instant` as the 2nd argument then the sim. start time will be passed to the mission model upon instantiation. If the `Instant` parameter (or `Registrar` parameter and optionally a "`Configuration`" parameter) is declared out-of-order then an actionable error message is now presented at compile-time. See the section below for examples.

More concretely, the following top-level mission model constructor parameters are allowed (lifted from `MissionModelParser.java`):
```java
    // - (Registrar, Instant, Configuration)
    // - (Registrar, Configuration)
    // - (Registrar, Instant)
    // - (Registrar)
```
As seen above, both sim. start time and mission model configuration parameters are optional. However, if both are defined "`Configuration`" must be declared last. This ensures backwards compatibility with existing mission models while allowing future mission models to make use of this `Instant`.

See the future work section for a discussion on alternative approaches for this PR.

## Verification
See `foo-missionmodel`'s [`MissionConfigurationTest`](https://github.com/NASA-AMMOS/aerie/pull/278/files#diff-712240ebbae24ed064a056ff8435c69eead17d86455f9851be99df69b161ba6b).

<details>
<summary>Error Handling</summary>

When the following `Mission` constructor parameters are declared in an [allowable](https://github.com/NASA-AMMOS/aerie/commit/5d210cbfca8492d3aa08f9a9c08c98d19becfaa5#diff-39f6ee1b28c111fa143558449ea4f23227bea8393e1d1274503876dfec5d2f1cR91-R95) order no compile-time error is thrown:
```java
public Mission(final Registrar registrar, final Instant planStart, final Configuration config) {
  ...
}
```

However, when `Instant` (or any of the above parameters) are declared out of expected order a custom error message will appear:
```
.../foomissionmodel/package-info.java:21: error: The top-level model's constructor is expected to accept [gov.nasa.jpl.aerie.merlin.framework.Registrar, java.time.Instant, gov.nasa.jpl.aerie.foomissionmodel.Configuration], found: [gov.nasa.jpl.aerie.merlin.framework.Registrar, gov.nasa.jpl.aerie.foomissionmodel.Configuration, java.time.Instant]
```
In this example, `Mission`'s constructor was declared with a plan start `Instant` in the wrong expected order:
```java
public Mission(final Registrar registrar, final Configuration config, final Instant planStart) {
  ...
}
```

</details>

## Documentation

- See `deployment/Environment.md` and `deployment/README.md` for information on `UNTRUE_PLAN_START`.
- See https://github.com/NASA-AMMOS/aerie/wiki/Developing-a-Mission-Model#mission-model-class
- See https://github.com/NASA-AMMOS/aerie/wiki/Configuring-a-Mission-Model#mission-model

## Future work

Initially a `@PlanStart` annotation was considered but in prototyping this implementation it was found that this change would require plan information to be passed in at our GQL layer (to allow any configured context to have plan information).
This would break existing GQL queries and more tightly couple our plan and model "domains".

See [Slack thread](https://jpl.slack.com/archives/GJ07FGRPH/p1661280990248519) for more context and discussion on this.

### DIFF
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/Mission.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/Mission.java
@@ -25,6 +25,7 @@ public final class Mission {
   // Need to generalize RealDynamics to nonlinear polynomials.
 
   public final Register<Double> foo = Register.forImmutable(0.0);
+  public final Register<Boolean> startingAfterUnixEpoch;
   public final Accumulator data = new Accumulator();
   public final Accumulator source = new Accumulator(100.0, 1.0);
   public final Accumulator sink;
@@ -37,9 +38,10 @@ public final class Mission {
   public final Clock utcClock = new Clock(Instant.parse("2023-08-18T00:00:00.00Z"));
   private final Registrar cachedRegistrar; // Normally bad practice, only stored to demonstrate built/unbuilt check
 
-  public Mission(final Registrar registrar, final Configuration config) {
+  public Mission(final Registrar registrar, final Instant planStart, final Configuration config) {
     this.cachedRegistrar = registrar;
 
+    this.startingAfterUnixEpoch = Register.forImmutable(planStart.compareTo(Instant.EPOCH) > 0);
     this.sink = new Accumulator(0.0, config.sinkRate);
 
     spawn(this::test);
@@ -52,6 +54,7 @@ public final class Mission {
     //   (This binding layer should also be the one responsible for feeding in constructor parameters.)
     registrar.discrete("/foo", this.foo, new DoubleValueMapper());
     registrar.discrete("/foo/conflicted", this.foo::isConflicted, new BooleanValueMapper());
+    registrar.discrete("/foo/starting_after_unix_epoch", startingAfterUnixEpoch, new BooleanValueMapper());
     registrar.real("/batterySoC", this.source.minus(this.sink));
     registrar.real("/data", this.data);
     registrar.real("/data/rate", this.data.rate);
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
@@ -18,6 +18,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.file.Path;
+import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
@@ -189,7 +190,7 @@ public final class LocalMissionModelService implements MissionModelService {
     }
 
     // TODO: [AERIE-1516] Teardown the mission model after use to release any system resources (e.g. threads).
-    return loadConfiguredMissionModel(message.missionModelId(), SerializedValue.of(config))
+    return loadConfiguredMissionModel(message.missionModelId(), message.startTime(), SerializedValue.of(config))
         .simulate(message.activityInstances(), message.samplingDuration(), message.startTime());
   }
 
@@ -246,7 +247,7 @@ public final class LocalMissionModelService implements MissionModelService {
   private MissionModelFacade loadConfiguredMissionModel(final String missionModelId)
   throws NoSuchMissionModelException, MissionModelLoadException
   {
-    return loadConfiguredMissionModel(missionModelId, SerializedValue.of(Map.of()));
+    return loadConfiguredMissionModel(missionModelId, Instant.EPOCH, SerializedValue.of(Map.of()));
   }
 
   /**
@@ -259,13 +260,20 @@ public final class LocalMissionModelService implements MissionModelService {
    * it contains may not abide by the expected contract at load time.
    * @throws NoSuchMissionModelException If no mission model is known by the given ID.
    */
-  private MissionModelFacade loadConfiguredMissionModel(final String missionModelId, final SerializedValue configuration)
+  private MissionModelFacade loadConfiguredMissionModel(
+      final String missionModelId,
+      final Instant planStart,
+      final SerializedValue configuration)
   throws NoSuchMissionModelException, MissionModelLoadException
   {
     try {
       final var missionModelJar = this.missionModelRepository.getMissionModel(missionModelId);
-      final var missionModel =
-          MissionModelLoader.loadMissionModel(configuration, missionModelDataPath.resolve(missionModelJar.path), missionModelJar.name, missionModelJar.version);
+      final var missionModel = MissionModelLoader.loadMissionModel(
+          planStart,
+          configuration,
+          missionModelDataPath.resolve(missionModelJar.path),
+          missionModelJar.name,
+          missionModelJar.version);
       return new MissionModelFacade(missionModel);
     } catch (final MissionModelRepository.NoSuchMissionModelException ex) {
       throw new NoSuchMissionModelException(missionModelId, ex);
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
@@ -34,13 +34,16 @@ public final class LocalMissionModelService implements MissionModelService {
 
   private final Path missionModelDataPath;
   private final MissionModelRepository missionModelRepository;
+  private final Instant untruePlanStart;
 
   public LocalMissionModelService(
       final Path missionModelDataPath,
-      final MissionModelRepository missionModelRepository
+      final MissionModelRepository missionModelRepository,
+      final Instant untruePlanStart
   ) {
     this.missionModelDataPath = missionModelDataPath;
     this.missionModelRepository = missionModelRepository;
+    this.untruePlanStart = untruePlanStart;
   }
 
   @Override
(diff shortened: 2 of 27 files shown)


mattdailis:
From reading the PR description, it sounds like the annotation approach would be preferable, but we're concerned that having different annotations for records vs regular classes would cause confusion. Am I reading that correctly?

Would it help to have two similarly named annotations in different names paces, or would that make it even more confusing?

Twisol:
> * I don't believe it's possible to define a single annotation for use as both a class member variable and constructor parameter but may be wrong.

I'd expect this to be possible. Annotations don't do anything on their own -- it's up to our processor how to interpret them.

If you're concerned about how to set `java.lang.annotation.Target` to specify where the annotation is allowed, `@Target` can accept a list of locations:

```java
@Target({ElementType.RECORD_COMPONENT, ElementType.FIELD})
@interface SimulationStart {
}
```

(Also, can you remind me what the difference is between the simulation start time and the plan start time?)

Twisol:
> * I'm not 100% sure the above `record` annotation example is possible (see https://stackoverflow.com/questions/67168624/constructor-annotation-on-java-records).

That question is asking about propagating annotations on `record`s to their synthesized constructors. I don't think this matters to us -- if we know it's a record, we can look at the record's components directly, which IIRC is how the other record-based analyses in our processor work.

pcrosemurgy:
>Also, can you remind me what the difference is between the simulation start time and the plan start time?

Sim. start time is passed through from a "create sim. message". Currently I believe they should always be the same but we're carefully making sure to use sim. start time here since I believe in the future we may have sim. and plan start time differ (for example when re-running part of a sim. I believe).

pcrosemurgy:
@Twisol 
>If you're concerned about how to set java.lang.annotation.Target to specify where the annotation is allowed, @ Target can accept a list of locations:

Thanks for the help! I'll rework this PR to use a single annotation with multiple targets. Having this annotation apply to only config. types will complicate our mapper code-gen. but an annotation does seem like the correct approach. :)

Twisol:
> > Also, can you remind me what the difference is between the simulation start time and the plan start time?
> 
> Sim. start time is passed through from a "create sim. message". Currently I believe they should always be the same but we're carefully making sure to use sim. start time here since I believe in the future we may have sim. and plan start time differ (for example when re-running part of a sim. I believe).

Hmm... well, why should we provide different input data to a simulation if we're just re-running a part of it? Presumably we're re-running it because of some *other* change, but we don't want to confound that with additional changes. Or we're running it for some *additional* time, and so no input should have changed, right?

It seems to me that plan start time is the relevant domain notion, no? Isn't it often meant as a base time used for SPICE calls and whatnot? Surely our trajectories shouldn't shift around our activities just because we've asked to run a certain subset of the plan.

pcrosemurgy:
If we're re-running a part of a plan then the sim. start time should be some time greater than the plan start time.

Twisol:
> If we're re-running a part of a plan then the sim. start time should be some time greater than the plan start time.

Yes, and isn't that bad? Doesn't that mean we could get wildly different results just because we're running a subset? That seems undesirable.

Twisol:
I edited this into my preceding comment, but I'm copying it into a new one so that it doesn't hide from email notifs:

> Isn't [plan start time] often meant as a base time used for SPICE calls and whatnot? Surely our trajectories shouldn't shift around our activities just because we've asked to run a certain subset of the plan.

pcrosemurgy:
Do you have some time for a quick call (via Slack)? Running a different subset of a plan inherently should result in (wildly) different results, so that seems appropriate to me.

Twisol:
Yes, sure. I've become freshly confused about the use-cases for providing this data to the model.

pcrosemurgy:
We've decided to use plan start time as using sim. start time should be an engine concern – i.e. we don't want our mission model to behave differently during a re-sim. (if/when re-sim. is supported).

pcrosemurgy:
Latest changes:
- Renamed `simuationStart` to `planStart`.
- Rebased with `develop`.

@mattdailis @Twisol this PR is now ready for re-review :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -246,7 +247,7 @@ private MissionModelFacade.Unconfigured<?> loadUnconfiguredMissionModel(final St
   private MissionModelFacade loadConfiguredMissionModel(final String missionModelId)
   throws NoSuchMissionModelException, MissionModelLoadException
   {
-    return loadConfiguredMissionModel(missionModelId, SerializedValue.of(Map.of()));
+    return loadConfiguredMissionModel(missionModelId, Instant.EPOCH, SerializedValue.of(Map.of()));
```
mattdailis: 🔶 I worry about this "bogus" value. I'm imagining a mission model using it to instantiate spice-related values, and blowing up because the kernels don't cover January 1st, 1970.

pcrosemurgy: That's a really good point. This is the one area I'm the most unsure about. Let me think about this for another minute.

pcrosemurgy: The only 2 alternatives I can think of are:
1. Have the mission model define a sane default `Instant` plan start time (through an annotation) denoting a time that should always "work" with the mission model.
2. Require plan ID whenever loading a "configured mission model". This change would bubble up to our GQL layer.

mattdailis: I agree with your assessment. I'm not sure which of the two options is preferable.

pcrosemurgy: I think option 2 above is sort of unavoidable, at least with how we have our mission model loading implemented.

pcrosemurgy: Another solution would be to represent "bogus" at the type-level. For example, by using an `Optional<Instant>` instead of an `Instant`. This implementation detail would leak to the mission modeler since they would conditionally use the time depending on if it's empty or not.

If an `Optional<Instant>` is frowned upon we could use a `sealed` type called `PlanStart` which could be `Empty` or `Provided` instead.

pcrosemurgy: Basically with this approach the mission modeler is aware of when a bogus start time is being used within the top-level mission model constructor. Not the best solution but may be good enough until we figure out how to refactor our mission model loading (which is a sizable amount of work).

pcrosemurgy: I'm actually pretty convinced this is a decent solution until we redesign our mission model loading.

With this PR a `PlanStart` parameter to top-level mission model c-tor is completely optional, so no backwards compat. is being broken and it's not a big deal if mission modelers have to be aware of when `PlanStart` is `instanceof PlanStart.Empty` for a few releases.

examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/Mission.java
```diff
@@ -37,9 +38,10 @@ public final class Mission {
   public final Clock utcClock = new Clock(Instant.parse("2023-08-18T00:00:00.00Z"));
   private final Registrar cachedRegistrar; // Normally bad practice, only stored to demonstrate built/unbuilt check
 
-  public Mission(final Registrar registrar, final Configuration config) {
+  public Mission(final Registrar registrar, Instant planStart, final Configuration config) {
```
mattdailis: For the sake of consistency, we can tack a `final` on here.

pcrosemurgy: Good catch! Thanks

pcrosemurgy: Done :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
pcrosemurgy: Pushed change to add an `UNTRUE_PLAN_START` env. var. This is threaded from the app driver to the mission model service.

pcrosemurgy:
Everything is now stable. Done making changes :)

goetzrrGit:
Just to understand it better the mission model will use the `UNTRUE_PLAN_START` by default unless specified by the user in their queries?

pcrosemurgy:
> Just to understand it better the mission model will use the `UNTRUE_PLAN_START` by default unless specified by the user in their queries?

The mission model will actually use `UNTRUE_PLAN_START` for every mission model instantiation besides the instantiation done for a simulation run. So any of these:
Screen Shot 2022-08-25 at 11 37 42 AM

pcrosemurgy:
When running a simulation the actual plan start time will be used :)

goetzrrGit:
Thanks for the info!! I also just read through the slack thread I missed between you, Matt, and Jonathan, which gave me more context.

pcrosemurgy:
Thanks Ryan, appreciate you spending time on this w/ us!

goetzrrGit (approved):
Not much I would improve upon. Just make sure you create another ticket for the mission model redesign.

mattdailis (approved):
Looks good to me. If you haven't already, let's make a ticket to come up with the longer term solution.

EDIT: Whoops, just saw that Ryan said the same thing

PR #279, opened 2022-08-05
[AERIE-1696]  Move secrets to enviornment variables

Mythicaeda:
* **Tickets addressed:** AERIE-1696
* **Review:** By commit  

## Description

The goal was to remove secrets (ie usernames/passwords) from files being committed to GitHub, with a focus on Docker. This was done by moving them to .env files. Combing the code base for secrets stored in other locations, but I think we got them all.

## Verification

Verified by building Aerie through Docker using the .env and checking that the end image does not contain secrets.

## Documentation

Updated Enviornment.md to no longer contain secrets. Updated both READMEs to tell users and developers that they need to set secrets in an .env if they want to use the default docker-compose.yml. Updated .gitignore to not automatically track any new .env files.

## Future work

None.


.env.template
```diff
@@ -0,0 +1,6 @@
+MERLIN_USERNAME=
```
camargo: Can we prefix `MERLIN_USERNAME` and `MERLIN_PASSWORD` with `AERIE_` instead of `MERLIN_`? Merlin is a specific service and the since this is a system-wide default I think the name should reflect that.

camargo:
We will need to come up with a way to generate an `.env` for the e2e tests in CI. These docs might be helpful for this: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#creating-an-environment

camargo:
The UI e2e test docker-compose will also need to be updated and a .env must be generated from secrets: https://github.com/NASA-AMMOS/aerie-ui/blob/develop/docker-compose-test.yml

We will also need to update UI deployment docs.

deployment/README.md
```diff
@@ -17,7 +17,7 @@ Before you can deploy Aerie, you must install and configure the following produc
 
 ## Environment Variables
 
-Each container has environment variables that can be used to fine-tune your deployment. See the [environment variable documentation](./Environment.md) for the complete set of variables. See the example [docker-compose.yml](./docker-compose.yml) file for examples on how to set the environment variables. **We highly recommend changing the default environment variable passwords before deploying Aerie**.
```
camargo, line 20: Since this step will be required before a deployment will work we should probably add more specific instructions here for how to do it. In general I'm a little worried that this will make it harder for first-time users to deploy Aerie. CC @bashbash

.env.template
Mythicaeda: I've done that change! Feel free to resolve this conversation if you're happy with it.

deployment/README.md
Mythicaeda: I've modified the README to make the instructions more clear and to make the setting up the environment step more obvious.

camargo:
This is looking great! 
~~Can you also update the [building.md](https://github.com/NASA-AMMOS/aerie/blob/develop/docs/building.md) file with instructions on how to set the .env.template for development?~~
Never mind looks like you did this!

.env.template
```diff
@@ -0,0 +1,6 @@
+AERIE_USERNAME=
+AERIE_PASSWORD=
+
+POSTGRES_DB=
+POSTGRES_USERNAME=
```
camargo: I'm wondering if we should call this `POSTGRES_USER` instead of `POSTGRES_USERNAME` to align with the same env var `postgres` actually uses?

Mythicaeda: I initially set it to `USERNAME` instead of `USER` so it was more obvious that this was a second, separate set of credentials, completely independent from the `AERIE` credentials. If you feel like this won't be an issue then I've got no problems changing it, although if you change it in the Github environment let me know so I can modify `test.yml` accordingly.

camargo: Is the assumption that we will always pass the `POSTGRES_USERNAME` env to the `POSTGRES_USER` env in the `postgres` container?

Mythicaeda: Sort of. `POSTGRES_USERNAME` is only used in the `postgres` container's `POSTGRES_USER` env. However, `POSTGRES_USER` is also an env var used by the `aerie-gateway` container, this time with a value of `AERIE_USERNAME`.

camargo: Ah yeah that is a little confusing. I think that is more of an issue with the name of the gateway env var tho. We should probably rename at some point. 
I think since `POSTGRES_USERNAME` in the `.env` is really meant to define the `POSTGRES_USER` for the `postgres` container we should probably just call it `POSTGRES_USER` to reflect that.

Mythicaeda: Will do then. Again, just to clarify, will you be updating the name in e2e-test as well?

camargo: Yup will do it now!

docs/building.md
```diff
@@ -42,6 +42,7 @@ The `merlin` container is sourced by building the container anew using the build
 process. This is evidenced by the `build:` directive in the `merlin`
 section in docker-compose.yml.
 
+Prior to building docker-compose.yml, it is necessary to create a .env file using the template found in .env.template in the Aerie root directory.
```
pcrosemurgy, line 45: So a `.env` file in a working directory effectively sets environment variables for that working directory? The only mention of Unix `.env` files I can find are ones that reside in `$HOME/.env`, so I believe they fairly global environment variables (per-user).

.env.template
Mythicaeda: Done!

docs/building.md
pcrosemurgy: From the [AERIE-1696](https://jira.jpl.nasa.gov/browse/AERIE-1696) ticket I believe we want to [use secrets in compose](https://docs.docker.com/engine/swarm/secrets/#use-secrets-in-compose) instead of environment variables. From the previous link:
>Docker secrets do not set environment variables directly. This was a conscious decision, because environment variables can unintentionally be leaked between containers (for instance, if you use `--link`).

camargo: I think we should move this PR forward. And look at maybe using secrets later.

pcrosemurgy: Okay! In that case I'm still unsure if we're going about this the right way (see 1st comment in this chain).

Mythicaeda: When you run a `docker compose`, Docker looks for a .env in the current working directory and loads it into Docker as though you had used the -e tag, as shown by the documentation [here](https://docs.docker.com/compose/environment-variables/):
> If you have multiple environment variables, you can substitute them by adding them to a default environment variable file named .env or by providing a path to your environment variables file using the --env-file command line option.

If the .env was added to the machine's root directory, then they would appear in your global environment, but because it is in a subdirectory, unless explicitly told to nothing in it will be `export`ed. 

While using secrets in compose could also work, we chose to use environment variables instead for compatibility with Amazon's ECS system, [which prefers environment variables](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-data-security-container-task/):
> As a security best practice, pass sensitive information to containers as environment variables.

pcrosemurgy: To use Docker secrets it looks like we just need `.txt` files laying around:
```yaml
secrets:
   db_password:
     file: db_password.txt
   db_root_password:
     file: db_root_password.txt
```

camargo: It's a Docker compose thing: https://docs.docker.com/compose/environment-variables/#the-env-file

pcrosemurgy: Ohh gotcha! Thanks for the explanation @Mythicaeda :) That makes a lot of sense.

camargo: This also works nicely with GH environment secrets

deployment/.env
```diff
@@ -1,2 +1,10 @@
...
 REPOSITORY_DOCKER_URL=ghcr.io/nasa-ammos
+DOCKER_TAG=latest
+
+# Provide Usernames and Passwords Below
+AERIE_USERNAME=
+AERIE_PASSWORD=
+
+POSTGRES_DB=
+POSTGRES_USER=
+POSTGRES_PASSWORD=
```
pcrosemurgy, line 10: Sorry, I'm kind of confused about this file. It's checked into git, yet these fields have to be modified per-user? In that case, shouldn't this not be checked in?

docs/building.md
Mythicaeda: No problem @pcrosemurgy!

deployment/.env
pcrosemurgy: Or maybe we should change this to `deployment/.env.template` and suggest users use a similar pattern as the top-level `.env.template`.

camargo: We need it because it's the default .env file for the deployment. It comes with the deployment .zip as an example of what env vars are needed to deploy.

pcrosemurgy: Ohh I see! Thanks again, this makes sense!

pcrosemurgy (approved):
Looks good! Thanks for walking me through some of those changes.

camargo (approved):
LGTM!

PR #282, opened 2022-08-09
[AERIE-1940]  scheduler populate source metadata field

dyst5422:
* **Tickets addressed:** AERIE-1940
* **Review:** By commit  

## Description

Adjusted the GraphQL queries to use parameterized queries. Built a map of goals to activities, used that map to store the goal id for each activity that had a source goal.

## Verification

Tested locally.

## Documentation

None

## Future work

There are a bunch more graphql queries that could really use some cleaning up.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
@@ -18,6 +18,7 @@ import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchPlanException;
 import gov.nasa.jpl.aerie.scheduler.server.http.InvalidEntityException;
 import gov.nasa.jpl.aerie.scheduler.server.http.InvalidJsonException;
 import gov.nasa.jpl.aerie.scheduler.server.http.SerializedValueJsonParser;
+import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
 import gov.nasa.jpl.aerie.scheduler.server.models.MerlinActivityInstance;
 import gov.nasa.jpl.aerie.scheduler.server.models.MerlinPlan;
 import gov.nasa.jpl.aerie.scheduler.server.models.MissionModelId;
@@ -285,7 +286,8 @@ public record GraphQLMerlinService(URI merlinGraphqlURI) implements PlanService.
   @Override
   public Pair<PlanId, Map<ActivityInstance, ActivityInstanceId>> createNewPlanWithActivities(
       final PlanMetadata planMetadata,
-      final Plan plan
+      final Plan plan,
+      final Map<ActivityInstance, GoalId> activityToGoalId
   )
   throws IOException, NoSuchPlanException, PlanServiceException
   {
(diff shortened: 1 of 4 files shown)


dyst5422:
@mattdailis and @adrienmaillard I think you two are most familiar with this section of the codebase and so you both drew the lucky straws of reviewing this PR.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -106,6 +108,41 @@ private Optional<JsonObject> postRequest(final String gqlStr) throws IOException
...
+      if (httpResp.statusCode() != 200) {
+        //TODO: how severely to error out if aerie cannot be reached or has a 500 error or json is garbled etc etc?
+        return Optional.empty();
+      }
+      final var respBody = Json.createReader(httpResp.body()).readObject();
+      if (respBody.containsKey("errors")) {
+        throw new PlanServiceException(respBody.toString());
+      }
+      return Optional.of(respBody);
+    } catch (final InterruptedException e) {
```
mattdailis, line 138: We should at least reset the interrupted flag here (Thread.currentThread().interrupt())

My limited understanding is that InterruptedException is thrown when the thread receives a request to terminate (ctrl+c or similar...?), so I don't think we'll want to retry in that scenario.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -114,9 +151,16 @@ private Optional<JsonObject> postRequest(final String gqlStr) throws IOException
...
+    final var query = """
+        query GetPlanRevision($id: Int!) {
+          plan_by_pk(id: $id) {
+            revision
+          }
+        }
+        """;
+    final var variables = Json.createObjectBuilder().add("id", planId.id()).build();
+
+    final var response = postRequest(query, variables).orElseThrow(() -> new NoSuchPlanException(planId));
```
mattdailis, line 163: I believe this was already the case before this PR, so I won't insist on fixing it now, but I think that throwing NoSuchPlanException every time anything goes wrong is misleading and makes it difficult to determine what actually went wrong.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
dyst5422: I 100% copy pasted the new request with arguments method from the existing request by string method, so I will 100% defer to you on this element and happy to make the change in both methods

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
dyst5422: I think in this case, it is only when the postRequest returns an empty optional, in which case it "should" interpret as the plan doesn't exist. I think the issue is in our liberal returning of Optional.empty in the postRequest methods for cases that are not actually an "empty" return value, but an error that occurred during resolution of the response (ie non 200 status codes, and interruptions)

dyst5422: Generally operating fast and loose with failure modes

dyst5422:
@mattdailis since both of the above concerns were pre-existing, how do we feel about merging this in and addressing them in a tech-debt ticket I can slot in for this coming release?

mattdailis (approved):

goetzrrGit (approved):
I don't see anything wrong. I like how you clean up the gql calls. They are more readable now. 

Just make sure you have a ticket in JIRA to go back and update the `TODO`

PR #284, opened 2022-08-11
[AERIE-1941] Rename activity table to activity_directive

dyst5422:
* **Tickets addressed:** AERIE-1941
* **Review:** By commit  

## Description

A lot of cmd-shift-f to find all the replacement locations, then iteratively going through and finding missed sports until the tests were all green again.

If you think you work with a part/path of the code that isn't currently a part of one of the e2e tests, speak up so we don't end up accidentally missing a necessary refactor.

## Verification

Tests supported refactor


camargo:
This will require a bunch of changes in the aerie-ui as well. Please do not merge until we have a PR for those since it will break the UI.

goetzrrGit (approved):
The commanding side checks out. No errors on my end.

I was too afraid to test the planning side :P

camargo (approved):

PR #285, opened 2022-08-11
[AERIE-2002]  Generate a standalone sequence seqjson

goetzrrGit:
* **Tickets addressed:** AERIE-2002 
* **Review:** By commit  

## Description
Some users will like to a create sequence files outside of the Falcon Editor using their own tool suite or text editor. These users will need to have a way to generate a sequence seqjson file out of these standalone sequences. 

Below is the standard file format that should be followed by the falcon editor when saving sequences files into AERIE. The Falcon editor should be able to wrap a UI around this format. The users working outside of AERIE should use this format in their sequence file. 

```
export default () =>
  Sequence.new({
    seqId: "test_00001",
    metadata: {},
    commands: [
        BAKE_BREAD,
        A`2020-060T03:45:19`.PREHEAT_OVEN(100),
    ],
  });
```

When generating a sequence seqjson, the Command Expansion server will need a command dictionary and sequence file like the one above.

Here is the GQL call 

```
query GenerateEDSL($commandDictionaryID: Int!, $edslBody: String!) {
  generateSequenceSeqJsonEDSL(commandDictionaryID: $commandDictionaryID, edslBody: $edslBody) {
    id
    metadata
    steps {
      args
      metadata
      stem
      time {
        tag
        type
      }
      type
    }
  }
}

{
  "commandDictionaryID": 1,
  "edslBody": "export default () =>\n  Sequence.new({\n    seqId: \"test_00001\",\n    metadata: {},\n    commands: [\n        BAKE_BREAD,\n        A`2020-060T03:45:19`.PREHEAT_OVEN(100),\n    ],\n  });"
}

```

The above GQL call will generate the sequence seqjson file below:

```
{
      "id": "test_00001",
      "metadata": {},
      "steps": [
        {
          "args": [],
          "metadata": {},
          "stem": "BAKE_BREAD",
          "time": {
            "type": "COMMAND_COMPLETE"
          },
          "type": "command"
        },
        {
          "args": [
            100
          ],
          "metadata": {},
          "stem": "PREHEAT_OVEN",
          "time": {
            "tag": "2020-060T03:45:19.000",
            "type": "ABSOLUTE"
          },
          "type": "command"
        }
      ]
    }
```

## Verification
I wrote a `Unit test` to test the new endpoint in the Command Expansion Server. Also, you can run the GQL above and get the same output.

## Documentation
I will have to add a new documentation page on the Command Expansion Wiki addressing the generating of a sequence seqjson file outside of AERIE. 

## Future work
Probably hook this up to Falcon editor at some point.

### DIFF
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -19,7 +19,7 @@ import {
 } from './lib/batchLoaders/simulatedActivityBatchLoader.js';
 import { expansionSetBatchLoader } from './lib/batchLoaders/expansionSetBatchLoader.js';
 import Piscina from 'piscina';
-import type { executeExpansion, typecheckExpansion } from './worker.js';
+import type { executeExpansion, typecheckExpansion, executeEDSL } from './worker.js';
 import { isRejected, isResolved } from './utils/typeguards.js';
 import { expansionBatchLoader } from './lib/batchLoaders/expansionBatchLoader.js';
 import type { UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
@@ -433,7 +433,45 @@ export interface SeqBuilder {
   ): Sequence;
 }
 
-app.post('/get-seqjson-for-sequence', async (req, res, next) => {
+//generate a sequence JSON from a sequence standalone file
+app.post('/get-seqjson-for-sequence-standalone', async (req, res, next) => {
+  const commandDictionaryID = req.body.input.commandDictionaryID as number;
+  const edslBody = req.body.input.edslBody as string;
+
+  let commandTypes;
+  try {
+    const context: Context = res.locals['context'];
+    commandTypes = await context.commandTypescriptDataLoader.load({ dictionaryId: commandDictionaryID });
+  } catch (e) {
+    res.status(200).json({
+      message: 'Error loading command dictionary',
+      cause: (e as Error).message,
+    });
+    return next();
+  }
+
+  const result = await ((await piscina.run(
+    {
+      edslBody,
+      commandTypes,
+    },
+    { name: 'executeEDSL' },
+  )) as ReturnType<typeof executeEDSL>);
+
+  const sequenceJson = result.sequenceJson;
+  if (sequenceJson != null && result.errors.length === 0) {
+    res.status(200).json(sequenceJson);
+  } else {
+    res.status(200).json({
+      message: 'Failure to generate sequence JSON',
+      cause: result.errors,
+    });
+  }
+
+  return next();
+});
+
+app.post('/get-seqjson-for-seqid-and-simulation-dataset', async (req, res, next) => {
   // Get the specified sequence + activity instance ids + commands from the latest expansion run for each activity instance (filtered on simulation dataset)
   // get start time for each activity instance and join with activity instance command data
   // Create sequence object with all the commands and return the seqjson
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -61,7 +61,10 @@ export interface SequenceSeqJson {
 }
 
 declare global {
-  class Command<A extends ArgType[] | { [argName: string]: any } = [] | {}> {
+  class Command<
+    A extends ArgType[] | { [argName: string]: any } = [] | {},
+    M extends Record<string, any> = Record<string, any>,
+  > {
     public static new<A extends any[] | { [argName: string]: any }>(opts: CommandOptions<A>): Command<A>;
 
     public toSeqJson(): CommandSeqJson;
@@ -72,6 +75,17 @@ declare global {
 
     public relativeTiming(relativeTime: Temporal.Duration): Command<A>;
   }
+
+  class Sequence {
+    public readonly seqId: string;
+    public readonly metadata: Record<string, any>;
+    public readonly commands: Command[];
+
+    public static new(opts: { seqId: string; metadata: Record<string, any>; commands: Command[] }): Sequence;
+
+    public toSeqJson(): SequenceSeqJson;
+  }
+
   type Context = {};
   type ExpansionReturn = Arrayable<Command>;
 
--- a/command-expansion-server/src/worker.ts
+++ b/command-expansion-server/src/worker.ts
@@ -8,7 +8,7 @@ import { UserCodeError, UserCodeRunner } from '@nasa-jpl/aerie-ts-user-code-runn
 
 import getLogger from './utils/logger.js';
 import type { SimulatedActivity } from './lib/batchLoaders/simulatedActivityBatchLoader.js';
-import type { Command } from './lib/codegen/CommandEDSLPreface.js';
+import type { Command, SequenceSeqJson } from './lib/codegen/CommandEDSLPreface.js';
 import type { Mutable } from './lib/codegen/CodegenHelpers';
 import { deserializeWithTemporal } from './utils/temporalSerializers.js';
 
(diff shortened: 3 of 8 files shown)


command-expansion-server/src/app.ts
```diff
@@ -433,6 +433,44 @@ export interface SeqBuilder {
   ): Sequence;
 }
 
+//generate a sequence JSON from a sequence file
+app.post('/get-sequence-edsl', async (req, res, next) => {
```
dyst5422: I think the name of this endpoint is a little confusing. I would have it reference that we are going from edsl -> seqjson

command-expansion-server/src/app.ts
```diff
@@ -433,6 +433,44 @@ export interface SeqBuilder {
...
+  const result = await ((await piscina.run(
+    {
+      edslBody,
+      commandTypes,
+    },
+    { name: 'executeEDSL' },
+  )) as ReturnType<typeof executeEDSL>);
+
+  const sequenceJson = result.sequenceJson;
+  if (sequenceJson != null) {
```
dyst5422: I'm going to claim that errors should be the differentiating check here instead of sequenceJson existence. Err on the side of aggressive failure rather than silent success and suppression of errors.

command-expansion-server/src/app.ts
```diff
@@ -433,6 +433,44 @@ export interface SeqBuilder {
...
+    res.status(200).json({
+      message: 'Failure to generate sequence JSON',
+      cause: result.errors,
+    });
+  }
+
+  return next();
+});
+
 app.post('/get-seqjson-for-sequence', async (req, res, next) => {
```
dyst5422: Given the above comment about the new api endpoint, I think it might make sense to rename this one as well to be something more along the lines of `get-seqjson-for-seqid-and-simulation-dataset` or some less verbose version of that

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -72,6 +75,17 @@ declare global {
 
     public relativeTiming(relativeTime: Temporal.Duration): Command<A>;
   }
+
+  class Sequence {
```
dyst5422, line 79: I worry about exposing Sequence as a global construct that is accessible in expansions. I think there is an argument to be made that we should split at least the exposed types between expansions and sequence authoring.

command-expansion-server/src/worker.ts
```diff
@@ -110,3 +110,41 @@ export async function executeExpansion(opts: {
...
+  sequenceJson: SequenceSeqJson | null;
+  errors: ReturnType<UserCodeError['toJSON']>[];
+}> {
+  try {
+    const result = await codeRunner.executeUserCode(
+      opts.edslBody,
+      [],
+      'Sequence',
+      [],
+      3000,
```
dyst5422: 2 things:

1. I think 3 seconds for sequence authoring to execute is probably too lenient - if something is taking longer than maybe 1 second, I think it's more indicative of either a mistake in implementation of the sequence or problematic programming constructs. That said, 3 seconds isn't likely to be all that much more problematic than 1 second.
2. We should probably make these configuration variables at some point, not necessarily in this PR

dyst5422 (approved):
Nothing broken or critical I can see. Just some thoughts for discussion.

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I am not sure how to limit `class Sequence` to only the `sequence authoring` side as both `expansion` and `sequence authoring` use the same preface. We could create a sequence-specific preface that is only used when expanding to a sequence. The prefaces are built when you upload a command dictionary so that means we will have to build 2 of them - one for expansion and one for sequences. 

We might need a separate ticket to discuss/implement this unless there is an easier way to limit the exposure in TS? The preface makes the EDSL very flexible for us to reuse a lot of the same code and the user. Maybe this is the side effect of that though.

dyst5422: I think that's fine to take this on in a different PR. Yes, we'd need to split to multiple files in order to make it work. I think the best implementation would be to keep that same implementation file and multiple global declaration files to only expose the parts we want exposed (perhaps even some shared global declarations and some differing ones).

command-expansion-server/src/app.ts
goetzrrGit: I can't think of a less verbose name so I changed it to `get-seqjson-for-seqid-and-simulation-dataset` which matches the new endpoint `get-seqjson-for-sequence-standalone`

command-expansion-server/src/app.ts
goetzrrGit: This should be better :P

`get-seqjson-for-sequence-standalone`

command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Would this inheritance affect the `aerie-ts-runner`? I think we can just pass more file paths into it but I wasn't sure how that works with the Global namespace of TS. 

```
#Commands
await codeRunner.executeUserCode(
      opts.edslBody,
      [],
      'Commands',
      [],
      1000,
      [
        ts.createSourceFile('global-types.ts', opts.globalTypes, compilerTarget),
        ts.createSourceFile('command-types.ts', opts.commandTypes, compilerTarget)
        ts.createSourceFile('TemporalPolyfillTypes.ts', temporalPolyfillTypes, compilerTarget),
      ],

#Sequence
await codeRunner.executeUserCode(
      opts.edslBody,
      [],
      'Sequence',
      [],
      1000,
      [
        ts.createSourceFile('global-types.ts', opts.gloablTypes, compilerTarget),
        ts.createSourceFile('sequence-types.ts', opts.commandTypes, compilerTarget)
        ts.createSourceFile('TemporalPolyfillTypes.ts', temporalPolyfillTypes, compilerTarget),
      ],
```

dyst5422: Yeah, we just pass additional files with `declare global {}` blocks. I think we might need a `export {}` at the end or some such thing to ensure they're interpreted correctly.

dyst5422: In fact. We could probably figure out a way to have those declaration files be static and not needed to be generated through some intelligent use of importing from the generated file.

deployment/hasura/metadata/actions.graphql
```diff
@@ -132,6 +132,13 @@ type Query {
   ): SequenceSeqJson!
 }
 
+type Query {
+  generateSequenceSeqJsonEDSL(
```
camargo: I'm wondering if we can call this action `getUserSequenceSeqJson`?
And maybe later in a separate PR we can rename the exisiting `getSequenceSeqJson` to `getExpansionSequenceSeqJson`? This way each type is more explicit and paves the way for keeping both sequence types in the database.

dyst5422: I thought about this as well when reviewing. I just wasn't sure what the most obvious names would be. In both cases, we are generating SeqJson from the EDSL, but in one, we take a Sequence EDSL, in the other, we generate from sequence constructs in the plan.

goetzrrGit: Sure I can make this change. Yeah, coming up with a name is tricky as I don't want to make it super long, but also want it to be meaningful :) I will also rebase with develop - I saw your slack message @camargo .

dyst5422: Generally speaking, as long as it's not so long that it causes serious formatting issues, I am a fan of verbose and descriptive naming - even when it may occasionally seem excessive.

camargo (approved):
LGTM

PR #288, opened 2022-08-14
Javadocs for the parser combinator library

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR is based on the discussion begun at https://github.com/NASA-AMMOS/aerie/discussions/276. It consists largely of new and improved Javadocs for the JSON parser combinator library, but also includes some source-level refactoring and one bugfix (concerning the name of the tag field in the schema generated for a `sumP`).

I'm planning to add more Javadocs for the `ProductParsers` and the individual parsers in `BasicParsers`, as well as for the `JsonParseResult` that bubbles parse results up through the composed parsers, but there should be enough here to get some initial feedback.

In particular, the new `package-info.java` contains a full-on tutorial on authoring a JSON format for a simple expression DSL. Feedback on that in particular would be great.


Twisol:
@JoelCourtney, it's a bit late, but here are the initial docs we discussed. 👋 Eager to hear your thoughts, even though I still have some more to do.

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/Uncurry.java
```diff
@@ -5,29 +5,42 @@
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
+/**
+ * A set of utility functions for transforming the fields of an object parser.
+ *
+ * <p> The parsers produced from {@link ProductParsers} work with iteratively nested tuples, which are inconvenient
+ * to work with directly. The static functions in this class let you use an n-ary function where a nested tuple
+ * would normally be given, or return an n-tuple rather than a series of nested 2-tuples. </p>
```
JoelCourtney: Nitpick:

> work with iteratively nested {@link Pair}s

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/package-info.java
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * <p> This library provides the basic building blocks for defining serialization formats for any Java type,
+ * together with "parsers" which can convert both to and from these formats. A format is defined grammatically, by
+ * composing simpler pre-existing formats into more complex assemblies with the help of general-purpose combiners.
+ * We prioritize ease-of-use and expressivity over efficiency: parsers defined with this library may be slower than
+ * those defined with other libraries, but our goal is for them to be more maintainable in certain dimensions
+ * than alternatives. </p>
+ *
+ * <p> We take the position that serialization is a separate and orthogonal concern from domain modeling. Domain types
+ * ought not favor a particular mode of serialization, as business needs can easily change change over time. Formats
```
JoelCourtney: change change

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/package-info.java
```diff
@@ -0,0 +1,181 @@
...
+ *     return chooseP(
+ *         numP,
+ *         sumP("op", intExprClass, List.of(
+ *             new Variant<>("+", Expr.Add.class, addP(integerExprP)),
+ *             new Variant<>("-", Expr.Negate.class, negateP(integerExprP))
+ *         )));
+ *   }
+ *
+ *   final JsonParser<Expr<Integer>> integerExprP
+ *       = recursiveP(selfP -> integerExprP(selfP)); }
```
JoelCourtney, line 127: I think that would be a compile error because they both have the name `integerExprP`

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/BasicParsers.java
```diff
@@ -58,9 +43,27 @@ public JsonValue unparse(final JsonValue value) {
     }
   };
 
+  public static final JsonParser<Void> noneP = new JsonParser<>() {
```
JoelCourtney, line 46: Cool. It makes mechanical sense, but when would we ever use it?

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/JsonObjectParser.java
```diff
@@ -0,0 +1,34 @@
...
+
+public interface JsonObjectParser<T> extends JsonParser<T> {
+  @Override
+  JsonObject unparse(final T value);
+
+  @Override
+  default <S> JsonObjectParser<S> map(final Iso<T, S> transform) {
+    Objects.requireNonNull(transform);
+
+    final var self = this;
```
JoelCourtney, line 16: Why not use `this`?

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/JsonParser.java
```diff
@@ -3,31 +3,130 @@
...
     return Json
         .createObjectBuilder()
         .add("$schema", Json.createValue("https://json-schema.org/draft/2020-12/schema"))
-        .addAll(Json.createObjectBuilder(this.getSchema(new HashMap<>())))
+        .addAll(Json.createObjectBuilder(this.getSchema(new SchemaCache())))
         .build();
   }
 
+  /**
+   * Adapts this parser losslessly from one domain type to another.
```
JoelCourtney: In practice we don't use this losslessly :|. I believe in hasura actions we sometimes map `string -> UNIT` and `UNIT -> ""` because its only used for deserialization and we don't care about the field, only that it exists. I guess that's not a issue with the docs though.

JoelCourtney (approved):
dank tutorial, thank you!

I do have some nitpicks but I'm preemptively approving anyway

Twisol:
Sorry @JoelCourtney, somehow I totally missed the notification that you'd reviewed this! I'm picking this back up now.

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/package-info.java
Twisol: Actually, nope! Methods and variables live in different namespaces. That's how records like `record Foo(A a, B b) {}` can have both `foo.a` and `foo.a()` be defined.

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/Uncurry.java
Twisol: Good point. I've changed it in the first instance (the one you quoted) but will keep "tuple" elsewhere as a descriptive word.

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/package-info.java
Twisol: fixed fixed

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/BasicParsers.java
Twisol: Mmm, I can come up with any number of contrived cases. One halfway-plausible one is that you might have a function that builds a parser based on some other information. If the information is inconsistent in some way, you still have to return a parser, so you can return `noneP` instead of using `null` or wrapping it in an `Optional`. (I am assuming the inconsistency isn't the kind that would motivate an exception instead.)

Another minor example is if you have a list of options you want to combine in some way; in the denegerate case where you have no options, you can again return `noneP`. It's like how, if you're summing over a list, you have to ask what the sum of an empty list should be -- even though you normally don't *intentionally* sum over an empty list, having a zero makes the behavior consistent across the board.

Things like this usually come up as edge cases, where you want the edges to play nicely with the "normal" behavior. It's nice that it exists, but it's not something you'd reach for daily.

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/JsonObjectParser.java
Twisol: On the next line, we construct an anonymous object. Within the methods of that object, `this` would refer to that object, not the outer parser on which `map` was invoked. So `self` makes the outer parser accessible.

Technically speaking, I could replace all uses of `self` with `JsonObjectParser.this`, which disambiguates to Java which object we're referring to. (The anonymous object is actually an instance of an anonymous class, so `JsonObjectParser.this` really does refer to the outer object uniquely.) But I don't think it actually makes things any clearer to the human reader.

Twisol: (It's a common idiom [in JavaScript, too](https://stackoverflow.com/a/338106), which is where I picked it up. It's less necessary these days in JS because of the prevalence of arrow functions.)

parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/JsonParser.java
Twisol: Well, there's two questions of lossiness here. Is the *Iso* lossy, and is the *parser* lossy?

The case you're talking about is, I believe, a case where the *parser* is lossy. That just means that if you parse and then unparse the result, you might not get the same JSON object back. This is *explicitly allowed* in the `JsonParser` contract; only the other round-trip needs to be exact.

Conversely, an `Iso` really does need to be lossless, since you're just reshaping the data between a generic form easy for the parser combinators to work with, and a domain-specific form easy for *you* to work with.

Twisol: Nah, hang on, I see the point you're making. Let me stew on this.

Twisol: You're right. Looking at this one for instance:

```java
  public static final JsonParser<HasuraAction<HasuraAction.ActivityInput>> hasuraActivityActionP
      = hasuraActionP(hasuraActivityInputP)
      . map(Iso.of(
          untuple((name, input, session, requestQuery) -> new HasuraAction<>(name, input, session)),
          $ -> tuple($.name(), $.input(), $.session(), "")));
````

The parser is perfectly well-formed, as far as I can tell. An instance of `HasuraAction<HasuraAction.ActivityInput>` should unparse and then parse back to an equivalent instance. This is not true of parsing then unparsing (which will replace `requestQuery` with an empty string), but as noted, it doesn't need to be true.

That means that "isomorphism" is not the correct notion here. We just want the parsing direction to undo the unparsing direction -- that is, parsing is a "left inverse" of unparsing. The notion of isomorphism specializes to a two-sided inverse, but that's too strong for us.

I'll make some changes and come up with a better name.

Comment (Twisol) [reply parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/JsonParser.java]: I've renamed `Iso` to `Convert`, adjusted its description to reflect the weaker requirements, and (as a bonus) replaced all uses of `map(Iso.of(x, y))` with a simpler `map(x, y)` form.

**Comment by Twisol** [comment]
Per discussion with Pat, I'm merging this in now. I can always come back and add more documentation, but I need to focus on simulation system documentation next.

**Review by patkenneally — APPROVED** [review]

File: parsing-utilities/src/main/java/gov/nasa/jpl/aerie/json/package-info.java [hunk 7]
```diff
@@ -0,0 +1,181 @@
...
+ * ought not favor a particular mode of serialization, as business needs can easily change over time. Formats can be
+ * authored using this library without any need to modify the domain types they operate on. Moreover, where other
+ * libraries may infer details about the serialization format by inspecting the class definitions of domain types, we
+ * accept some potential overhead in repeating ourselves to cleanly separate the concern of what <i>developers</i> call
+ * a field from what <i>clients</i> call it -- and whether their data is structually organized the same way at all. </p>
+ *
+ * <p> We <b>eschew reflection and convention</b> in favor of explicit control over how Java values are modeled in JSON.
+ * Other libraries may require little to no configuration by relying on reflection and convention, but needs outside
+ * the happy path begin to look very different (and ad-hoc) from where you started. This library tries to provide a
+ * uniform and consistent authoring experienc: if you can author a simple format, odds are you know everything you need
```
patkenneally: experienc -> experience

Twisol: Thanks; fixed. (Also fixed the &lt;/li&gt; that should have been a &lt;/p&gt;.)

PR #289, opened 2022-08-16
Remove {POSTGRES_DB} enviornment variable

Mythicaeda:
* **Review:** By commit  

## Description

Removed the POSTGRES_DB environment variable, as it doesn't need to be a secret.

## Verification

Composed the project and ran `./gradlew test`.

## Documentation

Removed mention of the POSTGRES_DB environment variable from the README and .env.template.


JoelCourtney:
Actually, it seems weird to me to make the db name configurable, especially since it affects the container name. What was the rationale behind making this configurable?

I think it would be better to put only the username and password in `.env`, because the db name is not a secret.

goetzrrGit:
It is a good idea, but I agree with Joel, I feel like having the container name in a .env is more trouble than it is worth. I don't think there is a need to keep the DB name a secret for security reasons, but I could be wrong.

Mythicaeda:
I've modified the PR so `POSTGRES_DB` is no longer in the `.env` @JoelCourtney and @goetzrrGit

goetzrrGit (approved):
Looks good to me.

skovati:
FYI @Mythicaeda, you can apply the `hotfix` label to reduce required approvals to one ;)

skovati (approved):
That being said, LGTM

PR #290, opened 2022-08-16
[AERIE-2004] extra quotes in constraints codegen

JoelCourtney:
* **Tickets addressed:** AERIE-2004
* **Review:** By commit  

## Description

Constraints codegen has a bug where activities with enum type parameters will have generated typescript that has an extra pair of quotes: `Discrete<| ""Option1"" | ""Option2"">`.

Commits:
1. Duplicates the ValueSchemaJsonParser in scheduler-server into the incorrect version in merlin-server
2. Moves the ValueSchemaJsonParser into merlin-driver json package
3. Adds an activity with an enum parameter to the StubMissionModel for testing

## Verification
I created commit 3 first. Without the fix, building fails with the error caused by extra quotes.


dyst5422 (approved):

JoelCourtney:
(force push was just to rename the first commit)

mattdailis (approved):
Looks good, thanks, Joel :+1:

PR #292, opened 2022-08-17
[AERIRE 1918] Add schema property to real profiles

Mythicaeda:
* **Tickets addressed:** AERIE-1918
* **Review:** By commit  

## Description

Unlike discrete profiles, real profiles weren't posting their schemas in the `profile` table, just their type. RealProfiles were modified to keep track of their ValueSchema, and their JSON parser was changed to return the type and schema in a pair similar to how DiscreteProfiles are processed. They now appear like so in the database:

image

## Verification

The test `simulationResultsTest` in `IncrementalSimulationTest.java` was updated to reflect the change in the structure of RealProfiles. Besides that, the changes were ran and tested locally.

## Future work

There's an open discussion regarding setting up simulation end-to-end tests.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationResults.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationResults.java
@@ -17,7 +17,7 @@ import java.util.SortedMap;
 
 public final class SimulationResults {
   public final Instant startTime;
-  public final Map<String, List<Pair<Duration, RealDynamics>>> realProfiles;
+  public final Map<String, Pair<ValueSchema, List<Pair<Duration, RealDynamics>>>> realProfiles;
   public final Map<String, Pair<ValueSchema, List<Pair<Duration, SerializedValue>>>> discreteProfiles;
   public final Map<String, List<Pair<Duration, SerializedValue>>> resourceSamples;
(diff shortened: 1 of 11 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationResults.java
```diff
@@ -46,13 +46,14 @@ public SimulationResults(
...
   private static Map<String, List<Pair<Duration, SerializedValue>>>
   takeSamples(
-      final Map<String, List<Pair<Duration, RealDynamics>>> realProfiles,
+      final Map<String, Pair<ValueSchema, List<Pair<Duration, RealDynamics>>>> realProfiles,
       final Map<String, Pair<ValueSchema, List<Pair<Duration, SerializedValue>>>> discreteProfiles)
   {
     final var samples = new HashMap<String, List<Pair<Duration, SerializedValue>>>();
 
-    realProfiles.forEach((name, profile) -> {
+    realProfiles.forEach((name, p) -> {
```
goetzrrGit, line 54: Totally a style thing for readablity. I would change `p` to `profile`

goetzrrGit (approved):
Looks good to me!!!

mattdailis (approved):
Looks good to me based on visual inspection

PR #294, opened 2022-08-18
[HOTFIX] Fix bad npm registry

dyst5422:
* **Review:** By commit  

## Description

When I updated the user code runner, my npm registry was set to JPL artifactory, so that got encoded into the package-lock files. This fixes that blunder.

Also locks version for typescript-json-schema to work around a dependency error with the 0.53.1 version

### DIFF
--- a/merlin-server/constraints-dsl-compiler/package-lock.json
+++ b/merlin-server/constraints-dsl-compiler/package-lock.json
@@ -19,30 +19,43 @@
         "@types/node": "^17.0.21"
       }
     },
-    "node_modules/@cspotcode/source-map-consumer": {
-      "version": "0.8.0",
-      "resolved": "https://registry.npmjs.org/@cspotcode/source-map-consumer/-/source-map-consumer-0.8.0.tgz",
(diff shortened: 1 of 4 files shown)


goetzrrGit (commented):
These files were deleted but don't we want to create new lock files?

dyst5422:
> These files were deleted but don't we want to create new lock files?

🤦 ~amended the commit~ - ~ugh why is this such a PITA, gimme a sec~ - THERE WE GO

goetzrrGit (commented):
That looks better :D One more question what about the command-expansion-server lock file? Was that one not messed up?

dyst5422:
This has turned into a dependency nightmare. I think I've got it right this time.

dyst5422:
> That looks better :D One more question what about the command-expansion-server lock file? Was that one not messed up?

Great question. I have no idea. It's clearly right here https://github.com/NASA-AMMOS/aerie/blob/hotfix/bad-npm-source/command-expansion-server/package-lock.json#L1046

Perhaps I had already installed the 0.3.1 locally on a different day when my registry was correctly set and so the package-lock didn't change.

dyst5422:
@goetzrrGit finally got it

skovati (approved):
🥵

goetzrrGit (approved):
Merge, Merge, Merge

PR #298, opened 2022-08-23
[AERIE-1997] Add custom PR logic action

skovati:
* **Tickets addressed:** AERIE-1997
* **Review:** By commit  

## Description

The following is a description of the underlying action, as this PR just brings [aerie-pr-action](https://github.com/NASA-AMMOS/aerie-pr-action) into the `aerie` repo.

`aerie-pr-action` currently does two main things. On PR creation it will automatically assign the PR opener as the assignee. This is done so that Github assignees are treated akin to Jira assignees, i.e. the person mainly responsible for commits, testing, and pushing the PR over the finish line.

Secondly, on label or unlabed triggers, this action will automatically give one PR approval to PRs labeled with either `documentation` or `hotfix`. This is done so that the PR will only need one additional approval from a "real human" i.e. codeowner, speeding up the merge velocity for small PRs.

Architecturally, this action was designed to be somewhat monolithic, i.e. it handles multiple paths of custom logic and switches based on the type of trigger event, e.g. `opened` or `labeled`. This was done so that we could easily add new logic to this action, e.g. commenting a welcome message with a link to `CONTRIBUTING.md` for PRs opened by new contributors, without having to create and add several new actions to `aerie` workflows.

## Verification

Lots of manual verification. Testing framework has been setup for the underlying action, which can be added to as we add more custom logic. 

## Documentation

None invalidated, although this new functionality could be documented in e.g. `CONTRIBUTING.md` for new open source contributors.

## Future work

Any additional custom PR logic we desire can be added to this action.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,14 @@
+name: "Aerie PR Action"
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - ready_for_review
+      - labeled
+      - unlabeled
+jobs:
+  pr_logic:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: nasa-ammos/aerie-pr-action@main


skovati:
I added the `hotfix` label temporarily to demo the automatic approval by the `github-actions` bot, now this PR would just need one approval from a codeowner to be merged in.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,14 @@
+name: "Aerie PR Logic"
```
camargo: I'm wondering if we should just call this `Aerie PR Action` to reflect that is what is being used here.

camargo (approved):

.github/workflows/pr.yml
skovati: Good call, change has been made

mattdailis:
I really like this functionality. I worry that it could incentivize us to use the "HOTFIX" label inappropriately (also - small changes can break things just as easily as big changes), but we'll just need to keep each other honest about that :+1:

skovati:
> I really like this functionality. I worry that it could incentivize us to use the "HOTFIX" label inappropriately (also - small changes can break things just as easily as big changes), but we'll just need to keep each other honest about that 👍

@mattdailis I totally agree, that was my main concern making it label based, but I couldn't think of a good heuristic that could categorize PRs as hotfixes otherwise. It'll just have to be an "on your honor" kind of thing...

pcrosemurgy (approved):

PR #300, opened 2022-08-25
[AERIE 1815] Replace org.json dependency with javax.json.JsonObject dependency

Mythicaeda:
* **Tickets addressed:** AERIE-1815
* **Review:** By commit  

## Description

When we write to the typescript compiling node process, we were using several calls to `org.json.JSONObject.quote()` to build up a string describing a JSON. This replaces those calls by using `javax.json` to build a `JsonObject` with the same entries. This also removes any references to `org.json` from the gradle files.

## Verification

- Ran the gradle tests
- Built the project and did a visual comparison between the output of both the original and the new techniques, confirming that they are functionally identical.

## Future work

We ought to consider giving `SchedulingDSLCompilationService.compileSchedulingGoalDSL()` and `ConstraintsDSLCompilationService.compileConstraintsDSL()` a timeout and/or the ability to be interrupted, because as of right now they block on `output.readline()` indefinitely until the node process outputs a new-line-terminated string.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
@@ -7,9 +7,9 @@ import gov.nasa.jpl.aerie.merlin.server.http.InvalidEntityException;
 import gov.nasa.jpl.aerie.merlin.server.http.InvalidJsonException;
 import gov.nasa.jpl.aerie.merlin.server.models.ConstraintsCompilationError;
 import gov.nasa.jpl.aerie.constraints.json.ConstraintParsers;
-import org.json.JSONObject;
 
 import javax.json.Json;
+import javax.json.JsonObject;
 import javax.json.stream.JsonParsingException;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


pcrosemurgy (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
```diff
@@ -52,8 +52,11 @@ public void close() {
   synchronized public ConstraintsDSLCompilationResult compileConstraintsDSL(final String missionModelId, final String constraintTypescript)
   throws MissionModelService.NoSuchMissionModelException
   {
-    final var missionModelGeneratedCode = JSONObject.quote(this.typescriptCodeGenerationService.generateTypescriptTypesFromMissionModel(missionModelId));
-
+    final var missionModelGeneratedCode = this.typescriptCodeGenerationService.generateTypescriptTypesFromMissionModel(missionModelId);
+    final JsonObject typescriptCode = Json.createObjectBuilder()
```
mattdailis: I'd suggest renaming `typescriptCode` to something like `messageJson`, to signify that this is the payload being sent to the subprocess. (When I read `typescriptCode`, I expect that to be a string that contains the contents of a typescript file).

Brainstorming other names: `payload`, `payloadJson`, `jsonPayload`, `messageToSubprocess`....

mattdailis (approved):
Looks good to me! Just one variable rename request - after addressing that, feel free to merge.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationService.java
Mythicaeda: I called it that because it's a JSON wrapper on generated Typescript code, but I can see what you mean. Change is done!

mattdailis: Looks great, thanks

PR #301, opened 2022-08-26
[FIX] Fix scheduler

adrienmaillard:
* **Tickets addressed:** FIX
* **Review:** By commit  

## Description

While trying to reproduce AERIE-1904, I found two issues that make the scheduler currently broken :
- it was possible to set the arguments of an activity instance by directly assigning to a map. Thus it was possible to pass an immutable map. We expect these maps to be mutable. 
- when the user creates an activity with a controllable duration and does not set its duration, the scheduler was updating its duration after running scheduling because it was not pulling effective arguments. 

Here is what I (with help from @mattdailis) have done (4 commits):
- clear and putall arguments of activity instance instead of copying the reference of passed maps. 
- use effective arguments in the scheduler which gets default values in case the user has not set some 
- prevent the scheduler from updating and deleting activities. Originally, the intent was to be general but it is not supposed to be supported yet. 
- add a test for this particular case in the e2e-tests 

## Verification

Manual testing of the scheduler via the aerie-ui + a new test in e2e-tests.

## Documentation

None.

## Future work

None.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
@@ -4,6 +4,7 @@ import gov.nasa.jpl.aerie.json.BasicParsers;
 import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
+import gov.nasa.jpl.aerie.merlin.protocol.types.InvalidArgumentsException;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 import gov.nasa.jpl.aerie.scheduler.model.ActivityInstance;
@@ -240,7 +241,7 @@ public record GraphQLMerlinService(URI merlinGraphqlURI) implements PlanService.
    */
   @Override
   public MerlinPlan getPlanActivityDirectives(final PlanMetadata planMetadata, final Problem problem)
-  throws IOException, NoSuchPlanException, PlanServiceException, InvalidJsonException
+  throws IOException, NoSuchPlanException, PlanServiceException, InvalidJsonException, InvalidArgumentsException
   {
     final var merlinPlan = new MerlinPlan();
     final var request =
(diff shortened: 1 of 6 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/model/ActivityInstance.java
```diff
@@ -362,15 +362,15 @@ public Map<String, VariableArgumentComputer> getVariableArguments() {
    * @param arguments a name/value map of arguments
    */
   public void setArguments(Map<String, SerializedValue> arguments) {
-    this.arguments = arguments;
+    this.arguments.putAll(arguments);
```
mattdailis, line 384: Is setArguments expected to remove all existing arguments, or is it an overlay?

E.g. if this.arguments has { "a": 1 }, and we call setArguments({ "b": 2 }), do we expect the arguments to have both keys, or just "b"? The current implementation will have both keys.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -436,23 +436,32 @@ public void updateActivityDirective(
...
-          update_activity_directive_by_pk(pk_columns: {id: $id, plan_id: $planId}, _set: {arguments: $arguments, start_offset: $startOffset, source_scheduling_goal_id: $goalId}) {
-            affected_rows
+    String query;
+    final var variables = Json.createObjectBuilder()
+                    .add("id", instanceId.id())
+                    .add("planId", planId.id())
+                    .add("arguments", arguments.build())
+                    .add("startOffset", activity.startTimestamp().toString());
+    if(goalId != null) {
+      query = """
```
dyst5422: I would expect that we can have a null value for goalId and use the same query in both cases. We just need to do a null check before calling `.id()`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/model/ActivityInstance.java
mattdailis: The two options I see here are either clearing the map before calling `putAll` (if it's important that the reference remain the same)
```suggestion
    this.arguments.clear();
    this.arguments.putAll(arguments);
```

Or, if it's okay to create a new reference, we could simply create a new map:
```suggestion
    this.arguments = new HashMap<>(arguments);
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
mattdailis: It sounds like we might not even need a null check, as long as the json object builder accepts `.add("goalId", null)`. If so, I recommend dropping the second commit.

Side note: the `updateActivityDirective` shouldn't be called unless the scheduler is updating an activity that's already in the plan, right? So in theory, even if it's broken, none of our codepaths should trigger it. If we are hitting that codepath, that means we have another issue upstream. 🤔

Twisol: https://docs.oracle.com/javaee/7/api/javax/json/JsonObjectBuilder.html

> This class does not allow null to be used as a name or value while building the JSON object

The docs are admittedly outdated, but for some reason my IntelliJ won't let me jump to the sources -- just to decompiled bytecode. Anyway, this matches my understanding last I looked.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/model/ActivityInstance.java
adrienmaillard: You are right, it should remove existing arguments. I'll make the change.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: Finally, these changes have been reverted. I have also commented calls to delete and update. If we end up in this branch anyway, we'll get an exception, the scheduler is not supposed to do that (yet).

adrienmaillard:
I have pushed new commits and updated the description of this PR.

mattdailis (approved):

dyst5422 (approved):
I suspect I introduced this bug, so thanks for cleaning up after me!

PR #302, opened 2022-08-30
[AERIE-1764] Return more information about errors.

goetzrrGit:
* **Tickets addressed:** AERIE-1764
* **Review:** By commit  

## Description
Just a tiny change to return more information about the errors and the error object to the user. It makes it easier to debug and allows our users to understand the errors they receive from the worker. 

https://hasura.io/docs/latest/actions/action-handlers/#returning-an-error-response

## Verification
Manually created errors to see how they show up in Hasura

## Documentation
none

## Future work
none

### DIFF
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -582,7 +582,15 @@ app.post('/get-seqjson-for-seqid-and-simulation-dataset', async (req, res, next)
 // General error handler
 app.use((err: any, _: Request, res: Response, next: NextFunction) => {
   logger.error(err);
-  res.status(err.status ?? err.statusCode ?? 500).send({ message: err.message });
+
+  res.status(err.status ?? err.statusCode ?? 500).send({
+    message: err.message,
+    extensions: {
+      ...(err.cause ? { cause: err.cause } : {}),
+      ...(err.stack ? { stack: err.stack } : {}),
+      object: err,
+    },
+  });
   return next();
 });
 


command-expansion-server/src/app.ts
```diff
@@ -582,7 +582,15 @@ app.post('/get-seqjson-for-seqid-and-simulation-dataset', async (req, res, next)
 // General error handler
 app.use((err: any, _: Request, res: Response, next: NextFunction) => {
   logger.error(err);
-  res.status(err.status ?? err.statusCode ?? 500).send({ message: err.message });
+
+  res.status(err.status ?? err.statusCode ?? 500).send({
+    message: err.message,
+    extensions: {
+      ...(err.cause ? { cause: err.cause } : {}),
+      ...(err.stack ? { cause: err.stack } : {}),
```
camargo: I think the `stack` case needs to be `{ stack: err.stack }`. Otherwise it will overwrite `cause` if both `stack` and `cause` exist.

dyst5422:
I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.

command-expansion-server/src/app.ts
goetzrrGit: good catch.

goetzrrGit:
> I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.

Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?

command-expansion-server/src/app.ts
goetzrrGit: I made the change

camargo (approved):

dyst5422:
> > I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.
> 
> Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?

Lee Byron - the creator of GraphQL - has specifically stated on multiple occasions that the error mechanisms for GraphQL are NOT intended for known error modes and those should be managed in the return types specifically. Here's a decent article on it https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors

Basically, the "caught errors" should only be used for exceptional errors, not business-logic-defined errors.

Mythicaeda:
> > > I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.
> > 
> > 
> > Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?
> 
> Lee Byron - the creator of GraphQL - has specifically stated on multiple occasions that the error mechanisms for GraphQL are NOT intended for known error modes and those should be managed in the return types specifically. Here's a decent article on it https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors
> 
> Basically, the "caught errors" should only be used for exceptional errors, not business-logic-defined errors.

According to the article you linked, unless we're actually using this to report something like the constraints check failed, this looks like the correct place to report errors such as exceptions/timeouts/unavailable services.

dyst5422:
> > > > I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.
> > > 
> > > 
> > > Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?
> > 
> > 
> > Lee Byron - the creator of GraphQL - has specifically stated on multiple occasions that the error mechanisms for GraphQL are NOT intended for known error modes and those should be managed in the return types specifically. Here's a decent article on it https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors
> > Basically, the "caught errors" should only be used for exceptional errors, not business-logic-defined errors.
> 
> According to the article you linked, unless we're actually using this to report something like the constraints check failed, this looks like the correct place to report errors such as exceptions/timeouts/unavailable services.

Yes, but not for things like "Hey you have some type errors in your sequence and we weren't able to convert it to seqjson"

goetzrrGit:
> > > > > I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.
> > > > 
> > > > 
> > > > Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?
> > > 
> > > 
> > > Lee Byron - the creator of GraphQL - has specifically stated on multiple occasions that the error mechanisms for GraphQL are NOT intended for known error modes and those should be managed in the return types specifically. Here's a decent article on it https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors
> > > Basically, the "caught errors" should only be used for exceptional errors, not business-logic-defined errors.
> > 
> > 
> > According to the article you linked, unless we're actually using this to report something like the constraints check failed, this looks like the correct place to report errors such as exceptions/timeouts/unavailable services.
> 
> Yes, but not for things like "Hey you have some type errors in your sequence and we weren't able to convert it to seqjson"

For my understanding wont. `..(err.cause ? { cause: err.cause } : {}),` return the cause for the sequence generation failure from `aerie-ts-user-code-runner` Unless you want to capture any errors outside of `aerie-ts-user-code-runner` which should show up under `message: err.message`

dyst5422:
> > > > > > I like this change, but I don't think we should be using this mechanism for operations we know are fallible. So this is a good change to have in addition to correcting our responses to be fallible with error fields.
> > > > > 
> > > > > 
> > > > > Hi Dylan I don't think I follow your statement. Are you saying we should not default to this catch-all error reporting and do a more specific error reporting in each endpoint?
> > > > 
> > > > 
> > > > Lee Byron - the creator of GraphQL - has specifically stated on multiple occasions that the error mechanisms for GraphQL are NOT intended for known error modes and those should be managed in the return types specifically. Here's a decent article on it https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors
> > > > Basically, the "caught errors" should only be used for exceptional errors, not business-logic-defined errors.
> > > 
> > > 
> > > According to the article you linked, unless we're actually using this to report something like the constraints check failed, this looks like the correct place to report errors such as exceptions/timeouts/unavailable services.
> > 
> > 
> > Yes, but not for things like "Hey you have some type errors in your sequence and we weren't able to convert it to seqjson"
> 
> For my understanding wont. `..(err.cause ? { cause: err.cause } : {}),` return the cause for the sequence generation failure from `aerie-ts-user-code-runner` Unless you want to capture any errors outside of `aerie-ts-user-code-runner` which should show up under `message: err.message`

The return type under graphql does not enable typed error returns though. The contract GraphQL is stating is that it will return seqjson unless there is some internal error, which doesn't cover the case of "there are type errors here that need to be addressed", for which there may be multiple errors that each have a message, stack, and location.

I think in line with the GraphQL recommendation for errors, we specify for actions like `getUserSequenceSeqJson` we return a type that looks like 
```graphql
type GetUserSequenceSeqJsonResponse {
  seqJson: SequenceSeqJson,
  errors: [UserCodeError!]
}
```

Where when successful, errors = null, and when errors, seqJson = null

pcrosemurgy (approved):

pcrosemurgy:
@goetzrrGit Does this conform to the Hasura GQL error reporting format? I'm having trouble telling if it is from the TS diff but I haven't looked too deep.

goetzrrGit:
@pcrosemurgy Yes, this does conform to what Hasura suggests. There is going to be tickets to catch fallible errors so we don't have to use this catch-all in the PR too often.

goetzrrGit:
Going to merge as we are moving forward with the failable Hasura action in Dylan's PR. This one will be used to catch anything that isn't failable.

PR #304, opened 2022-08-30
[AERIE-1982] feat: add user sequence table

camargo:
* **Tickets addressed:** [AERIE-1982](https://jira.jpl.nasa.gov/browse/AERIE-1982)
* **Review:** By commit  

## Description

- Add first-draft of `user_sequence` table for storing user-defined sequences for 0.13.0

## Verification

Test the `user_sequence` table is added to the commanding database

## Future work

Possible renaming:

- `aerie_commanding` database to `aerie_sequencing`
- `sequence` table to `expansion_sequence`

### DIFF
--- a/command-expansion-server/sql/commanding/tables/user_sequence.sql
+++ b/command-expansion-server/sql/commanding/tables/user_sequence.sql
@@ -1,0 +1,39 @@
+create table user_sequence (
+  authoring_command_dict_id integer not null,
+  created_at timestamptz not null default now(),
+  definition text not null,
+  id integer generated always as identity,
+  name text not null,
+  owner text not null default 'unknown',
+  updated_at timestamptz not null default now(),
+
+  constraint user_sequence_primary_key primary key (id)
+);
+
+comment on column user_sequence.authoring_command_dict_id is e''
+  'Command dictionary the user sequence was created with.';
+comment on column user_sequence.created_at is e''
+  'Time the user sequence was created.';
+comment on column user_sequence.definition is e''
+  'The user sequence definition string.';
+comment on column user_sequence.id is e''
+  'ID of the user sequence.';
+comment on column user_sequence.name is e''
+  'Human-readable name of the user sequence.';
+comment on column user_sequence.owner is e''
+  'Username of the user sequence owner.';
+comment on column user_sequence.updated_at is e''
+  'Time the user sequence was last updated.';
+
+create or replace function user_sequence_set_updated_at()
+returns trigger
+security definer
+language plpgsql as $$begin
+  new.updated_at = now();
+  return new;
+end$$;
+
+create trigger set_timestamp
+before update on user_sequence
+for each row
+execute function user_sequence_set_updated_at();
--- a/deployment/hasura/metadata/databases/AerieCommanding/tables/public_user_sequence.yaml
+++ b/deployment/hasura/metadata/databases/AerieCommanding/tables/public_user_sequence.yaml
@@ -1,0 +1,13 @@
+table:
+  name: user_sequence
+  schema: public
+object_relationships:
+- name: command_dictionary
+  using:
+    manual_configuration:
+      remote_table:
+        schema: public
+        name: command_dictionary
+      insertion_order: null
+      column_mapping:
+        authoring_command_dict_id: id
--- a/command-expansion-server/sql/commanding/init.sql
+++ b/command-expansion-server/sql/commanding/init.sql
@@ -11,5 +11,6 @@ begin;
   \ir tables/activity_instance_commands.sql
   \ir tables/sequence.sql
   \ir tables/sequence_to_simulated_activity.sql
+  \ir tables/user_sequence.sql
 
 end;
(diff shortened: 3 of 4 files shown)


command-expansion-server/sql/commanding/tables/user_sequence.sql
```diff
@@ -0,0 +1,39 @@
+create table user_sequence (
+  command_dict_id integer not null,
```
dyst5422: I think we should mirror our expansion behavior here and have this be the `authoring_command_dict_id`. We don't want to confine sequence usage to a single dictionary, but rather be able to hint to the UI what dictionary to use when loading it.

deployment/hasura/metadata/databases/AerieCommanding/tables/public_user_sequence.yaml
```diff
@@ -0,0 +1,3 @@
+table:
+  name: user_sequence
+  schema: public
```
dyst5422, line 3: We should add in the cross-db association to the command dictionary (not necessary in this PR though)

dyst5422 (changes requested):

goetzrrGit:
I was thinking about this for a couple of minutes and I realized we never talked about user-defined sequences with respect to their sequence ID. On the command expansion side, we make those unique but on the user side, they can create multiple sequences for a sequence ID with no sequence ID checking. 

Use Case 1:

Create a new sequence with seqID A, save - no issue
I want to update the sequence, save - no issue

Use Case 2:

Create a new sequence with seqID A, save - no issue
I create a new sequence with seqID B, save - no issue

Use Case 3:

Create a new sequence with seqID A, save - no issue
I create a brand new sequence with seqID A, save - Problem: two sequences with the same seqID

This can get more complicated as there can be different states a sequence can be in ( Draft, Accepted, Rejected, etc.)


I am not saying we should address this in this ticket but something we should have a meeting about in the future.

dyst5422:
> I was thinking about this for a couple of minutes and I realized we never talked about user-defined sequences with respect to their sequence ID. On the command expansion side, we make those unique but on the user side, they can create multiple sequences for a sequence ID with no sequence ID checking.
> 
> Use Case 1:
> 
> Create a new sequence with seqID A, save - no issue I want to update the sequence, save - no issue
> 
> Use Case 2:
> 
> Create a new sequence with seqID A, save - no issue I create a new sequence with seqID B, save - no issue
> 
> Use Case 3:
> 
> Create a new sequence with seqID A, save - no issue I create a brand new sequence with seqID A, save - Problem: two sequences with the same seqID
> 
> This can get more complicated as there can be different states a sequence can be in ( Draft, Accepted, Rejected, etc.)
> 
> I am not saying we should address this in this ticket but something we should have a meeting about in the future.

Sequence IDs are very specifically NOT something we are managing. That is on the missions to determine. The unique ID for sequences we have are disconnected from the seqid and in fact the seqid isn't necessarily part of what is tracked by AERIE at all. I would think the workflow would haver the seqid somewhere in the `name` field, but ultimately, the seqid is determined by the seqid property on the sequence object in the body of the sequence file.

goetzrrGit:
> > I was thinking about this for a couple of minutes and I realized we never talked about user-defined sequences with respect to their sequence ID. On the command expansion side, we make those unique but on the user side, they can create multiple sequences for a sequence ID with no sequence ID checking.
> > Use Case 1:
> > Create a new sequence with seqID A, save - no issue I want to update the sequence, save - no issue
> > Use Case 2:
> > Create a new sequence with seqID A, save - no issue I create a new sequence with seqID B, save - no issue
> > Use Case 3:
> > Create a new sequence with seqID A, save - no issue I create a brand new sequence with seqID A, save - Problem: two sequences with the same seqID
> > This can get more complicated as there can be different states a sequence can be in ( Draft, Accepted, Rejected, etc.)
> > I am not saying we should address this in this ticket but something we should have a meeting about in the future.
> 
> Sequence IDs are very specifically NOT something we are managing. That is on the missions to determine. The unique ID for sequences we have are disconnected from the seqid and in fact the seqid isn't necessarily part of what is tracked by AERIE at all. I would think the workflow would haver the seqid somewhere in the `name` field, but ultimately, the seqid is determined by the seqid property on the sequence object in the body of the sequence file.

You are right, I forgot we are not managing sequence ID and that a mission should do that. :P Let me close that can of worms and bury it.

command-expansion-server/sql/commanding/tables/user_sequence.sql
goetzrrGit: +1

deployment/hasura/metadata/databases/AerieCommanding/tables/public_user_sequence.yaml
goetzrrGit: Don't forget to create a ticket for this

goetzrrGit (changes requested):
Nothing much to add besides creating a ticket from Dylan's comment.

deployment/hasura/metadata/databases/AerieCommanding/tables/public_user_sequence.yaml
camargo: Agree. This would be just a regular intra-db relationship right since the command_dictionary table exists in the same database as the user_sequence? Happy to add in for this PR since it's simple.

camargo: Done

command-expansion-server/sql/commanding/tables/user_sequence.sql
camargo: Done

dyst5422 (approved):

goetzrrGit (approved):
Looks great!!!

PR #308, opened 2022-08-31
[AERIE-1967] Global Scheduling Conditions

mattdailis:
* **Tickets addressed:** AERIE-1967
* **Review:** By commit  

## Description

Use case: Users want to help the scheduler avoid violating constraints. Currently, the scheduler has no knowledge of constraints (in the merlin sense of "constraints").

While it would be straightforward to give the scheduler access to merlin's constraints, we argue this would not provide much benefit with respect to the use case above. While the scheduler could _check_ that, after placing an activity, it has violated a constraint, it cannot _predict_ which regions of the plan would be safe for placing a given activity. We believe that a "guess and check" approach would be intractable - the scheduler needs some more help from the user.

This "help" that the user can provide the scheduler comes in the form of a "Global Scheduling Condition". This is a piece of typescript code, akin to a goal. Like a goal, it can be added to a scheduling specification, enabled, and disabled. Unlike a goal, it returns a `Windows` object, which uses the Constraints language to establish windows in which the scheduler is allowed to place activities. Also, unlike goals, Global Scheduling Conditions have no notion of priority.

## Verification

This PR has been validated through a rather tedious manual test. If I can get someone to pair with me, I'd love to turn this into an e2e test.
<details>
<summary>Test procedure</summary>

1. Clean build
2. Upload banananation
3. Check mission model id, you'll need it later
4. Create a new plan
5. Check the scheduling spec id, you'll need it later.
6. Create a new Goal
```typescript
export default (): Goal => {
    activityTemplate: ActivityTemplates.BiteBanana({biteSize: 1}),
    interval: 24 * 60 * 60 * 1000 * 1000
  })
}
```
7. Run the following (fill in your model_id):
```graphql
mutation InsertGSC {
  insert_scheduling_condition_one(object:{
    model_id: 1
    name: "My first scheduling condition"
    definition: "export default (): Windows => Real.Resource('/fruit').greaterThan(5.0)"
  }) {
    id
  }
}
```
Note the id returned.
8. Run the following (fill in your spec id and condition id)
```graphql
insert_scheduling_specification_conditions_one(object:{
  condition_id: 1
  specification_id: 1
  enabled: true
}) {
  __typename
}
```
9. Run scheduling, check that it has not placed any activities.
10. Run the following to change the threshold from 5 to 1:
```graphql
mutation UpdateGSC {
  update_scheduling_condition_by_pk(
    pk_columns: {id:1},
    _set: {
      definition: "export default (): Windows => Real.Resource('/fruit').greaterThan(0.0)"
    }
  ) {
    __typename
  }
}
```
11. Run scheduling, check that it has placed activities. (Note that it should stop placing activities when the /fruit resource reaches 0.0)
</details>

## Documentation

- [x] Docs have not been written, but absolutely should be before merging this PR.

UPDATE: I have added a "Global Scheduling Conditions" section to the [scheduling guide](https://github.com/NASA-AMMOS/aerie/wiki/Scheduling-Guide)

## Future work

- This PR lays some groundwork for associating scheduling conditions with particular activity types (e.g. do not place "battery intensive" activities when the battery is low, but it's okay to place a "recharge" activity)

### DIFF
--- a/scheduler-server/scheduling-dsl-compiler/src/main.ts
+++ b/scheduler-server/scheduling-dsl-compiler/src/main.ts
@@ -1,8 +1,8 @@
 import fs from 'fs';
-import * as readline from 'readline';
 import ts from 'typescript';
-import { UserCodeRunner } from '@nasa-jpl/aerie-ts-user-code-runner';
-import type { Goal } from './libs/scheduler-edsl-fluent-api.js';
+import {UserCodeRunner} from '@nasa-jpl/aerie-ts-user-code-runner';
+import type {Goal} from './libs/scheduler-edsl-fluent-api.js';
+import * as readline from 'readline';
 
 const codeRunner = new UserCodeRunner();
 const schedulerEDSL = fs.readFileSync(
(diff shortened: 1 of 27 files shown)


scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/GlobalConstraint.java
```diff
@@ -8,12 +8,12 @@
  * Abstract class defining methods that must be implemented by global constraints such as mutex or cardinality
```
mattdailis: TODO: change comment to say interface

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/scheduling/GlobalConstraintWithIntrospection.java
```diff
@@ -9,11 +9,11 @@
  * Abstract class defining methods that must be implemented by global constraints such as mutex or cardinality
```
mattdailis: TODO: change comment to say interface

scheduler-server/scheduling-dsl-compiler/src/main.ts
```diff
@@ -25,80 +25,69 @@ const tsConfig = JSON.parse(fs.readFileSync(new URL('../tsconfig.json', import.m
...
+
+async function handleRequest(data: Buffer) {
   try {
-    const { goalCode, schedulerGeneratedCode, constraintsGeneratedCode } = JSON.parse(data) as { goalCode: string, schedulerGeneratedCode: string, constraintsGeneratedCode: string };
+    // Test the health of the service by responding to "ping" with "pong".
+    if (data.toString() === 'ping') {
+      process.stdout.write('pong\n');
+      lineReader.once('line', handleRequest);
+      return;
+    }
```
goetzrrGit, line 51: More of a general question. Is this the only service in AERIE doing a ping pong check? Is there a hook somewhere in the UI that will notify the user if a service is down?

goetzrrGit (approved):
The walkthrough helped a lot. Everything looks good. Just a more general question below

scheduler-server/scheduling-dsl-compiler/src/main.ts
mattdailis: Good question - this check is only between the Java process and the node subprocess (so the UI can't send a "ping" message). The check is mostly supposed to be a "fail fast" mechanism, so that if the node subprocess is misconfigured, we find out at startup time rather than waiting for the first compilation request to come in.

Now, unfortunately, if the subprocess is unresponsive, the whole application hangs, since we haven't figured out how to set a timeout.

dyst5422: This is at least partly handled by the user code runner having a timeout, but the process as a whole does not.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
```diff
@@ -100,17 +101,14 @@ export default () => Goal.ActivityRecurrenceGoal({
...
-                }),
-                interval: -25 // one day in microseconds
-              })
-              """, true)));
+      final var results = runScheduler(BANANANATION, List.of(), List.of(new SchedulingGoal(new GoalId(0L), """
+          export default () => Goal.ActivityRecurrenceGoal({
+            activityTemplate: ActivityTemplates.PeelBanana({
+              peelDirection: "fromStem",
+            }),
+            interval: -25 // one day in microseconds
```
patkenneally: :large_blue_circle: Is this comment still correct?

scheduler-server/sql/scheduler/tables/scheduling_condition.sql
```diff
@@ -0,0 +1,54 @@
+create table scheduling_condition (
+  id integer generated always as identity,
+  revision integer not null default 0,
```
patkenneally, line 3: :large_blue_circle: I've looked through the code base (albeit quickly) and don't see that we use the revision field here, or on the scheduling goals. Is it used somewhere I'm not seeing (likely)? Or perhaps is it being included with an eye to a future capability that I'm not seeing?

patkenneally (approved):

scheduler-server/sql/scheduler/tables/scheduling_condition.sql
mattdailis: You're right - we don't currently use the `revision` field on scheduling goals or conditions - but it is auto-updated via trigger, and it could be useful (in the future) to know that a particular scheduling run used an older version of one of these inputs. Additionally, users can use these values via the API to track these revisions themselves, if they like. So - I'm going to go with:

> it being included with an eye to a future capability

even though if I'm being honest, I didn't think about it until you asked.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/services/SchedulingIntegrationTests.java
mattdailis: It is not correct, thank you. I have updated it.

mattdailis:
Update: I have added a "Global Scheduling Conditions" section to the [scheduling guide](https://github.com/NASA-AMMOS/aerie/wiki/Scheduling-Guide)

PR #309, opened 2022-09-01
[REFACTOR] Try with resource refactoring in database actions

patkenneally:
* **Tickets addressed:** Try-with-resource refactoring in database actions
* **Review:** By commit  

## Description
In the interest of hunting down places where we could be leaking resources, I did a pass (guided by a set of SonarLint results) over the database actions package. I also did a number of simple clean ups in the first set of commits. The first six commits are light clean up, and the last two commits contain the try-with-resources changes. Eager to hear where I may have gotten it wrong/there are better options. 

## Verification
Test suites (unit and e2e) were run. Also deployed, upload mission model, create plan, run sim, check constraints.

## Documentation
No new documentation needed and no changes needed.

## Future work
None

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
@@ -44,33 +44,34 @@ import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
     PreparedStatements.setTimestamp(this.statement, 2, simulationStart);
     PreparedStatements.setTimestamp(this.statement, 3, planStart);
 
-    final var results = this.statement.executeQuery();
-    if (!results.next()) throw new FailedInsertException("simulation_dataset");
-    final Status status;
-    try {
-      status = Status.fromString(results.getString(2));
-    } catch (final Status.InvalidSimulationStatusException ex) {
-      throw new Error("Simulation Dataset initialized with invalid state.");
+    try (final var results = this.statement.executeQuery()) {
+      if (!results.next()) throw new FailedInsertException("simulation_dataset");
+      final Status status;
+      try {
+        status = Status.fromString(results.getString(2));
+      } catch (final Status.InvalidSimulationStatusException ex) {
+        throw new Error("Simulation Dataset initialized with invalid state.");
+      }
+
+      final var datasetId = results.getLong(1);
+      final var reason = results.getString(3);
+
+      final var state = new SimulationStateRecord(
+          status,
+          reason);
+      final var canceled = results.getBoolean(4);
+
+      final var simulationDatasetId = results.getLong(5);
+
+      return new SimulationDatasetRecord(
+          simulationId,
+          datasetId,
+          state,
+          canceled,
+          offsetFromPlanStart,
+          simulationDatasetId
+      );
     }
-
-    final var datasetId = results.getLong(1);
-    final var reason = results.getString(3);
-
-    final var state = new SimulationStateRecord(
-        status,
-        reason);
-    final var canceled = results.getBoolean(4);
-
-    final var simulationDatasetId = results.getLong(5);
-
-    return new SimulationDatasetRecord(
-        simulationId,
-        datasetId,
-        state,
-        canceled,
-        offsetFromPlanStart,
-        simulationDatasetId
-    );
   }
 
   @Override
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/EventGraphUnflattener.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/EventGraphUnflattener.java
@@ -69,7 +69,7 @@ public final class EventGraphUnflattener {
           self.children.put(tag.index(), Trie.of(tag.suffix(), event));
         }
       } else {
-        throw new IllegalArgumentException("Unexpected variant %s of type $s".formatted(this, Trie.class));
+        throw new IllegalArgumentException("Unexpected variant %s of type %s".formatted(this, Trie.class));
       }
     }
 
@@ -83,7 +83,7 @@ public final class EventGraphUnflattener {
         }
         return graph;
       } else {
-        throw new IllegalArgumentException("Unexpected variant %s of type $s".formatted(this, Trie.class));
+        throw new IllegalArgumentException("Unexpected variant %s of type %s".formatted(this, Trie.class));
       }
     }
 
@@ -97,7 +97,7 @@ public final class EventGraphUnflattener {
         }
         return graph;
       } else {
-        throw new IllegalArgumentException("Unexpected variant %s of type $s".formatted(this, Trie.class));
+        throw new IllegalArgumentException("Unexpected variant %s of type %s".formatted(this, Trie.class));
       }
     }
 
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/EventGraphUnflattener.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/EventGraphUnflattener.java
@@ -3,7 +3,6 @@ package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
 import gov.nasa.jpl.aerie.merlin.driver.timeline.EventGraph;
 import org.apache.commons.lang3.tuple.Pair;
 
-import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
@@ -50,12 +50,14 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.
   }
 
   private Map<String, SerializedValue> parseActivityArguments(final Reader stream) {
-    final var json = Json.createReader(stream).readValue();
-    return activityArgumentsP
-        .parse(json)
-        .getSuccessOrThrow(
-            failureReason -> new Error("Corrupt activity arguments cannot be parsed: " + failureReason.reason())
-        );
+    try(final var reader = Json.createReader(stream)) {
+      final var json = reader.readValue();
+      return activityArgumentsP
+          .parse(json)
+          .getSuccessOrThrow(
+              failureReason -> new Error("Corrupt activity arguments cannot be parsed: " + failureReason.reason())
+          );
+    }
   }
 
   @Override
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
@@ -1,18 +1,19 @@
 package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
 
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.json.JsonParseResult;
+import gov.nasa.jpl.aerie.json.JsonParser;
 import org.intellij.lang.annotations.Language;
 
 import javax.json.Json;
-import java.io.Reader;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
+import java.sql.ResultSet;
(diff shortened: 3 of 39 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/EventGraphUnflattener.java
```diff
@@ -69,7 +69,7 @@ default void add(Optional<Tag> tag$, T event) throws CollisionException {
           self.children.put(tag.index(), Trie.of(tag.suffix(), event));
         }
       } else {
-        throw new IllegalArgumentException("Unexpected variant %s of type $s".formatted(this, Trie.class));
+        throw new IllegalArgumentException("Unexpected variant %s of type %s".formatted(this, Trie.class));
```
Twisol, line 72: Oops. IntelliJ even linted this as a problem. Sorry.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
```diff
@@ -50,12 +50,14 @@ public List<ActivityInstanceRecord> get(final long planId) throws SQLException {
   }
 
   private Map<String, SerializedValue> parseActivityArguments(final Reader stream) {
-    final var json = Json.createReader(stream).readValue();
-    return activityArgumentsP
-        .parse(json)
-        .getSuccessOrThrow(
-            failureReason -> new Error("Corrupt activity arguments cannot be parsed: " + failureReason.reason())
-        );
+    try(final var reader = Json.createReader(stream)) {
```
Twisol: According to the JSON-P Javadocs for [`JsonReader#close`](https://javadoc.io/static/javax.json/javax.json-api/1.1.4/javax/json/JsonReader.html#close--),

> This method closes the underlying input source.

The underlying input source here is the `stream` argument, which comes from the caller. Although this is a private helper method, it still feels weird to construct part of the reader pipeline at the caller and part of it here, and then take ownership of the whole thing to close it here.

I'd actually suggest changing the signature of `parseActivityArguments` to `(ResultSet results, String key)`, and moving the `getCharacterStream` call inside this method.

```java
private static Map<String, SerializedValue>
parseActivityArguments(final ResultSet results, final String column) throws SQLException {
  try (final var reader = Json.createReader(results.getCharacterStream(column))) {
    // ...
  }
}
```

Twisol: For bonus points, we can generalize what `parseActivityArguments` is doing into a general-purpose helper (like those on `PreparedStatements`) for reading structured data out of a JSON column.

```java
private static <V> JsonParseResult<V>
getJsonColumn(final ResultSet results, final String column, final JsonParser<V> parser) throws SQLException {
  try (final var reader = Json.createReader(results.getCharacterStream(column))) {
    return parser.parse(reader.readValue());
  }
}
```

Usage:

```java
        activities.add(new ActivityInstanceRecord(
            results.getLong("id"),
            results.getString("type"),
            results.getLong("start_offset_in_micros"),
            getJsonColumn(results, "arguments", activityArgumentsP)
                .getSuccessOrThrow($ -> new Error("Corrupt activity arguments cannot be parsed: " + $.reason()))
        ));
```

This has the extra features of (a) consolidating this pattern of resource handling into one consistent place, and (b) moving the error message into the more domain-relevant block of code.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityTypesAction.java
```diff
@@ -58,40 +58,49 @@ public List<ActivityType> get(final long missionModelId) throws SQLException {
...
-            failureReason -> new Error("Corrupt activity type parameters cannot be parsed: " + failureReason.reason())
-        );
-    return parametersMap
-        .entrySet()
-        .stream()
-        .map(entry -> new ParameterRecord(entry.getKey(), entry.getValue().getKey(), entry.getValue().getValue()))
-        .sorted(Comparator.comparingInt(ParameterRecord::order))
-        .map(record -> new Parameter(record.name(), record.schema()))
-        .toList();
+    try(final var reader = Json.createReader(stream)) {
```
Twisol: Same recommendation for this file as for the other file. In particular, if we have the generic `getJsonColumn` I mentioned, all three of these helper methods can go away / be inlined into the call site:

```java
        activityTypes.add(new ActivityType(
            results.getString("name"),
            getJsonColumn(results, "parameters", parameterRecordP)
                .getSuccessOrThrow($ -> new Error("Corrupt activity type parameters cannot be parsed: " + $.reason()))
                .entrySet()
                .stream()
                .map(entry -> new ParameterRecord(entry.getKey(), entry.getValue().getKey(), entry.getValue().getValue()))
                .sorted(Comparator.comparingInt(ParameterRecord::order))
                .map(record -> new Parameter(record.name(), record.schema()))
                .toList(),
            getJsonColumn(results, "required_parameters", listP(stringP))
                .getSuccessOrThrow($ -> new Error("Corrupt activity type required parameters cannot be parsed: " + $.reason())),
            getJsonColumn(results, "computed_attributes_value_schema", valueSchemaP)
                .getSuccessOrThrow($ -> new Error("Corrupt activity type computed attribute schema cannot be parsed: " + $.reason()))
        ));
```

Only the `parameterRecordP` needs to be defined as a static field on the class.

```java
  private static final JsonParser<Map<String, Pair<Integer, ValueSchema>>> parameterRecordP =
      mapP(
          productP
              .field("order", intP)
              .field("schema", valueSchemaP));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
```diff
@@ -326,8 +327,8 @@ private static PlanDatasetRecord associatePlanWithDataset(
   private static <T> T parseJson(final String subject, final JsonParser<T> parser)
   throws InvalidJsonException, InvalidEntityException
   {
-    try {
-      final var requestJson = Json.createReader(new StringReader(subject)).readValue();
+    try (JsonReader reader = Json.createReader(new StringReader(subject))) {
```
Twisol, line 330: I think this is actually a no-op, as `JsonReader` simply closes the underlying reader, here a `StringReader`, which itself has an essentially no-op `close` operation.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
```diff
@@ -44,33 +44,34 @@ public SimulationDatasetRecord apply(
...
 
-    final var results = this.statement.executeQuery();
-    if (!results.next()) throw new FailedInsertException("simulation_dataset");
-    final Status status;
-    try {
-      status = Status.fromString(results.getString(2));
-    } catch (final Status.InvalidSimulationStatusException ex) {
-      throw new Error("Simulation Dataset initialized with invalid state.");
-    }
+    try(final var results = this.statement.executeQuery()) {
```
Twisol: This change is fine, but note from the `ResultSet` javadocs:

> A ResultSet object is automatically closed when the Statement object that generated it is closed, re-executed, or used to retrieve the next result from a sequence of multiple results.

So this is at worst a *bounded* leak, since the next call to `apply` will automatically release the previously-generated `ResultSet`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationTopicsAction.java
```diff
@@ -35,25 +35,25 @@ public List<Triple<Integer, String, ValueSchema>> get(
...
-          )
-      );
+    try (final var resultSet = this.statement.executeQuery()) {
+      final var topics = new ArrayList<Triple<Integer, String, ValueSchema>>();
+      while (resultSet.next()) {
+        topics.add(
+            Triple.of(
+                resultSet.getInt(1),
+                resultSet.getString(2),
+                parseValueSchema(resultSet.getString(3))
```
Twisol: This action doesn't follow the pattern established by all of the other actions, which use `getCharacterStream` instead of `getString`. I know it's not the fault of this PR, but I'd recommend changing this to use `getCharacterStream` in line with everything else.

Twisol (approved):
Looks great! I have a few comments that go a bit beyond the leak-fixing purpose of this PR, but since we're in the area, it might be good to make some minor tweaks in aid of the next dev to come by this code.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
patkenneally: Voodoo. Done.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityTypesAction.java
patkenneally: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
Twisol: From judo to voodoo... 👀

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
```diff
@@ -67,12 +67,14 @@ public <Dynamics> List<Pair<Duration, Dynamics>> get(
   }
 
   private <Dynamics> Dynamics parseDynamics(final Reader jsonStream, final JsonParser<Dynamics> dynamicsP) {
-    final var json = Json.createReader(jsonStream).readValue();
-    return dynamicsP
-        .parse(json)
-        .getSuccessOrThrow(
-            failureReason -> new Error(
-                "Corrupt profile dynamics: " + failureReason.reason()));
+    try(final var reader = Json.createReader(jsonStream)) {
```
Twisol: FWIW, there are a few other places in 07b414e35fae1d0baf2edd44ff17240adfe7f3ec that my earlier comment on closing a received reader applies to, and they would all benefit from the new `getJsonColumn`. I could comment on each if you like, but they're all visible at that commit.

Twisol:
@patkenneally, your commits are not associated with your GitHub username. You should double-check that the email you use with Git (`git config --global user.email`) is associated to your GitHub account.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
patkenneally: True. Thank you 🙏 . Change made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationTopicsAction.java
patkenneally: Made redundant by refactoring with the new getJsonColumn method you suggested.

Twisol: 🚀

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateSimulationDatasetAction.java
patkenneally: Astute. Happy to undo if we think there's value in say readability or something like that.

Twisol: No, I think this is fine. More noting that, if you're looking for leaks that eventually lead to destabilization of the system, this probably isn't one. But it's good to address anyway.

patkenneally: 👍

pcrosemurgy (approved):
Thanks Pat! These changes look sane.

Some `try` statements need a space between the `try` and leading `(`. Ugly one-liner to fix this if you don't mind:
```
egrep -rn . -e ' try\(' --include \*.java -l | xargs perl -pi -e 's/try\(/try \(/g'
```

patkenneally:
Thank you @pcrosemurgy ! Also thank you for the clever snippet there!

PR #311, opened 2022-09-01
AERIE-2041 - TS convert functional programing UNIT to {}

goetzrrGit:
* **Tickets addressed:** AERIE-2041
* **Review:** By commit  

## Description
In the mission model, the activity's effect model that doesn't return a computed attribute (aka void run() method of the activity) uses UNIT which represents a non-null empty value. This is used internally by AERIE

https://en.wikipedia.org/wiki/Unit_type

This UNIT value will be exposed to our users when **command authoring**. Our users are not functional programmers so this causes confusion.

This will change the computed attribute to `{}` which makes more sense to the user.

## Verification
Ran the command expansion test locally without any issues

## Documentation
None

## Future work
None

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/rulesets/BasicValueMappers.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/rulesets/BasicValueMappers.java
@@ -23,14 +23,21 @@ import gov.nasa.jpl.aerie.contrib.serialization.mappers.PrimitiveLongArrayValueM
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.PrimitiveShortArrayValueMapper;
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.ShortValueMapper;
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.StringValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.UnitValueMapper;
 import gov.nasa.jpl.aerie.merlin.framework.ValueMapper;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Unit;
 
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
 public final class BasicValueMappers {
+
+  // Unit is an enum, so `$unit` needs to be defined before `$enum()`
+  // in order to override the latter's representation.
+  public static ValueMapper<Unit> $unit() { return new UnitValueMapper(); }
+
   public static ValueMapper<Boolean> $boolean() {
     return new BooleanValueMapper();
   }
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
@@ -10,7 +10,6 @@ import com.squareup.javapoet.ParameterSpec;
 import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
-import gov.nasa.jpl.aerie.contrib.serialization.mappers.EnumValueMapper;
 import gov.nasa.jpl.aerie.merlin.framework.ModelActions;
 import gov.nasa.jpl.aerie.merlin.framework.RootModel;
 import gov.nasa.jpl.aerie.merlin.framework.Scoped;
@@ -18,6 +17,7 @@ import gov.nasa.jpl.aerie.merlin.framework.Scoping;
 import gov.nasa.jpl.aerie.merlin.framework.ValueMapper;
 import gov.nasa.jpl.aerie.merlin.processor.MissionModelProcessor;
 import gov.nasa.jpl.aerie.merlin.processor.Resolver;
+import gov.nasa.jpl.aerie.merlin.processor.TypePattern;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.ActivityTypeRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.ConfigurationTypeRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.EffectModelRecord;
@@ -812,7 +812,10 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
       }
       computedAttributesTypeName = TypeName.get(typeMirror.get());
     } else {
-      effectModelReturnMapperBlock = Optional.of(CodeBlock.of("new $T($T.class)", EnumValueMapper.class, Unit.class));
+      effectModelReturnMapperBlock = new Resolver(
+          this.typeUtils, this.elementUtils, missionModel.typeRules).applyRules(
+          new TypePattern.ClassPattern(ClassName.get(ValueMapper.class),
+                                       List.of(new TypePattern.ClassPattern(ClassName.get(Unit.class), List.of()))));
       computedAttributesTypeName = TypeName.get(Unit.class);
     }
     return Optional.of(new ComputedAttributesCodeBlocks(
--- a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
+++ b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
@@ -1,13 +1,12 @@
 package gov.nasa.jpl.aerie.merlin.server.models;
 
-import gov.nasa.jpl.aerie.contrib.serialization.mappers.EnumValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.UnitValueMapper;
 import gov.nasa.jpl.aerie.foomissionmodel.generated.GeneratedMissionModelFactory;
 import gov.nasa.jpl.aerie.merlin.driver.DirectiveTypeRegistry;
 import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
 import gov.nasa.jpl.aerie.merlin.protocol.types.InvalidArgumentsException;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Parameter;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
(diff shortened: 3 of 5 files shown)


command-expansion-server/src/lib/codegen/ActivityTypescriptCodegen.ts
```diff
@@ -53,8 +53,12 @@ function convertSchemaType(schema: Schema): string {
           .join('\n'),
       );
     case SchemaTypes.Variant:
-      if (schema.variants.length === 1 && schema.variants[0]?.key === 'VOID') {
-        return 'null';
+      if (
+        schema.variants.length === 1 &&
+        (schema.variants[0]?.key === 'VOID' ||
+          (schema.variants[0]?.key === 'UNIT' && schema.variants[0]?.label === 'UNIT'))
```
Twisol: Would you see any downsides to removing the specialization to the particular keys `VOID` or `UNIT`? The latter in particular is automatically derived from the name of a Java identifier, and it makes me uncomfortable to imagine that refactoring a name might cause a user-visible change (and a bug, at that).

Alternatively, should we consider adding a new type of `ValueSchema` that explicitly denotes the lack of a useful value? That would make this situation easily detectable without causing all one-variant enums to be treated as trivial.

Twisol: *Alternatively* alternatively, perhaps we represent UNIT as an empty struct instead of a one-variant enum? Then we have a very close match indeed to the representation here as `{}`.

Maybe that would mean we don't have to change anything at all in the typescript, because I imagine it *already* turns empty maps into `{}`.

I think I like this the best. Thoughts?

Mythicaeda: I was also wondering why we need to check length and key. Can we have a schema with a `key` and `label` of `UNIT` but not a `length` of 1? If not, then I don't think it ought to be `&&`'d with the length.

Twisol: @Mythicaeda, in principle, nothing's stopping a modeler from writing an `enum Foo { UNIT; AAA; BBB }`, where it has three variants but one of them happens to be called "UNIT". (Another reason I don't like the name-sniffing.)

Twisol: For demonstration, I pushed a branch [`feature/unit-serialization`](https://github.com/NASA-AMMOS/aerie/commit/0ebb9928f6ef5eb983c513bbc3e8200b148e7200). This should override the representation of `Unit.UNIT` so that it is represented as `{}` from the very beginning, and so ostensibly should not require any changes downstream here. Please do share your thoughts 😃

Mythicaeda: Does `{}` have a length of 0 or 1?

Twisol: Zero. `Unit` is currently serialized as a list of variants (of length 1); I'm proposing serializing it as a struct with no fields (so 0).

But the code branch this comment is on only applies to variants, not structs, so this would push `Unit` into a different branch entirely. The relevant preexisting case is copied here:
```ts
    case SchemaTypes.Struct:
      return objectDeclaration(
        Object.entries(schema.items)
          .map(([key, value]) => `${key}: ${convertSchemaType(value)};`)
          .join('\n'),
      );
```
Since `items` will be empty, this should produce an empty object `{}` without any further effort on our part.

goetzrrGit: Hi Jonathan,

Thank you for all the feedback. I do like your idea of a value mapper that changes it upstream. I wasn't sure how to hook up the value mapper and your branch helps a lot. It is cleaner than my approach. Let me play around with the code as I want to learn a little bit more about the valuemappers. I haven't really touched that part of AERIE.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/rulesets/BasicValueMappers.java
```diff
@@ -23,14 +23,19 @@
...
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Unit;
 
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 
 public final class BasicValueMappers {
+
+  public static ValueMapper<Unit> $unit() { return new UnitValueMapper(); }
```
Twisol, line 39: I would highly recommend retaining a comment on this method explaining how its declaration order relative to `$enum()` is critically important. Declaration order in Java is not *usually* semantically relevant, but it definitely is here.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -812,7 +812,7 @@ private Optional<ComputedAttributesCodeBlocks> getComputedAttributesCodeBlocks(
       }
       computedAttributesTypeName = TypeName.get(typeMirror.get());
     } else {
-      effectModelReturnMapperBlock = Optional.of(CodeBlock.of("new $T($T.class)", EnumValueMapper.class, Unit.class));
+      effectModelReturnMapperBlock = Optional.of(CodeBlock.of("$T.$L", BasicValueMappers.class,"$unit()"));
```
Twisol: The fact that this change exists suggests to me that we should use `Resolver` in this branch, too, so that we're *always* consistent with explicitly-declared types.

```java
      effectModelReturnMapperBlock = new Resolver(this.typeUtils, this.elementUtils, missionModel.typeRules)
          .applyRules(new TypePattern.ClassPattern(ClassName.get(Unit.class), List.of()));
```

We do need to make `applyRules` public, but I think that's acceptable.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```diff
@@ -47,18 +46,20 @@ public void shouldGetActivityTypeList() {
...
-        assertThat(typeList).containsAllEntriesOf(expectedTypes);
+        registry.taskSpecTypes().forEach((name, specType) -> activityTypes.put(
+            name,
+            new ActivityType(name,
+                             specType.getParameters(),
+                             specType.getRequiredParameters(),
+                             specType.getReturnValueSchema())));
+
+      // THEN
+        assertThat(activityTypes).containsAllEntriesOf(expectedTypes);
```
Twisol, line 62: Slightly wonky indentation here. (Also, this block is a pure-whitespace change, so I had to squint to make sure it wasn't actually changing anything amidst the other changes.)

Twisol (commented):
(I might be jumping the gun a bit on this review; sorry if so.)

goetzrrGit:
I learned a lot from this ticket. Now I found the code-gen part of Merlin. This gets propagated down to the command expansion part now and will show up as:

```
declare global {
	interface ThrowBanana {
		readonly type: 'ThrowBanana';
		readonly duration: Temporal.Duration;
		readonly startOffset: Temporal.Duration;
		readonly startTime: Temporal.Instant;
		readonly endTime: Temporal.Instant;
		readonly attributes: {
			readonly arguments: {
				readonly speed: number;
			}
			readonly computed: {
			};
		}
	}
	type ActivityType = ThrowBanana;
}

export {};
```

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
goetzrrGit: I also changed `-> { activityTypes` to `-> activityTypes` but very tiny refactor let me change that weird whitespace issue.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/rulesets/BasicValueMappers.java
goetzrrGit: Sure I can add the comment back. Sorry, I thought that was more for me to understand better. :P

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
goetzrrGit: I also removed some unused imports :P

Twisol: Ah, true. I guess I'd consider that a formatting change (so, still semantically inert), but you're right that it's not whitespace-only. Thanks for highlighting that extra change for me.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
goetzrrGit: Let me look into that.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/rulesets/BasicValueMappers.java
Twisol: Why not both? ^_^

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
Twisol: > I also removed some unused imports :P

Yeah, that's fine, I was focusing on the small block of changes around the `// THEN` in this comment :)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
goetzrrGit: Got this working now with the below code thanks for all the help: 

```
effectModelReturnMapperBlock = new Resolver(
          this.typeUtils, this.elementUtils, missionModel.typeRules).applyRules(
          new TypePattern.ClassPattern(ClassName.get(ValueMapper.class),
                                       List.of(new TypePattern.ClassPattern(ClassName.get(Unit.class), List.of()))));
```

Twisol (approved):

patkenneally (approved):

PR #312, opened 2022-09-02
[AERIE-2082] Use unconfigured models where possible

mattdailis:
* **Tickets addressed:** AERIE-2082
* **Review:** By commit  

## Description

This PR is the result of a discussion we had today about which actions need "fully instantiated" mission models, and which actions only need "loaded mission models". The terminology I found in the code used the words "Configured Mission Model" to mean one whose constructor had been called, and "Unconfigured Mission Model" for mission models whose factories had been loaded, and directive types extracted, but no constructor had been called.

The strategy was to make use of `loadUnconfiguredMissionModel` wherever possible (everywhere except for `getResourceSchemas`). I added new methods to the `MissionModelFacade.Unconfigured` class, since they seemed to group well with the existing methods there.

## Verification

~~In progress - first, let's see if the e2e tests pass on this PR, next I'll spin up a local aerie instance and play around with the UI and make sure all of these actions are functioning properly.~~

e2e tests have passed, and a cursory playing around in the UI seemed to work okay.

## Documentation

Something we should document is the relationship that Aerie has with global state - how and when is it acceptable to read or write to global state?

## Future work

- Move resource declaration out of the mission model constructor, so that `getResourceSchemas` can make use of the unconfigured model as well


Twisol:
> Something we should document is the relationship that Aerie has with global state - how and when is it acceptable to read or write to global state?

At least in principle, simulations are not to have mutable state *at all*. Any state that needs to be mutated as simulation proceeds must be tracked in cells, and any other state must be fixed at instantiation time and scoped to that particular model instance. I believe this is documented... somewhere... in the mission modeler guide.

This applies equally to SPICE, with the caveat that since SPICE is global, it has to be managed even more carefully. Global state has the capacity to persist beyond the lifetime of a single model instance, and may feed across into other models instantiated alongside it or later on. I'm very reluctant to provide strict guarantees on the behavior and availability of global state, but due to the prevalence and necessity of SPICE, we do have to provide *some* answers -- or at least a documented safe pattern of use.

Twisol:
> * Move resource declaration out of the mission model constructor, so that `getResourceSchemas` can make use of the unconfigured model as well

I have thoughts on what this could look like, but I don't see a way to do it that won't require changes to how mission models declare resources now. Gavin was aware of this line of attack, but I don't know what kinds of hedging the team may have in place for when it actually happens.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -70,7 +70,7 @@ public Map<String, Constraint> getConstraints(final String missionModelId) throw
   }
 
   @Override
-  public Map<String, ValueSchema> getStatesSchemas(final String missionModelId)
+  public Map<String, ValueSchema> getResourceSchemas(final String missionModelId)
```
Twisol, line 77: _Thank you._

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/ConfigurationType.java
```diff
@@ -26,4 +26,28 @@ public UnconstructableConfigurationException(final String message) {
       super(message);
     }
   }
+
+  /**
+   * This method must behave as though implemented as:
```
Twisol, line 31: Noted during our huddle, but I'll come through and javadoc all this up separately. All that's here now is the equivalent of "you'd better not re-implement this method".

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -130,7 +132,19 @@ public List<String> validateActivityArguments(final String missionModelId, final
   public Map<ActivityInstanceId, String> validateActivityInstantiations(final String missionModelId, final Map<ActivityInstanceId, SerializedActivity> activities)
   throws NoSuchMissionModelException, MissionModelLoadException
   {
-    return this.loadConfiguredMissionModel(missionModelId).validateActivityInstantiations(activities);
+    final var failures = new HashMap<ActivityInstanceId, String>();
```
Twisol, line 145: `MissionModelFacade#validateActivityInstantiations` is no longer used anywhere. Can we remove it? (The `MissionModelFacade` class is kind of a façade on top of a façade, so I'm eager to collapse some of this extra layering if we can get away with it.)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -113,7 +114,8 @@ public List<String> validateActivityArguments(final String missionModelId, final
   {
     try {
       // TODO: [AERIE-1516] Teardown the missionModel after use to release any system resources (e.g. threads).
-      return this.loadConfiguredMissionModel(missionModelId).validateActivity(activity);
+      return this.loadUnconfiguredMissionModel(missionModelId)
+                 .validateActivityArguments(activity);
```
Twisol: `MissionModelFacade#validateActivityInstantiations` is no longer used anywhere except in tests. Can we remove it, and update the tests to use the unconfigured variant you've added here?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -130,7 +132,19 @@ public List<String> validateActivityArguments(final String missionModelId, final
...
   throws NoSuchMissionModelException, MissionModelLoadException
   {
-    return this.loadConfiguredMissionModel(missionModelId).validateActivityInstantiations(activities);
+    final var failures = new HashMap<ActivityInstanceId, String>();
+
+    for (final var entry : activities.entrySet()) {
+      final var id = entry.getKey();
+      final var act = entry.getValue();
+      try {
+        getActivityEffectiveArguments(missionModelId, act); // The return value is intentionally ignored - we are only interested in failures
```
Twisol: Add `this.` for consistency?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -141,10 +155,10 @@ public Map<String, SerializedValue> getActivityEffectiveArguments(final String m
          InvalidArgumentsException
   {
     try {
-      return this.loadConfiguredMissionModel(missionModelId)
-                 .getActivityEffectiveArguments(activity.getTypeName(), activity.getArguments());
-    } catch (final MissionModelFacade.NoSuchActivityTypeException ex) {
-      throw new NoSuchActivityTypeException(activity.getTypeName(), ex);
+      return this.loadUnconfiguredMissionModel(missionModelId)
+                 .getActivityEffectiveArguments(activity);
```
Twisol: `MissionModelFacade#getActivityEffectiveArguments` is no longer used (except by other dead code starting at `validateActivityInstantiations`). Can we remove it?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -154,15 +168,17 @@ public List<String> validateModelArguments(final String missionModelId, final Ma
          MissionModelLoadException,
          InvalidArgumentsException
   {
-    return this.loadConfiguredMissionModel(missionModelId)
-               .validateConfiguration(arguments);
+    return this.loadUnconfiguredMissionModel(missionModelId)
+               .validateMissionModelArguments(arguments);
```
Twisol: `MissionModelFacade#validateConfiguration` is no longer used anywhere. Can we remove it?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -154,15 +168,17 @@ public List<String> validateModelArguments(final String missionModelId, final Ma
...
   }
 
+
   @Override
   public List<Parameter> getModelParameters(final String missionModelId)
   throws NoSuchMissionModelException, MissionModelLoadException
   {
-    return loadUnconfiguredMissionModel(missionModelId).getParameters();
+    return this.loadUnconfiguredMissionModel(missionModelId)
+               .getParameters();
```
Twisol: Just to confirm, I think this is only a whitespace change, yes?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -171,9 +187,9 @@ public Map<String, SerializedValue> getModelEffectiveArguments(final String miss
          MissionModelLoadException,
          InvalidArgumentsException
   {
-    return this.loadConfiguredMissionModel(missionModelId)
-               .getEffectiveArguments(arguments);
```
Twisol, line 175: `MissionModelFacade#getEffectiveArguments` is no longer used anywhere. Can we remove it?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -213,9 +229,8 @@ public void refreshActivityTypes(final String missionModelId)
   throws NoSuchMissionModelException
   {
     try {
-      final var activityTypes =
-          loadUnconfiguredMissionModel(missionModelId)
-              .getActivityTypes();
+      final var activityTypes = this.loadUnconfiguredMissionModel(missionModelId)
+                                    .getActivityTypes();
```
Twisol: This is also just a whitespace change, yes?

Twisol (commented):
Comments mostly surrounding dead code. I think it is architecturally *very telling* that most if not all of the methods of `MissionModelFacade` are used in only one place, and I'd really like to take a crack at flattening it out altogether. (But maybe not totally in this PR. Just the particular code paths we're making dead.)

mattdailis:
Latest force push:
- Splits out style tweaks into a separate commit
- Adds additional commits to remove dead code from the MissionModelFacade (which required updating three tests)
- I noticed I had missed the `MissionModelFacade::getStatesSchemas` method, so the first commit has been updated to rename that as well

EDIT: One more force push to add `this.` to `getEffectiveArguments`, per Jonathan's request

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Removed :+1:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Tests have been updated, and method has been removed :+1:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Method has been removed

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Removed

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Yes, my bad - I've gone back and split out whitespace changes into their own commit

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Removed

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Yes - see commit called "Minor style tweaks to LocalMissionModelService" - I'd also be okay with dropping that commit if you don't think it adds much value.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -141,8 +155,8 @@ public Map<String, SerializedValue> getActivityEffectiveArguments(final String m
          InvalidArgumentsException
   {
     try {
-      return this.loadConfiguredMissionModel(missionModelId)
-                 .getActivityEffectiveArguments(activity.getTypeName(), activity.getArguments());
+      return this.loadUnconfiguredMissionModel(missionModelId)
+                 .getActivityEffectiveArguments(activity);
     } catch (final MissionModelFacade.NoSuchActivityTypeException ex) {
```
Mythicaeda: I noticed that we have two versions of `NoSuchActivityTypeException`s, one from `MissionModelFacade` (seen here), and one in `MissionModelService` (seen in the `validateActivityInstantions` above). The behavior in both seem to be approximately the same, and neither depend on anything from the class/interface they reside in. Not this PR, but we should talk about potentially pulling these out to some kind of top-level exception class.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -154,15 +168,16 @@ public List<String> validateModelArguments(final String missionModelId, final Ma
          MissionModelLoadException,
          InvalidArgumentsException
   {
-    return this.loadConfiguredMissionModel(missionModelId)
-               .validateConfiguration(arguments);
+    return this.loadUnconfiguredMissionModel(missionModelId)
+               .validateMissionModelArguments(arguments);
```
Mythicaeda: This is the only use of the `validateMissionModelArguments` method. Given that it's a one-liner, should we inline it here and remove it from `MissionModelFacade`?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -171,9 +186,9 @@ public Map<String, SerializedValue> getModelEffectiveArguments(final String miss
          MissionModelLoadException,
          InvalidArgumentsException
   {
-    return this.loadConfiguredMissionModel(missionModelId)
-               .getEffectiveArguments(arguments);
-}
+    return this.loadUnconfiguredMissionModel(missionModelId)
+               .getMissionModelEffectiveArguments(arguments);
```
Mythicaeda: Same story and question here as with `validateMissionModelArguments`.

mattdailis:
I've gone ahead and inlined MissionModelFacade - if any reviewers think it clutters up this PR, I can easily pull those two commits into a follow-up PR

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
Twisol: Nah, I'm fine with whitespace changes. I just wanted to make sure my eyes weren't playing tricks on me. Thanks for factoring them out!

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/TaskSpecType.java
```diff
@@ -27,4 +27,27 @@ public UnconstructableTaskSpecException(final String message) {
...
+
+  /**
+   * This method must behave as though implemented as:
+   * {@snippet :
+   * return this.getValidationFailures(this.instantiate(activity));
+   * }
+   */
+  default List<String> validateArguments(final Map<String, SerializedValue> activity)
+  throws InvalidArgumentsException {
+    return this.getValidationFailures(this.instantiate(activity));
```
Twisol: Minor nitpick, but of the methods introduced to these interfaces, this is the only one that calls its parameter `activity`. I'd recommend naming it "arguments" here even if only because `instantiate()` (the first method in the composition this wraps) also calls it `arguments`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -178,15 +188,15 @@ public List<String> validateModelArguments(final String missionModelId, final Ma
          InvalidArgumentsException
   {
     return this.loadUnconfiguredMissionModel(missionModelId)
-               .validateMissionModelArguments(arguments);
+        .getConfigurationType()
+        .validateArguments(arguments);
```
Twisol: Very minor nitpick, but the indentation here is just a little inconsistent. I'd expect the method call dots to be vertically aligned, either in the way it was before (with the first method on the same line as `this`) or by moving the first method onto its own line.

This happens in a few places in this commit ("Inline MissionModelFacade.Unconfigured out of existence"). Don't worry about it unless it really bothers you too.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```diff
@@ -61,14 +51,18 @@ public void shouldGetActivityTypeList() {
...
-        final Map<String, ActivityType> typeList = unconfiguredMissionModel.getActivityTypes();
+      final var activityTypes = new HashMap<String, ActivityType>();
+      registry.taskSpecTypes().forEach((name, specType) -> {
+        activityTypes.put(name, new ActivityType(name, specType.getParameters(), specType.getRequiredParameters(), specType.getReturnValueSchema()));
+      });
+      final Map<String, ActivityType> typeList = activityTypes;
 
         // THEN
         assertThat(typeList).containsAllEntriesOf(expectedTypes);
     }
```
Twisol: This block seems to have been dedented relative to the surrounding code, probably because this file was originally set at 4-space indentation. I'm fine with either, as long as it's consistent :)

(This happened throughout this file.)

Twisol:
> I've gone ahead and inlined MissionModelFacade - if any reviewers think it clutters up this PR, I can easily pull those two commits into a follow-up PR

Seems like the inline went through pretty easily! I like it.

Twisol (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -238,46 +278,44 @@ private MissionModelFacade.Unconfigured<?> loadUnconfiguredMissionModel(final St
   }
 
   /**
-   * Load an {@link MissionModel} from the mission model repository using the mission model's default mission model configuration,
-   * and wrap it in an {@link MissionModelFacade} domain object.
+   * Load an {@link MissionModel} from the mission model repository using the mission model's default mission model configuration
```
Mythicaeda: You missed an "an"->"a" change here. Only commenting on it as I see it being changed elsewhere.

Mythicaeda (approved):
LGTM

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: Good catch - I've switched it back to "a" to match "a MissionModel".

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: After inlining, there is only one NoSuchActivityType exception - I've left it as `MissionModelService` because I see it as part of the interface. (I wouldn't put up a fight about pulling it out into a separate class - but I feel that there's no longer a compelling reason to, since the "duplicate" exception has now been deleted).

PR #315, opened 2022-09-06
[AERIE 2071]  Resolve Code Security Alerts

Mythicaeda:
* **Tickets addressed:** AERIE-2071
* **Review:** By commit  

## Description

Updated dependencies based on Dependabot warnings. Changed `aerie-commanding` to only log the database, host, and port, instead of also logging the username/password to the console in clear.

## Verification

Ran `./gradlew test clean`. Composed Aerie. Checked the output of `aerie-commanding` to make sure it was only logging what was expected.


skovati (approved):
LGTM! It wouldn't be a bad idea to introduce the Dependabot automatic dependency upgrades to catch stuff like this in the future too.

e2e-tests/package-lock.json
```diff
@@ -1714,9 +1714,9 @@
       }
     },
     "node_modules/jpeg-js": {
-      "version": "0.4.3",
-      "resolved": "https://registry.npmjs.org/jpeg-js/-/jpeg-js-0.4.3.tgz",
-      "integrity": "sha512-ru1HWKek8octvUHFHvE5ZzQ1yAsJmIvRdGWvSoKV52XKyuyYA437QWDttXT8eZXDSbuMpHlLzPDZUPd6idIz+Q=="
+      "version": "0.4.4",
```
camargo: Why does this `package-lock.json` have updates if we did not update the corresponding `package.json`?

Mythicaeda: I believe it's because I modified `node_modules/playwright-core` to use version 4.4 for `jpeg-js` on line 1923

camargo: Ah got it. So you should actually never edit `package-lock.json` directly since it is a generated file. Same with `node-modules` since they are installed. Instead try updating the [@playwright/test dependency in package.json](https://github.com/NASA-AMMOS/aerie/blob/develop/e2e-tests/package.json#L10), then `npm install` to generate the new `package-lock.json` file. This was [already fixed in Playwright](https://github.com/microsoft/playwright/pull/14942) so updating to the latest (`1.25.2`) should do the trick.

Mythicaeda: Thanks for the heads-up! I think I got confused by some version dates when I initially checked if I could change Playwright. I've changed the `package.json` as you've suggested.

camargo (approved):
LGTM

e2e-tests/package.json
```diff
@@ -7,7 +7,7 @@
     "test": "playwright test"
   },
   "dependencies": {
-    "@playwright/test": "1.20.0",
+    "@playwright/test": "1.25.2",
```
camargo, line 10: Here is the commit from the UI when we updated Playwright to >= [1.21.0](https://github.com/microsoft/playwright/releases/tag/v1.21.0), which added native ESM support when running TypeScript tests: https://github.com/NASA-AMMOS/aerie-ui/commit/4734ecf29aa1497e35063ce563e145339ddaecd8
We need to make similar updates to this commit to make the tests pass.

PR #318, opened 2022-09-08
Remove some unnecessary simulation system complexity (and fix a bug)

Twisol:
* **Tickets addressed:** [AERIE-2086](https://jira.jpl.nasa.gov/browse/AERIE-2086)
* **Review:** By commit

## Description
This PR consists of a series of simplifications to the simulation system, including a fix for a major unreported bug in the interaction between spawning lambdas and spawning activities. (Look at that lines changed delta -- about a 400 line reduction in total!)

* Some elements of the `merlin.protocol` interface were made redundant by previous changes (#86, #170), including `Phantom`, `DirectiveTypeId`, and `DirectiveTypeRegistrar`. Ultimately, the `Registry` type parameter on `MissionModelFactory` and other interfaces has gone away entirely.
* One of the `Scoped` variables used by generated model code wasn't being propagated correctly to child tasks spawned from functions (as opposed to child activity instances), causing any further grandchild activities to fail to spawn. In particular, since `defer(...)` is implemented with a wrapping `spawn()` containing the desired delay, any attempt to defer a child activity would fail.

    The solution I ended up reasoning my way to was to remove the problematic `Scoped` variable entirely, and instead require `spawn()` actions to provide the model to invoke the activity on explicitly. This is no issue for any uses in-tree or in `aerie-lander`, as these actions always occur within another activity (whose `mission` argument is immediately available) or within a test method (which always has a `mission` field to make use of). If downstream usage patterns include spawning of activities from within submodels proper, these cases might need a reference to the top-level model to be passed down to the submodels -- but, this is a return to how things "normally" work in Java.

    This might seem like a step backward, but it actually opens the door to multi-mission activity definitions, as well as models that can be instantiated multiple times (think primary + backup radios) and yet share a common activity type (differing only in which instance it gets invoked on). It also raises the possibility of defining activities against individual submodels, rather than all taking the same top-level model type, which could improve the ability of different modeling teams to define their subsystems independently.
* I also replaced `waitFor()` with a fused `call()` action, and removed the `String` return value from `spawn()` and `defer()`. This has long been something we've wanted to do, and it moves us toward being able to return a spawned activity's return value from a `call()` without holding onto it (potentially bloating memory) for the entire duration of the simulation. Most uses of `waitFor()` can be replaced with a `call(() -> { ... })` which spawns any desired parallel tasks and automatically waits for their completion.

## Verification
All tests pass; I also made an analogous series of changes to `aerie-lander`, whose tests also pass.

## Documentation
- [ ] The mission modeler guide may need to be updated for the changes made to our JUnit extension.
- [ ] The mission modeler guide *definitely* needs to be updated for the new signatures of activity `spawn()`, `defer()`, etc.

## Future work
* Lots and lots of Javadoc documentation. I'm working on it! All these changes came about because I started documenting `MissionModelFactory` and realized there was a big chunk of stuff that simply didn't need to exist.
* The big bad breaking change down the road (ever so steadily creeping closer) is getting resources registered ahead of model instantiation. I think a lot of other things will change here, too, as we will need to introduce a location for user-defined code invoked before the model gets instantiated. Probably, activity type registration will move here (and out of `package-info.java`), though we will still provide mapper generation. (But, after this is done, the top-level model won't need a `Registrar` parameter any more, so we should be able to simplify the JUnit experience even more 👀)
* Change `call()` to return the return value of the child activity. We're very close now!


Twisol:
Latest push renames `MissionModelFactory` to `ModelType`, to parallel `DirectiveType` and `ConfigurationType` as reflective interfaces, and refactors `DirectiveType`, `ConfigurationType`, and `Resource` using new common `InputType` and `OutputType` reflective interfaces.

patkenneally (approved):
Thank you a 100 times, for splitting these changes each into their own bite sized commit.

Twisol:
I've rebased onto latest `develop` -- with one more review this will be ready to merge.

skovati (approved):
This is **awesome** work. I especially love seeing those sim driver types be public 👍

PR #321, opened 2022-09-09
Avoid leaking an open JarFile

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
I noticed an IntelliJ warning about a `JarFile` not being used with `try-with-resources`. I doubt this is a significant leak, but given the recent effort to reduce memory leaks, this PR ensures that the `JarFile` gets closed promptly.

## Verification
Tests pass.


mattdailis (approved):
Good find. I suppose the URL class loader handles reopening the jar as needed?

Twisol:
Yes, I think so. And the `jarFile` we open is anyway not used for classloading -- just to look up the contents of a file.

patkenneally (approved):

skovati (approved):
Nice catch!

PR #322, opened 2022-09-11
Facilitate model-scoped SPICE

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR enables each loaded model to acquire their *own* isolated instance of SPICE. (I kind of can't believe I'm writing those words!) The key ingredients are:

* The host program must *not* have JNISpice available, much less depend on JNISpice itself. It is absolutely crucial that the classes `cspice.basic.CSPICE` and `gov.nasa.jpl.aerie.spice.SpiceLoader` be loaded by the model JAR's `ClassLoader`, *not* the Aerie application classloader. As I found out the hard way, nested classloaders *check their parents first* before attempting any specialized behavior. So the parent classloader *must not* satisfy requests for JNISpice classes.
* The native library *must* be loaded from completely independent copies (with completely different canonical file paths) every time. It isn't sufficient to simply hard-link a file with a different name to the same `libJNISpice.so`; the JVM and the OS both must be convinced that they are *independent* images, or else either the JVM will throw an error (it won't tolerate explicit duplicates, or even distinct files with the same path) or the OS will silently provide a handle to the same library you've previously opened.

    ("But wait, Jonathan -- how can two distinct files have the same path???" Well, made-up interrogator, if you load a library at a path, delete that file, and create a new one in its place, your JVM will have a reference to a file that is no longer accessible to the filesystem, but *did* have a particular path on load, and you will have a *new* file that the JVM has not (yet) loaded at the same path!)

I've gone through the Aerie codebase and removed all indirect dependencies on SPICE that are not model-side. The SPICE classes used to be accessed through `merlin-sdk`, which meant they were available *everywhere*. I moved them to the `contrib` module, and cleaned up any minor dependencies on `contrib` from the rest of the system.

The current NAIF distribution of JNISpice for MacOS throws us another curveball: it compiles the library with extra flags (see "Future work") that cause the first-loaded library to greedily satisfy all matching `native` method resolutions, even if the JVM is specifically asking for a symbol from a different loaded library. I was able to make a custom build of libJNISpice.jnilib and confirm that, all together, these factors allow each model to have their own isolate of SPICE.

I've replaced the `libJNISpice.jnilib` in our source tree with my custom-built one. Neither this nor the old one is particularly easy to reproduce, so I don't think I'm making the situation any *worse* with this change. Nonetheless, it's worth noting that I *did* build both an Intel and an M1 version of the library and fuse them together (with [`lipo`](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary)), since the old one was similarly universal.

On Linux, [this StackOverflow discussion](https://stackoverflow.com/questions/40026804/loading-multiple-versions-of-java-classes-that-use-native-code) indicates that we should get the desired behavior without any changes. We'll see.

## Verification
I modified Banananation's `Mission` class to load SPICE and perform a `furnsh` bracketed by two `ktotal`s printed to screen. I then made a little `main()` in `simulation-driver`, confirmed that it didn't already observe `CSPICE` using the system classloader, and loaded `banananation.jar` three times in succession.

Normally this would output `0 1 1 2 2 3`, as each successive model instance observes the previous one's kernel loads.

Now, it outputs `0 1 0 1 0 1`. Isolation! (Holy smokes!)

## Documentation
* We need to document for users this new understanding about how native libraries interact with models. In particular, if models bundle their own native libraries and load them, they should *automatically* be scoped to that model. (As long as they don't include antagonizing flags like libJNISpice.jnilib did, at least!)
* This is *extremely delicate* right now; if any application-side Aerie component ever needs SPICE and needs to work with arbitrary mission models, the application-level SPICE is likely to be shared with all the models it loads. We ought to document this somewhere, and very clearly and carefully, because it is not the kind of thing that is very loud when it breaks.

    Maybe we should add a `ClassLoader.loadClass("spice.basic.CSPICE")` call in a reasonable location for the application-side components, and throw an error if it succeeds? At least then it will break loudly.

    I have a prototype solution for this, but it is *not* the kind of thing that can easily be placed on top of JNISpice. It's something that would need to be integrated pretty well into JNISpice proper. So I'm not sure how likely that is to happen.

## Future work
* Petition NAIF to remove the flags `-flat_namespace -undefined suppress` from their build script for `libJNISpice.jnilib`. As it stands, the `-flat_namespace` flag causes all native methods to get resolved to the first loaded copy of the library, making it impossible to interact with multiply-loaded SPICE libraries.

* Petition NAIF to distribute an M1 (arm64) version of JNISpice. If they distribute separate Intel and M1 versions with the above flags removed, we can easily fuse them together into a universal library using [`lipo`](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary).

* Implement a [child-first classloader](https://medium.com/@isuru89/java-a-child-first-class-loader-cbd9c3d0305) (see [also](https://stackoverflow.com/questions/5445511/how-do-i-create-a-parent-last-child-first-classloader-in-java-or-how-to-overr))to better insulate models from the application classpath. If both sources provide a class, we want the model to see the model's version, not the application's version. (This would likely allow the application to depend on CSPICE without causing global sharing to occur -- though some details around native library loading are still unclear.)

* Make the scheduler less dependent on the particular representation of controllable duration parameters. The model could always override the value mapper used for `Duration`, so the scheduler's use of `DurationValueMapper` was already suspect; my changes here just make that dependency explicit.

    We ought to instead add a method `fromDuration` to the `SchedulerModel` that allows the scheduler to turn a given `Duration` into a `SerializedValue` suitable for that parameter. I have a mock-up [here](https://gist.github.com/Twisol/b22fb42f07c69007ccbace4cac05e9e8).


Twisol:
Apparently the custom JNISpice I built is already universal.

```
❯ lipo -archs contrib/src/main/resources/gov/nasa/jpl/aerie/spice/libJNISpice.jnilib
x86_64 arm64
```

I'm not gonna question it.

Twisol:
With some effort, I've built an Intel (x86_64) version of JNISpice with the offending flags removed, and fused it with my M1 (arm64) build to recover a universal library. I've replaced the one in the tree; neither is particularly easy to reproduce, but...

examples/banananation/build.gradle
```diff
@@ -13,15 +13,24 @@ test {
...
+jar {
+  from {
+    configurations.runtimeClasspath.filter{ it.exists() }.collect{ it.isDirectory() ? it : zipTree(it) }
+  } {
+    exclude 'META-INF/LICENSE.txt'
+    exclude 'META-INF/NOTICE.txt'
+    exclude 'META-INF/LICENCE'
+  }
+}
+
```
mattdailis: This `jar` section seems to have been added to `banananation`, but not the other example mission models.

merlin-framework-processor/build.gradle
```diff
@@ -10,6 +10,7 @@ java {
 }
 
 dependencies {
+  implementation project(':merlin-sdk')
```
mattdailis, line 13: This dependency seems to have been added in a commit called "Remove [...] dependencies" 🤔  (more importantly, it seems like it might be unnecessary..?)

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
```diff
@@ -68,23 +67,16 @@ public static MerlinPlugin loadMissionModelProvider(final Path path, final Strin
...
 
-            // SAFETY: We checked above that MissionModelFactory is assignable from this type.
-            @SuppressWarnings("unchecked")
-            final var factoryClass = (Class<? extends MerlinPlugin>) factoryClass$;
-
-            return factoryClass.getConstructor().newInstance();
-        } catch (final ClassNotFoundException | NoSuchMethodException | InstantiationException
-            | IllegalAccessException | InvocationTargetException ex)
-        {
+            return (MerlinPlugin) factoryClass$.getConstructor().newInstance();
```
mattdailis, line 78: Figure no SAFETY comment necessary, because there's the assignable check above?

contrib/src/main/java/gov/nasa/jpl/aerie/spice/SpiceLoader.java
```diff
@@ -12,24 +12,23 @@ public final class SpiceLoader {
...
-        final String library = "JNISpice";
+        try (final InputStream in = SpiceLoader.class.getResourceAsStream(getResourcePathByOS())) {
+            if (in == null) throw new RuntimeException("JNISpice native library not found");
 
-        // Attempt to load our copy of JNISpice via `System.load`.
-        try (final InputStream in = SpiceLoader.class.getResource(getResourcePathByOS()).openStream()) {
             // Copy the JNISpice library to a temporary location on-disk, then load it into the JVM.
-            final Path file = Files.createTempFile(library + "-", ".tmp");
-            deleteFileOnExit(file);
+            // This needs to be a distinct file from any other load, or else the OS will happily give
```
mattdailis, line 19: Vague concern that I don't have a concrete suggestion for:

It doesn't seem like we're guaranteeing that this name is unique (and in general, we probably _can't_ guarantee it). Specifically I'm considering a naive implementation of `createTempFile` such as:
```python
def createTempFile(name):
    i = 0
    while exists(name + i):
        i += 1
    return makeFile(name + i)
```

I suppose the answer is "most createTempFile implementations don't do that, so stop worrying, Matt".

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -49,10 +48,10 @@
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import static gov.nasa.jpl.aerie.merlin.driver.json.ValueSchemaJsonParser.valueSchemaP;
 import static gov.nasa.jpl.aerie.scheduler.server.graphql.GraphQLParsers.parseGraphQLInterval;
 import static gov.nasa.jpl.aerie.scheduler.server.graphql.GraphQLParsers.parseGraphQLTimestamp;
 import static gov.nasa.jpl.aerie.scheduler.server.http.SerializedValueJsonParser.serializedValueP;
-import static gov.nasa.jpl.aerie.merlin.driver.json.ValueSchemaJsonParser.valueSchemaP;
```
Mythicaeda, line 55: Is this just moving line 55 to line 51?

scheduler/build.gradle
```diff
@@ -14,11 +14,10 @@ test {
 
 dependencies {
   api 'com.apollographql.apollo:apollo-api:2.5.11'
-  compileOnly project(':merlin-framework')
   compileOnly project(':merlin-driver')
   compileOnly project(':constraints')
-  compileOnly project(':contrib')
 
+  implementation 'org.apache.commons:commons-math3:3.6.1'
```
Mythicaeda, line 20: Where is this import being used?

merlin-framework-processor/build.gradle
Twisol: Somewhat frustratingly, it's not actually unnecessary -- the processor needs the merlin-sdk classes, and as best as I can tell, the compile-time classpath (the `compileOnly` dependency for models on `merlin-sdk`) is separate from the runtime classpath of the processor actively executing in the compiler's JVM. That is, the reflective APIs at processing time consider e.g. the `Duration` referenced by the model to be distinct from the `Duration` referenced by the processor.

I'll tweak the commit message to be more clear about what the dependency reshuffling is doing, though.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
Twisol: That's right. Would you prefer I add one anyway for clarity?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```diff
@@ -46,7 +45,7 @@ public void shouldGetActivityTypeList() {
                     new Parameter("y", ValueSchema.STRING),
                     new Parameter("vecs", ValueSchema.ofSeries(ValueSchema.ofSeries(ValueSchema.REAL)))),
                 List.of(),
-                new UnitValueMapper().getValueSchema()
+                ValueSchema.ofStruct(Map.of())
```
Mythicaeda, line 48: I know you already mentioned that this is very brittle, @Twisol, but I'm highlighting this for other reviewers, since your description doesn't mention that part of the brittleness comes from inlining certain methods from `contrib` classes.

contrib/src/main/java/gov/nasa/jpl/aerie/spice/SpiceLoader.java
Twisol: Ah, but I delete the file afterwards, so it *must* be a unique file (inode) every time, regardless of the name 😇

Twisol: To be more concrete: it's not the *name* that must be unique, it's the *file* (i.e. the inode representing that file in the filesystem). Naming files different things at different paths is simply one way to make sure you're always creating a new file and not accidentally overwriting the contents of another. But since I delete the file immediately after it's loaded, the inode remains but is totally inaccessible from the filesystem. So we're always creating new files, every time, regardless of name.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
Twisol: Yes. The import probably got reorganized when IntelliJ refactored `DurationValueMapper` out of the imports list. Notice that it's out of alphabetic order pre-commit.

I could revert this but I'm not particularly inclined to 🤔

scheduler/build.gradle
Twisol: It was indirectly imported via the `contrib` dependency that's been removed. It's used by the scheduler in a few math-heavy places; I don't know what they're about, only that they broke when `contrib` was removed, so I needed to add the `commons-math3` dependency back explicitly.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
Mythicaeda: That's fine, was just confirming.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
Twisol: The brittleness I was referring to is from the delicate nature of the classloader situation, where we need to avoid having CSPICE on the classpath at all costs.

I'd claim that this particular line is also not brittle, because we're testing specifically using the `foo-missionmodel` model, so if `foo-missionmodel` changes its representation, these tests will instantly break and we will update the tests, same as any coupling between unit tests and the unit they test.

I'll add information about the brittleness of the scheduler's `Duration` usage in the "Future work" section in a moment.

examples/banananation/build.gradle
Twisol: True. I'll go through and add it to the others, too.

Twisol:
FYI, I've updated the Future work section with some information about places where we'll want to make improvements later. Thanks to @Mythicaeda for a robust conversation on these topics.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
Mythicaeda: Understood. My concern came from inlining also occurring`SimulationFacade` and `GraphQLMerlinService` with the DurationValueMapper, which are used are used outside of testing (which you have now added to the `Future Work` section.

Twisol: Right, agreed. Since those other uses (unlike this one) all seem to be in the scheduler, I think it's fair to say that the future work I outlined will address that. Thanks for the detailed check :+1:

contrib/src/main/java/gov/nasa/jpl/aerie/spice/SpiceLoader.java
Twisol: Also, [the javadocs](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/io/File.html#createTempFile(java.lang.String,java.lang.String,java.io.File)) say this:

> Creates a new empty file in the specified directory, using the given prefix and suffix strings to generate its name. If this method returns successfully then it is guaranteed that:
> 1. The file denoted by the returned abstract pathname did not exist before this method was invoked, and
> 2. Neither this method nor any of its variants will return the same abstract pathname again in the current invocation of the virtual machine.

Which is quite cool.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
mattdailis: No need, I think it's quite clear as is.

contrib/src/main/java/gov/nasa/jpl/aerie/spice/SpiceLoader.java
mattdailis: That is quite cool! In that case, my concern is assuaged.

I hadn't considered that it might be the inode that needed to be unique - definitely could use some brushing-up on filesystems.

What led me to worry about the uniqueness of the filename was this snippet of code:

([source](https://github.com/openjdk/jdk18/blob/0f2113cee79b9645105b4753c7d7eacb83b872c2/src/java.base/share/classes/jdk/internal/loader/NativeLibraries.java#L196-L206))
```java
// find if this library has already been loaded and registered in this NativeLibraries
NativeLibrary cached = libraries.get(name);
if (cached != null) {
    return cached;
}


// cannot be loaded by other class loaders
if (loadedLibraryNames.contains(name)) {
    throw new UnsatisfiedLinkError("Native Library " + name +
            " already loaded in another classloader");
}
```

Now, I haven't done any digging into where `loadedLibraryNames` comes from - however, I am satisfied by guarantee number 2 above - "Neither this method nor any of its variants will return the same abstract pathname again in the current invocation of the virtual machine."

examples/banananation/build.gradle
```diff
@@ -13,15 +13,24 @@ test {
   useJUnitPlatform()
 }
 
+jar {
```
Mythicaeda, line 19: Reminder: Add a comment @Twisol

Mythicaeda (approved):
Overall looks good, although I'm a little concerned that the `.jnilib` changed size.

contrib/src/main/java/gov/nasa/jpl/aerie/spice/SpiceLoader.java
Twisol: Oh fascinating. When I did this with my custom JNI prototype, I was able to get away with not renaming the library file at all -- just putting it in a different directory was sufficient. 

(Also, look at you, linking to the OpenJDK sources on GitHub 👀)

Okay, so I think I need to update the factors listed in the summary. *Both* the inode *and* the name must be unique -- the inode so the OS doesn't cheat, and the filename so Java doesn't scream. (You can have the same inode with a different name using a hard link; this causes the OS to cheat and say "no, I know I already opened this, you can have a new handle to it though.")

I really thought inode uniqueness would be sufficient. Huh.

Twisol: Yeah, the caller (in our flow from `System.load` anyway) of the function you linked computes `name` as `file.getCanonicalPath()`, so the name is sensitive to the entire path, not just the basename. That explains why both my directory name scheme and the `SpiceLoader`'s random temp filename scheme get past this requirement.

examples/banananation/build.gradle
mattdailis: One more question - is this `jar` an optimization, or is it semantically significant? Should we advise users to add this to their own build.gradle files?

Twisol: It is necessary. If a model has dependencies that are not provided by its environment -- which is *typical*, since we can't predict all needs of all models -- it must include those dependencies in the JAR it ships to our execution platform.

This is true even beyond SPICE -- it just happened that our models have been simple enough to not have many dependencies -- and lucky enough that those it did have, like commons-lang3, were incidentally shared by the execution platforms it ran on.

mattdailis (approved):
🚀 🎉 🥳 💯 🥇

examples/banananation/build.gradle
Twisol: Thanks; I've added a brief explainer comment to `banananation/build.gradle` about how models need to bundle their dependencies in general.

Mythicaeda:
@Twisol If you still have/remember what changes you made to Banananation during the verification of this PR, could you please share them? I'm trying to simulate a model that has SPICE calls, but am getting the following error: 

```
java.lang.UnsatisfiedLinkError: /tmp/JNISpice-15655851895714875059.tmp: /tmp/JNISpice-15655851895714875059.tmp: cannot open shared object file: No such file or directory (Possible cause: can't load AMD 64 .so on a AARCH64 platform)
```

Since it seems like you were able to run simulation on the modified Banananation model, I'd like to try running it in order to see if I get the same error.

Twisol:
The only temporary changes I made for verification were the ones listed here:

> I modified Banananation's Mission class to load SPICE and perform a furnsh bracketed by two ktotals printed to screen. I then made a little main() in simulation-driver, confirmed that it didn't already observe CSPICE using the system classloader, and loaded banananation.jar three times in succession.

It sounds like you're getting caught on the first step, the `SpiceLoader.loadSpice()`.

To confirm that I can still load and make SPICE calls today, I've just now added the following code to the constructor of banananation's `Mission` constructor. This doesn't test the ClassLoader isolation, but it should be enough to at least replicate your error:

```java
    // under banananation, in the constructor for Mission.java, add:
    try {
      SpiceLoader.loadSpice();
      System.out.println(CSPICE.ktotal("ALL"));
    } catch (final SpiceErrorException ex) {
      throw new Error(ex);
    }
```

When I then run banananation's test `Main` locally through IntelliJ, everything runs to completion and I don't observe any error. (I also observe an extra `0`, which is the result of calling `ktotal` here.)

Mythicaeda:
Thanks for sharing! When I run the modified Banananation, I get the same error. To try and figure out the possible variables, did you run it on an Intel Mac or an M1 Mac? I'm running it on an M1 Mac.

Update: I'm going to try temporarily removing the `Files.delete(file)` line in SpiceLoader, just to make sure that the error isn't being caused by my machine trying to access the temp file after deletion for some reason.

Update on the update: Confirmed, the error is happening even if the temp file is not deleted.

Twisol:
> Thanks for sharing! When I run the modified Banananation, I get the same error. To try and figure out the possible variables, did you run it on an Intel Mac or an M1 Mac? I'm running it on an M1 Mac.

I'm also running it on an M1 Mac -- specifically a 2020 Mac mini. 🤔

Twisol:
Just to confirm, are you running `Main` natively through IntelliJ or within a Docker container? I can imagine the latter could be a confounding variable.

Mythicaeda:
Do you mean Banananation's `Main`? I'm running it by pressing the UI's `Simulate` button (and yes, I'm rebuilding the Docker images/volumes each time I compile). If I just press the green `run` button next to `Main` in IntelliJ, there are no errors. 

To be explicit, my current flow is:
- Make edits, `./gradlew clean; ./gradlew assemble`
- Delete the prior Docker images/volumes that come from compiled code in the Aerie repo, `docker compose up -d`
- Upload the Bananantion model, create a new empty Plan from it via the UI
- Run simulation via the UI

Twisol:
> Do you mean Banananation's `Main`?

Yes, I do:

> When I then run banananation's test `Main` locally through IntelliJ, everything runs to completion and I don't observe any error.

So it sounds like this is an environment-specific issue under Docker deployments. Based on [this StackExchange answer](https://apple.stackexchange.com/a/451240), I suspect the Docker images in your deployment are built for the AMD64 architecture (essentially x86_64). (The reason that doesn't simply fail to run at all would be because MacOS' Rosetta virtualization layer can run binaries built for x86_64.)

That's weird, though, because the CSPICE jnilib I generated contains two slices, one for each architecture, so if that's the case, I don't know why the JVM would be preferring the wrong slice. Maybe that "Possible cause" bit is a red herring?

I don't normally use a Docker deployment when I'm working on local simulation components, so I don't have much advice on this front...

Mythicaeda:
> I suspect the Docker images in your deployment are built for the AMD64 architecture

I'm not certain about that. When Docker runs images that were built for AMD64 architecture, it gives a warning. For example, the `aerie_gateway` and `aerie_ui` containers come from images compiled on AMD64 architecture and have the pictured warning, while you can see that the `aerie_merlin` container (whose image are directly coming from code compiled on my machine's architecture) doesn't have that warning. 
image

Twisol:
Interesting; good to know. I wonder what makes the difference in the Docker deployment then?

Twisol:
Are you able to potentially run Banananation's `Main` under the relevant Docker image so we can exclude the rest of the system stack from scope?

Twisol:
Also, maybe this would be a good thing to now open a new Issue for.

Mythicaeda:
> Interesting; good to know. I wonder what makes the difference in the Docker deployment then?

That's what I'm trying to figure out. Do we know anything about how the class loader decides which slice of the .jnilib to run? Given that Intel Macs aren't having this issue, I'm wondering if there's something going wrong with that slice selection.

> Are you able to potentially run Banananation's Main under the relevant Docker image so we can exclude the rest of the system stack from scope?

How do I do that?

> Also, maybe this would be a good thing to now open a new Issue for.

Sure, I'll open one now 👍

Twisol:
> Do we know anything about how the class loader decides which slice of the .jnilib to run?

I don't think this is a classloader thing; it'll be a JVM thing at best, and I'd be surprised if it wasn't actually an OS thing. As I recall, a .jnilib is really just a .so; I think it has a different suffix because there's already a `.so` bundled with `SpiceLoader` for Linux.

Ah, hang on, that's probably it. The Docker images are Linux-based, so `SpiceLoader` is going to copy out the `.so` for Linux; but that's only been built for x86_64 and doesn't contain a slice for ARM.

> > Are you able to potentially run Banananation's Main under the relevant Docker image so we can exclude the rest of the system stack from scope?
> 
> How do I do that?

1. Figure out some way to have Gradle generate a JAR that includes test classes. (Or hack the dependencies around and move `Main` into the main section temporarily. Either way, get a JAR that includes `Main`.)
2. Double-check that `java -jar banananation.jar gov.nasa.jpl.aerie.banananation.Main` works. (The command-line might be wrong; I haven't had to run a JAR directly like this in a while.)
3. Grab the Merlin simulation-worker Docker image, and mount the directory containing the built JAR into a new container for that image. Then, run the above command from that container. This will end up something like `docker run --rm -it --mount <stuff> <worker-image> java -jar /path/to/banananation.jar gov.nasa.jpl.aerie.banananation.Main`.

PR #323, opened 2022-09-13
feat: increase command expansion server body parser limit

camargo:
## Description

- When trying to upload a `27mb` command dictionary I got a ` request entity too large` error
- This commit just allows for much larger command dictionaries by increasing the max body size limit for the upload server

### DIFF
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -37,7 +37,7 @@ const PORT: number = parseInt(getEnv().PORT, 10) ?? 27184;
 const app: Application = express();
 // WARNING: bodyParser.json() is vulnerable to a string too long issue. Iff that occurs,
 // we should switch to a stream parser like https://www.npmjs.com/package/stream-json
-app.use(bodyParser.json({ limit: '25mb' }));
+app.use(bodyParser.json({ limit: '100mb' }));
 
 DbExpansion.init();
 const db = DbExpansion.getDb();


dyst5422 (approved):

command-expansion-server/src/app.ts
```diff
@@ -37,7 +37,7 @@ const PORT: number = parseInt(getEnv().PORT, 10) ?? 27184;
 const app: Application = express();
 // WARNING: bodyParser.json() is vulnerable to a string too long issue. Iff that occurs,
 // we should switch to a stream parser like https://www.npmjs.com/package/stream-json
```
mattdailis, line 39: Do we feel like we understand this comment? "vulnerable to a string too long issue"? What risks (if any) are there associated with raising the limit?

mattdailis (approved):
Seems reasonable to me! I left one question just for my own education.

command-expansion-server/src/app.ts
camargo: No idea. I think it's low risk because we will just get an `entity to large` error if we try to send a string that is too large. @dyst5422 can probably elaborate.

dyst5422: The limit is essentially there to avoid abusive payloads. A stream parser is going to be significant processing-time overhead, so unless we find ourselves in a situation that legitimately needs it, it's best to avoid. The hard limit on string length is 256mb, so we are well below that with the 100mb limit.

mattdailis: Thanks, both of you :+1:

dyst5422: And yes, @camargo is right that it's not a server-breaking issue if we attempt to load pass something larger.

dyst5422: We hit the string too long issue after ~1.5 years of operations on M2020 in a case where we were returning all the simulation results from SSIM (a very large payload) and did the switch to stream-json at that point, which slows things down for smaller payloads, but was necessary there, so there's prior art on lab in doing that shift.

PR #324, opened 2022-09-13
[AERIE-2049]  Fallible seqJson actions

dyst5422:
* **Tickets addressed:** AERIE-2049
* **Review:** By commit  

## Description

Added some GraphQL interfaces for wrangling errors

## Verification

Tests updated and run locally

## Documentation

Handled by auto-doc generation

## Future work

Continue looking at our other Fallible responses

- [ ] Update command expansion documentation with new formats
- [ ] Update aerie UI with new response formats

### DIFF
--- a/command-expansion-server/src/app.ts
+++ b/command-expansion-server/src/app.ts
@@ -13,8 +13,8 @@ import { InferredDataloader, objectCacheKeyFunction, unwrapPromiseSettledResults
 import { commandDictionaryTypescriptBatchLoader } from './lib/batchLoaders/commandDictionaryTypescriptBatchLoader.js';
 import { activitySchemaBatchLoader } from './lib/batchLoaders/activitySchemaBatchLoader.js';
 import {
-  simulatedActivitiesBatchLoader,
   SimulatedActivity,
+  simulatedActivitiesBatchLoader,
   simulatedActivityInstanceBySimulatedActivityIdBatchLoader,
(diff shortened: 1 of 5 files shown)


goetzrrGit (approved):
Looks great. I am glad we had a meeting to go over this with the team.

command-expansion-server/src/app.ts
```diff
@@ -571,11 +577,26 @@ app.post('/get-seqjson-for-seqid-and-simulation-dataset', async (req, res, next)
     };
   });
 
+  const errors = sortedSimulatedActivitiesWithCommands.flatMap(ai => ai.errors ?? []);
```
Mythicaeda, line 620: Why do we only need to flatten the `errors` array here? Is there something special about what it might (or does) look like in this method?

dyst5422: Each activity instance included can have multiple errors (type check errors for example) and we are simply collecting them all together into a single array of all errors for the requested operation.

So making `{ errors: Error[], ...etc }[]` -> `Error[]`

Mythicaeda: Gotcha. So, just to confirm, we don't flatten elsewhere (for example, `get-seqjson-for-sequence-standalone`) because there we're only gathering one activity's errors, as opposed to here where we have an array of arrays?

dyst5422: Correct.

Mythicaeda (approved):
LGTM!

PR #327, opened 2022-09-14
[HOTFIX] Ensure Duration is Long prior to serialization

patkenneally:
* **Tickets addressed:** hot fix
* **Review:** By commit  

## Description
The Duration parameter for a controllable activity was being double quoted. 1st converted to a string, and then secondly being serialized by the JsonObject builder. This resulted in the Duration activity parameter in microseconds being inserted into the activity-directives table arguments jsonb column as a JSON string, and not the expected JSON number. As a result simulation would fail when trying to instantiate the scheduler created activity directives because the arguments to the duration parameters were Strings and not the expected Long.

This change removes the initial conversion to a String.

## Verification
I tested this but I'm creating a ticket to add a set of end-to-end tests on this aspect for the scheduler. 
Prior to testing the insert resulted in { "duration": "1800000000" }. With this change the json entered into the db is { "duration": 1800000000 }

## Documentation
NA

## Future work
None

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
@@ -558,7 +558,7 @@ public record GraphQLMerlinService(URI merlinGraphqlURI) implements PlanService.
       final var insertionObjectArguments = Json.createObjectBuilder();
       if(act.getType().getDurationType() instanceof DurationType.Controllable durationType){
         if(!act.getArguments().containsKey(durationType.parameterName())){
-          insertionObjectArguments.add(durationType.parameterName(), serializeForGql(act.getDuration()));
+          insertionObjectArguments.add(durationType.parameterName(), act.getDuration().in(Duration.MICROSECOND));
         }
       }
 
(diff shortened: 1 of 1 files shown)


Twisol:
Interesting timing on finding this -- @patkenneally, what are your thoughts about #322's Future Work item relating to controllable durations?

> * Make the scheduler less dependent on the particular representation of controllable duration parameters. The model could always override the value mapper used for `Duration`, so the scheduler's use of `DurationValueMapper` was already suspect; my changes here just make that dependency explicit.
>
>     We ought to instead add a method `fromDuration` to the `SchedulerModel` that allows the scheduler to turn a given `Duration` into a `SerializedValue` suitable for that parameter. I have a mock-up [here](https://gist.github.com/Twisol/b22fb42f07c69007ccbace4cac05e9e8).

In particular because of the comment removed in this PR:

> //TODO: can probably leverage some serializers from aerie

Twisol (approved):

mattdailis:
🤦 

Another relevant PR to this bug - #301 - takes a different tack - it calls `getEffectiveArguments` on every activity in the plan, so in theory, this if statement should never be triggered. (The if statement checks if the "controllable duration" parameter is omitted - but after effective arguments, my understanding is that all arguments will be present).

mattdailis (approved):
Approving, because I think there's no harm in merging this one and rebasing #301 onto it :+1:

PR #210, opened 2022-06-08
[AERIE-1879] Commands with Dynamic Repeating Arguments are Out of Order

goetzrrGit:
* **Tickets addressed:** AERIE-1879 & AERIE-1886
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

Clipper has special commands with dynamic repeating arguments. 

ex.
`SEQ_VAR_CMD( [dynamic_arg -> n]..., opcode);`

The above command can be written below

```
SEQ_VAR_CMD( dynamic_arg1, opcode);
SEQ_VAR_CMD( dynamic_arg1, dynamic_arg2, dynamic_arg3, opcode);
SEQ_VAR_CMD( dynamic_arg1, dynamic_arg2, dynamic_arg3, dynamic_arg4, dynamic_arg5, opcode);
etc.
```

The command dictionary typescript library generator already handles these special commands, but with Javascript's `Objects` ordering are not guaranteed, we need to preserve the order of these dynamic arguments. Also, the arguments are missing when expanding the sequence.

Here is an example with a new bananantion command - `PACKAGE_BANANA`

```
    // arguments in order 
    PACKAGE_BANANA({
        bundle_name1: "Chiquita",
        number_of_bananas1: 43,
        bundle_name2: "Dole",
        number_of_bananas2: 12,
        lot_number: 1093,
      }),
      
      // arguments are out of order
      PACKAGE_BANANA({
        bundle_name2: "Dole",
        number_of_bananas1: 43,
        bundle_name1: "Chiquita",
        lot_number: 1093,
        number_of_bananas2: 12
      }),

      PACKAGE_BANANA("Chiquita",43,"Dole",12,"Blue",1,1093)

```

When you expand a sequence you should have the arguments for the command in the correct order. Also, the `args` should have the correct values. 

```
   {
      type: 'command',
      stem: 'PACKAGE_BANANA',
      time: { type: 'COMMAND_COMPLETE' },
      args: ['Chiquita', 43, 'Dole', 12, 1093],
      metadata: { simulatedActivityId: simulatedActivityId4 },
    },

    {
      type: 'command',
      stem: 'PACKAGE_BANANA',
      time: { type: 'COMMAND_COMPLETE' },
      args: ['Chiquita', 43, 'Dole', 12, 1093],     <----- Arguments are in the correct order
      metadata: { simulatedActivityId: simulatedActivityId4 },
    },
   
    {
      type: 'command',
      stem: 'PACKAGE_BANANA',
[...]

PR #328, opened 2022-09-14
[Aerie 2021]  Remove Deprecated Time Formats

goetzrrGit:
* **Tickets addressed:** Aerie 2021
* **Review:** By commit  

## Description
It is time to remove the deprecated methods in the command authoring API. Now the below methods are invalid:

```
.absoluteTiming()
.relativeTiming()
.epochTiming()
```

Moving forward you will need to use the timing types A/R/E/C before calling the commands.

ex.

`PREHEAT_OVEN(350) --> C.PREHEAT_OVEN(350)`

Overall I've removed the command methods from `declare Global{}` but left their interfaces so the IDE's content assist will still work. I also went through and cleaned up the code generation for commands with dynamic arguments.

## Verification
I have updated the command expansion test and all pass. I also used Clipper command Dictionary to make sure there weren't any issues with a more fleshed-out dictionary. 

## Documentation
I have updated the documentation on the wiki to reflect these changes
https://github.com/NASA-AMMOS/aerie/wiki/Expansion-Logic-API#time

## Future work
None

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
+++ b/command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
@@ -12,7 +12,7 @@ function generateTypescriptCode(dictionary: ampcs.CommandDictionary): {
   declarations: string;
   values: string;
 } {
-  const typescriptFswCommands: { value: string; declaration: string }[] = [];
+  const typescriptFswCommands: { value: string; interfaces: string }[] = [];
   for (const fswCommand of dictionary.fswCommands) {
     typescriptFswCommands.push(generateFswCommandCode(fswCommand, dictionary.enumMap));
   }
@@ -20,7 +20,7 @@ function generateTypescriptCode(dictionary: ampcs.CommandDictionary): {
   // language=TypeScript
   const declarations = `
 declare global {
-${typescriptFswCommands.map(fswCommand => fswCommand.declaration).join('\n')}
+${typescriptFswCommands.map(fswCommand => fswCommand.interfaces).join('\n')}
 \tconst Commands: {\n${dictionary.fswCommands
     .map(fswCommand => `\t\t${fswCommand.stem}: typeof ${fswCommand.stem},\n`)
     .join('')}\t};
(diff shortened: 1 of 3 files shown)


command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -85,65 +84,72 @@ ${doc}
...
-      let repeatArgsDeclaration = '';
-      let repeatArgsNames: string[] = [];
+      let repeatArgs: Array<{ name: string; type: string }> = [];
+
       for (let n = 1; n < i; n++) {
         if (repeatArg.repeat) {
-          repeatArgsNames = repeatArgsNames.concat(repeatArg.repeat?.arguments.map(arg => `"${arg.name}${n}"`));
+          repeatArgs = repeatArgs.concat(
+            repeatArg.repeat.arguments.map(arg => {
+              return { name: `${arg.name}${n}`, type: mapArgumentType(arg, enumMap) };
```
pcrosemurgy: This likely doesn't matter too much since this is generated code but a character to more clearly separate the argument name and `n` may be good to have. Maybe:
```ts
${arg.name}_${n}
```

pcrosemurgy: Also, this `if (repeatArg.repeat)` can be lifted above the `for (...)`.

pcrosemurgy: >This likely doesn't matter too much since this is generated code but a character to more clearly separate the argument name and n may be good to have.

Ohh, looks like the format you have is sticking to some spec. that Clipper expects (see #210), is that right?

pcrosemurgy (approved):
Looks good! I'd suggest moving that `if` statement to wrap the `for` but otherwise all good.

command-expansion-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: Let me try lifting the `if` and run some tests to see if that trips up anything. I am not opposed to adding an `_` I don't see it breaking anything but I will have to do some testing.

goetzrrGit: I lifted the `if` and add `_` for the argument name:

ex. 

```
        bundle_name1: VarString<8, 1024>, --> bundle_name_1: VarString<8, 1024>
        number_of_bananas1: U8, -->  number_of_bananas_1 : U8
```

pcrosemurgy: Oh sweet, thanks Ryan!

dyst5422 (approved):
I like the refactor you have in here!

PR #332, opened 2022-09-15
[FIX][AERIE-2022] Scheduling does not see child activities

adrienmaillard:
* **Tickets addressed:** AERIE-2022
* **Review:** By commit  

## Description

The scheduler was not pulling generated activities from the simulations and thus could not satisfy any goal with these activities. 

- every time a simulation is performed, we pull all generated activities. we check that compared to previous simulations,  previously existing generated activities still exist. Then we add new generated activities to the plan. Activity instances are marked as generated. They are used for satisfaction during scheduling. 
- When pushing changes to the database, we do not push generated activity instances. Goals satisfied by a generated activity are associated with its top-level parent. 

## Verification

Two tests were added to scheduler integration tests. 

## Documentation

None. 

## Future work

We should work on the "continuous" satisfaction of goals. Here is an example: let's say we insert an activity A1 at time T1 to satisfy coexistence goal G1 based on the value of some state S1. Now we insert A2 to satisfy G2 at T2 < T1, and A2 is modifying the value of S1 at time T1. Maybe G1 becomes unsatisfied.   

With generated activities, the state of resource S1 could impact what activities are generated by A1 and thus the satisfaction of another goal supported by these activities. 

Checking the stationarity of goal satisfaction would increase robustness. The easiest way would be to run `Goal.getConflicts` for each previously processed goal before any new goal and checking that it stays constant. It might be costly. There may be other way of doing that with the notion of "support". An inserted activity to satisfy a goal could reference supports. It could be existing activities, resource values at a specific time... checking supports is much faster than checking conflicts.


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -440,20 +440,31 @@ private ScheduleResults collectResults(final Plan plan, final Map<ActivityInstan
...
-                  .collect(Collectors.toList()),
+                  .getAssociatedActivities().stream().map(activityInstance -> {
+                    ActivityInstanceId id;
+                    if(activityInstance.isGenerated()){
+                      id = instancesToIds.get(activitiesById.get(activityInstance.getParentActivity()));
+                    } else{
+                      id = instancesToIds.get(activityInstance);
+                    }
+                    return id;
+                  }).toList(),
```
pcrosemurgy: Can replace the `getInsertedActivities()` and `getAssociatedActivities()` logic with:
```java
.getAssociatedActivities().stream().map(activityInstance -> instancesToIds.get(activityInstance.isGenerated() ?
    activitiesById.get(activityInstance.getParentActivity()) :
    activityInstance)).toList(),
```

pcrosemurgy: @adrienmaillard I've pushed a commit to make `getParentActivity()` return an `Optional<SchedulingActivityInstanceId>`.

This makes client-side calls safer since they don't have to use `isGenerated()` guarding (which is not guaranteed to prevent `getParentActivity()` returning a `null` should `ActivityInstance`'s implementation change). Feel free to discard my commit, no hard feelings :)

pcrosemurgy (approved):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
adrienmaillard: It looks better, thank you.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -445,26 +445,20 @@ private ScheduleResults collectResults(final Plan plan, final Map<ActivityInstan
...
-                      id = instancesToIds.get(activityInstance);
-                    }
-                    return id;
-                  }).toList(),
+                  .getAssociatedActivities().stream()
+                      .map(activityInstance -> instancesToIds.get(
+                        activityInstance.getParentActivity()
+                            .map(activitiesById::get)
+                            .orElse(activityInstance))
+                    ).toList(),
```
Mythicaeda, line 526: Just double-checking, the parenthesis is closing the get on line 457, right?

pcrosemurgy: Oops, the line-breaking and parentheses-breaking here is formatted kind of strangely. That final parenthesis is for the `map`. Just pushed formatting change for this :)

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -90,6 +98,33 @@ public Optional<Duration> getActivityDuration(ActivityInstance activityInstance)
...
+    Map<ActivityInstance, SchedulingActivityInstanceId> generatedActivities = new HashMap<>();
+    this.lastSimDriverResults.simulatedActivities.forEach( (activityInstanceId, activity) -> {
+      final var sameActInConstraintsResults = this.lastSimConstraintResults.activities.stream().filter(act -> act.id == activityInstanceId.id()).findFirst().get();
+      final var rootParent = getIdOfRootParent(this.lastSimDriverResults, activityInstanceId);
+      if(!rootParent.equals(activityInstanceId)){
+        //generated activity
+        final var schedulingActId = planActInstanceIdToSimulationActInstanceId.entrySet().stream().filter(
+            entry -> entry.getValue().equals(rootParent)
+        ).findFirst().get().getKey();
+        final var activityInstance = new ActivityInstance(activityTypes.get(sameActInConstraintsResults.type), sameActInConstraintsResults.window.start, sameActInConstraintsResults.window.end.minus(sameActInConstraintsResults.window.start), sameActInConstraintsResults.parameters, schedulingActId);
```
mattdailis: 🔻 Let's put these arguments on separate lines - I don't have a hard rule for how long lines can be, but 299 characters is too long 😝

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -90,6 +98,33 @@ public Optional<Duration> getActivityDuration(ActivityInstance activityInstance)
...
+    } else {
+      return getIdOfRootParent(results, act.parentId());
+    }
+  }
+
+  public Map<ActivityInstance, SchedulingActivityInstanceId> getAllGeneratedActivities(Duration endTime){
+    computeSimulationResultsUntil(endTime);
+    Map<ActivityInstance, SchedulingActivityInstanceId> generatedActivities = new HashMap<>();
+    this.lastSimDriverResults.simulatedActivities.forEach( (activityInstanceId, activity) -> {
+      final var sameActInConstraintsResults = this.lastSimConstraintResults.activities.stream().filter(act -> act.id == activityInstanceId.id()).findFirst().get();
```
mattdailis: 🔶 What information do we need from `lastSimConstraintResults` that we can't get from `activity`? It seems like it has all the same information in a slightly different format.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -221,7 +225,42 @@ public void initializePlan() throws SimulationFacade.SimulationException {
...
+          //break at first occurrence. there may be several activities equal in properties.
+          break;
+        }
+      }
+    }
+
+    //copyNew contains only things that are new
+    //copyOld contains only present in old but absent in new
+    if(copyOld.size() != 0){
+      throw new Error("Activities have disappeared from simulation, failing");
```
mattdailis: 🔶 Do we think that this should always be an error? I see in the Future Work section you mentioned continuous satisfaction and supports - is that related to this code here?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -221,7 +225,42 @@ public void initializePlan() throws SimulationFacade.SimulationException {
...
+    final var copyOld = new ArrayList<>(this.generatedActivityInstances);
+    final var copyNew = new ArrayList<>(formattedNewGeneratedActivities);
+
+    for(final var pairOld: this.generatedActivityInstances){
+      for (final var pairNew : formattedNewGeneratedActivities){
+        if(pairOld.getLeft().equalsInProperties(pairNew.getLeft()) &&
+           pairNew.getRight().equals(pairOld.getRight())){
+          copyNew.remove(pairNew);
+          copyOld.remove(pairOld);
+          //break at first occurrence. there may be several activities equal in properties.
```
mattdailis, line 251: 🔶 Could you clarify this comment a little? It seems to me that even if there are several activities that are equal in properties, they would need to also have the same ids in order to be "matched" again.

mattdailis (changes requested):

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: `lastSimDriverResults` contains the temporal information as `Instant` and the constraint one as `Duration`.

mattdailis: Gotcha - I think we can convert from one to the other using `planningHorizon.toDur`, but I'll leave it up to you which one makes more sense.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: Yes it is related to this code. In the future, the fact that some activities have disappeared should change the execution flow. Maybe some goal becomes unsatisfied and the strategy could be to try processing this failed goal again to see if we can achieve the same level of satisfaction as before.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: In `equalsInProperties`, we do not check the id. It is possible that we do not associate the same activities at each run because we can't check the id.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: You're right. I'm gonna do that.

adrienmaillard: Actually, the start times of activities in simulation results were set with respect to the beginning of the planning horizon being Instant.now(). I have made a new commit to generate results in which the beginning of the planning horizon is passed to the generation of results.

mattdailis: That seems like a good idea :+1:

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
mattdailis: Right - so I would classify this under the same category as resources changing from simulation run to simulation run - it's not an error _unless_ it affects the satisfaction of a previously finished goal. Since we don't error for resources changing from run to run, do you think it might be possible to not error if the decomposition of an activity changes between two runs? Or does that create more problems than it solves?

adrienmaillard: I have commented out these lines and put a TODO to remember for an upcoming change related to this issue.

mattdailis (approved):
Tested locally, seems to be working :+1:

PR #334, opened 2022-09-19
[AERIE-2091] Fix "classic style" required `@Parameter` support

pcrosemurgy:
* **Tickets addressed:** AERIE-2091
* **Review:** By commit  

## Description

Previously the following required parameter would always result in an `InvalidArgumentsException` due to the internal argument value not being reassigned from (planner) provided arguments:
```java
@Parameter
public Integer z; // No default value specified, therefore this parameter is required
```

"Classic style" activities (defined by the `AllDefinedMethodMaker.java` generator) now correctly perform this assignment.

## Verification
Updated `MissionModelTest` and added a new `FooActivityTest` test.

`FooActivityMapper` now contains the following logic:
```java
          case "z":
            z = Optional.ofNullable(template.z = this.mapper_z.deserializeValue(entry.getValue())
                .getSuccessOrThrow(failure -> new UnconstructableArgumentException("z", failure)));
```
Previously the internal `z` value was not being reassigned:
```java
          case "z":
            template.z = this.mapper_z.deserializeValue(entry.getValue())
                .getSuccessOrThrow(failure -> new UnconstructableArgumentException("z", failure));
```

There's a chance that this bug was not caught sooner due to some ostensibly "required" parameters not actually being "required" due to their type being a Java primitive `int`/`long`/etc with a default value of `0`.

## Documentation
None.

## Future work
We could present the user with a warning message at compile time if they define a class member of a primitive Java type (`int`/`long`/etc.) with a `@Parameter` annotation and no assignment statement. For example:
```java
@Parameter
public int z; // No default value *explicitly* specified, however the type is `int` so this parameter will not be considered "required" by the planner
```

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/AllDefinedMethodMaker.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/AllDefinedMethodMaker.java
@@ -61,7 +61,9 @@ import java.util.Optional;
 
     methodBuilder = makeArgumentAssignments(methodBuilder, (builder, parameter) -> builder
         .addStatement(
-            "template.$L = this.mapper_$L.deserializeValue($L.getValue())$W.getSuccessOrThrow(failure -> new $T(\"$L\", failure))",
+            "$L = $T.ofNullable(template.$L = this.mapper_$L.deserializeValue($L.getValue())$W.getSuccessOrThrow(failure -> new $T(\"$L\", failure)))",
+            parameter.name,
+            Optional.class,
             parameter.name,
             parameter.name,
             "entry",
--- a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
+++ b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
@@ -24,150 +24,165 @@ public final class MissionModelTest {
 
   private DirectiveTypeRegistry<?, ?> registry;
 
-    @BeforeEach
-    public void initialize() {
-        this.registry = DirectiveTypeRegistry.extract(new GeneratedMissionModelFactory());
-    }
+  @BeforeEach
+  public void initialize() {
+      this.registry = DirectiveTypeRegistry.extract(new GeneratedMissionModelFactory());
+  }
 
-    @AfterEach
-    public void teardown() {
-      // TODO: [AERIE-1516] Teardown the model to release any system resources (e.g. threads).
-    }
+  @AfterEach
+  public void teardown() {
+    // TODO: [AERIE-1516] Teardown the model to release any system resources (e.g. threads).
+  }
 
-    @Test
-    public void shouldGetActivityTypeList() {
-        // GIVEN
-      final Map<String, ActivityType> expectedTypes = Map.of(
-            "foo", new ActivityType(
-                "foo",
-                List.of(
-                    new Parameter("x", ValueSchema.INT),
-                    new Parameter("y", ValueSchema.STRING),
-                    new Parameter("vecs", ValueSchema.ofSeries(ValueSchema.ofSeries(ValueSchema.REAL)))),
-                List.of(),
-                ValueSchema.ofStruct(Map.of())
-            ));
+  @Test
+  public void shouldGetActivityTypeList() {
+    // GIVEN
+    final Map<String, ActivityType> expectedTypes = Map.of(
+          "foo", new ActivityType(
+              "foo",
+              List.of(
+                  new Parameter("x", ValueSchema.INT),
+                  new Parameter("y", ValueSchema.STRING),
+                  new Parameter("z", ValueSchema.INT),
+                  new Parameter("vecs", ValueSchema.ofSeries(ValueSchema.ofSeries(ValueSchema.REAL)))),
+              List.of(),
+              ValueSchema.ofStruct(Map.of())
+          ));
 
-        // WHEN
-        final var activityTypes = new HashMap<String, ActivityType>();
-        registry.taskSpecTypes().forEach((name, specType) -> activityTypes.put(
-            name,
-            new ActivityType(name,
-                             specType.getParameters(),
-                             specType.getRequiredParameters(),
-                             specType.getReturnValueSchema())));
+    // WHEN
+    final var activityTypes = new HashMap<String, ActivityType>();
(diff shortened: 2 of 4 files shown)


pcrosemurgy:
Note that this PR's diff is slightly misleading: `MissionModelTest.java` had inconsistent indentation which was fixed along with the addition of the `FooActivity.z` parameter.

pcrosemurgy:
A diff of `MissionModelTest` excluding whitespace looks like:
```diff
diff --git a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
index 282bb575d..4a8f2806a 100644
--- a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
+++ b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
@@ -43,6 +43,7 @@ public final class MissionModelTest {
               List.of(
                   new Parameter("x", ValueSchema.INT),
                   new Parameter("y", ValueSchema.STRING),
+                  new Parameter("z", ValueSchema.INT),
                   new Parameter("vecs", ValueSchema.ofSeries(ValueSchema.ofSeries(ValueSchema.REAL)))),
               List.of(),
               ValueSchema.ofStruct(Map.of())
@@ -69,6 +70,7 @@ public final class MissionModelTest {
           List.of(
               new Parameter("x", ValueSchema.INT),
               new Parameter("y", ValueSchema.STRING),
+              new Parameter("z", ValueSchema.INT),
               new Parameter("vecs", ValueSchema.ofSeries(ValueSchema.ofSeries(ValueSchema.REAL)))),
           List.of(),
           ValueSchema.ofStruct(Map.of())
@@ -119,7 +121,8 @@ public final class MissionModelTest {
     // GIVEN
     final var typeName = "foo";
     final var parameters = new HashMap<>(Map.of("x", SerializedValue.of(0),
-                                                    "y", SerializedValue.of("test")));
+                                                "y", SerializedValue.of("test"),
+                                                "z", SerializedValue.of(1)));

     // WHEN
     final var activity = new SerializedActivity(typeName, parameters);
```

This can be generated by running:
```
$ git diff -w develop fix/aerie-2091-classic-activity-style-req-params -- merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
```diff
@@ -24,150 +24,153 @@ public final class MissionModelTest {
...
-    throws MissionModelService.NoSuchActivityTypeException, InvalidArgumentsException
-    {
-        // GIVEN
-        final var typeName = "foo";
-        final var parameters = new HashMap<>(Map.of("x", SerializedValue.of(0),
-                                                    "y", SerializedValue.of("test")));
-
-        // WHEN
+    // THEN
+    assertThat(thrown).isInstanceOf(InvalidArgumentsException.class);
```
mattdailis, line 155: 🔶 I think it would communicate the intent of this test better if we add a couple more assert statements here, showing that `thrown.missingArguments` contains `z`, and that `unconstructableArguments` contains `y`.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/AllDefinedMethodMaker.java
```diff
@@ -61,7 +61,9 @@ public MethodSpec makeInstantiateMethod() {
 
     methodBuilder = makeArgumentAssignments(methodBuilder, (builder, parameter) -> builder
         .addStatement(
-            "template.$L = this.mapper_$L.deserializeValue($L.getValue())$W.getSuccessOrThrow(failure -> new $T(\"$L\", failure))",
+            "$L = $T.ofNullable(template.$L = this.mapper_$L.deserializeValue($L.getValue())$W.getSuccessOrThrow(failure -> new $T(\"$L\", failure)))",
```
mattdailis, line 64: 🔵 As you're well aware - this assumption of `null` === "missing parameter" doesn't work if parameters are primitive types. We should make a ticket to handle those correctly (or, alternatively, Deprecate "Classic" style in favor of record style activities)

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
pcrosemurgy: Good idea! Will do

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/AllDefinedMethodMaker.java
Twisol: Deprecation has my vote 👀 If we had records when Aerie was started, I doubt we ever would have supported anything else.

Twisol: The one thing you can't do with records is do subclassing, because the Record superclass already fills the one inheritance slot Java gives us. I don't think this is something people should be doing with activities, but it would be worth scanning for such uses anyway.

pcrosemurgy: Just created this ticket to look into deprecating classic style activities and/or warning the mission modeler when defining a `@Parameter` of a primitive type with no explicit assignment statement: https://jira.jpl.nasa.gov/browse/AERIE-2122.

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/models/MissionModelTest.java
pcrosemurgy: Just pushed this change :)

mattdailis (approved):
Nice work, thanks Paul 💯

goetzrrGit (approved):
Thanks for walking this over with me in the huddle. Looks good to me

pcrosemurgy:
Thanks @goetzrrGit & @mattdailis!

PR #338, opened 2022-09-22
[FEAT](AERIE-2103) Convert seqJson to seq EDSL

dyst5422:
* **Tickets addressed:** AERIE-2103
* **Review:** By commit  

## Description

Added endpoint to convert SeqJson to SeqEDSL `/get-edsl-for-seqjson`

## Verification

Added some tests for the seqjson -> sequence and sequence -> seqedsl conversions.

## Documentation

- [ ] Need to add documentation on the new endpoint for API users

## Future work

I think we need to have clear documentation that this is more of a migration tool than something to be generally relied upon as part of standard workflows, preferring instead the SeqEDSL for human centric use cases. It doesn't retain formatting and comments the same way our SeqEDSL does. I think the exception to this is programmatically generated sequences from other tooling where SeqJson is a more appropriate interchange format


goetzrrGit (approved):
I really like how easy it is to go back and forth between seqjson and the EDSL. You really designed a great system!!!

camargo (approved):
LGTM! Tested locally and opened the corresponding UI PR: https://github.com/NASA-AMMOS/aerie-ui/pull/156/

PR #339, opened 2022-09-22
[AERIE-2115] Use IntervalMap for Linear and Discrete Profiles

JoelCourtney:
* **Tickets addressed:** AERIE-2115
* **Review:** By commit  

## Description

I've rebased this on top of the gaps in external profiles PR, and marked this as draft until that is merged.

This PR makes `IntervalMap` the common backing container for linear profiles, discrete profiles, and windows. This makes operations for future improvements to the constraints eDSL easier, simplifies some existing operations, and allows any future performance improvements to IntervalMap to benefit all three types.

I've also removed all the bounds arguments from profile operations. Previously, the plan bounds were passed into operations like `profile.plus(other, bounds)` or `profile.rate(bounds)`, so that the profile would bound its results. Now these calls are of the form `profile.rate().select(bounds)`, which has some performance penalty, but separates concerns. IntervalMap and the wrappers that use it don't care about planning horizons, so that concern is now dealt with by the constraints and scheduling AST evaluations.

Also, the value used by segments of LinearProfiles are now represented in point slope form. The time value of the point can be outside of the segment its associated with, without causing problems, but in practice that should never happen.

## Verification
All tests pass, and I verified that nothing broke in the UI by creating a simple plan in banananation and running a resource-based constraint.

## Future work
Use our fancy new profiles


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalMap.java
```diff
@@ -213,6 +235,27 @@ IntervalMap<R> map2(
...
+   * @param right right operand
+   * @param transform a function that transforms an interval and two {@link Optional}s of the left and right operands' types to an
+   *                  optional of a new type.
+   * @param <V1> value type of the left operand
+   * @param <V2> value type of the right operand
+   * @param <R> value type of the result
+   * @return a new IntervalMap, the result of applying the transform
+   */
+  public static <V1, V2, R>
+  IntervalMap<R> map2(
```
mattdailis, line 261: This new variant of `map2` doesn't seem to be used anywhere - is there a reason to keep it?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/resources/StateQueryParam.java
```diff
@@ -37,27 +38,27 @@ public SerializedValue getValue(SimulationResults simulationResults, Plan plan,
...
       //TODO: improve the profile data structure to allow fast time-keyed query
       //TODO: improve Interval to allow querying containment in interval directly
       //for now we look for the last matching profile segment, if any (last to get latest if any overlaps)
       final var profile = simulationResults.discreteProfiles.get(this.resourceName);
-      final Predicate<DiscreteProfilePiece> containsQueryTimeP = (piece)
-          -> !Interval.intersect(piece.interval, Interval.at(queryT)).isEmpty();
-      final var matchPiece = Lists.reverse(profile.profilePieces).stream()
+      final Predicate<Segment<SerializedValue>> containsQueryTimeP = (piece)
+          -> !Interval.intersect(piece.interval(), Interval.at(queryT)).isEmpty();
+      final var matchPiece = profile.profilePieces.stream()
```
mattdailis: Just trying to understand - is the reason this reverse is unnecessary because there can never be overlaps? (And there never were overlaps before, right, so this code was overly defensive about that?).

There's a comment a few lines up that becomes stale after this change - I suggest we just delete it to reduce future confusion. (The one about "for now we look for the last matching profile segment, if any")

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalMap.java
JoelCourtney: I created it because I thought I had a use case when refactoring the operations on linear profiles, but when I changed it to using point-slope form I no longer needed it. So while its currently not used, its not unreasonable.

Also it mirrors the variants for `map`, which are all used.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/resources/StateQueryParam.java
JoelCourtney: Yes, I think the reverse was there in case of overlaps, which wasn't really a concern before either. I'll edit the comment

JoelCourtney: updated comment, also improved the predicate. (Interval's interface has improved since this was written.)

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalMap.java
```diff
@@ -196,6 +197,27 @@ public <R> IntervalMap<R> map(final BiFunction<V, Interval, R> transform) {
     return builder.build();
   }
 
+  /**
+   * A similar operation to {@link IntervalMap#map(Function)}, except it allows producing any number of segments
+   * per interval. Each segment is then flattened into a new IntervalMap.
+   */
+  public <R> IntervalMap<R> flatMap(final Function<V, Stream<Segment<R>>> transform) {
```
mattdailis, line 207: Just musing - it is important that the segments returned by `transform` be contained in the original overlapping interval?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -1,203 +1,134 @@
...
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
-public final class LinearProfile implements Profile<LinearProfile> {
+public final class LinearProfile implements Profile<LinearProfile>, Iterable<Segment<LinearEquation>> {
   // IMPORTANT: Profile pieces must be non-overlapping, and increasing (based on interval field)
```
mattdailis: I think this comment is now stale, since this invariant is now the responsibility of `IntervalMap`

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -1,203 +1,134 @@
...
-  public LinearProfile(final List<LinearProfilePiece> profilePieces) {
+  public LinearProfile(final IntervalMap<LinearEquation> profilePieces) {
     this.profilePieces = Objects.requireNonNull(profilePieces);
   }
 
-  public LinearProfile(final LinearProfilePiece... profilePieces) {
-    this(List.of(profilePieces));
+  @SafeVarargs
+  public LinearProfile(final Segment<LinearEquation>... profilePieces) {
+    this.profilePieces = IntervalMap.of(profilePieces);
```
mattdailis: Thoughts on having these two constructors delegate to `this(profilePieces)`?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
```diff
@@ -0,0 +1,180 @@
...
+      builder.set(
+          Interval.between(Duration.MIN_VALUE, Inclusive, t, Exclusive),
+          false
+      );
+      builder.set(Interval.at(t), true);
+      builder.set(
+          Interval.between(t, Exclusive, Duration.MAX_VALUE, Inclusive),
+          false
+      );
+      return new Windows(builder.build());
```
mattdailis: Consider replacing the builder here with the Windows varargs constructor:

```suggestion
      return new Windows(
          Segment.of(Interval.between(Duration.MIN_VALUE, Inclusive, t, Exclusive), false),
          Segment.of(Interval.at(t), true),
          Segment.of(Interval.between(t, Exclusive, Duration.MAX_VALUE, Inclusive), false));
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -1,203 +1,134 @@
...
-        if (rightIter.hasNext()) {
-          rightPiece = rightIter.next();
+    public Windows changePoints() {
+      final var result = IntervalMap.<Boolean>builder().set(this.profilePieces.map(LinearEquation::changing));
+      for (int i = 0; i < this.profilePieces.size(); i++) {
+        final var segment = this.profilePieces.get(i);
+        final var startTime = segment.interval().start;
+        if (i == 0) {
+          if (!segment.interval().contains(Duration.MIN_VALUE)) {
+            result.unset(Interval.at(segment.interval().start));
```
mattdailis: It seems like `startTime` is defined to be `setment.interval().start()`, but then isn't used here. I would suggest either using `startTime` here, or inlining `segment.interval().start` into all uses of `startTime`.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -1,203 +1,134 @@
...
-      while (Interval.compareEndToStart(rightPiece.interval, leftPiece.interval) < 0) {
-        if (rightIter.hasNext()) {
-          rightPiece = rightIter.next();
+    public Windows changePoints() {
+      final var result = IntervalMap.<Boolean>builder().set(this.profilePieces.map(LinearEquation::changing));
+      for (int i = 0; i < this.profilePieces.size(); i++) {
+        final var segment = this.profilePieces.get(i);
+        final var startTime = segment.interval().start;
+        if (i == 0) {
+          if (!segment.interval().contains(Duration.MIN_VALUE)) {
```
mattdailis, line 112: Could you help me understand why we need to check for `Duration.MIN_VALUE` here? This may underlie my lack of understanding for what exactly `changePoints` means for a linear profile in the presence of gaps.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
```diff
@@ -1,203 +1,134 @@
...
-            rightPiece = rightIter.next();
-            continue;
+          final var previousSegment = this.profilePieces.get(i-1);
+
+          if (Interval.meets(previousSegment.interval(), segment.interval())) {
+            if (previousSegment.value().valueAt(startTime) != segment.value().valueAt(startTime)) {
+              result.set(Interval.at(segment.interval().start), true);
+            }
+          } else {
+            result.unset(Interval.at(segment.interval().start));
```
mattdailis: Still trying to wrap my head around what `changePoints` means in the presence of gaps, but it does seem strange to me that we're unsetting `segment.interval().start` but not `previousSegment.interval().end`. Thoughts?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
```diff
@@ -0,0 +1,180 @@
...
+  public boolean changing() {
+    return this.rate != 0;
+  }
+
+  private Optional<Duration> intersectionPointWith(final LinearEquation other) {
+    if (this.rate == other.rate) return Optional.empty();
+    return Optional.of(
+        this.initialTime.plus(
+            Duration.roundNearest(
+                (other.valueAt(this.initialTime) - this.initialValue) / (this.rate - other.rate)*1000000,
```
mattdailis: Why do we need a `1000000` scaling factor here? Could we instead use a different unit on the following line?

If we do keep the scaling factor, it would help my understanding if we could add parentheses to this expression - I'm not sure off the top of my head whether this is `a / (b * c)` or `(a / b) * c`.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
JoelCourtney: gaps mean "unknown" value. So if a segment is surrounded by gaps, we don't know if the first point of the segment is different from the point before, but we do know if the last point of the segment is different from the point before.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
JoelCourtney: I guess we don't have a consensus about what `Duration.MIN_VALUE` and `Duration.MAX_VALUE` mean conceptually. I see two options

1. The extrema refer to specific times, so no segments can extend beyond MIN and MAX.
2. Any segment that touches MIN and MAX is probably _intended_ to be unbounded, because realistically no bounded segment is going to extend that far. We can't actually represent unbounded yet (that's a future improvement I want to do)

I prefer the second interpretation, for that reason and because `shiftBy` under interpretation 1 would lead to unwanted gaps around the extremes. I've been making changes assuming interpretation 2.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
mattdailis: Ah, I see - the key definition I was missing was the definition of changePoint:
> a changePoint is defined as any point whose value is known to be different from the point before.

Through that lens, this method makes sense :+1: (I was thinking maybe it was a point where a change occurs either before or after, and was getting tied up in a knot. I was also wondering whether the 'clusivity mattered....). 

Perhaps a comment somewhere with this definition would help? Either in the constraints documentation, or just a javadoc comment either on this method, or on the `Profile` interface.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
mattdailis: That makes sense, thanks for the explanation :+1: I'm okay with treating Duration.MIN_VALUE as "negative infinity" for the time being, as long as we're careful and consistent about it.

JoelCourtney:
I've rebased this on top of the gaps in external profiles PR, and marked this as draft until that is merged.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalMap.java
JoelCourtney: As long as the intervalmap user is OK with last-one-wins behavior, no.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearProfile.java
JoelCourtney: No opinion. I'll make the change

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
JoelCourtney: The rate parameter has the units `1/second`. I'll try to make it a little more readable, and use `Duration.SECOND` instead

camargo (approved):

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
```diff
@@ -0,0 +1,174 @@
...
+  private Optional<Duration> intersectionPointWith(final LinearEquation other) {
+    if (this.rate == other.rate) return Optional.empty();
+    return Optional.of(
+        this.initialTime.plus(
+            Duration.roundNearest(
+                RATE_CONVERSION_TO_MICROSECONDS * (other.valueAt(this.initialTime) - this.initialValue)
+                  / (this.rate - other.rate),
+                Duration.MICROSECONDS
+            )
+        )
```
mattdailis: I _think_ this is equivalent, and a bit clearer:

```suggestion
        this.initialTime.plus(
            Duration.roundNearest(
                (other.valueAt(this.initialTime) - this.initialValue)
                  / (this.rate - other.rate),
                Duration.SECONDS
            )
        )
```

After reading `roundNearest`, I interpret it to mean essentially `Duration.of(float, unit)`, with the understanding that it will be rounded to the nearest microsecond.

mattdailis (approved):

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
JoelCourtney: Ah that's a misleading function name. Ok, will do

PR #340, opened 2022-09-23
Include stack trace in failure reason

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

When simulation or scheduling fails, we set the status to "failed", and include a reason. When it fails in an unexpected way, we save only the name of the exception, or, in the best case, "exception.toString()". This makes diagnosing the issue into a game of 20 questions, wasting the time of everyone involved.

This PR adds a new variant of `failWith` that accepts a `Throwable`, collects the stack trace into a string, and passes it along to the existing `failWith(String)` method. I updated the small handful of places that were stringifying exceptions before calling `failWith`.

I got the approach for stringifying an exception from [this stackoverflow answer](https://stackoverflow.com/a/1149712).

## Verification

I intentionally introduced a typo into one of the database actions, and checked to see that the full stack trace was logged to the console:

Screen Shot 2022-09-22 at 8 48 17 PM

## Documentation

We should describe our error visibility philosophy in some documentation at some point - in particular our differentiation between business logic errors intended to be addressed by users, and system errors intended to be addressed by deployment engineers or the Aerie team itself.

## Future work

There is no distinction between the kinds of errors from the UI's perspective - we should distinguish between user error and system error, so we can display an informative toast for one, and log to the console for the other, perhaps displaying a toast with "something went wrong, please call for help".

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
@@ -2,6 +2,9 @@ package gov.nasa.jpl.aerie.merlin.server;
 
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
 public final class ResultsProtocol {
   private ResultsProtocol() {}
 
@@ -35,6 +38,12 @@ public final class ResultsProtocol {
     //   of the underlying resource in order to deallocate it.
     void succeedWith(SimulationResults results);
     void failWith(String reason);
+
+    default void failWith(final Throwable throwable) {
+      final var stringWriter = new StringWriter();
+      throwable.printStackTrace(new PrintWriter(stringWriter));
+      this.failWith(stringWriter.toString());
+    }
   }
 
   public interface OwnerRole extends ReaderRole, WriterRole {}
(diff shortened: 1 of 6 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
```diff
@@ -35,6 +38,11 @@ public interface WriterRole {
     //   of the underlying resource in order to deallocate it.
     void succeedWith(SimulationResults results);
     void failWith(String reason);
+    default void failWith(final Throwable throwable) {
```
Twisol, line 42: A newline between the abstract methods and the default (helper) method might be nice. 😃

Twisol (approved):
Nice -- looks great to me!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
mattdailis: Newline added 🖖

camargo:
Nice. Are these more verbose errors stored in the database?

mattdailis:
> Nice. Are these more verbose errors stored in the database?

Yep :+1: They're in the same "reason" column as before - it's just a longer string now.

camargo (approved):

PR #345, opened 2022-09-28
docs: add ui-custom-base-path document

camargo:
## Description

* Add doc for building a custom UI image with a base path
* Add docs for specifying activity directive metadata
* Add `sphinx-markdown-tables` extension for displaying tables via markdown extensions
* Commit lock file update
* Remove unused aws-deployment page and associated images


Mythicaeda (approved):
Looks good! Just one quick request: Can you also remove the `aws-deployment` page? (Or `:orphan:` it and remove it from deployment's `index.rst` if we intend to use it later.) It has no content aside from a header, and many of the image files you deleted look like they would be associated with that page.

PR #348, opened 2022-09-28
[AERIE-2111] Return AnalysisID as part of Schedule Action

Mythicaeda:
* **Tickets addressed:** AERIE-2111
* **Review:** By commit  

## Description

The `schedule` Hasura action now includes the `analysisId` of the request in the returned JSON. Additionally, the unused `UncachedSchedulerService` and `InMemoryResultsCell` classes were removed.

## Verification

Manually checked the return of `Schedule` in Hasura after run scheduling tasks on several plans.

The "Run Scheduling" test in `scheduler.test.ts` now checks that `analysisId` is returned and does not change while awaiting an incomplete request to become complete. The GQL query it runs was modified so that `analysisId` is returned with the other fields.


pcrosemurgy (approved):
Looks good!

I'm cool with your removal of `scheduler/server/remotes/InMemoryResultsCell.java` and `scheduler/server/services/UncachedSchedulerService.java` as they're just stubs but I think we should create a proper ticket for implementing these if we eventually want to have them around (ultimately to support unit testing, for example unit testing the scheduler HTTP endpoints without spinning up PG).

skovati (approved):
LGTM! Thanks for removing those old classes as well

PR #351, opened 2022-09-30
[FIX][AERIE-2132] Incremental simulation incorrectly updated its timeline 

adrienmaillard:
* **Tickets addressed:** AERIE-2132
* **Review:** By commit  

## Description

When evaluating the anchor expression (a simple `GreaterThan` or `LessThan`) of a `CoexistenceGoal`, the expression returned no intervals instead of 2 as expected.

When looking at the `SimulationResults` produced, the value of the resource was never changing during the planning horizon (which is incorrect, the value is actually changing) and thus never moved to regions where the expression would evaluate to `True`. 

There are two methods for making the simulation go forward in `IncrementalSimulationDriver` : 
- `simulateSchedule` taking a list of activities as argument and simulating them until the end of the last one
- `simulateUntil` taking only an end time as argument and simulating until this time 

In absence of activities, `simulateUntil` was used. And this method was incorrectly updating the timeline with a delta always equal to 0. And thus values were never updated. This is the bug. It is fixed in the first commit. 

The second commit is not related to the bug. It moves methods about conversion of `SimulationResults` out of the `SimulationFacade` to make the latter more readable and maintainable. It also correctly sets the start timestamp of results with the start of the planning horizon. 

## Verification

Simulation results coming from the `SimulationDriver` and `IncrementalSimulationDriver` were compared on a 0.13.0 clipper-specific mission model containing resources that evolve with time. 
Can't add this test because it depends on this specific mission model. 

## Documentation

Nothing

## Future work

None

### DIFF
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
@@ -106,6 +106,14 @@ public class IncrementalSimulationDriver<Model> {
   }
 
 
+  /**
+   * Simulate an activity
+   * @param activity the activity to simulate
+   * @param startTime the start time of the activity
+   * @param activityId the activity id for the activity to simulate
+   * @throws TaskSpecType.UnconstructableTaskSpecException
+   * @throws InvalidArgumentsException
+   */
   public void simulateActivity(SerializedActivity activity, Duration startTime, ActivityInstanceId activityId)
   throws TaskSpecType.UnconstructableTaskSpecException, InvalidArgumentsException
   {
@@ -130,11 +138,11 @@ public class IncrementalSimulationDriver<Model> {
 
   /**
    * Get the simulation results from the Duration.ZERO to the current simulation time point
-   * @param startTime the epoch start time for these results
+   * @param startTimestamp the timestamp for the start of the planning horizon. Used as epoch for computing SimulationResults.
    * @return the simulation results
    */
-  public SimulationResults getSimulationResults(Instant startTime){
-    return getSimulationResultsUpTo(curTime, startTime);
+  public SimulationResults getSimulationResults(Instant startTimestamp){
+    return getSimulationResultsUpTo(startTimestamp, curTime);
   }
 
   public Duration getCurrentSimulationEndTime(){
@@ -144,20 +152,20 @@ public class IncrementalSimulationDriver<Model> {
   /**
    * Get the simulation results from the Duration.ZERO to a specified end time point.
    * The provided simulation results might cover more than the required time period.
-   * @param endTime the duration onto which these results are computed
-   * @param startTime the epoch start time for these results
+   * @param startTimestamp the timestamp for the start of the planning horizon. Used as epoch for computing SimulationResults.
+   * @param endTime the end timepoint. The simulation results will be computed up to this point.
    * @return the simulation results
    */
-  public SimulationResults getSimulationResultsUpTo(Duration endTime, Instant startTime){
+  public SimulationResults getSimulationResultsUpTo(Instant startTimestamp, Duration endTime){
     //if previous results cover a bigger period, we return do not regenerate
     if(endTime.longerThan(curTime)){
       simulateUntil(endTime);
     }
 
-    if(lastSimResults == null || endTime.longerThan(lastSimResultsEnd)) {
+    if(lastSimResults == null || endTime.longerThan(lastSimResultsEnd) || startTimestamp.compareTo(lastSimResults.startTime) != 0) {
       lastSimResults = SimulationEngine.computeResults(
           engine,
-          startTime,
+          startTimestamp,
           endTime,
           activityTopic,
           timeline,
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
@@ -52,7 +52,6 @@ public class IncrementalSimulationDriver<Model> {
     this.missionModel = missionModel;
     plannedDirectiveToTask = new HashMap<>();
     initSimulation();
-    simulateUntil(Duration.ZERO);
   }
 
   /*package-private*/ void initSimulation(){
@@ -95,8 +94,8 @@ public class IncrementalSimulationDriver<Model> {
       if(batch.offsetFromStart().longerThan(endTime) || endTime.isEqualTo(Duration.MAX_VALUE)){
         break;
       }
-      curTime = batch.offsetFromStart();
       final var delta = batch.offsetFromStart().minus(curTime);
+      curTime = batch.offsetFromStart();
       timeline.add(delta);
       // Run the jobs in this batch.
       final var commit = engine.performJobs(batch.jobs(), cells, curTime, Duration.MAX_VALUE, missionModel);
--- a/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
+++ b/scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
@@ -1,41 +1,26 @@
 package gov.nasa.jpl.aerie.scheduler.simulation;
 
-import com.google.common.collect.Maps;
-import gov.nasa.jpl.aerie.constraints.model.DiscreteProfile;
-import gov.nasa.jpl.aerie.constraints.model.DiscreteProfilePiece;
-import gov.nasa.jpl.aerie.constraints.model.LinearProfile;
-import gov.nasa.jpl.aerie.constraints.model.LinearProfilePiece;
-import gov.nasa.jpl.aerie.constraints.time.Interval;
 import gov.nasa.jpl.aerie.merlin.driver.ActivityInstanceId;
 import gov.nasa.jpl.aerie.merlin.driver.MissionModel;
 import gov.nasa.jpl.aerie.merlin.driver.SerializedActivity;
-import gov.nasa.jpl.aerie.merlin.driver.SimulatedActivity;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
 import gov.nasa.jpl.aerie.merlin.protocol.model.TaskSpecType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
 import gov.nasa.jpl.aerie.merlin.protocol.types.InvalidArgumentsException;
-import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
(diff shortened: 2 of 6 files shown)


scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
```diff
@@ -70,7 +70,7 @@ private static Problem makeTestMissionAB() {
     return new Problem(fooMissionModel, h, new SimulationFacade(h, fooMissionModel), SimulationUtility.getFooSchedulerModel());
   }
 
-  private final static PlanningHorizon h = new PlanningHorizon(TimeUtility.fromDOY("2025-001T01:01:01.001"), TimeUtility.fromDOY("2030-001T01:01:01.001"));
+  private final static PlanningHorizon h = new PlanningHorizon(TimeUtility.fromDOY("2025-001T01:01:01.001"), TimeUtility.fromDOY("2025-005T01:01:01.001"));
```
Mythicaeda, line 73: Why is the duration of this test's PlanningHorizon being changed from five years to four days?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
```diff
@@ -133,16 +132,16 @@ public void simulateActivity(SerializedActivity activity, Duration startTime, Ac
...
    * @return the simulation results
    */
-  public SimulationResults getSimulationResults(){
-    return getSimulationResultsUpTo(curTime);
+  public SimulationResults getSimulationResults(Instant startTimestamp){
+    return getSimulationResultsUpTo(startTimestamp, curTime);
   }
 
   /**
    * Get the simulation results from the Duration.ZERO to a specified end time point.
```
Mythicaeda, line 153: Does this comment still accurately describe this method's behavior? Just checking since `Duration.ZERO` doesn't appear to be used here.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationResultsConverter.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.stream.Collectors;
+
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
+
+public class SimulationResultsConverter {
+
+  /**
+   * convert a simulation driver SimulationResult to a constraint evaluation engine SimulationResult
+   *
+   * @param driverResults the recorded results of a simulation run from the simulation driver
```
Mythicaeda, line 30: [Minor:] Can this comment please be updated to include the new parameter?

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationResultsConverter.java
```diff
@@ -0,0 +1,105 @@
...
+   * @param id the name of the activity instance
+   * @param driverActivity the completed activity instance details from a driver SimulationResult
+   * @return an activity instance suitable for a constraint model SimulationResult
+   */
+  public static gov.nasa.jpl.aerie.constraints.model.ActivityInstance convertToConstraintModelActivityInstance(
+      long id, SimulatedActivity driverActivity, final Instant startTime)
+  {
+    final var startT = Duration.of(startTime.until(driverActivity.start(), ChronoUnit.MICROS), MICROSECONDS);
+    final var endT = startT.plus(driverActivity.duration());
+    final var activityWindow = startT.isEqualTo(endT)
```
Mythicaeda: Why was this variable renamed? It's of type `Interval`, not `Window`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -83,11 +68,11 @@ public Optional<Duration> getActivityDuration(ActivityInstance activityInstance)
       logger.error("You need to simulate before requesting activity duration");
       return Optional.empty();
     }
-    var duration = driver.getActivityDuration(planActInstanceIdToSimulationActInstanceId.get(activityInstance.getId()));
-    if(duration.isEmpty()){
-      logger.error("Incremental simulation is probably outdated, check that no activity is removed between simulation and querying");
-    }
-    return duration;
+     var duration = driver.getActivityDuration(planActInstanceIdToSimulationActInstanceId.get(activityInstance.getId()));
```
Mythicaeda: [Minor:] Does the whitespace change here have significance?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
```diff
@@ -36,11 +37,11 @@ public void init() throws TaskSpecType.UnconstructableTaskSpecException, Invalid
   @Test
   public void simulationResultsTest(){
     //ensures that simulation results are generated until the end of the last act;
-    var simResults = incrementalSimulationDriver.getSimulationResults();
+    var simResults = incrementalSimulationDriver.getSimulationResults(Instant.now());
     assert(simResults.realProfiles.get("/utcClock").getRight().get(0).getLeft().isEqualTo(endOfLastAct));
     /*ensures that when current simulation results cover more than the asked period and that nothing has happened
     between two requests, the same results are returned*/
-    var simResults2 = incrementalSimulationDriver.getSimulationResultsUpTo(Duration.of(7,SECONDS));
+    var simResults2 = incrementalSimulationDriver.getSimulationResultsUpTo(Instant.now(), Duration.of(7, SECONDS));
```
Mythicaeda, line 55: [Clarification:] Does `simResults2` return the same object in `simResults` because its duration is a subset of `simResults`'s?

JoelCourtney (approved):
@Mythicaeda's review covered all of my comments, so I approve

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
adrienmaillard: Because simulating 5 years of mission model is too much in terms of computing time. When scheduling starts, it needs to simulate the whole planning horizon (for evaluating coexistence expressions for example).

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
adrienmaillard: Yes. `Duration.ZERO` is the origin for simulation. The interval is `[Duration.ZERO, curTime]`.

adrienmaillard: `startTimeStamp` is there only to set the start timestamps of activities in `SimulationResults`

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationResultsConverter.java
adrienmaillard: I have renamed it to `activityInterval`.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: No, I am reverting this.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
Mythicaeda: Yes, but isn't running scheduling over a plan multiple years long a use case of ours? If so, we need to be testing how the scheduler handles that. Is there another test that runs covers this?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationTest.java
adrienmaillard: Absolutely, it is not recomputed because of what you mentioned.

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
adrienmaillard: Performance is definitely a concern but I am not sure we should have to wait for a long time every time we run unit tests.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
Mythicaeda: Ah gotcha! Could you please add the `@param`s for the parameters here and on the function directly above?

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
Mythicaeda: My concern is less with performance and more that if we never test the scheduler against a duration greater than a few days, then we can end up missing bugs/issues that only appear on longer plans.

scheduler/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/IncrementalSimulationDriver.java
adrienmaillard: Done

scheduler/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
adrienmaillard: I do not think simulation has a month/year-long test. Maybe we could have another type of tests that would not be ran with other unit tests but only in some context ?

mattdailis: Did this test recently start to take a longer time to run? Or has it always been taking a long time? Also - how long is a long time?

adrienmaillard: Yes, it started to take a much longer time as a consequence of this fix because the timeline is now accurately go forward in time. A 5-year simulation is a lot. I am not sure I did see the end of it.

adrienmaillard: And if you wonder why: there are two methods actually triggering simulation. One with activities. One without activities. The method that is fixed here is the second one. These tests start with empty plans. Which trigger the faulty method (as there are no activities) to simulate the whole horizon before computing conflicts.

adrienmaillard: I have made one change concerning returning different simulation results if the start timestamp is different. And added tests about that. And added a 5-year test with Banananation.

Mythicaeda (approved):

PR #355, opened 2022-10-03
Add logo and favicon to docs

jeffpamer:
* **Tickets addressed:** AERIE-2079
* **Review:** By commit

## Description
Added the Aerie logo to the header and footer of the docs template, as well as included the favicon. This uses the same favicon combination as the `aerie-ui` project: Users should get a favicon that matches their light/dark system theme preference. Browsers that don't support SVG favicons (i.e. Safari) will fallback to a default icon that works against either a dark or light background.

image


docs/sphinx_scylladb_theme/static/css/main.css
```diff
@@ -1 +1 @@
...
...33333%}.grid-margin-x>.medium-offset-1{margin-left:calc(8.33333% + .9375rem)}.medium-offset-2{margin-left:16.66667%}.grid-margin-x>.medium-offset-2{margin-left:calc(16.66667% + .9375rem)}.medium-offset-3{margin-left:25%}.grid-margin-x>.medium-offset-3{margin-left:calc(25% + .9375rem)}.medium-offset-4{margin-left:33.33333%}.grid-margin-x>.medium-offset-4{margin-left:calc(33.33333% + .9375rem)}.medium-offset-5{margin-left:41.66667%}.grid-margin-x>.medium-offset-5{margin-left:calc(41.66667% + .9375rem)}.medium-offset-6{margin-left:50%}.grid-margin-x>.medium-offset-6{margin-left:calc(50% + .9375rem)}.medium-offset-7{margin-left:58.33333%}.grid-margin-x>.medium-offset-7{margin-left:calc(58.33333% + .9375rem)}.medium-offset-8{margin-left:66.66667%}.grid-margin-x>.medium-offset-8{margin-left:calc(66.66667% + .9375rem)}.medium-offset-9{margin-left:75%}.grid-margin-x>.medium-offset-9{margin-left:calc(75% + .9375rem)}.medium-offset-10{margin-left:83.33333%}.grid-margin-x>.medium-offset-10{margin-left:calc(83.33333% + .9375rem)}.medium-offset-11{margin-left:91.66667%}.grid-margin-x>.medium-offset-11{margin-left:calc(91.66667% + .9375rem)}}@media print,screen and (min-width:64em){.large-offset
```
camargo: I'm wondering if we should just Prettier format this file since we will be editing it. It will make reviewing changes easier which is probably more important than the ~33kb of space we save by removing the whitespace.

camargo: Since we are leaving the `.hero__img` div empty can we remove the `max-width` from `.hero__title`? It will make the "Welcome to Aerie Documentation" title cleaner I think by taking up more horizontal space.

jeffpamer: For a while I was assuming there was a non-minified version of the stylesheet that ended up compiled into this at some point, but I couldn't find evidence of that anywhere. So if we may be doing additional edits here I definitely agree we should go ahead and prettier/un-minify it.

camargo: Yeah unfortunately I don't think `main.css` (or anything in that static directory for that matter) is compiled. It looks like it's just statically included as part of the `sphinx_scylladb_theme`.

camargo (approved):
Looks so good thank you Jeff!

Mythicaeda (commented):
Looks awesome! One quick thing, though: I opening the site in incognito-Chrome to test the light-vs-dark logo, and it used the light logo. Is this supposed to happen? 
image

jeffpamer:
> Looks awesome! One quick thing, though: I opening the site in incognito-Chrome to test the light-vs-dark logo, and it used the light logo. Is this supposed to happen?

Ohh so I'm guessing you have a light system theme, but that dark background is coming from the incognito window specifically, is that right? I'm not sure how we could handle that case ...

Might need to do a bit of research to see if there's anything we can do to support that.

Mythicaeda:
> Ohh so I'm guessing you have a light system theme, but that dark background is coming from the incognito window specifically, is that right? I'm not sure how we could handle that case ...

Yes, although this persists when I changed my Chrome theme color from "default" to "black", so it might be a side effect of Chrome not considering any of its darker-colored themes to be a "Dark" mode? Rather than specifically an Incognito thing.

jeffpamer:
> Yes, although this persists when I changed my Chrome theme color from "default" to "black", so it might be a side effect of Chrome not considering any of its darker-colored themes to be a "Dark" mode? Rather than specifically an Incognito thing.

Ah yep that makes sense. The icon is using a media query to choose colors based on the OS theme, which I don't think would be altered by changing the Chrome theme specifically. Might be worth some exploration to see if there's any way to handle that, but it doesn't seem _super_ likely.

camargo:
Going to merge to keep the docs moving, and per @jeffpamer: we can address support for non-os-level dark themes as an edge case at a later time.

PR #362, opened 2022-10-06
[AERIE-2119] Structure `failWith()` reasons

pcrosemurgy:
* **Tickets addressed:** AERIE-2119
* **Review:** By commit  

## Description

Motivated by [discussion](https://jpl.slack.com/archives/GJ07FGRPH/p1664308622371219) with @camargo and @duranb. The UI and API clients would like to have some structure within our `simulation_dataset.reason` and `scheduling_request.failure_reason` columns.

This PR introduces a serialization of these `reason`s with `ScheduleFailure` and `SimulationFailure`. Notably, there's now a `message` field to act as a human-readable summary of the failure.

This PR also renames `scheduling_request.failure_reason` -> `scheduling_request.reason`.

## Verification

- End-to-end tests passing
- Manual inspecting `simulation_dataset.reason` and `scheduler_request.failure_reason` columns after end-to-end tests.

After simulating with an activity (`BiteBanana`) with extraneous arguments `simulation_dataset.reason` contains:
```json
{
    "type": "PLAN_CONTAINS_UNCONSTRUCTABLE_ACTIVITIES",
    "message": "Plan contains unconstructable activities",
    "data": {
        "success": false,
        "errors": {
            "2": {
                "reason": {
                    "success": false,
                    "errors": {
                        "extraneousArguments": [
                            "bad_three",
                            "bad_one",
                            "bad_two"
                        ],
                        "unconstructableArguments": [],
                        "missingArguments": []
                    },
                    "arguments": {
                        "biteSize": 1.0
                    }
                }
            }
        }
    },
    "trace": "",
    "timestamp": "2022-10-10T18:05:27.5648916Z.564891"
}

```

@camargo / @duranb: Note that the inner `data.errors` in this case is now the exact same structure that's reported from `getActivityEffectiveArguments`/`getModelEffectiveArguments`. This should hopefully make ingesting this data easier.

## Documentation

TODO: will update this section once this PR has general/informal approval.

## Future work
- Find other services/areas where our error reporting could be more human-friendly.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ScheduleFailure.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ScheduleFailure.java
@@ -1,0 +1,45 @@
+package gov.nasa.jpl.aerie.scheduler.server.services;
+
+import javax.json.JsonValue;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public record ScheduleFailure(
+    String type,
+    String message,
+    JsonValue data,
+    String trace
+) {
+  public static final class Builder {
+    private String type = "";
+    private String message = "";
+    private String trace = "";
+    private JsonValue data = JsonValue.EMPTY_JSON_OBJECT;
+
+    public Builder type(final String type) {
+      this.type = type;
+      return this;
+    }
+
+    public Builder message(final String message) {
+      this.message = message;
+      return this;
+    }
+
+    public Builder trace(final Throwable throwable) {
+      final var sw = new StringWriter();
+      throwable.printStackTrace(new PrintWriter(sw));
+      this.trace = sw.toString();
+      return this;
+    }
+
+    public Builder data(final JsonValue data) {
+      this.data = data;
+      return this;
+    }
+
+    public ScheduleFailure build() {
+      return new ScheduleFailure(type, message, data, trace);
+    }
+  }
+}
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ScheduleFailure.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ScheduleFailure.java
@@ -3,12 +3,14 @@ package gov.nasa.jpl.aerie.scheduler.server.services;
 import javax.json.JsonValue;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.time.Instant;
 
 public record ScheduleFailure(
     String type,
     String message,
     JsonValue data,
-    String trace
+    String trace,
+    Instant timestamp
 ) {
   public static final class Builder {
     private String type = "";
@@ -39,7 +41,7 @@ public record ScheduleFailure(
     }
 
     public ScheduleFailure build() {
-      return new ScheduleFailure(type, message, data, trace);
+      return new ScheduleFailure(type, message, data, trace, Instant.now());
     }
   }
 }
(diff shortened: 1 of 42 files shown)


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -169,7 +173,10 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
         }
       }
       if (!failedGoals.isEmpty()) {
-        writer.failWith(failedGoals.toString());
+        writer.failWith(b -> b
+            .type("SCHEDULING_GOALS_FAILED")
```
pcrosemurgy, line 177: It's maybe not ideal that these `type` fields are hardcoded string literals but I figured this is pretty innocuous right now as
1. there are no duplicate failure types right now, i.e., there's only ever one `failWith()` call per failure type.
1. every `failWith()` invocation is within this method – with the exception of one `failWith()` within the `ThreadedSchedulerAgent` that is called as a catch-all for any `Throwable`.

mattdailis: > there's only ever one failWith() call per failure type

I'm in favor of this property because I think it will help make the errors more traceable 👍

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ThreadedSchedulerAgent.java
```diff
@@ -64,7 +64,9 @@ public void run() {
               this.schedulerAgent.schedule(req.request(), req.writer());
             } catch (final Throwable ex) {
               ex.printStackTrace(System.err);
-              req.writer().failWith(ex);
+              req.writer().failWith(b -> b
+                  .type("UNEXPECTED_SCHEDULER_EXCEPTION")
```
mattdailis, line 68: 🔵 I'm wondering if it's worth having a message here - "Something went wrong while scheduling" - I could go either way.

mattdailis (approved):
Looks good to me overall

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ScheduleFailure.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public Builder trace(final Throwable throwable) {
+      final var sw = new StringWriter();
+      throwable.printStackTrace(new PrintWriter(sw));
+      this.trace = sw.toString();
+      return this;
+    }
+
+    public Builder data(final JsonValue context) {
```
mattdailis: 🔵 Curious about the choice to name the method `data` but the field `context` - any reason for them to be different?

pcrosemurgy: Oops, good catch! Just changed this to `data`.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ThreadedSchedulerAgent.java
pcrosemurgy: That's a good idea, just pushed a change to add a `.message()` here.

duranb (commented):
@pcrosemurgy This looks great! One thing that we forgot to realize is that the design has timestamps for the errors. Is that something that is easy to add to this?

Also, I haven't done any Java in yeeears so I might sound crazy, but does it make sense to make the `Failure` records implement some base generic `Failure` interface for when we eventually add more `Failure` types?

pcrosemurgy:
@duranb

>One thing that we forgot to realize is that the design has timestamps for the errors. Is that something that is easy to add to this?

Ahh okay! I'll add a timestamp field to the `simulation_dataset.reason` and `scheduling_request.failure_reason` columns.

>does it make sense to make the Failure records implement some base generic Failure interface for when we eventually add more Failure types?

So with this PR `SimulationFailure` and `ScheduleFailure` serve as this interface because every `writer.failWith()` call (anything that can write to the `reason` column) must build a `SimulationFailure`/`ScheduleFailure` (depending on which server we're talking about). :)

duranb:
@pcrosemurgy 
> Ahh okay! I'll add a timestamp field to the `simulation_dataset.reason` and `scheduling_request.failure_reason` columns.

Thank you!

> So with this PR `SimulationFailure` and `ScheduleFailure` serve as this interface because every `writer.failWith()` call (anything that can write to the `reason` column) must build a `SimulationFailure`/`ScheduleFailure` (depending on which server we're talking about). :)

Got it! Thanks for clearing that up for me!

pcrosemurgy:
Verification section has been updated to show this new `timestamp` as a field within the reason JSON

duranb (approved):
These changes are good with me 💯, but I'd like to see if @camargo has any thoughts.

PR #365, opened 2022-10-10
[FIX] Consistently serialize invalid arguments exception

pcrosemurgy:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

A while back (#258) `InvalidArgumentsException` exceptions were serialized when caught within `MerlinBindings`. There were a few validation endpoints that were not changed to use this serialization and instead turned the exception into a raw string. This PR correctly and consistently serializes all `InvalidArgumentException`s.

@camargo @duranb With these changes the Merlin endpoints `validateActivityArguments` and `validateModelArguments` will return an errors like so when the arguments are invalid (invalid in the sense that they are incorrectly-typed, not in the sense that they fail `@Validation`):
```
"errors": {
                "extraneousArguments": [],
                "missingArguments": [
                  "glutenFree"
                ],
                "unconstructableArguments": [
                  {
                    "failure": "Expected boolean, got StringValue[value=true]",
                    "name": "glutenFree"
                  }
                ]
              },
```
Previously this was reported as a JSON array:
```
"errors": [
    "some long java exception turned into a string"
]
```

## Verification

When validating `BakeBananaBread` with a `glutenFree` value of `"true"` (note, incorrectly typed as it should be a bool instead of a string) we now have a properly serialized `InvalidArgumentsException` instead of the exception turned into a raw string:
```json
"response": {
            "body": {
              "arguments": {
                "tbSugar": 34,
                "temperature": 99
              },
              "errors": {
                "extraneousArguments": [],
                "missingArguments": [
                  "glutenFree"
                ],
                "unconstructableArguments": [
                  {
                    "failure": "Expected boolean, got StringValue[value=true]",
                    "name": "glutenFree"
                  }
                ]
              },
              "success": false
            },
```

## Documentation
None.

## Future work
None.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
@@ -166,8 +166,7 @@ public final class MerlinBindings implements Plugin {
     } catch (final InvalidEntityException ex) {
       ctx.status(400).result(ResponseSerializers.serializeInvalidEntityException(ex).toString());
     } catch (final InvalidArgumentsException ex) {
-      ctx.status(400)
-          .result(ResponseSerializers.serializeFailures(List.of(ex.getMessage())).toString());
+      ctx.status(400).result(ResponseSerializers.serializeInvalidArgumentsException(ex).toString());
     } catch (final NoSuchPlanException ex) {
       ctx.status(404).result(ResponseSerializers.serializeNoSuchPlanException(ex).toString());
     } catch (final MissionModelService.NoSuchMissionModelException ex) {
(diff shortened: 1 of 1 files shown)


camargo:
This one will need a corresponding UI update before we merge since the UI assumes `errors` is an array. I think the only update will be here:
https://github.com/NASA-AMMOS/aerie-ui/blob/develop/src/components/activity/ActivityFormPanel.svelte#L268

pcrosemurgy:
This appears to be a broader issue that quite a few endpoints suffer from (for example, when serializing "no such mission model messages" without conforming to the expected GQL response type). [Make all Merlin Hasura endpoints conform to their expected GQL](https://jira.jpl.nasa.gov/browse/AERIE-2163) has been created to address this.

PR #366, opened 2022-10-10
[FEAT](AERIE-2146) Include EDSL API docs in docs

dyst5422:
Includes the autogenerated EDSL API documentation in our documentation site

* **Tickets addressed:** AERIE-2146
* **Review:** By commit  

## Description

Makes EDSL documentation part of the docsite generation and includes them in the docsite outputs. Includes some links to them in the existing docsite files.

Release procedure:

Ensure the current docs for EDSLs are in `docs/edsl-api-docs/<constraints|schedling>-edsl-api/develop` (can run `./gradlew publishDocs` to generate it there) and copy it to a versioned directory (ie `docs/edsl-api-docs/<constraints|schedling>-edsl-api/v0.13.2`). Note that this versioned directory MUST have the same name as the release tag for it to properly be linked in the site docs.

This new versioned directory needs to be committed to git so we maintain the older versions of edsl docs.

## Verification

Generated them locally and was able to link locally - will have to build + deploy to see if it works in CI

## Documentation

None

### DIFF
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -43,7 +43,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: "temurin"
-          java-version: "18"
+          java-version: "19"
       - name: Set up Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
       - name: Set up Gradle
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -6,6 +6,8 @@ on:
       - develop
     paths:
       - "docs/**"
+      - "merlin-server/constraints-dsl-compiler/src/**"
+      - "scheduler-server/scheduling-dsl-compiler/src/**"
   workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -35,10 +35,32 @@ jobs:
         run: make -C docs multiversion
       - name: Build redirects
         run: make -C docs redirects
+
+      # Generate EDSL documentation
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "18"
+      - name: Set up Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Set up Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Build Docs
+        run: ./gradlew publishDocs
+      - name: Copy Docs To Publish Dir
+        run: cp -r docs/_build/dirhtml publish-docs-tmp; cp -r docs/edsl-api-docs/. publish-docs-tmp
+      # Upload to GitHub Pages
       - name: Upload artifacts
         uses: actions/upload-pages-artifact@v1
         with:
-          path: "./docs/_build/dirhtml/"
+          path: "./publish-docs-tmp"
+      - name: Cleanup
+        run: rm -rf publish-docs-tmp
       - name: Deploy docs to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v1
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -8,6 +8,8 @@ on:
       - develop
     paths:
       - "docs/**"
+      - "merlin-server/constraints-dsl-compiler/src/**"
+      - "scheduler-server/scheduling-dsl-compiler/src/**"
 
 jobs:
   build:
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/setup-java@v3
         with:
           distribution: "temurin"
-          java-version: "18"
+          java-version: "19"
       - name: Set up Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
       - name: Set up Gradle
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -26,3 +26,18 @@ jobs:
         run: make -C docs setupenv
       - name: Build docs
         run: make -C docs test
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "18"
+      - name: Set up Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Set up Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Build EDSL Docs
+        run: ./gradlew publishDocs
--- a/docs/source/activity-plans/scheduling-guide.rst
+++ b/docs/source/activity-plans/scheduling-guide.rst
@@ -126,7 +126,7 @@ a goal is enabled or disabled.
 
 Scheduling DSL Documentation
 ============================
-Here you will find the full set of features in the scheduling DSL.
+Here you will find the full set of features in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__.
 
 
 .. warning::
--- a/docs/source/activity-plans/scheduling-guide.rst
+++ b/docs/source/activity-plans/scheduling-guide.rst
@@ -126,7 +126,7 @@ a goal is enabled or disabled.
 
 Scheduling DSL Documentation
 ============================
-Here you will find the full set of features in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__.
+The full set of features can be found in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__.
 
 
 .. warning::
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -7,6 +7,7 @@ from datetime import date
 from sphinx_scylladb_theme.utils import multiversion_regex_builder
 
 sys.path.insert(0, os.path.abspath(".."))
+sys.path.append(os.path.abspath("./_ext"))
 
 # -- Global variables
 
@@ -33,6 +34,7 @@ extensions = [
     "sphinx_scylladb_theme",
     "sphinx_multiversion",  # optional
     "myst_parser",  # for converting .md to .rst,
+    "rst_jinja"
 ]
 
 # The suffix(es) of source filenames.
--- a/merlin-server/build.gradle
+++ b/merlin-server/build.gradle
@@ -42,6 +42,21 @@ task generateDocumentation(type: NpmTask) {
   inputs.dir('src')
 }
 
+task cleanDocsFromDocsDir(type: Delete) {
+  delete rootDir.toPath().resolve('docs/edsl-api-docs/constraints-edsl-api/develop')
+}
+
+task copyDocsToDocsDirectory(type: Copy) {
+  from rootDir.toPath().resolve('merlin-server/constraints-dsl-compiler/build/docs')
+  into rootDir.toPath().resolve('docs/edsl-api-docs/constraints-edsl-api/develop')
+}
+
+task publishDocs {
+  dependsOn cleanDocsFromDocsDir
+  dependsOn generateDocumentation
+  dependsOn copyDocsToDocsDirectory
+}
+
 assemble {
   dependsOn assembleConstraintsDSLCompiler
   dependsOn generateASTJsonSchema
--- a/merlin-server/constraints-dsl-compiler/package.json
+++ b/merlin-server/constraints-dsl-compiler/package.json
@@ -4,11 +4,11 @@
   "description": "Typescript compiler for the constraints DSL",
   "type": "module",
   "scripts": {
-    "build": "tsc",
+    "build": "rm -rf build; tsc",
     "test": "npm run build && node build/main.js",
     "generate-ast-schema": "npx typescript-json-schema src/libs/constraints-ast.ts \"*\" --aliasRefs --out ./build/constraints.schema.json",
     "reformat": "npx prettier --config ./.prettierrc -w ./src",
-    "generate-doc": "npx typedoc src/libs/constraints-edsl-fluent-api.ts --excludePrivate --excludeInternal --out ./build/docs --name \"Constraints eDSL documentation\""
+    "generate-doc": "rm -rf build/docs; npx typedoc --plugin typedoc-plugin-extras --customTitle \"AERIE DOCUMENTATION\" --customTitleLink ../index.html src/libs/constraints-edsl-fluent-api.ts --excludePrivate --excludeInternal --out ./build/docs --name \"Constraints eDSL documentation\""
   },
   "author": "",
   "dependencies": {
@@ -17,7 +17,8 @@
     "source-map": "^0.7.3",
     "typescript": "^4.5.5",
     "typescript-json-schema": "0.53.0",
-    "typedoc": "^0.23.10"
+    "typedoc": "^0.23.10",
+    "typedoc-plugin-extras": "^2.3.0"
   },
   "devDependencies": {
     "@tsconfig/node16-strictest-esm": "^1.0.2",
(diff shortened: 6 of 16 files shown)


Mythicaeda (changes requested):
Major Issue: the EDSL `index.html` is currently replacing our site's main homepage, both when I build with `make preview` and `make multiversionpreview`. 

Less major: we ought to include a link to the EDSL documentation on our homepage, perhaps as another topic box, as well as in the navigation sidebar and/or at the bottom of the template our standard documentation pages follow. Similarly, we should have a way to get back to the main docs from the EDSL page.

Minor enough to not hold up the PR: is it possible to set the favicon/title of the pages to be in line with the other part of the docs?

merlin-server/constraints-dsl-compiler/package.json
```diff
@@ -8,7 +8,7 @@
     "test": "npm run build && node build/main.js",
     "generate-ast-schema": "npx typescript-json-schema src/libs/constraints-ast.ts \"*\" --aliasRefs --out ./build/constraints.schema.json",
     "reformat": "npx prettier --config ./.prettierrc -w ./src",
-    "generate-doc": "npx typedoc src/libs/constraints-edsl-fluent-api.ts --excludePrivate --excludeInternal --out ./build/docs --name \"Constraints eDSL documentation\""
+    "generate-doc": "npx typedoc src/libs/constraints-edsl-fluent-api.ts --excludePrivate --excludeInternal --out ./build/docs/constraints-dsl-compiler --name \"Constraints eDSL documentation\""
```
Mythicaeda: For both here and in `scheduler-server/scheduling-dsl-compiler/package.json`, what does this change do?

dyst5422: This puts it in a subdirectory - which we need so we don't have various files stomping on each other when we add them into the docsite with `html_extra_files`.

dyst5422:
> Major Issue: the EDSL `index.html` is currently replacing our site's main homepage, both when I build with `make preview` and `make multiversionpreview`.
> 
> Less major: we ought to include a link to the EDSL documentation on our homepage, perhaps as another topic box, as well as in the navigation sidebar and/or at the bottom of the template our standard documentation pages follow. Similarly, we should have a way to get back to the main docs from the EDSL page.
> 
> Minor enough to not hold up the PR: is it possible to set the favicon/title of the pages to be in line with the other part of the docs?

~Missed the multiversionpreview when adding the generation to the makefile.~ Correction, it already inherits that from muultiversion
I'll take a look at preview, but I specifically changed the outputs of our typedoc generation to fix that - you might have old docs generated under those right now that are being pulled in

Yeah, I wasn't sure where we wanted to link the docs, so I just put them in their corresponding EDSL general documentation pages for now.

Getting a link back out used to be possible, but I didn't immediately find it for the typedoc generation yesterday. I'll take a deeper look today. edit: Looks like that was moved to a plugin for typedoc, so I'll look at bringing that in and working on that change.

dyst5422:
Added the favicon we use for the rest of the docsite, ensured everything was building and being put in the right locations (definitely need to do a clean, and I added clearing out the build and docs to our scripts for the edsl compilers to ensure no stale outputs), and added the title link to go back to the primary docsite

dyst5422:
@Mythicaeda I think I'm inclined to have them not as their own sections on the main page with the topic boxes as I think its more of a sub piece of the respective systems. And I think its more of a detail of the DSL guides we have where we would want them going to those sections first before they look at the API documentation which is much more low-level. Thoughts?

Mythicaeda:
> @Mythicaeda I think I'm inclined to have them not as their own sections on the main page with the topic boxes as I think its more of a sub piece of the respective systems. And I think its more of a detail of the DSL guides we have where we would want them going to those sections first before they look at the API documentation which is much more low-level. Thoughts?

From my understanding, these are important docs that users are going to be referring to on a regular bases. Sure, for a first-time user we might want them to see our normal guidelines first, but if any other user wants to look at the EDSL documentation, they shouldn't have to navigate through four clicks from our home page to get there (assuming they have the click path memorized). Plus, I personally consider them to be their own topic, since they cover a different level of information, as you pointed out. Therefore, we should have a topic box called "EDSL API Documentation" (with the description highlighting that these are the in-depth docs for the EDSL) on the landing page, but as the last one, since we want users to look elsewhere first.

If you're worried about a first-time user ending up stuck in the low-level EDSL Docs and not being able to find the higher-level docs, then is it possible to include on the sidebar of the EDSL Docs a link specifically to the higher-level docs?

Mythicaeda:
> I'll take a look at preview, but I specifically changed the outputs of our typedoc generation to fix that - you might have old docs generated under those right now that are being pulled in

Yep, once I cleaned my gradle build files the problem was resolved.

dyst5422:
> > @Mythicaeda I think I'm inclined to have them not as their own sections on the main page with the topic boxes as I think its more of a sub piece of the respective systems. And I think its more of a detail of the DSL guides we have where we would want them going to those sections first before they look at the API documentation which is much more low-level. Thoughts?
> 
> From my understanding, these are important docs that users are going to be referring to on a regular bases. Sure, for a first-time user we might want them to see our normal guidelines first, but if any other user wants to look at the EDSL documentation, they shouldn't have to navigate through four clicks from our home page to get there (assuming they have the click path memorized). Plus, I personally consider them to be their own topic, since they cover a different level of information, as you pointed out. Therefore, we should have a topic box called "EDSL API Documentation" (with the description highlighting that these are the in-depth docs for the EDSL) on the landing page, but as the last one, since we want users to look elsewhere first.
> 
> If you're worried about a first-time user ending up stuck in the low-level EDSL Docs and not being able to find the higher-level docs, then is it possible to include on the sidebar of the EDSL Docs a link specifically to the higher-level docs?

Unfortunately, the doc generator is pretty limited in what it displays, but perhaps we can put a link at the top of the readme it brings in for the landing page on the typedocs with a message "hey if you found yourself here and think you're in over your head, you might want to go to these other docs first"

Mythicaeda:
> ensured everything was building and being put in the right locations (definitely need to do a clean, and I added clearing out the build and docs to our scripts for the edsl compilers to ensure no stale outputs)

So the files are being put in the right locations, but your link to the EDSL documentation is broken (as shown by the results of this `make linkcheck`:

`(activity-plans/scheduling-guide: line  129) *broken*   ../../scheduling-dsl-compiler/index.html`

If you insert a [`:doc:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-documents) or [`:ref:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations) (maybe [`:any:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-anything) if all else fails) reference in front of that link, it should work. You can verify if the link is working by running `make clean`, `make dochtml`, then `make linkcheck`, or manually by checking the link in _both_ `make preview` and `make multiversionpreview` mode (relative links end up directing to different locations between the modes, so something that appears to work in one version doesn't necessarily work in the other. Between the two, having the link work in `make multiversionpreview` is more important).

Mythicaeda:
> Unfortunately, the doc generator is pretty limited in what it displays, but perhaps we can put a link at the top of the readme it brings in for the landing page on the typedocs with a message "hey if you found yourself here and think you're in over your head, you might want to go to these other docs first"

That would work! My only concern is that it appears to be bringing in our GitHub README as the landing page, and we might not want our GitHub landing page speaking as if it's the root of the EDSL documentation.

docs/Makefile
```diff
@@ -88,11 +92,11 @@ multiversionpreview: multiversion
...
 .PHONY: test
-test: setup
+test: setup build-typedoc
 	$(SPHINXBUILD) -b dirhtml $(TESTSPHINXOPTS) $(BUILDDIR)/dirhtml
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/dirhtml."
 
 .PHONY: linkcheck
-linkcheck: setup
+linkcheck: setup build-typedoc
```
Mythicaeda: We shouldn't rebuild the typedoc docs when we run `make linkcheck`.

Mythicaeda: Actually, in general I'm wondering if we need to rebuild the typedoc docs whenever we build the rest of the site, or if we could instead include it as a separate `make` command that you only run when you're working specifically on those docs, or as a first step if the EDSL docs are not present in `aerie/schedule-server/scheduling-dsl-compiler/build/docs/scheduling-dsl-compiler` ([which appears to be possible to check with MAKEFILE syntax ](https://stackoverflow.com/questions/5553352/how-do-i-check-if-file-exists-in-makefile-so-i-can-delete-it)using the [`wildcard`](https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html) command).

dyst5422: I wanted them to build because building documentation typically isn't on the minds of the devs working on the EDSLs and I want to ensure we publish a up-to-date version.

Mythicaeda: For ensuring that we publish an up-to-date version, we can modify the GitHub workflow to always build the EDSL docs before running `make multiversion`. My concern here was on the increased build time for devs working on the docs, especially since the EDSL files are autogenerated. It's not PR-blocking, but the difference felt significant enough to bring up.

dyst5422: I think that's a fine solution.

dyst5422:
@Mythicaeda I think I have this in a place where we can merge to test the deploy GitHub action behavior.

.github/workflows/docs-pages.yaml
```diff
@@ -35,10 +37,32 @@ jobs:
...
+      - name: Set up Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: Set up Node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Build Docs
+        run: ./gradlew publishDocs
+      - name: Copy Docs To Publish Dir
+        run: cp -r docs/_build/dirhtml publish-docs-tmp; cp -r docs/edsl-api-docs/. publish-docs-tmp
```
Mythicaeda, line 58: Can you confirm what the contents of publish-docs-tmp look like after this step?

docs/source/activity-plans/scheduling-guide.rst
```diff
@@ -126,7 +126,7 @@ a goal is enabled or disabled.
 
 Scheduling DSL Documentation
 ============================
-Here you will find the full set of features in the scheduling DSL.
+Here you will find the full set of features in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__.
```
Mythicaeda: Minor, but could this sentence be changed to "The full set of features can be found in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__"?

docs/source/conf.py
```diff
@@ -33,6 +34,7 @@
     "sphinx_scylladb_theme",
     "sphinx_multiversion",  # optional
     "myst_parser",  # for converting .md to .rst,
+    "rst_jinja"
```
Mythicaeda, line 37: What does rst_jinja do? Is it how you use the template links?

.github/workflows/docs-pages.yaml
dyst5422: Screen Shot 2022-10-14 at 3 18 17 PM

scheduler-server/scheduling-dsl-compiler/package.json
```diff
@@ -1,13 +1,12 @@
 {
   "name": "scheduling-dsl-compiler",
-  "version": "1.0.0",
```
Mythicaeda, line 3: Why is `version` being removed?

merlin-server/build.gradle
```diff
@@ -42,6 +42,21 @@ task generateDocumentation(type: NpmTask) {
   inputs.dir('src')
 }
 
+task cleanDocsFromDocsDir(type: Delete) {
+  delete rootDir.toPath().resolve('docs/edsl-api-docs/constraints-edsl-api/develop')
```
Mythicaeda, line 46: Is this path the part that will need to be updated in order to support multiversioning on the EDSL docs?

Mythicaeda (commented):
I'm having difficulties getting the docs to build correctly. Could you please provide the steps?

docs/source/conf.py
dyst5422: It enables use to do templating in the rst files and specifically adds int the env var that we need from the multi version extension

scheduler-server/scheduling-dsl-compiler/package.json
dyst5422: We don't want version information held in the package files - they are misleading at best.

merlin-server/build.gradle
dyst5422: So we decided that as part of the release of a new version, we take the existing develop and move it to a versioned dir that gets committed. So in that way, we always have develop building, and the versioned ones we just handle manually for now. (typedoc doesn't have the cool git integration sphinx does)

dyst5422:
Build steps:

```
cd docs
make clean
make multiversion
cd ..
./gradlew publishDocs
rm -rf publish-docs-tmp
cp -r docs/_build/dirhtml publish-docs-tmp; cp -r docs/edsl-api-docs/. publish-docs-tmp
```

Mythicaeda:
Thanks for the steps! Unfortunately, they lead to a broken link if the branch of the documentation that you enter on is not `develop`. Can you clarify what `current version` is supposed to be set to?

dyst5422:
> Thanks for the steps! Unfortunately, they lead to a broken link if the branch of the documentation that you enter on is not `develop`. Can you clarify what `current version` is supposed to be set to?

`current_version` is set as whatever the current version of the sphinx documentation being built is. So it should build correctly for develop, replacing it with `develop`, and any of our tags should replace it with the tag name. In this way, when we store off versions of the EDSL docs, they simply need to match the naming scheme of our sphinx docs to be pulled in.

Do you want to walk through it briefly on a call so we can talk through it and enumerate any adjustments we want to make?

Mythicaeda:
> current_version is set as whatever the current version of the sphinx documentation being built is. So it should build correctly for develop, replacing it with develop, and any of our tags should replace it with the tag name. In this way, when we store off versions of the EDSL docs, they simply need to match the naming scheme of our sphinx docs to be pulled in

So just to be explicit, `current_version` will be `develop` while looking at the develop docs and `v0.13.1` while looking at the v0.13.1 docs and so on and so forth?

docs/source/activity-plans/scheduling-guide.rst
```diff
@@ -126,7 +126,7 @@ a goal is enabled or disabled.
 
 Scheduling DSL Documentation
 ============================
-Here you will find the full set of features in the `Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>`__.
+The full set of features can be found in the Scheduling EDSL API Documentation <../../scheduling-edsl-api/{{current_version}}/index.html>__.
```
Mythicaeda: You accidentally dropped the backticks around the link ;-)

dyst5422:
> > current_version is set as whatever the current version of the sphinx documentation being built is. So it should build correctly for develop, replacing it with develop, and any of our tags should replace it with the tag name. In this way, when we store off versions of the EDSL docs, they simply need to match the naming scheme of our sphinx docs to be pulled in
> 
> So just to be explicit, `current_version` will be `develop` while looking at the develop docs and `v0.13.1` while looking at the v0.13.1 docs and so on and so forth?

Yep!

docs/source/activity-plans/scheduling-guide.rst
dyst5422: Darn copy paste doing weird string sanitization

Mythicaeda (approved):
LGTM!

patkenneally (approved):

.github/workflows/docs-pages.yaml
```diff
@@ -35,10 +37,32 @@ jobs:
         run: make -C docs multiversion
       - name: Build redirects
         run: make -C docs redirects
+
+      # Generate EDSL documentation
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "18"
```
camargo: Can we update to 19?

.github/workflows/docs-pr.yml
```diff
@@ -26,3 +28,18 @@ jobs:
         run: make -C docs setupenv
       - name: Build docs
         run: make -C docs test
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "18"
```
camargo: Can we update to 19?

camargo (approved):
LGTM. A couple simple Java version updates and rebase with dev this should be good to go.

PR #367, opened 2022-10-10
[AERIE-1930] Include awaiting execution state activity tasks in results

patkenneally:
* **Tickets addressed:** AERIE-1930
* **Review:** By commit  

## Description
ComputeResults incorrectly assumes that tasks are either in the Terminated or InProgress state. There is a third state,  AwaitingChildren, which is like "Unfinished" in that it doesn't have an end time, but it's like "Finished" in that it has computed attributes. The check for Task state is extended to catch this third state so that all activities are included in simulation results.

## Verification
Added a test which simulates a small plan which has been constructed such that there will always be a parent activity awaiting a child. A new activity type, which decomposes using spawn, is added to Banananation for this purpose.

## Documentation
No changes or additions needed

## Future work
None

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -531,6 +531,18 @@ public final class SimulationEngine implements AutoCloseable {
             activityChildren.getOrDefault(activityId, Collections.emptyList()),
             (activityParents.containsKey(activityId)) ? Optional.empty() : Optional.of(activityId)
         ));
+      } else if (state instanceof ExecutionState.AwaitingChildren<?> e){
+        final var inputAttributes = taskInfo.input().get(task.id());
+        unfinishedActivities.put(activityId, new UnfinishedActivity(
+            inputAttributes.getTypeName(),
+            inputAttributes.getArguments(),
+            startTime.plus(e.startOffset().in(Duration.MICROSECONDS), ChronoUnit.MICROS),
+            activityParents.get(activityId),
+            activityChildren.getOrDefault(activityId, Collections.emptyList()),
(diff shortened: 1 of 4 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -531,6 +531,16 @@ public static SimulationResults computeResults(
...
+      } else if (state instanceof ExecutionState.AwaitingChildren<?> e){
+        final var inputAttributes = taskInfo.input().get(task.id());
+        unfinishedActivities.put(activityId, new UnfinishedActivity(
+            inputAttributes.getTypeName(),
+            inputAttributes.getArguments(),
+            startTime.plus(e.startOffset().in(Duration.MICROSECONDS), ChronoUnit.MICROS),
+            activityParents.get(activityId),
+            activityChildren.getOrDefault(activityId, Collections.emptyList()),
+            (activityParents.containsKey(activityId)) ? Optional.empty() : Optional.of(activityId)
+        ));
```
pcrosemurgy, line 543: Now that we're exhaustively checking for each of the possible `ExecutionState` types it may be good to add an `else` here to guarantee every subtype is being handled when testing:
```java
    } else {
      throw new Error("Unexpected subtype of %s: %s".formatted(ExecutionState.class, state.getClass()));
    }
```
When that JEP for pattern matching on sealed types is available this can be taken out.

pcrosemurgy (approved):
Thanks Pat, looks good!

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
patkenneally: Good point! I've added this.

mattdailis (approved):

PR #369, opened 2022-10-11
[HOTFIX] Canonicalize activity template arguments

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This was a regression reported by @camargo, where the Recurrence goal would fail to recognize existing activities in the plan and go ahead and place more activities.

The root cause was traced to `.equals` returning `false` when comparing `SerializedValue.IntValue(325)` with `SerializedValue.RealValue(325.0)`. This existing issue was exacerbated #301, where we now call `getEffectiveArguments` on every activity in the plan to fill in its full argument set.

This PR augments `makeActivityTemplate` to call `getEffectiveArguments`. Looking forward to Activity Expressions (which are allowed to have missing arguments), this implementation catches `IllegalArgumentException` and ignores missing arguments.

## Verification

I added a test to make sure that `325.0` results in a SerializedValue.RealValue rather than an IntValue.

Unfortunately, I wasn't able to directly write a test that reproduces the issue at the integration test level, since the effective arguments is called as part of the PlanService, which is mocked out. We can add an e2e test for this if we think it would be valuable.

## Documentation

This is a bugfix, shouldn't need any docs updates.

## Future work

I wonder if comparing the serialized values is the move long term, or if we should be comparing values at the types defined in the model.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
@@ -1,10 +1,12 @@
 package gov.nasa.jpl.aerie.scheduler.server.remotes.postgres;
 
-import gov.nasa.jpl.aerie.constraints.time.Windows;
 import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
 import gov.nasa.jpl.aerie.constraints.tree.WindowsWrapperExpression;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
+import gov.nasa.jpl.aerie.merlin.protocol.types.InvalidArgumentsException;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.Range;
 import gov.nasa.jpl.aerie.scheduler.constraints.TimeRangeExpression;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityCreationTemplate;
@@ -21,6 +23,9 @@ import gov.nasa.jpl.aerie.scheduler.model.PlanningHorizon;
 import gov.nasa.jpl.aerie.scheduler.server.models.SchedulingDSL;
 import gov.nasa.jpl.aerie.scheduler.server.models.Timestamp;
 import gov.nasa.jpl.aerie.scheduler.server.services.UnexpectedSubtypeError;
+
+import java.util.HashMap;
+import java.util.Map;
 import java.util.function.Function;
 public class GoalBuilder {
   private GoalBuilder() {}
@@ -28,7 +33,7 @@ public class GoalBuilder {
       final SchedulingDSL.GoalSpecifier goalSpecifier,
       final Timestamp horizonStartTimestamp,
       final Timestamp horizonEndTimestamp,
-      final Function<String, ActivityType> lookupActivityType) {
+      final Function<String, ActivityType> lookupActivityType) throws GoalBuilderFailure {
     final var hor = new PlanningHorizon(
         horizonStartTimestamp.toInstant(),
         horizonEndTimestamp.toInstant()).getHor();
@@ -127,7 +132,8 @@ public class GoalBuilder {
   }
   private static ActivityCreationTemplate makeActivityTemplate(
       final SchedulingDSL.ActivityTemplate activityTemplate,
-      final Function<String, ActivityType> lookupActivityType) {
+      final Function<String, ActivityType> lookupActivityType) throws GoalBuilderFailure
+  {
     var builder = new ActivityCreationTemplate.Builder();
     final var type = lookupActivityType.apply(activityTemplate.activityType());
     if(type.getDurationType() instanceof DurationType.Controllable durationType){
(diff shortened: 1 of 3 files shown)


scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/SynchronousSchedulerAgent.java
```diff
@@ -177,14 +179,26 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
...
+                  specification.horizonStartTimestamp(),
+                  specification.horizonEndTimestamp(),
+                  problem::getActivityType);
+        } catch (GoalBuilder.GoalBuilderFailure e) {
+          writer.failWith(b -> b
+              .type("SCHEDULING_GOAL_BUILDER_FAILED")
+              .message(e.getMessage())
+              .data(Json.createObjectBuilder()
+                        .add("goal_id", compiledGoal.getKey().id())
+                        .build()));
```
pcrosemurgy, line 199: You can add a `.trace(e)` at the tail-end of this builder chain to include the stack trace in the failure `reason`.

Not a big deal but eventually we can always revisit this builder pattern and maybe make things a bit more explicit to reduce the chance of an exception's trace being omitted. For example, we could separate this method into `failWithException()` and `failWith()` where `failWithException` must take an exception as an argument.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
```diff
@@ -138,9 +144,49 @@ private static ActivityCreationTemplate makeActivityTemplate(
...
-    for (final var argument : activityTemplate.arguments().entrySet()) {
+    for (final var argument : canonicalizeSerialization(type, activityTemplate.arguments()).entrySet()) {
       builder = builder.withArgument(argument.getKey(), argument.getValue());
     }
     return builder.build();
   }
+
+  /**
+   * Take a set of arguments and run it through getEffectiveArguments to make sure that the SerializedValues
+   * are the correct type. E.g. Int -> Double.
```
pcrosemurgy, line 155: > `E.g. Int -> Double`

This is an example of `SerializedValues` that are not correct, right?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
```diff
@@ -138,9 +144,49 @@ private static ActivityCreationTemplate makeActivityTemplate(
...
+      return res;
+    } catch (InvalidArgumentsException e) {
+      // If an exception is thrown, we still want to extract the valid arguments. Extraneous arguments are left as they were.
+      for (final var argument : e.validArguments) {
+        // This fills in the "correct" serialization of the argument. E.g. Int -> Double
+        res.put(argument.parameterName(), argument.serializedValue());
+      }
+      if (!e.extraneousArguments.isEmpty()) {
+        // If we hit this case, that means the typescript static checking has failed to catch extraneous arguments.
+        throw new GoalBuilderFailure("Activity Template had extraneous arguments: %s".formatted(e.extraneousArguments));
```
pcrosemurgy, line 180: I'd put this `if (...) throw` before the `for` loop above, that way you don't construct `res` and then just throw :)

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GoalBuilder.java
pcrosemurgy: Ohh it's an example of types that TS does not differentiate

dyst5422: There are _ways_ to make TS differentiate, but its definitely not a simple thing and can mess with type usage generally (see https://www.typescriptlang.org/play#example/nominal-typing)

dyst5422: We do it with our timestrings in the command expansion library here https://github.com/NASA-AMMOS/aerie/blob/develop/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts#L325-L326, but its an isolated case where we handle it all internal to our code

pcrosemurgy: Thanks!!

dyst5422: I _think_ we _could_ reliably do this with an int check to cast it to int internal to our type system, but again I'm not sure that's really a solution for this case as it doesn't get passed through to json

camargo:
Opened https://github.com/NASA-AMMOS/aerie-ui/pull/179 to reflect this change in the UI. The e2e tests in the UI should fail until this PR is merged. Once we merge this PR we can retest the UI PR against the develop aerie image and merge there if everything passes. I tested locally against this branch and the UI branch and everything works as expected.

pcrosemurgy (approved):
Assuming those 2 changes I mentioned get in this looks good to me :)

mattdailis:
@pcrosemurgy thanks for the review - after chatting with Adrien, we decided that he will open a PR with an alternate approach. We can weigh some pros/cons of the two approaches and close one of the PRs (likely this one).

Twisol:
I kind of like the canonicalization approach better. The input arguments are just input, a partial (generally) description of an activity instance that you can feed to the model to get a true instance. if you want to recognize a constructed activity, it's always going to be more reliable to either compare the instances themselves (which we don't require the model to support, currently) or to extract the canonical arguments for the instances and compare them.

If you just want to see if the same inputs have been used, you can compare inputs; but there's always going to be situations where different inputs (even wildly so) produce the same instance.

mattdailis:
Closing for now - I think we need to revisit the comparisons of serialized values more wholistically in the future.

PR #371, opened 2022-10-11
Rename `InvalidArgumentsException` -> `InstantiationException `

pcrosemurgy:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

**Note**: this PR's renaming changes the sig. of a mission model interface, therefore mission models must be updated.

This PR renames `InvalidArgumentsException` -> `InstantiationException` to minimize confusion between arguments being "invalid" (unconstructable, missing, or extraneous) and our "validation" endpoint responses indicating arguments that are "invalid" (failed an `@Validation` check).

## Verification
All tests passing.

## Documentation
Related docs updated.

## Future work
None.


Twisol:
I feel like "erroneous" is still roughly synonymous; I could still see confusion arising here.

How about `UnconstructibleDirectiveException`? That seems to describe the situation at hand more directly -- these arguments are not merely invalid, but they actually prevent the construction of the desired directive.

Twisol:
Or, since the exception occurs directly on the `instantiate` method, we could call it `InstantiationException`, which is nice and short.

pcrosemurgy:
`InstantiationException` is cool with me, thanks for the input :)
Just pushed this change.

mattdailis (approved):

goetzrrGit (approved):
Looks good to me.

PR #372, opened 2022-10-12
[HOTFIX] Fix json<->SerializedValue int/double ambiguity when parsing

adrienmaillard:
* **Review:** By commit  

## Description

Problem: when parsing the arguments of a `ActivityTemplate` to transform them into SerializedValue, a json number will be parsed in the following block (from `SerializedValueJsonParser`): 
```
case NUMBER -> {
        final var num = (JsonNumber) value;
        yield (num.isIntegral())
            ? SerializedValue.of(num.longValue())
            : SerializedValue.of(num.doubleValue());
      }
```
If `300.0` is parsed, it will result into a `IntValue` even though the mission model specifies that this argument should be a `RealValue`. And then collide with effective arguments (that are correctly typed) later. 

Approach: create a "strict" `SerializedValueJsonParser` taking a `ValueSchema` as argument to treat this real/int ambiguity. To pass the correct `ValueSchema` to this strict parser when deserializing, I also created a specific parser for `ActivityTemplate` that goes and picks the right `ValueSchema` for each of the argument. 

Then I annotated the safe and unsafe used of `SerializedValueJsonParser` in the scheduler. Safe is when deserializing, unsafe is when serializing. 

Finally, I have updated the tests. Even though it was transparent because of the use of `SerializedValue.of()`, the bug was actually present there also as the argument with the name `subsubfield1` has type `Real` but was always returned as an `Integer`. 

https://github.com/NASA-AMMOS/aerie/pull/369 by @mattdailis is trying to fix the same bug. 

## Verification

Tests are passing.

## Documentation

None, this is a bug. 

## Future work

I believe this bug exists everywhere arguments are parsed. It may not be visible if the arguments are then going through effective arguments as the effective arguments will "come back" with the correct type.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/ActivityTemplateJsonParser.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/ActivityTemplateJsonParser.java
@@ -1,0 +1,71 @@
+package gov.nasa.jpl.aerie.scheduler.server.http;
+
+import gov.nasa.jpl.aerie.json.Breadcrumb;
+import gov.nasa.jpl.aerie.json.JsonParseResult;
+import gov.nasa.jpl.aerie.json.JsonParser;
+import gov.nasa.jpl.aerie.json.SchemaCache;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.scheduler.server.models.SchedulingDSL;
+import gov.nasa.jpl.aerie.scheduler.server.services.MissionModelService;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.json.JsonValue;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ActivityTemplateJsonParser implements JsonParser<SchedulingDSL.ActivityTemplate> {
+
+  private final Map<String, MissionModelService.ActivityType> activityTypesByName = new HashMap<>();
+
+  public ActivityTemplateJsonParser(MissionModelService.MissionModelTypes activityTypes){
+    activityTypes.activityTypes().forEach((actType)-> activityTypesByName.put(actType.name(), actType));
+  }
+
+  @Override
+  public JsonObject getSchema(final SchemaCache anchors) {
+    return Json
+        .createObjectBuilder()
+        .add("type", "object")
+        .build();
+  }
+
+  @Override
+  public JsonParseResult<SchedulingDSL.ActivityTemplate> parse(final JsonValue json) {
+    if (!(json instanceof JsonObject)) return JsonParseResult.failure("expected object");
+    final var asObject = json.asJsonObject();
+    if(!asObject.containsKey("activityType") || !asObject.containsKey("args")) return JsonParseResult.failure("expected fields activityType and args");
+    final var activityType = asObject.getString("activityType");
+    final var args = asObject.getJsonObject("args");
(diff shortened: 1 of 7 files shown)


camargo:
Opened https://github.com/NASA-AMMOS/aerie-ui/pull/179 to reflect this change in the UI. The e2e tests in the UI should fail until this PR is merged. Once we merge this PR we can retest the UI PR against the develop aerie image and merge there if everything passes.

mattdailis (approved):

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/ActivityTemplateJsonParser.java
```diff
@@ -0,0 +1,61 @@
...
+    final var asObject = json.asJsonObject();
+    if(!asObject.containsKey("activityType") || !asObject.containsKey("args")) return JsonParseResult.failure("expected fields activityType and args");
+    final var activityType = asObject.getString("activityType");
+    final var args = asObject.getJsonObject("args");
+    final var map = new HashMap<String, SerializedValue>(json.asJsonObject().size());
+    for (final var field : args.entrySet()) {
+      final var parsedValue = new StrictSerializedValueJsonParser(this.activityTypesByName.get(activityType).parameters().get(field.getKey())).parse(field.getValue());
+      if(parsedValue.isFailure()){
+        return JsonParseResult.failure("error while parsing argument with key "+ field.getKey());
+      }
```
mattdailis: Let's use `prependBreadcrumb` here to avoid losing the information from the `parsedValue` failure message.

```suggestion
      if (parsedValue instanceof JsonParseResult.Failure<SerializedValue> failure){
        failure.prependBreadcrumb(Breadcrumb.ofString(field.getKey()));
        return failure.cast();
      }
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/ActivityTemplateJsonParser.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+
+  @Override
+  public JsonValue unparse(final SchedulingDSL.ActivityTemplate activityTemplate) {
+    final var builder = Json.createObjectBuilder();
+    builder.add("activityType", activityTemplate.activityType());
+    final var argumentsBuilder = Json.createObjectBuilder();
+    for (final var entry : activityTemplate.arguments().entrySet())
+      argumentsBuilder.add(entry.getKey(), new StrictSerializedValueJsonParser(this.activityTypesByName.get(activityTemplate.activityType()).parameters().get(entry.getKey())).unparse(entry.getValue()));
```
mattdailis: I think it may help readability to 1) add braces around the for loop and 2) add some newlines:
```suggestion
    for (final var entry : activityTemplate.arguments().entrySet()) {
      argumentsBuilder.add(
          entry.getKey(),
          new StrictSerializedValueJsonParser(
              this.activityTypesByName
                  .get(activityTemplate.activityType())
                  .parameters()
                  .get(entry.getKey()))
              .unparse(entry.getValue()));
    }
```

adrienmaillard: Done

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/ActivityTemplateJsonParser.java
adrienmaillard: Done

PR #376, opened 2022-10-14
[AERIE-2165]  fix command typelib generation

dyst5422:
* **Tickets addressed:** AERIE-2165
* **Review:** By commit  

## Description

The command generation was all kinds of fouled up.

Specific changes:
- added FixedString type alias to preface
- wrapped all argument names in quotes so that ts-invalid characters don't screw it up
- Fixed the type annotation outputs that were invalid

Regarding the invalid type annotations, the array args version of declarations were incorrectly being written out as
```ts
function SomeCommand(args: [
  number: F<64>,
  string: FixedString,
]): SomeCommand;
```

where now it will write out correctly as

```ts
function SomeCommand(args: [
  F<64>,
  FixedString,
]): SomeCommand;
```

## Verification

Added a snapshot test for command generation and ran all tests
Also ran on Psyche dictionary locally where we found this issue initially

## Documentation

None should be necessary

### DIFF
--- a/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/command-expansion-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -92,6 +92,7 @@ declare global {
   type I32 = I<32>;
   type I64 = I<64>;
   type VarString<PrefixBitLength extends number, MaxBitLength extends number> = string;
+  type FixedString = string;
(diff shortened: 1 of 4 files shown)


mattdailis (approved):
This seems like a reasonable change to me

camargo (approved):

PR #378, opened 2022-10-16
Javadocs for merlin.protocol

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR contains a bunch of actual Javadocs for the `merlin.protocol` package. It's still in progress, but this is the kind of thing that can be done incrementally!

Also, a fair amount of "documentation" came in #318, which reorganized and renamed many of the entities in `merlin.protocol` to give a better baseline of self-documenting code. That isn't to say that prose documentation isn't needed (heavens, no!) -- only that readers of the prose documentation herein will greatly benefit from the much tighter naming and organization prepared in #318.

As usual, this PR also comes with a (small) prefix of refactors that came up during the main work. (Really, #318 was the prefix for this work, but it was separated out because of how much there was to it -- and the bug it ultimately fixed.) Here, I've cleaned up some Gradle dependencies in `merlin-sdk` and removed a type parameter that no longer matters.

## Verification
It all compiles and runs; the changes made in this PR are either documentation-only or do not impinge on business logic. (For instance, the type parameter removal is the kind of thing where, if it's wrong, the system should simply not compile. Type checking is a form of testing!)

## Documentation
Lots of javadocs were added, but nothing existing needs to change.

## Future work
Document all the things!

image

### DIFF
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Initializer.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Initializer.java
@@ -4,15 +4,75 @@ import gov.nasa.jpl.aerie.merlin.protocol.model.CellType;
 import gov.nasa.jpl.aerie.merlin.protocol.model.OutputType;
 import gov.nasa.jpl.aerie.merlin.protocol.model.Resource;
 import gov.nasa.jpl.aerie.merlin.protocol.model.Task;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
-import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 
 import java.util.function.Function;
 
+/**
+ * An interface to the driver during model instantiation.
+ *
+ * <p> An {@code Initializer} allows models to {@linkplain #allocate(Object, CellType, Function, Topic) allocate}
+ * storage for simulation-aware state, which is the only mutable state that can safely be shared among
+ * stateful cells that can be read and updated from {@link Task}s. The model must not interact with an {@code Initializer}
+ * beyond the scope in which it was received. </p>
+ *
+ * <p> Models can also export resources (time-varying values) and topics (streams of discrete events) via this interface.
+ * These observable outputs of the model provide feedback to users, displaying the effect on the model of the input
+ * directives and configuration. Observing the behavior of resources and topics is very nearly the whole point of
+ * simulating a model, so it is essential to register these outputs for clients to make use of. </p>
+ *
+ * <p> In a future revision of this API, exportation of resources and topics may be moved to the {@link gov.nasa.jpl.aerie.merlin.protocol.model.ModelType}
+ * interface, so that the existence and shape of these outputs cannot vary with model configuration. </p>
+ */
 public interface Initializer {
+  /**
+   * Gets the current state of the cell with the given ID.
+   *
+   * <p> Since cells cannot be mutated during initialization, this method always returns the initial state provided to
+   * {@link #allocate(Object, CellType, Function, Topic)} for the given cell ID. </p>
+   *
+   * @param <State>
+   *   The type of state held by the cell.
+   * @param cellId
+   *   The ID of the cell to query.
+   * @return
+   *   The current state of the cell.
+   */
   <State>
   State getInitialState(CellId<State> cellId);
 
+  /**
+   * Allocates a stateful cell with the given initial state, which may evolve over time and under effects. The cell is
+   * automatically subscribed to the provided topic, whose events are interpreted as effects by the given function.
+   *
+   * <p> All mutable state referenced by a model must be allocated via this method. In turn, this method returns
+   * {@link gov.nasa.jpl.aerie.merlin.protocol.driver.CellId}s, which identify the cell to the driver in future
+   * read/write interactions. This makes it possible to give each concurrent task a transactional view of the model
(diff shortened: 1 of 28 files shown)


mattdailis:
I was going to put this comment on #318 , but it may be more appropriate here since you're in the process of explaining all the things 😉 (Let me know if you think otherwise and I can re-post on the other PR)

My first question was regarding Task vs TaskFactory - previously, ActivityActions would call `spawn(DirectiveId, Task)`, and now they call `spawn(TaskFactory)`. I have no concerns about this, but I'm not sure I appreciate the nuance - could you help me understand the reasoning here?

Just thinking out loud, the one observable difference I can think of is that previously calling `spawn(new UserDefinedActivity())` during Mission model initialization would result in an `IllegalStateException`, whereas now it will spawn that activity's effect model as a daemon task...? 🤔 

Also - I remember we went back and forth on how to consolidate the two `Scheduler::spawn` methods (one of which was for directives, one of which was for anonymous tasks). Could you comment on how you were able to consolidate them both into `spawn(TaskFactory)`?

Twisol:
> I was going to put this comment on #318 , but it may be more appropriate here since you're in the process of explaining all the things 😉 (Let me know if you think otherwise and I can re-post on the other PR)

Here is fine!

> My first question was regarding Task vs TaskFactory - previously, ActivityActions would call `spawn(DirectiveId, Task)`, and now they call `spawn(TaskFactory)`. I have no concerns about this, but I'm not sure I appreciate the nuance - could you help me understand the reasoning here?
> 
> Just thinking out loud, the one observable difference I can think of is that previously calling `spawn(new UserDefinedActivity())` during Mission model initialization would result in an `IllegalStateException`, whereas now it will spawn that activity's effect model as a daemon task...? 🤔

Well, the difference between `Task` and `TaskFactory` is not so great; if you have a `TaskFactory`, you can always get a `Task`. The big change was removing `DirectiveTypeId` (or whatever it was called), which was how we *previously* associated a directive type to its input and output serialization routines for sim result extraction. The simulation engine doesn't use the `DirectiveTypeId` anymore, due to an even earlier PR that identifies directive start/stop events by their registered topic name (which is something a downstream client in principle ought to be doing instead, but I digress).

The difference between a daemon task and regular task delegation is simply *when* `spawn` is called, not *how* it is called. When you spawn a task during initialization, there is no true parent task. Put differently, model construction must complete before simulation can proceed, so you can't do some things tasks normally can (and you can do some things that tasks normally cannot, like cell allocation).

Also, using `TaskFactory` everywhere instead of `Task` made it easier to consolidate task creation on the engine side, where the root model is always easily available. Using the `Task` version of `spawn` in `ActivityActions#spawn` (and friends) required us to also fetch the root model (so we could use its stored executor for any spawned `ThreadedTask`s), which simply isn't necessary anymore. This is what let me get rid of the second `Scoped` variable entirely in #318.

> 
> Also - I remember we went back and forth on how to consolidate the two `Scheduler::spawn` methods (one of which was for directives, one of which was for anonymous tasks). Could you comment on how you were able to consolidate them both into `spawn(TaskFactory)`?

I think we had debated adding a special `DirectiveTypeId` for anonymous tasks, so that we could unify everything into the two-argument form. Since we have no need of `DirectiveTypeId` at all anymore, everthing naturally unified into the single-argument form.


Does that answer your questions? 🤔

Twisol:
Also, one of the original motivations for `TaskFactory` was that model construction might be expensive; `TaskFactory` lets us reuse the same constructed model across multiple simulations, so long as they're configured the same, since we can create fresh daemon tasks for each simulation.

It turns out that `TaskFactory` is simply more convenient than `Task` in other cases, too, so it's started to take over.

I'm really considering making the `CallingTask` variant of `TaskStatus` take a `TaskFactory` too, for the same reason. (Also it makes it less easy to confuse the child for the parent's continuation.)

Twisol:
Huh! Looks like I never made `Scheduler#spawn` take a `TaskFactory`. We're currently threading the executor through `ReplayingTask` even though that class doesn't use it, just so that we can propagate it to spawned tasks -- I think we can remove that flow entirely by switching to `TaskFactory` in `Scheduler#spawn` and `TaskStatus#calling`. Let's find out!

Twisol:
Well, that escalated quickly. #382 is the culmination of moving more things to using `TaskFactory`. It is very nice.

Twisol:
Could we get one more approval on these Javadocs? 👀

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Initializer.java
```diff
@@ -4,35 +4,134 @@
...
 
+/**
+ * An interface to the driver during model instantiation.
+ *
+ * <p> An {@code Initializer} allows models to {@linkplain #allocate(Object, CellType, Function, Topic) allocate}
+ * storage for simulation-aware state, which is the only mutable state that can safely be shared among
+ * stateful cells that can be read and updated from {@link Task}s. The model must not interact with an {@code Initializer}
+ * beyond the scope in which it was received. </p>
+ *
+ * <p> Models can also export resources (time-varying values) and topics (streams of discrete events via this interface.
```
mattdailis: Missing closing paren

mattdailis (approved):

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/driver/Initializer.java
Twisol: Ah, good find! I'll get this rebased and fix this paren later today.

PR #379, opened 2022-10-17
Update to Java 19

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR updates our Gradle scripts and and Dockerfiles to use Java 19, which was released on September 20th, 2022. At this point, [Eclipse's Adoptium](https://adoptium.net/temurin/releases?version=19) has released its Temurin distribution of OpenJDK 19, including Docker base images.

The [features in Java 19](https://openjdk.org/projects/jdk/19/) are mostly at the preview and incubator stage, but this will allow us and our users to begin experimenting early with Loom's virtual threads, which we have been eagerly anticipating for possible performance gains and architectural simplifications (re: threaded tasks in simulation).

(For those for whom "what about LTS" comes to mind, please see Ron Pressler's thoughts on this: [[1]](https://news.ycombinator.com/item?id=33028988), [[2]](https://news.ycombinator.com/item?id=33046148). Ron is the technical lead on OpenJDK's Project Loom, and has been beating this particular drum for *years*.)

## Verification
Build, test, and assembly all complete successfully. We'll see if the GitHub Actions deployment behaves as well.

## Documentation
- [x] Mission modelers will need to install a Java 19 JDK in order to use builds of the Merlin libraries after this change, but existing built mission models should continue to work. Customer deployments which take our Docker images should continue to do so; no additional effort should be required.

## Future work
* Experiment with Loom!

### DIFF
--- a/constraints/build.gradle
+++ b/constraints/build.gradle
@@ -4,7 +4,7 @@ plugins {
 
 java {
   toolchain {
-    languageVersion = JavaLanguageVersion.of(18)
+    languageVersion = JavaLanguageVersion.of(19)
   }
 }
 
--- a/contrib/build.gradle
+++ b/contrib/build.gradle
@@ -5,7 +5,7 @@ plugins {
 
 java {
   toolchain {
-    languageVersion = JavaLanguageVersion.of(18)
+    languageVersion = JavaLanguageVersion.of(19)
   }
 
   withJavadocJar()
--- a/db-tests/build.gradle
+++ b/db-tests/build.gradle
@@ -8,6 +8,10 @@ java {
   }
 }
 
+test {
+  useJUnitPlatform()
+}
+
 task e2eTest(type: Test) {
   useJUnitPlatform()
 }
--- a/db-tests/build.gradle
+++ b/db-tests/build.gradle
@@ -4,7 +4,7 @@ plugins {
 
 java {
   toolchain {
-    languageVersion = JavaLanguageVersion.of(18)
+    languageVersion = JavaLanguageVersion.of(19)
   }
 }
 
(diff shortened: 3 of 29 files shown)


Twisol:
I somewhat expected this, but my attempt to put the Dockerfiles back on `alpine` failed because they use `apt` -- hence the above deployment failures. I've pinned them to an (ideally, reproducible) Ubuntu base version instead.

It's really nice that our CI system catches stuff like this, by the way :)

skovati (approved):
Loom 👀 👀. Good call upgrading Javalin too, lets us get that virtual thread goodness right away. Should we make the transition to alpine based container images a separate chunk of work?

camargo (approved):
LGTM. Thank you!

Twisol:
> Loom 👀 👀. Good call upgrading Javalin too, lets us get that virtual thread goodness right away.

Well, we won't necessarily get virtual threads in Javalin unless we run the server with `--enable-preview`. Which we could, I guess! But the real reason I upgraded Javalin is because previous versions would eagerly try to use Loom if it was present, even if doing so would cause an exception due to the absence of said flag. So upgrading just keeps the server from dying on startup. (I have words on this in the commit message, too.)

> Should we make the transition to alpine based container images a separate chunk of work?

Yeah, I was just trying to do so opportunistically while I was in the area. Since it would require changing a bit more than I'm comfortable doing in a Java version bump, I settled for an Ubuntu base like before.

Twisol:
I saw in another PR that a release is planned very soon. Should I hold off on merging this until that happens?

camargo:
My vote is we get it into the point release tomorrow so we can get it into users hands sooner. @patkenneally @mattdailis what are your thoughts?

Twisol:
I'll let one of you all hit the button depending on what your decision is. If you end up needing to rebase, this branch shouldn't be very difficult to do so with.

PR #380, opened 2022-10-17
[AERIE 2138]  Plan Duplicate and Snapshots

Mythicaeda:
* **Tickets addressed:** AERIE-2138
* **Review:** By commit  

## Description

Implements three features necessary for Plan Merge: 

- Duplicating plans to create child plans and 
- Taking snapshots of plans
- Finding the history of plans and snapshots

This PR also modifies the database as follows:

- updates the primary key of the `activity_directive` and `activity_directive_validations` tables from `id` to `(id, plan_id)`
- adds a `parent_id` column to `plan`
- adds the tables `plan_snapshot`, `plan_snapshot_parent`, `plan_latest_snapshot`, and `plan_snapshot_activities`

## Verification

- Created a new test file, `PlanCollaborationTests.java`, to unit test the new features. 
- Modified test utility file `ActivityDirective.ts` to use the new primary key for `activity_directive`.
- Modified `simulatedActivityInstanceBatchLoader.spec.ts` and `sequence-generation.spec.ts` to reflect updates to `ActivityDirective.ts`

## Documentation

To be completed with #399.

## Future work

Merge Requests and Documentation


merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
+-- TODO list:
+--   - plan snapshot table (done)
+--   - duplicate function (done)
+--        - duplicate temporal subset of plan
+--   - merge request table
+--   - diff function
+--   - history tracking (tables and function) (done)
+--   - merge base function
```
JoelCourtney: please remove done things from the list

merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
...
+
+create table plan_snapshot_parent(
+  snapshot_id integer,
+  parent_snapshot_id integer
+);
+
+create table plan_latest_snapshot(
+  plan_id integer,
+  snapshot_id integer
+);
```
JoelCourtney: please add foreign and primary key constraints

merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
...
+
+-- Snapshot is a collection of the state of all the activities as they were at the time of the snapshot
+-- as well as any other properties of the plan that can change
+create table plan_snapshot(
+  snapshot_id integer generated always as identity,
+
+  plan_id integer not null,
```
JoelCourtney: please add key constraints

merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
...
+  revision integer not null,
+  name text not null,
+  duration interval not null,
+  start_time timestamptz not null
+);
+
+create table plan_snapshot_activities(
+  snapshot_id integer,
```
JoelCourtney, line 57: please add key constraints

merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
...
+  returns integer -- snapshot id inserted into the table
+  language plpgsql as $$
+  declare
+    validate_planid integer;
+    inserted_snapshot_id integer;
+begin
+  select id from plan where plan.id = plan_id into validate_planid;
+  if validate_planid is null then
+    raise exception 'Plan % does not exist.', plan_id;
+  end if;
```
JoelCourtney, line 81: this should be taken care of by foreign key constraints

merlin-server/sql/merlin/plan_collaboration.sql
```diff
@@ -0,0 +1,188 @@
...
+  language plpgsql as $$
+  declare
+    validate_plan_id integer;
+    new_plan_id integer;
+    created_snapshot_id integer;
+begin
+  select id from plan where plan.id = duplicate_plan.plan_id into validate_plan_id;
+  if(validate_plan_id is null) then
+    raise exception 'Plan % does not exist.', plan_id;
+  end if;
```
JoelCourtney, line 124: this should be covered by key constraints

merlin-server/sql/merlin/tables/plan.sql
```diff
@@ -8,6 +8,7 @@ create table plan (
 
   -- TODO: Remove 'start_time'; its purpose will be served by a model-specific entry in 'simulation.arguments'.
   start_time timestamptz not null,
+  parent_id integer,
```
JoelCourtney: key constraint pls

JoelCourtney (changes requested):
Seems pretty slick except for the key constraints we talked about. I put comments on all of them just to be thorough. Other than that I like it!

merlin-server/sql/merlin/plan_collaboration.sql
JoelCourtney: also looks like it could be `not null`.

merlin-server/sql/merlin/plan_collaboration.sql
JoelCourtney: also add `not null` when possible

merlin-server/sql/merlin/tables/plan.sql
JoelCourtney: In this case with a key constraint it would be nice to explicitly say `parent_id integer null` just to be clear.

Mythicaeda: I don't think this can be a foreign key column because `parent_id` can be null. It can also be the id of a plan that no longer exists.

merlin-server/sql/merlin/plan_collaboration.sql
Mythicaeda: Will add a primary key notice to snapshot_id, but because snapshots persist after plans are deleted, I believe that excludes plan_id from being a foreign key on plan.

JoelCourtney: Why do snapshots persist?

merlin-server/sql/merlin/tables/plan.sql
JoelCourtney: I did a little research before, and I thought it could be. It only enforces the constraint if it is not null. I haven't tried it myself though.

JoelCourtney:
```sql
CREATE TABLE some_child (
    parent_id integer references parent(id) on delete set null
);
```

merlin-server/sql/merlin/plan_collaboration.sql
JoelCourtney: regardless, you can still use `on delete set null` and allow `plan_id` to be null. It wouldn't make sense to reference a deleted plan anyway, right?

Mythicaeda: TL;DR: To not break merges. If I'm currently merging your plan into mine and you delete your plan, I should still be able to finish my merge. Also, the same snapshot can be in the history of several plans (they have a many to many relationship), so it doesn't make sense to delete on plan delete.

Also:
In general, we decided that once a merge request has been made, no changes made to the plan supplying changes will impact what the plan getting merged into sees when they go to merge. Not having snapshots would violate that decision.

Similarly, snapshot history and finding the merge base can break if we have snapshots deleted with their plan. For a simple example, if we:

1. create a plan
2. create a child
3. create a grandchild from the child
4. create a merge request to merge the grandchild into the grandparent plan
5. delete the parent plan
6. attempt to begin the merge between the grandchild and grandparent

Then our merge base, which was the creation snapshot of the child plan, has been unexpectedly deleted. This means that deleting one plan has affected a merge request involving two different plans. Similarly, if we had actually begun the merge between the grandchild and grandparent and the child plan was deleted at some point between when the grandparent plan is locked and `begin_merge`/the UI finish wanting to get data from the merge base snapshot, then we're risking an erroneous state.

merlin-server/sql/merlin/tables/plan.sql
Mythicaeda: I think we ought to decide as a team if we want to orphan the children of deleted plans. The other two options are 

1. keep the parent_id and preserve the relationship between plans
2. have the grandparent plan 'adopt' the immediate children of deleted plans, which preserves the list of which plans can merge into each other, if not the relationship between them

merlin-server/sql/merlin/plan_collaboration.sql
Mythicaeda: Assuming that the key constraint you're referring to is the one you proposed on the plan table, since we work with another table before attempting the insert, I disagree with removing this check.

merlin-server/sql/merlin/plan_collaboration.sql
Mythicaeda: The check here and in `duplicate_plan` verify that the value of `plan_id` is not null, which is something that is not enforced by foreign key constraints. While I agree that if we add FKCs, this check can be simpler, we will still need to catch the null case here.

merlin-server/sql/merlin/plan_collaboration.sql
JoelCourtney: Ok, that all makes sense. You can still add a foreign key constraint though, because it doesn't make sense to reference a plan that doesn't exist.

merlin-server/sql/merlin/tables/plan.sql
JoelCourtney: 1. You can't preserve the relationship between plans, because after the parent is deleted, `parent_id` is a less-than-meaningless number. It implies that the parent exists, and the `parent_id` is how you can find it. It even has the potential to end up referencing the wrong parent, because the fact that `generated always` produces sequential numbers is just an implementation detail and not something we should rely on.
2. I'm down with the adopting approach, because it permits a foreign key constraint.

I'm pushing for these things because relying on ourselves to keep the database in a valid state is the fastest path to an invalid state. "Make invalid states unrepresentable."

Mythicaeda: The adoption strategy has been implemented.

JoelCourtney (approved):
<pre>
    __4___
 _  \ \ \ \
<'\ /_/_/_/
 ((____!___/)
  \0\0\0\0\/
 ~~~~~~~~~~~
</pre>

camargo (approved):

PR #382, opened 2022-10-17
Move the simulation thread pool from the model to the driver

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR moves the `ExecutionService` for thread pooling from the model into the driver -- specifically, the `SimulationEngine`. Anything in the model that spawns a task now passes a `TaskFactory` back to the driver, rather than constructing a `Task` locally. This avoids a great deal of the work previously necessary to make the `ExecutionService` available wherever it was needed in the model, and generally streamlines interactions between the two sides.

As a result of these changes:
* `RootModel` is no longer necessary (as it simply bundle the `ExecutionService` alongside the mission's own model).
* The `MerlinExtension` no longer needs a model provided to it in test cases, so `MerlinTextContext` is no longer necessary. Also, the type parameters on `MerlinExtension` have gone away, meaning we can always use `@ExtendWith(MerlinExtension.class)` rather than the awkward field-based `@RegisterExtension`. Hurray!
* We can **automatically use Loom's virtual threads** if we detect that we're running in a JVM with them enabled. (We don't currently deploy images with `--enable-preview` present, but now it should be easier than ever to trial simulation on Loom.)

## Verification
Build succeeds. I haven't actually tried enabling Loom on this PR because it's kind of a pain to configure Gradle, but if Loom isn't available it will fall back to the existing thread pool anyway. (I did actually try Loom just before JDK 19 was released, and it worked without hiccups. Anyway, if you run with `--enable-preview`, you're explicitly opting in to potential instability, so...)

## Documentation
Probably some of the testing documentation needs to change. I also updated a code snippet for `ActivityMapper` that referenced `RootModel` when I was removing it.

## Future work
* Move registration of resources and topics out of the model constructor. This will, as a side-effect, streamline tests *even more*, because we won't need to pass in a `Registrar`.
* Allow models to `allocate` cells during simulation instead of only at construction time. This would avoid the need for a test constructor entirely. Entirely! (But also, it opens up some interesting patterns for shared local simulation state between tasks that there isn't much reason to forbid.)


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/TaskFactory.java
```diff
@@ -0,0 +1,7 @@
+package gov.nasa.jpl.aerie.merlin.protocol.model;
+
+import java.util.concurrent.ExecutorService;
+
+public interface TaskFactory<Return> {
+  Task<Return> create(ExecutorService executor);
```
mattdailis: For my understanding - why is `ExecutorService` the right interface to use here? Would it make any sense to pare down the methods we expose here, or is giving the task full control of the executor important?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -321,6 +350,8 @@ public void close() {
         r.state.release();
       }
     }
+
+    this.executor.shutdownNow();
```
mattdailis, line 354: 👌 This method alone justifies this change to me - coupling the cleanup of the executor with the cleanup of the simulation engine makes a lot of sense.

examples/foo-missionmodel/src/test/java/gov/nasa/jpl/aerie/foomissionmodel/SimulateMapSchedule.java
```diff
@@ -42,25 +41,21 @@ void simulateWithMapSchedule() {
...
 
-      simulationResults.simulatedActivities.forEach((name, activity) -> {
-        System.out.println(name + ": " + activity.start() + " for " + activity.duration());
-      });
-    } finally {
-      missionModel.getModel().close();
-    }
+    simulationResults.simulatedActivities.forEach((name, activity) -> {
+      System.out.println(name + ": " + activity.start() + " for " + activity.duration());
+    });
```
mattdailis, line 63: I was going to ask why we don't need this try/finally anymore, but I realize now that because `simulate` calls the sim engine's `.close()` method via its own try/finally, we don't need to think about that from out here. It's so _clean_.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/InitializationContext.java
```diff
@@ -4,25 +4,23 @@
...
-    this.executor = Objects.requireNonNull(executor);
+  public InitializationContext(final Initializer builder) {
     this.builder = Objects.requireNonNull(builder);
   }
 
   public static <T>
-  T initializing(final ExecutorService executor, final Initializer builder, final Supplier<T> initializer) {
-    try (final var restore = ModelActions.context.set(new InitializationContext(executor, builder))) {
+  T initializing(final Initializer builder, final Supplier<T> initializer) {
+    try (final var restore = ModelActions.context.set(new InitializationContext(builder))) {
```
mattdailis, line 23: Realizing now that every time we load a mission model just to interrogate its static properties, we will no longer need to spin up an executor service 🕺 I don't know that it was necessarily an expensive operation, but it's certainly semantically clean.

Though, as I write that, I'm remembering that the goal is to not call the constructor at all unless we're running a simulation - but I think this is still a significant step in the right direction.

merlin-framework-junit/src/main/java/gov/nasa/jpl/aerie/merlin/framework/junit/MerlinExtension.java
```diff
@@ -50,23 +45,22 @@ public void beforeAll(final ExtensionContext context) {
...
-  public MerlinTestContext<UNUSED, Model> resolveParameter(final ParameterContext parameterContext, final ExtensionContext extensionContext)
+  public Object resolveParameter(final ParameterContext parameterContext, final ExtensionContext extensionContext)
   throws ParameterResolutionException
   {
     final var state = this.getState(extensionContext);
     final var paramType = parameterContext.getParameter().getType();
 
-    if (!paramType.equals(MerlinTestContext.class)) {
+    if (paramType.equals(Registrar.class)) {
+      if (state.registrar == null) state.registrar = new Registrar(state.builder);
```
mattdailis, line 59: Curious about the null check - could you elaborate briefly?

examples/foo-missionmodel/src/test/java/gov/nasa/jpl/aerie/foomissionmodel/FooActivityTest.java
```diff
@@ -23,27 +22,21 @@
...
-
   private final Mission model;
 
   // Initializers and the test class constructor are executed in an "initialization" Merlin context.
   // This means that models can be created (and cell storage allocated, and daemons spawned),
   //   but simulation control actions like `waitFor`, `delay`, and `emit` cannot be performed.
-  // The `MerlinTestContext` does not need to be declared as a parameter, but will be injected if declared.
-  public FooActivityTest(final MerlinTestContext<ActivityTypes, Mission> ctx) {
+  // The `Registrar` does not need to be declared as a parameter, but will be injected if declared.
+  public FooActivityTest(final Registrar registrar) {
```
mattdailis, line 37: Ah, I see how this improves ergonomics - we definitely need to mention this in a migration guide though, since every single test file will need to be updated :+1:

mattdailis (approved):

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/TaskFactory.java
Twisol: That *is* a good point! The `ExecutorService` is only directly used in `ThreadedTask`, I think, and only to invoke `execute`.

Previously, the `ExecutorService` was itself owned by the framework, so refining the interface didn't make sense. But now I can see an argument for it!

To your point: no, there's no reason to give the modeling side full control of the executor. The model just needs to dispatch the task to us.

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/InitializationContext.java
Twisol: Incremental progress is still good progress :)

merlin-framework-junit/src/main/java/gov/nasa/jpl/aerie/merlin/framework/junit/MerlinExtension.java
Twisol: Ah, it's possible in general for multiple test methods to request a parameter of the same type. We generally don't see this happen for our testing use cases, but I want to make sure that we only instantiate one Registrar and share it amongst everyone associated with the test model.

examples/foo-missionmodel/src/test/java/gov/nasa/jpl/aerie/foomissionmodel/FooActivityTest.java
Twisol: This change also means all tests consistently use `@ExtendWith` instead of sometimes needing to use `@RegisterExtension`. I consider that an ergonomics win too :)

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
```diff
@@ -8,21 +8,21 @@
...
 import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executor;
 import java.util.function.Supplier;
 
 public final class ThreadedTask<Return> implements Task<Return> {
   private final Scoped<Context> rootContext;
   private final Supplier<Return> task;
-  private final ExecutorService executor;
+  private final Executor executor;
```
Twisol, line 17: Wow, I feel stupid. It's nice to know this interface exists 🤦‍♂️

mattdailis: I stumbled on it by accident 😅

camargo (approved):

pcrosemurgy (approved):

PR #383, opened 2022-10-18
[AERIE-2177] Serialize Temporal.Duration as integer in scheduling edsl

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

~~This PR walks #187 back slightly. That PR changed all Durations in the scheduling EDSL to use Temporal.Duration - but this breaks for ActivityTemplate, because the parameters passed to ActivityTemplate are passed unchanged to the activity mapper's `instantiate` method.~~

~~This PR continues to use Temporal.Duration for scheduler-defined values, but goes back to integer microseconds for Activity Parameter values.~~

~~This is a stepping stone to avoid introducing regressions to 0.13.2. We're looking into serializing the Temporal.Duration to microseconds on the Javascript side.~~

UPDATE: After some iteration, this PR instead uses the JSON.stringify `replacer` function to define the serialization of Temporal.Duration.

## Verification

I've added a test to catch the regression - there is now a `DurationParameterActivity` in Banananation.

## Documentation

~~The examples in the [Scheduling Guide](https://nasa-ammos.github.io/aerie/stable/activity-plans/scheduling-guide.html) continue to show controllable duration activity parameters as integers, so I think we're good on that front for now.~~

- [x] The above examples need to be updated prior to release.

~~I've updated the type alias from `Duration` to `IntegerMicroseconds` to hopefully serve as additional documentation while we still have this way of expressing durations.~~

## Future work

- [x] We toyed around with how to serialize Temporal.Duration to microseconds in javascript - @dyst5422 I forget if this is something we had explored previously.

### DIFF
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/DurationParameterActivity.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/DurationParameterActivity.java
@@ -1,0 +1,22 @@
+package gov.nasa.jpl.aerie.banananation.activities;
+
+import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.AutoValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+/**
+ * This activity type intentionally takes a duration as a parameter, but is not a ControllableDuration activity
+ */
+@ActivityType("DurationParameterActivity")
+public record DurationParameterActivity(Duration duration) {
+
+  @EffectModel
+  public ComputedAttributes run(Mission mission) {
+    return new ComputedAttributes(duration);
+  }
+
+  @AutoValueMapper.Record
+  public record ComputedAttributes(Duration duration) {}
+}
--- a/docs/source/activity-plans/scheduling-guide.rst
+++ b/docs/source/activity-plans/scheduling-guide.rst
@@ -104,7 +104,7 @@ activity template and an interval. Let's add those:
   export default (): Goal => {
       return Goal.ActivityRecurrenceGoal({
           activityTemplate: null,
-          interval: 24 * 60 * 60 * 1000 * 1000 // 24 hours in microseconds
+          interval: Temporal.Duration.from({ hours: 24 })
       })
   }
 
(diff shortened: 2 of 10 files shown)


examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/DurationParameterActivity.java
```diff
@@ -0,0 +1,2 @@
+package gov.nasa.jpl.aerie.banananation.activities;public class DurationParameterActivity {
+}
```
Twisol: wat

mattdailis: oop. thanc

patkenneally (approved):

camargo (approved):
LGTM

mattdailis:
I've pushed a commit that re-enables Temporal.Duration in parameters by adding a temporalToMicros function to the scheduling-dsl-compiler's `main.ts` file:

https://github.com/NASA-AMMOS/aerie/blob/604f5ef479d4a87f4f168edb0fda2e8a7672d0f6/scheduler-server/scheduling-dsl-compiler/src/main.ts#L107-L127

We call that function just before calling JSON.stringify:
https://github.com/NASA-AMMOS/aerie/blob/604f5ef479d4a87f4f168edb0fda2e8a7672d0f6/scheduler-server/scheduling-dsl-compiler/src/main.ts#L94

Let me know what you guys think - I can roll the last three commits back easily.

mattdailis:
I think I've found the idiomatic javascript way to add custom serialization logic to `JSON.stringify`:

https://github.com/NASA-AMMOS/aerie/blob/37bdc362876ed3a603da9a16f9f7462f03f4aac7/scheduler-server/scheduling-dsl-compiler/src/main.ts#L94-L102

This feels less hacky than the AST traversal in the comment above.

dyst5422:
@mattdailis I assume this is a temporary fix while we get durations parsing correct on the Java side? I ask because I'm not hot on obfuscated durations (ie non-ISO string serialization) as its less obvious what the context of the value is (especially when it's a well standardized thing like a Duration)

mattdailis:
> @mattdailis I assume this is a temporary fix while we get durations parsing correct on the Java side? I ask because I'm not hot on obfuscated durations (ie non-ISO string serialization) as its less obvious what the context of the value is (especially when it's a well standardized thing like a Duration)

 @dyst5422 Yes, and no. The issue we ran into is that the arguments provided to the activity template in the eDSL are passed directly to the activity type's `instantiate` method (defined in Java). If that activity type uses our `Duration` type, it expects the duration to be serialized as an integer in microseconds.

So - yes, this change is meant to conform to the expectations set on the Java side, but I'm not sure that we have plans to change the serialized `Duration` type in merlin to be a string.

PR #384, opened 2022-10-18
build: update gradle wrapper version

camargo:
## Description

- Use Gradle 7.6 snapshot so we can use Java 19 to build Aerie locally
- Follow up to https://github.com/NASA-AMMOS/aerie/pull/379
- See: https://github.com/gradle/gradle/issues/20372
- See: https://services.gradle.org/distributions-snapshots/
- See: https://services.gradle.org/distributions/ (Will return to pulling from here when Gradle 7.6 is released)

## Future

When Gradle 7.6 is officially released we can switch to use it here

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -67,7 +67,7 @@ jobs:
       uses: actions/setup-java@v3
       with:
         distribution: "temurin"
-        java-version: "18"
+        java-version: "19"
     - run: |
         ./gradlew testClasses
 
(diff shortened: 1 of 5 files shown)


patkenneally (approved):

Twisol:
I didn't need to do this on my computer -- I'm still using 7.5.

Did you set your JAVA_HOME so that Gradle would use Java 18 or earlier? It's no different to IntelliJ running on Java 11 (admittedly its own bundled Java 11). Different programs need different versions.

camargo:
@Twisol No I set my `JAVA_HOME` to use Java 19

Twisol:
That would be why. You can keep using stable Gradle if you set your JAVA_HOME to Java 18.

camargo:
@Twisol Yes I realize. This is just for version consistency.

PR #387, opened 2022-10-19
fix: update activity table to use start_time_doy

camargo:
## Description

- Needed to be in sync with latest UI
- Also use `text` instead of `agTextColumnFilter`


duranb (approved):
👍🏻

PR #391, opened 2022-10-19
[FEAT](AERIE-2146) Include EDSL API docs redo

camargo:
Adds back https://github.com/NASA-AMMOS/aerie/pull/366 following the revert https://github.com/NASA-AMMOS/aerie/pull/390. Please do not merge yet.

### DIFF
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -6,6 +6,8 @@ on:
       - develop
     paths:
       - "docs/**"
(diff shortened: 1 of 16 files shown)


dyst5422:
Theresa and I explored a perhaps more fruitful way to building docs which is ~80% there. She's taking over on it, but I would let her determine if that path is going to work better for us before taking action on this PR.

Mythicaeda:
Covered by #404.

PR #394, opened 2022-10-20
build: upgrade command expansion server deps

camargo:
## Description

- Also reorder scripts
- Cleanup package.json
- Update temporal serializer type


dyst5422 (approved):
All changes make sense to me. I am still working on double checking things locally, will update this comment once I've completed that.

dyst5422 (approved):
All changes make sense to me. Tested on my computer and the changes to the test snapshots are good.

mattdailis (approved):
Approving because Dylan approved - I'm not sure I follow why all those double backslashes were removed in the snapshot.

dyst5422:
> Approving because Dylan approved - I'm not sure I follow why all those double backslashes were removed in the snapshot.

I understand why the expectation is that they shouldn't be there. What I don't understand is how they got there in the first place.

PR #396, opened 2022-10-20
[DEV] minor workflow fixes

dyst5422:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Switches command validation to run in dev mode with hot reloading when code is changed.

Adds a note about environment variables to readme

## Verification

Works on my machine!

## Documentation

Pretty self explanitory

## Future work

None???

### DIFF
--- a/command-expansion-server/Dockerfile
+++ b/command-expansion-server/Dockerfile
@@ -1,4 +1,3 @@
 FROM node:lts-alpine
-COPY . /app
 WORKDIR /app
-CMD [ "npm", "start" ]
+CMD [ "npm", "run", "dev" ]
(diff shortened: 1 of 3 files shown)


command-expansion-server/Dockerfile
```diff
@@ -1,4 +1,3 @@
 FROM node:lts-alpine
-COPY . /app
 WORKDIR /app
-CMD [ "npm", "start" ]
+CMD [ "npm", "run", "dev" ]
```
camargo, line 3: Do we want to do `npm run dev` as the default here? This implies the default production image will run in dev when the container starts.

dyst5422: For some reason I thought we had a different docker image for production. With that in mind, I'm going to completely redo this PR and its going to be more necessary. As it is, the docker image is not sufficient for production (it doesn't do a build internal to the docker image, who knows where its getting the build files).

dyst5422:
Closing in favor of #398

PR #397, opened 2022-10-20
[AERIE-2145] Gaps in external profiles

JoelCourtney:
* **Tickets addressed:** AERIE-2145
* **Review:** By commit  

## Description
by commit:

1. The `nullableP` parser currently cannot unparse nulls because it throws a `NoSuchElementException` when unwrapping. The workaround is to throw a `ClassCastException` which the `chooseP` parser will catch try the next option. @Twisol was this the intended pattern? I don't see another way to do it, but throwing for control flow on the happy path seems iffy.
2. This allows the dynamics in `ProfileSet`s to be Optional. This allows external dataset profiles to have gaps, but also requires all segments of simulated profiles to be wrapped in `Optional.of()`. Also, the `Post/GetProfileSegmentsAction`s were updated to write to the new `is_gap` column of the `profile_segment` table.
3. Due to conversions between our many formats for representing profiles, the duration of the last segment is lost when storing and retrieving from the database. This isn't an issue for simulated profiles because they end at the plan horizon, but external profiles can end in the middle of the plan. So I've added an extra gap segment at the end of all profiles to keep the duration. This required a small change in the UI [here](https://github.com/NASA-AMMOS/aerie-ui/pull/197).

## Verification
I'm in progress writing an e2e test(s) for this.

## Documentation
I have added tutorial docs for uploading external datasets and using them in constraints. I will add a page for scheduling when we enable that.

## Future work
- assignGaps
- assignGaps
- visual representation of violation gaps in UI
- assignGaps


Twisol:
> 1. @Twisol was this the intended pattern?

I'm not sure what you mean here. The `develop` implementation of `nullableP` invokes `Optional::get`, which will of course throw on empty. This doesn't seem to be a problem with the parser system itself.

Here is a custom parser (no combinators) that I think should work. I haven't tested this, and I'm not confident that my `getSchema()` uses the right JSON-Schema syntax, but it's the basic idea.

```java
    return new JsonParser<>() {
      @Override
      public JsonObject getSchema(final SchemaCache anchors) {
        return Json.createObjectBuilder()
            .add("any", Json.createArrayBuilder()
                .add(JsonValue.NULL)
                .add(parser.getSchema()))
            .build();
      }

      @Override
      public JsonParseResult<Optional<T>> parse(final JsonValue json) {
        if (json.getValueType() == JsonValue.ValueType.NULL) return JsonParseResult.success(Optional.empty());

        return parser.parse(json).mapSuccess(Optional::of);
      }

      @Override
      public JsonValue unparse(final Optional<T> value) {
        return value.map(parser::unparse).orElse(JsonValue.NULL);
      }
    };

**Comment by JoelCourtney** [comment]
@Twisol When unparsing `chooseP(A, B)`, choose's unparser will only check B's unparser if a `ClassCastException` is thrown. In cases where the Java types associated with `A` and `B` are different, this is fine because the exception will be caused by chooseP itself. But in this case, both `A` and `N` operate on `Optional<T>`, so the only way for `A` to tell  `chooseP` to try `B` when it encounters `Optional.empty()` is to manually throw a `ClassCastException`.

**Comment by Twisol** [comment]
Right. Because `chooseP` depends on class casts to pick the right parser, it isn't appropriate for this case. Put another way, the individual parsers are supposed to convert to and from values of some type, but one of them only handles some of them (failing on the rest) and the other is lossy because it sends everything to null (and then back to empty).

The individual parsers ought to be total and lossless, so chooseP is not a good combinator to use to decompose this problem with.

**Comment by Twisol** [comment]
I really want a proper sumP akin to productP but with Either instead of Pair. Unfortunately there are serious ergonomics issues with the design I've tried up to this point. And I don't think it would help in this particular situation (since it uses a tag field to disambiguate).

**Comment by JoelCourtney** [comment]
@Twisol just for the sake of getting something working, I reworked it to add `chooseWithPredicateP` with allows each variant to decide whether it can be unparsed from the given value. There's a wee bit of "Don't Repeat Yourself Except When You Need To" because for the life of me I can't figure out how to map a generic varargs to another varargs array without losing all the type information. Thoughts on this solution?

**Comment by Twisol** [comment]
Did you have a chance to evaluate [my alternative solution](https://github.com/NASA-AMMOS/aerie/pull/397#issuecomment-1286220281) from before?

**Comment by JoelCourtney** [comment]
Wow. That there the whole time? I completely missed it. Please hold

**Comment by JoelCourtney** [comment]
That'll do it 🤦 . Thanks!

**Comment by Twisol** [comment]
Glad to be of assistance :)

**Comment by JoelCourtney** [comment]
Turns out we already had a docs page for external datasets, and its better written! I'm going to update it and move it to the new directory instead of using my own.

edit: done

File: merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java [hunk 1]
```diff
@@ -11,13 +11,14 @@
...
+import java.util.Optional;
 
 import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatements.setTimestamp;
 
 public final class PostProfileSegmentsAction implements AutoCloseable {
   private final @Language("SQL") String sql = """
-      insert into profile_segment (dataset_id, profile_id, start_offset, dynamics)
-      values (?, ?, ?::timestamptz - ?::timestamptz, ?)
+      insert into profile_segment (dataset_id, profile_id, start_offset, dynamics, is_gap)
+      values (?, ?, ?::timestamptz - ?::timestamptz, ?, ?)
```
pcrosemurgy: Can change this to:
```sql
values (?, ?, ?::timestamptz - ?::timestamptz, ?::json, ?)
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
```diff
@@ -46,13 +47,32 @@ public <Dynamics> void apply(
       this.statement.setLong(2, profileRecord.id());
       setTimestamp(this.statement, 3, timestamp);
       setTimestamp(this.statement, 4, simulationStart);
-      this.statement.setString(5, serializeDynamics(dynamics, dynamicsP));
+      if (dynamics.isPresent()) {
+        this.statement.setString(5, serializeDynamics(dynamics.get(), dynamicsP));
+        this.statement.setBoolean(6, false);
+      } else {
+        this.statement.setString(5, "null");
```
pcrosemurgy:
```sql
dynamics jsonb null
```
Given this definition of `dynamics` what's the semantic difference between a `null` `dynamics` and a `dynamics` with a JSON value of `null`?

merlin-server/sql/merlin/tables/profile_segment.sql
```diff
@@ -4,6 +4,7 @@ create table profile_segment (
 
   start_offset interval not null,
   dynamics jsonb null,
+  is_gap bool not null default false,
```
pcrosemurgy, line 7: Following our other tables it'd be nice to have comments for each of these columns. [`profile.sql`](https://github.com/NASA-AMMOS/aerie/tree/develop/merlin-server/sql/merlin/tables/profile.sql) could be a decent example.

This would help devs/users understand what a null-able column may mean in the context of our data model.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
JoelCourtney: Its very blurry, and it doesn't actually matter whether the column is `null` or `not null`.

In the database, pg will convert JSON nulls into SQL nulls, but specifying `jsonb not null` doesn't restrict your ability to put JSON nulls into the column. So in effect, `jsonb not null` is a lie. If you say that the SQL value is `null`, that's an error. But if you say the value is a json object of `null`, it says "lol whatever you want boss" and promptly converts it into a SQL null without error. Its madness

Conceptually speaking from the user perspective, `null` is a legitimate value that they can use at their discretion. "Gaps" specifically mean unknown value, and I guess its more closely related to `void`.

pcrosemurgy: Ohh! Interesting. Thanks for the explanation, that's very good to know.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
```diff
@@ -50,23 +52,37 @@ public <Dynamics> List<Pair<Duration, Dynamics>> get(
...
+              .getSuccessOrThrow(
+                  failureReason -> new Error("Corrupt profile dynamics: " + failureReason.reason())));
+        } else {
+          dynamics = Optional.empty();
+        }
       }
 
       final var duration = simulationDuration.minus(offset);
       segments.add(Pair.of(duration, dynamics));
     }
```
pcrosemurgy, line 84: Wouldn't hurt to change this `}` to:
```java
    } else {
      throw new Error("No profile segments found for `dataset_id` (%d) and `profile_id` (%d)".formatted(datasetId, profileId));
    }
```
This way a legible error would be shown in this case instead of `segments.remove()` below causing an array index out of bounds exception.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
```diff
@@ -24,25 +25,33 @@ static ProfileSet getProfiles(
...
+                getDiscreteProfileSegments(connection, record.datasetId(), record.id(), simulationWindow)
+            )
+        );
         default -> throw new Error("Unrecognized profile type");
       }
     }
 
-    return new ProfileSet(realProfiles, discreteProfiles);
+    final var result = new ProfileSet(realProfiles, discreteProfiles);
+    System.out.println(result);
```
pcrosemurgy: Maybe this was added for testing and was meant to be removed. If you'd like to keep it around it may help to give this message some context. :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
pcrosemurgy: (I'm referring to line 82 here btw)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
JoelCourtney: ha, yup that's me debugging

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
```diff
@@ -46,13 +47,32 @@ public <Dynamics> void apply(
...
+      }
 
       this.statement.addBatch();
 
       accumulatedOffset = Duration.add(accumulatedOffset, duration);
     }
 
+    // Profiles must explicitly end in a gap, because the last segment's duration data is lost when
+    // inserting into the database.
+    if (segments.get(segments.size()-1).getValue().isPresent()) {
```
mattdailis: 🔶 Just wondering - what would be the pros/cons of this approach vs adding a start/end time to the `profile` table?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
```diff
@@ -50,23 +52,39 @@ public <Dynamics> List<Pair<Duration, Dynamics>> get(
...
+                  failureReason -> new Error("Corrupt profile dynamics: " + failureReason.reason())));
+        } else {
+          dynamics = Optional.empty();
+        }
       }
 
       final var duration = simulationDuration.minus(offset);
       segments.add(Pair.of(duration, dynamics));
+    } else {
+      throw new Error("No profile segments found for `dataset_id` (%d) and `profile_id` (%d)".formatted(datasetId, profileId));
```
mattdailis, line 83: 🔶 Is my understanding correct that this is an `Error` because we're guaranteed to have at least one segment - the "final gap" segment that denotes the end time of the profile?

docs/source/index.rst
```diff
@@ -67,6 +67,15 @@
 
   Deploy Aerie for individual users, small and large missions.
 
+.. topic-box::
+  :title: External Datasets
+  :link: external-datasets
+  :icon: scylla-icon scylla-icon--resource-center
+  :class: large-4
```
mattdailis: 🔵 Any reason this one should be large-4 while the others are large-6?

Screen Shot 2022-10-28 at 8 16 03 AM

docs/source/external-datasets/index.rst
```diff
@@ -0,0 +1,13 @@
+=================
+External Datasets
+=================
+
+.. toctree::
+   :hidden:
```
mattdailis: 🔵 Would exposing this table of contents be helpful, or do you figure it's redundant with the sidebar?

My initial reaction on pulling up the docs was "oh, that's it?", and then I noticed the arrow on the bottom right pointing that took me to "Managing External Datasets" - and only afterwards did I realize I could have used the sidebar to find the child pages.

I'm sure we'll iron out our "docs best practices" when we kick off the next sprint, so I'm okay with leaving this as is for now.

mattdailis (commented):
> Key:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
JoelCourtney: Theoretically none but it would take a bit more refactoring. Currently the `dataset` holds the start time, and its assumed that all profiles in the dataset start at the same time and end at infinity (but get truncated at the planning horizon). We could store a `duration` in `profile`? That seems a little cleaner than I have here. I only took this approach because it seemed like the smallest change that would do it.

docs/source/index.rst
JoelCourtney: whoops I was experimenting with shrinking all the boxes, forgot to undo it.

docs/source/external-datasets/index.rst
JoelCourtney: I agree in this case it would be helpful. But in longer sections it can be a gigantic wall of text because it includes all subheadings. Maybe a collapsible like this would do it:


<details>
<summary>hey click me</summary>
<ul>
<li>this is my</li>
<li>favorite trick on github</li>
</ul>
</details>

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfileSegmentsAction.java
JoelCourtney: Yes, but I think its also safe to say that if you don't find any segments for a profile that's supposed to exist, something has gone pretty wrong.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
mattdailis: If it's not too much work, I do think it would be less surprising for the `duration` to be defined at the profile level, 

I've been waffling on this, wondering if your way would be more conducive to streaming, but I think the profile duration interpretation would be more consistent in either case (no profile segment would have special meaning as the "end" segment)

mattdailis: More musings that might not be moving us towards a decision...

I suppose the interpretation is that the last segment applies until infinity. If your last segment is non-null, you are claiming that that segment applies until infinity. If your last segment is null, you are explicitly saying "I don't know the value from here until infinity". 

That interpretation matches your current implementation reasonably well - the part I'm not sure about is the requirement that the last segment be null...

pcrosemurgy (approved):
Nice work! Sounds like e2e tests will be added to this PR prior to merge.

JoelCourtney:
Uncovered a problem while testing, somewhere along the way it makes an assumption that the external dataset starts at plan start. Will fix.

docs/source/external-datasets/index.rst
mattdailis: We can try replacing `:hidden:` with `:maxdepth: 2` or `:maxdepth: 1` and see how that looks:
<details>
<summary><code>:maxdepth: 1</code></summary>

Screen Shot 2022-10-28 at 5 06 36 PM
</details>
<details>
<summary><code>:maxdepth: 2</code></summary>

Screen Shot 2022-10-28 at 5 06 57 PM
</details>

(I like your trick 💯)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
Twisol: Yes, I'd also like to make sure that when we begin streaming results into the database from the simulation worker, we can easily append new segments to the profile with minimal overhead.

I'm a little nervous that including an explicit gap could require *replacing* that row when streaming new data into that area (e.g. if the current segment is merely extended rather than the profile taking on a new value). If I recall correctly, we don't partition the table of profile segments on a per-resource basis (that's far too fine-grained), only on a per-simulation basis. That means we might have to go hunting for the null row that needs to get updated; it's not guaranteed to be at the end.

For that reason, I prefer giving the `profile` field a duration -- acknowledging that this may also require more writes by the worker to subsequently update that duration. I think an append-only regime is more important, and the `profiles` table is going to be far smaller (and better indexed for this update) than the segments table.

Regardless of approach, I do think we need to delimit the interval of time over which we *actually* have results; if we haven't determined the value of a profile past time T, we ought not claim that the last segment extends beyond it (especially not to infinity). It *might*, but we don't know.

mattdailis: Thanks for weighing in, @Twisol , that makes me feel better about our decision to go forward with the `duration` column on profile.

docs/source/external-datasets/index.rst
JoelCourtney: I missed most of the discussion yesterday. Did we talk about anything relevant to this?

docs/source/index.rst
JoelCourtney: done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostProfileSegmentsAction.java
JoelCourtney: done

JoelCourtney:
- Fixed problem that was forcing datasets to start at the plan start.
- Added e2e test
    - It doesn't test the full process of adding and querying an external dataset because the code for retrieving datasets is not exposed as a hasura action. So instead it uploads a dataset and queries it directly from hasura's autogenerated `plan_dataset_by_pk` query.
    - This tests that we correctly store datasets that don't start at the plan start, and that we store gaps in the middle of profiles.
- Added tutorial docs
- Switched to using a `duration` column on the profile table instead of an extra segment.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PreparedStatements.java
```diff
@@ -29,6 +31,13 @@ private PreparedStatements() {}
           .appendOffset("+HH:mm:ss", "+00")
           .toFormatter();
 
+  private static final @Language("SQL") String intervalStylePrefix = "set intervalstyle = ";
+
+  public static void setIntervalStyle(final Connection connection, final String intervalStyle) throws SQLException {
```
pcrosemurgy: This is very minor but I would suggest either renaming this something like `setIntervalStyleISO8601` and removing the `String` param. or making `intervalStyle` an enum with currently one option, `iso_8601`.

pcrosemurgy (approved):
Re-reviewing after the recent fixes, documentation, and e2e tests. Looks good to me!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PreparedStatements.java
JoelCourtney: There's four options, but yeah for now we're only using iso8601. I'll make an enum.

pcrosemurgy: Thanks :) Feel free to resolve this when done

mattdailis (approved):

PR #398, opened 2022-10-20
docs: update command server Dockerfile and deployment readme

dyst5422:
* **Review:** By commit  

## Description

Fixes the Dockerfile for command expansion to be suitable for production
- copies source into image
- installs in image
- runs from built files

## Verification

Built the image locally and executed locally

## Documentation

Self explanatory

### DIFF
--- a/command-expansion-server/.dockerignore
+++ b/command-expansion-server/.dockerignore
@@ -1,0 +1,3 @@
+.gradle
+sql
+test
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -29,12 +29,12 @@ They are, as follows:
 | `REPOSITORY_DOCKER_URL` | The URL used to fetch images of Aerie packages.                                                                              |
 | `DOCKER_TAG`            | The version of the Aerie images to fetch.                                                                                    |
 | `AERIE_USERNAME`        | The username used for Aerie services when they access the Postgres database.                                                 |
-| `AERIE_PASSWORD`        | The password used for Aerie services when they access the Postgres database.                                                 ||
-| `POSTGRES_USER`         | The username of the superuser for the Postgres database.                                                                     |
+| `AERIE_PASSWORD`        | The password used for Aerie services when they access the Postgres database.                                                 |
+| `POSTGRES_USER`         | The username of the superuser for the Postgres database. This cannot be the same as the `AERIE_USERNAME` value.              |
 | `POSTGRES_PASSWORD`     | The password of the superuser for the Postgres database.                                                                     |
 | `UNTRUE_PLAN_START`     | Temporary solution to provide plan start time to models, should be set to a time that models will not fail to initialize on. |
 
-### **At minimum, you _must_ assign values to the environment variables already present in [.env](./.env) in order to deploy Aerie**.
+### **At minimum, you _must_ assign values to the environment variables already present in [.env](./.env) in order to deploy Aerie. Additionally, the `POSTGRES_USER` value _cannot_ be the same as the `AERIE_USERNAME` value.**.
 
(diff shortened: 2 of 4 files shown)


deployment/README.md
```diff
@@ -29,12 +29,12 @@ They are, as follows:
...
-| `POSTGRES_USER`         | The username of the superuser for the Postgres database.                                                                     |
+| `AERIE_PASSWORD`        | The password used for Aerie services when they access the Postgres database.                                                 |
+| `POSTGRES_USER`         | The username of the superuser for the Postgres database. This cannot be the same as the `AERIE_USERNAME` value               |
 | `POSTGRES_PASSWORD`     | The password of the superuser for the Postgres database.                                                                     |
 | `UNTRUE_PLAN_START`     | Temporary solution to provide plan start time to models, should be set to a time that models will not fail to initialize on. |
 
-### **At minimum, you _must_ assign values to the environment variables already present in [.env](./.env) in order to deploy Aerie**.
+### **At minimum, you _must_ assign values to the environment variables already present in [.env](./.env) in order to deploy Aerie. Additionally, the `POSTGRES_USER` value CANNOT be the same as the `AERIE_USERNAME` value.**.
```
Mythicaeda: Can we either have CANNOT be _cannot_ or _must_ be MUST for consistency?

command-expansion-server/.dockerignore
```diff
@@ -0,0 +1,6 @@
+build
+cdict
+node_modules
+.gradle
+sql
+test
```
Mythicaeda: Github's saying this file's missing a terminating newline.

command-expansion-server/Dockerfile
```diff
@@ -1,4 +1,9 @@
 FROM node:lts-alpine
+
 COPY . /app
 WORKDIR /app
+
+RUN npm ci
```
camargo: I think it's best practice to not actually build the app when we build the Docker image. The Dockerfile should assume a build has already occurred so we can call `npm ci` in actual CI (on the GH action server in this case) and cache the node_modules as needed.

dyst5422: The problem I see with that is that the eventual execution environment is the docker container, NOT the CI environment. So for things like self-compiling binary packages installed with rpm, you can have a mismatch on target platform.

camargo: Yes true. I think if we ever need to export binary packages we can revisit. The general pattern in the repo is Gradle creates the build products in CI and our Dockerfiles copy them into place at Docker image build time. I don't want to diverge from that right now.

dyst5422: It's not about us exporting binaries, but about packages we depend on. Any package we depend on that has a build process using node-gyp can cause us problems. Really any package that uses platform specific binaries. And because of node package resolution, it could be a dependency of a dependency.

camargo: Ah yeah I hear you. For now we will just have to keep an eye on the system requirements in our dependencies and transitive dependencies and make sure our CI agents supports them.

camargo: Just to elaborate @dyst5422 I don't think this is a bad idea. My main thing is I just want our build processes to be as consistent as possible across the different services in the repo so it's easier to maintain. So if we decide to add build steps to our Dockerfiles I think we should do it throughout the repo, and also spend some time to research a bit more on how to do it correctly so we can take advantage of things like agent dependency caching.

dyst5422: I hear and agree with you. I have never done a workflow where build occurs in CI and then copied into containers, so apologies if I seem alarmist.

camargo (approved):

PR #399, opened 2022-10-21
[AERIE-2175]  Plan Merging 

Mythicaeda:
* **Tickets addressed:** AERIE-2175
* **Review:** By commit  

## Description

This PR depends on #380.

With this PR, merge requests and the merge process is introduced. Merges place their in-progress work into two tables: a staging area and a conflicting activities table. What goes where is described [here](https://github.com/NASA-AMMOS/aerie/discussions/349#discussioncomment-3909537). Once all conflicts are resolved, a merge request may be committed. When a merge request is committed, the resolved conflicts are copied into the staging area, which is used to update the target plan (the one receiving changes). 

This PR updates the database in the following ways:
- adds the column `is_locked` to `plan`, which holds a boolean
- adds a function to the triggers in `activity_directive` to prevent UPDATES, INSERTS, or DELETES for activities from a locked plan
- updates the DELETE trigger on `plan` to prevent deleting a locked plan
- adds the following tables: `merge_request`, `merge_request_comment`, `merge_staging_area`, `conflicting_activities`

## Verification

Added tests for:
- Plan Locking/Unlocking and its database triggers
- Finding the Merge Base between two plans
- Functions in `merge_request.sql`
- Beginning Merges, and specifically what is and isn't a conflict
- Committing Merges
- Ensuring that MergeRequests States can only transition as defined in [this image](https://github.com/NASA-AMMOS/aerie/discussions/349#discussioncomment-3843039)

## Documentation

This PR will supply the documentation to the website for this and #380. For a preview of important information while that is being written, see [this discussion](https://github.com/NASA-AMMOS/aerie/discussions/349).

## Future work

Adding support for duplicating/merging temporal subsets of plans.


merlin-server/sql/merlin/plan_merge.sql
```diff
@@ -0,0 +1,592 @@
...
+       left join activity_directive act on m.plan_id_receiving_changes = act.plan_id and c.activity_id = act.id
+union all
+select request_id as merge_request_id,
+       activity_id,
+       change_type as change_type_source,
+       'none' as change_type_target,
+       'source' as resolution,
+       snap_act as source,
+       act as target,
+       'false'::boolean as is_conflict
```
AaronPlave: Can someone explain this `is_conflict = false` case to me a bit more? Trying to understand if this needs to be displayed in the client. And on a related note – I'm not yet understanding why you'd end up with merge review activities where `is_conflict = false` + `resolution = source or target`. Does that indicate that these have been automatically resolved and no action is needed? And if so why make them merge review activities?

AaronPlave: Or maybe the join to create this view here has an issue with the logic?

mattdailis: That's a fair point. The thought process there was that non-conflicting activities have an implicit resolution of `source`, so that after a merge is committed, the mental model is that all of the activities with a resolution of `source` will match the `source` plan, and all activities with a resolution of `target` will match the version in the `target` plan.

This sort of makes sense thinking forward to a future capability to selectively accept/reject all changes (not just conflicts), but I can see that it doesn't make a lot of sense now.

Happy to rework this to better match UI needs/make invalid states unrepresentable. This view could be simplified to only roll up the conflicting activities, and then the UI would make two queries (which it could do with only one round trip) - one to get the non-conflicting activities from `merge_staging_area`, and one to get conflicting activities.

camargo (approved):
LGTM! Let's get this merged in today so we can do a `0.13.3` release so they can start playing with it.

merlin-server/sql/merlin/merge_request.sql
```diff
@@ -0,0 +1,85 @@
...
+
+create table merge_request(
+      id integer generated always as identity
+        primary key,
+      plan_id_receiving_changes integer,
+      snapshot_id_supplying_changes integer,
+      merge_base_snapshot_id integer not null,
+      status merge_request_status default 'pending',
+      requester_username text not null,
+      reviewer_username text not null default ''
```
mattdailis: I'm curious about pros/cons of using the empty string as the default here, rather than `null`.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -723,5 +968,1640 @@ void grandparentAdoptsChildrenOnDelete() throws SQLException{
...
+
+      deleteActivityDirective(basePlan, deleteContestedReceivingActId);
+      updateActivityName("Delete Contested Receiving Child", deleteContestedReceivingActId, childPlan);
+
+      final int mergeRQ = createMergeRequest(basePlan, childPlan);
+      beginMerge(mergeRQ);
+
+      assertEquals(4, getConflictingActivities(mergeRQ).size());
+      setResolution(mergeRQ, modifyContestedSupplyingActId, "supplying");
+      setResolution(mergeRQ, modifyContestedReceivingActId, "receiving");
```
mattdailis, line 2061: Did we standardize on `source` and `target`?

skovati (approved):
Really nice work on this! I'll have to study the `plpgsql` a bit more to really grok it, but LGTM. Getting this feature into `0.13.3` lets us iron out any potential issues before `1.0.0`

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: My understanding of the situation was that the backend uses "supplying" and "receiving" while the front end displays "source" and "target". Or did we decide to standardize between the two?

mattdailis: I see - I guess it would be good for the two to be consistent, but maybe not completely required.

The `setResolution` jumped out at me particularly because I was under the impression that it is visible via the api - when the client updates the conflicting_activities table to set a resolution - am I overlooking a translation step here?

Mythicaeda: IIRC, the inconsistency arose because the DB wants to be clear while the UI abstracted for the sake of conciseness/not being ugly. Unless you mean that our Hasura functions should be consistent with the UI terminology? In which case, I don't disagree, although it does beg the question as to if we should add Hasura functions specifically to do this `source`->`supplying` `target`->`receiving` mapping, since currently the [example query for resolving conflicts](https://github.com/NASA-AMMOS/aerie/pull/399/commits/3795c44f8002c69f589284af866349aab517321c#diff-a0597b372569429c046ef7d6732fc976ac99296101ad8bdb066697775709d65cR409-R441) is just doing a straight table update (and therefore using the DB terms).

As for if you're missing a translation point, maybe? This `setResolution` is taking place in our `dbTests`. [The method](https://github.com/NASA-AMMOS/aerie/pull/399/files#diff-269125a2369e062ea57b1a0ae448f5e159d3c8f69552c13ace30c65533e246b2R416-R426) queries the database directly rather than going through the API, which is why it's using the DB term.

merlin-server/sql/merlin/merge_request.sql
Mythicaeda: This was mostly a style choice instead of a pros/cons choice, as we can get equivalent behavior with either `reviewer_username text not null default ''` or `reviewer_username text default null`. The main two differences I can think of are:
- Whoever works with the data is guaranteed that getting `NULL` means that something's wrong
- `reviewer_username` can never go from having a value to `NULL`.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: > the [example query for resolving conflicts](https://github.com/NASA-AMMOS/aerie/pull/399/commits/3795c44f8002c69f589284af866349aab517321c#diff-a0597b372569429c046ef7d6732fc976ac99296101ad8bdb066697775709d65cR409-R441) is just doing a straight table update (and therefore using the DB terms)

👆 yes - this is what I was thinking about - the "straight table update" is part of our api (since we're exposing the table via hasura), so I'm proposing that all client-visible terminology be consistent. 

>  it does beg the question as to if we should add Hasura functions specifically to do this

Yes - this is what I'm imagining a "translation point" would look like - though I would advocate for changing the DB terms to be consistent with the API terms rather than adding a function to translate between them. Is there an advantage to using one set for the API and a different set internally?

merlin-server/sql/merlin/merge_request.sql
mattdailis: Gotcha - thanks for clarifying :+1: I would suggest we use `null` rather than an empty string to mean "no data" here - we used a similar pattern for the `reason` field in simulation_dataset (which is always null unless the simulation `status` is `failed`):
https://github.com/NASA-AMMOS/aerie/blob/c834e9c05d21b32cd07f5018c2ccb47ad39de94a/merlin-server/sql/merlin/tables/simulation_dataset.sql#L21

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: Mainly clarity. One thing that has come up often during this work is that `source` and `target` _are_ confusing, while `snapshot_supplying_changes` and `plan_receiving_changes` keep it obvious what exactly something _is_ and what exactly it's _doing_ in the merge. Personally, I'd rather have the code be understandable at the cost of a translation point at either the UI or API level rather than consistent at all points, especially with something like merging, where the code is already complicated enough without adding in a prerequisite of having to keep clear what `source` and `target` are.

While we could change the definition of `conflict_resolution` to `{ none, source, target }`, this would just move the translation point to inside the database, which I'd argue is worse than having it on the communications interface.

A question I have: does the API _need_ to use the UI display terms?

merlin-server/sql/merlin/merge_request.sql
Mythicaeda: Ah, if we're already using that pattern, then sure, I can change it to `reviewer_username text default null`.

Mythicaeda: Done

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: > A question I have: does the API need to use the UI display terms?

It technically doesn't - certainly if we ever add internationalization, the words displayed in the UI will differ from the terms in the API. However, there is a lot of benefit to having a [ubiquitous language](https://www.agilealliance.org/glossary/ubiquitous-language) that is shared among developers, product people, and customers - a shared vocabulary reduces friction and helps everyone be on the same page.

If we do go the route of adding a translation layer, we'll need to hide the ~~`merge_staging_area` and~~ `conflicting_activities` table~~s~~, and provide views that translate between the terms.

mattdailis: I guess we already wrap queries to those tables, so we just need to add the resolution wrapper

Mythicaeda: That was exactly what I was going to comment on. Should I add the wrapper to `conflict_resolution` to finalize the translation layer, or is there a different resolution we ought to discuss first?

mattdailis: That sounds good, one more wrapper and at least the API-facing terminology will be consistent :+1:

Twisol: > However, there is a lot of benefit to having a [ubiquitous language](https://www.agilealliance.org/glossary/ubiquitous-language) that is shared among developers, product people, and customers - a shared vocabulary reduces friction and helps everyone be on the same page.

It's worth noting that the ubiquitous language comes *after* the bounded context. That is, it's ubiquitous *within* a context. For example, while Merlin is predominantly used in a planning context where "activities" are the unit of control, Merlin can be used in other contexts, e,g. sequence simulation, where "commands" are the unit of control. We use the term "directive" internally to encompass both of these.

merlin-server/sql/merlin/plan_merge.sql
```diff
@@ -0,0 +1,619 @@
...
+    - uploading a resolved conflict with either the tag 'delete' or 'modify' depending on context
+      -  (none and add cannot cause conflict)
+      -  (a new version of the conflict resolution will replace the version currently in the table
+    - deletes are placed as the full activity, but are 'marked to delete' essentially
+-- Because of this, to get the full list of activities involved in an in-progress merge requires querying here and conflicting_activities
+ */
+-- TODO: Consider removing some of the defaults from this table? Decision depends on if we can have an activity
+--        that doesn't have a field in this area somehow.
+create table merge_staging_area(
+    request_id integer
```
AaronPlave: Can we rename this to `merge_request_id` to be consistent with the `merge_request_id` column in the `conflicting_activities` table?

Mythicaeda: Done

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: > That sounds good, one more wrapper and at least the API-facing terminology will be consistent 👍

Wrapper has been implemented. It's named `resolution_type` to avoid having to do explicit schemas whenever we convert. It's in `hasura_functions.sql`.

mattdailis (approved):
Great work, Theresa 🎉 I'm excited to see this in action.

PR #402, opened 2022-10-24
[AERIE-2141] Add during expression by updating ForEach and Spans 

adrienmaillard:
* **Tickets addressed:** AERIE-2142
* **Review:** By commit  

## Description

I have followed what was proposed by @JoelCourtney 
- A `Spans.ForEachActivity` function/AST node has been added in the eDSL that allows the lambda to return a `Spans` object
- Spans now have an optional activity metadata associated with each interval
- The `window()`, `start()` and `end()` methods in `ActivityInstance` in eDSL have been updated to return `Spans` with activity id metadata
- `ViolationsOf` is now `ViolationsOfWindows`. 

Finally, the `Windows.During(ActivityTypes[]): Windows` was added to the constraints eDSL. 

## Verification

Many tests were updated. 

## Documentation

- [ ] Update the scheduling documentation when this is accepted.

## Future work

The mutex global scheduling condition (AERIE-2142) in #410.


merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -602,29 +624,29 @@ export class ActivityInstance<A extends ActivityType> {
   /**
    * Produces a window for the duration of the activity.
    */
-  public window(): Windows {
-    return new Windows({
-      kind: AST.NodeKind.WindowsExpressionActivityWindow,
+  public window(): Spans {
+    return new Spans({
+      kind: AST.NodeKind.SpansExpressionActivityWindow,
```
JoelCourtney: This should probably be renamed to span()

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -602,29 +624,29 @@ export class ActivityInstance<A extends ActivityType> {
...
   /**
    * Produces an instantaneous window at the end of the activity.
    */
-  public end(): Windows {
+  public end(): Spans {
     return new Windows({
       kind: AST.NodeKind.WindowsExpressionEndOf,
       alias: this.__alias
-    });
+    }).spans();
```
JoelCourtney: Please refactor start() and end() the same way you refactored `window()` so that the spans can get the activity metadata.

Alternatively this can just be syntactic sugar for `start() = window().starts()` and `end() = window().ends()`, which is already implemented for Spans. Whichever you prefer

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -7,65 +7,108 @@
...
+        .stream(iter.spliterator(), false)
+        .filter($ -> !$.getLeft().isEmpty())
+        .forEach(this.intervals::add);
+  }
+
+  public void addAllWithMetadata(Iterable<Pair<Interval, Optional<Metadata>>> iter) {
+    StreamSupport
+      .stream(iter.spliterator(), false)
+      .filter($ -> !$.getLeft().isEmpty())
+      .forEach(this.intervals::add);
```
JoelCourtney: looks like these two ended up with identical behavior

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -7,65 +7,108 @@
...
 import gov.nasa.jpl.aerie.constraints.time.Interval.Inclusivity;
+import org.apache.commons.lang3.tuple.Pair;
 
 /**
  * A collection of intervals that can overlap.
  */
-public class Spans implements IntervalContainer<Spans>, Iterable<Interval> {
-  private final List<Interval> intervals;
+public class Spans implements IntervalContainer<Spans>, Iterable<Pair<Interval, Optional<Spans.Metadata>>> {
+  private final List<Pair<Interval, Optional<Metadata>>> intervals;
```
JoelCourtney: I'd slightly prefer reusing the `Segment<T>` record that `IntervalMap` uses, but its not a big deal. You can ignore this if it would be a lot of work to change or if you don't have time.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ActivityWindow.java
```diff
@@ -3,26 +3,24 @@
...
-import gov.nasa.jpl.aerie.constraints.time.Windows;
+import gov.nasa.jpl.aerie.constraints.time.Spans;
+import org.apache.commons.lang3.tuple.Pair;
 
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 
-public final class ActivityWindow implements Expression<Windows> {
+public final class ActivityWindow implements Expression<Spans> {
```
JoelCourtney: Please rename to ActivitySpan

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -47,11 +47,10 @@ export interface ForEachActivity {
   kind: NodeKind.ForEachActivity;
   activityType: string;
   alias: string;
-  expression: Constraint;
+  expression: SpansExpression;
```
JoelCourtney, line 59: I think this should be `WindowsExpression | SpansExpression`

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -72,13 +71,15 @@ export type WindowsExpression =
...
+  | IntervalsExpressionEnds
 
 export type SpansExpression =
+  | SpansExpressionActivityWindow
   | SpansExpressionSplit
   | IntervalsExpressionStarts
   | IntervalsExpressionEnds
-  | SpansExpressionFromWindows;
+  | SpansExpressionFromWindows
+  | ForEachActivity
```
JoelCourtney: please put the semicolons back

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -7,65 +7,108 @@
...
+
+  public record Metadata(long id){}
```
JoelCourtney, line 23: I think this needs to support a list of ids. Otherwise we can't provide both activities in a violation of a mutex constraint.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -31,36 +30,15 @@ export class Constraint {
...
-  public static ForbiddenActivityOverlap(activityType1: Gen.ActivityType, activityType2: Gen.ActivityType): Constraint {
-    return Constraint.ForEachActivity(
+  public static ForbiddenActivityOverlap(activityType1: Gen.ActivityType, activityType2: Gen.ActivityType): Spans {
+    return Spans.ForEachActivity(
         activityType1,
-        activity1 => Constraint.ForEachActivity(
+        activity1 => Spans.ForEachActivity(
             activityType2,
-            activity2 => Windows.All(activity1.window(), activity2.window()).invert()
+            activity2 => Windows.All(activity1.window().windows(), activity2.window().windows()).invert().spans()
```
JoelCourtney: We're gonna need some more operations for Spans so we can do this without converting to Windows and back. As-is this will lose the activity metadata.

Not your concern for this PR though, just a note for the future

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -73,6 +51,26 @@ export class Windows {
     this.__astNode = expression;
   }
 
+  /**
+   * Produces windows for each activity present in the plan and belonging to one of the activity types passed
+   *
+   * @param activityTypes the activity types
+   */
+  public static duringAll<A extends Gen.ActivityType>(list: Gen.ActivityType[]) : Windows {
+    return Windows.All(
```
JoelCourtney: I think you intended to use Windows.Any? `All` is intersection/and, `Any` is union/or.

I've realized recently that the names `All` and `Any` are kinda vague. I plan to open a bikeshedding PR this week that changes these names back to `And` and `Or`, like they were before I came along and thought I knew better ;)

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -808,6 +832,20 @@ declare global {
...
      */
     public windows(): Windows;
+
+
+    /**
+     * Applies an expression for each instance of an activity type.
+     *
+     * @param activityType activity type to check
+     * @param expression function of an activity instance that returns a Spans
+     * @constructor
```
JoelCourtney: please update this (and the implementation's comment) to describe that it produces Spans with associated activity IDs.

JoelCourtney (changes requested):
thanks! I have one major concern that I should have seen coming. I think we need to reverse how spans translate to violations. Otherwise we lose the activity metadata in the `invert()` call associated with specific spans. I saw in your implementation that you keep the IDs around and then associate all the violations with all of the IDs, which is better than nothing, but definitely a regression from `Constraint.ForEachActivity`. I messaged you on Slack because its probably easier to discuss this over huddle

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: revised: we should have both window() and span() since we are keeping Constraint.ForEachActivity

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ActivityWindow.java
JoelCourtney: Revised: better for now to have both ActivityWindow and ActivitySpan

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: edit: my comment was incorrect to begin with :), and even more wrong with the revision

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
adrienmaillard: Segment is fine.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsFromSpans.java
```diff
@@ -29,4 +30,18 @@ public String prettyPrint(final String prefix) {
         this.expression.prettyPrint(prefix + "  ")
     );
   }
+
+  @Override
+  public boolean equals(Object obj) {
```
mattdailis: Based on my understanding of records, I would expect these `equals` and `hashCode` methods to be identical to the default record behavior. Is that not the case?

https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/Record.html#equals(java.lang.Object)

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: We now have both.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: Took the syntactic sugar road.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
adrienmaillard: Removed the second one.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ActivityWindow.java
adrienmaillard: Yes.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
adrienmaillard: done

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: @adrienmaillard I guess this can be safely ignored for now, since we reduced the scope.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -7,65 +7,102 @@
...
+                 .forEach(itv -> this.intervals.add(Segment.of(itv, Optional.empty())));
   }
 
   public Spans(final Interval... intervals) {
-    this.intervals = Arrays.stream(intervals).filter($ -> !$.isEmpty()).toList();
+    this.intervals = new ArrayList<>();
+    Arrays.stream(intervals).filter($ -> !$.isEmpty()).forEach(itv -> this.intervals.add(Segment.of(itv, Optional.empty())));
+  }
+
+  public Spans(final Segment<Optional<Metadata>>... intervals) {
```
mattdailis, line 49: Because the varargs type is generic, and Java arrays do not handle generics well, Java gives us a warning unless we pinky promise that all the operations we're doing are type safe:

```suggestion
  @SafeVarargs
  public Spans(final Segment<Optional<Metadata>>... intervals) {
```
```
warning: [unchecked] unchecked generic array creation for varargs parameter of type Segment<Optional<Metadata>>[]
warning: [unchecked] Possible heap pollution from parameterized vararg type Segment<Optional<Metadata>>
```

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -73,6 +73,26 @@ export class Windows {
     this.__astNode = expression;
   }
 
+  /**
+   * Produces windows for each activity present in the plan and belonging to one of the activity types passed
+   *
+   * @param activityTypes the activity types
+   */
+  public static duringAll<A extends Gen.ActivityType>(list: Gen.ActivityType[]) : Windows {
```
mattdailis: Thoughts on maybe calling this `duringAny`?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -325,6 +347,28 @@ export class Spans {
...
+   *
+   * @param activityType activity type to check
+   * @param expression function of an activity instance that returns a Constraint
+   * @constructor
+   */
+  public static ForEachActivity<A extends Gen.ActivityType>(
+      activityType: A,
+      expression: (instance: ActivityInstance<A>) => Spans,
+  ): Spans {
+    let alias = 'activity alias ' + Spans.__numGeneratedAliases;
```
mattdailis: Trying to justify to myself whether or not it's okay for `Spans.ForEachActivity` and `Constraint.ForEachActivity` to produce overlapping aliases.

I suppose since they're used for distinct use cases, it's probably fine. It would give me a bit of peace of mind if we prefix the alias, like `'spans activity alias ' + Spans__numGeneratedAliases`, but I won't insist if you think it's unnecessary.

mattdailis (commented):
Dropping a round of comments that are ready - I still have a few files to review

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
```diff
@@ -581,22 +583,18 @@ public void testNestedForEachActivity() {
         Map.of()
     );
 
-    final var violation = new Violation(List.of(), List.of(), new Windows(interval(4, 6, SECONDS), true));
-    final var result = new ForEachActivity(
+    final var spans = new Spans(interval(4, 6, SECONDS));
+    final var result = new ForEachActivitySpans(
```
mattdailis, line 620: Since we've decided to keep both ForEachActivityViolations and ForEachActivitySpans, perhaps we should add a new test, rather than edit this one?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
Mythicaeda: Or `duringOr`, depending on the order we merge this and #407.

JoelCourtney: my two cents is, I'd prefer to just make `during` take varargs

Also in the constraints eDSL the convention is to capitalize static functions (`Windows.During(...)`). I don't know if that's standard, but at least we're consistent.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/WindowsFromSpans.java
adrienmaillard: Yes I don't even remember why I've added those.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: 3 comments, 3 different ideas :) I'll go with `during` and the capitalization.

adrienmaillard: `Windows.During` and `Windows.DuringAny`. And varargs.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: Okay, will do

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
adrienmaillard: Yes.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -329,14 +336,25 @@ static JsonParser<WindowsFromSpans> windowsFromSpansF(final JsonParser<Expressio
...
+        .field("kind", literalP("ForEachActivitySpans"))
+        .field("activityType", stringP)
+        .field("alias", stringP)
+        .field("expression", spansExpressionP)
+        .map(
+            untuple((kind, actType, alias, expression) -> new ForEachActivitySpans(actType, alias, expression)),
+            $ -> tuple(Unit.UNIT, $.activityType, $.alias, $.expression));
+  }
+
+  static JsonParser<ForEachActivityViolations> forEachActivityWindowsF(final JsonParser<Expression<List<Violation>>> violationListExpressionP) {
```
JoelCourtney: I think this should be `forEachActivityViolationsF` or `forEachActivityConstraintsF` instead of `...WindowsF`.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ViolationsOfSpans.java
```diff
@@ -0,0 +1,62 @@
...
+    satisfiedWindows.iterator().forEachRemaining(element -> {
+      if(element.value().isPresent()) {
+        activityIds.add(element.value().get().id());
+      }
+      spansList.add(new Windows(element.interval(), true).not().select(bounds).intoSpans());
+    });
+    final var allIntervals = new ArrayList<Interval>();
+    spansList.iterator().forEachRemaining(a-> a.iterator().forEachRemaining(b -> allIntervals.add(b.interval())));
+    return List.of(new Violation(activityIds, new ArrayList<>(), allIntervals));
+  }
```
JoelCourtney: I thought since we reduced scope by leaving `Constraint.ForEachActivity`, we wouldn't be needing this anymore.

BTW this: `new Windows(element.interval(), true).not().select(bounds).intoSpans()` should always produce an exception because the `Windows(interval, bool)` constructor makes everything a gap outside of the interval, and `select(bounds)` will set gaps outside the bounds anyway; then `intoSpans` throws an exception because it can't be called on windows with gaps.

JoelCourtney: I'm seeing a bit of an issue coming up everywhere in the constraints DSL because of this. We eagerly restrict and truncate evaluation to the plan bounds, but this is problematic for operations that don't like gaps, or for `shiftBy` which can lose data by shifting it outside of the bounds (and back).

My solution is I'd like to remove all the `select` operations in the AST nodes except for the top-level constraint nodes, but that doesn't matter for _right_ now. Instead for now I think we just don't need this class.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -80,6 +80,26 @@ export class Windows {
...
+
+  /*
+  * Produces windows for each activity of the argument type present in the plan
+  *
+  * @param activityType the argument type
+  */
+  public static During(activityType: Gen.ActivityType) : Windows {
+    return Spans.ForEachActivity(activityType, (activity) => activity.span()).windows();
+  }
+
```
JoelCourtney: These can be combined into a single `During(...ActivityType)` function because `DuringAny` can be called with one argument.

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
```diff
@@ -38,9 +38,9 @@ export class ActivityInstance<A extends ActivityType> {
     this.parameters = 5;
   }
 
-  public window(): Windows {
-    return new Windows({
-      kind: AST.NodeKind.WindowsExpressionActivityWindow,
+  public span(): Spans {
+    return new Spans({
+      kind: AST.NodeKind.SpansExpressionActivitySpan,
       alias: this.__alias
```
JoelCourtney: The `ActivityInstance` placeholder code here can be completely removed, I must have forgotten to remove it when we did that rework of the types and made it so this class didn't need to be generated.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ViolationsOfSpans.java
adrienmaillard: I can remove it. 
Regarding your other comment : what nodes are top-level ? We can use any node and the contract of `evaluate` is that if we pass bounds, we get results only inside these bounds, no ?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: yes

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
adrienmaillard: but we use this `span()` in the `during()`.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
adrienmaillard: ok

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
adrienmaillard: Oh. In the mission model.

JoelCourtney: `ActivityInstance` is now in the fluent-api file. This is a duplicate class that isn't used.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ViolationsOfSpans.java
JoelCourtney: I'm proposing that we should remove that contract. The `ViolationsOfWindows` and `ForEachActivityViolations` nodes can restrict the bounds, because we know no more operations will be done. Similarly for scheduling, the consumer of the windows objects in the scheduler can decide whether or not to truncate the result to the plan.

But it doesn't make sense to truncate the result of `profile.equals(5)` to the plan bounds immediately, because if it was an external profile, there might be data outside of the plan that will be moved in by a `shiftBy`.

adrienmaillard: We can look at when we introduced `bounds`, I think it was for performance purposes, it could have a big impact if you think that some resources can be sometimes span several years, you do not want to evaluate the expression on the whole horizon. What you say is that the user should explicitly say when to restrict the evaluation horizon of the expression ?

adrienmaillard: Maybe you can make a ticket or a discussion on the slack and we can discuss about this outside of this PR ?

adrienmaillard: I've just deleted this file and pushed.

JoelCourtney: Ok, yes this is worthy of a github discussion.

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
```diff
@@ -1,7 +1,7 @@
 // The following are only to satisfy the type checker before any code is generated.
 // All of this will be overwritten before any constraints are evaluated.
 
-import { Windows, Discrete } from "./constraints-edsl-fluent-api.js";
+import { Spans, Discrete } from "./constraints-edsl-fluent-api.js";
```
JoelCourtney: I think the Spans import is unused

JoelCourtney (approved):
Awesome! Sorry we changed the scope halfway through.

mattdailis (approved):
Great work - I have one request (can be before or after merging): could you update the PR description to reflect the decisions made, separating the changes that are being merged from the ones that are now "future work"?

I forget why I added the `breaking change` label - if you don't think there are any breaking changes, we can remove that label.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ViolationsOfSpans.java
Twisol: FWIW, Joel and I had previously discussed a potential future direction for the `IntervalMap` data structure that would compute lazily / on-demand rather than eagerly (and potentially wastefully) computing across the whole timeline on every operation on a profile. I think that would fit quite snugly into the proposed GitHub discussion.

PR #404, opened 2022-10-26
[Aerie 2146] Include eDSL API docs on website

Mythicaeda:
* **Tickets addressed:** AERIE-2146
* **Review:** By commit  

## Description

This PR re-introduces building our eDSL docs as a part of our Sphinx deployment (see #391). As a side-effect, the generateDocs task now builds the docs into MarkDown.

Regarding CI:
If either of the eDSL docs changes structure, `buildExternalDocs.zsh` will need to be updated to reflect the new layout of the docs. This shouldn't lead to any backwards compatibility issues, however, as Sphinx runs the version of `buildExternalDocs.zsh` present on the branch/tag it has currently checked out.

## Verification

Generated both single-version and multiversion of the docs and ensured that internal page links were present, especially with the Scheduling docs, which cross-references the Constraints docs.

## Future work

We can hopefully use similar techniques to include the JavaDocs on the website.

### DIFF
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -6,6 +6,8 @@ on:
       - develop
     paths:
       - "docs/**"
+      - "merlin-server/constraints-dsl-compiler/src/**"
+      - "scheduler-server/scheduling-dsl-compiler/src/**"
   workflow_dispatch:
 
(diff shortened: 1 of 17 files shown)


camargo (approved):
LGTM thanks!

PR #406, opened 2022-10-26
[AERIE-2140] Scheduler workers

pcrosemurgy:
* **Tickets addressed:** AERIE-2140
* **Review:** By file

## Description

This PR factors a scheduler worker out of `scheduler-server` in a similar pattern as `merlin-server` and `merlin-worker` (#107). Changes here aim to keep refactoring changes ("2nd order refactoring changes") as low as possible in order to mirror `merlin-server`/`worker` as closely as possible.

Suggested to review by file due to the large number of changes.
The last 4 commits perform some general Gradle dependency housekeeping.

## Verification

- `e2e-tests` passing with 2 scheduler workers.

## Documentation

- [x] Documentation created with similar content as `merlin-worker`'s documentation.
- [x] Updated `TCP Port Requirements` documentation section.
- [x] Documented deployment environment variables for `scheduler-worker`.

## Future work

- `aerie-ui`: `src/types/scheduling.d.ts` must accomadate the new `pending` status.
- `scheduler-server` and `scheduler-worker` could use some general refactoring and possibly relocation of some classes to `scheduler-driver`.


scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
```diff
@@ -0,0 +1,64 @@
...
+      final var jsonValue = Json.createReader(new StringReader(payload)).readValue();
+      final var notificationPayload = postgresSchedulingRequestNotificationP
+          .parse(jsonValue)
+          .getSuccessOrThrow();
+      try {
+        notificationQueue.put(notificationPayload);
+      } catch (InterruptedException ex) {
+        // We are unable to handle this exception here so the thread interrupt flag is set to true
+        // so that code higher up the call stack can see that an interrupt was issued.
+        Thread.currentThread().interrupt();
```
pcrosemurgy, line 55: My understanding is that this allows the PGSQL driver (`jdbc`) to log that the listener thread was interrupted if it wants. I'm assuming the PGSQL driver logs some sort of message but it might just quietly chug along – both of which are honestly fine.

docker-compose.yml
```diff
@@ -159,6 +159,56 @@ services:
...
+    environment:
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      SCHEDULER_LOCAL_STORE: /usr/src/app/scheduler_file_store
+      SCHEDULER_WORKER_DB: "aerie_scheduler"
+      SCHEDULER_WORKER_DB_PASSWORD: "${AERIE_PASSWORD}"
+      SCHEDULER_WORKER_DB_PORT: 5432
+      SCHEDULER_WORKER_DB_SERVER: postgres
+      SCHEDULER_WORKER_DB_USER: "${AERIE_USERNAME}"
+      SCHEDULER_OUTPUT_MODE: UpdateInputPlanWithNewActivities
+      SCHEDULER_RULES_JAR: /usr/src/app/merlin_file_store/scheduler_rules.jar
```
JoelCourtney: Are you able to remove any of these variables from aerie_scheduler? Ok if not

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SpecificationLoadException.java
```diff
@@ -1,10 +1,9 @@
 package gov.nasa.jpl.aerie.scheduler.server.exceptions;
 
+import java.util.List;
```
mattdailis: The "implement schedule workers" commit seems to have a broad swath of non-functional changes in it (imports reordering, marking classes as final, adding/removing whitespace). Would it be possible to split them out, to make the functional changes easier to review?

(Interestingly enough, your imports autosorter disagrees with my imports autosorter... sounds like we need to standardize 😅)

I took note of some of the files that I found with no functional changes, just import re-ordering. I think the commit would be a lot less intimidating-looking without them:
- SpecificationLoadException.java
- GraphQLParsers.java
- ActivityTemplateJSONParser.java
- InvalidEntityException.java
- SchedulerParsers.java
- SerializedValueJsonParser.java
- StrictSerializedValueJsonParser.java
- MerlinActivityInstance.java
- MerlinPlan.java
- PlanMetadata.java
- SchedulingCompilationError.java
- SchedulingDSL.java
- Specification.java
- CancelSchedulingRequestAction.java
- CreateRequestAction.java
- DeleteRequestAction.java
- GetCreatedActivitiesAction.java
- GetGoalSatisfaction.java
- GetRequestAction.java
- GetSatisfyingActivitiesAction.java
- GetSpecificationAction.java
- GetSpecificationConditionsAction.java
- GetSpecificationGoalsAction.java
- GoalBuilder.java
- InsertCreatedActivitiesAction.java
- InsertGoalSatisfactionAction.java
- InsertSatisfyingActivitiesAction.java
- PostgresParsers.java
- PostgresSpecificationRepository.java
- PreparedStatements.java
- SetRequestStateAction.java
- SpecificationRecord.java
- GraphQLMerlinService.java
- MissionModelService.java
- PlanService.java
- ThreadedSchedulerAgent.java
- GraphQLParsersTest.java

pcrosemurgy: Wow, thanks for taking the time to compile this! Just pushed a separate commit for this that precedes the `Implement scheduler workers` commit.

pcrosemurgy:
@Mythicaeda Any chance you know what may be causing this doc `make` failure? https://github.com/NASA-AMMOS/aerie/actions/runs/3347770365/jobs/5546115927

docs/buildExternalDocs.bash
```diff
@@ -12,7 +12,7 @@ cp -a ../merlin-server/constraints-dsl-compiler/build/docs/. ./source/constraint
 rm -f ./source/constraints-edsl-api/.nojekyll
 
 mkdir -p ./source/scheduling-edsl-api
-cp -a ../scheduler-server/scheduling-dsl-compiler/build/docs/. ./source/scheduling-edsl-api
+cp -a ../scheduler-worker/scheduling-dsl-compiler/build/docs/. ./source/scheduling-edsl-api
```
Mythicaeda, line 15: `docs-pages.yaml` and `docs-pr.yaml` need this same change in the line `"scheduler-server/scheduling-dsl-compiler/src/**"`

docs/source/deployment/single-host-deployment.md
```diff
@@ -21,7 +21,7 @@ This product is using Docker containers to run the application. The Docker conta
 
 Each Aerie service is configured with environment variables. A description of those variables is found in the [Environment Variable Documentation](https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/Environment.md)
 
-Of note, the aerie-merlin, aerie-scheduler, and aerie_merlin_worker(s) can be provided additional JVM arguments as environment variables. For example one may choose to configure the JVM allocated heap size. On must provide any desired JVM flags to the `JAVA_OPTS` environment variable for the container being configured.
+Of note, the aerie-merlin, aerie-scheduler, aerie_merlin_worker(s), and aerie_merlin_scheduler(s) can be provided additional JVM arguments as environment variables. For example one may choose to configure the JVM allocated heap size. On must provide any desired JVM flags to the `JAVA_OPTS` environment variable for the container being configured.
```
JoelCourtney: aerie_merlin_scheduler?

docker-compose.yml
```diff
@@ -159,6 +159,56 @@ services:
...
+      JAVA_OPTS: >
+        -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
+        -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
+        -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=INFO
+        -Dorg.slf4j.simpleLogger.logFile=System.err
+    ports: ["5009:5005", "27189:8080"]
```
JoelCourtney, line 178: deploy/docker compose says 27188

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
```diff
@@ -40,10 +31,6 @@ public record SchedulerBindings(
     ScheduleAction scheduleAction,
     GenerateSchedulingLibAction generateSchedulingLibAction
 ) implements Plugin {
-  public SchedulerBindings {
-    Objects.requireNonNull(schedulerService, "schedulerService must be non-null");
-    Objects.requireNonNull(scheduleAction, "scheduleAction must be non-null");
-  }
```
JoelCourtney: why remove this requirement?

(I may have written this comment twice, I can't remember. My bad if you get this twice)

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/mocks/InMemoryResultsCellRepository.java
```diff
@@ -1,19 +1,24 @@
...
   {
     throw new UnsupportedOperationException(); // TODO stubbed method must be implemented
   }
 
+  @Override
+  public Optional<ResultsProtocol.OwnerRole> claim(SpecificationId specificationId)
+  {
+    throw new UnsupportedOperationException(); // TODO stubbed method must be implemented
+  }
+
```
JoelCourtney, line 21: Does this have to be implemented for this PR or is this future work?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -44,12 +47,43 @@ public ResultsProtocol.OwnerRole allocate(final SpecificationId specificationId)
...
 
+  @Override
+  public Optional<ResultsProtocol.OwnerRole> claim(final SpecificationId specificationId)
+  {
+    try (
+        final var connection = this.dataSource.getConnection();
+        final var claimSimulationAction = new ClaimRequestAction(connection)
+    ) {
+      claimSimulationAction.apply(specificationId.id());
+      logger.info("Claimed scheduling request with specification id {}", specificationId);
```
JoelCourtney: please either move this log until after we know the claim is successful or change it to "Attempted to claim..."

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/ClaimRequestAction.java
```diff
@@ -0,0 +1,38 @@
...
+  public ClaimRequestAction(final Connection connection) throws SQLException {
+    this.statement = connection.prepareStatement(sql);
+  }
+
+  public void apply(final long specificationId) throws SQLException, UnclaimableRequestException {
+    this.statement.setLong(1, specificationId);
+
+    final var count = this.statement.executeUpdate();
+    if (count < 1) {
+      throw new UnclaimableRequestException(specificationId);
```
JoelCourtney, line 27: Since this is an intentional race condition, I'd prefer not to include `throw`s on the happy path. Can you change this to return a boolean instead of throw?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ConstraintsTypescriptCodeGenerationService.java
```diff
@@ -0,0 +1,37 @@
+package gov.nasa.jpl.aerie.scheduler.server.services;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+import gov.nasa.jpl.aerie.constraints.TypescriptCodeGenerationService;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
+
+public final class ConstraintsTypescriptCodeGenerationService
```
JoelCourtney: I think this name might be misleading because it doesn't actually generate the code.

JoelCourtney:
Can you move scheduling-dsl-compiler/src/libs/README out of src please?

scheduler-worker/build.gradle
```diff
@@ -0,0 +1,136 @@
...
+task deleteConstraintsTypescript(type: Delete) {
+  delete project.projectDir.toPath().resolve('scheduling-dsl-compiler/src/libs/constraints')
+}
+
+task copyConstraintsTypescript(type: Copy) {
+  dependsOn deleteConstraintsTypescript
+  dependsOn ':merlin-server:assembleConstraintsDSLCompiler'
+  from rootDir.toPath().resolve('merlin-server/constraints-dsl-compiler/src/libs')
+  into project.projectDir.toPath().resolve('scheduling-dsl-compiler/src/libs/constraints')
+}
```
JoelCourtney, line 32: I think we ended up with two copies of the constraints API floating around, even before this PR. @mattdailis do you know why we copy into `scheduling-dsl-compiler/src/libs/constraints` but also have the files in `scheduling-dsl-compiler/src/libs`?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulerWorkerAppDriver.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+    }
+  }
+
+  private static String getEnv(final String key, final String fallback){
+    final var env = System.getenv(key);
+    return env == null ? fallback : env;
+  }
+
+  private static WorkerAppConfiguration loadConfiguration() {
```
JoelCourtney, line 107: Can you rename `WorkerAppConfiguration` in merlin and scheduler to `MerlinWorkerAppConfiguration` and `SchedulerWorkerAppConfiguration`? I know it technically doesn't matter, I just want it to be a little clearer

docs/source/deployment/single-host-deployment.md
pcrosemurgy: Oops! Good catch. Fixed.

docker-compose.yml
pcrosemurgy: Good catch. Since the deployment stack only has 1 merlin worker and 1 scheduler server theres a small gap in the sequential port numbering in `deployment`. When I wrote this I figured we'd not introduce a gap in the port numbering, but I think having this stack's ports consistent with our development stack makes things far less confusing so I've changed fixed this :)

`deployment/`'s port is now 27189 for the scheduler worker 1.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
pcrosemurgy: Adding `requireNonNull()` guards to `record` classes kind of depends on personal style, but since this PR should not attempt to make "2nd order refactoring changes" (refactoring on top of refactoring scheduler server) I'll revert this change.

In general I think these `requireNonNull()`s on `record` definitions mostly add clutter.

pcrosemurgy: Reverted this change. Although I did take out the optional `"schedulerService must be non-null"` messages since they are redundant when these `requireNonNull` asserts fail.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/mocks/InMemoryResultsCellRepository.java
pcrosemurgy: Future work :) On the Merlin side of things this in-memory results cell repo. exists to help with testing without a true PG DB up and running. We can eventually support this for the scheduler as well.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
pcrosemurgy: Ahh true, there are a few lines below this that could cause the whole claim to fail. I've moved this to right before the final `return` in this method.

Thanks!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ConstraintsTypescriptCodeGenerationService.java
pcrosemurgy: 100% agree. It's kind of a utility class. Any preference on the name? Maybe `ConstraintsTypescriptCodeGenerationHelper`/`Util`?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/ClaimRequestAction.java
pcrosemurgy: I completely agree with this change but I'd like to do this as part of a follow-up PR that makes this change in both `merlin-worker` and `scheduler-worker`. I've avoided making refactoring changes to `merlin-worker` in this PR and would like to keep both workers in lock-step w.r.t. changes as well as we can. :)

pcrosemurgy: I'll note this in the future work section and will make sure to capture this in a ticket.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulerWorkerAppDriver.java
pcrosemurgy: While this is a minor change I think this would be best for a follow-up PR that makes this change in both `merlin-worker` and `scheduler-worker`.

We likely need to have a broader discussion about if we should / how much we should qualify subproject-specific names within subprojects that mirror other subprojects' architecture (for example `*-server`s and `*-worker`s having similar internal names).

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
JoelCourtney: Do records require non null automatically?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulerWorkerAppDriver.java
pcrosemurgy: I'm also a little weary about qualifying things to this degree, mostly since I'd like naming conventions to be as consistent as possible.

For example, if we make this change then shouldn't we also change `PostgresResultsCellRepository` within both subprojects to `MerlinPostgresResultsCellRepository` and `SchedulerPostgresResultsCellRepository`? Maybe the answer is "yes", but it seems like a lot of clutter to add these qualifiers to every class that has the same name across multiple subprojects, ya know?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ConstraintsTypescriptCodeGenerationService.java
JoelCourtney: Helper sounds good

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
pcrosemurgy: They don't. So I should really say it's really a matter of preference, not style.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/ConstraintsTypescriptCodeGenerationService.java
pcrosemurgy: Thanks! Done.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
pcrosemurgy: This is one area I kind of sway back & forth on. I think @Twisol also felt a bit ambivalent here too

pcrosemurgy:
>Can you move scheduling-dsl-compiler/src/libs/README out of src please?

Done :)

scheduler-worker/build.gradle
mattdailis: Here's where I added that: https://github.com/NASA-AMMOS/aerie/pull/179#issuecomment-1138060052 

We give the user code runner a list of filenames - I'm not sure that we can provide it filenames in any sort of namespaced way (i.e. imitating directory structure). However, during development, I think it's helpful to keep the copied constraints files in their own directory so they don't get mixed up with scheduling eDSL files.

So, only the files in the `scheduling-dsl-compiler/src/libs/constraints` matter at runtime.

The imports in `scheduling-edsl-fluent-api.ts` are written as though the constraints files were in the same directory (since they will appear to be at runtime). However, at build time, the typescript compiler complains about missing imports, so the similarly named files in `scheduling-dsl-compiler/src/libs/` simply import and re-export the copied files in `constraints`.

Happy to entertain alternative solutions - this is just what I came up with at the time to satisfy all of my constraints (no pun intended).

docker-compose.yml
pcrosemurgy: Good catch. I've removed these from the scheduler in the latest commit.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SpecificationLoadException.java
mattdailis: Thanks for splitting it out - you've unblocked my ability to review this PR, so I'll try to get on that now :+1: 

I do wonder whether it would make sense to remove those changes from this PR entirely. So far, the only policy I'm aware of is that we like to avoid `.*` imports (which it looks like your import auto formatter inserted). We have talked about maybe making a style guide - if you feel strongly about import order, maybe you can help start drafting a style guide so we can start to standardize.

pcrosemurgy: Ahh I see. My intent here was to remove any unused imports, any stylistic changes were a side-effect, no strong opinion on import style here :)

pcrosemurgy: But looks like there are only a few files that had unused imports lol (like [GetRequestAction.java](https://github.com/NASA-AMMOS/aerie/pull/406/commits/c240ddccfb88c9f458516af173a380d0a146f504#diff-b19bd10556a8ec67369d66a5a8443a1f37ac6b7a8d3aa58d458f2b2e268a5d2c)). I'm happy to drop this commit altogether since it's 90% stylistic side-effects and 10% removing unused imports.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -265,9 +314,9 @@ public void cancel() {
     @Override
     public boolean isCanceled() {
       try (final var connection = dataSource.getConnection()) {
-        return getRequest(connection, specId, specRevision).canceled();
-      } catch (final NoSuchRequestException ex) {
-        throw new Error("Scheduling request no longer exists");
+        return getRequest(connection, specId, specRevision)
+            .map(RequestRecord::canceled)
+            .orElseThrow(() -> new Error("Request corrupted"));
```
mattdailis: I like the use of `Optional` here instead of exceptions - could you help me understand the reasoning behind the changed error message?

Before: "Scheduling request no longer exists"
After: "Request corrupted"

pcrosemurgy: Weird, I've grepped around trying to find where the `"Request corrupted"` string came from and can't find it. I changed these strings back to their original form.

deployment/docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
         -Dorg.slf4j.simpleLogger.logFile=System.err
       UNTRUE_PLAN_START: "2000-01-01T11:58:55.816Z"
     image: "${REPOSITORY_DOCKER_URL}/aerie-merlin-worker:${DOCKER_TAG}"
+    ports: ["27187:8080"]
```
mattdailis, line 77: ~~This ports line seems spurious~~

Nevermind, good catch! As a reviewer, this caught my eye as an "unrelated change" - I won't request it be split out into its own commit now, but just a note for the future.

mattdailis (approved):
Approving based on visual inspection. If I have time I'll also try to spin up a local test.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SpecificationLoadException.java
pcrosemurgy: Commit dropped now :)

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
Twisol: Yeah, I really wish records automatically had non-null checks. I'd like to say I add them depending on how far the nulls could potentially propagate... but usually it's a matter of "do I really want to have all this boilerplate?".

I'd say non-null checks are ideal, but not worth it for certain kinds of records. This one is the type that would fail instantly on attempting to interact with the service, so we don't necessarily need to crash out much sooner than that.

scheduler-worker/build.gradle
JoelCourtney: Is this due to a limitation of the aerie user code runner, not letting us put the files in "virtual" directories?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulerWorkerAppDriver.java
JoelCourtney: Alright, I'm ok with it

JoelCourtney (approved):
dank, thank you!

scheduler-worker/build.gradle
JoelCourtney: @pcrosemurgy this is not a blocker for merging

pcrosemurgy: Got it, marking this as "resolved" but it sounds like we should continue this conversation on this PR after it's closed, through Slack, or through GH discussions. :)

PR #407, opened 2022-10-26
[AERIE-2183] Revert name changes to boolean operations

JoelCourtney:
* **Tickets addressed:** AERIE-2183
* **Review:** By file  

**Changelog entry:** Rename `Windows.Any` to `Windows.Or`, `Windows.All` to `Windows.And`, and `windows.invert` to `windows.not` in Constraints eDSL

## Description
- The main change is to revert my change to the constraints eDSL 4 months ago: renaming All -> And, Any -> Or, Invert -> Not. This is because with the introduction of gaps, I think the most helpful mental model is that of a boolean profile. So we should have boolean suggestive names. (I forget who disagreed with me four months ago, but you were right :).
- I also created new methods in the DSL that delegate to the static And/Or functions. i.e. `left.and(right) -> Windows.And(left, right)`

## Verification
The new methods have a test in the CompilationServiceTests.

## Documentation
I brought all of the DSL doc-comments up to date, and added a few extra.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -261,30 +261,30 @@ public final class ConstraintParsers {
               untuple((kind, left, right) -> new GreaterThanOrEqual(left, right)),
               $ -> tuple(Unit.UNIT, $.left, $.right));
 
-  static JsonParser<All> allF(final JsonParser<Expression<Windows>> windowsExpressionP) {
+  static JsonParser<And> andF(final JsonParser<Expression<Windows>> windowsExpressionP) {
     return productP
-        .field("kind", literalP("WindowsExpressionAll"))
+        .field("kind", literalP("WindowsExpressionAnd"))
         .field("expressions", listP(windowsExpressionP))
         .map(
-            untuple((kind, expressions) -> new All(expressions)),
+            untuple((kind, expressions) -> new And(expressions)),
             $ -> tuple(Unit.UNIT, $.expressions));
   }
 
-  static JsonParser<Any> anyF(final JsonParser<Expression<Windows>> windowsExpressionP) {
+  static JsonParser<Or> orF(final JsonParser<Expression<Windows>> windowsExpressionP) {
     return productP
-        .field("kind", literalP("WindowsExpressionAny"))
+        .field("kind", literalP("WindowsExpressionOr"))
         .field("expressions", listP(windowsExpressionP))
         .map(
-            untuple((kind, expressions) -> new Any(expressions)),
+            untuple((kind, expressions) -> new Or(expressions)),
             $ -> tuple(Unit.UNIT, $.expressions));
   }
 
-  static JsonParser<Invert> invertF(final JsonParser<Expression<Windows>> windowsExpressionP) {
+  static JsonParser<Not> notF(final JsonParser<Expression<Windows>> windowsExpressionP) {
     return productP
-        .field("kind", literalP("WindowsExpressionInvert"))
+        .field("kind", literalP("WindowsExpressionNot"))
         .field("expression", windowsExpressionP)
         .map(
-            untuple((kind, expr) -> new Invert(expr)),
+            untuple((kind, expr) -> new Not(expr)),
             $ -> tuple(Unit.UNIT, $.expression));
   }
 
@@ -365,9 +365,9 @@ public final class ConstraintParsers {
         equalF(discreteProfileExprP),
         notEqualF(linearProfileExprP),
         notEqualF(discreteProfileExprP),
-        allF(selfP),
-        anyF(selfP),
-        invertF(selfP),
+        andF(selfP),
+        orF(selfP),
+        notF(selfP),
         shiftByF(selfP),
         startsF(selfP),
         endsF(selfP),
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -36,7 +36,7 @@ export class Constraint {
         activityType1,
         activity1 => Constraint.ForEachActivity(
             activityType2,
-            activity2 => Windows.All(activity1.window(), activity2.window()).invert()
+            activity2 => Windows.And(activity1.window(), activity2.window()).not()
         )
     )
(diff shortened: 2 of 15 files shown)


Twisol:
> [...] I think the most helpful mental model is that of a boolean profile. So we should have boolean suggestive names. (I forget who disagreed with me four months ago, but you were right :).

👋 😉

Twisol (approved):
I'm a fan!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -261,30 +261,30 @@ static JsonParser<ShorterThan> shorterThanP(JsonParser<Expression<Windows>> wind
               untuple((kind, left, right) -> new GreaterThanOrEqual(left, right)),
               $ -> tuple(Unit.UNIT, $.left, $.right));
 
-  static JsonParser<All> allF(final JsonParser<Expression<Windows>> windowsExpressionP) {
+  static JsonParser<And> allF(final JsonParser<Expression<Windows>> windowsExpressionP) {
```
Mythicaeda: Regarding `allF`, `anyF`, and `invertF`: We ought to consider renaming these so that they remain consistent (especially `invertF`, which was renamed from `notF` during the `not->invert` conversion)

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -74,44 +81,41 @@ export class Windows {
...
   }
 
   /**
-   * Produce a window when any argument produces a window.
+   * Performs the boolean Or operation on any number of Windows.
    *
-   * Performs the union of the argument windows.
-   *
-   * @param windows one or more windows expressions
+   * @param windows any number of windows expressions
```
Mythicaeda, line 98: Can we call this with 0 windows expressions?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -122,6 +122,14 @@ export class Windows {
   /**
    * Invert all the windows produced this.
    */
```
Mythicaeda, line 128: I think this comment is supposed to be attached to the `not` method below.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -122,6 +122,14 @@ export class Windows {
...
+  public and(...windows: Windows[]): Windows {
+    return Windows.And(this, ...windows);
+  }
+
+  public or(...windows: Windows[]): Windows {
+    return Windows.Or(this, ...windows);
+  }
+   */
```
Mythicaeda, line 142: Extra */

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -697,6 +705,10 @@ declare global {
     /**
      * Invert all the windows produced this.
```
Mythicaeda: Same as above: I think this is supposed to be attached to `public not() Windows;` below.

Mythicaeda (commented):
Nice work! Just a couple of minor things, mainly docs-related.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
JoelCourtney: ah good catch

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Yes. `Windows.Or()` returns false and `Windows.And()` return true. Which is kinda vacuously correct. None of the arguments to Or were true, and all of the arguments to And were true. 🤷

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: I updated all the doc comments in the next commit, I think you must be viewing commit 2

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: same as above, this is fixed in commit 3

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: same as above, this is fixed in commit 3

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
Mythicaeda: Ah, I was

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
Mythicaeda: Okay, just to confirm, when all the arguments to `Or` are false, `Or` returns false, and when all of the arguments to `And` are false, `And` returns true?

JoelCourtney: whoops, typo: `Windows.And()` returns true because all of the arguments were _true_

Mythicaeda: Gotcha. And calling `Or` without arguments means all arguments are false, while calling `And` with no arguments means all arguments are true?

JoelCourtney: Yeah. `Or` returns true if any of the arguments are true, and `And` returns false if any of the arguments are false. Even with no arguments this is still technically correct, which is the best kind of correct

Mythicaeda (approved):

PR #409, opened 2022-10-31
style: format command expansion server

camargo:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

* Run Prettier format over command expansion server
* Change 'reformat' script to 'format'
* Alphabetize .prettierrc
* Regenerated test snapshot 
* No functional difference

## Verification

N/A

## Documentation

N/A

## Future work

N/A

### DIFF
--- a/command-expansion-server/test/__snapshots__/command-types.spec.ts.snap
+++ b/command-expansion-server/test/__snapshots__/command-types.spec.ts.snap
@@ -245,7 +245,7 @@ export class Command<
 
   private static argumentsToString<A extends ArgType[] | { [argName: string]: any } = [] | {}>(args: A): string {
     if (Array.isArray(args)) {
-      const argStrings = args.map((arg) => {
+      const argStrings = args.map(arg => {
         if (typeof arg === 'string') {
           return \`'\${arg}'\`;
         }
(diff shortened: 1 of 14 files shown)


mattdailis:
Thoughts on starting a `.git-blame-ignore-revs` file? https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

EDIT: Now that I think about it, we could do this retroactively, so we don't need to decide now

camargo:
@mattdailis No preference. I don't use blame often personally. If you want to add this in a separate commit please go for it.

skovati (approved):
Looking quite pretty 👍

command-expansion-server/test/__snapshots__/command-types.spec.ts.snap
```diff
@@ -299,20 +297,16 @@ export class Sequence {
...
-  return (
-\`export default () =>
+    return \`export default () =>
   Sequence.new({
     seqId: '\${this.seqId}',
     metadata: \${JSON.stringify(this.metadata)},
     commands: [\${commandsString}
     ],
-  });\`
-  );
```
Mythicaeda, line 315: Why was this return wrapped in `();`?

mattdailis:
I'll add a commit, one sec :+1:

mattdailis (approved):
:+1: Thanks, I do use git blame quite heavily

Mythicaeda (approved):
LGTM!

command-expansion-server/test/__snapshots__/command-types.spec.ts.snap
camargo: So this is the generated file but the [file that generates this code](https://github.com/NASA-AMMOS/aerie/pull/409/files#diff-224d12ae69810ebad6855e95127d0a73a31d2653d5207744b5f8f68370bc0727L304) had parens in the `return` statement. They are not needed tho so Prettier removed them.

PR #416, opened 2022-11-02
[FIX] Use property for publishing version string

pcrosemurgy:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

The previous solution using `rootProject.ext.versionForPublishing` failed when building Aerie locally from a separate Gradle project, likely due to `rootProject` taking on a different value/meaning in that context.

Using a global property here we're able to specify the publishing version in the top-level build and guarantee that any uses of it are localized to subprojects of our top-level build.

## Verification
- Aerie is compiling and passing tests.
- Running `gradle assemble -PuseLocalAerie=true -PlocalAeriePath=../../aerie` from a cloned `aerie-simple-mission-model` compiles the mission model without issues.

## Documentation
None.

## Future work

We'll want to apply this patch to the https://github.jpl.nasa.gov/Aerie/aerie-simple-mission-model repo.:
```diff
diff --git a/build.gradle b/build.gradle
index f47411b..3695094 100644
--- a/build.gradle
+++ b/build.gradle
@@ -27,19 +27,19 @@ jar {
 }

 dependencies {
-//  if (useLocalAerie == "true") {
-//    compileOnly project(':contrib')
-//    compileOnly project(':merlin-framework')
-//    implementation project(':parsing-utilities')
-//    annotationProcessor project(':merlin-framework-processor')
-//
-//    testImplementation project(':contrib')
-//    testImplementation project(':merlin-framework')
-//    testImplementation project(':merlin-sdk')
-//    testImplementation project(':merlin-driver')
-//    testImplementation project(':merlin-framework-junit')
-//  }
-//  else {
+  if (useLocalAerie == "true") {
+    compileOnly project(':contrib')
+    compileOnly project(':merlin-framework')
+    implementation project(':parsing-utilities')
+    annotationProcessor project(':merlin-framework-processor')
+
+    testImplementation project(':contrib')
+    testImplementation project(':merlin-framework')
+    testImplementation project(':merlin-sdk')
+    testImplementation project(':merlin-driver')
+    testImplementation project(':merlin-framework-junit')
+  }
+  else {
     implementation 'gov.nasa.jpl.aerie:contrib:0.13.0-SNAPSHOT-bad4548'
     implementation 'gov.nasa.jpl.aerie:merlin-framework:0.13.0-SNAPSHOT-bad4548'
     implementation 'gov.nasa.jpl.aerie:parsing-utilities:0.13.0-SNAPSHOT-bad4548'
@@ -50,8 +50,7 @@ dependencies {
     testImplementation 'gov.nasa.jpl.aerie:merlin-framework-junit:0.13.0-SNAPSHOT-bad4548'

     implementation 'gov.nasa.jpl.aerie:merlin-sdk:0.13.0-SNAPSHOT-bad4548'
-
-//  }
+  }
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,14 +49,15 @@ configure(subprojects.findAll { it.projectDir.toPath().resolve('sql').toFile().e
   archiveDeployment.dependsOn distributeSql
 }
 
-project.ext.versionForPublishing = {
+// Set publishing version property globally for this project
+setProperty('publishing.version', {
   if (findProperty("version.isRelease").toBoolean()) {
     return "${findProperty("version.number")}"
   } else {
     def hash = 'git rev-parse --short HEAD'.execute().text.trim()
     return "${findProperty("version.number")}-SNAPSHOT-$hash"
   }
-}()
+}())
 
 subprojects {
   apply plugin: 'com.github.ben-manes.versions'
--- a/contrib/build.gradle
+++ b/contrib/build.gradle
@@ -34,7 +34,7 @@ dependencies {
 publishing {
   publications {
     library(MavenPublication) {
-      version = rootProject.ext.versionForPublishing
+      version = findProperty('publishing.version')
       from components.java
     }
   }
--- a/examples/banananation/build.gradle
+++ b/examples/banananation/build.gradle
@@ -42,7 +42,7 @@ javadoc.options.tags = [ "contact", "subsystem" ]
 publishing {
   publications {
     library(MavenPublication) {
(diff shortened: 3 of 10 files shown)


mattdailis (approved):
I'm not too familiar with the nuance of properties, but I do think this accomplishes the intent we're going for :+1:

Really it's just a way to re-use this version logic. I had originally pasted the whole if statement into each subproject, but then decided it would be better to define it in one place.

PR #419, opened 2022-11-03
[AERIE-2205] Rename condition table to "constraint"

JoelCourtney:
* **Tickets addressed:** AERIE-2205
* **Review:** By commit  

**Changelog entry:** Rename condition table to constraint. GraphQL queries that interact with the `condition` table will need to be refactored.

## Description
The constraints in the database are still stored in a table named `condition`, which is a relic name from the ancient past (probably like the cretaceous, I donno). Now with scheduling conditions, this is extra confusing. So I changed the table name to `"constraint"`. Unfortunately `constraint` is a keyword, so we have to put quotes around it, but other than that this is much less confusing.

## Verification
I manually created and ran a constraint in the UI just for sanity.

## Documentation
This allows the graphql documentation to call constraints by what they actually are :)

## Future work
I'm opening a sister PR in the UI ([here](https://github.com/NASA-AMMOS/aerie-ui/pull/212))

### DIFF
--- a/e2e-tests/src/utilities/gql.ts
+++ b/e2e-tests/src/utilities/gql.ts
@@ -142,7 +142,7 @@ const gql = {
           startOffset: start_offset
           type
         }
-        constraints: conditions {
+        constraints {
           definition
           description
           id
@@ -158,7 +158,7 @@ const gql = {
             name
             parameters
           }
-          constraints: conditions {
+          constraints {
             definition
             description
             id
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_constraint.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_constraint.yaml
@@ -1,3 +1,3 @@
 table:
-  name: condition
+  name: constraint
   schema: public
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_mission_model.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_mission_model.yaml
@@ -20,12 +20,12 @@ array_relationships:
       table:
         name: activity_type
         schema: public
-- name: conditions
+- name: constraints
   using:
     foreign_key_constraint_on:
       column: model_id
       table:
-        name: condition
+        name: constraint
         schema: public
 - name: plans
   using:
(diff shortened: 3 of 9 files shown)


camargo:
IIRC I think the only reason it was called "condition" was because of the keyword issue. Maybe @Twisol can comment more here. I'm okay with this change as long as we don't think the keyword clash will cause other issues in Postgres.

JoelCourtney:
> as long as we don't think the keyword clash will cause other issues in Postgres.

[Somebody on stack overflow](https://stackoverflow.com/questions/7651417/escaping-keyword-like-column-names-in-postgres) gave a link to [this postgres docs page](https://www.postgresql.org/docs/current/sql-syntax-lexical.html), which says:

> There is a second kind of identifier: the delimited identifier or quoted identifier. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). A delimited identifier is always an identifier, never a key word. So "select" could be used to refer to a column or table named “select”, whereas an unquoted select would be taken as a key word and would therefore provoke a parse error when used where a table or column name is expected.

So it looks like this is a pretty standard way to use keywords as identifiers. Hasura handles it even better, because the generated graphql queries/mutations don't care about this keyword nonsense.

camargo (approved):

Twisol:
> IIRC I think the only reason it was called "condition" was because of the keyword issue. Maybe @Twisol can comment more here. I'm okay with this change as long as we don't think the keyword clash will cause other issues in Postgres.

Haha. Yes, hi. The keyword issue was certainly part of it, but also, I saw a semantic difference between a *condition* (which is simply a derived boolean resource profile) and a *constraint* (in which we pass or fail the plan based on whether the constraint is met).

I don't think this was ever a naming position held commonly on the team, and I'm not strongly bound to it. But I also hope it's not seen as retroactively wrongheaded 😄

e2e-tests/src/utilities/gql.ts
```diff
@@ -129,7 +129,7 @@ const gql = {
           startOffset: start_offset
           type
         }
-        constraints: conditions {
+        constraints: constraints {
```
camargo: You can remove the alias here and just do `constraints {`

mattdailis (approved):

PR #423, opened 2022-11-07
[REFACTOR] Avoid computing samples eagerly in SimulationResults

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This PR removes `resourceSamples` as a field on `SimulationResults`. Prior to this PR, this field was derived from `realProfiles` and `discreteProfiles` via the `takeSamples` method - upon calling the `SimulationResults` constructor, `takeSamples` would be called eagerly.

In typical usage, `resourceSamples` is unused - the simulation results are written to the database using the `realProfiles` and `discreteProfiles` directly. The only time it is used (outside of testing) is via the explicit `resourceSamples` hasura action.

This PR moves computation of resource samples out of the `SimulationResults` constructor, so that they can be computed when requested, but not by default.

I have not done any performance measurements to see whether or not this change is significant - nevertheless, I believe avoiding needless computations is generally a good thing.

## Verification

I spun up a local Aerie deployment and checked that running simulations and viewing their results worked, and I also called the `resourceSamples` endpoint directly.

## Documentation

No functional behavior has changed, so no documentation needs to be updated.

## Future work

No future work

### DIFF
--- a/examples/foo-missionmodel/src/test/resources/gov/nasa/jpl/aerie/foomissionmodel/plan.json
+++ b/examples/foo-missionmodel/src/test/resources/gov/nasa/jpl/aerie/foomissionmodel/plan.json
@@ -2,24 +2,29 @@
     "type": "foo",
     "arguments":
       { "x": 1,
-        "y": "test_1" } },
+        "y": "test_1",
+        "z": 4 } },
   { "defer": 50000,
     "type": "foo",
     "arguments":
       { "x": 2,
-        "y": "spawn" } },
+        "y": "spawn",
+        "z": 4 } },
   { "defer": 50000,
     "type": "foo",
     "arguments":
       { "x": 2,
-        "y": "test" } },
+        "y": "test",
+        "z": 4 } },
   { "defer": 100000,
     "type": "foo",
     "arguments":
       { "x": 2,
-        "y": "test_2" } },
+        "y": "test_2",
+        "z": 4 } },
   { "defer": 150000,
     "type": "foo",
     "arguments":
       { "x": 3,
-        "y": "test_3" } } ]
+        "y": "test_3",
+        "z": 4 } } ]
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
@@ -3,14 +3,13 @@ package gov.nasa.jpl.aerie.merlin.server.services;
 import gov.nasa.jpl.aerie.constraints.InputMismatchException;
 import gov.nasa.jpl.aerie.constraints.model.ActivityInstance;
 import gov.nasa.jpl.aerie.constraints.model.DiscreteProfile;
-import gov.nasa.jpl.aerie.constraints.model.DiscreteProfilePiece;
 import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
 import gov.nasa.jpl.aerie.constraints.model.LinearProfile;
-import gov.nasa.jpl.aerie.constraints.model.LinearProfilePiece;
 import gov.nasa.jpl.aerie.constraints.model.Violation;
 import gov.nasa.jpl.aerie.constraints.time.Interval;
 import gov.nasa.jpl.aerie.constraints.tree.Expression;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationFailure;
+import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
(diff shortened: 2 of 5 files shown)


camargo (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -78,7 +77,7 @@ public Map<String, List<Pair<Duration, SerializedValue>>> getResourceSamples(fin
     final var revisionData = this.planService.getPlanRevisionData(planId);
     return this.simulationService
         .get(planId, revisionData)
-        .map(r -> r.resourceSamples)
+        .map(SimulationResults::resourceSamples)
```
Mythicaeda: Could you clarify where `resourceSamples` gets the `SimulationResults` parameter it needs here?

Mythicaeda (commented):
Looks fine, although seeing it appear in the mission models makes me wonder how often this calculation will be called per SimulationResults object. Would it be worth it to instead make `resourceSamples` a private field that is not assigned at construction, and have it accessed through a getter that assigns the output of `takeSamples` into it if `resourceSamples` is null? Then we only run `takeSamples` at most once per SimulationResults object.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
mattdailis: Sure thing - `map` provides the simulation results as an argument - the code below is equivalent: 

```suggestion
        .map(results -> SimulationResults.resourceSamples(results))
```

Mythicaeda: Thanks!

mattdailis:
> Looks fine, although seeing it appear in the mission models makes me wonder how often this calculation will be called per SimulationResults object. Would it be worth it to instead make `resourceSamples` a private field that is not assigned at construction, and have it accessed through a getter that assigns the output of `takeSamples` into it if `resourceSamples` is null? Then we only run `takeSamples` at most once per SimulationResults object.

If we expected this computation to happen multiple times, absolutely, I think caching it in a field would be a valid approach. 💯 

As it stands, we've moved away from this method (it provides strictly less information than `realProfiles` and `discreteProfiles` do, while using more memory) - so we don't expect this method to get more use in the future. If I had my druthers, we'd remove this method entirely, but we're keeping the `resourceSamples` hasura action in the API as a convenience for clients that don't want to work with continuous profiles.

Mythicaeda:
> As it stands, we've moved away from this method (it provides strictly less information than `realProfiles` and `discreteProfiles` do, while using more memory) - so we don't expect this method to get more use in the future. If I had my druthers, we'd remove this method entirely, but we're keeping the `resourceSamples` hasura action in the API as a convenience for clients that don't want to work with continuous profiles.

So this refactor is setup for an eventual deprecation? If so, then I think I'm fine with making it the responsibility of whoever calls the public method to cache its return value.

mattdailis:
> So this refactor is setup for an eventual deprecation? If so, then I think I'm fine with making it the responsibility of whoever calls the public method to cache its return value.

When I brought up the possibility of deprecation, @patkenneally explained that we should keep the capability in case API clients want it, but that we should encourage people to use the continuous profiles directly. So - yes, it's "discouraged" if not exactly "deprecated for removal".

Twisol:
Personally, I'd like to see the `resourceSamples` code moved out of `SimulationResults` entirely. It looks like the only place where we access it outside of testing/development code is that API endpoint, and the computation doesn't really feel like the responsibility of `SimulationResults` anymore.

mattdailis:
> Personally, I'd like to see the resourceSamples code moved out of SimulationResults entirely

Done :+1: The sampling logic is now the responsibility of the `GetSimulationResultsAction`, which was the only client of this functionality. I inlined it into the `getResourceSamples` method.

Twisol:
Awesome! Thanks for doing that!

examples/foo-missionmodel/src/test/resources/gov/nasa/jpl/aerie/foomissionmodel/plan.json
```diff
@@ -2,24 +2,29 @@
     "type": "foo",
     "arguments":
       { "x": 1,
-        "y": "test_1" } },
+        "y": "test_1",
+        "z": 4 } },
```
Mythicaeda, line 6: Quick sanity check: where did the new `z` argument come from?

mattdailis: Good question :+1: It came from #334, when we wanted to include a "required" parameter in our sample mission models - since we had a regression where required parameters would fail to instantiate.

This plan.json is used in the main method in `SimulateMapSchedule` - it's not used for any automated tests, but it can be used for manual testing. We should have added the z parameter in #334. 

It's unrelated to this PR - so I would accept the argument that it should be split into its own PR :+1: But it is such a small, inconsequential change, that I'd also be okay with bundling it into this PR.

Twisol: It's in its own commit, so I think it's fine. IMHO, we should feel comfortable cleaning up things as we come in contact with code, even if it's technically not required for the task at hand.

Twisol: It would be / would have been good to include that context in the commit message. (Remember, the title needs to be short, but you can have as long a commit body as you desire.)

Mythicaeda (approved):

PR #424, opened 2022-11-07
[FIX] Load scheduling DSL TypeScript libraries in `scheduler-server`

pcrosemurgy:
* **Tickets addressed:** None
* **Review:** By commit  

## Description
Since splitting `scheduler-server` into `scheduler-worker` – and thereby relocating the scheduling DSL Node project – the scheduler server has not had access to static DSL libraries. This PR loads scheduling DSL TypeScript libraries from the scheduler server JAR instead of requiring Docker to be configured to copy in the directory and set an environment variable.

## Verification
- [x] Unit test added to test loading `.ts` files from the `scheduler-server` JAR.
- [x] Manually tested end-to-end.
- [x] Automated end-to-end tests.

## Documentation
None.

## Future work
None.

### DIFF
--- a/scheduler-server/build.gradle
+++ b/scheduler-server/build.gradle
@@ -1,8 +1,9 @@
+import java.nio.file.Path
+
 plugins {
   id 'java'
   id 'java-test-fixtures'
   id 'application'
-  id 'com.github.node-gradle.node' version '3.5.0'
 }
 
 java {
@@ -40,6 +41,11 @@ dependencies {
   testImplementation 'org.glassfish:javax.json:1.1.4'
 
   testFixturesImplementation project(':merlin-driver')
+}
+
+processResources {
+  // Copy in scheduling DSL compiler static libraries when processing resources
+  from project(':scheduler-worker').projectDir.toPath().resolve(Path.of('scheduling-dsl-compiler', 'src', 'libs'))
 }
 
 test {
--- a/e2e-tests/src/tests/scheduler.test.ts
+++ b/e2e-tests/src/tests/scheduler.test.ts
@@ -32,6 +32,11 @@ test.describe('Scheduling', () => {
     await delay(2000);
   });
 
+  test('Get scheduling DSL TypeScript', async ({ request }) => {
+    const schedulingDslTypes = await req.getSchedulingDslTypeScript(request, mission_model_id);
+    expect(schedulingDslTypes.typescriptFiles.length).toEqual(7);
+  });
+
   test('Create Plan', async ({ request }) => {
     const plan_input : CreatePlanInput = {
       model_id : mission_model_id,
--- a/e2e-tests/src/types/scheduling-goal.d.ts
+++ b/e2e-tests/src/types/scheduling-goal.d.ts
@@ -1,8 +1,13 @@
 type SchedulingDslTypesResponse = {
   reason: string;
   status: 'failure' | 'success';
-  typescript: string;
+  typescriptFiles: TypescriptFile[];
 };
+
+type TypescriptFile = {
+  filePath: string;
+  content: string;
+}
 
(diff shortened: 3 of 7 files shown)


JoelCourtney:
> Manually tested end-to-end.

Can you automate it instead just for good measure? We already have a set of scheduler e2e tests that should do the busy work of uploading a model and creating goals.

pcrosemurgy:
@JoelCourtney Sure! I'll give it a shot.

scheduler-server/build.gradle
```diff
@@ -42,6 +43,11 @@ dependencies {
   testFixturesImplementation project(':merlin-driver')
 }
 
+processResources {
+  // Copy in scheduling DSL compiler static libraries when processing resources
+  from project(':scheduler-worker').projectDir.toPath().resolve(Path.of('scheduling-dsl-compiler', 'src', 'libs'))
```
camargo, line 48: Why does `scheduling-dsl-compiler` live in the worker directory and not the server directory? Asking because it is the opposite of how the merlin `constraints-dsl-compiler` is organized.

pcrosemurgy: Good question. So with `merlin-*` the server executes the Node DSL code. However, with `scheduler-*` the worker is the service that actually executes Node DSL code. The `scheduler-server` just needs access to static TS files but does not perform any TS execution.

pcrosemurgy: This is why we're restricting `scheduler-server`'s access to be limited to just `scheduling-dsl-compiler/src/libs/`.

pcrosemurgy: Might be a good idea to eventually move `constraints-dsl-compiler` and `scheduling-dsl-compiler` to the top-level Aerie repo directory since both are used by more than 1 subproject (ex: `constraints-dsl-compiler` is actually used within the `scheduling-dsl-compiler` through some Gradle copy-on-build business). But I think that should be done outside of this PR.

camargo (approved):
Tested this branch against the UI e2e tests as well and all pass!

pcrosemurgy:
Not sure why this is failing:
```
1) src/tests/scheduler.test.ts:35:3 › Scheduling › Get scheduling DSL TypeScript =================

    ReferenceError: mission_model_id is not defined

      34 |
      35 |   test('Get scheduling DSL TypeScript', async ({ request }) => {
    > 36 |     typescriptFiles = await req.getSchedulingDslTypeScript(request, mission_model_id);
         |                                 ^
      37 |     expect(typescriptFiles).toBeDefined();
      38 |   });
      39 |

        at Object.getSchedulingDslTypeScript (file:///home/runner/work/***/***/e2e-tests/src/utilities/requests.ts:[176](https://github.com/NASA-AMMOS/aerie/actions/runs/3414650049/jobs/5682881013#step:13:177):97)
        at file:///home/runner/work/***/***/e2e-tests/src/tests/scheduler.test.ts:36:33
```
The `test('Get scheduling DSL TypeScript', ...)` comes right after `test('Upload jar and create mission model', ...)` which sets `mission_model_id`.

**Edit**: I think I've fixed this. The above error is strange as it makes it look like `mission_model_id` within the context of `await req.getSchedulingDslTypeScript(request, mission_model_id);` is undefined. However, it was actually undefined within `utilities/requests.ts`. The trace preview here is misleading IMO.

pcrosemurgy:
> > Manually tested end-to-end.
> 
> Can you automate it instead just for good measure? We already have a set of scheduler e2e tests that should do the busy work of uploading a model and creating goals.

@JoelCourtney Done :)

mattdailis (approved):
Did a quick manual check locally - intellisense is working in the scheduling goal editor again 🙏 Thanks for the quick turnaround, Paul 🎉

PR #434, opened 2022-11-08
Use a record instead of a pair for profile segments

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Profile segments in simulation results are currently represented as a `Pair<Duration, Dynamics>` - it is difficult to tell from this representation what the meaning of the "Duration" should be. By convention, we interpret the `Duration` to be the `extent` of the Profile Segment - i.e. the length of this segment. Other plausible interpretations would be the offset from the start of the plan, as well as the length of the _previous_ segment.

This PR introduces a `ProfileSegment` record to make it clear which interpretation is being used, and to pave the way for changing that interpretation in the near future. (The current interpretation is somewhat incompatible with streaming - since as it stands you cannot emit a segment until its end time is known).

This PR updates `SimulationResults` as well as all uses of profiles that have `merlin-driver` as a dependency. It draws the line at the `constraints` package, which does not depend on `merlin-driver`.

As @JoelCourtney pointed out, we have numerous representations of profiles, and it's hard to keep them all straight - while this PR does not reduce the number of representations, it hopefully makes one of those representations a little easier to work with.

## Verification

I spun up a local aerie and ran constraints checking, just to make sure that pipeline still works. Otherwise, I consider the fact that the code compiles, and that the tests pass on this PR to be sufficient verification.

## Documentation

The `ProfileSegment` class has associated javadoc, but aside from that no documentation has been updated.

## Future work

Future work: simulation results streaming

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ProfileSet.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ProfileSet.java
@@ -1,6 +1,6 @@
 package gov.nasa.jpl.aerie.merlin.server.models;
 
-import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.driver.engine.ProfileSegment;
 import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
@@ -15,12 +15,12 @@ import java.util.stream.Collectors;
 import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
 
 public record ProfileSet(
-    Map<String, Pair<ValueSchema, List<Pair<Duration, Optional<RealDynamics>>>>> realProfiles,
-    Map<String, Pair<ValueSchema, List<Pair<Duration, Optional<SerializedValue>>>>> discreteProfiles
+    Map<String, Pair<ValueSchema, List<ProfileSegment<Optional<RealDynamics>>>>> realProfiles,
+    Map<String, Pair<ValueSchema, List<ProfileSegment<Optional<SerializedValue>>>>> discreteProfiles
 ) {
   public static ProfileSet ofNullable(
-      final Map<String, Pair<ValueSchema, List<Pair<Duration, Optional<RealDynamics>>>>> realProfiles,
-      final Map<String, Pair<ValueSchema, List<Pair<Duration, Optional<SerializedValue>>>>> discreteProfiles
+      final Map<String, Pair<ValueSchema, List<ProfileSegment<Optional<RealDynamics>>>>> realProfiles,
+      final Map<String, Pair<ValueSchema, List<ProfileSegment<Optional<SerializedValue>>>>> discreteProfiles
   ) {
     return new ProfileSet(
         realProfiles,
@@ -28,8 +28,8 @@ public record ProfileSet(
     );
   }
   public static ProfileSet of(
-      final Map<String, Pair<ValueSchema, List<Pair<Duration, RealDynamics>>>> realProfiles,
-      final Map<String, Pair<ValueSchema, List<Pair<Duration, SerializedValue>>>> discreteProfiles
+      final Map<String, Pair<ValueSchema, List<ProfileSegment<RealDynamics>>>> realProfiles,
+      final Map<String, Pair<ValueSchema, List<ProfileSegment<SerializedValue>>>> discreteProfiles
   ) {
     return new ProfileSet(
         wrapInOptional(realProfiles),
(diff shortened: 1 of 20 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ProfileSet.java
```diff
@@ -66,7 +66,7 @@ public static <T> Map<String, Pair<ValueSchema, List<Pair<Duration, T>>>> unwrap
                 $.getValue().getLeft(),
                 $.getValue().getRight()
                  .stream()
-                 .map(untuple((duration, dynamics) -> Pair.of(duration, dynamics.get())))
+                 .map(untuple(segment -> new ProfileSegment<>(segment.extent(), segment.dynamics().get())))
```
JoelCourtney: I don't think untuple is necessary here anymore

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -162,14 +162,14 @@ public Map<String, List<Violation>> getViolations(final PlanId planId)
         .orElseGet(Collections::emptyMap);
     final var discreteProfiles = new HashMap<String, DiscreteProfile>(_discreteProfiles.size());
     for (final var entry : _discreteProfiles.entrySet()) {
-      discreteProfiles.put(entry.getKey(), DiscreteProfile.fromSimulatedProfile(entry.getValue().getRight()));
+      discreteProfiles.put(entry.getKey(), DiscreteProfile.fromSimulatedProfile(entry.getValue().getRight().stream().map($ -> Pair.of($.extent(), $.dynamics())).toList()));
```
JoelCourtney: Better to update `fromSimulatedProfile` and `fromExternalProfile` instead of converting to the old format, because with this change all use sites will be using the new format.

JoelCourtney (changes requested):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
mattdailis: The issue I ran into there is that `fromSimulatedProfile` is defined in the `constraints` subproject, which doesn't depend on the `merlin-driver` subproject (which is where I put the `ProfileSegment` record).

Options are:
- Convert between the two in `merlin-server`, which depends on both subprojects
- Make one depend on the other
- Put `ProfileSegment` in a different subproject that both `constraints` and `merlin-driver` depend on.

There is one subproject that fits the bill for the last point - `merlin-sdk`(i.e. `merlin.protocol`) - though its intended purpose is to provide an interface between the mission model and the merlin-driver - not between the merlin-driver and downstream processes... But I could convince myself that resource profiles are a central enough concept to warrant membership in the `merlin.protocol`.

Twisol: Since `merlin.protocol` is provided by the runtime environment and not bundled with the model, I could see an argument for putting it there. But I don't really think it's something a model ever needs, so I'm not really a fan of that argument.

On the other hand, what if we just merged `constraints` into `merlin-driver`? Are there any clients of one that aren't already clients of the other? And either way, do we expect it to be a burden?

JoelCourtney: I like putting it in merlin-sdk. I think putting it next to the types that it's made of (`Duration`, `RealDynamics`, `SerializedValue`) makes a lot of sense.

mattdailis: I'm piecing together that `ProfileSegment` needs to depend on `merlin-sdk` (as you pointed out, Joel, it is composed of pieces from there), but that it's a downstream detail that should not be made available to mission models (as you pointed out, Jonathan).

Merging `constraints` into `merlin-driver` is a bit bigger of a change than I was hoping to make in this PR, so I'll take the middle ground of making `constraints` depend on `merlin-driver`. We've had to duplicate everything to do with `SimulationResults` between the two, and this discussion is highlighting that anew. I'd be interested in discussing further where our "lines in the sand" are best drawn beyond the context of this PR - perhaps a Github discussion (or a thread in public slack 🥺) is warranted.

I'll report back once I've made the change

mattdailis: For ease of review, I've made the changes in a separate commit: https://github.com/NASA-AMMOS/aerie/commit/5c990bf9b7bbbaed550b6cb9aa290eafe4763adb

JoelCourtney (approved):

camargo (approved):

PR #360, opened 2022-10-04
[Aerie-2118] Database Migration Script Prototype

goetzrrGit:
* **Tickets addressed:** AERIE-2118
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description
I prototyped a migration strategy from our GitHub discussion on [Database migration](https://github.com/NASA-AMMOS/aerie/discussions/346).

The PR has examples of database changes, `up.sql` are migrations, and `down.sql` are DB reverts. The idea is that each up/down.sql will have 3 sections:

* Database changes - alteration to the database such as creation or deletion of tables, columns, etc.
* Data Migration - any data that needs to be migrated to a new table should go here 
* Database Check - Any checks you can do with the database after the alteration is made. This is the first line of defense in preventing a corrupted database.

The migration script is a python CLI tool 

```
./aerie_db_migration -h                      
usage: aerie_db_migration [-h] [-a] [-r] [-db DB_NAME]

Migrate an AERIE Database

optional arguments:
  -h, --help            show this help message and exit
  -a, --apply           apply all migration steps to all databases if one isn't provided
  -r, --revert          revert all migration steps to all databases if one isn't provided
  -db DB_NAME, --db-name DB_NAME
                        name of a specific DB to apply or revert all migration steps
```

If you want to run a manual mode you should run the script without any arguments:
`Not Present` means you haven't applied the migration to the database yet

```
./aerie_db_migration
#########################
Manual DB Migration Mode
#########################

0) 

QUIT MANUAL MODE...

1)

Database: AerieScheduler
No migration steps

2)

Database: AerieMerlin
No migration steps

3)

Database: AerieCommanding
VERSION        NAME                   SOURCE STATUS  DATABASE STATUS
1664478381304  rename_mission_column  Present        Not Present
1664478481307  add_column_expand_to   Present        Not Present

4)

Database: AerieUI
No migration steps

Which DB would you like to apply migrations to (0-4):        
```

To apply all migration steps automatically per database run the command:
```
./aerie_db_migration -a
```

To revert all migration steps automatically per database run the command:
```
./aerie_db_migration -r
```

To target a specific database to apply all the migration steps automatically run the command:
```
./aerie_db_migration -a -db AerieCommanding
```

## Verification
Here is the full output from the script when applying the migration across all databases:

```

rrgoetz@MT-306238 aerie % ./aerie_db_migration -a
INFO hasura cli                                    version=v2.12.0

#################
AerieScheduler
#################

No migration steps present...

#################
[...]

PR #438, opened 2022-11-08
[AERIE 2131] Database Migration Script

Mythicaeda:
* **Tickets addressed:** AERIE-2118, AERIE-2131
* **Review:** By commit  

## Description

Finishes the python CLI tool began in #360.

The script now:

- avoids displaying databases that don't have migrations
- avoids attempting to apply/revert changes that have already been applied
- prevents the user from skipping a migration when in step-by-step
- has the user start the tool in either `apply` or `revert` mode, rather than allowing them to choose per-migration-step
- no longer raises an exception if not given an integer on the select database screen
- allows the user to select a list of databases to migrate, rather than either one or all
- displays an error in the event that a user attempts to migrate a database when they are missing some of the migration steps that exist on the server

For more information on the layout of the `up.sql`/`down.sql`, see #360. 

## Verification

Ran the script against the migrations found in `deployment/sample-migrations` and verified that they ran both up and down without issues. (Note: `test_migration` in AerieMerlin requires two plans to exist prior to being able to apply it, as it's a demonstration that we can insert/delete/modify data in migrations, rather than a schema change)

## Documentation

* Added a new page to `Upgrade Guides`, `Aerie Database Migration Script`.
* Updated the index page for `Upgrade Guides` to display the Table of Contents.


deployment/aerie_db_migration
```diff
@@ -0,0 +1,239 @@
...
+      to_migrate_set.discard(db)
+
+if to_migrate_set:
+  print("\033[91mError\033[0m: The following Database(s) do not contain migrations:\n\t"
+                          +"\n\t".join(to_migrate_set))
+  sys.exit()
+
+if not migration_db:
+  print("\033[91mError\033[0m: No database migrations found.")
+  sys.exit()
```
pcrosemurgy: I'd make this a `sys.exit(1)` to indicate that the process failed to the shell.

pcrosemurgy: This applies to all the other exits as well :)

Mythicaeda: Done, although I left the `sys.exit()` following the user entering a quit key untouched ;)

pcrosemurgy: Thanks! Good call

camargo (approved):

pcrosemurgy (approved):
Nice work! This looks good to me.

PR #439, opened 2022-11-08
refactor: rename command-expansion-server

camargo:
* **Tickets addressed:** [AERIE-2055](https://jira.jpl.nasa.gov/browse/AERIE-2055)
* **Review:** By commit  

## Description

* Rename `command-expansion-server` to `sequencing-server`
* No downstream API changes necessary because table names are the same
* Will require updating downstream deployment docker compose files or custom deployment scripts

## Verification

* Ran all UI e2e tests to make sure nothing broke

## Documentation

* Updated as part of this PR where applicable

## Future work

* Update the `sequence` table name to `expansion_sequence`
* Update the `user_sequence` table name to `authored_sequence`


mattdailis (approved):
I'm waffling on the word "sequencing" vs the word "commanding" - not sure which is better with respect to the domain.

I see your point that this service does more than command expansion now, so it does warrant a rename.

I'd like to add this commit to `.git-blame-ignore-revs` - I can do that now.

AaronPlave (approved):

PR #440, opened 2022-11-09
[HOTFIX] Commit merge with delete-modify conflict resolution edge cases

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

When doing some manual testing of the new merge UI, I noticed an unexpected behavior:

1. I deleted an activity in the parent
2. I modified that same activity in the child
3. During merge conflict resolution, I chose to use the activity from the child
4. After accepting the changes, the parent plan should have contained this activity, but it did not

For each merge staging area entry with the `'modify'` change type, the `commit_merge` function updates the corresponding activity in the target plan. However, if there are no corresponding activities in the target plan (as in the case of a delete-modify conflict), no change is applied to the target plan. This is not quite what we want.

The desired behavior is to "update the corresponding activity if it exists, and insert it if it does not". This "update or insert" behavior (colloquially known as an "upsert") is supported by postgres via the "[insert on conflict](https://www.postgresql.org/docs/devel/sql-insert.html#SQL-ON-CONFLICT)" statement, which we use in other places, such as the [CreateActivityType](https://github.com/NASA-AMMOS/aerie/blob/0d70641dadfcb949d8ff9a8cb631e89e77a92164/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateActivityTypeAction.java#L16-L22) action.

This PR starts by adding a failing test (described below), then making the changes necessary to fix it.

One note is that I modified the `generate_activity_directive_name` trigger function to avoid overwriting a name that has been provided. This makes its behavior more like "generated by default" and allows us to copy activity directives into the table without losing their names.

## Verification

Two new cases have been added to the existing `modifyAndDeletesApplyCorrectly` test - to help cover the options of:
- Delete activity in base, modify in child
- Modify activity in base, delete in child

And, for each of those:
- Resolve supplying
- Resolve receiving

(two of the four cases were already covered by the existing test - this PR just adds the remaining two)

## Documentation

No documentation updates were made.

## Future work

Perhaps a survey of our triggers for gotchas like the generated activity directive name would be helpful. We may also consider adding a test case that adds are applied correctly.

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
@@ -2045,8 +2045,8 @@ public class PlanCollaborationTests {
       final int deleteUncontestedActId = insertActivity(basePlan);
       final int modifyContestedSupplyingActId = insertActivity(basePlan);
       final int modifyContestedReceivingActId = insertActivity(basePlan);
-      final int deleteContestedSupplyingActId = insertActivity(basePlan);
-      final int deleteContestedReceivingActId = insertActivity(basePlan);
+      final int deleteContestedSupplyingResolveSupplyingActId = insertActivity(basePlan);
+      final int deleteContestedReceivingResolveReceivingActId = insertActivity(basePlan);
       final int childPlan = duplicatePlan(basePlan, "Child");
 
       assertEquals(6, getActivities(basePlan).size());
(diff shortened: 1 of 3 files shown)


camargo (approved):
Good find. Tested this flow locally in the UI and works as expected.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -2165,6 +2131,22 @@ void commitMergeCleansUpSuccessfully() throws SQLException{
     }
   }
 
+  private static void assertActivityEquals(final Activity expected, final Activity actual) {
```
Mythicaeda: Can this be moved up to where the other helper methods are?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -2030,12 +2029,14 @@ void modifyAndDeletesApplyCorrectly() throws SQLException {
       final int deleteUncontestedActId = insertActivity(basePlan);
       final int modifyContestedSupplyingActId = insertActivity(basePlan);
       final int modifyContestedReceivingActId = insertActivity(basePlan);
-      final int deleteContestedSupplyingActId = insertActivity(basePlan);
-      final int deleteContestedReceivingActId = insertActivity(basePlan);
+      final int deleteContestedSupplyingActIdResolveSupplying = insertActivity(basePlan);
```
Mythicaeda: Minor refactor: Can these variables be renamed so that `ActId` is at the end (ie: `deleteContestedSupplyingResolveSupplyingActId`)?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Done :+1:

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Makes sense! Nouns come after adjectives in English 🤦 Done 👍

Mythicaeda (approved):

PR #475, opened 2022-11-11
Add constraints to scheduling specifications

### LINKED ISSUE #471: Add constraints to the scheduling_specification and scheduling_specification_goals tables
1. Add a constraint to the [scheduling_specification](https://github.com/NASA-AMMOS/aerie/blob/develop/scheduler-server/sql/scheduler/tables/scheduling_specification.sql#L5) table requiring the `plan_id` to be unique so only one specification can be created per plan.
1. Add a constraint to the [scheduling_specification_goals](https://github.com/NASA-AMMOS/aerie/blob/develop/scheduler-server/sql/scheduler/tables/scheduling_specification_goals.sql#L3) table requiring the `goal_id` to be unique such that you cannot add the same goal to multiple specifications (and thus multiple plans).

mattdailis:
* **Tickets addressed:** Closes #471
* **Review:** By commit  

## Description

After consulting with users, we determined that for the present state of the product, sharing goals across specifications, and specifications across plans, would do more harm than good.

These constraints will help avoid those scenarios. We can remove these constraints at a later date if we deem them unnecessary. It is easier from a database migration perspective to remove constraints than it is to add them later.

## Verification

Still working on verification...

## Documentation

We should mention that this is 

## Future work

No future work

### DIFF
--- a/scheduler-server/sql/scheduler/tables/scheduling_specification.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_specification.sql
@@ -9,7 +9,9 @@ create table scheduling_specification (
   simulation_arguments jsonb not null,
   analysis_only boolean not null,
   constraint scheduling_specification_synthetic_key
-    primary key(id)
+    primary key(id),
+  constraint scheduling_specification_unique_per_plan
+    unique (plan_id)
 );
 
 comment on table scheduling_specification is e''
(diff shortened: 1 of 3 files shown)


JoelCourtney (commented):
I like what I see so far but I'll wait until its done to approve.

I think that this is _technically_ not in normal form, because the constraints on the spec_goals and spec_conditions tables effectively turns the primary key into just `goal_id` and `condition_id` respectively. And any time you have a foreign primary key you should merge the tables because their rows match one-to-one.

But that would make it more difficult to undo this change, so I'm ambivalent.

mattdailis:
Can't muster the brainpower to update the tests 😅 Closing for now

PR #487, opened 2022-11-16
Add external dataset CSV example

### LINKED ISSUE #453: Add better documentation for adding external profiles with "real world" examples
Right now the [documentation for adding external profiles](https://nasa-ammos.github.io/aerie/develop/user-guide/ui-api-guide/external-datasets/managing.html) is too simple. We need to show a more complex "real world" example. One common use case Clipper has is converting a CSV into external profiles for upload. We should get an example from them that we can highlight in our docs.

Internal ticket with real CSV data: https://jira.jpl.nasa.gov/browse/AERIEQS-423

camargo:
* Closes https://github.com/NASA-AMMOS/aerie/issues/453
* **Review:** By commit  

## Description

* Add new example on how to import CSV to external dataset
* Cleanup document - formatting and wording
* Remove API example and just keep mutation local to the managing doc

### DIFF
--- a/docs/source/user-guide/ui-api-guide/external-datasets/managing.md
+++ b/docs/source/user-guide/ui-api-guide/external-datasets/managing.md
@@ -1,27 +1,38 @@
 # Managing External Datasets
 
-External profiles are quite similar to the profiles produced by a Merlin simulation. They describe the behavior of some dynamic characteristic over time and can be viewed in the Aerie UI along with a plan. External profiles come from the user, rather than Merlin's simulation capabilities, and can be uploaded at any time, before or after simulation.
+This document describes how to add external datasets to Aerie.
 
-Ultimately, the purpose of external profiles is up to the user. You might be looking to include some geometry information to assist in building a plan, or you may simply be adding power and thermal modeling results to be viewable with existing simulation results. Just be aware that at this time external profiles are not accessible to the mission model during simulation, and are simply for viewing purposes in the UI.
+External datasets contain resource profiles that are quite similar to the profiles produced by a Merlin simulation. A resource profile describes the behavior of a resource over time. Each profile has a name, a type, a start time, and a list of profile segments that conform to that type. Each profile segment defines the value of the resource over a duration of time (referred to as its "dynamics"). The first segment starts at the start time of the profile, and every subsequent segment starts at the end of the previous segment.
 
-## How To Upload A Dataset
-To upload external profiles, you will need to add them to a specific plan. A collection of profiles that all start at the same time are called an "External Dataset". The dataset must be associated with a plan, and will persist as long as the plan exists, or until it is deleted by a user. An external dataset can be added via the GraphQL mutation `addExternalDataset`.
+External datasets come from the user rather than Merlin's simulation capabilities, and can be uploaded at any time before or after simulation. They can be viewed in the Aerie UI along with a plan.
 
-### GraphQL Mutation: `addExternalDataset`
-The `addExternalDataset` GraphQL mutation takes three parameters as specified below:
+Ultimately the purpose of external datasets is up to the user. You might be looking to include some geometry information to assist in building a plan, or you may simply be adding power and thermal modeling results to be viewable with existing simulation results. Just be aware that at this time external profiles are not accessible to the mission model during simulation, and are simply for viewing purposes in the UI.
 
-| Parameter      | Type    | Description                                                                      |
-|----------------|---------|----------------------------------------------------------------------------------|
-| `planId`       | Integer | The ID of the plan to associate the external dataset with.                       |
-| `datasetStart` | String  | The UTC timestamp the dataset is based from.<br/>UTC Format: `yyyy-dddThh:mm:ss` |
-| `profileSet`   | Object  | The set of precomputed profiles that make up the external dataset.               |
+## Add External Dataset Mutation
 
-The profile set to be uploaded should have one entry for each profile, indexed by a unique name mapping to an object specifying the details of the profile. 
-Each profile should have a `type` field, which specifies whether the profile is real- or discrete-valued. 
-It must also contain a `schema` field, which specifies the schema of the values it takes on. 
-For discrete profiles, these are not limited to basic types, but can take on any complex structure made up using our `ValueSchema` construct (for more information, see our [ValueSchema documentation](../../mission-modeler-guide/custom-value-types/index.rst)). 
+To upload an external dataset you need to add it to a specific plan. The dataset will persist as long as the plan exists, or until it is explicitly deleted. An external dataset can be added via the GraphQL mutation `addExternalDataset`:
+
+```{eval-rst}
+  .. include:: ../api-examples.rst
+    :start-after: begin add external dataset
+    :end-before: end add external dataset
+```
+
+The `addExternalDataset` GraphQL mutation takes three query variables as specified below:
+
+| Parameter       | Type    | Description                                                                       |
+| --------------- | ------- | --------------------------------------------------------------------------------- |
+| `$planId`       | Integer | The ID of the plan to associate the external dataset with                         |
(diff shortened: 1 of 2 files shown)


docs/source/user-guide/ui-api-guide/external-datasets/managing.md
```diff
@@ -1,27 +1,36 @@
...
 
-## How To Upload A Dataset
-To upload external profiles, you will need to add them to a specific plan. A collection of profiles that all start at the same time are called an "External Dataset". The dataset must be associated with a plan, and will persist as long as the plan exists, or until it is deleted by a user. An external dataset can be added via the GraphQL mutation `addExternalDataset`.
+## Add External Dataset Mutation
 
-### GraphQL Mutation: `addExternalDataset`
-The `addExternalDataset` GraphQL mutation takes three parameters as specified below:
+To upload an external dataset you need to it to a specific plan. The dataset will persist as long as the plan exists, or until it is explicitly deleted. An external dataset can be added via the GraphQL mutation `addExternalDataset`:
```
mattdailis:
```suggestion
To upload an external dataset you need to add it to a specific plan. The dataset will persist as long as the plan exists, or until it is explicitly deleted. An external dataset can be added via the GraphQL mutation `addExternalDataset`:
```

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
```diff
@@ -1,27 +1,36 @@
 # Managing External Datasets
 
-External profiles are quite similar to the profiles produced by a Merlin simulation. They describe the behavior of some dynamic characteristic over time and can be viewed in the Aerie UI along with a plan. External profiles come from the user, rather than Merlin's simulation capabilities, and can be uploaded at any time, before or after simulation.
+External datasets contain profiles that are quite similar to the profiles produced by a Merlin simulation. They describe the behavior of some dynamic characteristic over time and can be viewed in the Aerie UI along with a plan. External profiles come from the user, rather than Merlin's simulation capabilities, and can be uploaded at any time before or after simulation.
```
mattdailis: While _we_ know the shape of profiles produced by merlin simulation - I think we should assume that someone coming to this page might not know. We can either include a link to where that format is described, or describe it here.

In particular, I think it's important to define the word `profile`, as well as the concept of a `profile_segment` somewhere.

(May be out of scope of your current effort - we can follow up on this)

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
```diff
@@ -1,27 +1,36 @@
...
 
-Ultimately, the purpose of external profiles is up to the user. You might be looking to include some geometry information to assist in building a plan, or you may simply be adding power and thermal modeling results to be viewable with existing simulation results. Just be aware that at this time external profiles are not accessible to the mission model during simulation, and are simply for viewing purposes in the UI.
+Ultimately the purpose of external profiles is up to the user. You might be looking to include some geometry information to assist in building a plan, or you may simply be adding power and thermal modeling results to be viewable with existing simulation results. Just be aware that at this time external profiles are not accessible to the mission model during simulation, and are simply for viewing purposes in the UI.
```
mattdailis:
```suggestion
Ultimately the purpose of external datasets is up to the user. You might be looking to include some geometry information to assist in building a plan, or you may simply be adding power and thermal modeling results to be viewable with existing simulation results. Just be aware that at this time external profiles are not accessible to the mission model during simulation, and are simply for viewing purposes in the UI.
```

mattdailis (approved):

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
camargo: Happy to add it to this commit. Do we have a good working definition for each?

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
```diff
@@ -1,27 +1,36 @@
...
-The `addExternalDataset` GraphQL mutation takes three parameters as specified below:
+To upload an external dataset you need to add it to a specific plan. The dataset will persist as long as the plan exists, or until it is explicitly deleted. An external dataset can be added via the GraphQL mutation `addExternalDataset`:
 
-| Parameter      | Type    | Description                                                                      |
-|----------------|---------|----------------------------------------------------------------------------------|
-| `planId`       | Integer | The ID of the plan to associate the external dataset with.                       |
-| `datasetStart` | String  | The UTC timestamp the dataset is based from.<br/>UTC Format: `yyyy-dddThh:mm:ss` |
-| `profileSet`   | Object  | The set of precomputed profiles that make up the external dataset.               |
+```
+mutation AddExternalDataset($planId: Int!, $datasetStart: String!, $profileSet: ProfileSet!) {
```
Mythicaeda: This mutation and `DeleteExternalDataset` [already existed in `api-examples.rst`](https://github.com/NASA-AMMOS/aerie/blob/develop/docs/source/user-guide/ui-api-guide/api-examples.rst#add-an-external-dataset). Instead of deleting it from that file, could we instead update the mutation defined there and use an `include` directive?

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
mattdailis: Not sure about a working definition, I'll try to brainstorm something... (feel free to weigh in @joelcourtney) some of this may be redundant with the rest of this page, I'm just spitballing:

A resource profile describes the behavior of a resource over time. Each profile has a name, a type, a start time, and a list of profile segments that conform to that type. 

Each profile segment defines the value of the resource over a duration of time (referred to as its "dynamics"). The first segment starts at the start time of the profile, and every subsequent segment starts at the end of the previous segment.

A segment can also represent a gap in data - in which case it would have a duration but no dynamics.

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
camargo: No I intentionally removed it in this case. I don't think we need it in both places.

Mythicaeda: I thought `api-examples.rst` was a reference page containing all of our non-example API queries (`non-example` meaning 'not tailored to a specific example', like the mutations originally under the `Example 1/2/3` headers on this page). If this is a mistaken understanding then could you clarify what `api-examples.rst` is supposed to contain?

camargo: Previously our `api-examples` page existed to show queries we did not have explained elsewhere (or some different way to show a query we had elsewhere). In this case these queries are explained in this doc, and I don't think there is much utility in having them in `api-examples` without the context. When users ask about how to add external profiles in Aerie we should point them to this page and not the API examples page.

Mythicaeda: Right, but when we finish the documentation I don't know if there's any queries currently on `api-examples` that wouldn't also be on a explanation page. This is where my understanding of a "quick syntax reference page" comes from. If we're concerned about users needing the context, however, we can link to the appropriate page per header, along with adding a statement at the top of the page along the lines of "This page only contains example GraphQL queries used with Aerie. For an explanation of how each query works and when to use it, see the corresponding documentation." We could also rename the page to better reflect this change in expected use.

Additionally, wouldn't having the queries all hosted in one spot make it easier for us as developers to update the queries (especially if we link to the explanation page)?

JoelCourtney:
Headings like "Example 1" "Example 2" will not be very informative on a table of contents. Can you change them to more descriptive headings please?

docs/source/user-guide/ui-api-guide/external-datasets/managing.md
camargo: All good points @Mythicaeda. Let's discuss more next week.

Issue #464, opened 2022-11-10
Prototype lazy evaluation for Profiles in constraints


JoelCourtney:
Second half of the solution outlined in #417.

PR #488, opened 2022-11-17
Lazy profiles

JoelCourtney:
* **Tickets addressed:** #464 
* **Review:** By commit  

## Description

This prototype refactors profiles to be evaluated lazily. For context to the carnage below, the main reason for the design is that the bounds of evaluation are not known when a profile is created, only when it is used.

This branch doesn’t compile in its current state. It did compile and pass some tests in an old state, but I started updating usage sites and squashed some commits, so for now you’ll have to take me at my word that it does actually compile and produce correct results.

The meat is in the first commit. The second commit is a half-finished attempt to update use sites to the new interface.

### Design

In the old design, profiles are concrete `IntervalMap` lists of segments, and operations transform those lists. In the new design, profiles are objects that can lazily produce a stream of segments on request, and that stream can be reified into a concrete interval map.

The design revolves around the `Profile<V>` functional interface:

```java
interface Profile<V> {
	Stream<Segment<V>> stream(Interval bounds);

	default IntervalMap<V> evaluate(final Interval bounds) {
  		final var stream = this.stream(bounds);
 		final var result = IntervalMap.<V>builder();
	        stream.forEachOrdered(result::set);
		return result.build();
	}
	
	// … many other default and static methods
}
```

Until `evaluate` is called, the only parts of a profile that are concretely stored are the base components of an expression, such as simulated resources, activity parameters, or constants. Operations produce new profiles that transform the streams of other profiles. Here’s an example from `Windows`:

```java
interface Windows extends Profile<Boolean> {
  default Windows not() {
    return bounds -> this.stream(bounds) // this is a stream of segments
      .map(segment -> segment.mapValue($ -> !$));
  }
}
```

Since `Profile` is a functional interface, and `Windows` does not add any extra required methods, `Windows` is a functional interface too.

It took me a while to understand how this worked, even after I designed it. So here’s the play-by-play of what the expression `myWindows.not().evaluate(bounds)` does:
- myWindows is some instance of an anonymous class that implements `Windows`, so it can produce a stream of segments given bounds.
- `not` wraps that object in another instance of an anonymous Windows implementor, which produces a stream by calling `stream(bounds)` on its child and transforming the result.
- `evaluate` calls `stream`, which collapses the stack of anonymous Profile implementors into a mirror stack of Stream implementors, which it then drains and reifies into an `IntervalMap`.

This design makes the idea that `Windows` is just a boolean profile with extra operations a reality. Similarly `LinearProfile` extends `Profile<LinearEquation>` and has the same relationship with a few overridden default methods.

### Consequences

- There is no explicit interface for `DiscreteProfile`, because all of its potential behavior could be included in `Profile`. (Some consequences can be good :)
- Boolean profiles that were created as `Profile<Boolean>` implementors instead of `Windows` implementors can’t be downcast to `Windows` (just like any java object), but they CAN be converted using `(Windows) myBooleanProfile::stream`. For example, this is the _actual_ `not` method: 
  ```java 
  default Windows not() { 
    return mapValues($ -> !$)::stream; 
  }
  ```
  `mapValues` is a more general helper method on `Profile` that returns a `Profile` object. But since `Windows` doesn’t add any extra interface requirements beyond the stream method, the `::stream` method can be converted to the `Windows` functional sub-interface. (Same thing for LinearProfiles.)
- Some methods that are only used as generalized helper methods for more specific operations are exposed publicly because interface methods cannot be `protected` or `private`. (Such as `mapValues` above.) This is not necessarily a bad thing, because they are still valid operations, but it might be TMI.
- The process of fitting these objects into the AST and parser tree is hairy and non-trivial. I still haven’t figured out a satisfactory solution.

### Future Opportunities

- If we do segment streaming from simulation, this refactor would be one step toward streaming constraints so they could be pipelined together.
- If we want to make simulation optional for scheduling goals that don’t need it, this refactor could help by lazily triggering a re-sim only if a resource is referenced.

## Verification
I wrote a couple tests on the way, but haven’t done much in the way of running these profiles through the existing unit tests.

## Documentation
yikes


JoelCourtney:
As for #417, check out `Windows.shiftEdges`. I'm 95% sure its mathematically correct, and I'm 20% sure I could prove it in a fight 😬

Twisol:
Way to go! I'd really like to take some time to review this; do you need reviews in particularly soon?

JoelCourtney:
> do you need reviews in particularly soon?

No, this is low-ish priority and no short-term future work depends on it directly. So anytime before 1.1 is great (which is around the end of January).

JoelCourtney:
I have found it impossible to update the parsers to accommodate the generic type argument in `Profile<V>`. I bet its possible for someone a little smarter, but I'm not gonna shovel that work onto someone else because this wasn't a necessary feature anyway. I'm just gonna leave this branch to get buried in the repo, and we can resurrect it later if I have any genius ideas.

PR #495, opened 2022-11-21
Add standalone Aerie Docker images

### LINKED ISSUE #463: Create Docker images called aerie-postgres and aerie-hasura
Currently our docker-compose deployment of Aerie requires the use of host mounts that mount Hasura metadata and Postgres initialization SQL into the respective containers. Host mounts are not supported on serverless deployments like AWS fargate, and also require new users to either have the `deployment.zip` or the repo cloned with SQL distributed to startup Aerie.

This ticket should create `aerie-postgres` and `aerie-hasura` images that are a simple wrapper around these base images with the data included in the image. We will still need to provide the base `deployment.zip` for custom deployments.

After we have these new images we can create a docker-compose-standalone that does not require the .zip file for first-time users looking to run Aerie locally.

camargo:
* **Tickets addressed:** Resolves https://github.com/NASA-AMMOS/aerie/issues/463
* **Review:** By commit  

## Description

* Create Hasura docker image with Aerie data
* Create Postgres Docker image with Aerie data
* Used for quick-starting with Aerie

## Verification

N/A

## Documentation

README included.

## Future work

Add a `docker-compose-standalone.yml` to our deployment directory that uses the new images.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 *
 
 # Explicitly whitelist files relevant to image creation.
+!deployment
 !merlin-server/build/distributions/*.tar
 !merlin-worker/build/distributions/*.tar
 !scheduler-server/build/distributions/*.tar
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,31 @@
+# Docker
+
+This directory contains additional Dockerfiles for images built by Aerie.
+
+- [Dockerfile.hasura](./Dockerfile.hasura) - A Hasura Docker image with bundled Aerie-specific Hasura metadata
+- [Dockerfile.postgres](./Dockerfile.postgres) - A Postgres Docker image with bundled Aerie-specific SQL
+
+## Build
+
+First build Aerie to make sure the SQL files are properly added to the [deployment](../deployment/) directory:
+
+```sh
+cd aerie
+./gradlew assemble
+```
+
+Next, still from the top-level Aerie directory, build the images from the provided Dockerfiles:
+
+```sh
+docker build -t aerie-hasura -f ./docker/Dockerfile.hasura .
+docker build -t aerie-postgres -f ./docker/Dockerfile.postgres .
+```
+
+## Run
+
+To run the images you can use the following commands. Note these are just for testing purposes:
+
+```sh
+docker run --name aerie-hasura -d -p 8080:8080 aerie-hasura
+docker run --name aerie-postgres -d -p 5432:5432 --env-file ./.env aerie-postgres
+```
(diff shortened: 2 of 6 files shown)


docker/README.md
```diff
@@ -0,0 +1,31 @@
...
+Next, still from the top-level Aerie directory, build the images from the provided Dockerfiles:
+
+```sh
+docker build -t aerie-hasura -f ./docker/Dockerfile.hasura .
+docker build -t aerie-postgres -f ./docker/Dockerfile.postgres .
+```
+
+## Run
+
+To run the images you can use the following commands. Note these are just for testing purposes:
```
Mythicaeda, line 26: Regarding the "note these are just for testing purposes": Are the files in `docker` meant for us to use internally or are they meant to be used by users?

.dockerignore
```diff
@@ -2,6 +2,7 @@
 *
 
 # Explicitly whitelist files relevant to image creation.
+!deployment
```
Mythicaeda, line 5: Is `deployment` now being whitelisted because it's the source of the Hasura metadata/Postgres SQL files?

camargo: Yes exactly

docker/README.md
camargo: The resulting images are meant for users. The README here is more for developers looking at this directory.

JoelCourtney (approved):
seems legit

docker/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```sh
+docker run --name aerie-hasura -d -p 8080:8080 aerie-hasura
```
Mythicaeda, line 29: In our normal `docker-compose.yml`, Hasura uses some variables from the `.env`. Why doesn't it need that file in this statement?

camargo: Because those `run` commands are really just for testing the `Dockerfiles` build correctly (i.e. not for actual use with Aerie).

Mythicaeda: Also, when I run this command, I get a `Fatal Error: Either of --metadata-database-url or --database-url option expected` from the `aerie-hasura` container.

Mythicaeda: > Because those `run` commands are really just for testing the Dockerfiles build correctly (i.e. not for actual use with Aerie).

Gotcha. Why does the Postgres image actually need `.env` then? That was why the Hasura `run` missing it stood out.

camargo: That's fine. The main thing you can test if you want is `docker exec -it [CONTAINER_ID] /bin/sh` to make sure the container has the proper metadata files. Don't worry too much about this tho - the main contribution of this commit is the images resulting from the publish workflow.

camargo: > Why does the Postgres image actually need .env then? 

Because that container will not even start without it

Mythicaeda (approved):

PR #497, opened 2022-11-22
Provide Markdown Syntax in Documentation Examples

Mythicaeda:
* **Review:** By commit  

## Description

* How to build a local preview of the site is now on the index page of `contribute/documentation`, as well as a link to how to test changes.
* Modified all the pages in `Examples` so it now has examples on some of the available markup for both Markdown and reStructuredText
* Removed the `PanelBox` directive, which was a custom Scylla directive that wasn't visually distinct from a bulleted list.
* Minor language changes throughout `contribute/documentation`

## Verification

* `make clean ; make dirhtml-ext` throws no errors
* `make clean ; make dirhtml-ext ; make linkcheck` throws only the expected errors
* `make clean ; make test` throws no errors.

## Future Work
There are some entries on `aerie-glossary` that need to be filled in, especially in the `Aerie Client User Interface` section.


docs/source/aerie-glossary.md
```diff
@@ -1,98 +1,137 @@
...
 
-**Argument**: a value given to a function or assigned to an Activity Parameter.
+Argument
+  A value given to a function or assigned to an Activity Parameter.
 
-**Call**: is a function in the Merlin simulation modeling API. Call() executes a provided task and blocks until that task has completed. Typically Call() is used by a mission modeler who wants to execute an activity and wait for it's completion before continuing execution of their modeling code. Call() has a few function signatures so that it can be used to call an activity type and arguments or a Java lambda/Runnable.
+Call
+  A function in the Merlin simulation modeling API. Call() executes a provided task and blocks until that task has completed. Typically Call() is used by a mission modeler who wants to execute an activity and wait for it's completion before continuing execution of their modeling code. Call() has a few function signatures so that it can be used to call an activity type and arguments or a Java lambda/Runnable.
```
mattdailis: Tiny typo that caught my eye:
```suggestion
  A function in the Merlin simulation modeling API. Call() executes a provided task and blocks until that task has completed. Typically Call() is used by a mission modeler who wants to execute an activity and wait for its completion before continuing execution of their modeling code. Call() has a few function signatures so that it can be used to call an activity type and arguments or a Java lambda/Runnable.
```

docs/sphinx_scylladb_theme/__init__.py
```diff
@@ -11,7 +11,6 @@
     hero_box,
     multiversion,
     navigation,
-    panel_box,
```
mattdailis, line 14: Curious about this removal (not concerned, but I didn't remember it being called out in the PR description). Could you briefly elaborate?

docs/source/contribute/documentation/examples/tabs.md
```diff
@@ -0,0 +1,224 @@
...
+    
+    sed -i -e 's/README/index/g' index.rst
+
+  .. code-tab:: shell Windows (Powershell)
+    
+    (Get-Content index.rst) -replace 'Hello', 'World' | Out-File -encoding ASCII index.rst
+
+```
+
+`````{code-tab} md
```
mattdailis, line 189: Just an observation - it seems like the tabs are coupled to each other (i.e. selecting markdown on one tab changes all the tabs on the page to be markdown). That's not necessarily bad, but I found it surprising.

https://user-images.githubusercontent.com/1189602/203616475-e8a1292c-734d-4195-a289-26b394eb6094.mov

mattdailis (approved):
Approving, because none of my comments should block merge

docs/source/contribute/documentation/examples/tabs.md
Mythicaeda: Yup, that's intentional behavior! `code-tab`s are a special type of `group-tab` that only contains a code-block. [There is the `tab` option as well](https://sphinx-tabs.readthedocs.io/en/latest/#basic-tabs), but I couldn't think of when we'd want to use that instead of `group-tab` (since we generally use `tabs` for API vs UI or RST vs MD documentation, where a user's selection ought to carry throughout the page), so I excluded mentioning it.

docs/sphinx_scylladb_theme/__init__.py
Mythicaeda: Ah, I can add it to the PR description, but basically this was a custom Scylla component that wasn't visually distinct from a bulleted list.

docs/source/aerie-glossary.md
Mythicaeda: Fixed, thanks!

PR #503, opened 2022-11-29
Bulk SeqJSON Actions

### LINKED ISSUE #489: Create getSequenceSeqJsons action to generate seqJsons in bulk
Currently the [getSequenceSeqJson action](https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/hasura/metadata/actions.yaml#L42) ([endpoint in sequencing server](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/src/app.ts#L542)) takes only a single `seqId` and returns a single `seqJson`. Our users currently call this API in a loop and automate the generation of `seqJson` for many sequences (currently 50-100 per plan). We need a bulk version of this API that allows for passing a list of `seqIds` to return a list of generated `seqJsons`.

For API consistency we should also do this for the [getUserSequenceSeqJson action](https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/hasura/metadata/actions.yaml#L46).

dyst5422:
* **Tickets addressed:** Resolves #489
* **Review:** By commit  

## Description

Added new queries for the bulk behavior and deprecated the singular endpoints (for lower long term maintenance)

Added `getSequenceSeqJsonBulk` for Sequence based on seqId and simulation dataset id
Added `getUserSequenceSeqJsonBulk` for Sequence from EDSL body
Added `getEdslForSeqJsonBulk` for EDSL body from seqJson 

A slight change had to be made to the massive SQL that drives the plan expansion so that it would pick out data based on pairs of seqid + simulation dataset id. `node-postgres` doesn't natively support `in tuple` queries (as is necessary for this operation) so we had to add pg-format to do the input cleansing and formatting into the required tuple format in the string template (I hate that it doesn't go through the same parameterization mechanisms, but a bunch of googling led me to this as the best solution).

## Verification

Added tests for the bulk behavior

## Documentation

Need to update the documentation for sequence generation. Specifically, API calls to `getEdslForSeqJson`, `getUserSequenceSeqJson`. `getSequenceSeqJson`

## Future work

Remove the deprecated code after a few releases.


sequencing-server/src/app.ts
```diff
@@ -662,14 +718,180 @@
...
+      return {
+        status: FallibleStatus.FAILURE,
+        seqJson: null,
+        errors: [promise.reason],
+      };
+    }
+  }));
+
+  return next();
+});
```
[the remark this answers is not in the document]
dyst5422: Pretty sure we intend to rate-limit at the Hasura level rather than the individual server level

dyst5422:
rebased

dyst5422:
@camargo @mattdailis Who is best to review this? I assume @goetzrrGit  is still out on paternity. (Just trying to ensure I follow through on any support to reviewing this etc to get it merged in)

camargo:
@dyst5422 I will take a look. Going forward @cohansen is onboarding and will be the new main owner of the sequencing service and will provide review as well.

deployment/hasura/metadata/actions.graphql
```diff
@@ -144,14 +144,36 @@ type Query {
   getSequenceSeqJson(
     seqId: String!
     simulationDatasetId: Int!
-  ): GetSeqJsonResponse!
+  ): GetSeqJsonResponse! @deprecated(reason: "Use bulk query getSequenceSeqJsonBulk instead")
```
camargo, line 147: These `@deprecated` fields seem to be ignored by Hasura right now :(
Good to have for documentation purposes tho.

sequencing-server/src/app.ts
```diff
@@ -505,7 +497,7 @@ app.post('/get-seqjson-for-sequence-standalone', async (req, res, next) => {
     const context: Context = res.locals['context'];
     commandTypes = await context.commandTypescriptDataLoader.load({ dictionaryId: commandDictionaryID });
   } catch (e) {
-    res.status(200).json({
+    res.status(500).json({
```
camargo, line 500: There seems to be a mix of returning 200s and 500s in this file when catching. Was this 200 here for a reason before?

camargo (approved):
Tested locally and the new actions work. LGTM thanks for the updates!

deployment/hasura/metadata/actions.graphql
dyst5422: womp womp 😢

sequencing-server/src/app.ts
dyst5422: Good point! I think we can rework the this error mode in general. Right now, we catch all errors from the dataloader, which includes command dictionary not found errors as well as network errors. The command dictionary not found errors fall along the lines of the "business logic driven errors" that should be reflected in fallable error returns (which use 200). The network errors we want to just throw and be auto handled by hasura.

So we have a CommandDictionaryNotFound error that should be a 200 on the server and other stuff which should be a 500.

Now a complicating element here is that the command dictionary not found error is NOT the error type we specify in the graphql api [here](https://github.com/NASA-AMMOS/aerie/blob/feature/489--bulk-seqJson/deployment/hasura/metadata/actions.graphql#L181) so ideally we'd have a union error type for the API where we include a CommandDictionaryNotFound error type (or a generic error type) that at a minimum does not contain a code location as the UserCodeError type does as its definitely not relevant to the error of CommandDictionaryNotFound. I don't think Hasura allows user created union types, so we can't currently do that.

Because of this, I think it's best to return as a 500 so that it circumvents the shape represented in our graphql api.

We should probably add failure mode test cases at some point (probably a good intro task for onboarding).

cohansen (approved):
I'm still learning the code base but it looks good!

PR #507, opened 2022-11-30
Spans based scheduling basic AST

### LINKED ISSUE #466: Implement backend representation for Spans-based scheduling AST

JoelCourtney:
* **Tickets addressed:** closes #466 
* **Review:** By commit  

## Description

This adds the basic Java data structure that represents a Spans-based scheduling goal. It also adds a `SpansAlias` node to the constraints AST that accesses a string alias for a spans object in the EvaluationEnvironment, similar to how we handle activity aliases.

This PR does not include any behavior, including getting conflicts or parsing. So `SpansGoal` does not yet inherit from Goal because there is nothing to put in getConflicts. SpansGoal can be added to the inheritance tree when its behavior is implemented.

## Verification
SpansAlias got a basic test, but that's it. SpansGoal doesn't yet have behavior to test.

## Future work
yes.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
@@ -31,6 +31,7 @@ public final class ForEachActivitySpans implements Expression<Spans> {
       if (activity.type.equals(this.activityType)) {
         final var newEnvironment = new EvaluationEnvironment(
             new HashMap<>(environment.activityInstances()),
+            environment.spansInstances(),
             environment.realExternalProfiles(),
             environment.discreteExternalProfiles()
         );
@@ -61,8 +62,7 @@ public final class ForEachActivitySpans implements Expression<Spans> {
 
   @Override
   public boolean equals(Object obj) {
-    if (!(obj instanceof ForEachActivitySpans)) return false;
-    final var o = (ForEachActivitySpans)obj;
+    if (!(obj instanceof final ForEachActivitySpans o)) return false;
 
     return Objects.equals(this.activityType, o.activityType) &&
            Objects.equals(this.alias, o.alias) &&
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
@@ -1,0 +1,31 @@
+package gov.nasa.jpl.aerie.constraints.tree;
+
+import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.Segment;
+import gov.nasa.jpl.aerie.constraints.time.Spans;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+
+import java.util.Objects;
+import java.util.Set;
+
+public record SpansAlias(String spansAlias) implements Expression<Spans> {
+
+  @Override
+  public Spans evaluate(final SimulationResults results, final Interval bounds, final EvaluationEnvironment environment) {
+    return environment.spansInstances().get(this.spansAlias);
+  }
+
+  @Override
+  public void extractResources(final Set<String> names) {}
+
+  @Override
+  public String prettyPrint(final String prefix) {
+    return String.format(
+        "\n%s(spansAlias %s)",
+        prefix,
+        this.spansAlias
+    );
+  }
+}
--- a/constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
+++ b/constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
@@ -461,7 +461,7 @@ public class ASTTests {
         Map.of(),
         Map.of()
     );
-    final var environment = new EvaluationEnvironment(Map.of("act", act), Map.of(), Map.of());
+    final var environment = new EvaluationEnvironment(Map.of("act", act), Map.of(), Map.of(), Map.of());
 
     final var result = new RealParameter("act", "p1").evaluate(simResults, environment);
 
@@ -726,6 +726,7 @@ public class ASTTests {
                 Interval.between(4, 8, SECONDS))
         ),
         Map.of(),
+        Map.of(),
         Map.of()
     );
 
@@ -756,6 +757,7 @@ public class ASTTests {
                 Map.of(),
                 Interval.between(4, 8, SECONDS))
         ),
+        Map.of(),
         Map.of(),
         Map.of()
     );
@@ -788,6 +790,7 @@ public class ASTTests {
                 Map.of(),
                 Interval.between(4, 8, SECONDS))
         ),
+        Map.of(),
         Map.of(),
         Map.of()
     );
@@ -879,6 +882,7 @@ public class ASTTests {
 
     final var environment = new EvaluationEnvironment(
         Map.of(),
+        Map.of(),
         Map.of(
             "real1", new LinearProfile(Segment.of(Interval.at(1, SECONDS), new LinearEquation(Duration.of(1, SECONDS), 0, 1))),
             "real2", new LinearProfile(Segment.of(Interval.at(2, SECONDS), new LinearEquation(Duration.of(2, SECONDS), 0, 1))),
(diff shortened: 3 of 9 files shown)


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/SpansGoal.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public Builder allowReuse(final boolean allowReuse) {
+      this.allowReuse = allowReuse;
+      return this;
+    }
+
+    public SpansGoal build() {
+      if (spansExpression == null) {
+        throw new IllegalArgumentException(
+            "creating spans goal requires non-null \"spansExpression\"");
```
Mythicaeda: Very minor, but can `creating` be changed to `Creating` in the errors?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
```diff
@@ -61,8 +62,7 @@ public String prettyPrint(final String prefix) {
 
   @Override
   public boolean equals(Object obj) {
-    if (!(obj instanceof ForEachActivitySpans)) return false;
-    final var o = (ForEachActivitySpans)obj;
+    if (!(obj instanceof final ForEachActivitySpans o)) return false;
```
Mythicaeda, line 65: Could you clarify what is happening in the `if` statement here?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
```diff
@@ -0,0 +1,31 @@
...
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.Segment;
+import gov.nasa.jpl.aerie.constraints.time.Spans;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+
+import java.util.Objects;
+import java.util.Set;
+
+public record SpansAlias(String spansAlias) implements Expression<Spans> {
```
Mythicaeda, line 13: Based on the test added, am I correct in assuming that SpansAlias is a class for referring to a span by a name?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Override
+  public Spans evaluate(final SimulationResults results, final Interval bounds, final EvaluationEnvironment environment) {
+    return environment.spansInstances().get(this.spansAlias);
+  }
+
+  @Override
+  public void extractResources(final Set<String> names) {}
```
Mythicaeda, line 21: Why is this method empty?

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
```diff
@@ -923,6 +923,31 @@ public void testExternalRealResource() {
...
+    );
+
+    final var environment = new EvaluationEnvironment(
+        Map.of(),
+        Map.of("my spans", spans),
+        Map.of(),
+        Map.of()
+    );
+
+    final var result = new SpansAlias("my spans").evaluate(simResults, environment);
```
Mythicaeda, line 956: I notice that `SpansAlias.evaluate` has three parameters, but its called here with only two. How are we able to do that?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
JoelCourtney: It doesn't depend on any resources. (other nodes that do modify the names set)

constraints/src/test/java/gov/nasa/jpl/aerie/constraints/tree/ASTTests.java
JoelCourtney: We have default overloads for `Expression.evaluate` in the Expression interface.

Mythicaeda: Ah, missed that, thank you!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
JoelCourtney: This is a non-behavioral change. Its syntactic sugar for context-aware downcasting.

```java
if (obj instanceof MySubclass) {
    final MySubclass obj2 = (MySubclass) obj;
    // here you have a downcasted binding
}
// here you don't

// is the same as

if (obj instanceof final MySubclass obj2) {
    // here you have a downcasted binding
}
// here you don't
```

Its even funkier if you check that it is *not* an instanceof the subclass, *AND* the if statement always breaks/returns/throws to disrupt control flow:

```java
if (!(obj instanceof MySubclass)) {
    // here you don't have a downcasted binding
    return; // necessary
}
final MySubclass obj2 = (MySubclass) obj;
// here you do

// is the same as

if (!(obj instance of final MySubclass obj2)) {
    // here you don't have a downcasted binding
    return; // necessary
}
// here you do
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
Mythicaeda: Can Spans in general potentially modify resources? If so, in order to extract them, would I need to call something like `SpanAlias("My Span").evaluate(mySimResults, myEnviornment).extractResources`?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
Mythicaeda: Thanks for the explanation on context-aware downcasting!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
JoelCourtney: I'm not really sure what you're saying here.

> Can Spans in general potentially modify resources?

No, nothing in the constraints package can modify resources. Some operations can transform profiles, but the resources in simulation results are never changed.

> SpanAlias("My Span").evaluate(mySimResults, myEnviornment).extractResources

`extractResources` is a method on the expression interface. Its used in the constraints hasura action to associate the resources that an expression used with the violations it produced.

JoelCourtney: I guess I don't know how familiar you are with constraints evaluation. Let me know if I'm not being very clear and I can give you a crash course on the package if you want.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
JoelCourtney: Yes. The constraints package is essentially a very (very) basic unoptimized interpreter for the constraints AST. Unlike a basic Javascript or Python interpreter, we don't have source code that we need to lex and parse into the AST; the constraints typescript API generates the AST directly. The `EvaluationEnvironment` record is how we store variables. The environment comes preloaded with resource bindings from simulation results - sort of like auto-imported global variables like `console` in JS. Then during evaluation we can define more variables like activity aliases in `ForEachActivity`.

In this case, if a spans expression is being evaluated in the context of a Spans based scheduling goal, the `getConflicts` method will preload the environment with both the simulation results and a spans alias, which can be queried using the `SpansAlias` expression AST node.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansAlias.java
Mythicaeda: I think the question I meant to ask was "Can Spans potentially *depend* on resources?"

Either way, I'll take that crash course 🙂.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/SpansGoal.java
JoelCourtney: done

Mythicaeda (approved):

PR #508, opened 2022-12-01
Added the path param to schema types

### LINKED ISSUE #481: Command expansion fails for a plan with an activity that contains a 'path' parameter
When trying to run command expansion on a plan that contains an activity with a `path` typed parameter the expansion fails. Additionally it fails even if the activity with the `path` parameter is not used in the expansion (i.e. does not have an associated sequence, or expansion rule/set).

## Repro Steps

1. Create a plan from Banananation
2. Add a PeelBanana activity from Banananation
3. Add a [LineCount](https://github.com/NASA-AMMOS/aerie/blob/develop/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/LineCountBananaActivity.java#L17) activity from Banananation
4. Add a command dictionary
5. Create an expansion rule against PeelBanana
6. Create an expansion set with the PeelBanana rule from (5)
7. Create a sequence
8. Assign the sequence from (7) to the PeelBanana activity
9. Run expansion
10. You should see an error

## Error

The following error is logged in the `aerie-sequencing` server:

```
ERROR app - Unknown schema type: path
Error: Unknown schema type: path
    at convertType (file:///app/build/lib/batchLoaders/simulatedActivityBatchLoader.js:138:19)
    at file:///app/build/lib/batchLoaders/simulatedActivityBatchLoader.js:98:32
    at Array.reduce (<anonymous>)
    at mapGraphQLActivityInstance (file:///app/build/lib/batchLoaders/simulatedActivityBatchLoader.js:95:78)
    at file:///app/build/lib/batchLoaders/simulatedActivityBatchLoader.js:37:81
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 11)
    at async Promise.all (index 0)
```

## Possible Solution

Add the `path` parameter schema type to the [convertType function](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts#L218) in the `sequencing-server`. Additionally consider reviewing and adding any additional missing `SchemaTypes`.

cohansen:
* **Review:** By commit

## Description
Closes #481. The `path` type was missing from the Schema types, so it was added.

## Verification
Tested manually by adding and removing activities that had a path parameter and then running the expansion.

### DIFF
--- a/sequencing-server/src/lib/batchLoaders/activitySchemaBatchLoader.ts
+++ b/sequencing-server/src/lib/batchLoaders/activitySchemaBatchLoader.ts
@@ -57,6 +57,7 @@ export enum SchemaTypes {
   Series = 'series',
   Struct = 'struct',
   Variant = 'variant',
+  Path = 'path',
 }
 
 export type Schema =
@@ -67,7 +68,8 @@ export type Schema =
   | StringSchema
   | SeriesSchema
   | StructSchema
-  | VariantSchema;
+  | VariantSchema
+  | PathSchema;
 
 interface BaseSchema<T extends SchemaTypes | unknown> {
   type: T;
@@ -78,6 +80,7 @@ type BooleanSchema = BaseSchema<SchemaTypes.Boolean>;
 type IntSchema = BaseSchema<SchemaTypes.Int>;
 type RealSchema = BaseSchema<SchemaTypes.Real>;
 type DurationSchema = BaseSchema<SchemaTypes.Duration>;
+type PathSchema = BaseSchema<SchemaTypes.Path>;
 
 interface SeriesSchema extends BaseSchema<SchemaTypes.Series> {
   items: Schema;
--- a/sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
+++ b/sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
@@ -4,6 +4,7 @@ import { ErrorWithStatusCode } from '../../utils/ErrorWithStatusCode.js';
 import parse from 'postgres-interval';
 import { GraphQLActivitySchema, Schema, SchemaTypes } from './activitySchemaBatchLoader.js';
 import type { activitySchemaBatchLoader } from './activitySchemaBatchLoader.js';
+import { assertUnreachable } from '../../utils/assertions.js';
 
 export const simulatedActivitiesBatchLoader: BatchLoader<
   { simulationDatasetId: number },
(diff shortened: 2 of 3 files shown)


sequencing-server/src/lib/batchLoaders/activitySchemaBatchLoader.ts
```diff
@@ -14,13 +14,13 @@ export const activitySchemaBatchLoader: BatchLoader<
       activity_type(where: {
         _or: [
           ${keys
-            .map(
-              key => `{
+      .map(
```
camargo: Was this format changed for a reason?

cohansen: I think it got hit by prettier, I didn't touch it manually.

camargo: Is it using [this prettier file](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/.prettierrc)?

cohansen: Ah, it was defaulting to using the language formatting. I fixed it in my IDE and it reverted the change.

camargo (approved):
LGTM!

sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
```diff
@@ -228,6 +228,8 @@ function convertType(value: any, schema: Schema): any {
       return Temporal.Duration.from(parse(value).toISOString());
     case SchemaTypes.Boolean:
       return value;
+    case SchemaTypes.Path:
```
Mythicaeda, line 232: Not blocking (and technically not this PR), but is there a reason we don't use fall-through/`case A || B` syntax for the Real, Boolean, Path, and String cases? It's fine if the answer's style, it just stood out that we have four cases that all are the same behavior.

Mythicaeda (approved):
LGTM, although you mentioned that you did your testing manually. Is it possible to add a quick test for this in the e2e tests?

sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
cohansen: I was going to modify this, but with the `assertUnreachable` it actually breaks if we go that route now because it has parsing that all the cases have been hit.

PR #511, opened 2022-12-02
Propagate child return value to parent on call()

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR changes the `call()` simulation action to return the result of the called task, rather than `void`. This ends a line of work that started with the removal of the `waitFor(String)` method, which ensured that we wouldn't need to hold on to all task return values forever just in case somebody calls `waitFor` later.

## Verification
- [ ] I still have to test this 🤣 I wanted to get the PR up so that I can accommodate any feedback as soon as possible.

## Documentation
* I need to scan the documentation to see if anything needs to change. Probably an *example* of using the return value should be added; I don't think this invalidates any existing documentation though.

## Future work
* I'd like to discuss whether the current behavior of `call`, which resumes the caller one tick *after* the child completes, is actually the desired behavior. I'll put more notes in a separate Discussion topic.


Twisol:
FYI, this:

> [Twisol](https://github.com/Twisol) [had a problem deploying](https://github.com/NASA-AMMOS/aerie/actions/runs/3641538513/jobs/6147592494)

should just be because I haven't recompiled the aerie-lander JAR.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/Task.java
```diff
@@ -20,7 +25,174 @@
...
+   * @param <A>
+   *   The input type of the given function.
+   * @param <B>
+   *   The output type of the given function.
+   * @param step
+   *   The pure function to perform as a task.
+   * @return
+   *   A task performing the given pure function.
+   */
+  // LAW: compose(lift(f), lift(g)) === lift(f.andThen(g))
```
mattdailis, line 101: If I'm following this correctly, the intent behind this comment was:
```suggestion
  // LAW: compose(lift(f), lift(g)) === lift(f).andThen(lift(g)))
```

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
```diff
@@ -1,176 +1,249 @@
...
-      if (request instanceof TaskRequest.Resume resumeRequest) {
+      if (request instanceof Request.Resume<Midput> r) {
         // We've been told to continue executing.
-        return resumeRequest.scheduler;
-      } else if (request instanceof TaskRequest.Abort) {
+        return r;
+      } else if (request instanceof Request.Abort) {
         // We've been told to bail out and release this thread ASAP.
         //
         // We'll throw an exception to get as far up and out of the task as we can.
```
mattdailis, line 249: Later in this comment - we say "We'll just keep throwing whenever `yield()` is called" - which I think is slightly inaccurate - we'll throw whenever any method is called on `TaskContext`.

In fact, I think the TODO is satisfied now, no?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ReplayingTask.java
```diff
@@ -1,71 +1,227 @@
...
     this.rootContext = Objects.requireNonNull(rootContext);
     this.task = Objects.requireNonNull(task);
   }
 
   @Override
-  public TaskStatus<Return> step(final Scheduler scheduler) {
-    final var handle = new ReplayingTaskHandle();
-    final var context = new ReplayingReactionContext(this.rootContext, this.memory, scheduler, handle);
+  public TaskStatus<Output> step(final Scheduler scheduler, final Input input) {
+    // Use locals to avoid holding a reference on `this` from the thread's lambda.
```
mattdailis, line 30: In this particular case, there are no new threads going on - so perhaps this comment could be
```suggestion
    // Use locals to avoid holding a reference on `this` from the lambda.
```

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ReplayingTask.java
```diff
@@ -1,71 +1,227 @@
...
+        final History history,
+        final Scheduler scheduler,
+        final Function<Context, Output> task
+    ) {
+      history.rewind();
+      final var context = new TaskContext<>(task, scheduler, history);
+      try {
+        return TaskStatus.completed(task.apply(context));
+      } catch (final Yield ignored) {
+        // TODO: Extract the new TaskStatus from `context`.
```
mattdailis, line 148: Does this TODO still apply?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ReplayingTask.java
```diff
@@ -1,71 +1,227 @@
...
-    private Scheduler yield(final TaskStatus<Return> status) {
-      this.status = status;
-      throw Yield;
+    public static <Output>
+    TaskStatus<Output> replay(
+        final History history,
+        final Scheduler scheduler,
+        final Function<Context, Output> task
+    ) {
+      history.rewind();
```
mattdailis, line 143: I'm noticing that there's no way to reset `isYielding` back to `false` once it's been set to `true`. Should we reset `isYielding` when we're about to replay?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -779,39 +804,44 @@ static ConditionJobId forCondition(final ConditionId condition) {
   }
 
   /** The lifecycle stages every task passes through. */
-  private sealed interface ExecutionState<Return> {
-    /** The task is in its primary operational phase. */
-    record InProgress<Return>(Duration startOffset, Task<Return> state)
-        implements ExecutionState<Return>
+  private sealed interface ExecutionState<Output> {
+    record InProgress<Input, Output>(
```
mattdailis, line 811: Any thoughts on the pros/cons of an "InProgress" with a mutable input, vs an explicit "Blocked" state?

If we do go with the MutableObject - it might be helpful to spell out the invariants in a comment.

Taking a quick stab at the invariants I'm inferring - the contents of input will be immediately `Unit` for all tasks whose input is `Unit` - and for tasks with other types of input, the contents will be initially `null`, and then must be updated to the correct value before signaling the task?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -214,62 +219,78 @@ private <Return> void stepTaskHelper(
...
+    final var blocked = progress.blockWith(s.continuation());
+    this.tasks.put(task, blocked);
+    this.waitingTasks.subscribeQuery(task, Set.of(SignalId.forTask(child)));
+
+    // Ready to run (with the given Input); will produce Midput
+    this.tasks.put(child, new ExecutionState.InProgress<>(
+        currentTime,
+        s.child().create(this.executor),
+        new MutableObject<>(s.input()),
+        blocked.input()::setValue));
```
mattdailis, line 287: I keep looking for a cast somewhere - because when I prototyped this myself about a year ago, I needed to cast the child's return value to the parent's expectation. However, I think you've managed to use generics to guarantee that the types line up. Nice! 🚀

mattdailis (commented):
Here's my first-pass review - most of my comments are about comments. Once we're happy with the new semantics in this PR, we should take another pass and evaluate whether it's worth kick-starting some protocol versioning here.

<hr />
FYI - commit "Propagate a called task's final output to its parent" doesn't compile due to java not being able to infer types in ThreadedTask::call:

```diff
-final var request = this.<Unit>yield($ -> TaskStatus.calling(child, $));
+final var request = this.<Midput>yield($ -> TaskStatus.calling(child, $));
```
The following commit includes the fix: "Move the Arguments parameter of getTaskFactory into the returned factory"

camargo:
What would the upgrade guide look like for this?

camargo:
Closing for now, please reopen when you think we can get this in

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/Task.java
Twisol: No, the idea is that compose-then-lift should behave the same as lift-then-compose. (In Haskell-y syntax, `fmap g . fmap f == fmap (g . f)`.)

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ThreadedTask.java
Twisol: > Later in this comment - we say "We'll just keep throwing whenever `yield()` is called" - which I think is slightly inaccurate - we'll throw whenever any method is called on `TaskContext`.

True! In a previous life, the `if (this.isAborting)` check was (more or less) centralized to `yield`, and this comment made more sense.

Twisol: > In fact, I think the TODO is satisfied now, no?

Which TODO? I don't see a TODO in this vicinity -- are you referring to the one you've commented on elsewhere?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ReplayingTask.java
Twisol: Uh, hmm. No?

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ReplayingTask.java
Twisol: I don't think so. Since we create a new `TaskContext` every time we step the task, we're going to throw away the one that has `isYielding` stuck true anyway.

Issue #501, opened 2022-11-17
Add Temporal.Duration in Constraints eDSL


camargo:
The scheduling and command expansion eDSLs use `Temporal.Duration` to represent durations. The constraints language should as well. Since the scheduling eDSL includes the constraints eDSL, perhaps we simply move the `Temporal.Duration` typing file into the constraints eDSL.

PR #520, opened 2022-12-06
Fix serialization of Duration to avoid rounding to millisecond

### LINKED ISSUE #513: Profile segments with a duration of 1 microsecond cause simulation results to fail to save
## Steps to reproduce
1. Modify BiteBanana's effect model in the following way:

```diff
public ComputedAttributes run(final Mission mission) {
    final var bigBiteSize = biteSize > 1.0;
    final var newFlag = bigBiteSize ? Flag.B : Flag.A;
    mission.flag.set(newFlag);
-    mission.fruit.subtract(biteSize);
+    mission.fruit.subtract(biteSize / 2);
+    delay(1, MICROSECOND);
+    mission.fruit.subtract(biteSize / 2);
    return new ComputedAttributes(bigBiteSize, newFlag);
  }
```

2. Compile, upload mission model, make a plan
3. Add BiteBanana to the plan and simulate

This gets the following error:
> duplicate key value violates unique constraint "profile_segment_19_dataset_id_profile_id_start_offset_key"

Full stack trace from the merlin worker:
```
gov.nasa.jpl.aerie.merlin.server.remotes.postgres.DatabaseException: Failed to store simulation results
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresResultsCellRepository$PostgresResultsCell.succeedWith(PostgresResultsCellRepository.java:622)
	at gov.nasa.jpl.aerie.merlin.server.services.SynchronousSimulationAgent.simulate(SynchronousSimulationAgent.java:98)
	at gov.nasa.jpl.aerie.merlin.worker.MerlinWorkerAppDriver.main(MerlinWorkerAppDriver.java:81)
Caused by: java.sql.BatchUpdateException
	at com.impossibl.postgres.jdbc.IntegerBatchResults.getException(BatchResults.java:85)
	at com.impossibl.postgres.jdbc.PGPreparedStatement.executeBatch(PGPreparedStatement.java:502)
	at com.impossibl.postgres.jdbc.PGPreparedStatement.executeBatch(PGPreparedStatement.java:417)
	at com.zaxxer.hikari.pool.ProxyStatement.executeBatch(ProxyStatement.java:127)
	at com.zaxxer.hikari.pool.HikariProxyPreparedStatement.executeBatch(HikariProxyPreparedStatement.java)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostProfileSegmentsAction.apply(PostProfileSegmentsAction.java:60)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.ProfileRepository.postRealProfileSegments(ProfileRepository.java:159)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.ProfileRepository.postProfileSegments(ProfileRepository.java:137)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.ProfileRepository.postResourceProfiles(ProfileRepository.java:117)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresResultsCellRepository.postSimulationResults(PostgresResultsCellRepository.java:459)
	at gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresResultsCellRepository$PostgresResultsCell.succeedWith(PostgresResultsCellRepository.java:619)
	... 2 more
Caused by: com.impossibl.postgres.jdbc.PGSQLIntegrityConstraintViolationException: duplicate key value violates unique constraint "profile_segment_19_dataset_id_profile_id_start_offset_key"
	at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:192)
	at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:174)
	at com.impossibl.postgres.jdbc.ErrorUtils.makeSQLException(ErrorUtils.java:161)
	at com.impossibl.postgres.jdbc.PGDirectConnection.execute(PGDirectConnection.java:514)
	at com.impossibl.postgres.jdbc.PGDirectConnection.execute(PGDirectConnection.java:466)
	at com.impossibl.postgres.jdbc.PGPreparedStatement.finishRequests(PGPreparedStatement.java:562)
	at com.impossibl.postgres.jdbc.PGPreparedStatement.executeBatch(PGPreparedStatement.java:499)
	... 11 more
```

Postgres container logs:
```
postgres_1                  | 2022-12-05 15:23:21.777 UTC [43] ERROR:  duplicate key value violates unique constraint "profile_segment_19_dataset_id_profile_id_start_offset_key"
postgres_1                  | 2022-12-05 15:23:21.777 UTC [43] DETAIL:  Key (dataset_id, profile_id, start_offset)=(19, 6114, 39:38:15) already exists.
postgres_1                  | 2022-12-05 15:23:21.777 UTC [43] STATEMENT:  insert into profile_segment (dataset_id, profile_id, start_offset, dynamics, is_gap)
postgres_1                  |     values ($1, $2, $3::interval, $4::json, $5)
```

mattdailis:
* **Tickets addressed:** Fixes #513 
* **Review:** By commit  

## Description

This PR is in response to a bug found in #513 where profile segments shorter than one millisecond would cause writing simulation results to the database to fail.

The root cause was found to be in the serialization logic from Durations to iso8601 strings:
https://github.com/NASA-AMMOS/aerie/blob/59f3fcde69a88cb6f413b3995efbce7474c8432b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java#L500-L502

This rounds Durations from microseconds to milliseconds - causing a collision when the start offsets of two profile segments fall within the same millisecond.

The first commit updates this logic, as well as the corresponding deserialization logic, to use nanoseconds instead of milliseconds.

The subsequent commits move the (de)serialization logic out of the Duration class - serialization is not be one of its concerns. (I should have brought this up originally in #187, but it did not catch my eye).

#### Before making a fix, I reproduced the issue as follows:

I modified BiteBanana in the following way:

```diff
+@Parameter
+public int delayMicros = 1;
@EffectModel
  public ComputedAttributes run(final Mission mission) {
    final var bigBiteSize = biteSize > 1.0;
    final var newFlag = bigBiteSize ? Flag.B : Flag.A;
    mission.flag.set(newFlag);
-    mission.fruit.subtract(biteSize);
+    mission.fruit.subtract(biteSize / 2);
+    delay(delayMicros, MICROSECONDS);
+    mission.fruit.subtract(biteSize / 2);
    return new ComputedAttributes(bigBiteSize, newFlag);
  }
```

I ran a local deployment of Aerie, uploaded my modified banananation, and added a single BiteBanana to the plan. Running simulation triggered the error described in #513 .

#### After making the fix...
I assembled the merlin worker, rebuilt the docker image, re-created the docker container, invalidated the simulation results by modifying a parameter, and ran simulation again. This time, no error occurred, and a visual inspection of the profile segments using IntelliJ's database viewer showed that they had the correct start offsets.

If anyone has thoughts on how to make this an automated test, I'm all ears.

## Documentation

No documentation needs to be updated for this bugfix... except possibly putting it on a list of known bugs for Aerie 1.0.0.

## Future work

None

### DIFF
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
@@ -1,5 +1,6 @@
 package gov.nasa.jpl.aerie.merlin.protocol.types;
 
+import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.List;
 
@@ -494,11 +495,12 @@ public final class Duration implements Comparable<Duration> {
   }
 
   public static Duration fromISO8601String(final String iso8601String){
-    return of(java.time.Duration.parse(iso8601String).toMillis(), MILLISECOND);
+    final var javaDuration = java.time.Duration.parse(iso8601String);
+    return of((javaDuration.getSeconds() * 1_000_000L) + (javaDuration.getNano() / 1000L), MICROSECONDS);
   }
 
   public String toISO8601String() {
-    return java.time.Duration.ofMillis(dividedBy(Duration.MILLISECONDS)).toString();
+    return java.time.Duration.of(this.durationInMicroseconds, ChronoUnit.MICROS).toString();
   }
 
   @Override
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
@@ -1,6 +1,5 @@
 package gov.nasa.jpl.aerie.merlin.protocol.types;
 
-import java.time.temporal.ChronoUnit;
 import java.util.Collections;
 import java.util.List;
 
@@ -497,10 +496,6 @@ public final class Duration implements Comparable<Duration> {
   public static Duration fromISO8601String(final String iso8601String){
     final var javaDuration = java.time.Duration.parse(iso8601String);
     return of((javaDuration.getSeconds() * 1_000_000L) + (javaDuration.getNano() / 1000L), MICROSECONDS);
-  }
-
-  public String toISO8601String() {
-    return java.time.Duration.of(this.durationInMicroseconds, ChronoUnit.MICROS).toString();
   }
 
   @Override
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
@@ -493,11 +493,6 @@ public final class Duration implements Comparable<Duration> {
     return this.isEqualTo(other);
   }
 
-  public static Duration fromISO8601String(final String iso8601String){
-    final var javaDuration = java.time.Duration.parse(iso8601String);
-    return of((javaDuration.getSeconds() * 1_000_000L) + (javaDuration.getNano() / 1000L), MICROSECONDS);
-  }
-
   @Override
   public int hashCode() {
     return Long.hashCode(this.durationInMicroseconds);
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PreparedStatements.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PreparedStatements.java
@@ -3,6 +3,7 @@ package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationFailure;
 import gov.nasa.jpl.aerie.merlin.protocol.model.InputType.Parameter;
 import gov.nasa.jpl.aerie.merlin.protocol.model.InputType.ValidationNotice;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.server.http.MerlinParsers;
 import gov.nasa.jpl.aerie.merlin.server.http.ResponseSerializers;
 import gov.nasa.jpl.aerie.merlin.server.models.Timestamp;
@@ -65,6 +66,11 @@ public final class PreparedStatements {
   public static void setTimestamp(final PreparedStatement statement, final int parameter, final Timestamp argument)
   throws SQLException {
     statement.setString(parameter, TIMESTAMP_FORMAT.format(argument.time()));
+  }
+
+  public static void setDuration(final PreparedStatement statement, final int parameter, final Duration argument) throws SQLException {
+    final var micros = argument.in(Duration.MICROSECONDS);
+    statement.setString(parameter, "PT%d.%06dS".formatted(micros / 1_000_000, micros % 1_000_000));
   }
 
   public static void setParameters(final PreparedStatement statement, final int parameter, final List<Parameter> parameters)
--- a/e2e-tests/src/tests/simulation-microsecond-resolution.test.ts
+++ b/e2e-tests/src/tests/simulation-microsecond-resolution.test.ts
@@ -1,0 +1,62 @@
+import { expect, test } from '@playwright/test';
+import req from '../utilities/requests.js';
+import time from "../utilities/time.js";
+
+test.describe('Simulation microsecond resolution', () => {
+  test('GrowBanana with duration of 1 microsecond should finish with a successful simulation', async ({ request }) => {
+    const rd = Math.random() * 100;
+    const plan_start_timestamp = "2021-001T00:00:00.000";
+    const plan_end_timestamp = "2021-001T12:00:00.000";
+    let jar_id: number;
+    let mission_model_id: number;
+    jar_id = await req.uploadJarFile(request);
+    const model: MissionModelInsertInput = {
+      jar_id,
+      mission: 'aerie_e2e_tests',
+      name: 'Banananation (e2e tests)',
+      version: rd + "",
+    };
+    mission_model_id = await req.createMissionModel(request, model);
+    const plan_input : CreatePlanInput = {
+      model_id : mission_model_id,
+      name : 'test_plan' + rd,
+      start_time : plan_start_timestamp,
+      duration : time.getIntervalFromDoyRange(plan_start_timestamp, plan_end_timestamp)
+    };
+    const plan_id = await req.createPlan(request, plan_input);
+    const simulation : SimulationCreation = {
+      plan_id: plan_id,
+      arguments : {},
+    };
+    await req.createSimulation(request, simulation);
+    const activityToInsert : ActivityInsertInput =
+        {
+          arguments : {
+            "duration": 1
+          },
+          plan_id: plan_id,
+          type : "ControllableDurationActivity",
+          start_offset : "1h"
+        };
+
+    const max_it = 10;
+    let it = 0;
+    await req.insertActivity(request, activityToInsert);
+    while (it++ < max_it) {
+      const { reason, status, simulationDatasetId } = await req.simulate(request, plan_id);
+      if (!(status == "pending" || status == "incomplete")) {
+        if (!(status === "complete")) {
+          console.error(JSON.stringify({reason}, undefined, 4));
+          expect(status).toEqual("complete");
+        }
+        return;
+      }
+      await delay(1000);
+    }
+    throw Error("Simulation timed out after " + max_it + " iterations");
+  });
+});
+
+function delay(ms: number) {
+  return new Promise( resolve => setTimeout(resolve, ms) );
+}
--- a/e2e-tests/src/types/simulation.d.ts
+++ b/e2e-tests/src/types/simulation.d.ts
@@ -36,3 +36,11 @@ type SimulationResponseActivity = Omit<ActivityDirective, 'id' | 'startTime'> & 
   computedAttributes: string;
   startTimestamp: string;
 };
+
+type SimulationResponseStatus = 'pending' | 'complete' | 'failed' | 'incomplete';
+
+type SimulationResponse = {
+  reason: any;
+  status: SimulationResponseStatus;
+  simulationDatasetId: number;
+};
--- a/e2e-tests/src/utilities/gql.ts
+++ b/e2e-tests/src/utilities/gql.ts
@@ -28,6 +28,14 @@ const gql = {
     }
   `,
 
+  SIMULATE: `#graphql
+  query Simulate($plan_id: Int!) {
+    simulate(planId: $plan_id){
+      status
+      reason
+    }
+  }`,
+
   CREATE_PLAN: `#graphql
     mutation CreatePlan($plan: plan_insert_input!) {
       insert_plan_one(object: $plan) {
@@ -36,6 +44,7 @@ const gql = {
       }
     }
   `,
+
   CREATE_SIMULATION: `#graphql
     mutation CreateSimulation($simulation: simulation_insert_input!) {
       insert_simulation_one(object: $simulation) {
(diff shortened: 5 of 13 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PreparedStatements.java
```diff
@@ -51,7 +52,6 @@ public String toString() {
       return sqlName;
     }
   }
-
```
Twisol: I think we lost this line by accident 🤔

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -493,14 +493,6 @@ public boolean equals(final Object o) {
...
   }
 
-  public static Duration fromISO8601String(final String iso8601String){
-    return of(java.time.Duration.parse(iso8601String).toNanos() / 1000, MICROSECONDS);
-  }
-
-  public String toISO8601String() {
-    return java.time.Duration.ofNanos(this.durationInMicroseconds * 1000).toString();
-  }
-
```
Twisol: Minor nit -- I think the commits that remove uses of these methods should also include the removal of these methods themselves. (So the commit with `PreparedStatements.setDuration` would come with a removal of `Duration.toISO8601String`.) That way, those commits both feel more like complete inlinings than one-half of an inlining.

Twisol (approved):
LGTM!

Twisol:
BTW, if you add `Fixes #513` in one of the git commit message bodies (probably the first one, here), GitHub will [automatically close](https://github.blog/2013-01-22-closing-issues-via-commit-messages/) the linked issue when the PR is merged.

(I prefer to put these in the body rather than the commit subject (first line) because it would otherwise clutter the `git log --graph --oneline` output.)

camargo:
To e2e test could we:

1. Add a new activity to banananation based on the edits you made to BiteBanana above 
2. Simulate a plan with that activity and note the simulation dataset ID
3. Subscribe to the simulation dataset ID and expect on the status and reason columns
    - Are the errors recorded in the status/reason simulation dataset?
    - If not where are they recorded?

Twisol:
The fault manifested at millisecond resolution, but for regression / e2e testing we really ought to check that we have at least microsecond resolution. So I agree with @camargo, but would change `delayMicros` to be fixed to 1 microsecond.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -494,11 +494,11 @@ public boolean equals(final Object o) {
   }
 
   public static Duration fromISO8601String(final String iso8601String){
-    return of(java.time.Duration.parse(iso8601String).toMillis(), MILLISECOND);
+    return of(java.time.Duration.parse(iso8601String).toNanos() / 1000, MICROSECONDS);
   }
 
   public String toISO8601String() {
-    return java.time.Duration.ofMillis(dividedBy(Duration.MILLISECONDS)).toString();
+    return java.time.Duration.ofNanos(this.durationInMicroseconds * 1000).toString();
```
Twisol: It occurs to me that multiplying by 1000 could actually cause arithmetic overflow in extreme cases -- a 64-bit integer can only span ~600 years (between approx. -300 to 300) so any time beyond that horizon is going to get slammed against the ceiling.

(Back when we defined `Duration`, we chose micros for our time resolution instead of nanos in part because a 300-year positive time horizon seemed somehow too short. I think this is because I hoped (and still hope) that the core modeling and simulation functionality could be used for much longer time-horizon applications. But also, I don't think anybody who would use Merlin actually desires nanosecond resolution 👀)

This issue doesn't affect the replacement of `toISO8601String` in a later commit, since that outputs micros directly. And `fromISO8601String` (and its equivalent that's been moved to `PreparedStatements`) will "only" cause a loss of precision, so distinct times occurring within the same microsecond would be treated as simultaneous -- but this could be fixed by just storing a number of microseconds in the database. (Postgres' `interval` type is much more calendar-sensitive than we really want here, in any case.)

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
mattdailis: I've moved the deletions to their respective commits and deleted the last commit.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
mattdailis: This is a good point - I've updated the first commit to avoid representing the whole duration as a long in nanoseconds.

Now that you bring it up - I think `fromISO8601String` still has the overflow vulnerability if the duration is too large to represent as a long in nanoseconds - simply calling `toNanos` will throw. From the `java.time.Duration::toNanos` javadoc:
> If this duration is too large to fit in a long nanoseconds, then an exception is thrown.

I've propagated the update to the last commit - so that we convert the `seconds` and `nanos` components of a Duration to microseconds separately - if we overflow in microseconds, that's to be expected, but we won't overflow if the duration is too big for nanoseconds (if I did my math right 🤞 )

Twisol: Yikes. Let's look at the trade-offs of replacing `interval` in Postgres soon >_<

JoelCourtney:
I'd like to keep the to/from converters, because I'd like to use them to deserialize Temporal.Duration. ISO8601 is intended to be a standard way to serialize/deserialize and communicate durations, so I'd like to use it instead of copying the post-processor bodge in the scheduler dsl main.ts that uses a JSON replacer on the AST object and converts Temporal.Duration objects to micros. Temporal.Duration serializes to ISO8601 by default.

JoelCourtney:
This should fix the rounding issue:

```java
  public static Duration fromISO8601String(final String iso8601String){
    return of(java.time.Duration.parse(iso8601String).get(ChronoUnit.MICROS), MICROSECOND);
  }

  public String toISO8601String() {
    return java.time.Duration.of(this.durationInMicroseconds, ChronoUnit.MICROS).toString();
  }
```

edit: nvm, this throws a runtime exception because the Micros unit is unsupported. Cool cool. will find another way.

edit 2: looks like you fixed it before deleting it. please put it back?

Twisol:
...Well, this is the implementation of `java.time.Duration#get`:
```
    /**
     * Gets the value of the requested unit.
     * <p>
     * This returns a value for each of the two supported units,
     * {@link ChronoUnit#SECONDS SECONDS} and {@link ChronoUnit#NANOS NANOS}.
     * All other units throw an exception.
     *
     * @param unit the {@code TemporalUnit} for which to return the value
     * @return the long value of the unit
     * @throws DateTimeException if the unit is not supported
     * @throws UnsupportedTemporalTypeException if the unit is not supported
     */
    @Override
    public long get(TemporalUnit unit) {
        if (unit == SECONDS) {
            return seconds;
        } else if (unit == NANOS) {
            return nanos;
        } else {
            throw new UnsupportedTemporalTypeException("Unsupported unit: " + unit);
        }
    }
```
That is to say, it doesn't actually support querying MICROS, just NANOS. (And importantly, getting NANOS doesn't *convert* to nanoseconds, it just gets the *nanosecond field* of the duration -- you miss anything in `seconds` or higher.)

Twisol:
My concern about the ISO-8601 format is that it's fundamentally [sensitive to calendrical quantities](https://en.wikipedia.org/wiki/ISO_8601#Durations) -- an ISO-8601 duration may add different real quantities of time to different initial times. For instance, adding "P1M" (one month) to a time on January 1st will net February 1st, while adding the same to February 1st will net March 1st. It is commonly known that January and February have a different number of days, so the string "P1M" does not actually represent a unique, fixed span of time.

In contrast, our `Duration` is just a number of microseconds, totally independent of any calendar. I'd argue that ISO-8601 is not appropriate for what we're trying to model with `Duration`. For the same reason, I also discussed with Matt that we probably ought to change the type of the `duration` column in Postgres from `interval` to a plain sized integer type.

> I'd like to keep the to/from converters, because I'd like to use them to deserialize Temporal.Duration. ISO8601 is intended to be a standard way to serialize/deserialize and communicate durations, so I'd like to use it instead of copying the post-processor bodge in the scheduler dsl main.ts that uses a JSON replacer on the AST object and converts Temporal.Duration objects to micros. Temporal.Duration serializes to ISO8601 by default.

1. Can you remind me what `Temporal.Duration` is used for here?
2. The `merlin.protocol` package is meant to be all and only what is necessary to communicate between a model and the simulation host. Matt mentioned some concerns about forward compatibility of built mission models as we continue upgrading Aerie, so I think we have a strong interest in keeping the `merlin.protocol` packages as small and stable as possible. I'd like to know more about your use-case here, but it doesn't sound immediately like it's relevant to the purpose of `merlin.protocol`.

As an aside, there's no need for this Duration class to be the one and only class modeling Durations in the whole of Aerie. In fact, its presence in `merlin.protocol` indicates that its use should generally be *avoided*.

Given that `Duration` is properly not much more than a data class with one field, I don't really see a problem with creating another `Duration` class elsewhere to cater to a separate class of needs.

Twisol:
I would be quite okay with copying the `Duration` we have now into `merlin.framework`, replacing it in `merlin.protocol` with a simple `record Duration(long microseconds) {}`, and implementing any application-specific duration models wherever they need to exist. The class as it stands now is too tempting to use everywhere (and hence be influenced from everywhere).

mattdailis:
There are two separate questions at play here:
#### 1. How should we serialize Temporal.Duration?

Background: In #187 - we introduced Temporal.Duration to the scheduling DSL and serialized it using ISO8601. It didn't work as well as we'd hoped (i.e. it broke completely for directives that had Duration parameters), and we had to push out a hotfix that would serialize it as an integer in microseconds: #383.

Perhaps that was a symptom of things being coupled that shouldn't be - the serialized arguments coming out of the scheduling DSL are directly passed into the activity mappers - but then again I have trouble picturing an intermediary step there that would allow Durations to flow in as strings on one end and out as integers on the other that would be any less of a "bodge".

We could try to serialize it differently depending on whether it's an argument to a directive or not - I imagine this could be done via the ActivityTemplates codegen on the scheduler side - but I don't think we would gain much from that additional complexity.

There is the option to use ISO8601 everywhere as our serialization of duration - however, for the reasons @Twisol mentioned above, and for reasons brought up in our recent "Rollover Time Management" discussion - I think we want to steer away from calendrical representations of time wherever we can. I'd like to dig a little deeper into this after consulting the Monte docs and other sources about time. I think this is all out of scope for #501 though.

#### 2. Where should we put that serialization logic?

The purpose of `merlin.protocol` is to define the "plugin" interface that a model must conform to - anything on top of that is cluttering the interface. Jonathan's suggestion above would be in the spirit of minimizing this interface: https://github.com/NASA-AMMOS/aerie/pull/520#issuecomment-1344914957 - since convenience methods for the model belong in the `merlin-framework`, and convenience methods for the rest of the system belong... elsewhere.

Currently - each pair of communicators defines their own serialization logic (i.e. merlin-server <---> db, hasura <---> merlin-server, scheduler-server <---> db, etc.) - with the caveat that the workers and the servers share the same code for talking to the database. This gives each pair the choice of how to serialize things - in some ways this is a Good Thing:tm:, since we have the flexibility to choose the representation that makes the most sense for each case. On the other hand this is a Bad Thing:tm: because perhaps we want to make the same choices most of the time.

If we're concerned about code re-use - we can make a library along the lines of "serialization-utils" that includes (de)serialization logic for common types, including Duration.

I'm supportive of the desire to be consistent across the system with how we serialize Durations - but I want to make sure that we choose a serialization format that leaves no room for ambiguity around "how long is a year/month/day" and "were there leap seconds in this interval". As it stands, an integer in microseconds permits no such ambiguities.

camargo (approved):

PR #521, opened 2022-12-06
Scheduling uniqueness migrations

### LINKED ISSUE #471: Add constraints to the scheduling_specification and scheduling_specification_goals tables

mattdailis:
* **Tickets addressed:** Closes #471
* **Review:** By commit  

## Description

This PR is a second attempt - the first attempt was #471 . The key difference is that the first attempt directly edited our init.sql files, while this PR makes the change as a database migration.

The purpose of this PR is to restrict the flexibility of scheduling specifications. Currently, it is possible for a single goal to be a member of multiple specifications - which means that editing a scheduling goal from the context of one plan will cause it to be changed for all plans whose scheduling specifications contain this goal. This was deemed undesirable in a recent use case working group. It was also deemed undesirable to allow multiple scheduling specifications per plan.

If we commit to this approach long-term, we should consider normalizing our data structures as pointed out by @JoelCourtney https://github.com/NASA-AMMOS/aerie/pull/475#pullrequestreview-1177885404. In the interest of not changing the API (yet) - I've opted to merely add constraints without restructuring the data. I'm also not sure that we've really ironed out the use cases around scheduling specifications - especially in the context of plan branching - so I'd like to keep our over-engineered data structures around a little longer.

This PR includes two migrations - one to add a constraint to the `scheduling_specification_goals` table to make sure that each goal id is present at most once - and another to add a constraint to the `scheduling_specification` table to make sure that no two specifications refer to the same plan.

### A note about migration numbering:
In this PR I have elected to number migrations sequentially. Hasura seems to be perfectly happy with this - so although hasura's "create" function will number migrations based on timestamp - manually renaming the folder to use a sequential ID seems to have no detrimental effects. As discussed here: #346 a sequential numbering scheme forces us to think carefully about the order in which migrations are applied - and will cause helpful collisions if we rebase onto develop after another branch has introduced a migration with the same ID. Whether we choose to use sequential IDs or timestamps, we will still need to re-number our migrations after rebase - and my impression is that renumbering sequential IDs is less error-prone. 

- [x] TODO: Before merging this PR, I'd like to figure out what changes should be made to our current `init.sql` and related files - we could move them all into a zeroth migration.

## Verification

This PR introduces migrations to our `db-tests` suite. Each migration has its own test file, in which the `helper.startDatabaseBeforeMigration` method is used to bring the schema up to but not including the migration under test. After inserting seed data, a test can use the `applyMigration`, `rollbackMigration`, and `runSqlFile` to verify that the migration behaves as expected.

## Documentation

- [ ] TODO instructions for how to run migrations are necessary

## Future work

- Put some thought into whether or not a similar change need be applied for templates
- Consider refactoring the tables to be more normalized

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
@@ -7,8 +7,16 @@ import com.zaxxer.hikari.HikariDataSource;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * Manages the test database.
@@ -19,16 +27,36 @@ public class DatabaseTestHelper {
   private final String dbName;
   private final String appName;
   private final File initSqlScriptFile;
+  private final Optional<Path> migrationsDirectory;
 
   public DatabaseTestHelper(String dbName, String appName, File initSqlScriptFile) {
+    this(dbName, appName, initSqlScriptFile, Optional.empty());
+  }
+
+  public DatabaseTestHelper(String dbName, String appName, File initSqlScriptFile, Path migrationsDirectory) {
+    this(dbName, appName, initSqlScriptFile, Optional.of(migrationsDirectory));
+  }
+
+  private DatabaseTestHelper(String dbName, String appName, File initSqlScriptFile, Optional<Path> migrationsDirectory) {
     this.dbName = dbName;
     this.appName = appName;
     this.initSqlScriptFile = initSqlScriptFile;
+    this.migrationsDirectory = migrationsDirectory;
   }
 
-  public void startDatabase() throws SQLException, IOException, InterruptedException {
+  public void startDatabaseWithLatestSchema() throws SQLException, IOException, InterruptedException {
     createDatabase();
-    runInitSql();
+    if (migrationsDirectory.isPresent()) {
+      applyAllMigrations();
+    } else {
+      runInitSql();
+    }
+    establishConnection();
+  }
+
+  public void startDatabaseBeforeMigration(final String migrationName) throws SQLException, IOException, InterruptedException {
+    createDatabase();
+    applyMigrationsBefore(migrationName);
     establishConnection();
   }
 
@@ -130,5 +158,80 @@ public class DatabaseTestHelper {
 
   public Connection connection() {
     return connection;
+  }
+
+  void applyMigration(final String migrationName) throws IOException, InterruptedException, SQLException {
+    assertMigrationExists(migrationName);
+    runSqlFile(migrationsDirectory.get().resolve(migrationName).resolve("up.sql").toString());
+  }
+
+  void rollbackMigration(final String migrationName) throws IOException, InterruptedException, SQLException {
+    assertMigrationExists(migrationName);
+    runSqlFile(migrationsDirectory.get().resolve(migrationName).resolve("down.sql").toString());
+  }
+
+  /**
+   * Applies all migrations alphabetically up to but not including the given migration
+   */
+  void applyMigrationsBefore(final String migrationName) throws IOException, InterruptedException, SQLException {
+    assertMigrationExists(migrationName);
+    Integer previousVersion = null;
+    for (final var migration : getSortedMigrations()) {
+      if (previousVersion == null) {
+        previousVersion = migration.version();
+      } else {
+        assertTrue(
+            migration.version() > previousVersion,
+            "Duplicate migration version found: %d is not greater than %d".formatted(
+                migration.version(),
+                previousVersion));
+      }
+      if (migration.name().equals(migrationName)) {
+        return;
+      }
+      applyMigration(migration.name());
+    }
+  }
+
+  /**
+   * Applies all migrations alphabetically
+   */
+  void applyAllMigrations() throws IOException, InterruptedException, SQLException {
+    Integer previousVersion = null;
+    for (final var migration : getSortedMigrations()) {
+      if (previousVersion == null) {
+        previousVersion = migration.version();
+      } else {
+        assertTrue(
+            migration.version() > previousVersion,
+            "Duplicate migration version found: %d is not greater than %d".formatted(
+                migration.version(),
+                previousVersion));
+      }
+      applyMigration(migration.name());
+    }
+  }
+
+  private List<Migration> getSortedMigrations() {
+    return Arrays
+        .stream(Objects.requireNonNull(migrationsDirectory.get().toFile().listFiles()))
+        .map(Migration::of)
+        .sorted(Comparator.comparing(Migration::name))
+        .toList();
+  }
+
+  void assertMigrationExists(final String migrationName) {
+    if (!new File(migrationsDirectory.get().resolve(migrationName).toString()).isDirectory()) {
+      throw new AssertionError("No such migration found: %s".formatted(migrationName));
+    }
+  }
+
+  record Migration(int version, String name) {
+    static Migration of(final String migrationName) {
+      return new Migration(Integer.parseInt(migrationName.split("_")[0]), migrationName);
+    }
+    static Migration of(final File file) {
+      return Migration.of(file.getName());
+    }
   }
 }
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
@@ -3,7 +3,6 @@ package gov.nasa.jpl.aerie.database;
 import com.impossibl.postgres.jdbc.PGDataSource;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
-import org.junit.jupiter.api.Assumptions;
 
 import java.io.File;
 import java.io.IOException;
@@ -27,46 +26,40 @@ public class DatabaseTestHelper {
     this.initSqlScriptFile = initSqlScriptFile;
   }
 
+  public void startDatabase() throws SQLException, IOException, InterruptedException {
+    createDatabase();
+    runInitSql();
+    establishConnection();
+  }
+
   /**
    * Sets up the test database
    */
-  public void startDatabase() throws SQLException, IOException, InterruptedException {
+  private void createDatabase() throws IOException, InterruptedException {
     // Create test database and grant privileges
-    {
-      final var pb = new ProcessBuilder("psql",
-                                        "postgresql://postgres:postgres@localhost:5432",
-                                        "-v", "ON_ERROR_STOP=1",
-                                        "-c", "CREATE DATABASE " + dbName + ";",
-                                        "-c", "GRANT ALL PRIVILEGES ON DATABASE " + dbName + " TO aerie;"
-      );
-
-      final var proc = pb.start();
-
-      // Handle the case where we cannot connect to postgres by skipping the tests
-      final var errors = new String(proc.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
-      Assumptions.assumeFalse(
-          (  errors.contains("Connection refused")
-          || errors.contains("role \"postgres\" does not exist")));
-      proc.waitFor();
-      proc.destroy();
-    }
+    runSubprocess(
+      "psql",
+      "postgresql://postgres:postgres@localhost:5432",
+      "-v", "ON_ERROR_STOP=1",
+      "-c", "CREATE DATABASE " + dbName + ";",
+      "-c", "GRANT ALL PRIVILEGES ON DATABASE " + dbName + " TO aerie;"
+    );
 
     // Grant table privileges to aerie user for the tests
     // Apparently, the previous privileges are insufficient on their own
-    {
-      final var pb = new ProcessBuilder("psql",
-                                        "postgresql://postgres:postgres@localhost:5432/" + dbName,
-                                        "-v", "ON_ERROR_STOP=1",
-                                        "-c", "ALTER DEFAULT PRIVILEGES GRANT ALL ON TABLES TO aerie;",
-                                        "-c", "\\ir %s".formatted(initSqlScriptFile.getAbsolutePath())
-      );
+    runSubprocess(
+      "psql",
+      "postgresql://postgres:postgres@localhost:5432/" + dbName,
+      "-v", "ON_ERROR_STOP=1",
+      "-c", "ALTER DEFAULT PRIVILEGES GRANT ALL ON TABLES TO aerie;"
+    );
+  }
 
-      pb.redirectError(ProcessBuilder.Redirect.INHERIT);
-      final var proc = pb.start();
-      proc.waitFor();
-      proc.destroy();
-    }
+  private void runInitSql() throws IOException, InterruptedException, SQLException {
+    runSqlFile(initSqlScriptFile.getAbsolutePath());
+  }
 
+  private void establishConnection() throws SQLException {
     final var pgDataSource = new PGDataSource();
 
     pgDataSource.setServerName("localhost");
@@ -85,27 +78,54 @@ public class DatabaseTestHelper {
   }
 
   /**
+   * Runs the given sql file
+   */
+  public void runSqlFile(final String filePath) throws IOException, InterruptedException, SQLException {
+    final var errors = runSubprocess(
+        "psql",
+       "postgresql://postgres:postgres@localhost:5432/" + dbName,
+       "-v", "ON_ERROR_STOP=1",
+       "-f", filePath);
+    if (errors.toLowerCase().contains("error:")) {
+      throw new SQLException(errors);
+    }
+  }
+
+  /**
    * Tears down the test database
    */
   public void stopDatabase() throws SQLException, IOException, InterruptedException {
 
-    Assumptions.assumeTrue(connection != null);
-    connection.close();
+    if (connection != null) {
+      connection.close();
+    }
 
     // Clear out all data from the database on test conclusion
     // This is done WITH (FORCE) so there aren't issues with trying
     // to drop a database while there are connected sessions from
     // dev tools
-    final var pb = new ProcessBuilder("psql",
-                                      "postgresql://postgres:postgres@localhost:5432",
-                                      "-v", "ON_ERROR_STOP=1",
-                                      "-c", "DROP DATABASE IF EXISTS " + dbName + " WITH (FORCE);"
+    runSubprocess(
+        "psql",
+        "postgresql://postgres:postgres@localhost:5432",
+        "-v", "ON_ERROR_STOP=1",
+        "-c", "DROP DATABASE IF EXISTS " + dbName + " WITH (FORCE);"
     );
+  }
 
-    pb.redirectError(ProcessBuilder.Redirect.INHERIT);
+  /**
+   * Run a subprocess and return its stderr output
+   */
+  private static String runSubprocess(final String ... args) throws IOException, InterruptedException {
+    final var pb = new ProcessBuilder(args);
     final var proc = pb.start();
-    proc.waitFor();
-    proc.destroy();
+    try {
+      proc.waitFor();
+      final var stderr = new String(proc.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
+      System.err.println(stderr);
+      return stderr;
+    } finally {
+      proc.destroy();
+    }
   }
 
   public Connection connection() {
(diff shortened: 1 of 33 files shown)


Twisol:
> This PR is a second attempt - the first attempt was #471 . The key difference is that the first attempt directly edited our init.sql files, while this PR makes the change as a database migration.
>
>  TODO: Before merging this PR, I'd like to figure out what changes should be made to our current init.sql and related files - we could move them all into a zeroth migration.

Won't we want both? My note about upgraders and newcomers [elsewhere](https://github.com/NASA-AMMOS/aerie/discussions/467#discussioncomment-4319395) becomes quite technically relevant here -- we need to migrate existing customers' databases, but also give new customers a direct path to the latest database. We could simply keep `init.sql` and its includes fixed forever, and apply all migrations on a new deployment, but this would make the existing SQL schemas significantly less useful from a development & reference perspective. Is the planned approach to schema maintenance documented anywhere?

camargo:
> We could simply keep init.sql and its includes fixed forever, and apply all migrations on a new deployment, but this would make the existing SQL schemas significantly less useful from a development & reference perspective. Is the planned approach to schema maintenance documented anywhere?

I agree with this I think we need both. We should keep our schemas up to date as the gold source references of our data model, and also maintain migrations to reflect those schemas if upgrading from an old version.

mattdailis:
> 



> > We could simply keep init.sql and its includes fixed forever, and apply all migrations on a new deployment, but this would make the existing SQL schemas significantly less useful from a development & reference perspective. Is the planned approach to schema maintenance documented anywhere?
> 
> I agree with this I think we need both. We should keep our schemas up to date as the gold source references of our data model, and also maintain migrations to reflect those schemas if upgrading from an old version.

I'm on board - but I'm concerned about them getting out of sync. Perhaps we could:
a. Dump the latest schema after applying all migrations (we'd need to be happy with its auto-formatting though)
b. Do some kind of automated schema diff as part of PR checks (bring up two databases, one with migrations, one with the hand-written schema, assert that the diff is empty)
c. Just be super careful

I could be talked into c... but I'm wary of relying on programmer discipline to keep two sources of truth in sync.

> give new customers a direct path to the latest database

Yes - I can see this desire. My thinking was - applying all the migrations on an empty database should take very little time, even if it's a long list of migrations. (Most likely seconds - minutes, tops?) We can even publish a docker container with these migrations pre-applied.

If we do go the route where new customers directly apply the latest state of the database - we'll need some way to indicate which migrations are logically "before" this schema version, and which (future) migrations are logically "after". We could try to write directly to hasura's migration-tracking table and assert that all previous migrations have been applied 🤔

mattdailis:
> I could be talked into c... but I'm wary of relying on programmer discipline to keep two sources of truth in sync.

That said... perhaps the schemas don't need to match perfectly - they just need to pass all our tests. So we can run tests against both a new database and an upgraded one.

Though, that does mean that future migrations could need to run against subtly different previous schemas, which sounds like a headache to me... 🤔

Twisol:
My preferred workflow is to edit the schemas directly, developing until I've completed the changes I wanted to make, then do a diff to guide the creation of a migration.

(That is, I follow "just be super careful". Which I'm not necessarily advocating for.)

dyst5422:
> My preferred workflow is to edit the schemas directly, developing until I've completed the changes I wanted to make, then do a diff to guide the creation of a migration.

Agreed on workflow, but we still need a mechanism for validation (people are flawed and thats ok).

I think it necessarily has to involve processing migrations and asserting that the resulting schema post migration route === resulting schema from a new deployment. Additionally, we'll have to get tricky about that mechanism since we'll have to reach back through the commit history to get the initial database setup state to apply migrations on since we'll be changing the initialization sql files as gold source.

Twisol:
> a. Dump the latest schema after applying all migrations (we'd need to be happy with its auto-formatting though)

I'm rarely particularly happy with auto-formatters, and I personally really want to have things split into files and be organized logically. Having a canonical copy of the desired database schema is a really valuable thing for development, and I'd want it to serve our needs, not just be a dump from a database.

Also, it's not just schematic definitions that need to be included, but also sometimes fixtures -- tables that contain pre-initialized data. Dumping from a real system means you need some way to distinguish fixtures (to be kept) from incidental data (to be excluded).

> b. Do some kind of automated schema diff as part of PR checks (bring up two databases, one with migrations, one with the hand-written schema, assert that the diff is empty)

I actually rather like this idea. We can dump from a system initialized both ways, and confirm that the dumps are equivalent. This might get subtle, but I think it's a good idea.

> c. Just be super careful

This is how I've traditionally worked, but I agree that it's not ideal. I'd rather explore something like (b) before giving up.

dyst5422:
Additional note - we'll want to have some way to validate that migrations are not destructive in a way that jeopardizes users data. My guess is some unit tests around each migration, or at least a known review gateway.

Twisol:
> Additional note - we'll want to have some way to validate that migrations are not destructive in a way that jeopardizes users data. My guess is some unit tests around each migration, or at least a known review gateway.

I think that's what Matt's tried to set up here:

> ## Verification
> 
> This PR introduces migrations to our `db-tests` suite. Each migration has its own test file, in which the `helper.startDatabaseBeforeMigration` method is used to bring the schema up to but not including the migration under test. After inserting seed data, a test can use the `applyMigration`, `rollbackMigration`, and `runSqlFile` to verify that the migration behaves as expected.

dyst5422:
> I think that's what Matt's tried to set up here:

🤦 Got too deep into the thread, forgot the original post content

Mythicaeda:
> TODO instructions for how to run migrations are necessary

Are these separate docs from our [Database Migration Script docs](https://nasa-ammos.github.io/aerie/stable/user-guide/upgrade-guides/db_migration_assistant.html)?

mattdailis:
I think we're all in agreement that:
- We want to maintain both migrations and the sql files representing the latest database schema
- We want to automate checking that they do in fact represent the same schema

@Mythicaeda and I are current discussing whether or not we want to use the migrations or the latest schema when starting up a new database.

Option 1: Run all migrations on the empty database
Note:
- Requires users to install hasura-cli in order to deploy Aerie
Pros:
- The database schema for all users is determined the same way
Cons:
- If a migration were to fail that would be a confusing start-up experience
- If our init.sql gets out of sync with our migrations, customer code won't break - thus low visibility

Option 2: Apply the latest schema directly to the empty database
Notes:
- This approach would need a strategy for how to indicate what migrations need to be applied
  - We could only provide migrations from Aerie version n-1 to n - that way new users never see old migrations
  - Users who need to upgrade through multiple Aerie versions at once would download the migrations for each version that they need to upgrade through
- Requires users to install hasura-cli prior to upgrading
Pros:
- New users easy startup
Cons:
- Should our init.sql get out of sync with migrations - we will have divergent database states between users who installed Aerie at different times

Twisol:
I tend to prefer Option 2, but I think you've summarized things fairly.

For what it's worth, based on how the docs seem to be positioned, Hasura Migrations seems better suited to a devops workflow where the developers also manage a workflow, than to a split-team scenario where we provide migrations and customers apply them. I wonder if it would be difficult to roll our own solution?

<details>
<summary>Just mocking out a thought...</summary>

Create our own table to mark the database version:
```sql
create table meta (major_version int not null, minor_version int not null, patch_version int not null);
insert into meta values (1, 0, 0);
```

Then we write migrations in files named e.g. `from_1.0.0_to_1.0.1.sql`, and when we need to apply a series of migrations we just look at the database's current version and the target version and load the migrations that get us to that point into the database (`./scripts/migrate-to 1.0.1`).
</details>

(FWIW, I did a little googling and found [this article](https://workingwithdevs.com/delivering-databases-migrations-vs-state/) comparing the schema-first and migration-first approaches. You might have already read it, but if you haven't, I thought it was interesting.)

mattdailis:
Updates:

I've added a `schema_migrations` table to the AerieScheduler database, borrowing from the terminology Ruby on Rails uses: https://blog.appsignal.com/2020/04/14/dissecting-rails-migrationsl.html

This table tracks which migrations have been applied to achieve the current schema. I provided two convenience functions for interacting with this table - `mark_migration_applied` and `mark_migration_rolled_back`, which insert and delete from the table respectively.

In this approach - every migration must call `mark_migration_applied` with its id prior to committing its transaction. Similarly, the "latest" schema represented by our files in `postgres-init-db` must also call `mark_migration_applied` for all of the migrations whose changes are represented in the latest schema.  I put these calls into `applied_migrations.sql`.

So - the process to add a new migrations becomes as follows:
- Declare a new migration with the next available id by creating a folder in `deployments/hasura/migrations/<DATABASE>/` with the naming convention `<id>_<brief_human_readable_name>`.
- Write an `up.sql` in that folder that defines the migration. Include a call to `mark_migration_applied('<id>')`
- Write a `down.sql` in that folder that returns the schema to the previous state. Include a call to `mark_migration_rolled_back('<id>')`.
- Make the corresponding changes to the "latest" schema
- Add a call to `mark_migration_applied('<id>')` to `applied_migrations.sql`

These steps can be performed in any order - and should probably all be done in one commit.

I added a test called `EnsureMigrationsAndSchemasMatch` (open to better names if people have suggestions). This test brings up two databases - one using just the "latest" schema - and one starting from an empty database, and applying all migrations in order. It proceeds to dump both schemas using `pg_dump`, and asserts that the strings match exactly. It also queries the contents of the `schema_migrations` tables and asserts that they contain the same entries.

Future work: either update or replace our `aerie_db_migration` script to use the `schema_migrations` table to check if a migration has been run.

With that, I think this is ready for review.

Mythicaeda:
Before I dive in, quick concern I have re: the PGDump: in the migration I'm working on for anchors, I need to drop a few columns from Hasura_Functions and then add them back later in order to "release the lock" on the activity_directive table (the columns in question use a row from that table as a column type). Do you know if this will still produce the same string (ie, does PGDump determine column order for the string alphabetically or by order inserted)?

On a positive note, we should be able to write a quick bash script to assert that all the `up`s and `down`s contain those `mark_migration_applied`/`mark_migration_rolled_back` calls, and from there we can add it to our `.github/workflows`.

mattdailis:
Whoops, I thought I had replied this morning; I had not. No need to get to this before the "early dismissal".

> Before I dive in, quick concern I have re: the PGDump: in the migration I'm working on for anchors, I need to drop a few columns from Hasura_Functions and then add them back later in order to "release the lock" on the activity_directive table (the columns in question use a row from that table as a column type). Do you know if this will still produce the same string (ie, does PGDump determine column order for the string alphabetically or by order inserted)?

Great question - pg_dump is sensitive to column order - so, in some ways, the "exact string match" may be more restrictive than we really need. However, a more nuanced check of schema "equivalence, modulo column order" would be significantly more complex. I'm open to suggestions for how to accomplish this - in the near term I would suggest going ahead and changing the column order in the "latest" schema to keep them in sync.

> On a positive note, we should be able to write a quick bash script to assert that all the `up`s and `down`s contain those `mark_migration_applied`/`mark_migration_rolled_back` calls, and from there we can add it to our `.github/workflows`.

Yes! That could work. We could also write a db-test that checks the state of the `schema_migrations` table after every migration.

Twisol:
FWIW, column order *is* something that various operations in the database can be sensitive to. From [this SO answer](https://stackoverflow.com/a/34458442):

> PostgreSQL implements primary keys with an underlying unique index. If the primary key contains more than one column, then PostgreSQL creates the underlying index with columns in the order you declare them.

And of course, `INSERT INTO tbl VALUES (a, b, c);` will use the default column order (contrast with `INSERT INTO tbl (x, y, z) VALUES (a, b, c)`).

So I'm not sure column order is something we will want to default to ignoring.

Twisol:
If we need to insert columns at a specific position, we can use [this approach](https://stackoverflow.com/a/39781155), in which we create a new table with the desired shape, then INSERT into it via a SELECT on the old table. IIRC, there are ways to defer or even wholly disable constraints during operations like this for efficiency, if necessary.

Mythicaeda:
From my understanding, @Twisol, primary keys, foreign keys, and constraints are included in the dump. We should then be able to check that the keys are exact matches, as based on what you shared, it sounds like the primary keys are the same iff the columns used in the primary keys are in the same relative order. I believe that bypasses the need for column order to be completely identical between the two versions, unless there's another condition where we'd need them to be identical (excluding any potential instances where we insert into a table without specifying the columns the data goes into, which we should fix anyway).

@mattdailis, regarding how to check if two databases are equal sans column order, [this `pgcmp` tool](https://github.com/cbbrowne/pgcmp) looks like a good place to start. I don't know if we want all of its functionality, but the section starting [here](https://github.com/cbbrowne/pgcmp/blob/master/pgcmp#L129) appears to categorize matches based on if there exists an object (as in columns, permissions, constraints, triggers, etc) with the same name, datatype, and definition in both databases, and from my understanding of [the dump script](https://github.com/cbbrowne/pgcmp/blob/master/pgcmp-dump#L74), it looks like each column/key/etc is included as a separate object, meaning that the comparisons should be order-independent.

mattdailis:
Good find, @Mythicaeda ! I'm having some difficulty getting pgcmp to work for me, so I was unable to verify how it treats column order (it says "no differences found" even when I intentionally inject major differences... I think I'm mis-using it somehow).

I suggest we continue to use the "strict" equality test for this first cut, and then consider loosening the check if it turns out to be too strict. The pgcmp README draws a distinction between the "Build process" use case (which closely resembles our own use case), from the "Validating upgrades" use case - which leverages more of pgcmp's "equivalence" checks:

> 1. Build process 
> ...
> - We need to ensure that the results of building from scratch are identical to those that come from upgrading the old version using upgrade scripts.
> - In this case, we do not expect any variations, aside from the fact that OIDs will vary between databases. These databases will be built within the same smoke test environment, and thus have the same sets of users, roles, and such, and there should not be any additional components introduced (as might be the case in production)
> - It is therefore considered a build-breaking error to find any variation between the two database schemas

We get these properties from the `pg_dump` equality check :+1:

The additional features of pgcmp I think will be useful when we get around to their second use case, "Validating Upgrades" - "Replication", "DBA Tooling", "Supplementary Applications", and "Specialized ownership" are all reasonable forms of drift that may occur between a long-lived production database and a development one. That's an important problem - but it's separate from the "the latest schema should match the result of running the migrations" problem that this PR tries to address. If it's okay with you, I'd like to mark the "validate user database prior to running migrations" use case as out of scope for this ticket.

Since we have control over both the migrations, and the "latest" schema, I suggest that we strive to keep them identical (not merely equivalent).

A note on the use case you're working with - the hasura_functions are a bit of an edge case - since those tables don't hold any data, and exist just for hasura to be able to inspect their schema. The two possible approaches I would explore are:
1. Write the migration as you've been writing it, and update the column order in the latest schema to match
2. Write the migration in a way that the column order doesn't change (i.e. drop and recreate the whole table - which is safe because it has no data)

In option 1, there will be a diff in the latest schema that reviewers will need to understand is merely a column reordering. In option 2, there will be no diff in the latest schema, but the migration script will include the full table re-definition.

mattdailis:
We made #567 to limit the scope of this PR.

We did manage to get pgcmp working locally :+1: Pro tip: make sure the PGCLABEL args match between the pgcmp-dump and the pgcmp commands.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
```diff
@@ -20,53 +27,67 @@ public class DatabaseTestHelper {
   private final String dbName;
   private final String appName;
   private final File initSqlScriptFile;
+  private final Optional<Path> migrationsDirectory;
```
mattdailis, line 30: TODO: Make this better.

Currently - the DatabaseTestHelper has branching logic based on whether migrationsDirectory was provided - if it is provided, it will ignore the `initSqlScriptFile` argument. This is unintuitive and can be better.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
```diff
@@ -84,31 +105,154 @@ public void startDatabase() throws SQLException, IOException, InterruptedExcepti
...
+  /**
+   * Runs the given sql file
+   */
+  public void runSqlFile(final String filePath) throws IOException, InterruptedException, SQLException {
+    final var errors = runSubprocess(
+        "psql",
+       "postgresql://postgres:postgres@localhost:5432/" + dbName,
+       "-v", "ON_ERROR_STOP=1",
+       "-f", filePath);
+    if (errors.toLowerCase().contains("error:")) {
```
mattdailis, line 117: TODO: Maybe we should check the return code of psql instead of (or in addition to) looking for the string "errors:" in the output.

deployment/hasura/migrations/AerieScheduler/1_make_scheduling_goals_unique/down.sql
```diff
@@ -0,0 +1,19 @@
+/*
+This script restores the schema to its state before the migration. It is to be used if the Aerie upgrade is deemed,
+unsuccessful, and the database needs to be restored to compatibility with the previous version of Aerie.
+
+Note that this migration does not affect Aerie compatibility - it does not violate any assumptions that Aerie makes.
```
Mythicaeda, line 5: Couple of things regarding  `this migration does not affect Aerie compatibility`:
1. The next sentence implies that "this migration" refers to the up migration. If this is the case, then this section of the comment should be in the `up.sql`.
2. If "this migration" _is_ the up migration, then there shouldn't be a note about running the down migration to "restore compatibility with the previous version of **Aerie**" in the prior paragraph, since the compatibility issues are actually with APIs based on this paragraph.

deployment/hasura/migrations/AerieScheduler/1_make_scheduling_goals_unique/up.sql
```diff
@@ -0,0 +1,17 @@
+/*
+make_scheduling_goals_unique
```
Mythicaeda, line 2: Why is the name of the migration included here?

scheduler-server/sql/scheduler/tables/schema_migrations.sql
```diff
@@ -0,0 +1,21 @@
+create table "schema_migrations" (
+  "version" varchar not null primary key
```
Mythicaeda, line 2: Remind me, do we plan on using non-integers to refer to which migrations have been applied?
Also, because migrations are associated with releases, the column name `version` makes me think of Aerie versions (ie v1.0.0, v0.13.2) rather than applied migration numbers. Is there another term we can use or is `version` considered standard?

deployment/hasura/migrations/AerieScheduler/0_init/tables/schema_migrations.sql
```diff
@@ -0,0 +1,23 @@
+create table "schema_migrations" (
+  "version" varchar not null primary key
+);
+
+create function mark_migration_applied(migration_version varchar)
+returns void
+language plpgsql as $$
```
Mythicaeda, line 7:
```suggestion
create procedure mark_migration_applied(migration_version varchar)
language plpgsql as $$
```

And similarly for `mark_migration_rolled_back`/other void functions.

deployment/hasura/migrations/AerieScheduler/0_init/tables/schema_migrations.sql
```diff
@@ -0,0 +1,23 @@
...
+create function mark_migration_rolled_back(migration_version varchar)
+returns void
+language plpgsql as $$
+begin
+  delete from schema_migrations
+  where version = migration_version;
+end;
+$$;
+
+select mark_migration_applied('0');
```
Mythicaeda, line 23: [Ignorable suggestion] Could this line be added to the bottom of the `up.sql` instead? It feels a little weird to mark this migration as done when this is run as the first step.

Mythicaeda (commented):
Feedback regarding the `.sql` files.

One additional question: Could you please explain why the file `0_init/down.sql` exists if it's empty?

deployment/hasura/migrations/AerieScheduler/0_init/tables/schema_migrations.sql
mattdailis: From what I understand, the primary purpose of procedures, which differentiates them from functions, is the ability to begin and commit transactions. If we don't need that power, I recommend that we use void functions instead.

Am I perhaps overlooking a reason not to use void functions?

Mythicaeda: From my understanding, while procedures *can* begin/end transactions, [based on how the documentation is written](https://www.postgresql.org/docs/14/xproc.html), it seems like that's just an additional thing it can do, while its main distinction is that, unlike a function, it doesn't have a return value. 

It's also worth noting that we currently have no void functions in our db (unless it didn't appear in search, in which case, my bad!), so I guess my response to "is there a reason not to use void functions" is "is there a reason *to* use void functions here?" If so, then I don't mind these staying as void functions.

scheduler-server/sql/scheduler/tables/schema_migrations.sql
Mythicaeda: Additional thought: AFAIK moving things out of the public schema is something that's on our roadmap, so this should probably be moved this to it's own "migrations" schema or something of the like, so that it doesn't have to be migrated out of `public` later.

mattdailis:
Closing as inactive - I'll reopen this when it's ready for review again

PR #531, opened 2022-12-09
Tree-structured plans for simulation

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description

This PR is based on #511, so please ignore any commits that are also in that PR. (At the time of writing, only the final commit is really owned by this PR.)

This prototype describes my proposed method of supporting activity-relative scheduling at simulation-time. The key insight is that *modeled* activities can *already* perform relative scheduling, by using the `call()` action to wait for another task to complete. We leverage this to simplify the top-level API surface of the simulation driver to a single top-level task that recursively decomposes itself into the constituents of the plan.

This is already something we've wanted to do for a while (since it leads toward making the SimulationEngine simpler to use), but it also happens to make activity-relative scheduling nearly trivial, since we're just exposing a pre-existing capability. Of course, this prototype says nothing about the representation of relatively-scheduled activities in the database; at this time, it just provides a simple `ActionTree.from()` helper to take the existing data of a plan and convert it into an `ActionTree`.


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/ActionTree.java
```diff
@@ -0,0 +1,186 @@
...
+
+    return new ActionTree(self.action(), startAnchor, self.endAnchor());
+  }
+
+  /**
+   * Lift as many descendant actions as possible into the parent's anchor, and return an ActionTree with all remaining
+   * actions lifted as high as possible.
+   *
+   * @param anchor
+   *   The anchor of the parent which this action tree is scheduled relative to.
```
Twisol: @Mythicaeda mentioned finding this line difficult to grasp.

Since this is a recursive method, I recommend understanding `anchor` in context of one of its call sites. [Lines 173-180](https://github.com/NASA-AMMOS/aerie/pull/531/commits/c96fc885285fb1ae9c1d07579e455bd461f83b0f#diff-fdc43d758da16322322b30bba1f8078d2c7891a41e22ba7a469615473b2130abR173-R180) show where we pass on a different value for `anchor` than what we received. In that block, since we can't resolve through a directive, the new "best place" to move actions to is the current end anchor.

Twisol:
I pushed a second commit onto this PR which gets rid of `resolveOffsets`, melding its logic into `toTask` directly. This should in principle make it more efficient (no need to create a whole new copy of the plan), and hopefully also easier to understand (less parent-pointer recursion magic).

Unfortunately, it's also less easy to test -- instead of transforming data into data, we've transformed data into a program, and the only way to verify a program is to run it. (At least with `resolveOffsets`, we could directly write down and compare against the resolved tree we expect.)

camargo:
Closing this for now since we are going a different direction for `1.1.0`. But we will keep this branch around for future work and inspiration.

Twisol:
Even if we go for a different high-level approach to relative scheduling of activities, I do think this branch provides lower-level improvements to the simulation engine API. Would you mind if I kept working on this in that direction?

camargo:
@Twisol Not at all! Just trying to keep the open PRs active.

PR #533, opened 2022-12-12
Make negative delays throw exceptions

### LINKED ISSUE #526: Negative delay moves simulation time backward
Labeling this as a bug - but maybe it's a feature...? 😅

## Description
Type-wise, everything is fine - delay is defined to work on Duration, and that includes negative durations. However, I think this violates other invariants of the system - such as the invariant that the start offsets of resource profile segments are monotonically increasing.

`TaskStatus.Delayed` is perfectly happy to accept a negative duration:
https://github.com/NASA-AMMOS/aerie/blob/ad2e9c4130a09afea401daf15b3d0dbf7ccd801b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/TaskStatus.java#L10

The SimulationEngine is perfectly happy to schedule that task to be resumed in the past:
https://github.com/NASA-AMMOS/aerie/blob/ad2e9c4130a09afea401daf15b3d0dbf7ccd801b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java#L236-L238

This of course puts that task at the front of the job queue, so the next call to `extractNextJobs` contains this task, as well as the time (in the past) at which this task is scheduled.

The "delta" computation in the simulation driver will compute a negative delta and add it to the timeline: https://github.com/NASA-AMMOS/aerie/blob/ad2e9c4130a09afea401daf15b3d0dbf7ccd801b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java#L73-L75

Interestingly - the timeline keeps a perfect record of all this - if time travel were somehow semantically acceptable, the timeline would capture it faithfully (particularly - it faithfully captures the direction of causality even when time steps backwards).

## Possible solutions
1. We can throw an error and tank simulation as soon as a negative delay is requested
2. We can clamp negative durations to ZERO - and perhaps issue a warning to alert the client that we've done this
3. (Far-fetched) We can try to give a reasonable semantics to time travel

mattdailis:
* **Tickets addressed:** Fixes #526
* **Review:** By commit  

## Description

This PR adds a check that when a Task delays, it does so with a non-negative duration. This use to be the behavior of the simulation engine, until it got inadvertently lost during a refactoring.

## Verification

I did write a unit test for this, but did not include it in this PR because I wasn't thrilled with how it turned out. This gist contains that test: https://gist.github.com/mattdailis/1f0d747a5505f0a10db23b37449c7e41

## Documentation

- [ ] TODO make a ticket to document this behavior

## Future work

None

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -99,6 +99,8 @@ public final class SimulationEngine implements AutoCloseable {
 
   /** Schedule a new task to be performed at the given time. */
   public <Return> TaskId scheduleTask(final Duration startTime, final TaskFactory<Return> state) {
+    if (startTime.isNegative()) throw new IllegalArgumentException("Cannot schedule a task before the start time of the simulation");
+
     final var task = TaskId.generate();
     this.tasks.put(task, new ExecutionState.InProgress<>(startTime, state.create(this.executor)));
     this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(startTime));
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -27,7 +27,6 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.lang3.tuple.Triple;
 
-import java.lang.reflect.InvocationTargetException;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
@@ -234,6 +233,8 @@ public final class SimulationEngine implements AutoCloseable {
       this.tasks.put(task, progress.completedAt(currentTime, children));
       this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(currentTime));
     } else if (status instanceof TaskStatus.Delayed<Return> s) {
+      if (s.delay().isNegative()) throw new IllegalArgumentException("Cannot schedule a task in the past");
+
       this.tasks.put(task, progress.continueWith(s.continuation()));
       this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(currentTime.plus(s.delay())));
     } else if (status instanceof TaskStatus.CallingTask<Return> s) {


Twisol:
> I did write a unit test for this, but did not include it in this PR because I wasn't thrilled with how it turned out. This gist contains that test: https://gist.github.com/mattdailis/1f0d747a5505f0a10db23b37449c7e41

yet more evidence that our simulation API is hot garbage. (it's my fault :) )

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -234,6 +233,7 @@ private <Return> void stepEffectModel(
       this.tasks.put(task, progress.completedAt(currentTime, children));
       this.scheduledJobs.schedule(JobId.forTask(task), SubInstant.Tasks.at(currentTime));
     } else if (status instanceof TaskStatus.Delayed<Return> s) {
+      if (s.delay().isNegative()) throw new IllegalArgumentException("Cannot schedule a task in the past");
```
Twisol, line 238: Very minor nit: a newline after this line would be nice!

Twisol (approved):
Thanks for the fix, and _mea culpa_ -- based on our digging, this happened when we significantly refactored the engine over a year ago 😅

Twisol:
> fix/prevent-time-travel

excellent branch name

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
cohansen: +1, but I'm approving as well!

cohansen (approved):

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Twisol, line 238: Can we add this check to `SimulationEngine.scheduleTask` as well?

mattdailis: Done :+1:

mattdailis: I added an analogous check to `scheduleTask` - it's a mildly different situation, since the Simulation Engine doesn't know the current time (side note - should it?), so instead of saying "the past", I tried to give it a message that conveyed exactly what it was checking (and had different text from the other message in case we need to debug-by-grep):
> Cannot schedule a task before the start time of the simulation

Twisol: > since the Simulation Engine doesn't know the current time (side note - should it?)

Mmh. I remember thinking about this now. I don't know; it probably should, but I think this gets implicated alongside all the other problems with the current simulation API.

camargo (approved):

PR #544, opened 2022-12-19
Add plan id and sim dataset id to seq JSON

### LINKED ISSUE #504: Expanded seqJSON should include metadata to trace back to source activity
- Add metadata fields to seqJSON header to indicate planID, simulation data set ID
- Add metadata per command to indicate simulated activity instance ID

cohansen:
* **Tickets addressed:** Closes #504.
* **Review:** By commit

## Description
Automatically adds `planId` and `simulationDatasetId` to the top level sequence metadata.

## Verification
Double checked in the UI and updated all the tests to include the new metadata properties.

## Documentation
NA

## Future work
NA


sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
```diff
@@ -171,6 +175,9 @@ export interface GraphQLSimulatedActivityInstance<
   ActivityArguments extends Record<string, unknown> = Record<string, unknown>,
   ActivityComputedAttributes extends Record<string, unknown> = Record<string, unknown>,
 > {
+  activity_directive: {
```
camargo: So even tho this is technically correct I wonder if it makes more sense to get the `plan_id` from `simulation_dataset.simulation.plan.id` since ultimately the sequence is generated for a given simulation dataset.

cohansen: Yeah I wasn't sure where the best place to get the `plan_id` was from, but I can update this!

sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
```diff
@@ -192,7 +199,14 @@ export function mapGraphQLActivityInstance(
   activityInstance: GraphQLSimulatedActivityInstance<any, any>,
   activitySchema: GraphQLActivitySchema,
 ): SimulatedActivity {
+  if (!activityInstance.duration) {
+    throw new Error('Plan does not have a valid duration');
```
dyst5422: This is a single activity instance, not the entire plan, so the message should be updated.

Furthermore, I think we should make sure that a failure here does not fail all expansion, but only the one expansion with the offending activity. Partial success is hugely important for these computer-assisted operator tools as the human skills to perform without them diminish over time and a tool blocking operations is very bad.

cohansen: Removed this for now, I'll create a ticket to fix this later.

camargo (approved):
Tested locally, works great. LGTM!

PR #546, opened 2022-12-20
ForEach activity template parameter from anchor

### LINKED ISSUE #443: forEach activity relative parameter querying
When using the `forEach` activity operator in scheduling goals, if you schedule one activity with respect to another we need to query the parameters of the anchor activity such that they can have the same parameters. Perfect use case is the target parameter. Profile or spans could be a solution here.

### LINKED ISSUE #445: Choose the parameter of a “to be scheduled activity” based on a resource
For scheduling goals we should be able to choose the parameter of a “to be scheduled activity” based on a resource. 

1. How do we reference the start time of another activity? Always reference the start and end time of some window. 
2. What does the logic statement for selecting activity parameters based on resource value look like? Constraint profiles can be a solution here.

adrienmaillard:
Resolves #443
Resolves #445

* **Review:** By commit  

### Features 
- The parameters in activity templates passed to a coexistence goal can now be assigned to values from the anchoring activity. Approach here is similar to the `ForEachActivity` pattern where an alias is defined for the coexistence goal and the activity template is a function of an `ActivityInstance` (from the constraints eDSL) with this alias as a parameter. This allows to express the reference to the anchoring activity to the backend and to evaluate expressions. 
- More generally, the parameters in activity templates can now be assigned to expressions producing a parameter of the desired type. And allow mixed definition in the case of complex data types (a subfield of a subfield can be an expression while the rest stay absolute values). Commit 3 add the building blocks for hierarchy of expressions. What happen is that even if the user is providing an absolute value, the activity template constructor is turning it into an Expression<Profile> that will be evaluated. Activity template do not directly contain `SerializedValue` anymore. 
- The `valueAt(profile, spans):DiscreteProfile` is a new expression returning the value of a profile at a specific time. In this case, the passed spans must be a single instant. I believe this solves #445.

### Fixes/changes:
- `AssignGaps` now evaluates on the forever horizon. 
- Discrete and Real activity parameters expressions are defined on the forever horizon. Queried at a time outside the activity bounds, they can return the desired parameter. This allows instantiating a new activity that is not overlapping with an anchoring activity but references a parameter of the anchor activity. 
- `ActivityInstance` (in scheduler-driver) has been refactored into a record. 
- Obsolete constraints have been removed and replaced with appropriate global constraints   
- `CoexistenceGoal` now takes an `Expression<Spans>` as input
- `ActivityExpression` now implements `Expressions<Spans>` so it can be used by `CoexistenceGoal`
- Because `CoexistenceGoal` takes an `Expression<Spans>`, `AerieLander` tests have been commented. The constraints does not have the necessary expressive power yet to express what was expressed in these tests. 
## Verification

A few tests were added but more are needed. I just want this PR to start. 

## Documentation

Probably quite a lot of scheduling doc has to be updated. 

## Future work

- Better representation of timepoints in the eDSL (some kind of DurationExpression) instead of Spans. 
- ?


camargo:
@adrienmaillard provided the following examples for the Banananation model:

```ts
export default () =>
  Goal.CoexistenceGoal({
    activityTemplate: growBananaActivity =>
      ActivityTemplates.PickBanana({ quantity: growBananaActivity.parameters.quantity }),
    startsAt: TimingConstraint.singleton(WindowProperty.END).plus(Temporal.Duration.from({ minutes: 5 })),
  });
```

```ts
export default () =>
  Goal.CoexistenceGoal({
    activityTemplate: growBananaActivity =>
      ActivityTemplates.ChangeProducer({
        producer: Discrete.Resource('/producer').valueAt(growBananaActivity.span().starts()),
      }),
    startsAt: TimingConstraint.singleton(WindowProperty.END).plus(Temporal.Duration.from({ minutes: 5 })),
  });
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Profile.java
```diff
@@ -1,11 +1,15 @@
...
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 
 public interface Profile<P extends Profile<P>> {
   Windows equalTo(P other);
   Windows notEqualTo(P other);
   Windows changePoints();
 
   P assignGaps(P def);
+
+  SerializedValue valueAt(Duration timepoint);
```
JoelCourtney: I think instead of throwing an error, it should return `Optional<SerializedValue>`, so that the consumer can decide if that's really a problem or not. In this case it is, but who knows for future use cases.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -157,6 +160,16 @@ public Spans split(final Interval bounds, final int numberOfSubSpans, final Incl
     });
   }
 
+  public boolean isSubsetOf(final Spans otherSpans){
+    return new HashSet<>(otherSpans.intervals).containsAll(this.intervals);
+  }
```
JoelCourtney: I'm not certain this is the behavior you want. What if one span in `this` is a subinterval of a span in `otherSpans`? Should that not count?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Windows.java
```diff
@@ -465,10 +466,12 @@ public Windows ends() {
...
    */
   public Spans intoSpans(final Interval bounds) {
-    final var firstInterval = this.segments.get(0).interval();
-    final var lastInterval = this.segments.get(-1).interval();
     boolean boundsStartContained = false;
     boolean boundsEndContained = false;
+    if(this.segments.size() == 1){
+      if (segments.get(0).interval().contains(bounds.start)) boundsStartContained = true;
+      if (segments.get(0).interval().contains(bounds.end)) boundsEndContained = true;
+    }
```
JoelCourtney, line 474: good catch

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -20,7 +23,7 @@
 public class Spans implements IntervalContainer<Spans>, Iterable<Segment<Optional<Spans.Metadata>>> {
   private final List<Segment<Optional<Metadata>>> intervals;
 
-  public record Metadata(long id){}
+  public record Metadata(long id, String type, Map<String, SerializedValue> arguments){}
```
JoelCourtney: Maybe we should just have a reference to the activity instance, instead of breaking up all of its fields?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/AssignGaps.java
```diff
@@ -14,8 +13,8 @@
 
   @Override
   public P evaluate(final SimulationResults results, final Interval bounds, final EvaluationEnvironment environment) {
-    final var originalProfile = this.originalProfile.evaluate(results, bounds, environment);
-    final var defaultProfile = this.defaultProfile.evaluate(results, bounds, environment);
+    final var originalProfile = this.originalProfile.evaluate(results, Interval.FOREVER, environment);
+    final var defaultProfile = this.defaultProfile.evaluate(results, Interval.FOREVER, environment);
```
JoelCourtney: With the change to make `split` not care about things outside the plan bounds, is this still necessary?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/DiscreteParameter.java
```diff
@@ -22,7 +22,7 @@ public DiscreteParameter(final String activityAlias, final String parameterName)
   public DiscreteProfile evaluate(final SimulationResults results, final Interval bounds, final EvaluationEnvironment environment) {
     final var activity = environment.activityInstances().get(this.activityAlias);
     return new DiscreteProfile(
-        Segment.of(activity.interval, activity.parameters.get(this.parameterName))
+        Segment.of(Interval.FOREVER, activity.parameters.get(this.parameterName))
```
JoelCourtney, line 25: ahhhhh good point. There go my dreams of not having to use `.if(activity.window())` on activity parameter constraints. Oh well

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ListExpressionAt.java
```diff
@@ -0,0 +1,51 @@
...
+import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.IntervalMap;
+import gov.nasa.jpl.aerie.constraints.time.Segment;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+public record ListExpressionAt(
```
JoelCourtney, line 15: I think this and the struct expression node deserve a doc comment explaining why it uses `valueAt`, because its not obvious at first why it can't just zip the profiles together until you take Real profiles into account.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ListExpressionAt.java
```diff
@@ -0,0 +1,51 @@
...
+    List<ProfileExpression<?>> elements
+) implements Expression<DiscreteProfile> {
+
+  @Override
+  public DiscreteProfile evaluate(final SimulationResults results, final Interval bounds, final EvaluationEnvironment environment) {
+    final var retList = new ArrayList<SerializedValue>();
```
JoelCourtney, line 28: Please put an assertion/error here if the bounds are not singleton. This node is only intended to be used when the bounds are a single point. (same for the Struct version)

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ValueAt.java
```diff
@@ -0,0 +1,66 @@
...
+import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.IntervalMap;
+import gov.nasa.jpl.aerie.constraints.time.Segment;
+import gov.nasa.jpl.aerie.constraints.time.Spans;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+
+import java.util.Set;
+
+public record ValueAt(
```
JoelCourtney: I don't understand why this needs to be a node. Will the user ever use it directly?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ValueAt.java
```diff
@@ -0,0 +1,66 @@
...
+
+
+  public static SerializedValue valueAt(Profile<?> profile, Duration time){
+    if(profile instanceof DiscreteProfile discreteProfile){
+      return discreteProfile.valueAt(time);
+    } else if (profile instanceof LinearProfile linearProfile){
+      return linearProfile.valueAt(time);
+    } else{
+      throw new Error("Profile is neither discrete or linear");
+    }
```
JoelCourtney: I think this is unnecessary, since the Profile interface already has the valueAt function. In `evaluate` you can just call `res.valueAt(timepoint)`.

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
```diff
@@ -25,3 +25,10 @@ export const ActivityTypeParameterMap = {
     })
   }),
 }
+
+export type ParameterTypeactivity = {
+  parameter: string
+}
+export type ActivityTypeParameterMap = {
+  [ActivityType.activity]: ParameterTypeactivity
+};
```
JoelCourtney, line 34: two things:
- I don't think this is used in constraints, only scheduling. So shouldn't these types go in scheduling instead? I don't like how we have such a big distinction between the two, but as long as we have the distinction we might as well abide by it.
- If the `ParameterType<activity>` type isn't going to be used directly by itself, I'd prefer to inline it into `ActivityTypeParameterMap` to avoid polluting the namespace.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ValueAt.java
JoelCourtney: Given the compilation service test you wrote, it looks like this won't ever be used by the user.

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
JoelCourtney: Or if you don't want to inline the types, you can just not export them.

JoelCourtney (changes requested):
Here's a review of the constraints side of things. I won't have time to go through all of the scheduler things today. Nice work!

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
adrienmaillard: 1) a) Okay, I'll move it. b) I tried inlining but I could not make it work, that's why it is this way. Will try again now that pieces are not moving so much around. 
2) Okay

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/AssignGaps.java
adrienmaillard: Has this change been merged ? If yes, I can test.

JoelCourtney: Yes, that was merged

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Profile.java
adrienmaillard: Absolutely.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ValueAt.java
adrienmaillard: in this goal for example : 
```export default () =>
  Goal.CoexistenceGoal({
    activityTemplate: growBananaActivity =>
      ActivityTemplates.ChangeProducer({
        producer: Discrete.Resource('/producer').valueAt(growBananaActivity.span().starts()),
      }),
    startsAt: TimingConstraint.singleton(WindowProperty.END).plus(Temporal.Duration.from({ minutes: 5 })),
  });```

File: docs/buildExternalDocs.bash [hunk 12]
```diff
@@ -80,6 +80,7 @@ do
   [ -e "$file" ] || continue
   tail -n +3 "$file" > "$file.tmp" && mv "$file.tmp" "$file"
   sed -i -e 's/README/index/g' $file
+  sed -i -e 's,../modules/Constraint_eDSL.,../../constraints-edsl-api/modules/,g' $file
```
Mythicaeda, line 85: Can this line be added to the processing done in `./source/scheduling-edsl-api/enums/` and `./source/scheduling-edsl-api/classes/`?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ValueAt.java
JoelCourtney: Oh I was imagining that coexistence goal execution would automatically get the value at when the activity starts. So you would just put `Discrete.Resource('/producer')`.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
adrienmaillard: Actually, in that precise case, I believe this is the behavior I want. The spans for coexistence can only either be the original ones or disappear one by one strictly (without leaving any subinterval) as new activities are anchored to each of them.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/AssignGaps.java
adrienmaillard: It is not necessary, I have removed it.

merlin-server/constraints-dsl-compiler/src/libs/mission-model-generated-code.ts
adrienmaillard: I tried and this is going to take time, I'd rather not do it now.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: Sorry for the late response. I'm OK with this behavior then, but I think it should have a different name and a doc comment because it could be misleading. Maybe `.isCollectionSubsetOf`? I'm not sure. I'm happy as long as its a weird enough name to force people to read the doc comment.

adrienmaillard: Done.

JoelCourtney (approved):
noice

PR #549, opened 2022-12-20
Run Prettier formatting over sequencing server

camargo:
## Description

* Also update NPM format script to include test directory


cohansen (approved):
Thanks, looks good!

PR #553, opened 2022-12-21
Move Temporal.Duration to constraints eDSL

### LINKED ISSUE #501: Add Temporal.Duration in Constraints eDSL

JoelCourtney:
* **Tickets addressed:** closes #501 
* **Review:** By commit  

## Description
- adds `Temporal` to the constraints edsl
- removes `Temporal` to the scheduling edsl
- updates the user code runner version number to 0.5.0
- removes the weird hot comment at the top of the `...-ast.ts` files
    - No tests fail, so I'm not sure what the point was, but it solved a very weird bug.

In other news, async functions have been supported in the user code runner in 0.5.0 for a while now, but now that we're upgrading it should be available to users.

## Verification
tests are updated

## Documentation
Will make sure auto generated and written docs are up to date with the change before merging. There is a chance that the generated docs will not be able to link to Temporal.


JoelCourtney:
tests are failing because the package-lock changes used artifactory. Will fix tomorrow

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -384,7 +384,9 @@ export default function() {
               ),
               new SchedulingDSL.ConstraintExpression.ActivityExpression("SampleActivity2"),
               Optional.of(new SchedulingDSL.ActivityTimingConstraint(TimeAnchor.START, TimeUtility.Operator.PLUS, SECOND, true)),
-              Optional.empty()
+              Optional.empty(
+
+              )
```
adrienmaillard: Why this formatting change ?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -150,7 +150,7 @@ export default function myConstraint() {
              return Real.Resource("state of charge");
           }
         """,
-        "TypeError: TS2322 Incorrect return type. Expected: 'Constraint', Actual: 'Real'."
+        "TypeError: TS2322 Incorrect return type. Expected: 'Constraint | Promise<Constraint>', Actual: 'Real'."
```
adrienmaillard, line 153: Can you described a case where the return would be a `Promise<Constraint>`, I am getting it.

adrienmaillard (approved):
Glad you could succeed at this, I can't find exactly what you did different from when we looked at it.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
JoelCourtney: not sure how that happened, will undo

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
JoelCourtney: TBH I'm not sure when our users would ever do that. The only example I can think of is if they queried some data from another server and needed to await it.

This change is just because we should be keeping up to date with the current aerie-ts-user-code-runner version, and not because it is needed for this PR. I was just messing with `package.json` and `package-lock.json` so much that I decided to update to 0.5.0.

dyst5422: Yeah, this is because with V0.4.0 of the user code runner, we enabled async returns. This message is simply reflecting that that is a valid return type. You can customize the error messaging on this with the message mapper passed to the constructor if you really want to remove the Promise<> part of the message https://github.com/NASA-AMMOS/aerie-ts-user-code-runner/blob/develop/src/UserCodeRunner.ts#L24

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -186,9 +186,10 @@ static JsonParser<Rate> rateF(final JsonParser<Expression<LinearProfile>> linear
 
   static final JsonParser<Duration> durationP =
       longP
-          . map(
-              microseconds -> Duration.of(microseconds, Duration.MICROSECONDS),
-              duration -> duration.in(Duration.MICROSECONDS));
+          .map(
```
mattdailis, line 189: I think this change ended up being spurious - looks like in the trip through ISO8601 and back we lost the `S`s at the end of `MICROSECOND`.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference path="./TemporalPolyfillTypes.ts"; />
```
mattdailis, line 1: Do we need this hot comment in `constraints-ast.ts`?

mattdailis (approved):
Good stuff :+1: Debugging our multi-lingual system is... not straightforward.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: I guess not? I'm confused as to why it was necessary. Before this PR we would get a Not Found error if this comment was removed - it was the first thing I tried when doing this task. But if I remove it now it is fine. Now this comment isn't in constraints or scheduling.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
JoelCourtney: I put the ~s's~ back
I put the ~~esses~~
the ~Ses~
~Ss~?

I undid the change

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
JoelCourtney: Actually no, this comment is required still. The tests failed, so I'm putting it back.

PR #559, opened 2022-12-22
Fix SPICE not loading on M1 Docker images

### LINKED ISSUE #541: Spice Fails to Load on M1 Macs during Simulation
## Summary:
When attempting to simulate a plan whose model loads SPICE, the following error occurs when SPICE is loaded:

```
java.lang.UnsatisfiedLinkError: /tmp/JNISpice-4439314142500713632.tmp: /tmp/JNISpice-4439314142500713632.tmp: cannot open shared object file: No such file or directory (Possible cause: can't load AMD 64 .so on a AARCH64 platform)
```

@Twisol and I have so far been discussing this in PR #322.

## How to Reproduce:
1) Build the Banananation model found on the `fix/spice_m1` branch
2) Compose up Aerie into Docker
3) Via the UI, upload the Banananation Model and create a new Plan from the model
4) Via the UI, attempt to simulate the new Plan

Mythicaeda:
* **Tickets addressed:** Fixes #541
* **Review:** By commit  

## Description

**Short explanation of the bug**: Our Docker containers consider their OS to be Linux and their Architecture to be that of the machine that compiled the image. Therefore, despite @Twisol creating a [universal Mac SPICE library](https://github.com/NASA-AMMOS/aerie/pull/322), when running simulation via the UI, if SPICE needed to be loaded, it would attempt to load the Linux-Intel library, which fails on Docker images created on M1 Macs (aka whenever one of us who have an M1 Mac runs `docker compose up`).

This PR adds a Linux-AArch `libJNISpice_M1.so`, and modifies SpiceLoader to, in the case of a Linux OS, pick between the Linux-Intel and Linux-AArch SPICE libraries based on the System's Architecture.

This PR also modifies Banananation to load Spice as part of the mission model constructor.

## Verification

I did the following on my M1 Mac, and @mattdailis did the same on his Intel Mac:
1. Checkout `fix/spice_m1`
2. `.gradlew clean assemble`
3. Compose up Aerie into Docker
4. Via the UI, upload the Banananation Model and create a new Plan from the model
5. Via the UI, attempt to simulate the new Plan

## Documentation

There are no docs changes, so instead I'm going to document here how I got the Linux-M1 SPICE build in case we need to repeat the steps in the future.

1. Download the [Linux JNISpice.tar.z from NAIF](https://naif.jpl.nasa.gov/pub/naif/misc/JNISpice/PC_Linux_GCC_Java1.8_64bit/packages/)
2. In the directory with `JNISpice.tar.z`, run:
```bash
docker run --name JNIBuilder --rm -it -v $PWD:/build ubuntu:latest
```
4. In a terminal on the host machine, in the directory with `JNISpice.tar.z`, run the following:
```bash
docker cp JNISpice.tar.z JNIBuilder:/JNISpice.tar.z
```
5. Inside the container's terminal, run the following: 
```csh
apt-get update
apt-get install -y openjdk-19-jdk gcc csh vim

uncompress JNISpice.tar
tar -xvf JNISpice.tar

cd JNISpice
csh
cd src/JNISpice
vim mkprodct.csh
```
Modify `set COMPILEOPTIONS` to this line: `set COMPILEOPTIONS   = "-c"`
Modify `set TKLINKOPTIONS` to this line: `set TKLINKOPTIONS = "-shared"`
Use `:wq` to exit Vim
```csh
csh mkjnijava.csh
cd ..
```
6. There are three options of what to do next:
    1. Go through each folder in `src` and in each `mkprodct.csh` change `set TKLINKOPTIONS` to this line: `set TKLINKOPTIONS= "-lm"` (this was what I did)
    2. Modify `../makeall.csh` to run `setenv TKLINKOPTIONS "-lm"` at the start of the script, then run `unsetenv TKLINKOPTIONS` before compiling JNISpice and then set it again to `"-lm"` after compiling JNISpice
    3. Rename TKLINKOPTIONS in `src/JNISpice/mkprodct.csh` so that you can set `TKLINKOPTIONS` in your environment
7. Inside the container's terminal, run the following:
```csh
cd ..
# After this, you should be in the top most JNISpice folder
setenv TKCOMPILEOPTIONS "-c -ansi -O2 -fPIC -DNON_UNIX_STDIO"
csh makeall.csh
```
8. In the terminal on the host machine, run:
```bash
docker cp JNIBuilder:/JNISpice/lib/libJNISpice.so .
```


camargo (approved):
Tested and this still works on x86. LGTM.

cohansen (approved):
Tested as well on x86, still worked fine.

JoelCourtney:
Tested on M1 arm, works

Twisol:
By any chance, were you able to confirm that consecutive loads of the model get fresh instances of SPICE? To do this, I normally load a kernel with FURNSH and then print the result of KTOTAL("ALL"); if the number doesn't grow with each loaded model, they should be isolated.

Twisol:
Also, sorry for not being around to answer your questions, @Mythicaeda. I just had a root canal and two crowns after unimaginable pain :)

Mythicaeda:
It's alright. Unfortunately, I don't understand SPICE calls enough to know how/where to do this check. I do know, however, that the `flat-namespace` flag that was causing issues on the `.jnilib` builds isn't used in the Linux build, and when I went through all the other compilation/link flags used, none of them would've caused similar functionality, which is why I didn't think to check it.

Twisol:
Yeah, that's true. Our customers are probably using a Linux-based environment too (either via Docker or AWS or whatnot), so ideally they'd have seen strangeness since the first feature rollout if it wasn't working. Since this is just an M1 build of the same artifact, hopefully it has the same overall dynamic linking behavior.

PR #566, opened 2023-01-03
Add multi-arch image builds in `publish` workflow

### LINKED ISSUE #555: Support multi-arch Docker images for all Aerie images
See: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
We should use https://github.com/docker/buildx in our publish action workflow.

Just noting eventually we also need to support M1 chipsets which is [not currently supported](https://github.com/actions/runner-images/issues/2187).

skovati:
* **Tickets addressed:** closes #555 
* **Review:** By commit  

## Description

Adds docker buildx support to our github `publish` action, which allows us to build multi-arch images, i.e. images that automagically work on both amd64 and arm64. This allows our images to easily run on [arm-based cloud VMs](https://aws.amazon.com/ec2/graviton/), as well as removes the need for M1 mac users to virtualize amd64 (though they still need to virtualize linux).

Additionally bumped the versions for most actions in the workflow, since those use deprecated versions of `node`. This should remove the ~50 warnings we get every time this workflow runs :laughing: 

## Verification

Unfortunately, github actions are hard to validate when only run on merge to `develop` but there are no breaking changes on version updates, and buildx should be able to push multi-arch images directly to `ghcr` according to their docs, since all our base images are also multi-arch

## Documentation

N/A

## Future work

N/A

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,6 +45,8 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       # Build, scan, and push aerie-hasura Docker artifacts.
       - name: Extract metadata (tags, labels) for aerie-hasura Docker image
         id: aerieHasura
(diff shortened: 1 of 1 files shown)


camargo:
Awesome!! Thanks for the quick turnaround on this. Can you also update the UI publish workflow and Gateway publish workflow with the same changes?

camargo (approved):

PR #570, opened 2023-01-04
Anchors

### LINKED ISSUE #449: Make activity directive start time relative to another activity directive start time
We need to be able to specify the start time of an activity directive relative to the start time of another activity directive plus or minus some offset.

Mythicaeda:
* **Tickets addressed:** Closes #449
* **Review:** By commit  

## Description

This is the database changes to allow for anchors.

Changes to DB: 
- adds the `anchor_ID` and `anchored_to_start` columns to `activity_directives`, `plan_snapshot_activities`, and `merge_staging_area`.
- adds a new table, `anchor_validation_status`. The `reason_invalid` column contains the reason why an activity's anchor is invalid.
- removes the "start time offsets must be nonnegative" constraints on tables `activity_directives` and `merge_staging_area`
- adds new constraint "anchor_in_plan" to `activity_directives`
- adds new functions in `hasura_functions` to resolve how to handle any activities anchored to an activity getting deleted

Note: `validate_anchors` raises INFO-level notes when it finds a invalid anchor that doesn't throw an exception (right before it writes to `anchor_validation_status`).

## Verification

- DatabaseTestHelper was modified to give `aerie` USAGE access on all schemas in a test database. Prior to this, it only had access to the `public` schema
- AnchorTests was created
- PlanCollaborationTests was modified:
     - Anchor-related fields are also tracked and checked during existing tests
     - AnchorMergeTests were added to test edge cases that can occur from merging plans

## Future work

- Updates to the simulation engine to use the relative offsets

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
@@ -55,7 +55,7 @@ public class DatabaseTestHelper {
     // Apparently, the previous privileges are insufficient on their own
     {
       final var pb = new ProcessBuilder("psql",
-                                        "postgresql://postgres:postgres@localhost:5432/" + dbName,
+                                        "postgresql://aerie:aerie@localhost:5432/" + dbName,
                                         "-v", "ON_ERROR_STOP=1",
                                         "-c", "ALTER DEFAULT PRIVILEGES GRANT ALL ON TABLES TO aerie;",
                                         "-c", "\\ir %s".formatted(initSqlScriptFile.getAbsolutePath())
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
@@ -2562,4 +2562,155 @@ public class PlanCollaborationTests {
       }
     }
   }
+
+  @Nested
+  class AnchorMergeTests{
+
+    // Method is taken from AnchorTests
+    private void setAnchor(int anchorId, boolean anchoredToStart, int activityId, int planId) throws SQLException {
+      try(final var statement = connection.createStatement()) {
+        if (anchorId == -1) {
+          statement.execute(
+              """
+                  update activity_directive
+                  set anchor_id = null,
+                      anchored_to_start = %b
+                  where id = %d and plan_id = %d;
+                  """.formatted(anchoredToStart, activityId, planId));
+        } else {
+          statement.execute(
+              """
+                  update activity_directive
+                  set anchor_id = %d,
+                      anchored_to_start = %b
+                  where id = %d and plan_id = %d;
+                  """.formatted(anchorId, anchoredToStart, activityId, planId));
+        }
+      }
+    }
+
+    @Test
+    void cantMergeCycle() throws SQLException{
+      final int planId = insertPlan(missionModelId);
+      final int activityA = insertActivity(planId);
+      final int activityB = insertActivity(planId);
+
+      final int childPlan = duplicatePlan(planId, "Cycle Test Plan");
+
+      // Plan chain: B -> A
+      setAnchor(activityA, true, activityB, planId);
+      // ChildPlan chain: A -> B
+      setAnchor(activityB, true, activityA, childPlan);
+
+      // Merge fails as it would establish B -> A -> B cycle
+      final int mergeRQ = createMergeRequest(planId, childPlan);
+      beginMerge(mergeRQ);
+      try {
+        commitMerge(mergeRQ);
+        fail();
+      } catch (SQLException ex) {
+        if(!ex.getMessage().contains("Cycle detected. Cannot apply changes.")){
+          throw ex;
+        }
+      }
+    }
+
+    @Test
+    void anchorMustBeInTargetPlanAtEndOfMerge() throws SQLException{
+      // Can't merge in a version of an activity that is anchored to an activity that is deleted from the target plan.
+      final int planId = insertPlan(missionModelId);
+      final int activityA = insertActivity(planId);
+      final int activityB = insertActivity(planId);
+
+      final int childPlan = duplicatePlan(planId, "Anchor Delete Test");
+      setAnchor(activityA, true, activityB, childPlan);
+
+      deleteActivityDirective(planId, activityA);
+
+      final int mergeRQ = createMergeRequest(planId, childPlan);
+      beginMerge(mergeRQ);
+      try{
+        commitMerge(mergeRQ);
+        fail();
+      } catch (SQLException ex){
+        if(!ex.getMessage().contains(
+            "insert or update on table \"activity_directive\" violates foreign key constraint \"anchor_in_plan\"")){
+          throw ex;
+        }
+      }
+
+    }
+
+    @Test
+    void deleteSubtreeDoesNotImpactRelatedPlans() throws SQLException{
+      final int planId = insertPlan(missionModelId);
+      final int activityAId = insertActivity(planId);
+      final int activityBId = insertActivity(planId);
+      setAnchor(activityAId, true, activityBId, planId);
+      final int childPlanId = duplicatePlan(planId, "Delete Chain Test");
+
+      final Activity activityABefore = getActivity(childPlanId, activityAId);
+      final Activity activityBBefore = getActivity(childPlanId, activityBId);
+
+      try(final var statement = connection.createStatement()) {
+        statement.execute(
+            """
+             select hasura_functions.delete_activity_by_pk_delete_subtree(%d, %d)
+             """.formatted(activityAId, planId));
+      }
+      assertEquals(0, getActivities(planId).size());
+      assertEquals(2, getActivities(childPlanId).size());
+      assertActivityEquals(activityABefore, getActivity(childPlanId, activityAId));
+      assertActivityEquals(activityBBefore, getActivity(childPlanId, activityBId));
+    }
+
+    @Test
+    void deletePlanReanchorDoesNotImpactRelatedPlans() throws SQLException{
+      final int planId = insertPlan(missionModelId);
+      final int activityAId = insertActivity(planId);
+      final int activityBId = insertActivity(planId);
+      setAnchor(activityAId, true, activityBId, planId);
+      final int childPlanId = duplicatePlan(planId, "Delete Chain Test");
+
+      final Activity activityABefore = getActivity(childPlanId, activityAId);
+      final Activity activityBBefore = getActivity(childPlanId, activityBId);
+
+      try(final var statement = connection.createStatement()) {
+        statement.execute(
+            """
+             select hasura_functions.delete_activity_by_pk_reanchor_plan_start(%d, %d)
+             """.formatted(activityAId, planId));
+      }
+      assertEquals(1, getActivities(planId).size());
+      assertEquals(2, getActivities(childPlanId).size());
+      assertActivityEquals(activityABefore, getActivity(childPlanId, activityAId));
+      assertActivityEquals(activityBBefore, getActivity(childPlanId, activityBId));
+    }
+    @Test
+    void deleteActivityReanchorDoesNotImpactRelatedPlans() throws SQLException{
+      final int planId = insertPlan(missionModelId);
+      final int activityAId = insertActivity(planId);
+      final int activityBId = insertActivity(planId);
+      final int activityCId = insertActivity(planId);
+      setAnchor(activityAId, true, activityBId, planId);
+      setAnchor(activityCId, true, activityAId, planId);
+      final int childPlanId = duplicatePlan(planId, "Delete Chain Test");
+
+      final Activity activityABefore = getActivity(childPlanId, activityAId);
+      final Activity activityBBefore = getActivity(childPlanId, activityBId);
+      final Activity activityCBefore = getActivity(childPlanId, activityCId);
+
+      try(final var statement = connection.createStatement()) {
+        statement.execute(
+            """
+             select hasura_functions.delete_activity_by_pk_reanchor_to_anchor(%d, %d)
+             """.formatted(activityAId, planId));
+      }
+      assertEquals(2, getActivities(planId).size());
+      assertEquals(3, getActivities(childPlanId).size());
+      assertActivityEquals(activityABefore, getActivity(childPlanId, activityAId));
+      assertActivityEquals(activityBBefore, getActivity(childPlanId, activityBId));
+      assertActivityEquals(activityCBefore, getActivity(childPlanId, activityCId));
+    }
+  }
 }
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
@@ -31,27 +31,11 @@ public class PlanCollaborationTests {
   private Connection connection;
   int fileId;
   int missionModelId;
-  int planId;
-  int activityId;
-  int simulationTemplateId;
-  int simulationWithTemplateId;
-  int simulationWithoutTemplateId;
-  int datasetId;
-  gov.nasa.jpl.aerie.database.SimulationDatasetRecord simulationDatasetRecord;
-  gov.nasa.jpl.aerie.database.PlanDatasetRecord planDatasetRecord;
 
   @BeforeEach
   void beforeEach() throws SQLException {
     fileId = insertFileUpload();
     missionModelId = insertMissionModel(fileId);
-    planId = insertPlan(missionModelId);
-    activityId = insertActivity(planId);
-    simulationTemplateId = insertSimulationTemplate(missionModelId);
-    simulationWithTemplateId = insertSimulationWithTemplateId(simulationTemplateId, planId);
-    simulationWithoutTemplateId = insertSimulationWithoutTemplateId(planId);
-    planDatasetRecord = insertPlanDataset(planId);
-    datasetId = insertDataset();
-    simulationDatasetRecord = insertSimulationDataset(simulationWithTemplateId, datasetId);
   }
 
   @AfterEach
@@ -124,10 +108,6 @@ public class PlanCollaborationTests {
   }
 
   int insertPlan(final int missionModelId) throws SQLException {
-    return insertPlan(missionModelId, "2020-1-1 00:00:00");
-  }
-
-  int insertPlan(final int missionModelId, final String start_time) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -135,7 +115,7 @@ public class PlanCollaborationTests {
                   INSERT INTO plan (name, model_id, duration, start_time)
                   VALUES ('test-plan-%s', '%s', '0', '%s')
                   RETURNING id;"""
-                  .formatted(UUID.randomUUID().toString(), missionModelId, start_time)
+                  .formatted(UUID.randomUUID().toString(), missionModelId, "2020-1-1 00:00:00")
           );
       res.next();
       return res.getInt("id");
@@ -161,108 +141,17 @@ public class PlanCollaborationTests {
       return res.getInt("id");
     }
   }
-
-  int insertSimulationTemplate(final int modelId) throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                  INSERT INTO simulation_template (model_id, description, arguments)
-                  VALUES ('%s', 'test-description', '{}')
-                  RETURNING id;"""
-                  .formatted(modelId)
-          );
-      res.next();
-      return res.getInt("id");
-    }
-  }
-
-  int insertSimulationWithTemplateId(final int simulationTemplateId, final int planId) throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                  INSERT INTO simulation (simulation_template_id, plan_id, arguments)
-                  VALUES ('%s', '%s', '{}')
-                  RETURNING id;"""
-                  .formatted(simulationTemplateId, planId)
-          );
-      res.next();
-      return res.getInt("id");
-    }
-  }
-
-  int insertSimulationWithoutTemplateId(final int planId) throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                  INSERT INTO simulation (plan_id, arguments)
-                  VALUES ('%s', '{}')
-                  RETURNING id;"""
-                  .formatted(planId)
-          );
-      res.next();
-      return res.getInt("id");
-    }
-  }
-
-  int insertDataset() throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                    INSERT INTO dataset
-                    DEFAULT VALUES
-                    RETURNING id;"""
-          );
-      res.next();
-      return res.getInt("id");
-    }
-  }
-
-  gov.nasa.jpl.aerie.database.PlanDatasetRecord insertPlanDataset(final int planId) throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                  INSERT INTO plan_dataset (plan_id, offset_from_plan_start)
-                  VALUES ('%s', '0')
-                  RETURNING plan_id, dataset_id;"""
-                  .formatted(planId)
-          );
-      res.next();
-      return new gov.nasa.jpl.aerie.database.PlanDatasetRecord(res.getInt("plan_id"), res.getInt("dataset_id"));
-    }
-  }
-
-  gov.nasa.jpl.aerie.database.SimulationDatasetRecord insertSimulationDataset(final int simulationId, final int datasetId) throws SQLException {
-    try (final var statement = connection.createStatement()) {
-      final var res = statement
-          .executeQuery(
-              """
-                  INSERT INTO simulation_dataset (simulation_id, dataset_id, offset_from_plan_start)
-                  VALUES ('%s', '%s', '0')
-                  RETURNING simulation_id, dataset_id;"""
-                  .formatted(simulationId, datasetId)
-          );
-      res.next();
-      return new gov.nasa.jpl.aerie.database.SimulationDatasetRecord(
-          res.getInt("simulation_id"),
-          res.getInt("dataset_id"));
-    }
-  }
   //endregion
 
   //region Helper Methods
-  private boolean updateActivityName(String newName, int activityId, int planId) throws SQLException {
+  private void updateActivityName(String newName, int activityId, int planId) throws SQLException {
     try(final var statement = connection.createStatement()) {
-      return statement.execute(
-          """
-          update activity_directive
-          set name = '%s'
-          where id = %d and plan_id = %d;
-          """.formatted(newName, activityId, planId));
+      statement.execute(
+        """
+        update activity_directive
+        set name = '%s'
+        where id = %d and plan_id = %d;
+        """.formatted(newName, activityId, planId));
     }
   }
 
@@ -486,7 +375,9 @@ public class PlanCollaborationTests {
           res.getString("type"),
           res.getString("arguments"),
           res.getString("last_modified_arguments_at"),
-          res.getString("metadata")
+          res.getString("metadata"),
+          res.getString("anchor_id"),
+          res.getBoolean("anchored_to_start")
       );
     }
   }
@@ -514,7 +405,9 @@ public class PlanCollaborationTests {
             res.getString("type"),
             res.getString("arguments"),
             res.getString("last_modified_arguments_at"),
-            res.getString("metadata")
+            res.getString("metadata"),
+            res.getString("anchor_id"),
+            res.getBoolean("anchored_to_start")
         ));
       }
       return activities;
@@ -544,7 +437,9 @@ public class PlanCollaborationTests {
             res.getString("type"),
             res.getString("arguments"),
             res.getString("last_modified_arguments_at"),
-            res.getString("metadata")
+            res.getString("metadata"),
+            res.getString("anchor_id"),
+            res.getBoolean("anchored_to_start")
         ));
       }
       return activities;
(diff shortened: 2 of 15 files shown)


merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -13,15 +13,21 @@ create table activity_directive (
   last_modified_arguments_at timestamptz not null default now(),
   metadata merlin_activity_directive_metadata_set default '{}'::jsonb,
 
+  anchor_ID integer default null,
```
mattdailis: Minor naming convention comment - we tend to keep everything lowercase

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -13,15 +13,21 @@ create table activity_directive (
...
   constraint activity_directive_natural_key
     primary key (id, plan_id),
   constraint activity_directive_owned_by_plan
     foreign key (plan_id)
     references plan
     on update cascade
     on delete cascade,
-  constraint activity_directive_start_offset_is_nonnegative
-    check (start_offset >= '0')
+  -- An activity cannot anchor to an activity in another plan
```
mattdailis, line 25: I agree with this comment (I had not considered the possibility to anchor outside of the plan - I do agree that that kind of integrity should be enforced).

I'm having trouble seeing how the constraint below relates to this comment, could you help guide me through that?

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -68,6 +80,23 @@ $$;
 comment on procedure plan_locked_exception(plan_id integer) is e''
   'Verify that the plan corresponding to the activity being updated is unlocked, throwing an exception if not.';
 
+create procedure nothing_anchored_before_delete(activity_id integer)
```
mattdailis: I wonder if we can get this "restrict deletion" behavior by leveraging the foreign key, up on line 30. I think if we say `on delete restrict`, we'll get the same behavior as with this trigger

mattdailis (commented):
Here's a first round of comments - I need to spend some more time with `anchor_validation_status.sql` before I can review it intelligently

merlin-server/sql/merlin/tables/activity_directive.sql
mattdailis: Ah, un-quoted identifiers are case-insensitive, right? So this is more of a style comment than a naming one :+1:

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: Sure! The constraint says that if we have activity A with the following (relevant) fields: `id`, `plan_id`, `anchor_id`, that there must be another activity B in `activity_directives` such that `B.id = A.anchor_id` and `B.plan_id = A.plan_id` (plus or minus the foreign key null-value allowances).

Although looking it back over, I'm not certain if I want `on delete/update cascade`.

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: Either way, thanks for the catch. It is fixed now.

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: Looking into this more, it looks like we can make the order of operations within a transaction a bit more flexible with [`on delete no action`](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK) and [`deferrable initially deferred`](https://www.postgresql.org/docs/current/sql-set-constraints.html) and delay checking if the foreign key is satisfied until the COMMIT step of the transaction. Thoughts?

mattdailis: Thanks for sharing those docs - it's good to know that we have options. I like your thought about the order of delete statements - it is true that deferring the check could allow deletes to occur in multiple orders - but users shouldn't be doing bulk deletes in separate statements. The most idiomatic way to do a bulk delete is the following:

```sql
delete from activity_directive where plan_id = 1 and id in (1, 2, 3, 4, 5);
```

(or the graphql equivalent, using `_in: [1, 2, 3, 4, 5]`)

This deletes the whole set atomically - so there is no observable state that could violate our constraints.

So, let’s go for “on delete restrict” here until we encounter a compelling reason to defer the check.

Mythicaeda: Good to note, but I was thinking about the case where we want to delete an anchor and reanchor the thing anchored to it.

ie B is anchored to A and we want to delete A. With `on delete restrict`, the transaction must be
```sql
update activity_directive
set anchor_id = null
where (id, plan_id) = (B.id, B.plan_id);

delete from activity_directive
where (id, plan_id) = (A.id, A.plan_id);
```

But with the deferral, we could instead chose to reverse the order of those statements if we want.

camargo:
@duranb brought this up as he is implementing anchors. In the UI we currently compute an absolute start_time from the plan_start and start_offset for activity directives. However with anchors computing the absolute start time becomes harder since the absolute start_time depends on anchor offsets. Would it be possible to add a column to the directive that stores the absolute start time based on the anchor (if it exists)? If no anchor then based on the plan_start? It would be more ergonomic to not have to compute it in the client.

Moved convo to Slack: https://jpl.slack.com/archives/CJ07FGRPH/p1673390923129769

Twisol:
> However with anchors computing the absolute start time becomes harder since the absolute start_time depends on anchor offsets

Not just harder -- if something is anchored off the end time of another activity, we simply don't know what the absolute start time is without simulation.

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: Follow up on this: while we do get the same behavior using `on delete restrict` as we do with the trigger, I think the error message gotten from `on delete restrict` is more difficult to understand.

on delete restrict: `[23503] ERROR: update or delete on table "activity_directive" violates foreign key constraint "anchor_in_plan" on table "activity_directive" Detail: Key (id, plan_id)=(1, 1) is still referenced from table "activity_directive".`

Trigger: `[P0001] ERROR: Cannot delete. Other activity directives are anchored to Activity Directive 1.`

mattdailis: That's a good point :+1: I think that's a broader question than this ticket though - all of our constraint violation error messages pass the postgres error messages to the API client.

Looking around, hasura has an open ticket regarding abilities to customize error messages, so at least we know other people have had this thought as well https://github.com/hasura/graphql-engine/issues/3658

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
```diff
@@ -67,6 +67,27 @@ public void startDatabase() throws SQLException, IOException, InterruptedExcepti
       proc.destroy();
     }
 
+    // Grant usage privileges to all aerie for all schemas for tests
```
mattdailis: For my own understanding, what goes wrong if we don't grant these privileges, and why is does this occur in tests, but not in production?

Mythicaeda: If we don't grant `USAGE` privileges in tests, then the `aerie` user cannot see anything not inside the public schema, meaning that, for example, it does not have access to `hasura_functions` when testing against AerieMerlin (which is needed for the `AnchorDeletion` and `AnchorMerge` tests).

From what I can tell looking between `init-aerie.sh` and `DatabaseTestHelper`, the difference between test and production seems to be that the `postgres` user creates all the objects in the db in Test, while the `aerie` user creates them in production (meaning it would automatically have access to all objects).

deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
```diff
@@ -70,3 +70,21 @@
   configuration:
     custom_root_fields:
       function: set_resolution_bulk
+- function:
+    name: delete_activity_by_pk_reanchor_plan_start
```
mattdailis, line 74: minor: We can expose these without the hasura_functions prefix if we add a `custom_root_fields` here.

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -248,3 +260,81 @@ create trigger check_locked_on_delete_trigger
 before delete on activity_directive
 for each row
 execute procedure check_locked_on_delete();
+
```
mattdailis, line 263: extraneous newline here

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -248,3 +260,81 @@ create trigger check_locked_on_delete_trigger
...
+    with recursive history(activity_id, plan_id, anchor_id, total_offset) as (
+      select ad.id, ad.plan_id, ad.anchor_id, ad.start_offset
+      from activity_directive ad
+      where (ad.id, plan_id) = (_activity_id, _plan_id)
+      union
+      select ad.id, ad.plan_id, ad.anchor_id, h.total_offset + ad.start_offset
+      from activity_directive ad, history h
+      where (ad.id, ad.plan_id) = (h.anchor_id, h.plan_id)
+        and h.anchor_id is not null
+    )select total_offset
```
mattdailis:
```suggestion
    ) select total_offset
```

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -248,3 +260,81 @@ create trigger check_locked_on_delete_trigger
...
+  declare
+    _total_offset interval;
+  begin
+    if _plan_id is null then
+      raise exception 'Plan ID cannot be null.';
+    end if;
+    if _activity_id is null then
+      raise exception 'Activity ID cannot be null.';
+    end if;
+    if not exists(select id from public.activity_directive where (id, plan_id) = (_activity_id, _plan_id)) then
```
mattdailis: Just curious - is the `public.` required here?

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -248,3 +260,81 @@ create trigger check_locked_on_delete_trigger
...
+      where (ad.id, plan_id) = (_activity_id, _plan_id)
+      union
+      select ad.id, ad.plan_id, ad.anchor_id, h.total_offset + ad.start_offset
+      from activity_directive ad, history h
+      where (ad.id, ad.plan_id) = (h.anchor_id, h.plan_id)
+        and h.anchor_id is not null
+    )select total_offset
+    from history
+    where history.anchor_id is null
+    into _total_offset;
```
mattdailis: When trying to wrap my head around this code, I ended up re-styling it a bit:

```sql
with recursive path_to_root(activity_id, plan_id, anchor_id, start_offset) as (
  select id, plan_id, anchor_id, start_offset
    from activity_directive
    where (id, plan_id) = (_activity_id, _plan_id)
  union
  select ad.id, ad.plan_id, ad.anchor_id, ad.start_offset
    from activity_directive ad join path_to_root p
    on (ad.id, ad.plan_id) = (p.anchor_id, p.plan_id)
)
select sum(start_offset) from path_to_root into _total_offset;
```
Rather than accumulate the sum in the recursive query - I split it out into a separate statement. It's a few extra cpu cycles, but I think it helps readability.

merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -248,3 +260,81 @@ create trigger check_locked_on_delete_trigger
...
+          anchored_to_start = true
+      where (anchor_id, plan_id) = (_activity_id, _plan_id)
+      returning *;
+  end
+  $$;
+comment on function public.anchor_direct_descendents_to_plan(_activity_id integer, _plan_id integer) is e''
+'Given the primary key of an activity, reanchor all anchor chains attached to the activity to the plan.\n'
+'In the event of an end-time anchor, this function assumes all simulated activities have a duration of 0.';
+
+create function anchor_direct_descendents_to_ascendant_anchor(_activity_id int, _plan_id int)
```
mattdailis: Perhaps `anchor_direct_descendants_to_ancestor`? I'm not sure ascendant is the word we're looking for 😅

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+  from activity_directive
+  where (id, plan_id) = (_activity_id, _plan_id)
+  into _anchor_id, _start_offset, _anchored_to_start;
+
+
+  if (_start_offset < '0'                 -- if the offset is nonnegative, we don't have to check it
+    and _anchor_id is not null)           -- if the activity is anchored to the plan, then it doesn't matter the value of the offset
+  then
+    /*
+      Postgres if's don't "short-circuit" -- all clauses are evaluated. Therefore, this query is placed here so that
```
mattdailis: nitpick: It's not "if" that doesn't short-circuit, but "and".

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+      it only runs iff the outer 'if' is true
+    */
+    with recursive anchors(plan_id, activity_id, anchor_id, anchored_to_start, start_offset, total_offset) as (
+      select _plan_id, _activity_id, _anchor_id, _anchored_to_start, _start_offset, _start_offset
+      union
+      select ad.plan_id, ad.id, ad.anchor_id, ad.anchored_to_start, ad.start_offset, anchors.total_offset + ad.start_offset
+      from activity_directive ad, anchors
+      where anchors.anchor_id is not null                               -- stop at plan
+        and (ad.plan_id, ad.id) = (anchors.plan_id, anchors.anchor_id)
```
mattdailis: The convention in other parts of this PR is to put plan_id on the right: (id, plan_id). Suggest updating this line to be consistent with that convention

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+  _start_offset interval;
+  _anchored_to_start boolean;
+begin
+  select anchor_id, start_offset, anchored_to_start
+  from activity_directive
+  where (id, plan_id) = (_activity_id, _plan_id)
+  into _anchor_id, _start_offset, _anchored_to_start;
+
+
+  if (_start_offset < '0'                 -- if the offset is nonnegative, we don't have to check it
```
mattdailis: I'm not sure I follow why we don't have to check it - could you catch me up? I may suggest adding context to the comment as well

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+    _anchor_id integer;
+    _start_offset interval;
+    _anchored_to_start boolean;
+  begin
+    select anchor_id, start_offset, anchored_to_start
+    from activity_directive
+    where (id, plan_id) = (_activity_id, _plan_id)
+    into _anchor_id, _start_offset, _anchored_to_start;
+
+    if(_start_offset < '0'  and _anchored_to_start) then -- only need to check if anchored to start or something with a negative offset
```
mattdailis: Suggest whitespace tweak
```suggestion
    if (_start_offset < '0' and _anchored_to_start) then -- only need to check if anchored to start or something with a negative offset
```

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+
+/*
+  When an activity directive is anchored to another activity directive's end time, it must simplify to a positive offset,
+  as simulation can't handle a negative offset (since simulation won't know its start time until the anchoring activity has finished,
+  which will always place the anchored activity's start time in the past)
+
+  Throws an exception if:
+    - The activity is anchored to itself
+    - A cycle is detected
+  For other erroneous states that can wait until simulation time to be resolved, it writes to 'anchor_validation_status's 'reason_invalid' field and then returns.
```
mattdailis: wording suggestion - feel free to wordsmith further. I found the existing wording a bit unclear
```suggestion
  For other erroneous states that we allow to exist, as long as they're resolved before simulation time, it writes to 'anchor_validation_status's 'reason_invalid' field and then returns.
```

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+  returns trigger
+  security definer
+  language plpgsql as $$
+declare
+  end_anchor_id integer;
+  invalid_descendant_act_ids integer[];
+  offset_from_end_anchor interval;
+  offset_from_plan_start interval;
+begin
+  -- Clear the reason invalid field (if an exception is thrown, this will be rolled back)
```
mattdailis, line 151: Curious - why does it make sense to upsert a row with an empty string, as opposed to deleting the row, if it exists?

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+
+  -- An activity cannot anchor to itself
+  if(new.anchor_id = new.id) then
+    raise exception 'Cannot anchor activity % to itself.', new.anchor_id;
+  end if;
+
+  -- Validate that no cycles were added
+  if exists(
+      with recursive history(plan_id, activity_id, anchor_id, is_cycle, path) as (
+        select new.plan_id, new.id, new.anchor_id, false, array[row(new.plan_id, new.id)]
```
mattdailis: Suggest swap order - so that plan_id is on the right. Alternatively, I'm not sure we need the plan_id in this array.

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+comment on column anchor_validation_status.reason_invalid is e''
+  'If null, the anchor is valid. If not null, this contains a reason why the anchor is invalid.';
+
+
+/*
+    An activity directive may have a negative offset from its anchor's start time.
+    If its anchor is anchored to the end time of another activity (or so on up the chain), the activity with a
+    negative offset must come out to have a positive offset relative to that end time anchor.
+*/
+create function is_net_negative_end_offset(_activity_id integer, _plan_id integer)
```
mattdailis: Since is_net_negative_end_offset is used exclusively for its side effects (i.e. updating the anchor_validation_status table), I suggest making it a procedure, and giving it a name that makes it clear it has side effects. Perhaps `validate_net_end_offset` or similar. Same suggestion for `is_net_negative_plan_start`

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+  */
+
+  -- Get collection of dependent activities, with offset relative to this activity
+  create temp table dependent_activities as
+  with recursive d_activities(plan_id, activity_id, anchor_id, anchored_to_start, start_offset, total_offset) as (
+    select
+      ad.plan_id, ad.id, ad.anchor_id, ad.anchored_to_start, ad.start_offset, ad.start_offset
+    from activity_directive ad
+    where
+        (ad.plan_id, ad.anchor_id) = (new.plan_id, new.id) -- select all activities anchored to this one
```
mattdailis: Suggest moving plan_id to the right side. (Though, this PR does put plan_id on the left more than I realized. I believe it was plan merge that established the plan_id on the right convention. I don't mind either way, but I would appreciate consistency)

merlin-server/sql/merlin/tables/anchor_validation_status.sql
```diff
@@ -0,0 +1,341 @@
...
+    union
+    select ad.plan_id, ad.id, ad.anchor_id, ad.anchored_to_start, ad.start_offset, eta.total_offset + ad.start_offset
+    from activity_directive ad, end_time_anchor eta
+    where eta.anchor_id is not null                               -- stop at plan
+      and (ad.plan_id, ad.id) = (eta.plan_id, eta.anchor_id)
+      and eta.anchored_to_start                                   -- or stop at end time anchor
+  ) select into end_anchor_id, offset_from_end_anchor
+        anchor_id, total_offset from end_time_anchor eta -- get the id of the activity that the selected activity is anchored to
+  where not eta.anchored_to_start and eta.anchor_id is not null
+  order by activity_id desc
```
mattdailis: I'm suspicious of sorting by activity_id - what was the intent here?

merlin-server/sql/merlin/hasura_functions.sql
```diff
@@ -231,3 +231,91 @@ begin
...
+
+  return query
+    with recursive descendents(activity_id, p_id) as (
+    select _activity_id, _plan_id
+    from activity_directive ad
+    where (ad.id, ad.plan_id) = (_activity_id, _plan_id)
+    union
+        select ad.id, ad.plan_id
+        from activity_directive ad, descendents d
+        where (ad.anchor_id, ad.plan_id) = (d.activity_id, d.p_id)
```
mattdailis: Not sure what the best indentation is for this kind of query - here's one that seems okay 😅

```suggestion
    with recursive descendents(activity_id, p_id) as (
      select _activity_id, _plan_id
        from activity_directive ad
        where (ad.id, ad.plan_id) = (_activity_id, _plan_id)
    union
      select ad.id, ad.plan_id
        from activity_directive ad, descendents d
        where (ad.anchor_id, ad.plan_id) = (d.activity_id, d.p_id)
```
The `from` and `where` can be aligned with the select if you prefer - I'm mostly looking for indentation relative to the `with` and `union` parts of the recursive query

merlin-server/sql/merlin/hasura_functions.sql
```diff
@@ -231,3 +231,91 @@ begin
...
+    raise exception 'Activity Directive % does not exist in Plan %', _activity_id, _plan_id;
+  end if;
+
+    return query
+      with updated as (
+        select public.anchor_direct_descendents_to_ascendant_anchor(_activity_id := _activity_id, _plan_id := _plan_id)
+      )
+      select updated.*, 'updated'
+        from updated;
+    return query
```
mattdailis, line 281: I did not know that pl/pgsql allows multiple return statements in a function. Nifty!

merlin-server/sql/merlin/hasura_functions.sql
```diff
@@ -231,3 +231,91 @@ begin
...
+      with deleted as (
+        delete from activity_directive where (id, plan_id) = (_activity_id, _plan_id) returning *
+      )
+      select (deleted.id, deleted.plan_id, deleted.name, deleted.tags, deleted.source_scheduling_goal_id,
+              deleted.created_at, deleted.last_modified_at, deleted.start_offset, deleted.type, deleted.arguments,
+              deleted.last_modified_arguments_at, deleted.metadata, deleted.anchor_id, deleted.anchored_to_start)::activity_directive, 'deleted' from deleted;
+end
+$$;
+
+create function hasura_functions.delete_activity_by_pk_delete_chain(_activity_id int, _plan_id int)
```
mattdailis: Suggest using the word `subtree` instead of `chain` - `chain` is a bit misleading because implies a linear structure

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -2663,4 +2562,155 @@ void cancelCleansUpSuccessfully() throws SQLException {
...
+                  set anchor_id = %d,
+                      anchored_to_start = %b
+                  where id = %d and plan_id = %d;
+                  """.formatted(anchorId, anchoredToStart, activityId, planId));
+        }
+      }
+    }
+
+    @Test
+    void cantMergeCycle() throws SQLException{
```
mattdailis, line 2593: Whoa, good edge case. I hadn't considered that "valid" anchors in the source plan can become invalid in the target plan... especially since the merge process allows being selective about which directives are carried over...

mattdailis: The question arises - how is this situation made visible to the user, and how does a user remedy the situation? I think what you have here - simply don't allow the commit - is definitely reasonable for this stage. It may be worth thinking through what the "optimal" user experience would be 🤔

mattdailis (commented):
Great work 🚀 Most of my comments are clarifying questions or stylistic concerns.

This review is purely based on reading the code - I'd like to spend some time tomorrow actually exercising it a bit

deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
Mythicaeda: Can you clarify or give an example of how the layout is supposed to be? I copied the layout of the parameters from the contents already there from plan merge, including having a `custom_root_fields` in the `configuration` section, so I'm not sure why they're appearing in the GQL with the `hasura_functions` prefix.

merlin-server/sql/merlin/hasura_functions.sql
Mythicaeda: Yeah, because we're returning a `setof`, `return query` and `return next` contribute to the results set without actually terminating the function. [Here's the docs on it](https://www.postgresql.org/docs/14/plpgsql-control-structures.html#PLPGSQL-STATEMENTS-RETURNING) if you're curious.

deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
mattdailis: My mistake - my eyes glazed over and I didn't see the `custom_root_fields` right there 🤦

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: At the end of the transaction, I want to keep the row around even if there's no error, as the `reason_invalid` column being empty means that the activity has been checked at some point and had no anchor validation errors at that time, while the row not existing means that the activity's anchor has never been checked. Therefore, if this was a delete, then we would need to keep track of if we need to insert `(new.id, new.plan_id, '')`, which was more complicated than upserting the empty row and then overwriting that update if there's a reason to.

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: We need it for the `where` clause to make sure we're grabbing the right activity, but you're right, we don't need to store it in `history` at all.

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: Yeah, I think the convention I was using was that if I was comparing to something's anchor, I would do (plan, id) vs (plan, anchor_id), but I'll change it to be consistent with what was done in Plan Merge. 👍

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: I made an implicit assumption that should that select grab more than one element (which it shouldn't), that the activity with the higher ID would be further down the chain. I'm removing the `order by` statement now.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: Yeah, although I don't think this is something that can really be checked in the DB prior to the merge getting committed, since it doesn't have all the information regarding conflict resolution until a user goes to commit.

UX-wise, I think that as long as these errors are actually shown to the user (plus or minus some language tweaking to make it more obvious which activity is part of the cycle), that ought to be sufficient, as they don't lose any work from failing to commit.

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: Probably not anymore. I was originally working on these functions in the `hasura_functions` schema, and must've forgotten to remove the schema specification when moving the code into `activity_directive`

merlin-server/sql/merlin/hasura_functions.sql
Mythicaeda: Thanks. I really need to disable auto-indent for SQL files 🤦‍♀️

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: How about:

```
For all other invalid states, it writes to 'anchor_validation_status's 'reason_invalid' field and then returns.
If the activity's anchor is valid, then the 'reason_invalid' field on the activity's entry in 'anchor_validation_status' is set to ''.
```

mattdailis: Sounds great to me :+1:

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: You're right, we _do_ have to check it. I'll add a test and fix the function 👍.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java
mattdailis: I see, thanks for explaining. I think rather than grant aerie additional permissions in tests, we should use the aerie user to create those tables - especially since that helps mirror production a bit better.

I think we can do that by splitting this call in two - so that we leave the GRANT PRIVILEGES call as is, but run the `init-aerie.sh` under the `aerie` user (by replacing `postgres:postgres` in the url with `aerie:aerie`):

https://github.com/NASA-AMMOS/aerie/blob/37e6b229a852a618c7d243abb5acae4079526871/db-tests/src/test/java/gov/nasa/jpl/aerie/database/DatabaseTestHelper.java#L57-L61

Mythicaeda: Makes sense. Change has been done 👍

merlin-server/sql/merlin/tables/activity_directive.sql
Mythicaeda: I'd prefer to keep the accumulation in the recursive query, as that is its whole purpose. I can rename `total_offset` to `net_offset` and `history` to `get_net_offset` (or keep `total_offset` the same and rename `history` to `get_total_offset`) if that would help with understanding.

merlin-server/sql/merlin/tables/anchor_validation_status.sql
Mythicaeda: Changes have been made.

mattdailis (approved):

camargo (approved):
⚓

PR #577, opened 2023-01-06
Provide a single consistent capability for JSON (de)serialization of SerializedValue

Twisol:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
This PR takes the various copies and tweaked forms of the `SerializedValueJsonParser` and consolidates it into `merlin.driver.json`, where the closely-related `ValueSchemaJsonParser` already lives. In the process, we unify the internal representation of real and integer numbers in `SerializedValue` under the `BigDecimal` type, making it much easier to consistently serialize *and* deserialize numbers, be they represented as `4` or `4.0`.

This is labeled as a breaking change because it affects `merlin.protocol`. Custom `ValueMapper`s that use `SerializedValue.Visitor` directly may need to be updated to use `onNumeric` instead, but most mappers should be using the `asX` helper methods, of which none have been removed. Client models should "only" need a rebuild for compatibility with an updated host.

## Verification
At present, by successful build only. As far as I can tell, all but one of these instances of (de)serialization were identical, existing only because there wasn't already a parser defined in a common upstream package. The one distinct parser, the `StrictSerializedValueJsonParser`, seems (by consultation with @mattdailis) to have been a workaround for issues where an integer input argument may be transformed into a real effective argument; this should no longer be possible with the unified representation under `BigDecimal`, but we should probably double-check.

## Documentation
I don't think any documentation has been invalidated by this change. The `SerializedValue.Visitor` interface has changed, but that's a fairly low-level detail; most usage should be using the `asX()` helpers, of which none have been removed and only `asNumeric` (for the underlying `BigDecimal`) has been added.


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -101,8 +103,7 @@ public static JsonValue serializeStringList(final List<String> elements) {
   }
 
   public static JsonValue serializeArgument(final SerializedValue parameter) {
-    if (parameter == null) return JsonValue.NULL;
```
JoelCourtney, line 104: Looks like we lost this null check. Not sure why it was there in the first place, but we should probably keep it.

JoelCourtney (approved):
That's pretty slick, only one nitpick.

mattdailis:
I rebased this onto develop and added the null check back, per @JoelCourtney 's request

Here's the old version, for reference: https://github.com/NASA-AMMOS/aerie/tree/feature/consolidate-json--backup

mattdailis (approved):
I think this is ready to go :shipit:

Twisol:
Thanks for picking up this PR for me, Matt :+1: I didn't even notice Joel had commented; the notification must be somewhere in my email 🤔

camargo:
What would the upgrade guide look like for this?

mattdailis:
> What would the upgrade guide look like for this?

I would put:

> Any implementations of the `SerializedValue.Visitor` interface will need to replace their `onReal(double)` and `onInt(long)` methods with the new `onNumeric(BigDecimal)` method.

That said, I can't seem to find a single implementation of it anywhere outside of our own code, so I suspect for most customers it won't be a breaking change

I even think that it should continue to work with mission models compiled against a previous version of the interface, but I haven't tried that...

PR #582, opened 2023-01-09
Add schema_migrations table

mattdailis:
* **Tickets addressed:** This is related to #521 
* **Review:** By commit  

## Description

This PR adds the `schema_migrations` table to each of our four databases. It does it both by creating a migration, and by editing the latest schema.

## Verification

I brought up a fresh database using the latest schema, and checked that there were no SQL errors. I then visually inspected the database to see that the table was present.

I then brought down the system, removed volumes, checked out develop, and brought it up again. I then checked out this branch, and used our aerie_db_migrations script to apply the new migrations. I checked that no SQL errors occurred, and did another visual inspection of the database.

## Documentation

For now, the existing migrations documentation continues to apply. When we get around to modifying the `aerie_db_migrations` script we may want to include documentation on how it leverages the contents of the `schema_migrations` tables.

## Future work

- Automated migration testing

### DIFF
--- a/deployment/hasura/migrations/AerieMerlin/0_create_schema_migrations/up.sql
+++ b/deployment/hasura/migrations/AerieMerlin/0_create_schema_migrations/up.sql
@@ -1,0 +1,33 @@
+create schema migrations;
+create table migrations.schema_migrations (
+  migration_id varchar primary key
+);
+
+create procedure migrations.mark_migration_applied(_migration_id varchar)
+  language plpgsql as $$
+begin
+insert into migrations.schema_migrations (migration_id)
+values (_migration_id);
+end;
+$$;
+
+create procedure migrations.mark_migration_rolled_back(_migration_id varchar)
+  language plpgsql as $$
+begin
+delete from migrations.schema_migrations
+where migration_id = _migration_id;
+end;
+$$;
+
+comment on schema migrations is e''
+  'Tables and procedures associated with tracking schema migrations';
+comment on table migrations.schema_migrations is e''
+  'Tracks what migrations have been applied';
+comment on column migrations.schema_migrations.migration_id is e''
+'An identifier for a migration that has been applied';
+comment on procedure migrations.mark_migration_applied is e''
+  'Given an identifier for a migration, add that migration to the applied set';
+comment on procedure migrations.mark_migration_rolled_back is e''
+  'Given an identifier for a migration, remove that migration from the applied set';
+
+call migrations.mark_migration_applied('0');
--- a/deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
+++ b/deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
@@ -1,0 +1,33 @@
+create schema migrations;
+create table migrations.schema_migrations (
+  migration_id varchar primary key
+);
+
+create procedure migrations.mark_migration_applied(_migration_id varchar)
+  language plpgsql as $$
+begin
+insert into migrations.schema_migrations (migration_id)
+values (_migration_id);
+end;
+$$;
+
+create procedure migrations.mark_migration_rolled_back(_migration_id varchar)
+  language plpgsql as $$
+begin
+delete from migrations.schema_migrations
+where migration_id = _migration_id;
+end;
+$$;
+
+comment on schema migrations is e''
+  'Tables and procedures associated with tracking schema migrations';
+comment on table migrations.schema_migrations is e''
+  'Tracks what migrations have been applied';
+comment on column migrations.schema_migrations.migration_id is e''
+'An identifier for a migration that has been applied';
+comment on procedure migrations.mark_migration_applied is e''
+  'Given an identifier for a migration, add that migration to the applied set';
+comment on procedure migrations.mark_migration_rolled_back is e''
+  'Given an identifier for a migration, remove that migration from the applied set';
+
+call migrations.mark_migration_applied('0');
--- a/deployment/postgres-init-db/sql/ui/init.sql
+++ b/deployment/postgres-init-db/sql/ui/init.sql
@@ -1,4 +1,8 @@
 begin;
+  -- Schema migrations
+  \ir tables/schema_migrations.sql
+  \ir applied_migrations.sql
+
   -- Tables.
   \ir tables/view.sql
 end;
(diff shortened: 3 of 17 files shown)


deployment/postgres-init-db/sql/ui/init.sql
```diff
@@ -1,4 +1,8 @@
 begin;
```
Mythicaeda, line 1: Where do the UI files come from? I know for the other databases, they're filled in by the `DistributeSQL` Gradle task, but something feels weird with the UI DB only being defined in `deployment`.

deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
```diff
@@ -0,0 +1,33 @@
...
+$$;
+
+comment on schema migrations is e''
+  'Tables and procedures associated with tracking schema migrations';
+comment on table migrations.schema_migrations is e''
+  'Tracks what migrations have been applied';
+comment on column migrations.schema_migrations.migration_id is e''
+'An identifier for a migration that has been applied';
+comment on procedure migrations.mark_migration_applied is e''
+  'Given an identifier for a migration, track that migration as applied';
```
Mythicaeda: [Ignorable]
```suggestion
  'Given an identifier for a migration, add that migration to the applied set';
```

deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
```diff
@@ -0,0 +1,33 @@
...
+comment on procedure migrations.mark_migration_rolled_back is e''
+  'Given an identifier for a migration, make sure that migration is not considered applied';
```
Mythicaeda: [Ignorable]
```suggestion
  'Given an identifier for a migration, remove that migration from the applied set.';
```

deployment/hasura/migrations/AerieMerlin/0_create_schema_migrations/up.sql
```diff
@@ -0,0 +1,33 @@
+create schema migrations;
+create table migrations.schema_migrations (
```
Mythicaeda, line 2: Something I noticed when I applied the migrations is that Hasura doesn't automatically track the new tables. In the event we want it to (for example, in order to create a new relationship), do we know if there's a way to do this or will the user just have to copy in the latest schema's metadata and refresh it after they finish migrating?

Mythicaeda (approved):

deployment/hasura/migrations/AerieMerlin/0_create_schema_migrations/up.sql
Twisol: If we want Hasura to track a table, I think we can just add it to `deployment/hasura/metadata/databases/*/tables/`. The tables exposed to Hasura should only be those meant to be exposed to downstream clients and exported through the Aerie API. I'm not sure that's true of `schema_migrations`.

Twisol: (I think that once the `metadata/` is altered, customers need to reload the metadata through either the terminal or the Hasura frontend; I don't think it will autoload any changes. But it would be a one-step change as part of an upgrade in that case, and not something they need to go in and directly configure themselves.)

mattdailis: > will the user just have to copy in the latest schema's metadata and refresh it after they finish migrating?

Yeah - I think this is what will end up happening :+1: Thanks for bringing that up - there are no corresponding "metadata" migrations - so the metadata will need to be updated wholesale

Twisol: (Oops, sorry -- I was distracted and didn't realize I didn't completely read @Mythicaeda's full message. Sorry for missing the point 🤦‍♂️)

Mythicaeda: Huh. I know that when applying a migration via the Hasura CLI, it reloads the metadata, which to me implied that it might've been changeable alongside migrations, but reading the docs, I think it is just something that a user will need to replace and reload.

Also, looking at [the docs](https://hasura.io/docs/latest/migrations-metadata-seeds/migrations-metadata-setup/), there's apparently a way to mark a migration as applied without executing it:
```
# mark the squashed migration as applied on this server
hasura migrate apply --version "<squash-migration-version-timestamp-number>" --skip-execution --database-name <database-name>
```

Which should be useful for at least the aerie-helper migration script.

(And, yeah, I agree, `schema_migrations` probably doesn't need to be exposed and tracked, but in general if we add a table, we'll _probably_ want to track/expose it.)

Mythicaeda: Agh, I submitted that right before I got the email for your replies. Thanks for the clarification @mattdailis!

deployment/postgres-init-db/sql/ui/init.sql
mattdailis: The UI files are, in spirit, owned by the `aerie-ui` repo - so there is no corresponding gradle project for them in this repo.

The question of which things belong in `deployment`, vs in other subprojects is a good one - we used to keep everything in deployment, and then we decided to try to communicate ownership of the sql files by moving them into their corresponding project files. Our intentions were good, but `distributeSql` has been the source of much confusion, so perhaps it would be best to go back to keeping everything in deployment 🤷 

For the purposes of this PR, we're leaving `distributeSql` alone, and putting migrations in the locations expected by hasura. Fortunately, this is something that we do have the flexibility to change after the fact :+1:

deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
mattdailis: Thanks for the suggestion - I've made the change :+1:

deployment/hasura/migrations/AerieScheduler/0_create_schema_migrations/up.sql
mattdailis: Done :+1:

deployment/postgres-init-db/sql/ui/init.sql
Mythicaeda: Oh yeah, that is a weird situation, although I think it's mostly weird because Aerie's a mono repo _except_ for the Gateway/UI. Since it doesn't sound like anything in the UI repo is affecting/sharing ownership over these files, though, I'm no longer concerned.

mattdailis:
Latest force push drops the quotes from the `"migration_id"` column name

camargo (approved):
LGTM

PR #584, opened 2023-01-10
Add QEMU and buildx action for multi-arch container images

### LINKED ISSUE #555: Support multi-arch Docker images for all Aerie images

skovati:
* **Tickets addressed:** resolves #555 
* **Review:** By commit  

## Description

Adds `buildx` support to our `publish` pipeline to allow multi-arch (amd64, arm64, etc) images to be built. Uses QEMU to emulate architectures, since the built-in QEMU emulation in buildx fails otherwise. See #576 .

## Verification

Changes were applied to a [fork](https://github.com/skovati/testing/actions/runs/3877220045/jobs/6611978486) to make sure the action worked as intended.

## Documentation

N/A

## Future work

N/A

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,6 +45,12 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64
+      - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


Mythicaeda (approved):
I pulled the images on your fork. It's exciting to not get the Rosetta warning on a pulled Aerie image!

camargo (approved):
🔥

PR #589, opened 2023-01-11
Remove testing from publish workflow

camargo:
* **Review:** By commit  

## Description

* Already handled in test workflow
* Remove unnecessary ./gradlew classes

## Verification

Must merge to verify

## Future work

* Will add better test reporting in a follow up commit


skovati (approved):
Good call, no need to run these same tests again.

PR #591, opened 2023-01-12
Creation of windows/spans/profiles at absolute times

### LINKED ISSUE #590: Implement creation of windows/spans/profiles at absolute times
The main choice here is between `Temporal.Instant` and `Temporal.ZonedDateTime`. Since plan start time doesn't have time zone information, I don't think these times should either. So I'll be using `Temporal.Instant`.

JoelCourtney:
* **Tickets addressed:** closes #590 
* **Review:** By commit  

## Description
This PR allows creating individual windows, spans, and profile segments at absolute times in the eDSL using `Temporal.Instant`. For windows and profiles, I just expanded the implementation of `Windows.Value`, `Real.Value`, and `Discrete.Value` to allow an optional `Interval` class argument. For Spans I added a new function `Spans.FromInterval`.

The `Interval` class is very similar to the intervals we use in the constraints back-end, except it is defined with absolute times instead of offsets from plan start. I added a corresponding `AbsoluteInterval` record in the constraints package to have something to deserialize into.

The JS `Temporal.Instant` and the Java `java.time.Instant` classes are meant to play nice with each other. They represent the same concept (a point in time with no time-zone information) and (de)serialize to the same format (ISO date time) by default. So since we represent the plan start time with `java.time.Instant`, I am not expecting any timey-wimey wibbley-wobbley stuff to happen.

This also removes the json ast schema generation from both typescript packages, because it is unused and they are not smart enough to handle `Temporal.Instant`.

## Verification
I didn't implement every permutation of every test relevant to this feature. The tests do hit:
- Absolute Intervals created, serialized, deserialized, and converted to relative intervals
- Optionally providing intervals in a `.Value` function
- Not providing all fields for an interval
- The usual ast/parser/execution tests from `Spans.FromInterval`.

So I think we're good.

## Documentation
I will open another PR in the docs repo.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -129,9 +129,11 @@ public final class ConstraintParsers {
       productP
           .field("kind", literalP("RealProfileValue"))
           .field("value", doubleP)
+          .field("rate", doubleP)
+          .field("interval", absoluteIntervalP)
           .map(
-              untuple((kind, value) -> new RealValue(value)),
-              $ -> tuple(Unit.UNIT, $.value));
+              untuple((kind, value, rate, interval) -> new RealValue(value, rate, interval)),
+              $ -> tuple(Unit.UNIT, $.value(), $.rate(), $.interval()));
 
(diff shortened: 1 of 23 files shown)


adrienmaillard (approved):
LGTM. I liked especially how you made the definition of absolute interval as optional as possible everywhere.

PR #600, opened 2023-01-19
Update the all services to use Node LTS 18.0

### LINKED ISSUE #547: Upgrade all services to use Node LTS
They are currently on [16.x](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/build.gradle#L6) and LTS is on [18.13.0](https://nodejs.org/en/).

goetzrrGit:
* **Tickets addressed:** Resolves #547
* **Review:** By commit  

## Description
Updated the sequencing server to Node 18 which is the next LTS version

## Verification
I ran all the automatic test and they all passed

## Documentation
None

## Future work
None


skovati (approved):
LGTM! Any reason we're jumping to npm `v9.3.1` when `v8.19.3` is bundled with node LTS `v18.3.0`?

dyst5422:
@skovati Great point - I don't know what the difference between 8 and 9 are, but it seems that unless we have a specific use case for the difference there, we should just let our version of npm be dictated by the node version

goetzrrGit:
I can revert that change, The only reason for the bump was that my console displayed colorful text that I should update.

dyst5422:
@goetzrrGit Do we even need to specify the npm version? Can we omit that and allow the node version to drive it without specifying at all?

goetzrrGit:
@dyst5422 let me give it a try and see.

goetzrrGit:
It wasn't needed at all thanks for the suggestion guys.

dyst5422:
Should we additionally pin the docker images to use this same node version? I'm thinking yes. Currently, we have node:lts-alpine, which will drift from 18 as time goes on

goetzrrGit:
Good point, I wasn't even thinking about that. Docker is now using `node:18-alpine`

dyst5422:
This looks great to me.

I wonder how likely/risky the potential separation between 18-alpine in Docker and a hard versioned 18.13.0 in gradle is. I'm inclined to say not worth addressing, but thought I'd mention it.

skovati:
Awesome, looks even better! Seems like `merlin-server` and `scheduler-worker` are also on node 16 and could use a similar upgrade?

goetzrrGit:
@camargo are there tickets to update the `merlin-server `and` scheduler-worker` as well?

camargo:
Spoke with @goetzrrGit. We are just expanding the scope of this ticket a bit to include the other Aerie services.

cohansen:
We probably want an update to `.nvmrc` as part of this PR as well right?

camargo (approved):
LGTM

cohansen (approved):

PR #602, opened 2023-01-19
Accumulated duration for windows/spans

### LINKED ISSUE #592: Accumulated duration profile for windows/spans

JoelCourtney:
* **Tickets addressed:** closes #592 
* **Review:** By commit  

## Description
Implements an operation to count the accumulated time in windows and spans objects, as a real profile. The usage is `windows.accumulatedDuration(Temporal.Duration)`, where the argument tells the unit of time to count.

For example, if a Spans object has a single span `[0, 1]`, then the following:

```typescript
let milli = Temporal.Duration.from({milliseconds: 1});
spans.accumulatedDuration(milli)
```

will produce a real profile that is:
```
(-inf, 0): 0
[0, 1]:    1000 * t
(1, inf):  1000
```

where `t` is the time in seconds.

## Verification
Tests are added

## Documentation
The eDSL functions are docced.


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
```diff
@@ -27,7 +27,8 @@ public LinearEquation(final Duration initialTime, final double initialValue, fin
   }
 
   public double valueAt(final Duration time) {
-    return this.initialValue + this.rate*(time.minus(this.initialTime)).ratioOver(Duration.SECOND);
+    final var change = this.rate*time.ratioOver(Duration.SECOND) - this.rate*this.initialTime.ratioOver(Duration.SECOND);
+    return this.initialValue + change;
```
adrienmaillard, line 31: Can you explain why you made that change ?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -323,6 +323,26 @@ export class Windows {
...
+
+  /**
+   * Counts the cumulative amount of time spent in an active Span, as a multiple of some unit of time.
+   *
+   * WARNING: This method does not guarantee any particular zero point. Meaning, don't base behavior off of
+   * the total absolute time over the whole plan, because features such as temporal subset branches or
+   * simulation might change the zero point.
+   *
+   * Instead, use this for *relative* changes. Unfortunately, this requires subtraction, which is
+   * surprisingly difficult right now. We're working on it.
```
adrienmaillard: Can you detail this here in the discussion ? I am not grasping what you mean to explain. Maybe give an example ?

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -956,6 +998,20 @@ declare global {
      * @param defaultProfile boolean or windows to take default values from
      */
     public assignGaps(defaultProfile: Windows | boolean): Windows;
+
+    /**
+     * Counts the cumulative amount of time spent in an active Span, as a multiple of some unit of time.
+     *
```
adrienmaillard: Is this in the Windows class ? if yes, update comment from Span to Windows ?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
JoelCourtney: There is a possible long overflow in `time.minus(this.InitialTime)` that came up in testing. If `this.rate >= 1` then the overflow is unavoidable, but in this case rate was 0, so by distributing it we can avoid some overflows.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: Mathematically this should be treated like an [affine space](https://en.wikipedia.org/wiki/Affine_space). A good example from physics is the energy contained in a system. In most situations except those dealing with absolute zero, there isn't a universal choice of what should be labelled as "zero energy", because different observers can disagree on what would be the most natural zero point for their calculations. But all observers will ultimately agree on how the system behaves because it is only differences/changes in energy that matter to the evolution. In electronics, the decision that ground is zero volts is an arbitrary convention. It would be just as valid to declare that ground is 5 volts and all the predictions would be the same, although the math would be a little more annoying.

In our case, the duration profile will always start at 0 at the beginning of the simulation data, so it seems like you could write a constraint that says "don't do more than 1 hour of activity A in this plan". But if we had plan subset branches, the start time of simulation might change, some activities might be omitted, and `accumulatedDuration` will assign a different point the honor of being `0`. So its a lot more error-prone to reason about total time, and a lot more reliable to reason about changes/differences in time.

There's also a typo in that comment, thanks for making be re-read it :)

Twisol: Can we also just plaster your amazing explanation all over the simulation engine, please? Thanks 😂

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: done

adrienmaillard (approved):
LGTM.

PR #604, opened 2023-01-20
Catch long overflows when looking for intersection

JoelCourtney:
* **Tickets addressed:**
* **Review:** By commit  

## Description
Due to floating point math, it is possible for subtracting a number from itself to not result in 0. (i.e. `0.1+0.2-0.1-0.2`)

I personally don't like declaring an arbitrary cutoff like `|x| < 1e-15` means `x = 0`, because well, how do you decide?

So when finding intersections between two lines where the difference in slopes is `1e-20`, our options are:
1. use hand-written non-standard saturating math routines which are slow just to avoid using exceptions in the happy path
2. use exceptions in the happy path

2 is a lot easier.

## Verification
I updated the lessThan test to cause this exception.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
@@ -44,10 +44,26 @@ public final class LinearEquation {
 
   private Optional<Duration> intersectionPointWith(final LinearEquation other) {
     if (this.rate == other.rate) return Optional.empty();
+
+    /*
+    Floating point noise can cause rates to be extremely near zero, when they should
+    have been exactly zero. For example: `0.1 + 0.2 - 0.1 - 0.2 != 0`.
+
+    This can cause the denominator below to be tiny, leading to long overflow later.
+     */
+
+    // If the following causes an exception, something really has gone wrong, and we don't want to catch it.
+    final double numSeconds = (other.valueAt(this.initialTime) - this.initialValue) / (this.rate - other.rate);
+
+    // Check if numSeconds is too big before putting it in a long.
+    if (Math.abs(numSeconds) > ((double) Long.MAX_VALUE) / Duration.SECOND.dividedBy(Duration.MICROSECOND)) {
+      return Optional.empty();
+    }
+
     return Optional.of(
         this.initialTime.plus(
             Duration.roundNearest(
-                (other.valueAt(this.initialTime) - this.initialValue) / (this.rate - other.rate),
+                numSeconds,
                 Duration.SECONDS
             )
         )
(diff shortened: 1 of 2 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
```diff
@@ -44,14 +44,18 @@ public boolean changing() {
...
+    try {
+      return Optional.of(
+          this.initialTime.plus(
+              Duration.roundNearest(
+                  (other.valueAt(this.initialTime) - this.initialValue) / (this.rate - other.rate),
+                  Duration.SECONDS
+              )
+          )
+      );
+    } catch (ArithmeticException e) {
```
mattdailis: I think this is an adequate guard :+1: My only concern is whether or not it's over-broad - but after inspecting `valueAt`, I've convinced myself that the only source of `ArithmeticException` should be the attempt to fit a large (or infinite) `double` into a `long`.

mattdailis (approved):

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/LinearEquation.java
JoelCourtney: Actually that's not true, I just yesterday saw a different long overflow from `.valueAt(Duration.SECOND)` because it was called on a line with `initialTime = Duration.MIN_VALUE`. This is fixed in the accumulated duration operation PR.

JoelCourtney: Point being that computer arithmetic always finds a way.

JoelCourtney: How about this: if we call `i = a.intersectionPointWith(b)` where `abs(i - a.initTime) > LONG_MAX` then we will get an overflow. This can happen if `a.initTime = Duration.MIN` and `i` is some normal positive value.

Its a bit pathological, and AFAIK typical simulation results will never produce a line segment that starts at `Duration.MIN`, but technically there's still a way.

JoelCourtney: I've updated the check. It is no longer a try/catch. Instead it just checks this one case, just before the longs get involved.

There are still edge cases that can cause exceptions, like the above example, and I think we should just let those exceptions happen. If we somehow end up with an extremely long interval or an extremely steep slope in a real simulation, something really has gone wrong and returning `Optional.empty()`, meaning there is no intersection` is probably incorrect.

JoelCourtney: I tend to build a linear queue of expensive features I want to implement, which I know is not ideal, but here it is:

- In order to do a proper job of eliminating `Duration.MIN` and `Duration.MAX` from constraints/scheduling, my preference would be to use the lazy streaming profiles refactor idea, because that had very efficient and clean bounded evaluation.
- That refactor had major (major) problems with the parser system, that AFAICT made the objects behind interface that was simplest to use impossible to deserialize.
- Now that procedural scheduling/constraints is on the table, there is a long-term possibility of obviating the need for the scheduler/constraints parsers, enabling the lazy streaming interface again.

Therefore, in order to eliminate a couple idiosyncratic edge cases that could cause overflow exceptions, we should implement procedural constraints. I see no flaws in this logic.

mattdailis (approved):

camargo (approved):

PR #606, opened 2023-01-20
Add Kubernetes manifests

### LINKED ISSUE #572: Prototype Kubernetes manifests for aerie development/deployment
[Kubernetes](https://kubernetes.io/) is a popular open-source container orchestration engine developed by Google. Since `aerie` is deployed as a collection of containers, we could leverage features provided by k8s and its ecosystem such as container scheduling & scaling, load balancing, rollout/rollback of deployments, [mTLS](https://linkerd.io/2.12/features/automatic-mtls/#) between containers, networking metrics, [GitOps](https://fluxcd.io/), etc. This ticket should introduce a first iteration of kubernetes manifests that can be shared between local development using a tool like [Docker Desktop](https://docs.docker.com/desktop/kubernetes/) or [minikube](https://minikube.sigs.k8s.io/docs/start/), as well as robust cloud deployments with something like [EKS](https://aws.amazon.com/eks/).

skovati:
* **Tickets addressed:** close #572 
* **Review:** By commit  

## Description

This initial prototype of kubernetes manifests is essentially a one-for-one port of our `deployment/docker-compose.yml`, so it uses all pre-built images from `ghcr.io`.

## Verification

Kubernetes successfully deployed using a local [k3s](https://k3s.io) cluster, running in `docker` containers with the help of [k3d](https://k3d.io). 

## Documentation

No documentation invalidated, but new Kubernetes specific documentation regarding local deployment and AWS deployment will be added to [aerie-docs](https://github.com/NASA-AMMOS/aerie-docs).

## Future work

- A small but important quirk of Kubernetes is that it doesn't support underscores in pod names, since DNS records need to be compliant with [RFC1123](https://tools.ietf.org/html/rfc1123). This means we will need to change the naming scheme of our docker-compose files and hasura metadata, which will be completed in a follow-up PR.
- Secrets management should be handled using a templating scheme like the built-in [kustomize](https://kustomize.io/).
- Proper ingress support should be added, e.g. having http://gateway.localhost redirect to the gateway service or using `LoadBalancer` objects instead of relying on `NodePorts`
- Use environment file to hold all env vars instead of hardcoding into manifests.

### DIFF
--- a/deployment/kubernetes/aerie-file-store-persistentvolumeclaim.yaml
+++ b/deployment/kubernetes/aerie-file-store-persistentvolumeclaim.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  namespace: aerie-dev
+  name: aerie-file-store
+spec:
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 100Mi
--- a/deployment/kubernetes/aerie-gateway-deployment.yaml
+++ b/deployment/kubernetes/aerie-gateway-deployment.yaml
@@ -1,0 +1,54 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: aerie-dev
+  name: aerie-gateway
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: aerie-gateway
+  template:
+    metadata:
+      labels:
+        app: aerie-gateway
+    spec:
+      containers:
+        - name: aerie-gateway
+          image: ghcr.io/nasa-ammos/aerie-gateway:develop
+          ports:
+            - containerPort: 9000
+          volumeMounts:
+            - mountPath: /app/files
+              name: aerie-file-store
+          env:
+            - name: AUTH_TYPE
+              value: none
+            - name: GQL_API_URL
+              value: http://localhost:8080/v1/graphql
+            - name: LOG_FILE
+              value: console
+            - name: LOG_LEVEL
+              value: warn
+            - name: PORT
+              value: "9000"
+            - name: POSTGRES_AERIE_MERLIN_DB
+              value: aerie_merlin
+            - name: POSTGRES_HOST
+              value: postgres
+            - name: POSTGRES_PORT
+              value: "5432"
+            - name: POSTGRES_USER
+              valueFrom:
+                secretKeyRef:
+                  name: dev-env
+                  key: AERIE_USERNAME
+            - name: POSTGRES_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: dev-env
+                  key: AERIE_PASSWORD
+      volumes:
+        - name: aerie-file-store
+          persistentVolumeClaim:
+            claimName: aerie-file-store
--- a/deployment/kubernetes/aerie-ui-deployment.yaml
+++ b/deployment/kubernetes/aerie-ui-deployment.yaml
@@ -1,0 +1,35 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: aerie-dev
+  name: aerie-ui
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: aerie-ui
+  template:
+    metadata:
+      labels:
+        app: aerie-ui
+    spec:
+      containers:
+        - env:
+            - name: ORIGIN
+              value: http://127.0.0.1
+            - name: PUBLIC_AUTH_TYPE
+              value: none
+            - name: PUBLIC_GATEWAY_CLIENT_URL
+              value: http://localhost:9000
+            - name: PUBLIC_GATEWAY_SERVER_URL
+              value: http://aerie-gateway:9000
+            - name: PUBLIC_HASURA_CLIENT_URL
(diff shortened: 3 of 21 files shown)


deployment/kubernetes/aerie-file-store-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  namespace: aerie-dev
+  name: aerie-file-store
+spec:
+  accessModes:
+    - ReadWriteOnce
```
camargo: Why is `ReadWriteOnce` used here vs `ReadWriteMany` or `ReadWriteOncePod`?

skovati: Good find. This should be `ReadWriteMany` to allow for multi-node deployments. `ReadWriteOnce` allows one node to mount this volume as R/W, and since local development and the dev deployment only use a single node cluster, this fell through my testing.

skovati: Change has been made!

deployment/kubernetes/aerie-gateway-deployment.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: aerie-dev
```
camargo, line 4: Just curious about the process here. Is there an easy way to change the `namespace` if we want to deploy to another venue at some point?

deployment/kubernetes/aerie-ui-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+      app: aerie-ui
+  template:
+    metadata:
+      labels:
+        app: aerie-ui
+    spec:
+      containers:
+        - env:
+            - name: ORIGIN
+              value: http://127.0.0.1
```
camargo, line 19: How is `ORIGIN` set for the deployment?

deployment/kubernetes/aerie-gateway-deployment.yaml
skovati: As mentioned in the "Future Work" section, I'd like to implement a configuration mechanism like [kustomize](https://kustomize.io/), which is built in to `kubectl`. It would allow us to maintain patch files for dev/prod and apply them to respective environments.

e.g.
```sh
cat kustomization.yaml
```
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

# list of manifests to be applied
resources:
- deployment/kubernetes

# namespace to deploy all Resources to
namespace: aerie-prod
```

This would patch every manifest in our deployment dir, and could change namespaces, replica counts, environment variables, secrets, etc. Aerie users could then just maintain this one patch/overlay file, and apply it on top of our base manifests.

deployment/kubernetes/aerie-ui-deployment.yaml
skovati: `ORIGIN` is currently set to `http://localhost` for the `docker-compose.yaml` based dev deployment so I imagine it could stay the same if/when we switch to the k8s based deployment. Those current env var values are the result of a bit of trial and error getting CORS to work.

camargo: So the deployment docker-compose.yaml right now assumes you are running locally. If we were to use this to deploy to AWS how do we update these env vars? Is there a file on the host we can use to overwrite them?

camargo: Would this also use [kustomize](https://kustomize.io/) to update env vars?

skovati: Once an AWS EKS cluster is set up, it will provide authentication tokens that can be used with `kubectl` to control the cluster remotely, e.g. `kubectl apply -f deployment/kubernetes` would deploy Aerie to the cluster that `~/.kube/config` currently points to. To customize a deployment / environment variables, I see three main options:

- Users maintain their own manifests with hardcoded values (least attractive option)
- The environment variables could be patched onto our base manifests with `kustomize` and applied directly using `kubectl apply -k .`, i.e. users maintain their own `kustomize.yaml` file.
- The base manifests could be updated to to populate env vars them from a `Secret` or `ConfigMap`. End users could then maintain a `.env` file and create a `ConfigMap` with something like `kubectl -n aerie-dev create configmap dev-env --from-env-file=.env`. 

The last option is how I was managing secrets while testing these manifests; right now only "secrets" are managed with an `.env` file, e.g. `AERIE_PASSWORD` is stored in a secret that has to be manually created, and then `aerie-merlin-deployment` can [reference that secret](https://github.com/NASA-AMMOS/aerie/pull/606/files#diff-cb7b2a918b62b075a81c780d77733b7bfe0101ce1ca04ae16ad50a8c9a73ae91R45) to create the `MERLIN_DB_PASSWORD` env var. We could extend this same technique to all environment variables.

skovati: Note that these `kubectl` commands could easily be applied in a `deploy` GitHub action, so this whole thing could be automated.

camargo (approved):
LGTM

PR #617, opened 2023-01-24
Fixed the ordered of repeat args in sequence generation

### LINKED ISSUE #542: Repeat args out of order in commands for sequence eDSL and seq JSON
Commands written in the sequence eDSL with repeat arguments are requiring that the repeat arguments be at the beginning of the command even if they are not specified in the command dictionary at the beginning. Also the `seq.json` output is incorrectly adding the repeat arguments at the beginning of the command. See the examples below.

To fix this the eDSL should take into account the order of repeat arguments in the command dictionary and add them in the correct place. This might require wrapping repeat arguments in an array.

Part of the problem has been traced [to this code](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/src/lib/codegen/CommandTypeCodegen.ts#L127). Also [this old PR](https://github.com/NASA-AMMOS/aerie/pull/210) might be helpful.

### Command Dictionary

```xml
<?xml version="1.0" encoding="UTF-8"?>

<command_dictionary>
  <header mission_name="TEST_DICTIONARY" version="1.0.0.0" schema_version="1.0">
  </header>

  <enum_definitions>
    <enum_table name="some_enum">
      <values>
        <enum symbol="ENUM_A" numeric="1"/>
        <enum symbol="ENUM_B" numeric="2"/>
      </values>
    </enum_table>
  </enum_definitions>

  <command_definitions>
    <fsw_command stem="GNC_TEST_REPEAT_COMMAND" class="FSW">
      <arguments>
        <enum_arg name="arg_1" bit_length="8" enum_name="some_enum">
        </enum_arg>
        <unsigned_arg name="address" bit_length="32" units="None">
        </unsigned_arg>
        <enum_arg name="memory_space" bit_length="8" enum_name="some_enum">
        </enum_arg>
        <repeat_arg name="memory_to_write" prefix_bit_length="8">
          <repeat min="2" max="32">
            <arguments>
              <unsigned_arg name="data_word" bit_length="16" units="None">
                <range_of_values>
                  <include min="0" max="65535"/>
                </range_of_values>
              </unsigned_arg>
            </arguments>
          </repeat>
        </repeat_arg>
      </arguments>
    </fsw_command>
  </command_definitions>
</command_dictionary>
```

### Sequence

Notice the eDSL incorrectly requires the repeat args to be at the beginning of the command. Also notice the `seq.json` is outputing the repeat args at the beginning of the command.

#### eDSL

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    commands: [
      A`2025-012T00:00:00.000`.GNC_TEST_REPEAT_COMMAND(1, 2, 3, 4, "ENUM_A", 10, "ENUM_B")
    ]
  });
````

#### seq.json

```json
{
  "id": "",
  "metadata": {},
  "steps": [
    {
      "args": [
        1,
        2,
        3,
        4,
        "ENUM_A",
        10,
        "ENUM_B"
      ],
      "metadata": {},
      "stem": "GNC_TEST_REPEAT_COMMAND",
      "time": {
        "tag": "2025-012T00:00:00.000",
        "type": "ABSOLUTE"
      },
      "type": "command"
    }
  ]
}
```

cohansen:
Closes #542.

This fixes the order of repeat args in the generated seqjson, it also removes support for the existing list notation and moves to using object notation for adding command arguments.

* **Tickets addressed:** #542 
* **Review:** By commit

## Description
This started as a small fix for the repeat arg ordering, but expanded into quite a large change.

## Verification
Tests were slightly modified to use the new notation.

## Future work
We need to readdress this issue once we know what changes are going to be made to seqjson to support repeat args. Right now there's a case where we could have side-by-side repeat args that are the same type and we'd never be able to know which one the user is intending to use unless they're using the named object notation.

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -169,7 +169,7 @@ export class Command<
 
   public toSeqJson(): CommandSeqJson {
     return {
-      args: typeof this.arguments == 'object' ? Object.values(this.arguments) : this.arguments,
+      args: flatten(this.arguments),
       stem: this.stem,
       time:
         this.absoluteTime !== null
@@ -483,30 +483,30 @@ function commandsWithTimeValue<T extends TimingTypes>(
   }, {} as typeof Commands);
 }
 
-// @ts-ignore
-function orderCommandArguments(args: { [argName: string]: any }, order: string[]): any {
-  return order.map(key => args[key]);
-}
+function flatten(input: { [argName: string]: any }): any {
+  let flatList: any[] = [];
 
-// @ts-ignore: Used in generated code
-function findAndOrderCommandArguments(
-  commandName: string,
-  args: { [argName: string]: any },
-  argumentOrders: string[][],
-): any {
-  for (const argumentOrder of argumentOrders) {
-    if (argumentOrder.length === Object.keys(args).length) {
-      let difference = argumentOrder
-        .filter((value: string) => !Object.keys(args).includes(value))
-        .concat(Object.keys(args).filter((value: string) => !argumentOrder.includes(value))).length;
+  for (const element of Object.values(input)) {
+    // If our input was already flattened, don't try and flatten it again.
+    if (typeof element !== 'object') {
+      return input;
+    }
 
-      // found correct argument order to apply
-      if (difference === 0) {
-        return orderCommandArguments(args, argumentOrder);
+    const values = Object.values(element);
+
+    for (const value of values) {
+      if (Array.isArray(value)) {
+        // We've come across a repeat arg so we need to extract its values.
+        for (const repeat of value) {
+          flatList = flatList.concat([...Object.values(repeat)]);
+        }
+      } else {
+        flatList.push(value);
       }
     }
   }
-  throw new Error(`Could not find correct argument order for command: ${commandName}`);
+
+  return flatList;
 }
 
 function indent(text: string, numTimes: number = 1, char: string = '  '): string {
--- a/sequencing-server/package-lock.json
+++ b/sequencing-server/package-lock.json
@@ -1639,17 +1639,6 @@
         "npm": "1.2.8000 || >= 1.4.16"
       }
     },
-    "node_modules/body-parser/node_modules/qs": {
-      "version": "6.9.7",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.9.7.tgz",
-      "integrity": "sha512-IhMFgUmuNpyRfxA90umL7ByLlgRXu6tIfKPpF5TmcfRLlLCckfP/g3IQmju6jjpu+Hh8rA+2p6A27ZSPOOHdKw==",
-      "engines": {
-        "node": ">=0.6"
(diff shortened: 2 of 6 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -483,30 +483,30 @@ function commandsWithTimeValue<T extends TimingTypes>(
   }, {} as typeof Commands);
 }
 
-// @ts-ignore
-function orderCommandArguments(args: { [argName: string]: any }, order: string[]): any {
-  return order.map(key => args[key]);
-}
+function flatten(input: { [argName: string]: any }): any {
+  let flatList: any[] = [];
```
goetzrrGit, line 487: Super minor, probably want to create the `flatlist` object after the flatten check. No need to allocate an object if we are not going to use it.

goetzrrGit (approved):
It is sad we will lose unnamed parameter objects but this will cover commands with multiple repeating arguments which is super crazy that someone would build a command like that as we didn't even account for that possibility in the first design.

goetzrrGit:
One thing to mention in general about AERIE's commits style is we try to split things into different commits instead of one big one. The idea is to try to tell a story with the commits so it is easier for the reviewers to understand the changes taking place.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: Similarly, arrays can be allocated with `const` and pushed to - the array reference is immutable, but not the underlying contents. It's a bit of a sore point of TypeScript that has caused much conversation about the "scope" of the `const` implications and how it might differ from JavaScript behavior.

cohansen:
> One thing to mention in general about AERIE's commits style is we try to split things into different commits instead of one big one. The idea is to try to tell a story with the commits so it is easier for the reviewers to understand the changes taking place.

I'll start doing this, sorry for the confusion!

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
cohansen: The problem is we need to look at the first element in order to see if the list is already flattened, I can add a check like:
```
const elements = Object.values(input);

if (elements && elements.length > 0 && typeof elements[0] !== 'object') {
```

But I figured that was less readable than just allocating `flatList` first.

goetzrrGit: I would prefer readability since the allocating is so minor.

camargo (approved):
🔥

dyst5422:
> > One thing to mention in general about AERIE's commits style is we try to split things into different commits instead of one big one. The idea is to try to tell a story with the commits so it is easier for the reviewers to understand the changes taking place.
> 
> I'll start doing this, sorry for the confusion!

There is some great documentation regarding this [here](https://github.com/NASA-AMMOS/aerie/blob/develop/CONTRIBUTING.md#write-ups-about-good-commitprcode-review-practice) that is worth reading through at least once

PR #623, opened 2023-01-30
Cardinality goal does not respect applywhen

### LINKED ISSUE #457: Cardinality goal does not respect applyWhen
I want to write a cardinality goal that is only able to satisfy itself with / create activity instances within a given time bounds specified in an `applyWhen` clause. Currently, this goal finds all of activity instances in the plan even if they are very far from the specified window.

```ts
export default (): Goal => {
  return Goal.CardinalityGoal({
    activityTemplate: ActivityTemplates.ResumeBG(),
    specification: { occurrence: 1 },
  }).applyWhen(
    Windows.All(
      Real.Resource('/time/clock.seconds').greaterThan(0),
      Real.Resource('/time/clock.seconds').lessThan(3600),
    ),
  );
};
```

Internal ticket: https://jira.jpl.nasa.gov/browse/AERIEQS-355

adrienmaillard:
* **Tickets addressed:** closes #457 
* **Review:** By commit  

## Description

Card goal was using `ActivityCreationTemplate` to perform the activity search when it should have been `ActivityExpression`. 

## Verification

Added a test in which a card goal has a restricted application period and activities already present in the plan outside of it. The solver does not consider activities outside of the bounds and has to insert activities to satisfy the goal. 

## Documentation

None. 

## Future work

All the work on the differentiation between activity creation and the activity satisfaction.

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
@@ -334,6 +334,7 @@ public class ActivityExpression implements Expression<Spans> {
       endsIn = template.endRange;
       durationIn = template.durationRange;
       startsOrEndsIn = template.startOrEndRange;
+      startsOrEndsInW = template.startOrEndRangeW;
       arguments = template.arguments;
       return getThis();
     }
--- a/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
+++ b/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
@@ -2,12 +2,15 @@ package gov.nasa.jpl.aerie.scheduler;
 
 import com.google.common.testing.NullPointerTester;
 import com.google.common.truth.Correspondence;
+import gov.nasa.jpl.aerie.constraints.time.Interval;
 import gov.nasa.jpl.aerie.constraints.time.Windows;
 import gov.nasa.jpl.aerie.constraints.tree.WindowsWrapperExpression;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityCreationTemplate;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityExpression;
 import gov.nasa.jpl.aerie.scheduler.constraints.timeexpressions.TimeAnchor;
+import gov.nasa.jpl.aerie.scheduler.goals.CardinalityGoal;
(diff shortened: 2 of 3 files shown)


scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
```diff
@@ -228,4 +231,48 @@ public void getNextSolution_coexistenceGoalOnActivityWorks() {
...
+
+    //act at t=1hr and at t=2hrs
+    problem.setInitialPlan(makePlanA12(problem));
+
+    //and the goal window in between [0, 50 min[ so the activities already present in the plan can't count towards satisfying the goal
+    final var goalWindow = new Windows(false).set(List.of(
+        Interval.between(Duration.of(0, Duration.SECONDS), Duration.of(50, Duration.MINUTE))
+    ), true);
+
+    //unsatisfiable because 10 occurences asked when only one can be placed
```
mattdailis: Is this comment still accurate?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -334,6 +334,7 @@ Builder basedOn(@NotNull ActivityExpression template) {
       endsIn = template.endRange;
       durationIn = template.durationRange;
       startsOrEndsIn = template.startOrEndRange;
+      startsOrEndsInW = template.startOrEndRangeW;
```
mattdailis, line 337: Do you recall what the difference is between `startsOrEndsIn` and `startsOrEndsInW`? Not essential for this PR - it just strikes me as a possible point of confusion.

mattdailis (approved):
I verified that the test fails before this change, and passes afterwards :+1: I'm not sure I grasped the nuance though

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: Yes, one is just for one interval, the other for a set of intervals. One is reducible to the other. I am not why we have both but I believe this should be cleaned when we are going to work on the activity search vs activity creation.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
adrienmaillard: No, it is not, will remove before merge, thanks.

adrienmaillard:
The nuance is that the `ActivityExpression` uses `startsOrEndsInW` in the search but not `ActivityCreationTemplate` even though it can be specified. These two inherit from each other but they are hard to read currently. I'll probably separate them when we do this part of the work.

PR #632, opened 2023-02-01
Make scheduler's backtracking configuration accessible to user

### LINKED ISSUE #624: Controllable partial satisfaction for goals
When a goal can't be satisfied, either the solver backtracks or not. If it backtracks, it will remove activities inserted in the plan to try satisfying the goal. If it does not backtrack, it will let the activities inserted so far in the plan.

The first approach means "satisfy the goal completely or nothing". The second approach means "satisfy the goal as much as possible". 

The backend is already equipped with this mechanism but the edsl does not have the knob for the user to change this behavior for each goal. By default, the solver does not backtrack and follows a global non-backtracking policy. The idea behind this ticket is to allow the user to set this parameter for each goal.

adrienmaillard:
* **Tickets addressed:** Closes #624 
* **Review:** By commit  

## Description

See the ticket #624 for info on what this is. 

In the eDSL, we allow the user to set the parameter for each goal. 

In the backend, checked and updated the backtrack policy as well as making the name of the configuration parameter much clearer (`partialSatisfaction` -> `shouldRollbackIfUnsatisfied`).

By default, nothing has changed, as the default behavior is not to backtrack. 

## Verification

I have written several tests that demonstrate the behavior. 

## Documentation

I need to write documentation for this PR as well as the previous one. 

## Future work

Nothing regarding this particular aspect.

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
@@ -17,12 +17,26 @@ import java.util.List;
  * describes some criteria that is desired in the solution plans
  */
 public class Goal {
+  /**
+   * the human-legible identifier of the goal
+   * never null
+   */
+  protected String name;
 
+  /**
+   * the contiguous range of time over which the goal applies
+   */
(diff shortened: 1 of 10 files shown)


mattdailis (approved):
Looks reasonable to me

PR #638, opened 2023-02-01
Merge Sequence eDSL API with SeqJson interface

### LINKED ISSUE #610: Merge Sequence EDSL API and the  JsonSpec Interface.
The proposed plan is to incorporate the JSON spec into the Aerie Repo, granting EDSL access to the full JSON Spec, and to make the Sequence EDSL API use this new SPEC while preserving backward compatibility for now which will not break Clippers work.

```
Sequence.new({
   seqId: '',
   metadata: {},
   steps: [ commands, ground, loads, activates, etc. ],
   request: [],
   
   commands : [] <--- preserve backward compatibility for now
 });
```

goetzrrGit:
* **Tickets addressed:** Closes #610
* **Review:** By commit  

## Description
We are merging the Seq JSON Spec and the eDSL, adapting the Sequence eDSL to use the SeqJson Interface indicated in the Spec. This so far has not impacted Clipper compatibility.

## Verification
I ran the e2e test locally and there were no issues.

## Documentation
No documentation at this point as the user will not be aware of this change.

## Future work
Merge `CommandStem` eDSL API with `Command` Interface.

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -52,6 +52,26 @@ export interface CommandSeqJson<A extends ArgType[] = ArgType[]> {
 export type ArgType = boolean | string | number;
 export type Arrayable<T> = T | Arrayable<T>[];
 
+export interface SequenceOptions {
+  seqId: string;
+  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
+  locals?: [VariableDeclaration, ...VariableDeclaration[]];
+  // @ts-ignore : 'Metadata' found in JSON Spec
+  metadata: Metadata;
+  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
+  parameters?: [VariableDeclaration, ...VariableDeclaration[]];
+  // @ts-ignore : 'Step' found in JSON Spec
+  steps?: Step[];
+  // @ts-ignore : 'Request' found in JSON Spec
+  requests?: Request[];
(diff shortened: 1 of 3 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -265,30 +318,58 @@ export class CommandStem<
...
+    }
+    if (opts.requests) {
+      this.requests = opts.requests;
+    }
+    if (opts.immediate_commands) {
+      this.immediate_commands = opts.immediate_commands;
+    }
+    if (opts.hardware_commands) {
+      this.hardware_commands = opts.hardware_commands;
+    }
```
cohansen, line 363: How do you guys feel about this or using null coalescing for each value?
```
this.steps = opts.steps ?? undefined;
```

cohansen (approved):
Looks good, just 1 style comment from me.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I am in favor!!!

goetzrrGit: @cohansen I will make this change when I change CommandStem in the next PR. The `Sequence` Class has more to get `Commands` working for the `Step` field

cohansen: Sounds good!

PR #654, opened 2023-02-02
Update all the sequencing server packages

### LINKED ISSUE #536: Update all dependencies for the sequencing-server
There is a [security alert](https://github.com/NASA-AMMOS/aerie/security/dependabot/4) for the [sequencing-server](https://github.com/NASA-AMMOS/aerie/tree/develop/sequencing-server). We should just do a pass over [all the dependencies](https://github.com/NASA-AMMOS/aerie/blob/develop/sequencing-server/package.json) and update them to the latest versions.

cohansen:
Going to merge after the current release happens.

* **Tickets addressed:** Closes #536.
* **Review:** By commit

## Description

- Updated all the packages manually, and switched from `^` to `~`.
- Added name to data loaders to support package updates

## Verification
Unit tests were run, which failed so updates were made to the code to support the dependency updates.

## Documentation
NA

## Future work
Packages are set to `~` now, so only patch updates will happen automatically going forward.


sequencing-server/src/app.ts
```diff
@@ -64,11 +64,13 @@ app.use(async (_: Request, res: Response, next: NextFunction) => {
 
   const activitySchemaDataLoader = new DataLoader(activitySchemaBatchLoader({ graphqlClient }), {
     cacheKeyFn: objectCacheKeyFunction,
+    name: null,
```
goetzrrGit, line 67: can you tell me what the new name fields are for

goetzrrGit (approved):

sequencing-server/src/app.ts
cohansen: `The name given to this DataLoader instance. Useful for APM tools.` Just pulled from the docs, I don't think we need it but it's required as an option.

Issue #595, opened 2023-01-13
Long delay before UI acknowledges simulation request status


mattdailis:
This ticket to be filled out with more information as it becomes known.

AERIEQS ticket: [AERIEQS-427](https://jira.jpl.nasa.gov/browse/AERIEQS-427)

Sometimes, when clicking the simulate button, the UI does not display any simulation status for a very long time.

I've observed this a couple times locally - unfortunately I don't have screenshots or data at the moment - I will supply them if I see it again.

What I did observe was what appeared to be a deadlock. I ran the query below and saw that some queries (subscriptions from the UI, I think?) were both blocking and blocked by a `create partition` query. I interpret this to mean that the UI subscription to the simulation results is preventing the simulation from starting - and somehow neither query is able to go through. (For a deadlock to happen, there need to be at least two locks involved - so I'm a little unsure what the salient locks are here).

<details>
<summary>
SQL to see what queries are blocked by what other queries
</summary>

```sql
SELECT blocked_locks.pid     AS blocked_pid,
       blocked_activity.usename  AS blocked_user,
       blocking_locks.pid     AS blocking_pid,
       blocking_activity.usename AS blocking_user,
       blocked_activity.query    AS blocked_statement,
       blocking_activity.query   AS current_statement_in_blocking_process
FROM  pg_catalog.pg_locks         blocked_locks
          JOIN pg_catalog.pg_stat_activity blocked_activity  ON blocked_activity.pid = blocked_locks.pid
          JOIN pg_catalog.pg_locks         blocking_locks
               ON blocking_locks.locktype = blocked_locks.locktype
                   AND blocking_locks.database IS NOT DISTINCT FROM blocked_locks.database
                   AND blocking_locks.relation IS NOT DISTINCT FROM blocked_locks.relation
                   AND blocking_locks.page IS NOT DISTINCT FROM blocked_locks.page
                   AND blocking_locks.tuple IS NOT DISTINCT FROM blocked_locks.tuple
                   AND blocking_locks.virtualxid IS NOT DISTINCT FROM blocked_locks.virtualxid
                   AND blocking_locks.transactionid IS NOT DISTINCT FROM blocked_locks.transactionid
                   AND blocking_locks.classid IS NOT DISTINCT FROM blocked_locks.classid
                   AND blocking_locks.objid IS NOT DISTINCT FROM blocked_locks.objid
                   AND blocking_locks.objsubid IS NOT DISTINCT FROM blocked_locks.objsubid
[...]

PR #661, opened 2023-02-06
Minimize locks acquired for dataset partitions

mattdailis:
* **Tickets addressed:** Related to #595 
* **Review:** By commit  

## Description

**Problem:** Inserting simulation results claims locks that conflict with creating new dataset partitions
**Proposed solution:** Reduce the locks acquired when creating new dataset partitions

Steps taken:
1. Follow [this](https://stackoverflow.com/a/67016755) advice to replace the single-command partition creation with the two-command equivalent. This removes the need to acquire an AccessExclusiveLock on the whole partitioned table when creating a new dataset partition.
2. Replace the following foreign keys with constraint triggers: `profile_segment->profile`, `event->topic`, `span->dataset`. This removes the need to acquire a ShareRowExclusiveLock on `profile`, `topic`, and `dataset` when creating a new dataset partition.
3. Replace the `span->span` self-referencing foreign key with a separate foreign key per partition - e.g. `span_15->span_15`. This removes the need to acquire a ShareRowExclusiveLock on `span` when creating a new dataset partition.

Advantages:
- Inserting simulation results and creating new dataset partitions are independent - neither blocks the other. This allows us to focus on speeding up inserts without worrying about blocking concurrent partition creation.

Disadvantages:
- Triggers add complexity and maintenance burden. Attaching triggers and foreign keys to each partition by hand means that future migrations need to iterate over all partition tables and `alter` them separately.

Alternatives considered:
- Lock no more than one table at a time when inserting.
  This eliminates the deadlock, but attaching a new partition is still blocked until the lock is released - which, for `profile_segment`, can be a very long time
- Guarantee that all inserting transactions commit or rollback within some time limit.
   This is a good practice in general - and I do think we should take steps in this direction - but I do not like the idea of our design hinging on this limit. It introduces tension - the longer the limit, the longer the response time on simulation requests. This effect stacks up if you have alternating inserts and simulation requests. While we can work around this by attaching the partition in a concurrent thread - it becomes quite an intricate dance. At the other extreme, making the limit too short constrains the effort to speed up the writing of simulation results - which is a problem for which I'd like to leave the design space wide open, if we can.

## Verification

To reproduce the deadlock issue on develop:
1. Bring up aerie
9. Stop aerie_merlin_worker_2
10. Attach a debugger to aerie_merlin_worker_1 and set a breakpoint here:
    https://github.com/NASA-AMMOS/aerie/blob/602aeda0456d2fc4514b8966fe46fdc5adf9a49a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java#L117
11. Make a plan, run a simulation - observe that the worker stops at the breakpoint
12. Bring up aerie_merlin_worker_2
13. Edit the plan, run another simulation
14. Go back to the debugger for aerie_merlin_worker_1, and hit continue

The postgres log will contain the following:
```
postgres_1                   | 2023-02-06 20:11:49.017 UTC [500] ERROR:  deadlock detected at character 13
postgres_1                   | 2023-02-06 20:11:49.017 UTC [500] DETAIL:  Process 500 waits for RowExclusiveLock on relation 16624 of database 16386; blocked by process 509.
postgres_1                   |  Process 509 waits for ShareRowExclusiveLock on relation 16606 of database 16386; blocked by process 500.
postgres_1                   |  Process 500: insert into profile_segment (dataset_id, profile_id, start_offset, dynamics, is_gap)
postgres_1                   |    values ($1, $2, $3::interval, $4::json, $5)
postgres_1                   |  Process 509: create table profile_segment_63 partition of profile_segment for values in (63)
postgres_1                   | 2023-02-06 20:11:49.017 UTC [500] HINT:  See server log for query details.
postgres_1                   | 2023-02-06 20:11:49.017 UTC [500] STATEMENT:  insert into profile_segment (dataset_id, profile_id, start_offset, dynamics, is_gap)
postgres_1                   |    values ($1, $2, $3::interval, $4::json, $5)
```

To reproduce blocking issue - follow the same steps as above, but stop at this breakpoint instead:
https://github.com/NASA-AMMOS/aerie/blob/602aeda0456d2fc4514b8966fe46fdc5adf9a49a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java#L275

Follow the same steps on this branch, and verify that no deadlock or blocking occurs.

- [x] TODO: Prior to merging, I would like to add db-tests for the new triggers.

## Documentation

- [x] TODO: Prior to merging, add detailed comments to the triggers and prepared statement.

We can mention this in release notes, but I don't think there's a need for user-facing documentation.

I should probably mention this in the SDD though.

## Future work

- When speeding up simulation result writing, we'll likely want to experiment with delaying these triggers, or perhaps rewrite them as per-statement triggers


Twisol:
> Replace the following foreign keys with constraint triggers

Can we *heavily* document these triggers, ideally with `COMMENT` commands, to describe why we're not just using foreign keys? If somebody came along and saw this, their first thought *should* be "what are you doing, why are you reimplementing something Postgres gives you for free". We should anticipate this by explaining "well, *ackshually*, it's not free -- here's the cost and here's why ours is better for our needs".

I know it's all here in the PR (and related PRs), in principle, but in practice the code is the first (and sometimes last) port of call for understanding the code. We should front-load things a bit when we're doing something so obviously odd.

> 3\. Replace the `span->span` self-referencing foreign key with a separate foreign key per partition

Arguably, this is *more* precise, anyway, since you shouldn't have a span in one dataset parented to a span in another -- span trees should not span (heh) multiple datasets. :+1:

> Disadvantages:

Another disadvantage is that even less of the actual structure of a database instance is captured by our statically-defined SQL schemas. Before, the only deviation is that the `profile_segment` etc. tables had a class of `profile_segment_#` (etc.) tables associated to it; but those tables are the same shape as `profile_segment` (etc.), so we still had a clear picture. Now, the partitions are not exactly the same shape, so it is harder to know what's going on without looking at a runtime instance and making inferences.

This can be accommodated for, to some extent, by defining a stored procedure that captures the shape of the things we can add to the system -- and I'm glad to see that this is indeed part of the PR. Still, I think it's worth noting that the degree to which the system can deviate from the schemas has grown just a little bit.

Twisol:
Could we consider leveraging `CREATE TABLE ... LIKE ... INCLUDING ALL` to simplify `allocate_dataset_partitions`? Ideally, we would do a `CREATE TABLE profile_segment_TEMPLATE` (etc.) in the top-level schemas, so that all the constraints and such are defined tidily in one place, without needing all the string templating and concatenation. We could then use `CREATE TABLE LIKE` to effectively clone these templates into new partition instances.

mattdailis:
> Could we ... simplify `allocate_dataset_partitions`?

Yes! Turns out we can. We can have the triggers automatically be copied from the parent table. From the [postgres docs](https://www.postgresql.org/docs/current/sql-altertable.html):

> `ATTACH PARTITION` ... For each user-defined row-level trigger that exists in the target table, a corresponding one is created in the attached table.

So I moved those triggers to the top level table :+1: Unlike the foreign keys - copying triggers does not lock the referenced table :100:

The only thing I didn't move was the span_#->span_# self referencing foreign key - because copying this foreign key does take a lock on the parent table.

Twisol:
That's excellent! I'm glad something like what I suggested ended up being possible 🚀

Twisol:
That's excellent! I'm glad something like what I suggested ended up being possible 🚀

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
```diff
@@ -0,0 +1,41 @@
...
+        alter table profile_segment
+          attach partition profile_segment_%d for values in (%d);
+        """.formatted(datasetId, datasetId, datasetId, datasetId);
+    return sql;
+  }
+
+  @Override
+  public void close() throws SQLException {
+    // Nothing to clean up. The method is an intentionally-blank override.
+  }
```
Mythicaeda: We should probably close `statement` here.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
```diff
@@ -0,0 +1,41 @@
...
+    final @Language("SQL") String sql =
+        """
+        create table profile_segment_%d (
+           like profile_segment including defaults including constraints
+        );
+        alter table profile_segment
+          attach partition profile_segment_%d for values in (%d);
+        """.formatted(datasetId, datasetId, datasetId, datasetId);
+    return sql;
+  }
```
Mythicaeda: This method is no longer used.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
```diff
@@ -868,4 +870,321 @@ void shouldCancelSimulationOnSimulationTemplateUpdate() throws SQLException {
...
+        statement
+            .executeUpdate(
+              """
+              INSERT INTO profile_segment (dataset_id, profile_id, start_offset, dynamics, is_gap)
+              VALUES (%d, 1256, '0 seconds', '{}', false);
+              """.formatted(datasetId)
+          );
+      } catch (SQLException e) {
+        assertTrue(e.getMessage().contains("foreign key violation: there is no profile with id 1256 in dataset %d".formatted(datasetId)), e.getMessage());
+      }
```
Mythicaeda: With regards to this `assertTrue`, I'll just mention that I tend to express this as

```java
if (!e.getMessage().contains("<expected exception>"))
  throw e;
```

That way if there's a `SQLException` thrown that I'm not expecting, it's actually, y'know, thrown.

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,284 @@
+alter table profile_segment drop constraint profile_segment_owned_by_profile;
+alter table event drop constraint event_owned_by_topic;
+
+alter table span drop constraint span_has_parent_span;
+alter table span drop constraint span_owned_by_dataset;
+
+comment on table profile is e''
+  'Metadata describing a resource profile in the context of a dataset.';
+comment on column profile.type is e''
+  'The type of behavior this profile expresses. Any segments associated with this profile should abide by this type.';
```
Mythicaeda: These comments on `profile` do not match the comments on `profile` described in `profile.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,284 @@
+alter table profile_segment drop constraint profile_segment_owned_by_profile;
```
Mythicaeda, line 2: General style proposal for `up.sql`s: Can we have edits grouped by file, at least while we're still verifying them by hand? 

Having them grouped by file means that we can put the modified file and the up.sql side by side and see very quickly that all the changes in one are in the other, while having the changes grouped in another way in the `up.sql` means that each change in the modified file has to be searched for in the up file and likewise each change in the up file has to be found in one of the modified db files.

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,284 @@
...
+  returns trigger
+  security invoker
+  language plpgsql as $$begin
+  if old.id != new.id or old.dataset_id != new.dataset_id
+  then
+    update profile_segment
+    set profile_id = new.id,
+        dataset_id = new.dataset_id
+    where profile_segment.dataset_id = old.dataset_id and profile_segment.profile_id = old.id;
+  end if;
```
Mythicaeda, line 89: This sounds silly, but the function is indented differently here than in `profile.sql`, which I believe means they would fail a db equality check (if not, feel free to mark this as resolved!)

merlin-server/sql/merlin/tables/dataset.sql
```diff
@@ -32,3 +32,64 @@ create trigger delete_partitions_trigger
...
+      like event including defaults including constraints
+    );';
+  execute 'alter table event
+    attach partition event_' || dataset_id || ' for values in (' || dataset_id || ');';
+
+  execute 'create table span_' || dataset_id || ' (
+       like span including defaults including constraints
+    );';
+  execute 'alter table span
+    attach partition span_' || dataset_id || ' for values in (' || dataset_id || ');';
```
Mythicaeda, line 87: Very minor: [The docs](https://www.postgresql.org/docs/15/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN) recommend using variables and `USING` instead of concatenating the string, as "it avoids run-time overhead of converting the values to text and back". Was there a reason you chose to do concatenation instead?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
mattdailis: Good catch, done :+1:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
mattdailis: Method has been un-existed :+1:

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
mattdailis: That's a really great idea. Adopted 🚀

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Oops, that was from a previous approach I considered where we would have neither foreign keys nor triggers. Comment changes have been deleted :+1:

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Sounds reasonable to me - I added a new commit reorganizing the two files - if I have a spare moment I'll rework the commits so it looks like I did that all along.

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Not silly, good eye 👁️  I've updated the version in `profile.sql` since I think I like the indentation in `up.sql` better.

merlin-server/sql/merlin/tables/dataset.sql
mattdailis: Thanks for sharing that - I was unaware of that option. I'll take a look at it this afternoon and see if I think it's a good fit

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,286 @@
+-- profile_segment
+alter table profile_segment drop constraint profile_segment_owned_by_profile;
+
+create or replace function profile_segment_integrity_function()
+  returns trigger
+  security invoker
+  language plpgsql as $$begin
+  if not exists(
+      select from profile
+      where profile.dataset_id = new.dataset_id
```
Mythicaeda, line 10: This line is indented differently in `profile_segment.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,286 @@
...
+        and profile.id = new.profile_id
+        for key share of profile)
```
Mythicaeda: Again, different indentation level

merlin-server/sql/merlin/tables/dataset.sql
mattdailis: I think the challenge here is that we're using the dataset_id to determine the name of the table itself - the docs I think are describing the ability to define variables that can be dynamically populated with values (similarly to how prepared statements work) - but I don't think that gives us the ability to use that variable in the name of the table

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,286 @@
...
+
+create or replace function event_integrity_function()
+  returns trigger
+  security invoker
+  language plpgsql as $$begin
+  if not exists(
+      select from topic
+      where topic.dataset_id = new.dataset_id
+        and topic.topic_index = new.topic_index
+        for key share of topic)
```
Mythicaeda: Similar indentation differences between here and `event.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,286 @@
...
+  for each row
+execute function event_integrity_function();
+
+-- profile
+create or replace function delete_profile_cascade()
+  returns trigger
+  security invoker
+  language plpgsql as $$begin
+  delete from profile_segment
+  where profile_segment.dataset_id = old.dataset_id and profile_segment.profile_id = old.id;
```
Mythicaeda, line 70: Indentation differs between here and `profile.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+  'Trigger to simulate an ON UPDATE CASCADE foreign key constraint between event and topic. The reason to'
+  'implement this as a trigger is that this single trigger can propagate updates to any partitions of event.'
+  'If we used a foreign key, every new partition of event would need to add a new trigger to the topic'
+  'table - which requires acquiring a lock that conflicts with concurrent inserts. In order to allow adding new'
+  'partitions concurrently with inserts to referenced tables, we have chosen to forego foreign keys from partitions'
+  'to other tables in favor of these hand-written triggers';
+
+-- span
+alter table span drop constraint span_has_parent_span;
+alter table span drop constraint span_owned_by_dataset;
```
Mythicaeda, line 166: Minor, can ignore: These can be combined.

merlin-server/sql/merlin/tables/span.sql
```diff
@@ -42,3 +32,39 @@ comment on column span.type is e''
...
+  'dataset table when creating a new partition of span. This function checks that a corresponding dataset'
+  'exists for every inserted or updated span. A trigger that calls this function is added separately to each'
+  'new partition of span.';
+
+create constraint trigger insert_update_span_trigger
+  after insert or update on span
+  for each row
+execute function span_integrity_function();
+
+create or replace procedure span_add_foreign_key_to_partition(table_name varchar)
```
Mythicaeda: Should this just be `create procedure`, and not `create or replace`? Same with `span_integrity_function()` and in general with the added functions in this PR.

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+    dataset_id integer;
+  begin
+    for dataset_id in select id from dataset
+      loop
+        if exists(select from pg_tables where schemaname = 'public' and tablename = 'span_' || dataset_id) then
+          execute 'alter table span_' || dataset_id || ' add constraint span_has_parent_span foreign key (dataset_id, parent_id) references span_' || dataset_id || ' on update cascade on delete cascade;';
+        end if;
+      end loop;
+  end
+$$;
```
Mythicaeda: Is this essentially running `span_add_foreign_key_to_partition` to all of the existing partition tables?

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+  'Creates a self-referencing foreign key on a particular partition of the span table. This should be called'
+  'on every partition as soon as it is created';
+
+-- dataset
+create or replace function delete_dataset_cascade()
+  returns trigger
+  security definer
+  language plpgsql as $$begin
+  delete from span where span.dataset_id = old.id;
+  return old;
```
Mythicaeda: Indentation differs in `dataset.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+end$$;
+
+create or replace function delete_dataset_cascade()
+  returns trigger
+  security definer
+  language plpgsql as $$begin
+  delete from span where span.dataset_id = old.id;
+  return old;
+end$$;
```
Mythicaeda: Duplicated code

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+
+create trigger delete_dataset_trigger
+  after delete on dataset
+  for each row
+execute function delete_dataset_cascade();
```
Mythicaeda, line 230: Indentation differs in `dataset.sql`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
```diff
@@ -0,0 +1,290 @@
...
+  execute 'alter table span
+    attach partition span_' || dataset_id || ' for values in (' || dataset_id || ');';
+
+  -- Create a self-referencing foreign key on the span partition table. We avoid referring to the top level span table
+  -- in order to avoid lock contention with concurrent inserts
+  execute 'alter table span_' || dataset_id || ' add constraint span_has_parent_span
+    foreign key (dataset_id, parent_id)
+    references span_' || dataset_id || '
+    on update cascade
+    on delete cascade;';
```
Mythicaeda: This part of the method differs here from the version in dataset.sql

There, this part is: `call span_add_foreign_key_to_partition('span_' || dataset_id);`

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/down.sql
```diff
@@ -0,0 +1,75 @@
+-- profile_segment
+drop trigger if exists insert_update_profile_segment_trigger on profile_segment;
+drop function if exists profile_segment_integrity_function;
```
Mythicaeda: I don't think we should have the "if exists" modifier. I believed we've discussed on a prior PR that if the database state is wrong when we're migrating, don't we want to be loud about it?

merlin-server/sql/merlin/tables/dataset.sql
Mythicaeda: It looks like we could use the `format(%I)` approach to dynamically determine the name of the table. As mentioned towards the bottom of that section:
`A cleaner approach is to use format()'s %I specification to insert table or column names with automatic quoting`

It does look like we would end up concatenating the table names, though, so unless we can merge some of these statements, I'm not sure how useful that would be

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: It is indeed :+1: I've updated it to call that instead

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Thanks - I ran a diff and I think I've ironed out the indentation differences. (There are some differences left over from anchors - but I won't touch those in this PR)

merlin-server/sql/merlin/tables/span.sql
mattdailis: I suppose the idea is that this code only runs on an empty database - so there shouldn't be a difference (except in cases where we've duplicated a function inadvertently - which would be nice to know about). I can make the change for the new functions - there are likely old functions/triggers around that use `or replace`.

mattdailis: Done :+1:

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Deleted, good catch (this is where the `or replace` is masking mistakes 👍 )

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
mattdailis: Good catch - updated :+1:

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/down.sql
mattdailis: I went through and dropped if exists if it exists

Twisol: How meta!

merlin-server/sql/merlin/tables/dataset.sql
mattdailis: I gave it a shot, but ran into this restriction mentioned in the docs:
> Another restriction on parameter symbols is that they only work in optimizable SQL commands (SELECT, INSERT, UPDATE, DELETE, MERGE, and certain commands containing one of these). In other statement types (generically called utility statements), you must insert values textually even if they are just data values.

Looks like we're out of luck for `CREATE TABLE` and `ALTER TABLE`

mattdailis: That restriction is regarding the `$1` variables - format should still work

mattdailis: Here's what it looks like with format - thoughts on whether the added verbosity helps readability? https://github.com/NASA-AMMOS/aerie/commit/52b7e641ec8de817810a1f27639708defc64da01?diff=split

Mythicaeda: I don't think it does with how short out statements are. I do like the partition table names being created in advance, as it guarantees there's no typos and the hardest part to parse about the original was the concatenated table name, but it's not a really necessary change.

mattdailis: Sounds good - I'll drop it from this PR - but we can keep it in mind if we find ourselves writing more complex `execute` statements. (The `execute` statements in an earlier iteration of this PR were much more involved - I'm glad they're not too bad now)

deployment/hasura/migrations/AerieMerlin/2_drop_profile_topic_fks/up.sql
Mythicaeda: I've double-checked all of my "indentation differs" comments and left the three that still apply.

mattdailis: Ah, I see now, those indentation differences are not in the code of a function, so they won't cause problems in a diff :+1: (The reason I know that is I still have my local code to do the pg_dump diff) I'll adjust that indentation for the sake of consistency, but I don't believe it will cause any issues.

EDIT: I see one of them actually is in a function... interesting that I missed it in the diff. Will update.

Mythicaeda: Ah, no, no need! I complete forgot that it only matters in the body of the function!

EDIT: Two of them, actually, this one and the other one. (`create function delete_dataset_cascade()`)

Mythicaeda (approved):
LGTM!

mattdailis:
I re-ran the manual tests in the PR description, and practiced a migration from develop. Everything seems to be working okay.

Something that occurred in my manual testing: if the migration is run, but the docker containers are not updated, the system will mostly continue to function - the only difference is that new `span` partitions will be missing their self-referencing foreign keys. We can mention this in the upgrade guide - "Please bring down the `aerie_merlin_worker` docker containers before running this migration, and upgrade them before bringing them back up".

PR #665, opened 2023-02-08
Window reference in coexistence goal

### LINKED ISSUE #444: Specify the duration of an activity to match the duration of a window
For scheduling goals we should be able to define duration parameters in an activity template to match the duration of a window (as long as the scheduler can fill the duration).

JoelCourtney:
* **Tickets addressed:** closes #444 
* **Review:** By commit  

## Description

In no particular order:
- Refactored eDSL intervals into an expression. Now they can be literals or aliases.
- Made a `IntervalDuration(Expression<Interval>)` node, and a `DurationLiteral(Duration)` node, which are both `Expression<Duration>` (NOT `Expression<DiscreteProfile>`)
- Made a `DiscreteProfileFromDuration(Expression<Duration>)` node to convert durations into discrete profiles for internal use in the scheduler (not exposed to user).
- Activity template arguments now accept Duration expressions, which is automatically wrapped in `DiscreteProfileFromDuration` and evaluated when attempting to match activities/create conflicts.
- Because of this, I had to remove argument sanitation checks on CardinalityGoal, because the duration argument is computed after goal creation.

## Verification
I had to remove one test for CardinalityGoal, see last bullet point. Updated existing tests to hit duration and interval literals.

As it turns out, the single scheduling integration test that I added covers:
- Interval alias node
- Interval duration node
- coexistence goal interface
- activity expression with lazy duration
for eDSL execution, parsing, and evaluation.

## Documentation
- [x] documentation written
- [ ] documentation merged


JoelCourtney:
rebasing on develop caused an issue, looking into it

mattdailis:
> rebasing on develop caused an issue, looking into it

This seems to help with that particular issue:
https://github.com/NASA-AMMOS/aerie/commit/7538efae5a36d6990cc9e21b31c7d2611e9c9003

~I haven't looked into the merge conflicts yet~

EDIT: This rebase addresses the merge conflicts:

https://github.com/NASA-AMMOS/aerie/tree/5d25a778081cc1c56b2241e1d4dcd9f6a6bd7030

JoelCourtney:
@mattdailis I stole the commits from your branch, thanks

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -107,32 +107,8 @@ protected CardinalityGoal fill(CardinalityGoal goal) {
...
-    }
-
-    public boolean isUnsatisfiable(){
-      return this.durationRange != null &&
-             occurrenceRange != null &&
-             this.thereExists.getDurationRange() != null &&
-             this.thereExists.getDurationRange().start
-                 .times(occurrenceRange.getMinimum())
-                 .longerThan(durationRange.end);
-    }
```
adrienmaillard, line 134: Why a check of this flavor is not possible anymore ?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/EvaluationEnvironment.java
```diff
@@ -8,10 +9,15 @@
 public record EvaluationEnvironment(
     Map<String, ActivityInstance> activityInstances,
     Map<String, Spans> spansInstances,
+    Map<String, Interval> intervals,
```
adrienmaillard, line 12: Can you explain why you needed to add this ?

adrienmaillard (approved):
Nice work, I only have some questions.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/EvaluationEnvironment.java
JoelCourtney: I got tired of using collections to store single objects. Like using a Windows to store something that is guaranteed to be a single Window, or a Spans to store a guaranteed single Span. So I decided the most natural argument for the CoexGoal template factory would be an `Interval` instead of a `Windows` that only has one window. This caused all the Interval-related changes to the edsl. It also meant we need to store intervals in the environment so CoexGoal can generate the Interval, and the ast nodes in the factory can access it.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
JoelCourtney: They use `getDurationRange()`. I had to change the `Interval durationRange` field of `ActivityExpression` into `Expression<? extends Profile<?>> duration`. These unsatisfiability checks are sanity checks before the goal is run, but we can no longer check the duration before evaluation because the duration is lazy.

PR #672, opened 2023-02-14
Added handling for activities that stretch past a plans duration

### LINKED ISSUE #655: Expansion fails when plan has unfinished simulated activities
When a simulated plan includes unfinished activities (i.e., activities whose effect models don't return before the end time of the plan), command expansion breaks. As far as I can tell, this is because the command expansion server tries to parse the `null` values for end time and duration for the unfinished activity instances and breaks.

Internal ticket: https://jira.jpl.nasa.gov/browse/AERIEQS-439

cohansen:
* **Tickets addressed:** Closes #655
* **Review:** By commit

## Description
Now if the user is trying to expand and we come across an activity type whose effect model goes past the end of the plan the expansion will have a duration error for that specific activity.

## Verification
I added a test for this scenario alongside our existing tests.

## Documentation
N/A

## Future work
N/A

### DIFF
--- a/sequencing-server/src/app.ts
+++ b/sequencing-server/src/app.ts
@@ -365,6 +365,15 @@ app.post('/expand-all-activity-instances', async (req, res, next) => {
 
   const settledExpansionResults = await Promise.allSettled(
     simulatedActivities.map(async simulatedActivity => {
+      // The simulatedActivity's duration and endTime will be null if the effect model reaches across the plan end boundaries.
+      if (!simulatedActivity.duration && !simulatedActivity.endTime) {
+        return {
+          activityInstance: simulatedActivity,
+          commands: null,
+          errors: [{ message: 'Duration is null' }],
+        };
+      }
+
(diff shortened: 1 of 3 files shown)


goetzrrGit (approved):
looks great to me.

sequencing-server/package-lock.json
```diff
@@ -1642,6 +1642,17 @@
         "npm": "1.2.8000 || >= 1.4.16"
       }
     },
+    "node_modules/body-parser/node_modules/qs": {
```
camargo: Do we need this lock file change?

cohansen: Yeah, it looks like the qs patch version was updated.

dyst5422: Bummer that body-parser still uses qs despite URLSearchParams for a long while now.

https://dev.to/nerdyman/replacing-query-string-with-native-urlsearchparams-4kdg

camargo (approved):

PR #683, opened 2023-02-17
JSON Spec Updates and SEQ JSON to EDSL fix

### LINKED ISSUE #668: Uploading generated seqjson uses old list notation
When uploading generated seqjson inside the Sequence Editor it generates the old list notation rather than the new object notation.

We may need to wait until we've updated to using the latest seqjson before addressing this ticket.

goetzrrGit:
* **Tickets addressed:** Closes #668
* **Review:** By commit  

## Description
Updating how we convert from eDSL to SeqJSON and vice-versa. There are new changes in the SPEC and this PR is using version 1.0.16 for the JSON SPEC. This is for the point release next week. There are tweaks to repeat arguments in 1.0.17 of the SPEC which @cohansen will be taking care of by bringing back arguments by position. 

## Verification
Ran the e2e test and all checks out. 

## Documentation
Probably some documentation needs to be updated. 

## Future work
Upgrading to SPEC 1.0.17


sequencing-server/package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sequencing-server",
   "version": "1.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
camargo: @goetzrrGit What version of Node/NPM are you using? The lock file should not have this many updates with the change you made to package.json. Make sure you are using node LTS (which comes with NPM 8.x). The lockfile version should remain on `2`.

camargo:
Are there tickets associated with this PR or should we add this to the Project?

sequencing-server/package-lock.json
goetzrrGit: This was generated by` gradlew assemble `which is node 18.13 using npm that is bundled in 18.13

goetzrrGit:
@camargo  I accidentally hit `create PR` before updating the description with the ticket number. It should not link to an issue

cohansen (changes requested):
Can you go through and comment some of the newly added code in the preface just to explain what it does?

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -388,16 +388,120 @@ export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}
...
+            .map(arg => {
+              return this.convertArgsToInterfaces(arg);
+            })
+            .flat(),
+          type: 'repeat',
+          name: key,
+        };
+        result.push(repeatArg);
+      } else {
+        if (typeof value === 'string') {
```
cohansen: Would this be better suited as a switch? Might be more clear.

goetzrrGit:
@cohansen Sure I can do that

camargo:
What would the [upgrade guide](https://nasa-ammos.github.io/aerie-docs/upgrade-guides/1-0-2-to-1-1-0/) look like for this?

goetzrrGit:
@camargo you can add this to the upgrade guide.

The exported seqJSON now matches SPEC version 1.0.16, and a bug has been fixed in AERIE Sequence Editor that previously prevented users from transitioning from SeqJson to eDSL

cohansen (approved):
Looks good, thanks for adding those comments.

PR #688, opened 2023-02-21
Validate seq json file upload

### LINKED ISSUE #552: Validate seq JSON against the seq JSON schema for sequence editor
Currently the user sequence editor allows uploading invalid seq JSON which fails the user sequence generation. We should use the seq JSON schema to validate the incoming seq JSON and notify the user of validation errors if they occur.

cohansen:
* **Tickets addressed:** Closes #552.
* **Review:** By commit

## Description
This runs a validate on the user uploaded seqjson to make sure it matches the current schema we're using. I'm not super happy with the unit test I added, so if someone has a better idea for expecting the API to throw an error I'm all ears.

## Verification
Did manual testing and also added a unit test to include some invalid seqjson.

## Documentation
NA

## Future work
I commented out some currently broken seqjson upload tests that need to be reworked once we fix the import format for seqjson.

### DIFF
--- a/sequencing-server/src/app.ts
+++ b/sequencing-server/src/app.ts
@@ -1,4 +1,6 @@
 import * as ampcs from '@nasa-jpl/aerie-ampcs';
+import Ajv from 'ajv/dist/2020.js';
+import schema from '@nasa-jpl/seq-json-schema/schema.json' assert { type: 'json' };
 import type { CacheItem, UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
 import { Result } from '@nasa-jpl/aerie-ts-user-code-runner/build/utils/monads.js';
 import bodyParser from 'body-parser';
@@ -911,16 +913,36 @@ app.post('/bulk-get-seqjson-for-seqid-and-simulation-dataset', async (req, res, 
  */
 app.post('/get-edsl-for-seqjson', async (req, res, next) => {
   const seqJson = req.body.input.seqJson as SeqJson;
+  const validate = new Ajv({ strict: false }).compile(schema);
+
+  if (!validate(Sequence.fromSeqJson(seqJson))) {
+    throw new Error(
+      `POST /bulk-get-edsl-for-seqjson: Uploaded sequence JSON is invalid and does not match the current spec`,
+    );
+  }
 
   res.json(Sequence.fromSeqJson(seqJson).toEDSLString());
+
   return next();
 });
 
 // Generate Sequence EDSL from many sequence JSONs
 app.post('/bulk-get-edsl-for-seqjson', async (req, res, next) => {
   const seqJsons = req.body.input.seqJsons as SeqJson[];
+  const validate = new Ajv({ strict: false }).compile(schema);
 
-  res.json(seqJsons.map(seqJson => Sequence.fromSeqJson(seqJson).toEDSLString()));
+  res.json(
+    seqJsons.map(seqJson => {
+      if (!validate(Sequence.fromSeqJson(seqJson))) {
+        throw new Error(
+          `POST /bulk-get-edsl-for-seqjson: Uploaded sequence JSON is invalid and does not match the current spec`,
+        );
+      }
+
+      return Sequence.fromSeqJson(seqJson).toEDSLString();
+    }),
+  );
+
   return next();
 });
 
--- a/sequencing-server/test/seqjson-to-edsl.spec.ts
+++ b/sequencing-server/test/seqjson-to-edsl.spec.ts
@@ -59,41 +59,41 @@ describe('getEdslForSeqJson', () => {
 
   it('should throw an error if the user uploads an invalid seqjson', async () => {
     try {
-      await graphqlClient.request<{
-        getEdslForSeqJson: string;
-      }>(
-        gql`
-          query GetEdslForSeqJson($seqJson: SequenceSeqJsonInput!) {
-            getEdslForSeqJson(seqJson: $seqJson)
-          }
-        `,
-        {
-          seqJson: {
-            id: 'test_00001',
-            metadata: {},
-            steps: [
-              {
-                // expansion 1
-                type: 'command',
-                stem: 'BAKE_BREAD',
-                time: { type: TimingTypes.COMMAND_COMPLETE },
-                args: [],
-                metadata: {},
-              },
-              {
-                type: 'command',
-                stem: 'PREHEAT_OVEN',
-                time: { type: TimingTypes.ABSOLUTE, tag: '2020-060T03:45:19.000Z' },
-                args: [100],
-                metadata: {},
-              },
-            ],
+      expect(
+        await graphqlClient.request<{
+          getEdslForSeqJson: string;
+        }>(
+          gql`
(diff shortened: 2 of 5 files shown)


sequencing-server/test/seqjson-to-edsl.spec.ts
```diff
@@ -54,8 +55,49 @@ describe('getEdslForSeqJson', () => {
   });`,
     );
   });
+    */
+
+  it('should throw an error if the user uploads an invalid seqjson', async () => {
+    try {
+      await graphqlClient.request<{
```
dyst5422: [Try expect().rejects](https://stackoverflow.com/questions/47144187/can-you-write-async-tests-that-expect-tothrow)

sequencing-server/src/app.ts
```diff
@@ -902,16 +904,36 @@ app.post('/bulk-get-seqjson-for-seqid-and-simulation-dataset', async (req, res,
...
+
+  res.json(
+    seqJsons.map(seqJson => {
+      if (!validate(Sequence.fromSeqJson(seqJson))) {
+        throw new Error(
+          `POST /bulk-get-edsl-for-seqjson: Uploaded sequence JSON is invalid and does not match the current spec`,
+        );
+      }
+
+      return Sequence.fromSeqJson(seqJson).toEDSLString();
```
dyst5422, line 942: Thoughts on having the validation done inside Sequence.fromSeqJson so we are sure to validate any usage? Are we using that in cases where we might want best-effort conversion and not validate?

cohansen: I thought about putting it inside `fromSeqJson` but I was worried about introducing `Ajv` and the schema as dependencies inside the preface. I suppose we could pass those in if we wanted to validate inside that function though.

Any thoughts @goetzrrGit ?

dyst5422: Very good point I hadn't considered.

Since fromSeqJson is a static function, we could split fromSeqJson out into a separate function outside the preface and include the validation in it. I don't love that though as it leads to a bit of an anemic domain model. Passing an optional validator a la DI to fromSeqJson would be another option, but it requires carrying that DI around with us or doing partial application and managing that.

I'm not really in love with any of those other options and don't think they have material benefits over the current implementation. Really interested in what other creative thoughts others have though.

sequencing-server/test/seqjson-to-edsl.spec.ts
```diff
@@ -54,8 +55,49 @@ describe('getEdslForSeqJson', () => {
   });`,
     );
   });
+    */
+
+  it('should throw an error if the user uploads an invalid seqjson', async () => {
+    try {
```
dyst5422, line 61: You shouldn't need the try-catch anymore, and I suspect you may need to await the expect() call to ensure the test runner waits for the expectation to resolve.

100% spitballing based on experience and not running these tests though, so if you're seeing otherwise, that's great!

cohansen: Yeah, I tried removing the try-catch but it still threw the error, so I just left it to keep the testing logs clean!

dyst5422: I suspect that means the test is not setup correctly. You're awaiting inside the expect, which would end up throwing out of the expect call and so you aren't actually making any assertions in the test. What you want is to pass the Promise to expect rather than the result of that promise.

Try this
```typescript
it('should throw an error if the user uploads an invalid seqjson', async () => {
 await expect(
   graphqlClient.request<{
     getEdslForSeqJson: string;
   }>(
     gql`
       query GetEdslForSeqJson($seqJson: SequenceSeqJsonInput!) {
         getEdslForSeqJson(seqJson: $seqJson)
       }
     `,
     {
       seqJson: {
         id: 'test_00001',
         metadata: {},
         steps: [
           {
             // expansion 1
             type: 'command',
             stem: 'BAKE_BREAD',
             time: { type: TimingTypes.COMMAND_COMPLETE },
             args: [],
             metadata: {},
           },
           {
             type: 'command',
             stem: 'PREHEAT_OVEN',
             time: { type: TimingTypes.ABSOLUTE, tag: '2020-060T03:45:19.000Z' },
             args: [100],
             metadata: {},
           },
         ],
       },
     },
   ),
 ).rejects.toThrow();
});
```

sequencing-server/tsconfig.json
```diff
@@ -1,7 +1,14 @@
...
-  "exclude":["sequencing_file_store", "build","cdict"],
+  "exclude": [
+    "sequencing_file_store",
+    "build",
+    "cdict"
+  ],
   "compilerOptions": {
-    "lib": ["es2022"]
+    "lib": [
+      "esnext"
```
goetzrrGit, line 10: Are there any potential problems with using `esnext`? This version is essentially the next iteration before it becomes finalized as es2023.

Have you considered importing the schema using the following method instead?

```
"compilerOptions": {
    "resolveJsonModule" : true,
    "lib": ["es2022"]
  }
```

goetzrrGit (approved):
Although it is not possible to include the validation in the preface, we can still perform the check during the upload process on the server. By doing so, we can reject any invalid inputs before calling fromSeqJson.

sequencing-server/tsconfig.json
cohansen: I haven't seen any, yeah I originally tried with `es2022` but it didn't work for the schema import.

goetzrrGit (changes requested):
Sorry meant to click `Request Changes' not `Approve`

sequencing-server/tsconfig.json
goetzrrGit: Even with`"resolveJsonModule" : true`

dyst5422: That's surpising to me because I have definitely done the json module resolution with versions far prior to es2022.

camargo: Does it work without the import assertion (`assert { type: 'json' }`)?

cohansen: Yeah, I added `"resolveJsonModule": true` first before trying `esnext`.
`src/app.ts(3,60): error TS2821: Import assertions are only supported when the '--module' option is set to 'esnext' or 'nodenext'.`

No @camargo , it errors without the assertion.

dyst5422: Import assertions are notoriously poorly/weirdly supported.

camargo: Have you tried removing the import assertion and using `"esModuleInterop": true` as well?

camargo: Yeah I've def done json imports on past ES versions as well. Not sure why this is not working for `es2022`.

goetzrrGit: we can always use // @ts-ignore :P

cohansen: @camargo Just tried and it doesn't work either 😅

goetzrrGit: Although this seems to be the only viable option currently available, we can go ahead and approve it. Let's stay alert for the release of es2023 so we can switch to that version, should it become available.

goetzrrGit (approved):

PR #695, opened 2023-02-28
Activity Presets DB

### LINKED ISSUE #677: Create db tables for activity presets
This includes exposing the tables through our Hasura API, and making sure they are connected by foreign key to activity directives.

Mythicaeda:
* **Tickets addressed:** Closes #677 
* **Review:** By commit  

## Description

Creates the following tables:
- `activity_presets`: Stores an activity preset
- `preset_to_directive`: Association table between Presets and Activity Directives (One to Many)
- `preset_to_snapshot_directive`: Association table between Presets and Snapshot Directives (One to Many)

Creates a Hasura Function `apply_preset_to_activity`, which will check that the preset is for the correct activity type before updating the activity directive and upserting into `preset_to_directive`.

I chose to use association tables instead of adding a column to `activity_directive` because:
- It's a cleaner migration
- Presets are not really part of the definition of an activity. They are optional and ignored in merging, unlike the other metadata columns, which _are_ compared for equality.

## Verification

- Added tests to PlanMerge
- Added unit tests in PresetTests

## Documentation

Opened https://github.com/NASA-AMMOS/aerie-docs/issues/17

### DIFF
--- a/merlin-server/sql/merlin/tables/activity_presets.sql
+++ b/merlin-server/sql/merlin/tables/activity_presets.sql
@@ -1,0 +1,71 @@
+create table activity_presets(
+  id integer generated always as identity primary key,
+  model_id integer not null,
+  name text not null,
+  associated_activity_type text not null,
+  arguments merlin_argument_set not null,
+
+  foreign key (model_id, associated_activity_type)
+    references activity_type
+    on delete cascade,
+  unique (model_id, associated_activity_type, name)
+);
+
+comment on table activity_presets is e''
+  'A set of arguments that can be applied to an activity of a given type.';
+
+comment on column activity_presets.id is e''
+  'The unique identifier for this activity preset';
+comment on column activity_presets.model_id is e''
+  'The model defining this activity preset is associated with.';
+comment on column activity_presets.name is e''
+  'The name of this activity preset, unique for an activity type within a mission model.';
+comment on column activity_presets.associated_activity_type is e''
+  'The activity type with which this activity preset is associated.';
+comment on column activity_presets.arguments is e''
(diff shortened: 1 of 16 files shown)


merlin-server/sql/merlin/tables/activity_presets.sql
```diff
@@ -0,0 +1,71 @@
+create table activity_presets(
+  id integer generated always as identity primary key,
+  model_id integer not null,
+  name text,
```
mattdailis: Just pattern matching - any reason for `name` to be nullable?

Mythicaeda: Not that I can think of. I've added the `not null` constraint.

merlin-server/sql/merlin/tables/activity_presets.sql
```diff
@@ -0,0 +1,71 @@
...
+    on delete cascade,
+
+  constraint one_preset_per_activity_directive
+    unique (activity_id, plan_id),
+
+  primary key (preset_id, activity_id, plan_id)
+);
+
+comment on table preset_to_directive is e''
+  'Associates activity directiv with activity directives that have been assigned presets.';
```
mattdailis: Comment typo

```suggestion
  'Associates presets with activity directives that have been assigned presets.';
```

mattdailis (approved):
Looks good to me! Just one small typo that came up in my pg_dump diff

merlin-server/sql/merlin/tables/activity_presets.sql
Mythicaeda: Thanks for the catch. Fixed!

PR #696, opened 2023-02-28
Added a fix for when an activity type has the same name as a command

### LINKED ISSUE #635: Expansion Set fails on Activity Type with same name as a command
Aerie fails to create an expansion set which contains a rule for an activity type that has the same name as a command in the command dictionary (I hypothesize; that's the only logical link I've come up with). 

When creating the set, the API returns a TS2352 error (mildly redacted):

```
    Expansion set could not be type checked
    Caused by: [
        TypeError: TS2352 Conversion of type 'CommandStem<[{*command parameters*}], Record<string, any>>' to type '*activity type name*' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
        Type 'CommandStem<[{*command parameters*}], Record<string, any>>' is missing the following properties from type '*activity type name*': type, duration, startOffset, startTime, and 2 more.
    ]
```

cohansen:
* **Tickets addressed:** Closes #635.
* **Review:** By commit

## Description
There was an issue when creating an expansion set that had an activity type with the exact same name as a command in the command dictionary. I created a duplicate `GrowBanana` command so the name would match the activity type in the bananation model and confirmed that creating an expansion set was broken as well as a type error was being thrown in the editor.

To fix this I added a namespace for `ActivityTypes` so there is no longer a conflict.

## Verification
I tested manually in the UI and added a new unit test to double check this behavior against the updated command dictionary.

## Documentation
N/A

## Future work
N/A

### DIFF
--- a/sequencing-server/src/lib/codegen/ActivityTypescriptCodegen.ts
+++ b/sequencing-server/src/lib/codegen/ActivityTypescriptCodegen.ts
@@ -7,7 +7,7 @@ readonly startTime: Temporal.Instant;
 readonly endTime: Temporal.Instant;`;
 
 export function generateTypescriptForGraphQLActivitySchema(activitySchema: GraphQLActivitySchema): string {
-  const activityTypeAlias = `type ActivityType = ${activitySchema.name};`;
+  const activityTypeAlias = `type ActivityType = ActivityTypes.${activitySchema.name};`;
 
   const activityTypeDeclaration = `readonly type: '${activitySchema.name}';`;
 
@@ -27,8 +27,12 @@ export function generateTypescriptForGraphQLActivitySchema(activitySchema: Graph
   const propertyDeclarations = [activityTypeDeclaration, commonProperties, attributesDeclaration].join('\n');
 
   return (
-    globalDeclaration(`${interfaceDeclaration(activitySchema.name, propertyDeclarations)}\n${activityTypeAlias}`) +
-    '\n\nexport {};'
+    globalDeclaration(
+      `namespace ActivityTypes {\n${interfaceDeclaration(
+        activitySchema.name,
+        propertyDeclarations,
+      )}\n}\n${activityTypeAlias}`,
+    ) + '\n\nexport {};'
   );
 }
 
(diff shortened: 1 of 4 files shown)


sequencing-server/src/lib/codegen/ActivityTypescriptCodegen.ts
```diff
@@ -7,7 +7,7 @@ readonly startTime: Temporal.Instant;
 readonly endTime: Temporal.Instant;`;
 
 export function generateTypescriptForGraphQLActivitySchema(activitySchema: GraphQLActivitySchema): string {
-  const activityTypeAlias = `type ActivityType = ${activitySchema.name};`;
+  const activityTypeAlias = `type ActivityType = ActivityTypes.${activitySchema.name};`;
```
camargo, line 10: Does this imply the UI needs to change the default expansion starter code? 

https://github.com/NASA-AMMOS/aerie-ui/blob/c69d54048e0469bc233efb4713cc713a74c19da8/src/components/expansion/ExpansionRuleForm.svelte#L24

Or will the mapping still be good on the front end?

cohansen: Nope! The namespace gets picked up already:
Screenshot 2023-02-28 at 10 31 51 AM

camargo: Very nice

dyst5422: I don't think so. This type alias should be ok now that we've namespaced the actual activity name that customers can modify. Now if they make a command named `ActivityType`....

cohansen: > I don't think so. This type alias should be ok now that we've namespaced the actual activity name that customers can modify. Now if they make a command named `ActivityType`....

Yeah, that would cause a big problem! I think we can cross that bridge if it ever becomes an issue 😭

dyst5422: I think we should still encourage CamelCase activity types and CAPITALIZED command stems for clarity of readability as well.

dyst5422 (approved):

camargo (approved):
LGTM

mattdailis (approved):
🎉

PR #698, opened 2023-03-01
Add PGCMP Workflow

### LINKED ISSUE #567: Use pgcmp to compare migrations
In the interest of doing a more nuanced diff than just "pg_dump -> string comparison", we are looking into using [pgcmp](https://github.com/cbbrowne/pgcmp).

Current use case: Make sure that the schema that results from running all the migrations matches the latest schema.

Future use case: Provide the users with a way to detect database drift.

Source discussion: #427 

Two possible approaches are:
1. As of #427, the "pg_dump -> string comparison" test is in the `db-tests` suite - the pgcmp test could replace it
2. Alternatively, the pgcmp test could be split into a separate github workflow

Mythicaeda:
* **Tickets addressed:** Closes #567
* **Review:** By commit  

## Description

Added a workflow to:
1) Build the version of the DB from 1.0.1 (the base version for migrations), migrate it to current, then dump it
2) Build the version of the DB on the current branch, then dump it
3) Use `pgcmp` to compare the dumps for equality
4) Upload the results of `pgcmp` should at least one comparison fail.

## Verification

Successful running on this branch should indicate if it's working.

## Future work

Potentially refine the script to only look at the changes between develop and the migration defined in the PR.

### DIFF
--- a/.github/workflows/pgcmp.yml
+++ b/.github/workflows/pgcmp.yml
@@ -1,0 +1,334 @@
+name: "Database Migrations Check PR"
+
+on:
+  pull_request:
+    paths:
+      - 'deployment/hasura/migrations/**'
+      - 'merlin-server/sql/**'
+      - 'sequencing-server/sql/**'
+      - 'scheduler-server/sql/**'
+      - 'deployment/postgres-init-db/sql/**'
+  push:
+    paths:
+      - 'deployment/hasura/migrations/**'
+      - 'merlin-server/sql/**'
+      - 'sequencing-server/sql/**'
+      - 'scheduler-server/sql/**'
+      - 'deployment/postgres-init-db/sql/**'
+    branches:
+      - develop
+    tags:
+      - v*
+  workflow_dispatch:
+
+jobs:
+  dump-db-base:
+    runs-on: ubuntu-latest
+    environment: e2e-test
+    steps:
+      - name: Checkout v1.0.1
+        uses: actions/checkout@v3
+        with:
+          ref: 'v1.0.1'
+      - name: Clone PGCMP
+        uses: actions/checkout@v3
+        with:
+          repository: cbbrowne/pgcmp
+          path: pgcmp
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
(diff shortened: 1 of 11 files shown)


.github/workflows/pgcmp.yml
```diff
@@ -0,0 +1,176 @@
...
+
+on:
+  pull_request:
+    paths:
+      - 'deployment/hasura/migrations/**'
+      - 'merlin-server/sql/**'
+      - 'sequencing-server/sql/**'
+      - 'scheduler-server/sql/**'
+      - 'deployment/postgres-init-db/sql/**'
+      - '.github/**' # REMOVE BEFORE MERGING, BUT THIS SHOULD FORCE THE WORKFLOW TO RUN ON THIS PR
```
cohansen: Just tagging this so it's not forgotten :)

cohansen (changes requested):
Looks fine to me, I ran locally and Aerie ran successfully.

It looks like there's some failed checks though, and 1 remove before merge.

Mythicaeda:
> Looks fine to me, I ran locally and Aerie ran successfully.
You ran the workflow locally or you deployed Aerie from this branch successfully? Double-checking because this PR is adding a workflow to Aerie, not modifying any existing Aerie code

cohansen:
> > Looks fine to me, I ran locally and Aerie ran successfully.
> > You ran the workflow locally or you deployed Aerie from this branch successfully? Double-checking because this PR is adding a workflow to Aerie, not modifying any existing Aerie code

Ah, I misunderstood your instructions. I ran Aerie locally which I guess doesn't test this workflow. My comment above was because I saw some of the checks had failed against this branch.

Mythicaeda:
> Ah, I misunderstood your instructions. I ran Aerie locally which I guess doesn't test this workflow. My comment above was because I saw some of the checks had failed against this branch.

Yeah, that failing check is actually the one being added

.github/workflows/pgcmp.yml
Mythicaeda: It's been removed!

cohansen (approved):
Looks great and the checks passed! 🎉

PR #700, opened 2023-03-01
Added Ground_Block and Ground_Event to eDSL

goetzrrGit:
* **Review:** By commit  

## Description
Added `Ground_Block` and `Ground_Event` types to the `Sequence.step[]` object. You should be able to go back and forth between eDSL and SeqJSON also within the Sequence Editor.  

## Verification
 Ran the e2e test and manually tested 

## Documentation

```ts
C.GROUND_BLOCK("Block 1")
A`2050-45T00:12:00`.GROUND_BLOCK("Block 2")
E`10:00:00`.GROUND_BLOCK("Block 3")
R`04:39:23.000`.GROUND_BLOCK("Block 4")

C.GROUND_EVENT("EVENT 1")
A`2050-45T00:12:00`.GROUND_EVENT("EVENT 2")
E`10:00:00`.GROUND_EVENT("EVENT 3")
R`04:39:23.000`.GROUND_EVENT("EVENT 4")
```

You can also chain these objects' parameters

```ts
C.GROUND_BLOCK("").METADATA({}).ARGUMENTS([]).MODELS([]).DESCRIPTION("")
```

Here is an example of the SeqJSON generated

eDSL
```ts
R`02:02:00`
.GROUND_BLOCK('GroundBlock2')
.ARGUMENTS([{ name: 'intStartTime', type: 'number', value: 10 }])
.METADATA({ author: 'Ryan' })
.DESCRIPTION('Set the ground block time')
.MODELS([
	{
		offset: '00:10:00.001',
		value: true,
		variable: 'model_var_boolean',
	},
]);
```

```json
{
  "id": "test0000",
  "metadata": {
    "author": "XXXX",
    "name": "test sequence",
    "onboard_path": "/tmp/onboard/eis/"
  },
  "steps": [
    {
      "name": "GroundBlock2",
      "time": {
        "type": "COMMAND_RELATIVE",
        "tag": "02:02:00.000"
      },
      "args": [
        {
          "name": "intStartTime",
          "type": "number",
          "value": 10
        }
      ],
      "description": "Set the ground block time",
      "metadata": {
        "author": "Ryan"
      },
      "models": [
        {
          "offset": "00:10:00.001",
          "value": true,
          "variable": "model_var_boolean"
        }
      ],
      "type": "ground_block"
    }
  ]
}
```

## Future work
Add the last 2 step objects (Activate and Load)

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
@@ -6,6 +6,8 @@ import {
   HMS_STRING,
   Sequence,
   TimingTypes,
+  Ground_Event,
+  Ground_Block,
 } from './CommandEDSLPreface';
 
 describe('Command', () => {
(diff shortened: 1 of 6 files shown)


camargo:
So what happens if there is a command in the command dictionary called `GROUND_EVENT` or `GROUND_BLOCK`?

dyst5422:
> So what happens if there is a command in the command dictionary called `GROUND_EVENT` or `GROUND_BLOCK`?

We could de-facto namespace them with a special leading character like `$` (`$GROUND_EVENT`, `$GROUND_BLOCK`), but I'm also not a huge fan of that pattern either.

goetzrrGit:
Currently, I don't have a solution to the namespace collision issue we're facing. The proposed name for GroundBlock and GroundEvent was suggested by Basak, so I hope they won't use the same name for another FSW_Command. Although all FSW_Commands follow the SNAKE_UPPER_CASE pattern, changing these Ground steps to Snake_Case is a possibility, but not ideal as we were progressing towards a style of all SNAKE_UPPER_CASE for the eDSL.

One way to avoid namespace collisions is to use a unique prefix for all identifiers in the code. For example, you could prefix all FSW_Commands with "FSW_".  This will match HDW Commands in the Command dictionary "HDW_". This would help to differentiate them from any other namespaces or interfaces that may be used in the project.

Furthermore, we need to be careful of potential namespace collisions with the JSON Spec, as it has already claimed CamelCase for all its interfaces. So if a mission command collides with that then we have the same problem as above.

goetzrrGit:
We can probably encapsulate the JSON Spec into a namespace SPEC. To avoid collision with the JSON Spec.

`class Sequence implements SPEC.SeqJson`

I am trying to think how we can do this with Command namespace collisions. Maybe something like this as we are starting to chain things together now:

```
C.FSW.BAKE_BANANA()
C.GROUND_BLOCK() 
```

cohansen (approved):
This PR looks good to me, I'm also concerned with the namespacing issues especially after the one Clipper found, but it sounds like that needs a broader discussion and a separate ticket to address.

dyst5422:
@goetzrrGit can I recommend that we have the syntax for these special ones be Capitalized (GroundBlock, GroundEvent) such that they are immediately distinguishable from the UPPERCASED command stems as a first level of disambiguation? I worry about releasing a feature and then changing the API on it later.

PR #708, opened 2023-03-03
Make simulation optional for scheduling goals

### LINKED ISSUE #674: Make simulation optional for scheduling goals
To help with performance we need a way to make simulation optional for scheduling goals

JoelCourtney:
* **Tickets addressed:** closes #674 
* **Review:** By commit  

## Description

This makes re-simulation between goals optional. Each goal has two opportunities to simulate:
- before it evaluates, to make sure it has up-to-date results
- after it evaluates, to make sure the durations of the inserted activities are as expected

When there are multiple goals, some of the simulation is already avoided:
- Sim before goal 1
- Eval goal 1
- Sim after goal 1
- Sim before goal 2 (*no-op*, results are up-to-date)
- Eval goal 2
- Sim after goal 2

However, if "Sim after goal 1" is turned off, at "Sim before goal 2" it will recognize that results are out of date and resimulate, doing the same amount of work as before. This means that both "Sim after goal 1" and "Sim before goal 2" have to be turned off.

We already have a flag for turning off sim after a goal: `checkSimBeforeInsertingActivities`. I've added a second flag, `checkSimBeforeEvaluatingGoal` which does what it says it does. If goal 1 is set to not re-simulate after, then `checkSimBeforeInsertingActivites` will be `false` during goal 1, and `checkSimBeforeEvaluatingGoal` will be `false` during goal 2.

This PR also adds a column to the scheduling spec goals table, with a migration, and fixes a couple bugs I found on the way.

## Verification
I've added an integration test, which also doubles as a demonstration of how this feature turns off the correctness guarantee.

## Documentation
- [ ] This one's definitely worthy of an update to the docs.

## Future work
UI

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
@@ -174,7 +174,7 @@ public class Spans implements IntervalContainer<Spans>, Iterable<Segment<Optiona
 
   public Spans intersectWith(final Windows windows){
     final var ret = new Spans();
-    this.intervals.forEach(x -> windows.iterator().forEachRemaining(y -> ret.add(Interval.intersect(y.interval(), x.interval()), x.value())));
+    this.intervals.forEach(x -> windows.iterateEqualTo(true).iterator().forEachRemaining(y -> ret.add(Interval.intersect(x.interval(), y), x.value())));
     return ret;
   }
 
--- a/deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/down.sql
+++ b/deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/down.sql
@@ -1,0 +1,4 @@
+comment on column scheduling_specification_goals.simulate_after is null;
+alter table scheduling_specification_goals drop column simulate_after;
+
+call migrations.mark_migration_rolled_back('1');
--- a/deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/up.sql
+++ b/deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/up.sql
@@ -1,0 +1,6 @@
+alter table scheduling_specification_goals add column simulate_after boolean not null default true;
+
+comment on column scheduling_specification_goals.simulate_after is e''
+  'Whether to re-simulate after evaluating this goal and before the next goal.';
+
+call migrations.mark_migration_applied('1');
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
@@ -32,12 +32,16 @@ public class Goal {
    * state constraints applying to the goal
    */
   protected Expression<Windows> resourceConstraints;
-  
+
   /**
-   * Whether to resimulate after this goal or use stale results
+   * Whether to resimulate after this goal or make the next goal use stale results.
+   *
+   * True is the default behavior. If False, the durations of the activities inserted by
+   * *this* goal will not be simulated and checked, and the *next* goal will not have up-to-date
+   * sim results if this goal placed any activities.
    */
   public boolean simulateAfter;
-  
+
   /** Set to true if partial satisfaction is ok, the scheduler will try to do its best */
   private boolean shouldRollbackIfUnsatisfied = false;
 
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
@@ -32,6 +32,12 @@ public class Goal {
    * state constraints applying to the goal
    */
   protected Expression<Windows> resourceConstraints;
+  
+  /**
+   * Whether to resimulate after this goal or use stale results
+   */
+  public boolean simulateAfter;
+  
   /** Set to true if partial satisfaction is ok, the scheduler will try to do its best */
   private boolean shouldRollbackIfUnsatisfied = false;
 
@@ -148,6 +154,13 @@ public class Goal {
 
     boolean shouldRollbackIfUnsatisfied;
 
+    public T simulateAfter(boolean simAfter) {
+      this.simulateAfter = simAfter;
+      return getThis();
+    }
+
+    boolean simulateAfter = true;
+
     /**
      * uses all pending specifications to construct a matching new goal object
      *
@@ -205,6 +218,8 @@ public class Goal {
         final var windows = new Windows(false).set(Interval.between(starting, ending), true);
         goal.temporalContext = new WindowsWrapperExpression(windows);
       }
+
+      goal.simulateAfter = this.simulateAfter;
 
       return goal;
     }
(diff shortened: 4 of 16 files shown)


deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/down.sql
```diff
@@ -0,0 +1,5 @@
+alter table scheduling_specification_goals drop column simulate_after;
+
+comment on column scheduling_specification_goals.simulate_after is null;
```
Mythicaeda, line 1: This comment needs to be set to null before the column is dropped, otherwise an error is thrown when running the down.sql

deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/up.sql
```diff
@@ -0,0 +1,6 @@
+alter table scheduling_specification_goals add column simulate_after boolean not null default true;
+
+comment on column scheduling_specification_goals.simulate_after is e''
+  'Whether to re-simulate after evaluating this goal and before the next goal.';
+
+call migrations.mark_migration_applied('1');
```
Mythicaeda, line 6: This call also needs to happen at the bottom of `scheduler-server/sql/scheduler/applied_migrations.sql`.

JoelCourtney: done

deployment/hasura/migrations/AerieScheduler/1_optional_sim_per_goal/down.sql
JoelCourtney: done

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -861,7 +861,7 @@ public void testCoexistenceWindowsCutoffMidActivity() {
...
     partialPlan.add(SchedulingActivityDirective.of(actTypeA, planningHorizon.getStartAerie().plus(Duration.of(14, Duration.SECONDS)), Duration.of(4, Duration.SECONDS), null, true)); //create an activity that's 5 seconds long, 16s after start
     partialPlan.add(SchedulingActivityDirective.of(actTypeA, planningHorizon.getStartAerie().plus(Duration.of(19, Duration.SECONDS)), Duration.of(4, Duration.SECONDS), null, true)); //create an activity that's 5 seconds long, 16s after start
-    partialPlan.add(SchedulingActivityDirective.of(actTypeA, planningHorizon.getStartAerie().plus(Duration.of(25, Duration.SECONDS)), Duration.of(4, Duration.SECONDS), null, true)); //create an activity that's 5 seconds long, 16s after start
+    partialPlan.add(SchedulingActivityDirective.of(actTypeA, planningHorizon.getStartAerie().plus(Duration.of(25, Duration.SECONDS)), Duration.of(2, Duration.SECONDS), null, true)); //create an activity that's 5 seconds long, 16s after start
```
Mythicaeda: Very minor, but could you fix the comment here? I don't think this activity is 5s long and 16s after plan start.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Goal.java
```diff
@@ -32,6 +32,12 @@ public class Goal {
    * state constraints applying to the goal
    */
   protected Expression<Windows> resourceConstraints;
+  
+  /**
+   * Whether to resimulate after this goal or use stale results
```
Mythicaeda: Minor, but this comment feels slightly inaccurate, since it's not whether _this_ goal uses stale results, but whether the _next_ goal uses stale results.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -174,7 +174,7 @@ public boolean isCollectionSubsetOf(final Spans otherSpans){
 
   public Spans intersectWith(final Windows windows){
     final var ret = new Spans();
-    this.intervals.forEach(x -> windows.iterator().forEachRemaining(y -> ret.add(Interval.intersect(y.interval(), x.interval()), x.value())));
+    this.intervals.forEach(x -> windows.iterateEqualTo(true).iterator().forEachRemaining(y -> ret.add(Interval.intersect(x.interval(), y), x.value())));
```
Mythicaeda, line 177: I'd like to see a quick test of this method that would fail with the previous implementation but succeed with this fix, both because it's weird that all of our tests used to pass and to ensure we don't add back the same bug in the future.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -1682,6 +1686,59 @@ export default () => Goal.CoexistenceGoal({
     assertEquals(Duration.of(65, Duration.MINUTES), peelBanana.startOffset());
   }
 
+  @Test
+  void testOptionalSimulationAfterGoal() {
```
Mythicaeda: How does this prove that there was no simulation between the goals? Wouldn't a more accurate test be one where, by not simulating between the goals, we get a different result set than what we'd get if we _had_ simulated between each goal?

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
JoelCourtney: 👍

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
JoelCourtney: That's exactly what it does. I can add a comment explaining how though.

JoelCourtney: (There was a comment already there, but that was a mistake, I must have copy-pasted from somewhere else)

JoelCourtney: This test makes sure that resources are stale. I've added a second test that makes sure it doesn't care about directives that don't have corresponding spans. I've also modified both of them to make sure the results are different, compared to if it *is* resimulated.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
JoelCourtney: done

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -1682,6 +1686,132 @@ export default () => Goal.CoexistenceGoal({
...
+   * If the resource is resimulated, it will be different and Goal 1 will place one activity.
+   *
+   * Both options are tested here.
+   */
+  @Test
+  void testOptionalSimulationAfterGoal_staleResources() {
+    final var activityDuration = Duration.of(1, Duration.HOUR);
+    final var configs = Map.of(
+        false, 0, // don't simulate, expect 0 activities
+        true, 1 // do simualte, expect 1 activity
```
Mythicaeda: Typo
```suggestion
        true, 1 // do simulate, expect 1 activity
```

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -1682,6 +1686,132 @@ export default () => Goal.CoexistenceGoal({
...
+   * If it does resimulate, it should place one activity.
+   *
+   * Both options are tested here.
+   */
+  @Test
+  void testOptionalSimulationAfterGoal_unsimulatedActivities() {
+    final var activityDuration = Duration.of(1, Duration.HOUR);
+    final var configs = Map.of(
+        false, 0, // don't simulate, expect 0 activities
+        true, 1 // do simualte, expect 1 activity
```
Mythicaeda: Typo:
```suggestion
        true, 1 // do simulate, expect 1 activity
```

Mythicaeda (approved):
Looks great! Thanks for adding in those tests.

PR #709, opened 2023-03-03
Use correct join type when querying scheduling conditions

### LINKED ISSUE #679: Fix global scheduling conditions query in scheduler worker
To reproduce the issue, bring up aerie with two plans, and add a global scheduling condition to each one.

Add a scheduling goal to one of the plans, and attempt to run scheduling on that plan.

image

mattdailis:
* **Tickets addressed:** Fixes #679 
* **Review:** By commit  

## Description

`GetSpecificationConditionsAction` needs to get all of the conditions associated with a particular specification. It did this via a left join, which would return a `null` value for each scheduling condition in the database that isn't in the plan of interest. The intention of this query is to get the scheduling conditions for the given specification - other specifications are irrelevant - so an inner join is the appropriate choice here.

This PR replaces that left join with an inner join.

## Verification

To reproduce the issue, I brought up aerie with two plans, and added a global scheduling condition to each one.

I then added a scheduling goal to one of the plans, and attempted to run scheduling on that plan. This triggered the null pointer exception.

After making the fix, I rebuilt one of the scheduler workers, and left the other using the old code. By bringing down one of those workers at a time, I demonstrated that the old code exhibited this issue, while the new code ran with no errors.

## Documentation

No documentation updates, given that this is an unexpected bug.

## Future work

No future work - but mental note to choose my joins carefully

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationConditionsAction.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/GetSpecificationConditionsAction.java
@@ -17,7 +17,7 @@ import java.util.List;
       c.definition,
       c.revision
     from scheduling_specification_conditions as s
-      left join scheduling_condition as c
+      join scheduling_condition as c
       on s.specification_id = ?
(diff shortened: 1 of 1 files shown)


camargo (approved):
Tested locally. LGTM!

Issue #514, opened 2022-12-05
Workflow for Constraints checking request with no simulation data


mattdailis:
Transferred from JIRA: https://jira.jpl.nasa.gov/browse/AERIE-2204
AERIEQS: https://jira.jpl.nasa.gov/browse/AERIEQS-354

When a user makes a change to a plan and clicks "Check Constraints", without running simulation, they will get a silent "constraint checking succeeded" status with no violations, even if there should have been violations.

We should think through this workflow - first off, the constraints should fail when the simulation data is unavailable (or... it could leverage some of our new three-valued-logic operations...?). Secondly, what should the UI do when a user clicks check constraints on a stale simulation? Maybe a popup, "Hey, your simulation results are stale, want to run a simulation?" with a cancel button and a simulate button in the popup.

PR #710, opened 2023-03-03
Constraints: Load profiles by name

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Currently, when we check constraints, we load the entire SimulationResults from the database before we check any constraints. This draft PR proposes a way to load only the profiles that are referenced by the constraints being checked. In a scenario where a mission model produces very many profiles, but a user wants to check a specific constraint on just a few of them, this should make it possible to check that constraint more quickly.

The approach taken is to start out with empty `realProfiles` and `discreteProfiles`, and as we iterate through constraints, load the resource profiles referenced by that constraint from the database.

As a reviewer, I would start with the changes in `GetSimulationResultsAction::getViolations`.

This PR does not attempt to do a similar thing with external profiles, though that would be a logical extension of this idea.

This PR preserves the behavior that running constraints without a corresponding up-to-date simulation dataset results in a successful run with no violations. I had considered trying to "fix" that behavior in this PR, but then I realized that there may still be constraints that can be checked on external profiles, so simply throwing an error wouldn't be appropriate. I'll leave that work for when we get around to #514.

## Verification

I brought up Aerie, and used the banananation mission model to create a plan with BiteBananas and GrowBananas.

I used the following constraints for testing:

### Constraint that references zero resources
```typescript
  return Constraint.ForbiddenActivityOverlap(ActivityType.BiteBanana, ActivityType.GrowBanana);
}
```

### Constraint that references two resources
```typescript
  return Real.Resource("/fruit").lessThan(Real.Resource("/peel"));
}
```

### Constraint that references one of the same, and one different resources as the previous
```typescript
  return Real.Resource("/fruit").lessThan(5).if(Discrete.Resource("/producer").equal("Dole"));
}

```

I set breakpoints in `GetSimulationResultsAction::getViolations` and verified that for the first constraint, no resource profiles were loaded; for the second constraint, both resource profiles were loaded; and for the third constraint, only `/producer` was loaded, since `/fruit` was already available in memory.

I also verified that running constraints without an up-to-date simulation dataset results in a successful run with no violations.

## Documentation

There shouldn't be any observable differences in the system, aside from hopefully a slight improvement in constraints checking response time for certain workloads.

## Future work

Come up with a reasonable approach to #514 - the current solution of "just show that all the constraints are satisfied" doesn't quite cut it. (Perhaps, in the absence of simulation datasets, we can provide profiles that contain nothing but gaps to the constraints, and leverage their gap handling logic to report that the satisfaction is unknown?)


JoelCourtney:
__*oh my gawd we finally have a use for the extract resources method*__

Every time I add a new node I say "why do I have to maintain this method, we never use it, can't I just delete it?" But now the cumulative 5 minutes I've spent over the last year implementing this method have finally paid off!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -449,8 +408,77 @@ public PostgresResultsCell(
...
+                case INCOMPLETE -> new State.Incomplete(record.simulationDatasetId());
+                case FAILED -> new State.Failed(record.simulationDatasetId(), record.state().reason()
+                                                                                    .orElseThrow(() -> new Error(
+                                                                                        "Unexpected state: %s request state has no failure message".formatted(
+                                                                                            record.state().status()))));
+                case SUCCESS -> new State.Success(
+                    record.simulationDatasetId(),
+                    new SimulationResultsHandle() {
+                      @Override
+                      public SimulationResults getSimulationResults() {
```
skovati: Should this inline class instantiation be extracted to a `PostgresSimulationResultsHandle`?

skovati (commented):
LGTM! I'll have to merge these changes with #931 but otherwise this looks good to go.

mattdailis:
Looks like we've got a flaky test situation - I re-triggered the `e2e-test` run, and looks like you've got the green light now.

I "approve" :+1:

JoelCourtney:
Do we have any constraints integration tests? I'm not seeing any that would fail if this was broken. The code itself looks excellent, but this is the first real change to the evaluation process while I've been on aerie, and it looks like we never made any integration tests for the constraints action as a whole.

skovati:
Good call, I'll work on an e2e test for constraints 👍

skovati:
Hey @JoelCourtney, I added some (very) basic integration tests for adding and checking constraints. Let me know if you have other ideas for more rigorous e2e test we can add.

e2e-tests/src/tests/constraints.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  test("Check violation starts and ends as expected", async () => {
+    const plan_start_unix = time.getUnixEpochTime(plan_start_timestamp);
+    const plan_end_unix = time.getUnixEpochTime(plan_end_timestamp);
+    const plan_duration_micro = (plan_end_unix - plan_start_unix) * 1000;
+    const activity_offset_micro = activity_offset_hours * 60 * 60 * 1000 * 1000;
+
+    expect(violation.windows[0].start).toEqual(activity_offset_micro);
+    expect(violation.windows[0].end).toEqual(plan_duration_micro);
+  });
```
Mythicaeda, line 128: These tests are missing a "Teardown" step to clean the DB of testing artifacts (ie the things inserted during "Initialize")

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/PostgresSimulationResultsHandle.java
Mythicaeda, line 1: The design of this class doesn't sit right with me for a few reasons:
1) It forces formerly `private` methods in `PostgresResultsCellRepository` to now be declared as `public`. Additionally, it forces formerly `package-private` class `ProfileSet` to be declared `public`.
2) It doesn't wrap any new information that wasn't already in `SimulationResults`
3) Introducing "wrapper classes" can make the code harder to understand (ie the `SimulationFacade`, `ResumableSimulationDriver` scenario in the Scheduler)

I'd like clarification as to the reasoning behind this class's design and why its methods couldn't be implemented elsewhere (ie PostgresResultsCellRepository)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfilesByNameAction.java
```diff
@@ -0,0 +1,62 @@
...
+        and p.name = any(?)
+    """;
+
+  private final PreparedStatement statement;
+
+  public GetProfilesByNameAction(final Connection connection) throws SQLException {
+    this.statement = connection.prepareStatement(sql);
+  }
+
+  public List<ProfileRecord> get(final long datasetId, final Iterable<String> names) throws SQLException {
```
Mythicaeda: It's really interesting that the input of this function starts as a `HashSet`, gets "super"ed to an `Iterable`, is used to build an `ArrayList`, which is finally cast to a `String[]`. It makes me wonder if we can't do anything more efficient with these types 🤔.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -448,8 +407,29 @@ public PostgresResultsCell(
...
+        } else {
+          final var record = record$.get();
+          result = Optional.of(
+              switch (record.state().status()) {
+                case PENDING -> new State.Pending(record.simulationDatasetId());
+                case INCOMPLETE -> new State.Incomplete(record.simulationDatasetId());
+                case FAILED -> new State.Failed(record.simulationDatasetId(), record.state().reason()
+                                                                                    .orElseThrow(() -> new Error(
+                                                                                        "Unexpected state: %s request state has no failure message".formatted(
+                                                                                            record.state().status()))));
```
Mythicaeda: Very minor: Why the whitespace change on these three lines?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
```diff
@@ -21,8 +21,8 @@
 import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
 import static gov.nasa.jpl.aerie.merlin.server.http.ProfileParsers.realDynamicsP;
 
-/*package-local*/ final class ProfileRepository {
-  static ProfileSet getProfiles(
+/*package-local*/ public final class ProfileRepository {
```
Mythicaeda: The `package-local` comment represents the privacy of the class. If it is no longer `package-local` (ie how it is now declared `public`), the `package-local` comment needs to go away.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java
```diff
@@ -76,7 +78,7 @@ public Map<String, List<Pair<Duration, SerializedValue>>> getResourceSamples(fin
     final var revisionData = this.planService.getPlanRevisionData(planId);
     final var simulationResults$ = this.simulationService.get(planId, revisionData);
     if (simulationResults$.isEmpty()) return Collections.emptyMap();
-    final var simulationResults = simulationResults$.get();
+    final var simulationResults = simulationResults$.get().getSimulationResults();
```
Mythicaeda: `simulationResults$` needs to be renamed if it is no longer a `SimulationResults`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SimulationService.java
```diff
@@ -1,12 +1,12 @@
...
 import gov.nasa.jpl.aerie.merlin.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.merlin.server.models.SimulationResultsHandle;
+
+import java.util.Optional;
 
 public interface SimulationService {
   ResultsProtocol.State getSimulationResults(PlanId planId, RevisionData revisionData);
-  Optional<SimulationResults> get(PlanId planId, RevisionData revisionData);
+  Optional<SimulationResultsHandle> get(PlanId planId, RevisionData revisionData);
```
Mythicaeda, line 11: Is this type change a breaking change?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
```diff
@@ -18,7 +17,7 @@ record Pending(long simulationDatasetId) implements State {}
     record Incomplete(long simulationDatasetId) implements State {}
 
     /** Simulation complete -- results now available. */
-    record Success(long simulationDatasetId, SimulationResults results) implements State {}
+    record Success(long simulationDatasetId, SimulationResultsHandle results) implements State {}
```
Mythicaeda, line 20: Similar question here: Is this method signature change breaking?

Mythicaeda (changes requested):
Regarding Additional Tests: I'm seeing no tests of the feature introduced in this PR. It may make more sense for them to be unit tests given that it's about what's loaded in the Java environment, but I'd at bare minimum like to see the tests @mattdailis described in the PR's `Verification` part if possible.

e2e-tests/src/tests/constraints.test.ts
skovati: Good call, I've added a "Clean Up" test

Mythicaeda: Less important, but another e2e test that may be useful is that there are no constraint violations after fixing the violation and resimulating.

skovati: I've added this additional test

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/SimulationService.java
skovati: I don't think so, since the two spots this `SimulationResultsHandle` is consumed quickly transform it into `SimululationResults` and return that where expected.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/ResultsProtocol.java
skovati: Similarly, I don't think so. This allows `CachedSimulationService` and `UncachedSimulationService` to implement the `SimulationService` interface changes above, and when `ResultsProtocol.State.Success` is used in `GetSimulationResultsAction`, we simply grab the `simulationDatasetId` from it, and don't directly reference the `results`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/PostgresSimulationResultsHandle.java
skovati: Good points and I agree overall, I don't love the design of this class. I extracted `PostgresSimulationResultsHandle` to it's own class since it was otherwise an inline class instantiation within `PostgresResultsCellRepository`, which felt out of place considering `InMemorySimulationResultsHandle` had its own class directly implementing `SimulationResultsHandle`.

As for the broader point of whether or not this wrapper class should even exist, I'd be happy to explore in-lining these functions to another class. @Mythicaeda do you think `PostgresResultsCellRepository` and `InMemoryResultsCellRepository` make the most sense?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
skovati: No reason, I've reverted the change 👍

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/PostgresSimulationResultsHandle.java
Mythicaeda: As an initial reaction, either there or as an inner class of theirs, yeah.

.gitignore
```diff
@@ -51,6 +51,7 @@ deployment/postgres-init-db/sql/**/*.sql
 
 # Ignore Gradle build output directory
 build
+bin
```
Mythicaeda: Which of our Gradle build processes produces a `bin` directory that isn't inside an already-ignored directory?

skovati: Oops, looks like these folders aren't actually created by Gradle, but instead the [jdtls](https://github.com/eclipse/eclipse.jdt.ls) build process if it can't properly interface with Gradle. If no other developers experience this, I can move this to my global `.gitignore` instead.

Mythicaeda: Yeah, you may wanna double-check in the Slack Channel, but I don't _think_ anyone else is seeing this right now.

skovati:
Unfortunately, the manual validation @mattdailis did is deep in the internals of the `merlin-server` workflow, so automating that into a test would require either exposing a list of accessed profiles as a return value, or creating some intricate mock objects for unit testing. Since the e2e tests capture the inputs and outputs, additional tests that verify the internal workings of said function shouldn't be necessary.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetProfilesByNameAction.java
skovati: Good find, I've updated the chain of `getProfile` calls to take a `List<String>` instead.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/PostgresSimulationResultsHandle.java
skovati: I've converted both `PostgresSimulationResultsHandle` and `InMemorySimulationResultsHandle` to be inner classes, so we retain the wrapper class but keep these functions within the correct scope.

Mythicaeda: Thanks!

Mythicaeda (approved):
LGTM!

PR #723, opened 2023-03-09
Upload simulation results after scheduling

### LINKED ISSUE #676: Look into returning the last simulation results at the end of scheduling
Currently we re-simulate at the end of scheduling. We should look into returning the last simulation the scheduler runs instead.

adrienmaillard:
* **Tickets addressed:** resolves #676 
* **Review:** By commit  

## Description

After scheduling, the simulation results are posted to the merlin db via graphql. The id of the created dataset is returned as part of the schedule action and it is saved in the `request` table of the scheduler database. 

Also, the creation of partitions has been replaced with a trigger after insert on `dataset`, thank you @Mythicaeda. 

## Verification

Verified through local tests.
Right now, the user has to refresh the UI to see the new dataset after scheduling.

## Documentation

None. 

## Future work

Make the UI update the dataset id shown based of the id returned by the `schedule` call.

### DIFF
--- a/e2e-tests/src/tests/scheduler.test.ts
+++ b/e2e-tests/src/tests/scheduler.test.ts
@@ -2,6 +2,11 @@ import { expect, test } from '@playwright/test';
 import req from '../utilities/requests.js';
 import time from '../utilities/time.js';
 
+const eqSet = (xs, ys) =>
+    xs.size === ys.size &&
+    [...xs].every((x) => ys.has(x));
+
+
 test.describe('Scheduling', () => {
   const rd = Math.random() * 100;
   const plan_start_timestamp = "2021-001T00:00:00.000";
@@ -13,6 +18,7 @@ test.describe('Scheduling', () => {
   let second_goal_id: number;
   let plan_revision: number;
   let specification_id: number;
+  let dataset_id:number;
 
   test('Upload jar and create mission model', async ({ request }) => {
     //upload bananation jar
@@ -184,11 +190,12 @@ test.describe('Scheduling', () => {
     let it = 0;
     let reason_local: string;
     while (it++ < max_it && (status == 'pending' || status == 'incomplete')) {
-      const { reason, status, analysisId } = await req.schedule(request, specification_id);
+      const { reason, status, analysisId, datasetId } = await req.schedule(request, specification_id);
       status_local = status;
       reason_local = reason;
       expect(status).not.toBeNull();
       expect(status).toBeDefined();
+      dataset_id = datasetId
       await delay(1000);
     }
     if (status_local == "failed") {
@@ -196,6 +203,87 @@ test.describe('Scheduling', () => {
     }
     expect(status_local).toEqual("complete")
     expect(analysisId_local).toEqual(analysisId)
+    expect(dataset_id).not.toBeNull();
+  });
+
+  test('Verify posting of simulation results', async ({ request }) => {
+    let plan = await req.getPlan(request, plan_id)
+    let directiveIds = new Set<number>()
+    let startOffsetsActivityDirectives = new Set<string>()
+    for(const activity of plan.activity_directives){
+      directiveIds.add(activity['id'])
+    }
+    const activities = await req.getSimulationDatasetByDatasetId(request,dataset_id)
+    expect(activities.simulated_activities.length).toEqual(plan.activity_directives.length)
+    let refDirectiveIds = new Set<number>()
+    for(let simulated_activity of activities.simulated_activities){
+      refDirectiveIds.add(simulated_activity.activity_directive.id)
+      startOffsetsActivityDirectives.add(simulated_activity.start_offset)
+    }
+    //all directive have their simulated activity
+    expect(eqSet(refDirectiveIds,directiveIds))
+    const dataset = await req.getProfiles(request, dataset_id)
+    //expect one profile per resource in the banananation model
+    expect(dataset.length).toEqual(7)
+    for(let resource of dataset){
+      if(resource.name =="/fruit" || resource.name == "/peel"){
+        let startOffsetOfResource = new Set<string>()
+        for(let segment of resource.profile_segments){
+          startOffsetOfResource.add(segment.start_offset)
+        }
+        //these two resources are affected by the peel, grow and bite banana
+        expect(eqSet(startOffsetOfResource, startOffsetsActivityDirectives)).toEqual(true)
+      }
+    }
+    let topics = await req.getTopicsEvents(request, dataset_id)
+    let prefixInput = "ActivityType.Input."
+    let prefixOutput = "ActivityType.Output."
+    let peelInputIsThere = false
+    let biteInputIsThere = false
+    let growInputIsThere = false
+    let peelOutputIsThere = false
+    let biteOutputIsThere = false
+    let growOutputIsThere = false
+    //topic added in the mission model
+    let producerIsThere = false
+    for(let topic of topics){
+      switch (topic.name){
+        case prefixInput+"BiteBanana":
+          biteInputIsThere = true
+          expect(topic.events.length).toEqual(1)
+          break
+        case prefixOutput+"BiteBanana":
+          biteOutputIsThere = true
+          expect(topic.events.length).toEqual(1)
+          break
+        case prefixInput+"GrowBanana":
+          growInputIsThere = true
+          expect(topic.events.length).toEqual(1)
+          break
+        case prefixOutput+"GrowBanana":
+          growOutputIsThere = true
+          expect(topic.events.length).toEqual(1)
+          break
+        case prefixInput+"PeelBanana":
+          peelInputIsThere = true
+          expect(topic.events.length).toEqual(12)
+          break
+        case prefixOutput+"PeelBanana":
+          peelOutputIsThere = true
+          expect(topic.events.length).toEqual(12)
+          break
+        case "/producer":
+          producerIsThere = true
+          expect(topic.events.length).toEqual(0)
+      }
+    }
+    expect(peelInputIsThere)
+    expect(growInputIsThere)
+    expect(biteInputIsThere)
+    expect(peelOutputIsThere)
+    expect(growOutputIsThere)
+    expect(biteOutputIsThere)
+    expect(producerIsThere)
   });
 
   test('Get Plan', async ({ request }) => {
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
@@ -21,6 +21,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
 
@@ -51,6 +52,10 @@ public class SimulationFacade implements AutoCloseable{
     return lastSimConstraintResults;
   }
 
+  public SimulationResults getLatestDriverSimulationResults(){
+    return lastSimDriverResults;
+  }
+
   public SimulationFacade(final PlanningHorizon planningHorizon, final MissionModel<?> missionModel) {
     this.missionModel = missionModel;
     this.planningHorizon = planningHorizon;
@@ -68,6 +73,10 @@ public class SimulationFacade implements AutoCloseable{
   public void setActivityTypes(final Collection<ActivityType> activityTypes){
     this.activityTypes = new HashMap<>();
     activityTypes.forEach(at -> this.activityTypes.put(at.getName(), at));
+  }
+
+  public Map<SchedulingActivityDirectiveId, ActivityDirectiveId> getActivityIdCorrespondence(){
+    return planActDirectiveIdToSimulationActivityDirectiveId;
   }
 
   /**
--- a/scheduler-server/sql/scheduler/tables/scheduling_request.sql
+++ b/scheduler-server/sql/scheduler/tables/scheduling_request.sql
@@ -7,6 +7,7 @@ create table scheduling_request (
   status status_t not null default 'pending',
   reason jsonb null,
   canceled boolean not null default false,
+  dataset_id integer default null,
 
   specification_revision integer not null,
 
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
@@ -11,6 +11,7 @@ import gov.nasa.jpl.aerie.merlin.driver.ActivityDirectiveId;
 import gov.nasa.jpl.aerie.scheduler.server.ResultsProtocol;
 import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchRequestException;
 import gov.nasa.jpl.aerie.scheduler.server.exceptions.NoSuchSpecificationException;
+import gov.nasa.jpl.aerie.scheduler.server.models.DatasetId;
 import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
 import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
 import gov.nasa.jpl.aerie.scheduler.server.remotes.ResultsCellRepository;
@@ -174,7 +175,8 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
           specId.id(),
           specRevision,
           RequestRecord.Status.FAILED,
-          reason);
+          reason,
+          Optional.empty());
     }
   }
 
@@ -183,7 +185,8 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
       final SpecificationId specId,
       final long specRevision,
       final long analysisId,
-      final ScheduleResults results
+      final ScheduleResults results,
+      final Optional<DatasetId> datasetId
   ) throws SQLException {
     postResults(connection, analysisId, results);
     try (final var setRequestStateAction = new SetRequestStateAction(connection)) {
@@ -191,7 +194,8 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
           specId.id(),
           specRevision,
           RequestRecord.Status.SUCCESS,
-          null);
+          null,
+          datasetId);
     }
   }
 
@@ -265,7 +269,9 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
               request.reason()
                   .orElseThrow(() -> new Error("Unexpected state: %s request state has no failure message".formatted(request.status()))),
               request.analysisId());
-          case SUCCESS -> new ResultsProtocol.State.Success(getResults(connection, request.analysisId()), request.analysisId());
+          case SUCCESS -> new ResultsProtocol.State.Success(getResults(connection, request.analysisId()),
+                                                            request.analysisId(),
+                                                            request.datasetId());
         }
     );
   }
@@ -324,9 +330,9 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
     }
 
     @Override
-    public void succeedWith(final ScheduleResults results) {
+    public void succeedWith(final ScheduleResults results, final Optional<DatasetId> datasetId) {
       try (final var connection = dataSource.getConnection()) {
-        succeedRequest(connection, specId, specRevision, analysisId, results);
+        succeedRequest(connection, specId, specRevision, analysisId, results, datasetId);
       } catch (final SQLException ex) {
         throw new DatabaseException("Failed to update scheduling request state", ex);
       }
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PreparedStatements.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PreparedStatements.java
@@ -36,4 +36,13 @@ public final class PreparedStatements {
       return SchedulerParsers.scheduleFailureP.parse(reader.readValue()).getSuccessOrThrow();
     }
   }
+
+  public static Optional<Long> getDatasetId(final ResultSet results, final String columnLabel)
+  throws SQLException
+  {
+    final var datasetId = results.getObject(columnLabel);
+    return datasetId == null ?
+        Optional.empty() :
+        Optional.of(results.getLong(columnLabel));
+  }
 }
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SetRequestStateAction.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SetRequestStateAction.java
@@ -1,18 +1,22 @@
 package gov.nasa.jpl.aerie.scheduler.server.remotes.postgres;
 
+import gov.nasa.jpl.aerie.scheduler.server.models.DatasetId;
 import gov.nasa.jpl.aerie.scheduler.server.services.ScheduleFailure;
 import org.intellij.lang.annotations.Language;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.sql.Types;
+import java.util.Optional;
 
 /*package-local*/ final class SetRequestStateAction implements AutoCloseable {
   private static final @Language("SQL") String sql = """
     update scheduling_request
       set
         status = ?,
-        reason = ?::json
+        reason = ?::json,
+        dataset_id = ?
       where
         specification_id = ? and
         specification_revision = ?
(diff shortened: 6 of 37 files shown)


adrienmaillard:
Merlin db tests are failing. I am going to take a look at it.

Mythicaeda:
> Merlin db tests are failing. I am going to take a look at it.

It looks like the errors are caused by tests calling `allocateDataset()` (L1173-L1193 in `MerlinDatabaseTests`). Since this PR made calling `allocate_dataset` a trigger, you should be safe to remove or replace that in MerlinDBTests.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -265,7 +268,9 @@ private static Optional<ResultsProtocol.State> getRequestState(
               request.reason()
                   .orElseThrow(() -> new Error("Unexpected state: %s request state has no failure message".formatted(request.status()))),
               request.analysisId());
-          case SUCCESS -> new ResultsProtocol.State.Success(getResults(connection, request.analysisId()), request.analysisId());
+          case SUCCESS -> new ResultsProtocol.State.Success(getResults(connection, request.analysisId()),
+                                                            request.analysisId(),
+                                                            request.datasetId().orElseThrow(() -> new Error("Unexpected state: %s request state has no dataset id".formatted(request.status()))));
```
Mythicaeda: How will this `get or else throw` interact with #708? Ie, if all of the goals are set to not simulate before or after, what will happen?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PreparedStatements.java
```diff
@@ -36,4 +36,21 @@ private static ScheduleFailure deserializeScheduleFailure(final String failureJs
...
+    return datasetId == null ?
+        Optional.empty() :
+        Optional.of(results.getLong(columnLabel));
+  }
+
+  public static void setDatasetId(final PreparedStatement statement, final int parameter, final Long datasetId)
+  throws SQLException
+  {
+    statement.setObject(parameter, datasetId);
+  }
```
Mythicaeda: Is there a reason this method can't be inlined in `SetRequestStateAction.apply()`?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
```diff
@@ -254,6 +257,25 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
     }
   }
 
+  private DatasetId storeSimulationResults(PlanningHorizon planningHorizon, SimulationFacade simulationFacade, PlanMetadata planMetadata,
+                                           final Map<SchedulingActivityDirective, ActivityDirectiveId> instancesToIds)
+  throws PlanServiceException, IOException
+  {
+    //finish simulation until end of horizon before posting results
+    simulationFacade.computeSimulationResultsUntil(planningHorizon.getEndAerie());
+    final var instancesIdToMerlinIds = instancesToIds
```
Mythicaeda: Minor, but we're using the term `Activity Directives` now, not `Activity Instances`.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -1,28 +1,47 @@
...
 import gov.nasa.jpl.aerie.scheduler.server.http.InvalidJsonException;
+import gov.nasa.jpl.aerie.scheduler.server.models.ActivityAttributesRecord;
+import gov.nasa.jpl.aerie.scheduler.server.models.DatasetId;
 import gov.nasa.jpl.aerie.scheduler.server.models.GoalId;
 import gov.nasa.jpl.aerie.scheduler.server.models.MerlinPlan;
 import gov.nasa.jpl.aerie.scheduler.server.models.MissionModelId;
 import gov.nasa.jpl.aerie.scheduler.server.models.PlanId;
 import gov.nasa.jpl.aerie.scheduler.server.models.PlanMetadata;
+import gov.nasa.jpl.aerie.scheduler.server.models.ProfileSet;
+import gov.nasa.jpl.aerie.scheduler.server.models.Timestamp;
```
Mythicaeda: This import is unused.

Mythicaeda (changes requested):
Looks good overall, although I'd like to see at least one new e2e test that tests that everything is being uploaded to the DB correctly.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
adrienmaillard: From what I understand from https://github.com/NASA-AMMOS/aerie/pull/708 : let's say that simulation was disabled for all goals, there will be no activity in the simulation. Code from this PR takes activities from the simulation and tries to match them up with activities from the scheduler's plan. As simulation was disabled, it will not insert any new spans. But it will insert simulation results that are inconsistent with the state of the plan as it will compute the simulation results on the basis of the initial plan. I could add a step where we verify that there is a 1:1 correspondance between activities in the schedule and in the simulation before we store these results.

adrienmaillard: If there is no correspondance, we do not return a dataset.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PreparedStatements.java
adrienmaillard: not really, I can move it.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
adrienmaillard: Changed to `instancesIdToMerlinIds` to `schedulingActivityDirectiveToMerlinActivityDirective` and `instancesToIds` to `schedulingActivityDirectiveToMerlinActivityDirectiveIds`

adrienmaillard:
@Mythicaeda I have made the changes.

deployment/hasura/migrations/AerieScheduler/1_reference_dataset_id_from_scheduling_request/down.sql
```diff
@@ -0,0 +1,4 @@
+alter table scheduling_request
+  drop column dataset_id
+
+call migrations.mark_migration_applied('1');
```
Mythicaeda: This should be 
```suggestion
call migrations.mark_migration_rolled_back('1');
```

deployment/hasura/migrations/AerieMerlin/5_add_create_partition_trigger/up.sql
```diff
@@ -0,0 +1,14 @@
...
+    perform allocate_dataset_partitions(new.id);
+return new;
+end $$;
+
+create trigger create_partition_on_simulation
+  after insert on dataset
+  for each row
+  execute function call_create_partition();
+
+call migrations.mark_migration_applied('5');
```
Mythicaeda: This needs to be called in merlin-servers's applied_migrations.sql as well.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/SetRequestStateAction.java
```diff
@@ -28,12 +32,18 @@ public void apply(
...
+      final Optional<DatasetId> datasetId
   ) throws SQLException {
     this.statement.setString(1, status.label);
     PreparedStatements.setFailureReason(this.statement, 2, failureReason);
-    this.statement.setLong(3, specificationId);
-    this.statement.setLong(4, specificationRevision);
+    if(datasetId.isPresent()) {
+      statement.setLong(3, datasetId.get().id());
+    } else{
+      statement.setNull(3, Types.BIGINT);
```
Mythicaeda: Should be:

```suggestion
      statement.setNull(3, Types.INTEGER);
```

In order to match `scheduling_request.dataset`'s type.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
Mythicaeda: _If_ the user understands that the dataset posted to the DB may be inaccurate to the state of the plan if they turn off sim for some goals, I don't think we need to do 1:1 correspondence check. So this should be fine unless the Scheduler ever uses the last Simulation Results as its base, in which case `datasetId` should probably be `null`, since Scheduling didn't produce a dataset.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
```diff
@@ -254,6 +258,31 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
...
+    final var schedulingActivityDirectiveToMerlinActivityDirective = schedulingActivityDirectiveToMerlinActivityDirectiveIds
+        .entrySet()
+        .stream()
+        .collect(Collectors.toMap((a) -> new SchedulingActivityDirectiveId(a.getKey().id().id()), Map.Entry::getValue));
+    final var simulationActivityCorrespondance = simulationFacade.getActivityIdCorrespondance();
+    final var simulationActivityDirectiveIdToMerlinActivityDirectiveId =
+        simulationActivityCorrespondance.entrySet().stream().collect(Collectors.toMap(
+        Map.Entry::getValue,
+        (a) -> schedulingActivityDirectiveToMerlinActivityDirective.get(a.getKey())));
+    if(simulationActivityDirectiveIdToMerlinActivityDirectiveId.values().equals(schedulingActivityDirectiveToMerlinActivityDirectiveIds.values())){
```
Mythicaeda: The names, while verbose, do not actually match the types of the objects. Specifically:

- `schedulingActivityDirectiveToMerlinActivityDirective` is a `Map<SchedulingActivityDirectiveId, ActivityDirectiveId>`
- `simulationActivityDirectiveIdToMerlinActivityDirectiveId` is a `Map<ActivityDirectiveId, ActivityDirectiveId>`

As such, I'm not confident that this method is correct. Could you go over exactly how this check is supposed to work?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
```diff
@@ -254,6 +258,31 @@ public void schedule(final ScheduleRequest request, final ResultsProtocol.Writer
...
+                                                     final Map<SchedulingActivityDirective, ActivityDirectiveId> schedulingActivityDirectiveToMerlinActivityDirectiveIds)
+  throws PlanServiceException, IOException
+  {
+    //finish simulation until end of horizon before posting results
+    simulationFacade.computeSimulationResultsUntil(planningHorizon.getEndAerie());
+    final var schedulingActivityDirectiveToMerlinActivityDirective = schedulingActivityDirectiveToMerlinActivityDirectiveIds
+        .entrySet()
+        .stream()
+        .collect(Collectors.toMap((a) -> new SchedulingActivityDirectiveId(a.getKey().id().id()), Map.Entry::getValue));
+    final var simulationActivityCorrespondance = simulationFacade.getActivityIdCorrespondance();
```
Mythicaeda: Typo:
```suggestion
    final var simulationActivityCorrespondence = simulationFacade.getActivityIdCorrespondance();
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -65,6 +70,10 @@ public void setActivityTypes(final Collection<ActivityType> activityTypes){
     activityTypes.forEach(at -> this.activityTypes.put(at.getName(), at));
   }
 
+  public Map<SchedulingActivityDirectiveId, ActivityDirectiveId> getActivityIdCorrespondance(){
```
Mythicaeda: Typo: 
```suggestion
  public Map<SchedulingActivityDirectiveId, ActivityDirectiveId> getActivityIdCorrespondence(){
```

scheduler-server/sql/scheduler/tables/scheduling_request.sql
```diff
@@ -7,6 +7,7 @@ create table scheduling_request (
   status status_t not null default 'pending',
   reason jsonb null,
   canceled boolean not null default false,
+  dataset_id integer,
```
Mythicaeda: Your migration checks are failing because you need to specify `default null` here.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/remotes/postgres/PostgresResultsCellRepository.java
adrienmaillard: Yeah it is gonna be null unless there is a 1:1 correspondence.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
adrienmaillard: the first map has all the merlin ids of currently simulated activities in its values. The second map is from scheduling plan to merlin ids.  So if both sets are equal, we know all the activities in the plan are simulated, don't we ?

Mythicaeda: 1) How does this interact with decomposing activities? 
2) The variable names are confusing. It wasn't until I renamed everything that I was able to follow how all of the remappings worked.

```java
private Optional<DatasetId> storeSimulationResults(PlanningHorizon planningHorizon, SimulationFacade simulationFacade, PlanMetadata planMetadata,
                                                     final Map<SchedulingActivityDirective, ActivityDirectiveId> schedDirectiveToMerlinId)
  throws PlanServiceException, IOException
  {
    //finish simulation until end of horizon before posting results
    simulationFacade.computeSimulationResultsUntil(planningHorizon.getEndAerie());
    final var schedID_to_MerlinID =
        schedDirectiveToMerlinId.entrySet().stream()
        .collect(Collectors.toMap(
            (a) -> new SchedulingActivityDirectiveId(a.getKey().id().id()), Map.Entry::getValue));
    final var schedID_to_simID =
        simulationFacade.getActivityIdCorrespondance();
    final var simID_to_MerlinID =
        schedID_to_simID.entrySet().stream().collect(Collectors.toMap(
        Map.Entry::getValue,
        (a) -> schedID_to_MerlinID.get(a.getKey())));
    if(simID_to_MerlinID.values().equals(schedDirectiveToMerlinId.values())){
      return Optional.of(planService.storeSimulationResults(planMetadata,
                                                simulationFacade.getLatestDriverSimulationResults(),
                                                simID_to_MerlinID));
    } else{
      //schedule in simulation is inconsistent with current state of the plan (user probably disabled simulation for some of the goals)
      return Optional.empty();
    }
  }
```

adrienmaillard:
I need to write e2e tests for this.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
adrienmaillard: 1. `schedID_to_simID` only contains ids of activity directives. Child activities ids are not stored. So we check only the correspondence for activity directives. 
Then for storing results, we proceed exactly like the merlin server. Thus, spans of child activities are stored. 
2. I will use your naming.

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table scheduling_request
+  add column dataset_id integer default null
+
+call migrations.mark_migration_applied('2');
```
Mythicaeda: This migration needs to be marked as applied in the Scheduler's `applied_migrations.sql`.

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/down.sql
```diff
@@ -0,0 +1,4 @@
+alter table scheduling_request
+  drop column dataset_id
```
Mythicaeda: Missing `;` at the end of the line.

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
```diff
@@ -0,0 +1,4 @@
+alter table scheduling_request
+  add column dataset_id integer default null
```
Mythicaeda: Missing `;` at the end of the line.

e2e-tests/src/tests/scheduler.test.ts
```diff
@@ -184,20 +190,90 @@ test.describe('Scheduling', () => {
...
+        //these two resources are affected by the peel, grow and bite banana
+        expect(eqSet(startOffsetOfResource, startOffsetsActivityDirectives)).toEqual(true)
+      }
+    }
+    var topics = await req.getTopicsEvents(request, dataset_id)
+    var prefixInput = "ActivityType.Input."
+    var prefixOutput = "ActivityType.Output."
+    var peelIsThere = false
+    var biteIsThere = false
+    var growIsThere = false
```
Mythicaeda: These (and the other `vars` in this test), should be `let` or `const` instead of `var`, since `let` has a more restrictive scope, which helps avoid cross-test pollution. ([`let` is block-scoped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let), [as is `const`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const), while [`var` is global/function-scoped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var)).

e2e-tests/src/tests/scheduler.test.ts
Mythicaeda, line 1: While there are failing tests here (based on the error, it looks like the plan is getting deleted before the `Get Plan` test), I was wondering if you could explain why the scheduler tests test deleting a plan/mission model and uploading a jar?

e2e-tests/src/tests/scheduler.test.ts
```diff
@@ -184,20 +190,90 @@ test.describe('Scheduling', () => {
...
+          expect(topic.events.length).toEqual(12)
+          break
+      }
+    }
+    expect(peelIsThere)
+    expect(growIsThere)
+    expect(biteIsThere)
+  });
+
+
```
Mythicaeda: Extra blank line here.

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
adrienmaillard: done

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/down.sql
adrienmaillard: done

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
adrienmaillard: done

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: done

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: something was wrong in the query. I think deleting the plan/mission model and stuff is just to separate steps. I don't remember.

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: done

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
Mythicaeda: Change is not on GitHub

e2e-tests/src/tests/scheduler.test.ts
Mythicaeda: Not done, as I called out other `vars` in this comment that were not addressed (ie, line 246).

adrienmaillard: ok, i changed all the var to let.

deployment/hasura/migrations/AerieScheduler/2_reference_dataset_id_from_scheduling_request/up.sql
adrienmaillard: now it should be

e2e-tests/src/tests/scheduler.test.ts
```diff
@@ -184,18 +190,87 @@ test.describe('Scheduling', () => {
...
     while (it++ < max_it && (status == 'pending' || status == 'incomplete')) {
-      const { reason, status, analysisId } = await req.schedule(request, specification_id);
+      const { reason, status, analysisId, datasetId } = await req.schedule(request, specification_id);
       status_local = status;
       reason_local = reason;
       expect(status).not.toBeNull();
       expect(status).toBeDefined();
+      if(datasetId != null){
+        dataset_id = datasetId
+      }
```
Mythicaeda: Why is there a null check before assigning `dataset_id`?

e2e-tests/src/tests/scheduler.test.ts
```diff
@@ -184,18 +190,87 @@ test.describe('Scheduling', () => {
...
+          expect(topic.events.length).toEqual(1)
+          break
+        case prefixInput+"PeelBanana":
+          peelIsThere = true
+          expect(topic.events.length).toEqual(12)
+          break
+        case prefixOutput+"PeelBanana":
+          expect(topic.events.length).toEqual(12)
+          break
+      }
```
Mythicaeda: Regarding this switch:
- Can you add an `expect` for the size of `topic` (looks like it should be 6)?
- Can you add booleans to track that the prefixOutput topics occur as well?

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: because datasetId is null until scheduling is finished.

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: 1) No, topics can contain more stuff, in this example, it contains also a topic about a resource. And I am not sure exactly why. 
2) What do you mean ? in the switch, there both prefixInput and prefixOutput.

adrienmaillard: Oh I think I got it for the booleans.

mattdailis: > it contains also a topic about a resource. And I am not sure exactly why

The mission model can declare topics that it would like to include in simulation results:

https://github.com/NASA-AMMOS/aerie/blob/b96fce26102d88752e4edcc148f08ab5b87b1a85/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java#L41

e2e-tests/src/tests/scheduler.test.ts
Mythicaeda: Right, but what's the harm in assigning dataset_id to a null value?

e2e-tests/src/tests/scheduler.test.ts
Mythicaeda: Oh, if it also is supposed to have a resource topic, then those resource topics should also be checked that they exist. Otherwise, this part of the test isn't asserting that all of the topics were uploaded.

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: I don't remember at this point. I have removed the if.

e2e-tests/src/tests/scheduler.test.ts
adrienmaillard: ok I have added the check for the topic.

Mythicaeda (approved):
LGTM! Just needs the MerlinDB migration numbers updated one final time.

Issue #659, opened 2023-02-04
Include more information with simulation error messages


parkerabercrombie:
When a simulation fails, it would be really helpful to include the sim time at which the error occurred in the error message that displays in the error console, as well as the name and ID of the activity that threw the error. This would make it _much_ easier to track down issues in the simulation.

```[tasklist]
### Tasks
- [x] Display simulation time at which the error occurred
- [x] Display id of activity that threw the error, if the error was thrown by an activity
```

dandelany:
This came up in discussions yesterday & today with @joswig as something we'd like to improve before the clipper launch release, so I'm going to add the "clipper-launch" label to this & prioritize it. To clarify:

* The main outstanding requirement is: When simulation fails due to an error, the returned error message should include the **activity name** and **ID** of all current "in progress" activities, as well as any additional useful context information that would help debug.
  - Per @mattdailis ^this may be difficult for child activities, and we may only be able to return the name/ID of the **parent** activity, but this should be investigated. Not a blocker - even having parent ID would be an improvement.
* We have also discussed implementing a few other features that would improve this, but are out of scope of this ticket & will be tracked separately:
  - @mattdailis has done some work on a UI for viewing events sent by the server during sim, which would help with debugging - currently a draft but will be PR'ed when ready.
  - It should be possible to display a failed simulation on the timeline as a "partial" sim, ideally displaying the activities which were generated up until the point of the error, and a vertical guideline showing the time at which the error occurred. Will open a separate ticket for this.

dandelany:
Sending to @mattdailis for scoping exactly what the message should contain...

mattdailis:
Slightly workshopping the bullet point above:
> The main outstanding requirement is: When simulation fails due to an error, the returned error message should include the activity name and ID of all current "in progress" activities, as well as any additional useful context information that would help debug. 

I would suggest - When an exception is thrown in a task, identify the directive id that resulted in that task being spawned (i.e. walk the `taskParent` tree up to the directive), and include that information in the simulation exception.

As a starting point - `stepTaskHelper` can catch exceptions from the task's `step` method, and wrap those exceptions in a new exception type that includes the directive id (if one exists), or, if it doesn't, indicates that this is a daemon task.

This design may be impacted by #1174 , which removes the `taskParent` map 🤔 We may need to brainstorm ways to do this bookkeeping in that scenario.

PR #724, opened 2023-03-10
Report the time at which an exception occurs during simulation

mattdailis:
* **Tickets addressed:** Related to #659 
* **Review:** By commit  

## Description

Users have reported that it's difficult to debug an exception that occurs during simulation because they don't know where in the plan the exception occurred.

There are two aspects to this: reporting simulation time, and reporting a specific directive and/or child activity in which the exception occurred. The simulation time aspect of this is very straightforward, and I believe will deliver a lot of value even without the specifics of which activity caused the exception.

The approach taken is to introduce a SimulationException that holds on to the time at which the exception was thrown. The simulation loop in the driver is wrapped in a try/catch, which will rethrow any exception, but wrap it in a SimulationException. The simulation agent reports the wrapped exception as a new SIMULATION_EXCEPTION failure type, and attaches two fields to that - "elapsedTime", and "utcTimeDoy".

## Verification

I modified BiteBanana to throw an exception, and tested via the UI - note the JSON object above the trace.

Screenshot 2023-03-10 at 6 40 11 AM

This required a slight change in the UI - since it currently assumes that either there is a stack trace, or there is json data - this is a case where we want both.

## Documentation

No documentation was updated.

## Future work

Look into how we could report the specific directive id, and think about implications for daemon tasks or child activities.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationException.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationException.java
@@ -1,0 +1,71 @@
+package gov.nasa.jpl.aerie.merlin.driver;
+
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
+
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.HOUR;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTE;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.negate;
+
+public class SimulationException extends RuntimeException {
+  // This builder must be used to get optional subsecond values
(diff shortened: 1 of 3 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationException.java
```diff
@@ -0,0 +1,71 @@
...
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTE;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.negate;
+
+public class SimulationException extends RuntimeException {
+  // This builder must be used to get optional subsecond values
+  // See: https://stackoverflow.com/questions/30090710/java-8-datetimeformatter-parsing-for-optional-fractional-seconds-of-varying-sign
+  public static final DateTimeFormatter format =
+      new DateTimeFormatterBuilder()
+          .appendPattern("uuuu-DDD'T'HH:mm:ss")
```
Mythicaeda, line 22: How does `uuuu` differ from `yyyy`?

Mythicaeda (approved):
LGTM!

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationException.java
Twisol: Via [this SO answer](https://stackoverflow.com/a/41178418):

> `year-of-era` means the year within the given scope (era) [...]
>
> `year` is the signed year, where year 0 is 1 BC, year -1 is 2 BC, and so forth.

So `y` (year-of-era) should always be used in conjunction with `G` (era, "BC" vs. "AD" above). If `G` isn't appropriate, `u` (year) should be used instead of `y`.

PR #725, opened 2023-03-13
Implement Temporal Subset Simulation

### LINKED ISSUE #680: Allow setting simulation boundary on a subset of plan duration

Mythicaeda:
* **Tickets addressed:** Closes #680
* **Review:** By commit  

## Description

<details open><summary><h3>Database Changes</h3></summary>
<p>

- Columns for tracking simulation start time and simulation end time were added to `simulation` and `simulation_template`. They are both nullable, but between the two tables each column must have a value for a given row in `simulation`. For example, if a `simulation` row references a template, either the template or the simulation row can provide the sim start and end time, but if a `simulation` row doesn't reference a template, it must provide these values.
- A trigger was created to insert a row into `simulation` upon a row being inserted into `plan`
- A constraint was added to `simulation` to allow only one row per plan (If this constraint is currently violated in the DB, the migration introducing it will fail).
- `simulation_dataset` now has columns for what bounds the simulation was run over, as well as what the provided arguments were.
- `simulation_dataset`'s `offset_from_plan_start` is now solely maintained via trigger rather than by direct update.
-  The `simulated_activity` view was updated to have the correct start time for spans from temporal subset simulations

</p>
</details>

<details open><summary><h3>Merlin Changes</h3></summary>
<p>

- Methods for adjusting the start offset of a List of ActivityDirectives and filtering out ActivityDirectives with negative start offsets have been added to StartOffsetReducer.
- Plan, SimulationRecord, SimulationTemplateRecord, and SimulationDatasetRecord all have had simulationStartTime and simulationEndTime added as fields.
- `getPlan` was split into `getPlanForValidation` and `getPlanForSimulation` in order to avoid needlessly hitting up the database.
- SimulationResults now has a `duration` field.
- Simulation start and end time must now be specified by either the sim config or the template (see Database Changes for more details)

</p>
</details>

<details open><summary><h3>Breaking Changes</h3></summary>
<p>

The following breaking changes occurred:
1. The signature of SimulationDriver.simulate changed to
```java
  public static <Model> SimulationResults simulate(
      final MissionModel<Model> missionModel,
      final Map<ActivityDirectiveId, ActivityDirective> schedule,
      final Instant simulationStartTime,
      final Duration simulationDuration,
      final Instant planStartTime,
      final Duration planDuration
  )
```
2. The signature of CreateSimulationMessage changed to 
```java
  public CreateSimulationMessage {
    Objects.requireNonNull(missionModelId);
    Objects.requireNonNull(simulationStartTime);
    Objects.requireNonNull(simulationDuration);
    Objects.requireNonNull(planStartTime);
    Objects.requireNonNull(planDuration);
    Objects.requireNonNull(activityDirectives);
    Objects.requireNonNull(configuration);
  }
```
3. The structure of a SimulationResults changed to 
```java
public SimulationResults(
        final Map<String, Pair<ValueSchema, List<ProfileSegment<RealDynamics>>>> realProfiles,
        final Map<String, Pair<ValueSchema, List<ProfileSegment<SerializedValue>>>> discreteProfiles,
        final Map<SimulatedActivityId, SimulatedActivity> simulatedActivities,
        final Map<SimulatedActivityId, UnfinishedActivity> unfinishedActivities,
        final Instant startTime,
        final Duration duration,
        final List<Triple<Integer, String, ValueSchema>> topics,
        final SortedMap<Duration, List<EventGraph<Pair<Integer, SerializedValue>>>> events)
```
4. Simulation bounds need to be specified in the database by either the `simulation` or `template` rows. (See Database Changes for more information)

</p>
</details>

## Verification

- MerlinDatabaseTests were updated to reflect the added "one simulation row per plan" constraint and the "create simulation row when a plan is created" trigger.
- Tests were added to AnchorTests.StartOffsetReducerTests to test the new static methods. Notably, since the new methods and the constructor do some input validation/special behavior on `null` inputs, there are tests for these cases.
- A new set of tests, TemporalSubsetSimulationTests were added to verify that simulating a subset of a plan works. Since Aerie does not distinguish between full simulations and temporal subset simulations, the existing decomposition tests shouldn't need to be repeated here.

## Documentation

Documentation ticket for the API changes can be found [here](https://github.com/NASA-AMMOS/aerie-docs/issues/21).

## Future work

- Since the simulation-related fields are only initialized for a Plan object when calling `getPlanForSimulation`, I'd like to split them from the Plan class, either by putting them in their own class or by moving them into a subclass of Plan.
- Adding a Plan End timestamp column, likely maintained via a trigger like `simulation_dataset`'s `offset_from_plan_start` column.


e2e-tests/src/tests/simulation-microsecond-resolution.test.ts
```diff
@@ -38,10 +33,16 @@ test.describe('Simulation microsecond resolution', () => {
...
+    const simulationBounds : UpdateSimulationBoundsInput = {
+      plan_id : plan_id,
+      simulation_start_time: plan_start_timestamp,
+      simulation_end_time: plan_end_timestamp
+    };
 
     const max_it = 10;
     let it = 0;
     await req.insertActivity(request, activityToInsert);
+    await req.updateSimulationBounds(request, simulationBounds);
```
adrienmaillard, line 45: Why do you need to update the simulation bounds here ? Should it not fall back on plan boundaries without it ?

merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/TemporalSubsetSimulationTests.java
```diff
@@ -0,0 +1,961 @@
...
+  final Duration oneMinute = Duration.of(1, Duration.MINUTES);
+  final Duration fourAndAHalfHours = Duration.of(270, Duration.MINUTES);
+
+  private final Instant planStart = Instant.parse("2023-01-01T00:00:00Z");
+
+  private final Map<String, SerializedValue> arguments = Map.of("unusedArg", SerializedValue.of("test-param"));
+  private final SerializedActivity serializedDelayDirective = new SerializedActivity("DelayActivityDirective", arguments);
+  private final SerializedValue computedAttributes = new SerializedValue.MapValue(Map.of());
+
+  private static void assertEqualsSimulationResults(SimulationResults expected, SimulationResults actual){
```
adrienmaillard, line 46: Not checking resources ? Because if other things are the same, resource should be the same on the assumption that tasks are deterministic ?

adrienmaillard (approved):
Looks good, you answered most of my questions yesterday anyway. Great job.

e2e-tests/src/tests/simulation-microsecond-resolution.test.ts
Mythicaeda: The code currently does not falls back on plan bounds, and the start and end time columns of the simulation row  initialize to `NULL`. Therefore, we need to update the sim bounds here.

merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/TemporalSubsetSimulationTests.java
Mythicaeda: I don't compare them because the model doesn't have any resources. (Model is defined [here](https://github.com/NASA-AMMOS/aerie/blob/f6c9956c31d412de1d6f254a7ac5ebe09fd86a09/merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/AnchorSimulationTest.java#L1059-L1222))

mattdailis:
Could you fill out the Documentation section of the PR template? If no documentation has been written - could you link to the ticket to write this documentation?

I just pulled this branch and will play around a bit now - I don't expect to do a super detailed code review since we chatted about this a couple weeks ago, and it sounds like you guys had a good code walkthrough last week, but I would like to glance over it with the freshest pair of eyes I've had in a while 😅

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -44,34 +45,44 @@ public PostgresResultsCellRepository(final DataSource dataSource) {
...
+      if (startTime == null || endTime == null) {
+        if (simulation.simulationTemplateId().isPresent()) {
+          try (final var getSimulationTemplate = new GetSimulationTemplateAction(connection)) {
+            final var templateOptional = getSimulationTemplate.get(simulation.simulationTemplateId().get());
+            if (templateOptional.isEmpty()) {
+              throw new RuntimeException("TemplateRecord should not be empty");
+            }
+            template = templateOptional.get();
+            startTime = (startTime == null ? template.simulationStartTime() : startTime);
+            endTime = (endTime == null ? template.simulationEndTime() : endTime);
```
mattdailis, line 62: If there is a template, but the template's start or end time is null - that means `startTime` or `endTime` could be null after line 63, right? Is this a nominal scenario, or will this trigger a null pointer exception later on?

Mythicaeda: Thanks for the catch! So, error-wise, it'll throw an NPE in createSimulationDataset() on line 73 when it tries to parse the timestamp, but I'll change it to a better error.

Mythicaeda: Fixed!

mattdailis (approved):
Looks reasonable to me. I didn't pull out my fine-toothed comb, but Adrien's review, and the level of testing give me confidence in this change. I'm looking forward to exercising it more as we implement the UI portion of this feature 🚀

merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/TemporalSubsetSimulationTests.java
adrienmaillard: I see, I thought you wrote a general method for comparing simulation results.

e2e-tests/src/tests/simulation-microsecond-resolution.test.ts
adrienmaillard: OK.

PR #727, opened 2023-03-14
Use enums for "enum boolean" types for sequences

### LINKED ISSUE #726: Use actual enums for "enum boolean" types for sequences
Since https://github.com/NASA-AMMOS/aerie/commit/93236b75f0a62afa132eb1e87e6d8cdabc8cb40c ([v0.11.0](https://github.com/NASA-AMMOS/aerie/releases/tag/v0.11.0)) the sequence editor is detecting boolean-valued  enums and typing them as booleans rather than their enumerated value (e.g. "TRUE" or "FALSE"). This is causing some confusion with downstream tools where they are expecting the enumerated string instead of a boolean (as defined in the command dictionary). This task is to revert that change so enums are mapped to enums, instead of booleans.

camargo:
* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/726
* **Review:** By commit  

## Description

See the attached issue.

## Verification

Updated tests. To test for review please run an expansion with an enumerated boolean locally.

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
+++ b/sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
@@ -145,16 +145,7 @@ ${parameters.length > 0 ? parameters.join('\n') : '*'}
 function mapArgumentType(argument: ampcs.FswCommandArgument, enumMap: ampcs.EnumMap): string {
   switch (argument.arg_type) {
     case 'enum':
-      // boolean enum shouldn't be 'TRUE | FALSE' but of `boolean` type
-      if (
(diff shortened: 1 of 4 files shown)


joswig (commented):
My preference is to remove the special treatment and handle the arguments exactly as specified in the command dictionary. If we have requirements that contradict that, then I'll defer to those.

- doing custom type massaging is likely to differ from what other tools that load the dictionary do
- if the argument later has a 3rd value added then we'll break existing sequences

goetzrrGit (approved):
Looks great!!!

PR #733, opened 2023-03-14
Upgrade seq-json-schema in sequencing server to use 1.0.17

### LINKED ISSUE #728: Upgrade seq-json-schema in sequencing server to use 1.0.17

camargo:
* **Tickets addressed:** Resolves https://github.com/NASA-AMMOS/aerie/issues/728
* **Review:** By commit  

## Description

Updates to the latest seq-json-schema

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -540,17 +540,16 @@ export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}
     }
   }
 
-  // This function takes an array of Args interfaces and converts it into an object or an array depending on whether
-  // the repeat argument is true or false. The interfaces array contains objects matching the ARGS interface.
+  // This function takes an array of Args interfaces and converts it into an object.
+  // The interfaces array contains objects matching the ARGS interface.
   // Depending on the type property of each object, a corresponding object with the name and value properties is created
-  // and added to the output. If an object's type property is "repeat", then the value property is recursively passed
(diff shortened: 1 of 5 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -633,39 +686,15 @@ export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}
       if (Array.isArray(value)) {
         // @ts-ignore : 'RepeatArgument' found in JSON Spec
         let repeatArg: RepeatArgument = {
-          value: value
-            .map(arg => {
-              return this.convertArgsToInterfaces(arg);
-            })
-            .flat(),
+          value: value.map(arg => {
+            return this.convertRepeatArgs(arg);
```
goetzrrGit, line 688: What if the repeating arguments have nested repeating arguments? We probably want to have some recursion for that.

goetzrrGit (changes requested):
For the most part, it looks great. Just curious about nested repeating args. The original function used recursion to handle that extremely rare case.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
camargo: So actually we are not supporting nested repeat arguments because the AMPCS command dictionary does not support them

cohansen (approved):
Looks great to me, thanks for adding the changes to the EDSL (I thought I caught everything) and fixing the tests!

goetzrrGit (approved):

PR #735, opened 2023-03-14
Support Full export and import of SeqJson

### LINKED ISSUE #736: Full export and import of SeqJson
For a couple of releases, you were able to mix SeqJson and eDSL within the Sequence editor but full generation wasn't fully supported. We need to be able to export and import a full sequence spec into AERIE.

goetzrrGit:
* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/736
* **Review:** By commit  

## Description
Now that we have linked the eDSL with the Json Spec, exporting the full spec from the Sequence eDSL should be possible. I have updated the Sequence export to include the remaining SeqJson spec values such as local, parameters, hardware_commands, immediate_commands, and activate. With this update, you should now be able to export and import the complete SeqJson spec from within the Sequence editor.

However, please note that there are some limitations to the Spec due to embedded logic that is not captured by the TS conversion process. Some known limitations at this time include the fact that for Local and Parameter, the enum_name is only valid if it is of type ENUM. Additionally, for Activate, you can only have ground_epoch or time attributes, but not both simultaneously.

## Verification
Updated the e2e test

## Documentation
Note the limitation below:

For Local and Parameter fields, the enum_name is only valid if it is of type ENUM. Additionally, for Activate, you can only have ground_epoch or time attributes, but not both simultaneously.

## Future work
A more complex test could involve an exhaustive end-to-end test using a large eDSL that is converted to SeqJson and back again.

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -204,10 +204,10 @@ declare global {
 }
 
 /*
-  ---------------------------------
-			  Sequence eDSL
-  ---------------------------------
-  */
+	---------------------------------
+				Sequence eDSL
+	---------------------------------
+	*/
 // @ts-ignore : 'SeqJson' found in JSON Spec
 export class Sequence implements SeqJson {
   public readonly id: string;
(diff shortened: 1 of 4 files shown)


sequencing-server/test/__snapshots__/command-types.spec.ts.snap
goetzrrGit, line 1: This is auto-generated so you can ignore it.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -262,31 +262,122 @@ export class Sequence implements SeqJson {
...
+        ? {
+            requests: this.requests.map(request => {
+              return {
+                name: request.name,
+                steps: [
+                  request.steps[0] instanceof CommandStem ||
+                  request.steps[0] instanceof Ground_Block ||
+                  request.steps[0] instanceof Ground_Event
+                    ? request.steps[0].toSeqJson()
+                    : request.steps[0],
```
cohansen, line 277: When does this else case ever get hit, are there instances where we could have another type we're not accounting for here?

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -262,31 +262,122 @@ export class Sequence implements SeqJson {
...
+                  // @ts-ignore : 'step' found in JSON Spec
+                  ...request.steps.slice(1).map(step => {
+                    if (step instanceof CommandStem || step instanceof Ground_Block || step instanceof Ground_Event)
+                      return step.toSeqJson();
+                    return step;
```
cohansen, line 282: I'm confused how this whole piece of code works, if the first element in the list of request steps is a CommandStem, Ground_Block, or Ground_Event then we call `toSeqJson()`, and then we look at the rest of the array `slice(1)` and do the same check again?

Can't we just run `toSeqJson()` on every element of the array if it's one of those three types?

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -262,31 +262,122 @@ export class Sequence implements SeqJson {
...
-${indent(`seqId: '${this.id}',`, 2)}
-${indent(`metadata: ${JSON.stringify(this.metadata)},`, 2)}${commandsString.length > 0 ? `\n${indent(`steps: [${commandsString}`, 2)}\n${indent('],', 2)}` : ''}
-${indent(`});`, 1)}`;
+
+    const metadataString = Object.keys(this.metadata).length == 0 ? `{}` : `${objectToString(this.metadata)}`;
+
+    const localsString = this.locals ? `[\n${indent(this.locals.map(l => objectToString(l)).join(',\n'), 1)}\n]` : '';
+
+    const parameterString = this.parameters
+      ? `[\n${indent(this.parameters.map(l => objectToString(l)).join(',\n'), 1)}\n]`
```
cohansen, line 323: Is it possible to comment some of these template literals with examples of how they output? Trying to debug these is rough if the code or the output changes later.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -308,18 +399,50 @@ ${indent(`});`, 1)}`;
...
+                    ? // @ts-ignore : 'GroundEvent' found in JSON Spec
+                      Ground_Event.fromSeqJson(r.steps[0] as GroundEvent)
+                    : r.steps[0],
+                  // @ts-ignore : 'step : Step' found in JSON Spec
+                  ...r.steps.slice(1).map(step => {
+                    if (step.type === 'command') return CommandStem.fromSeqJson(step as CommandStem);
+                    else if (step.type === 'ground_block') return Ground_Block.fromSeqJson(step as Ground_Block);
+                    else if (step.type === 'ground_event') return Ground_Event.fromSeqJson(step as Ground_Event);
+                    return step;
+                  }),
```
cohansen, line 488: Same as above, not sure why we look at the array in this way.

cohansen (changes requested):

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Ex. There is a Request with only 1 step. The user can use the eDSL `C.BAKE_BANANA, GROUND_BLOCK, GROUND_EVENT` to define a step,  or they entered in a valid step of pure SeqJSON which we should pass it through.

Basically 2 options:

```
requests: [
      {
        steps: [C.ADD_WATER ],
        ...
      }
]
```

or this:

```
    requests: [
      {
        steps: [{
          "args": [],
          "stem": "ADD_WATER",
          "time": {
            "type": "COMMAND_COMPLETE"
          },
          "type": "command"
        }
        ],
       ...
    }
]
```

goetzrrGit: Also if they define a `Load` or `Activate` in SeqJSON it will get passed through as well, since those don't have eDSLs yet

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: The reason why you cannot simply use `toSeqJson()` is because the type is `steps: [Step, ...Step[]]` instead of `steps: Step[]`. It took some time to get it functioning properly as I initially attempted to run `toSeqJson()`. It seems that the schema intends for at least one step or more steps. It can't be empty

goetzrrGit: For the array slice, we can run `toSeqJson()` on everything.

goetzrrGit: I agree it is confusing!!!

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Added an example to all the literals

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Added an example to all the literals

cohansen (approved):
Looks good, thanks for answering my questions!

PR #739, opened 2023-03-15
Add activity presets to scheduling edsl

### LINKED ISSUE #678: Add activity presets to scheduling EDSL

JoelCourtney:
* **Tickets addressed:** closes #678 
* **Review:** By commit  

## Description

This serializes all presets for a mission model into a `ActivityPresets` object in the scheduling edsl, so goal authors can use then to construct activity templates. For example if they have a BiteBanana preset called "large bite", they can use it like:

```ts
export default (): Goal => {
  return Goal.CoexistenceGoal({
    forEach: ...,
    activityTemplate: ActivityTemplates.BiteBanana(ActivityPresets.BiteBanana["large bite"]),
    startsAt: ...
  });
}
```

`ActivityPresets.BiteBanana["large bite"]` is a concrete object containing the actual values of the preset, not an expression. This means it is easy for the author to override specific values such as:

```ts
let biteArgs = ActivityPresets.BiteBanana["large bite"];
biteArgs.biteSize += 10; // even bigger

// use AcitivityTemplates.BiteBanana(biteArgs)
```

## Verification

Added a compilation service test

## Documentation
- [x] documentation written
- [ ] documentation merged

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
@@ -10,6 +10,8 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 import org.apache.commons.lang3.tuple.Pair;
 
+import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
+
 public final class TypescriptCodeGenerationService {
   private TypescriptCodeGenerationService() { }
 
@@ -128,7 +130,8 @@ return (<T>makeAllDiscreteProfile(args))
         result.add(indent(indent("get \"%s\"(): %s {".formatted(preset.getKey(), serializedValueToType(SerializedValue.of(preset.getValue()))))));
         result.add(indent(indent(indent("return {"))));
         for (final var argument: preset.getValue().entrySet()) {
-          result.add(indent(indent(indent(indent("\"%s\": %s,".formatted(argument.getKey(), serializedValueToTypescript(argument.getValue())))))));
+          final var deserializedJson = serializedValueP.unparse(argument.getValue());
+          result.add(indent(indent(indent(indent("\"%s\": %s,".formatted(argument.getKey(), deserializedJson))))));
         }
         result.add(indent(indent(indent("};"))));
         result.add(indent(indent("},")));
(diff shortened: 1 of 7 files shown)


dyst5422:
You probably want to supply the presets as read only, otherwise you could end up in a situation where the user mutates a preset (as you do above with +=) and then have unexpected behavior the second time they use the preset.

JoelCourtney:
bleh. I was assuming that `const` meant it could not be mutated. Turns out `as const` will mostly do the trick though, because it asserts that its type is equal to its value, so you can't set it to anything else. Although it does leave the door open for chicanery like this:

```ts
let x = 5 as const; // has type: 5
x = 3 as 5;
console.log(x); // 3
```

dyst5422:
You can also look at the readonly keyword to avoid that for objects and arrays and Object.freeze for runtime guarantees

JoelCourtney:
Looks like `Object.freeze` isn't just a runtime guarantee. It recursively changes the field types to readonly, no interface required. [playground example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBCCuUCmAnGBeGB5ARgKyWCgDoAzFJJALyQAoBvAKBhZgEswxUAuGJ1gTABEpNkgA2AEyG8ATM1YBfRopgBDCDFCQoASgDcjRgmQpiHLmdETJGGABZ1m2YcbaIIcUmLiQAc1oTVHNOYOspAyA) Thanks!

edit Looks like the compile-time check is deep, but the runtime guarantee is shallow. Fascinating.

dyst5422:
Because we typecheck and treat type errors as runtime errors during execution through the coderunner, we don't actually need the runtime overhead if we have the type system enforcement. So you should be able to use a type utility to make things recursively readonly and get that enforced at runtime (since we typecheck at runtime).

JoelCourtney:
Honestly I don't trust typescript's type checker any more, because its too easy to fool, which is why I want runtime checks too. I'm satisfied with this solution:

- deep freeze prevents mutations at both compile and runtime
- It also produces a very restrictive type similar to `as const`, so looser type assertions on the fields (like `5 as number`) allow copies of presets to be mutated.

So this is not ok:
```ts
let biteArgs = ActivityPresets.BiteBanana["large bite"];
biteArgs.biteSize += 10;
```

but this is:
```ts
let biteArgs = {...ActivityPresets.BiteBanana["large bite"]};
biteArgs.biteSize += 10;
```

(I've pushed this as the third commit with an added test)

dyst5422:
Fair enough. We should just make sure we have tests that try to get around the immutability creatively to ensure we have it properly covered. I imagine someone running around in circles if they accidentally mutate a preset.

JoelCourtney:
Meh. I couldn't find a clean way to deal with activity arguments that are objects themselves. So the preset is now a `get` parameter, and the two outer objects are frozen. Now the author can mutate the presets however they want and nothing matters!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
```diff
@@ -265,4 +269,96 @@ public TypescriptType onVariant(final List<ValueSchema.Variant> variants) {
...
+      }
+
+      @Override
+      public String onBoolean(final boolean value) {
+        return Boolean.toString(value);
+      }
+
+      @Override
+      public String onString(final String value) {
+        return value;
```
mattdailis: Do we need to do any string escaping here?

JoelCourtney: Yes, good catch.

JoelCourtney: I just realized that we can skip the whole `serializedValueToTypescript` nonsense with `serializedValueP.unparse(value).toString()`. 👍

mattdailis (approved):
🎉

PR #746, opened 2023-03-16
Output non-matching eDSL objects to String

goetzrrGit:
* **Review:** By commit  

## Description

Address the problem of converting eDSL from SeqJSON, wherein, if the eDSL fails to recognize the SeqJSON object, it should populate the sequence with the placeholder [object]

In this example since `Activate` and `Load` are not a defined `Step` in the eDSL so it can't convert it to eDSL and outputs step. Now it will just preserve the SeqJSON by fully outputting the object.

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    steps: [
      [Object],
      C.BAKE_BREAD
      ]
  });
```

## Verification
e2e test work as well as my manual test

### DIFF
--- a/sequencing-server/test/__snapshots__/command-types.spec.ts.snap
+++ b/sequencing-server/test/__snapshots__/command-types.spec.ts.snap
@@ -309,7 +309,7 @@ export class Sequence implements SeqJson {
                 if (step instanceof CommandStem || step instanceof Ground_Block || step instanceof Ground_Event) {
                   return step.toEDSLString() + ',';
                 }
-                return step + ',';
+                return objectToString(step) + ',';
               })
               .join('\\n'),
             1,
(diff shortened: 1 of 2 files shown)


sequencing-server/test/__snapshots__/command-types.spec.ts.snap
goetzrrGit, line 1: Auto generated file

Mythicaeda:
What would I do to test this manually?

goetzrrGit:
@Mythicaeda 
1) Download this json file (https://github.com/NASA-AMMOS/seq-json-schema/blob/develop/test/valid-seq-json/all-possible-fields.seq.json)
2) navigate to Sequence Editor in AERIE
3) Hit the browser button for `Create Sequence from Seq JSON (optional)`
4) upload the json file you download
5) You shouldn't see `[object]` for all the `step` fields

camargo (approved):
Tested locally. LGTM! Thanks for the quick fix.

PR #806, opened 2023-03-20
Add hardware commands to Command eDSL

### LINKED ISSUE #744: Add hardware commands to the sequencing EDSL
They should only be allowed in the [hardware_commands](https://github.com/NASA-AMMOS/seq-json-schema/blob/develop/schema.json#L680) section.

goetzrrGit:
* **Tickets addressed:** closes https://github.com/NASA-AMMOS/aerie/issues/744
* **Review:** By commit  

## Description
Added hardware command generation and support to the eDSL from the command dictionary.

ex.
```ts
export default () =>
  Sequence.new({
    seqId: 'test000.0001',
    metadata: {},
    hardware_commands: [
      HDW_BLENDER_DUMP
      .DESCRIPTION('dump blender')
      .METADATA({
        author: 'rrgoetz',
      })
    ],
  });
```

## Verification
Tested generation with a clipper command dictionary and updated the e2e test.

## Documentation
none

## Future work
Immediate commands

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
@@ -1,5 +1,6 @@
 import {
   CommandStem,
+  HardwareStem,
   doyToInstant,
   DOY_STRING,
   hmsToDuration,
@@ -268,6 +269,31 @@ describe('Sequence', () => {
       }),
     ],
   });`);
+    });
+
+    it('should convert with Hardware,', () => {
+      const sequence = Sequence.new({
+        seqId: 'HW',
+        metadata: {},
+        hardware_commands: [
+          HardwareStem.new({ stem: 'HW_PYRO_DUMP' }).DESCRIPTION('Fire the pyros').METADATA({ author: 'Emery' }),
+        ],
+      });
+
+      expect(sequence.toEDSLString()).toEqual(
+        'export default () =>\n' +
+          '  Sequence.new({\n' +
+          "    seqId: 'HW',\n" +
+          '    metadata: {},\n' +
+          '    hardware_commands: [\n' +
+          '      HW_PYRO_DUMP\n' +
+          "      .DESCRIPTION('Fire the pyros')\n" +
+          '      .METADATA({\n' +
+          "        author: 'Emery',\n" +
+          '      })\n' +
+          '    ],\n' +
+          '  });',
+      );
     });
   });
 });
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -31,6 +31,14 @@ export type CommandOptions<A extends Args[] | { [argName: string]: any } = [] | 
   | {}
 );
 
+export type HardwareOptions = {
+  stem: string;
+  // @ts-ignore : 'Description' found in JSON Spec
+  description?: Description;
+  // @ts-ignore : 'Metadata' found in JSON Spec
+  metadata?: Metadata;
+};
+
 export type GroundOptions = {
   name: string;
   // @ts-ignore : 'Args' found in JSON Spec
@@ -150,6 +158,26 @@ declare global {
 
     // @ts-ignore : 'Description' found in JSON Spec
     public DESCRIPTION(description: Description): CommandStem<A>;
+    // @ts-ignore : 'Description' found in JSON Spec
+    public GET_DESCRIPTION(): Description | undefined;
+  }
+
+  // @ts-ignore : 'HardwareCommand' found in JSON Spec
+  class HardwareStem implements HardwareCommand {
+    stem: string;
+
+    public static new(opts: HardwareOptions): HardwareStem;
+
+    // @ts-ignore : 'Command' found in JSON Spec
+    public toSeqJson(): HardwareCommand;
+
+    // @ts-ignore : 'Metadata' found in JSON Spec
+    public METADATA(metadata: Metadata): HardwareStem;
+    // @ts-ignore : 'Metadata' found in JSON Spec
+    public GET_METADATA(): Metadata | undefined;
+
+    // @ts-ignore : 'Description' found in JSON Spec
+    public DESCRIPTION(description: Description): HardwareStem;
     // @ts-ignore : 'Description' found in JSON Spec
     public GET_DESCRIPTION(): Description | undefined;
   }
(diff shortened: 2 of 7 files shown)


sequencing-server/test/seqjson-to-edsl.spec.ts
```diff
@@ -250,13 +250,11 @@ describe('getEdslForSeqJson', () => {
...
     ],
     hardware_commands: [
-      {
-        description: 'FIRE THE PYROS',
-        metadata:{
-          author: 'rrgoetz',
-        },
-        stem: 'HDW_PYRO_ENGINE',
-      }
+      HDW_PYRO_ENGINE
```
cohansen, line 253: Same as my other comment about this command.

sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
```diff
@@ -269,5 +270,30 @@ describe('Sequence', () => {
...
+        ],
+      });
+
+      expect(sequence.toEDSLString()).toEqual(
+        'export default () =>\n' +
+          '  Sequence.new({\n' +
+          "    seqId: 'HW',\n" +
+          '    metadata: {},\n' +
+          '    hardware_commands: [\n' +
+          '      HW_PYRO_DUMP\n' +
```
cohansen, line 289: Should this command be HDW_BLENDER_DUMP to match the new command you added to banananation?

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -17,23 +17,39 @@ function generateTypescriptCode(dictionary: ampcs.CommandDictionary): {
...
   // language=TypeScript
   const values = `
 ${typescriptFswCommands.map(fswCommand => fswCommand.value).join('\n')}
+${typescriptHwCommands.map(hwCommands => hwCommands.value).join('\n')}\n
 export const Commands = {${dictionary.fswCommands
     .map(fswCommand => `\t\t${fswCommand.stem}: ${fswCommand.stem},\n`)
     .join('')}};
 
-Object.assign(globalThis, { A:A, R:R, E:E, C:Object.assign(Commands, STEPS), Sequence});
+export const Hwds = {${dictionary.hwCommands
```
cohansen: Any reason this isn't fully spelt out like `Commands` or `Sequence`?

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -121,23 +137,48 @@ function ${fswCommandName}(...args: [{ ${argsWithType.map(arg => arg.name + ': '
...
+    value,
+    interfaces,
+  };
+}
+
 /**
  * Creates a jsdoc style doc for the given command. Right now it just includes the args as
  * parameters.
  *
  * @param fswCommand The command we're generating documentation for.
```
cohansen: Super minor but the comment here could be updated to match the new name of the argument.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -154,6 +162,26 @@ declare global {
     public GET_DESCRIPTION(): Description | undefined;
   }
 
+  // @ts-ignore : 'HardwareCommand' found in JSON Spec
+  class HardwareStem implements HardwareCommand {
```
cohansen, line 166: Since some of these properties are shared now between `HardwardStem` and `CommandStem` do you think it's worth moving them to a shared interface? I don't know if we'll be adding more types that have `METADATA` and `DESCRIPTION`s.

cohansen (changes requested):
Some minor comments, otherwise looks great.

sequencing-server/test/seqjson-to-edsl.spec.ts
goetzrrGit: You can add a fake stem here because we are not using the command dictionary to go back to eDSL. This will be similar if the user uploads a random SeqJson that doesn't match an uploaded command dictionary.

sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
goetzrrGit: The same type of thing, but we going the other way from (eDSL -> SeqJson). We are mostly testing the Class methods of `HardwareStem` and these test aren't relying on a Command Dictionary.

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: No reason just shorthand for Hardware. Let me know if you want to change it.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I thought about it but wanted to hold off until we get the full spec implemented and then do that type of refactoring. I didn't want to open a can of worms and afraid of blowing up all the test.

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
cohansen: Yeah I think we should match just for consistency!

sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
cohansen: 👍

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
cohansen: Sure as long as you're keeping it in your mind! There's probably a point where it makes sense before we finish the spec though.

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: Thanks for catching that, I updated the parameter

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: Sounds good, I made the change

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Yep, I created a ticket a while ago.

goetzrrGit:
@cohansen Should be ready to go now. I made the requested changes and the test all pass.

cohansen (approved):
👍

PR #809, opened 2023-03-22
Close SimulationEngine when Scheduling Finishes

Mythicaeda:
* **Review:** By commit  

## Description

While we closed the SimulationEngine when we restarted simulation while solving a Scheduling problem, we weren't closing the Engine when we were done with the Driver. This PR adds a `close` method to ResumableSimulationDriver and SimulationFacade and changes the SynchronousSchedulerAgent to create the SimulationFacade in a try-with-resources block so that the Engine is properly disposed of post-scheduling.

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
@@ -27,7 +27,7 @@ import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
 /**
  * A facade for simulating plans and processing simulation results.
  */
-public class SimulationFacade {
+public class SimulationFacade implements AutoCloseable{
 
   private static final Logger logger = LoggerFactory.getLogger(SimulationFacade.class);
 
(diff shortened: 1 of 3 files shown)


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -129,7 +134,7 @@ public void removeActivitiesFromSimulation(final Collection<SchedulingActivityDi
       final var oldInsertedActivities = new HashMap<>(insertedActivities);
       insertedActivities.clear();
       planActDirectiveIdToSimulationActivityDirectiveId.clear();
-      driver = new ResumableSimulationDriver<>(missionModel, planningHorizon.getAerieHorizonDuration());
+      driver.initSimulation();
```
mattdailis: I'm not sure whether or not this underlies the test failures - but if we want to be extra sure that we're not altering scheduling semantics with this hotfix PR, we can make the following more conservative change:

```suggestion
      if (driver != null) driver.close();
      driver = new ResumableSimulationDriver<>(missionModel, planningHorizon.getAerieHorizonDuration());
```

mattdailis (approved):
:shipit:

PR #821, opened 2023-03-28
Fixed argument ordering when seqjson is generated

### LINKED ISSUE #719: Command Expansion/Sequencing EDSL: Command Arguments are Order-Dependent
# Summary

In the recently updated Sequencing EDSL, named parameters are allowed to be input in any order without the linter throwing any error or warning. However, changing the order of the named arguments changes the order of the arguments in the output SeqJson.  This syntax needs to either enforce argument order (both in the linter and at runtime) or lookup argument order when generating SeqJson so that the output arguments match the command dictionary.

# Environment

- Tools: Command Expansion and Sequence Editor
- Aerie Version: Tested in 1.0.2 and 1.2.0
- Models: Demonstrated in both Clipper's `data-complex` and main mission models

# Source

In Aerie 1.0.2, I believe the problem is in the [`flatten` method](https://github.com/NASA-AMMOS/aerie/blob/ba64bbf2abbe127852f85ab691fb8b6b970ecb0a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts#L487-L511).

In the latest mission model, I see that this has been refactored into the [`convertArgsToInterfaces` method](https://github.com/NASA-AMMOS/aerie/blob/3e4e3da247e08dca569cda502fcdb1c951bf3b4d/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts#L622-L672). I haven't tested it, but it looks subject to the same issue.

# Example

## v1.0.2

Input EDSL Sequence with two calls to the same command, but with named parameters specified in a different order (this passes the linter without errors):

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    commands: [
      C.IMGR_OBSERVE({
        resolution_mode: "LO_RES",
        frame_rate: 10,
        observation_duration: 20
      }),
      C.IMGR_OBSERVE({
        observation_duration: 20,
        resolution_mode: "LO_RES",
        frame_rate: 10
      })
    ]
  });
```

In the output SeqJson, we see that the arguments are output in the same order as the named parameters used in the EDSL:

```json
{
  "id": "",
  "metadata": {},
  "steps": [
    {
      "args": [
        "LO_RES",
        10,
        20
      ],
      "metadata": {},
      "stem": "IMGR_OBSERVE",
      "time": {
        "type": "COMMAND_COMPLETE"
      },
      "type": "command"
    },
    {
      "args": [
        20,
        "LO_RES",
        10
      ],
      "metadata": {},
      "stem": "IMGR_OBSERVE",
      "time": {
        "type": "COMMAND_COMPLETE"
      },
      "type": "command"
    }
  ]
}
```

## v1.2.0

Input EDSL:

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    steps: [
      C.IMGR_OBSERVE({
        resolution_mode: "LO_RES",
        frame_rate: 10,
        observation_duration: 20
      }),
      C.IMGR_OBSERVE({
        observation_duration: 20,
        resolution_mode: "LO_RES",
        frame_rate: 10
      })
    ]
  });
```

Output SeqJson:

```ts
{
  "id": "",
  "metadata": {},
  "steps": [
    {
      "args": [
        {
          "name": "resolution_mode",
          "type": "string",
          "value": "LO_RES"
        },
        {
          "name": "frame_rate",
          "type": "number",
          "value": 10
        },
        {
          "name": "observation_duration",
          "type": "number",
          "value": 20
        }
      ],
      "stem": "IMGR_OBSERVE",
      "time": {
        "type": "COMMAND_COMPLETE"
      },
      "type": "command"
    },
    {
      "args": [
        {
          "name": "observation_duration",
          "type": "number",
          "value": 20
        },
        {
          "name": "resolution_mode",
          "type": "string",
          "value": "LO_RES"
        },
        {
          "name": "frame_rate",
          "type": "number",
          "value": 10
        }
      ],
      "stem": "IMGR_OBSERVE",
      "time": {
        "type": "COMMAND_COMPLETE"
      },
      "type": "command"
    }
  ]
}
```

cohansen:
* **Tickets addressed:** Closes #719 
* **Review:** By commit

## Description
Adds ordering of arguments back to the generated seqjson, works for nested repeat arguments as well.

## Verification
Updated automated tests and ran manual tests as well.

## Documentation
NA

## Future work
NA

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -1671,7 +1671,7 @@ function commandsWithTimeValue<T extends TimingTypes>(
 	*/
 
 // @ts-ignore : Used in generated code
-function sortCommandArguments(args: { [argName: string]: any }, order: string[]): any {
+function sortCommandArguments(args: { [argName: string]: any }, order: string[]): { [argName: string]: any } {
   if (typeof args[0] === 'object') {
     return Object.keys(args[0])
       .sort((a, b) => order.indexOf(a) - order.indexOf(b))
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -1670,6 +1670,31 @@ function commandsWithTimeValue<T extends TimingTypes>(
 	---------------------------------
 	*/
 
+// @ts-ignore : Used in generated code
+function sortCommandArguments(args: { [argName: string]: any }, order: string[]): any {
+  if (typeof args[0] === 'object') {
+    return Object.keys(args[0])
+      .sort((a, b) => order.indexOf(a) - order.indexOf(b))
+      .reduce((objectEntries: { [argName: string]: any }, key) => {
+        if (Array.isArray(args[0][key])) {
+          const sortedRepeatArgs = [];
+
+          for (const test of args[0][key]) {
+            sortedRepeatArgs.push(sortCommandArguments([test], order));
+          }
+
+          objectEntries[key] = sortedRepeatArgs;
+        } else {
+          objectEntries[key] = args[0][key];
+        }
+
+        return objectEntries;
+      }, {});
+  }
+
+  return args;
+}
+
 function indent(text: string, numTimes: number = 1, char: string = '  '): string {
   return text
     .split('\n')
--- a/sequencing-server/test/sequence-generation.spec.ts
+++ b/sequencing-server/test/sequence-generation.spec.ts
@@ -60,7 +60,7 @@ describe('sequence generation', () => {
     export default function SingleCommandExpansion(props: { activityInstance: ActivityType }): ExpansionReturn {
       return [
         C.PREHEAT_OVEN({ temperature: 70 }),
-        C.PREPARE_LOAF({ tb_sugar: 50, gluten_free: "FALSE" }),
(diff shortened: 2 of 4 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -1555,6 +1555,31 @@ function commandsWithTimeValue<T extends TimingTypes>(
...
+          objectEntries[key] = args[0][key];
+        }
+
+        return objectEntries;
+      }, {});
+  }
+
+  return args;
+}
+
```
goetzrrGit: Maybe you want to return the object type instead of `any`. 

`function sortCommandArguments(args: { [argName: string]: any }, order: string[]) : { [argName: string]: any }`

If this breaks things then you can leave it as `any`

goetzrrGit (changes requested):
No major issues stick out.

sequencing-server/test/sequence-generation.spec.ts
```diff
@@ -91,17 +91,17 @@ describe('sequence generation', () => {
...
             },
             {
-              bundle_name: "Dole",
-              number_of_bananas: 12
+              number_of_bananas: 12,
+              bundle_name: "Dole"
             }
-          ]
+          ],
+          lot_number:  1093
```
goetzrrGit, line 105: In this test you are testing the order of the repeat arguments, maybe do that for a non-repeat arguments command too just to be sure it works for them also. I imagine it does.

goetzrrGit (approved):

PR #827, opened 2023-03-29
Add immediate commands to the eDSL

### LINKED ISSUE #745: Only allow FSW commands as immediate commands in the sequencing EDSL
Immediate commands are essentially FSW commands with no time tag. We need to only allow FSW commands in the immediate commands section of the sequencing EDSL.

goetzrrGit:
* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/745
* **Review:** By commit  

## Description
We had to make a change to how we define FSWCommands in the `CommandTypescriptLib.ts` file. Previously, FSWCommands in the Command Dictionary were defined as a single interface (`CommandStem`), but we now generate two interfaces: one for `ImmediateStems` and one for `CommandStems` in the eDSL.

```ts
ImmediateStem implants ImmediateCommand //from the TS Spec
CommandStem implements Command // from the TS Spec
```

When a user writes out a command in the editor, ex. `Bake_Banana` the IntelliSense presented a list of commands that are `immediateCommand` type. However, `immediateCommand` types cannot be added to the `step[]` in the `Sequence` as defined by the TS Spec, which includes [Command, Ground_Event, Ground_Block, Activate, Load].

To add an FSWCommand to the `step[]`, you will have to use time literal methods, such as `C.Bake_Banana`. This will present the users with a list of commands that are of `Command` type. The aim of this change is to enable the editor to ensure that you are not passing a command with time attributes into an `immediate_command[]` unnecessarily, and vice versa.

Furthermore, this change enables us to detect potential issues with invalid seqJSON that may be generated, which could adversely impact downstream processes. By restricting the types of commands that can be added to the `step[]` and `immediate_commands`, we can catch any potential problems earlier.

We are currently in discussions with the SeqJSON Spec team to add a `type: immediate` for the `ImmediateCommand` type. Within the `immediate_command[]` array, the editor is not flagging `C.<Command>`. In TypeScript, an object literal can be used as a value of an interface if the object has all the same properties, and optionally some more, with compatible types as specified in the interface. However, adding `type: immediate` will resolve this issue. I have included a check in the generation of SeqJSON to notify the user, but we also want the editor to notify the user earlier.

Ex. with immediate commands
```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    steps: [C.ADD_WATER],
    immediate_commands : [ADD_WATER]
  });
```

Ex. Invalid
```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    steps: [ADD_WATER],
  });

// or

export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    immediate_commands : [C.ADD_WATER]
  });

```

## Verification
Updated and ran e2e test

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
+++ b/sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
@@ -33,8 +33,8 @@ ${typescriptFswCommands.map(fswCommand => fswCommand.interfaces).join('\n')}${ty
     .join('')}\t};
 
 \tconst Hardwares : {\n${dictionary.hwCommands
-    .map(hwCommand => `\t\t${hwCommand.stem}: typeof ${hwCommand.stem}`)
-    .join(',\n')} \t};
+    .map(hwCommand => `\t\t${hwCommand.stem}: typeof ${hwCommand.stem},\n`)
+    .join('')} \t};
 }`;
 
   // language=TypeScript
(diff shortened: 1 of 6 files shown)


sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit, line 1: This should have been in the previous commit. This is just formatting adjustments when generating the commandLib file.

sequencing-server/test/__snapshots__/command-types.spec.ts.snap
goetzrrGit, line 1: Auto-generated file.

camargo (approved):
LGTM! Tested locally and ran UI e2e tests against it.

PR #833, opened 2023-03-30
Use org.postgresql jdbc driver

### LINKED ISSUE #819: Evaluate postgres jdbc drivers
The output of this ticket should be a trade study, with pros and cons, and a recommendation for which driver we should be using.

The trade study and decision should be summarized in the Aerie SDD.

Currently we use https://impossibl.github.io/pgjdbc-ng/ which claims to be the fastest postgres jdbc driver.

The "official" postgres jdbc driver seems to have been recently revamped, and may be a more attractive option now: https://jdbc.postgresql.org/

mattdailis:
* **Tickets addressed:** Related to #819 
* **Review:** By commit  

## Description

For this PR, I replaced impossibl.pgjdbc-ng with org.postgresql, and made the updates necessary to get the tests to pass.

Most changes were related to type casting in SQL (apparently pgjdbc-ng did some magic with inferring types), and replacing uses of pgjdbc-ng's Interval with PGInterval.

The listener had to be implemented slightly differently - though it ended up being fairly straightforward.

## Verification

Mostly relying on e2e-tests - I also did some manual spot checks. Would be good to get some more comprehensive manual testing done before merging.

- [x] More comprehensive manual testing

## Documentation

If we choose to pursue this route, I'll update the SDD:

- [ ] Update SDD

## Future work

None

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
@@ -1,6 +1,6 @@
 package gov.nasa.jpl.aerie.database;
 
-import com.impossibl.postgres.api.data.Interval;
+import org.postgresql.util.PGInterval;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
@@ -13,8 +13,6 @@ import java.io.File;
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.time.Duration;
-import java.time.Period;
 import java.util.ArrayList;
 import java.util.UUID;
 
@@ -118,7 +116,7 @@ public class AnchorTests {
                   INSERT INTO plan (name, model_id, duration, start_time)
                   VALUES ('test-plan-%s', '%s', '0', '%s')
                   RETURNING id;"""
-                  .formatted(UUID.randomUUID().toString(), missionModelId, "2020-1-1 00:00:00")
+                  .formatted(UUID.randomUUID().toString(), missionModelId, "2020-1-1 00:00:00+00")
           );
       res.next();
       return res.getInt("id");
@@ -173,7 +171,7 @@ public class AnchorTests {
     }
   }
 
-  private void updateOffsetFromAnchor(Interval newOffset, int activityId, int planId) throws SQLException {
+  private void updateOffsetFromAnchor(PGInterval newOffset, int activityId, int planId) throws SQLException {
     try(final var statement = connection.createStatement()) {
       statement.execute(
           """
@@ -187,16 +185,16 @@ public class AnchorTests {
   private Activity getActivity(final int planId, final int activityId) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement.executeQuery("""
-        SELECT *
+        SELECT id, plan_id, start_offset, anchor_id, anchored_to_start, approximate_start_time
         FROM activity_directive_extended
         WHERE id = %d
         AND plan_id = %d;
       """.formatted(activityId, planId));
-      res.first();
+      res.next();
       return new Activity(
           res.getInt("id"),
           res.getInt("plan_id"),
-          (Interval) res.getObject("start_offset"),
+          (PGInterval) res.getObject("start_offset"),
           res.getString("anchor_id"),
           res.getBoolean("anchored_to_start"),
           res.getString("approximate_start_time")
@@ -218,7 +216,7 @@ public class AnchorTests {
         activities.add(new Activity(
             res.getInt("id"),
             res.getInt("plan_id"),
-            (Interval) res.getObject("start_offset"),
+            (PGInterval) res.getObject("start_offset"),
             res.getString("anchor_id"),
             res.getBoolean("anchored_to_start"),
             res.getString("approximate_start_time")
@@ -244,7 +242,7 @@ public class AnchorTests {
         WHERE activity_id = %d
         AND plan_id = %d;
       """.formatted(activityId, planId));
-      res.first();
+      res.next();
       return new AnchorValidationStatus(
           res.getInt("activity_id"),
           res.getInt("plan_id"),
@@ -257,7 +255,7 @@ public class AnchorTests {
     return getValidationStatus(original.planId, original.activityId);
   }
 
-  int insertActivityWithAnchor(final int planId, final Interval startOffset, final int anchorId, final boolean anchoredToStart) throws SQLException {
+  int insertActivityWithAnchor(final int planId, final PGInterval startOffset, final int anchorId, final boolean anchoredToStart) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
@@ -286,7 +284,7 @@ public class AnchorTests {
   private record Activity(
       int activityId,
       int planId,
-      Interval startOffset,
+      PGInterval startOffset,
       String anchorId,  // Since anchor_id allows for null values, this is a String to avoid confusion over what the number means.
       boolean anchoredToStart,
       String approximateStartTime
@@ -299,8 +297,8 @@ public class AnchorTests {
   class AnchorCreationAndExceptions {
     @Test
     void createAnchor() throws SQLException {
-      final Interval oneDay = Interval.of(Period.ofDays(1));
-      final Interval tenMinutes = Interval.of(Duration.ofMinutes(10));
+      final PGInterval oneDay = new PGInterval("1 day");
+      final PGInterval tenMinutes = new PGInterval("10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int anchorActId = insertActivity(planId);
@@ -384,13 +382,13 @@ public class AnchorTests {
   class NetNegativeEndTimeStatus {
     @Test
     void negativeEndTimeOffsetWritesToStatus() throws SQLException {
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int grandparentActId = insertActivity(planId);
-      final int parentActId = insertActivityWithAnchor(planId, Interval.ZERO, grandparentActId, false);
+      final int parentActId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), grandparentActId, false);
       final int negOffsetActId = insertActivity(planId, minusTenMinutes.toString());
-      final int childActId = insertActivityWithAnchor(planId, Interval.ZERO, negOffsetActId, true);
+      final int childActId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), negOffsetActId, true);
       final int unrelatedActId = insertActivity(planId);
 
       // Invalid regarding Plan Start
@@ -437,9 +435,9 @@ public class AnchorTests {
 
     @Test
     void immediateDescendentBecomesInvalid() throws SQLException {
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be valid.
@@ -479,9 +477,9 @@ public class AnchorTests {
 
     @Test
     void childAndGrandchildBecomeInvalid() throws SQLException {
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be valid.
@@ -491,7 +489,7 @@ public class AnchorTests {
 
       // Create a chain
       final int childActId = insertActivityWithAnchor(planId, minusTenMinutes, baseActId, true);
-      final int grandchildActId = insertActivityWithAnchor(planId, Interval.ZERO, childActId, true);
+      final int grandchildActId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), childActId, true);
 
       final AnchorValidationStatus unrelatedValidation = getValidationStatus(planId, unrelatedActId);
       final AnchorValidationStatus baseValidation = getValidationStatus(planId, baseActId);
@@ -533,9 +531,9 @@ public class AnchorTests {
      */
     @Test
     void grandchildEndTimeAnchorIsIgnored() throws SQLException {
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be valid.
@@ -545,7 +543,7 @@ public class AnchorTests {
 
       // Create a chain
       final int childActId = insertActivityWithAnchor(planId, minusTenMinutes, baseActId, true);
-      final int grandchildActId = insertActivityWithAnchor(planId, Interval.ZERO, childActId, false);
+      final int grandchildActId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), childActId, false);
 
       final AnchorValidationStatus unrelatedValidation = getValidationStatus(planId, unrelatedActId);
       final AnchorValidationStatus baseValidation = getValidationStatus(planId, baseActId);
@@ -581,9 +579,9 @@ public class AnchorTests {
 
     @Test
     void onlyGrandchildBecomesInvalid() throws SQLException {
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be empty.
@@ -592,7 +590,7 @@ public class AnchorTests {
       final int baseActId = insertActivity(planId, fifteenMinutes.toString());
 
       // Create a chain
-      final int childActId = insertActivityWithAnchor(planId, Interval.ZERO, baseActId, true);
+      final int childActId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), baseActId, true);
       final int grandchildActId = insertActivityWithAnchor(planId, minusTenMinutes, childActId, true);
 
       final AnchorValidationStatus unrelatedValidation = getValidationStatus(planId, unrelatedActId);
@@ -629,9 +627,9 @@ public class AnchorTests {
 
     @Test
     void childIsInvalid() throws SQLException {
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be empty.
@@ -658,9 +656,9 @@ public class AnchorTests {
     @Test
     void farDescendantIsInvalid() throws SQLException {
       // Parent, base is anchored to end with negative offset, 100 anchors to start of base with 0 offset. Both parent and child should be invalid
-      final Interval fiveMinutes = Interval.of(Duration.ofMinutes(5));
-      final Interval fifteenMinutes = Interval.of(Duration.ofMinutes(15));
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
+      final PGInterval fiveMinutes = new PGInterval("5 minutes");
+      final PGInterval fifteenMinutes = new PGInterval("15 minutes");
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId); // Should always be valid.
@@ -670,9 +668,9 @@ public class AnchorTests {
       // Create a chain
       final int baseActId = insertActivityWithAnchor(planId, minusTenMinutes, parentActId, false);
       final int[] interimActIds = new int[100];
-      interimActIds[0] = insertActivityWithAnchor(planId, Interval.ZERO, baseActId, true);
+      interimActIds[0] = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), baseActId, true);
       for(int i = 1; i < 100; i++){
-        interimActIds[i] = insertActivityWithAnchor(planId, Interval.ZERO, interimActIds[i-1], true);
+        interimActIds[i] = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), interimActIds[i-1], true);
       }
       final int childActId = insertActivityWithAnchor(planId, fiveMinutes, interimActIds[99], true);
 
@@ -702,8 +700,8 @@ public class AnchorTests {
   class NetNegativePlanStartStatus {
     @Test
     void negativeToPlanStart() throws SQLException {
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
-      final Interval tenMinutes = Interval.of(Duration.ofMinutes(10));
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
+      final PGInterval tenMinutes = new PGInterval("10 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int activityId = insertActivity(planId, tenMinutes.toString());
@@ -728,15 +726,15 @@ public class AnchorTests {
 
     @Test
     void negativeToPlanStartDownChain() throws SQLException {
-      final Interval minusTenMinutes = Interval.of(Duration.ofMinutes(-10));
-      final Interval elevenMinutes = Interval.of(Duration.ofMinutes(11));
+      final PGInterval minusTenMinutes = new PGInterval("-10 minutes");
+      final PGInterval elevenMinutes = new PGInterval("11 minutes");
 
       final int planId = insertPlan(missionModelId);
       final int unrelatedActId = insertActivity(planId);
       final int grandparentActId = insertActivity(planId, elevenMinutes.toString());
       final int parentActId = insertActivityWithAnchor(planId, minusTenMinutes, grandparentActId, true);
       final int baseId = insertActivityWithAnchor(planId, elevenMinutes, parentActId, true);
-      final int childId = insertActivityWithAnchor(planId, Interval.ZERO, baseId, true);
+      final int childId = insertActivityWithAnchor(planId, new PGInterval("0 seconds"), baseId, true);
 
       // Base is currently valid regarding Plan Start
       final AnchorValidationStatus baseStatus = getValidationStatus(planId, baseId);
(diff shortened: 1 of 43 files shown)


Twisol:
I wonder if it's possible to get this battle-tested a bit before we merge? It would be nice to have some confidence that (1) the issue we saw previously isn't happening, (2) there aren't any obvious strange new bugs in the new driver, and (3) things still work pretty much as normal.

(I wonder if there are any performance implications for storing simulation results, in particular.)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
```diff
@@ -8,7 +8,7 @@
 
 /*package-local*/ final class CreateDatasetPartitionsAction implements AutoCloseable {
   private static final @Language("SQL") String sql = """
-    select from allocate_dataset_partitions(?);
+    select from allocate_dataset_partitions(?::integer);
```
Twisol: Oh, this is very depressing. Do you know what specific error comes up if you leave the type annotation off?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
```diff
@@ -192,11 +192,11 @@ private Activity getActivity(final int planId, final int activityId) throws SQLE
         WHERE id = %d
         AND plan_id = %d;
       """.formatted(activityId, planId));
-      res.first();
+      res.next();
```
Mythicaeda, line 193: Why the strong preference for `res.next` over `res.first`?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -594,24 +595,26 @@ void snapshotInheritsAllLatestAsParents() throws SQLException{
 
       //take n snapshots, then insert them all into the latest table
       final int numberOfSnapshots = 4;
-      final int[] snapshotIds = new int[numberOfSnapshots];
-      for(int i = 0; i < numberOfSnapshots; ++i){
-        snapshotIds[i] = createSnapshot(planId);
+      final var snapshotIds = new ArrayList<Integer>(numberOfSnapshots);
+      for(int i = 0; i < numberOfSnapshots; i++){
+        snapshotIds.add(createSnapshot(planId));
```
Mythicaeda: Why was `snapshotIds` changes from an int[] to an ArrayList?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -594,24 +595,26 @@ void snapshotInheritsAllLatestAsParents() throws SQLException{
...
-        assertTrue(res.isLast());
+        {
+          //assert that there is exactly one entry for this plan in plan_latest_snapshot
+          final var res = statement.executeQuery(
+              """
+                  select count(1) from plan_latest_snapshot where plan_id = %d;
+                  """.formatted(planId));
+          res.next();
+          assertEquals(1, res.getInt(1));
+        }
```
Mythicaeda: This change removes the assertion described in the comment. The old code checked that there was exactly one row. The new code checks that the first row's first column equals 1.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -621,28 +624,32 @@ void snapshotInheritsAllLatestAsParents() throws SQLException{
...
+          final var history = new ArrayList<Integer>();
+          while (res.next()) {
+            history.add(res.getInt(1));
+          }
+
+          final var expected = new ArrayList<>();
+          expected.add(finalSnapshotId);
+          expected.addAll(snapshotIds);
+
+          assertEquals(expected, history);
```
Mythicaeda: If this is a combination of `//assert that the snapshot history is n+1 long` and `//assert that res contains, in order: finalSnapshotId, snapshotId[0,1,...,n]`, then the comment should reflect that.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
```diff
@@ -27,38 +28,42 @@ public ListenSchedulerCapability(
...
+              try (final var reader = Json.createReader(new StringReader(payload))) {
+                final var jsonValue = reader.readValue();
+                final var notificationPayload = postgresSchedulingRequestNotificationP
+                    .parse(jsonValue)
+                    .getSuccessOrThrow();
+                try {
+                  this.notificationQueue.put(notificationPayload);
+                } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
+                  break;
```
Mythicaeda: This `break` is only going to break out of the `for` loop. If this thread is interrupted, we likely want to break out of the `while(true)` loop above (meaning the condition would be `while(!Thread.interrupted())`. Otherwise, the thread won't respond to its `interrupted()` flag being set on the line above.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
```diff
@@ -31,16 +30,14 @@ public static void main(String[] args) throws Exception {
...
+    hikariConfig.setDataSourceClassName("org.postgresql.ds.PGSimpleDataSource");
+    hikariConfig.addDataSourceProperty("serverName", postgresStore.server());
+    hikariConfig.addDataSourceProperty("portNumber", postgresStore.port());
+    hikariConfig.addDataSourceProperty("databaseName", postgresStore.database());
+    hikariConfig.addDataSourceProperty("applicationName", "Merlin Server");
     hikariConfig.setUsername(postgresStore.user());
     hikariConfig.setPassword(postgresStore.password());
-    hikariConfig.setDataSource(pgDataSource);
 
     final var hikariDataSource = new HikariDataSource(hikariConfig);
```
Mythicaeda, line 44: Interesting thing I noticed when looking at the [Hikari docs](https://www.javadoc.io/doc/com.zaxxer/HikariCP/latest/com/zaxxer/hikari/HikariDataSource.html): `HikariDataSource` has a `close` method that we never call. Although given that the while loop below only breaks on SIGINT or SIGKILL, I'm not sure when we _could_ call it 🤔.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
Mythicaeda, line 1: This class should be deleted, as it's unused as of #723.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsers.java
Mythicaeda, line 1: The an `intervalParser` Pattern defined near the top of this class that now goes unused.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
mattdailis: Good catch, thank you

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
mattdailis: Good question - according to the JDBC spec, by default ResultSets are "forward only", meaning that they support only the `next` operation. 

https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html

Here's the place in the new driver that sets that default:

https://github.com/pgjdbc/pgjdbc/blob/3c3e5b62349c42d9d3dfe1fe0102776d4de8c353/pgjdbc/src/main/java/org/postgresql/jdbc/PgConnection.java#L515-L519

And here's where it throws an exception if you try to scroll a non-scrollable result set:

https://github.com/pgjdbc/pgjdbc/blob/3c3e5b62349c42d9d3dfe1fe0102776d4de8c353/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L312-L319

So, the fact that impossibl's pgjdbc-ng allows us to call `first`, even though the statement's type is "forward only", is a deviation from the spec.

If we want to be able to call `first`, we can specify that when creating the statement, it would look something like this:

```java
connection.createStatement(
        ResultSet.TYPE_SCROLL_SENSITIVE,
        ResultSet.CONCUR_READ_ONLY);
```

My approach in this PR was to update our code to work with the default non-scrollable result sets

Twisol: ooooh, good find

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: My intent here was to be able to directly compare the two lists for equality - in the event that the assertion fails, it provides both lists in the message, rather than just the element that was unequal.

https://github.com/NASA-AMMOS/aerie/blob/63bb374ca4346d9c5ed5390b3e8e8a8c30c5b2dc/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java#L652

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Good catch - I can add an `assertFalse(res.next())` - or read the resultset into a list and check that its length is one

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Good catch, I'll update the comment :+1:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreateDatasetPartitionsAction.java
mattdailis: If I recall correctly, it was something to the effect of:
```
ERROR:  function allocate_dataset_partitions(bigint) does not exist
```

Mythicaeda: Makes sense. [The `apply` method in this class](https://github.com/NASA-AMMOS/aerie/pull/833/files#diff-86342dd34fb02577133fccc89429aee5b12a9c350b4e5a07a7d7f2c613a9e253L20-R23) takes (took?) a long, not an int.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Done

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
mattdailis: Comment updated

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
mattdailis: I swapped out the `break` for a `return` here

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsers.java
mattdailis: I ended up reverting some of the parsing changes on the scheduler - so this pattern is now used again.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
```diff
@@ -118,7 +118,7 @@ int insertPlan(final int missionModelId) throws SQLException {
                   INSERT INTO plan (name, model_id, duration, start_time)
                   VALUES ('test-plan-%s', '%s', '0', '%s')
                   RETURNING id;"""
-                  .formatted(UUID.randomUUID().toString(), missionModelId, "2020-1-1 00:00:00")
+                  .formatted(UUID.randomUUID().toString(), missionModelId, "2020-1-1 00:00:00+00")
```
Mythicaeda, line 119: Is the `+00` how you're forcing it to grab the results in UTC instead of in your local TZ?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
```diff
@@ -299,8 +299,10 @@ private record AnchorValidationStatus(int activityId, int planId, String reasonI
   class AnchorCreationAndExceptions {
     @Test
     void createAnchor() throws SQLException {
-      final Interval oneDay = Interval.of(Period.ofDays(1));
-      final Interval tenMinutes = Interval.of(Duration.ofMinutes(10));
+      connection.prepareStatement("set time zone 'UTC'").execute();
```
Mythicaeda: Why is this statement not needed in other tests?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
Mythicaeda, line 1: Minor: There are some now-unused imports in this file

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
```diff
@@ -7,10 +7,12 @@
...
 import org.junit.jupiter.api.TestInstance;
+import org.postgresql.util.PGInterval;
 
 import java.io.File;
 import java.io.IOException;
 import java.sql.Connection;
+import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.time.Duration;
```
Mythicaeda: Minor: Import statement is now unused

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
```diff
@@ -620,13 +622,23 @@ INSERT INTO plan_dataset (plan_id, dataset_id)
...
+      final var newOffsetFromPlanStart = new PGInterval(planDatasetInsertRes.getString("offset_from_plan_start"));
       planDatasetInsertRes.close();
 
-      final var calculatedOffset = offsetFromPlanStart.minus(Duration.ofMillis(newPlanStartTime.getTime()
-                                                                               - planStartTime.getTime()));
+      final var calculatedOffset = new PGInterval(offsetFromPlanStart.getValue());
+      calculatedOffset.setSeconds(calculatedOffset.getSeconds() - (newPlanStartTime.getTime() - planStartTime.getTime()) / 1000.0);
 
-      assertEquals(calculatedOffset, newOffsetFromPlanStart);
+      assertEquals(microsOfPGInterval(calculatedOffset), microsOfPGInterval(newOffsetFromPlanStart));
```
Mythicaeda: Why do we need to call `microsOfPGInterval` here?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
```diff
@@ -1142,14 +1156,16 @@ private int insertProfile(final int datasetId) throws SQLException {
   private int insertProfile(final int datasetId, final String name, final String type, final String duration)
   throws SQLException
   {
-    try (final var statement = connection.createStatement()) {
+    try (final var statement = connection.createStatement(
+        ResultSet.TYPE_SCROLL_SENSITIVE,
+        ResultSet.CONCUR_READ_ONLY)) {
```
Mythicaeda: Why do `TYPE_SCROLL_SENSITIVE` and `CONCUR_READ_ONLY` need to be specified here?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -10,6 +10,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.sql.Array;
```
Mythicaeda: Minor: Unused Import

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -594,24 +595,27 @@ void snapshotInheritsAllLatestAsParents() throws SQLException{
...
-            """
-            select snapshot_id from plan_latest_snapshot where plan_id = %d;
-            """.formatted(planId));
-        assertTrue(res.first());
-        assertTrue(res.isLast());
+        {
+          //assert that there is exactly one entry for this plan in plan_latest_snapshot
+          final var res = statement.executeQuery(
+              """
+                  select count(1) from plan_latest_snapshot where plan_id = %d;
```
Mythicaeda: What exactly is `count(1)` selecting?

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -621,28 +625,32 @@ void snapshotInheritsAllLatestAsParents() throws SQLException{
...
+        {
+          //assert that there is now only one entry for this plan in plan_latest_snapshot
+          final var res = statement.executeQuery(
+              """
+                  select count(1) from plan_latest_snapshot where plan_id = %d;
+                  """.formatted(planId));
           res.next();
-          assertEquals(res.getInt(1), snapshotId);
+          assertEquals(1, res.getInt(1));
+        }
```
Mythicaeda: Minor: This is the same check as in [L604-L613](https://github.com/NASA-AMMOS/aerie/pull/833/files#diff-269125a2369e062ea57b1a0ae448f5e159d3c8f69552c13ace30c65533e246b2R604-R613), so it should be written the same.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresParsers.java
```diff
@@ -24,6 +24,7 @@
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoField;
+import java.util.Date;
```
Mythicaeda: Minor: Unused import. Additionally, `java.time.Instant` is also an unused import.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresParsers.java
```diff
@@ -95,9 +96,21 @@ public JsonValue unparse(final Timestamp value) {
           realProfileTypeP);
 
   public static Duration parseOffset(final ResultSet resultSet, final int index, final Timestamp epoch) throws SQLException {
```
Mythicaeda: Minor: The parameter `epoch` is now unused.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresParsers.java
```diff
@@ -95,9 +96,21 @@ public JsonValue unparse(final Timestamp value) {
...
+      throw new IllegalArgumentException("Cannot convert years to microseconds");
+    }
+    if (interval.getMonths() != 0) {
+      throw new IllegalArgumentException("Cannot convert months to microseconds");
+    }
+    return interval.getMicroSeconds() +
+           (1_000_000L * interval.getWholeSeconds()) +
+           (1_000_000L * 60 * interval.getMinutes()) +
+           (1_000_000L * 3600 * interval.getHours()) +
+           (1_000_000L * 3600 * 24 * interval.getDays());
```
Mythicaeda, line 111: How does this work if this is called to parse, for example, the start offset of a span that starts two months into a year long simulation dataset? It looks like the method would throw an `IllegalArgumentException`, which I'd argue is unexpected behavior for calling [`GetSpanRecords.get()`](https://github.com/NASA-AMMOS/aerie/blob/prototype/use-org-postgresql-jdbc/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSpanRecords.java#L50) on a valid dataset.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
```diff
@@ -27,38 +28,42 @@ public ListenSchedulerCapability(
...
-      try {
-        notificationQueue.put(notificationPayload);
-      } catch (InterruptedException ex) {
-        // We are unable to handle this exception here so the thread interrupt flag is set to true
-        // so that code higher up the call stack can see that an interrupt was issued.
-        Thread.currentThread().interrupt();
+        while (true) {
+          final var pgConnection = connection.unwrap(PGConnection.class);
+          final var notifications = pgConnection.getNotifications(10000);
+          if (notifications != null && notifications.length > 0) {
```
Mythicaeda: Why do we need to check `notifications.length`? Not only should it not be 0 based on the description of `getNotifciations` (which states that it will return either an array with at least one element or null), but if a 0 length array is returned, then won't it skip the foreach loop?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
```diff
@@ -27,38 +28,42 @@ public ListenSchedulerCapability(
...
-      final var notificationPayload = postgresSchedulingRequestNotificationP
-          .parse(jsonValue)
-          .getSuccessOrThrow();
-      try {
-        notificationQueue.put(notificationPayload);
-      } catch (InterruptedException ex) {
-        // We are unable to handle this exception here so the thread interrupt flag is set to true
-        // so that code higher up the call stack can see that an interrupt was issued.
-        Thread.currentThread().interrupt();
+        while (true) {
```
Mythicaeda, line 40: This loop should be `while (!Thread.currentThread().isInterrupted())`, as otherwise we're waiting on trying to put something into the `notificationQueue` in order to detect an interrupted signal.

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/ListenSchedulerCapability.java
```diff
@@ -27,38 +28,42 @@ public ListenSchedulerCapability(
...
+              logger.info("Received PSQL Notification: {}, {}, {}", processId, channelName, payload);
+              try (final var reader = Json.createReader(new StringReader(payload))) {
+                final var jsonValue = reader.readValue();
+                final var notificationPayload = postgresSchedulingRequestNotificationP
+                    .parse(jsonValue)
+                    .getSuccessOrThrow();
+                try {
+                  this.notificationQueue.put(notificationPayload);
+                } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
```
Mythicaeda: The statement `Thread.currentThread().interrupt();` doesn't do anything here, since we return from the Thread's `run` method immediately after.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
mattdailis: In the merlin worker, we could conceivably call it using a try-with-resources, since the lifetime of the data source is in lexical scope. In the servers, I'm less sure how to go about this - I would ask a similar question about `jetty`.

Something to ponder :+1:

Mythicaeda: Yeah. From what I recall, a shutdown hook will work for SIGINT (and [this SO discussion](https://stackoverflow.com/questions/2541475/capture-sigint-in-java) seems to support that), but that doesn't affect SIGKILL.

Either way, I'm resolving this conversation so that it doesn't block this PR.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresParsers.java
mattdailis: I did some testing on `develop`, and observed that `spans` never use the `months` field of intervals :+1: I manually updated the start offset of a span to include months, and constraints checking threw an exception:

Screenshot 2023-05-31 at 5 10 55 PM

I did the same with years - and found an interesting difference between the old and new jdbc drivers. On develop, a span with a start offset of `1 year` in the database is rewritten as `P12M` by the jdbc driver, and throws the same error as in the screenshot above. (Mystifyingly, it does this only for prepared statements - a regular `executeQuery` with the same SQL results in `P1Y`, as you would expect). The new driver behaves as you would expect.

I'll poke at it a bit more - but the impression I'm getting is that this PR changes the error messages, but does not cause previously error-free code paths to error.

I would like to take a holistic look at how we use intervals, and make sure we're being consistent (or at least the inconsistencies are documented) - that can be a follow-up to this PR

Mythicaeda: Awesome! Would it be possible to include this check as a test for the `microsOfPGInterval` function?

I agree that further investigation is a separate body of work, and just including that test would be sufficient to resolve the concern my initial comment brought up.

mattdailis: Could you help me understand which check you're referring to?

Mythicaeda: The check where span records with a start offset greater than or equal to a month/year are produced and then fetched via `GetSpanRecords.get()`. (Or, equivalently, that `spans` never uses the 'month' field of intervals)

So something (really roughly) like:
```java
@Test
void microsFNTest(){
   final var planDuration = "2Years";
   createPlan(planDuration);
   for(int i = 0; i < 24; i++) 
        createActivity(i+" Months");
   createActivity("1 Year, 1 Month, 1 Day") // Feb 2nd effectively
   simulate(); // A simulation that will post to DB
   assertNotThrows(GetSpanRecords.get(simResults));
   // Further assertions could possibly be made about what the startOffsets looks like
   // but the important assertion is that fetching these results does not throw the IllegalArgumentException
}
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTests.java
mattdailis: I introduced it to try to get around this assertion error:
```
Expected :0 years 0 mons 0 days 0 hours 0 mins -3600.0 secs
Actual   :0 years 0 mons 0 days -1 hours 0 mins 0.0 secs
```

mattdailis: If we hard-code the expected value, we can remove the need for `microsOfPGInterval` here:

```java
assertEquals(new PGInterval("-1 hours"), newOffsetFromPlanStart);
```

The whole test would then become:

```java
@Test
void shouldCalculatePlanDatasetOffsetOnPlanDatasetInsertWithNonNullDatasetId() throws SQLException {
  final var newPlanId = insertPlan(missionModelId, "2020-1-1 01:00:00");

  final var planDatasetInsertRes = connection.createStatement()
                                             .executeQuery(
                                                 """
                                                     INSERT INTO plan_dataset (plan_id, dataset_id)
                                                     VALUES (%s, %s)
                                                     RETURNING *;"""
                                                     .formatted(newPlanId, planDatasetRecord.dataset_id())
                                             );
  planDatasetInsertRes.next();
  final var newOffsetFromPlanStart = new PGInterval(planDatasetInsertRes.getString("offset_from_plan_start"));
  planDatasetInsertRes.close();

  assertEquals(new PGInterval("-1 hours"), newOffsetFromPlanStart);
}
```

A potential downside here is that this test would now assume that the previously inserted plan had a start time of `2020-1-1 00:00:00`. Perhaps a comment stating this assumption would be sufficient.

Mythicaeda: ~~I don't think we need to state that in a comment, since it's part of the `insertPlan` statement (either implicitly if we use the `insertPlan(int)` or explicitly as part of the `final var newPlanId` line.~~

Sorry, misread what you were saying. I'm not against a comment, but I think it's still included if you look at the definition of `planDatasetRecord` (or repeat the plan and dataset insert of the `beforeEach` with locals at the top of this method)

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
mattdailis: The `+00` makes sure that the created plan's start time is interpreted as UTC - without it, I get that the inserted plan's start time is `2020-1-1 00:00:00-08`, which equates to `2020-1-1 08:00:00+00`

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
mattdailis: Hmm, without it I get this test failure, but I'm having trouble rationalizing it at the moment:
```
Expected :2020-01-02 00:00:00+00
Actual   :2020-01-01 16:00:00-08
```

Those two times have more than a timezone difference 🤔 Perhaps the timezone setting has an effect on the logic in `get_approximate_start_time`

mattdailis: This seems relevant, if a bit old: https://stackoverflow.com/a/30878862

Mythicaeda: No they don't. Midnight UTC is 16:00 the day prior ET (24-8=16)

Mythicaeda: Back to the original `connection.prepareStatement` line, it would make most sense for it to go in the `DatabaseTestHelper.startDatabase`, rather than trying to "catch" the side effects of not putting it there in various tests (ie here and in `MerlinDatabaseTests`).

mattdailis: > No they don't. Midnight UTC is 16:00 the day prior ET (24-8=16)

Whew! Thanks for checking me on that - that's a relief

db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
Mythicaeda: Is this still needed with declaring the TZ to UTC?

Mythicaeda (approved):
Look's great! Let's get this thing 🚢

PR #837, opened 2023-03-31
Typecheck Immediate and Hardware Commands

goetzrrGit:
* **Review:** By commit  

## Description
We are currently in discussions with the SeqJSON Spec team to add a `type: immediate` for the `ImmediateCommand` type and `type:hardware` for the `hardwareCommands`. Within the `immediate_command[]` array, the editor is not flagging `C.<Command>`. Within the `hardwareCommands[]` array the editor is not flagging `C.<Command>, <Command>`. In TypeScript, an object literal can be used as a value of an interface if the object has all the same properties, and optionally some more, with compatible types as specified in the interface. However, adding `type: immediate` and  `type: hardware` will resolve this issue. 

Trying to fix this:
Screenshot 2023-03-31 at 9 22 26 AM

- [x] Update Preface
- [ ] Update seq-json-schema 

## Verification
ran the e2e test and manually checked the Sequence Editro

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -177,6 +177,7 @@ declare global {
     // @ts-ignore : 'Args' found in JSON Spec
     args: Args;
     stem: string;
+    type: 'immediate';
 
     public static new<A extends any[] | { [argName: string]: any }>(opts: ImmediateOptions<A>): ImmediateStem<A>;
 
@@ -197,6 +198,7 @@ declare global {
   // @ts-ignore : 'HardwareCommand' found in JSON Spec
   class HardwareStem implements HardwareCommand {
     stem: string;
+    type: 'hardware';
 
     public static new(opts: HardwareOptions): HardwareStem;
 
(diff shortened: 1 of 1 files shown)


dyst5422:
You can probably do this without changing the SeqJson via simulated nominal typing. See here https://www.typescriptlang.org/play#example/nominal-typing

It's essentially what you are doing, but it doesn't actually make the additional field concrete, so it never exists at runtime, just typechecking time.

goetzrrGit:
Hi @dyst5422 let me look into this and see If I can get it working. One thing we changed is the ability to embed SeqJSON into the Sequence object. I am not sure how nominal typing will play with this as we need to support both the custom eDSL and SeqJson. 

Screenshot 2023-03-31 at 9 31 08 AM

mattdailis:
I came across this approach to disallowing extra properties - @dyst5422 do you think it would be viable here, or is it less applicable than I think it is? https://github.com/microsoft/TypeScript/issues/12936#issuecomment-368244671

dyst5422:
@goetzrrGit I had forgotten about that extension to support seqJson directly, it makes the simulated nominal types I suggested not work.

Yes, @mattdailis that's a good approach - specifically here where we KNOW the subset of stems that differentiate the command classes and have the narrow type for the stems being an enumeration of specific stems rather than a more general string.

```
type Exactify<T, X extends T> = T & {
    [K in keyof X]: K extends keyof T ? X[K] : never
}
```
This essentially says "The output type is such that: X must have all properties of T with same types (X extends T) AND for each key K in X ([K in keyof X]), K must be a key in T or not exist (K extends keyof T ? X[K] (it has its own type) : never (it doesn't exist)"

camargo:
Going a different direction here, closing

dyst5422:
@camargo 
> Going a different direction here, closing

Interested in some more info on the different direction if you get the chance

camargo:
@dyst5422 We will be doing this later in the year: https://github.com/NASA-AMMOS/seq-json-schema/issues/15

Issue #840, opened 2023-04-04
Finish MerlinBindingsTests


Mythicaeda:
Currently, `MerlinBindingsTests` does not test any of `MerlinBindings`, only `shouldEnableCors`. Completion of this ticket would involve either expanding this test class or creating end-to-end tests to test MerlinBindings.

PR #841, opened 2023-04-04
Cleanup Merlin Code

### LINKED ISSUE #651: Remove vestigial test classes from merlin-server

Mythicaeda:
* **Tickets addressed:** Closes #651.
* **Review:** By commit  

## Description

Read the commit history for a detailed breakdown of exactly what was removed or changed. Overall, however, this PR:
- Removes all unused classes and most unused methods present in `merlin-server`, `merlin-driver`, and `merlin-worker`. 
- Performs small tweaks to several other classes in those three folders (ie: removing unused imports or inverting a method whose return value is always negated when called)
- Flattened the interface and implementation of interfaces that only had one implementation.
- Removed the InMemoryStore option from the AerieAppDriver. This change was made as these classes are only functional insofar as they will pass tests. For example, since `InMemoryMissionModelRepository` isn't used in testing, nearly all of its methods are stubs.

## Verification

Tests that didn't test live code were removed, as well as tests that have been disabled since November 2021.

#840 was created to address a gap I noticed in `MerlinBindingsTests`.

#871 was created to address a gap noticed in `EventGraphEvaluator`.

## Documentation

No documentation changes should be necessary.

## Future work

Most of the `InMemory_` code is only used in tests of `InMemory` code, rather than being used to avoid having unit tests of things like the `SimulationDriver` turn into e2eTests. I'd like to take a finer tooth comb and see if we can either drop these classes entirely or actually use them in tests of live Merlin code.


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/timeline/RecursiveEventGraphEvaluator.java
```diff
@@ -4,8 +4,7 @@
 
 import java.util.Optional;
 
-public final class RecursiveEventGraphEvaluator implements EventGraphEvaluator {
-  @Override
+public final class RecursiveEventGraphEvaluator {
```
mattdailis: I have hit stack overflows with the recursive event graph evaluator before 🤔 It happened to be due to a bug in the mission model, but I suspect we may want to have the option to switch to the iterative version

Mythicaeda: Out of curiosity, could you quickly go over how we'd do that swap, assuming we'd need to? Would it be a matter of changing the code in `MissionModelBuilder`? If so, would doing the change there impact any existing models (I'm assuming not because it's in the driver, not the sdk)?

mattdailis: A few approaches come to mind - the simplest being an environment variable that can be set per worker or per deployment.

We could do something fancy and set a graph depth threshold - use the recursive variant for graphs up to a certain size, and the iterative variant otherwise. Similarly, we could catch a stack overflow and retry with the iterative variant.

As you mention, we could alternatively expose this as a knob to the mission model directly - this could be a backwards compatible change if that's a design constraint - we could use the recursive version by default, and allow the mission modeler to opt in to the iterative variant. This could be done at a whole-simulation level, or we could try to set it per topic (though as it stands we mix topics in a single event graph, so that may not be straightforward)

Mythicaeda: Resolving as the relevant commit has been dropped, but this discussion may be useful if we have to determine if we want to keep or scrap `IterativeEventGraphEvaluator` in #871.

camargo (approved):
LGTM based on your overview yesterday. Let's get sign off from @mattdailis as well.

mattdailis (approved):

PR #843, opened 2023-04-04
Added time ordering with respect to commands in expanded sequences

### LINKED ISSUE #828: Time order commands in expanded sequences
When running command expansion where **multiple** activities target a **single** sequence, we need to time-order sort commands for that sequence. Currently we do no time ordering for any expanded sequences. Clipper wants time ordered commands in this special case. 

We only time order commands if **all** the commands in the target sequence have relative or absolute time. We will not time order the sequence if it contains complete or epoch-relative commands (even if it only has one).

cohansen:
* **Tickets addressed:** Closes #828 
* **Review:** By commit

## Description
Consulted with @camargo on how the times work which I previously didn't fully understand and read the ticket.

## Verification
Added some unit tests to check sorting and non-sorting cases.

## Documentation
NA

## Future work
Depends on how we want to support this for other missions, for now all the code just lives in the `defaultSeqBuilder`.

### DIFF
--- a/sequencing-server/src/defaultSeqBuilder.ts
+++ b/sequencing-server/src/defaultSeqBuilder.ts
@@ -48,7 +48,7 @@ export const defaultSeqBuilder: SeqBuilder = (
 // Order commands by time in expanded sequences.
 function sortCommandsByTime(commands: CommandStem<{} | []>[]): CommandStem<{} | []>[] {
   const relativeTimeTracker: Record<string, Temporal.Instant> = {};
-  let previousAbsoluteTimestamp: Temporal.Instant;
+  let previousAbsoluteTimestamp: Temporal.Instant | undefined = undefined;
 
   for (const command of commands) {
     // If the command is epoch-relative, complete, or the first command and relative then short circuit and don't try and sort.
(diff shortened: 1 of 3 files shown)


sequencing-server/src/defaultSeqBuilder.ts
```diff
@@ -42,3 +44,44 @@ export const defaultSeqBuilder: SeqBuilder = (
...
+    if (
+      command.epochTime ||
+      (!command.absoluteTime && !command.epochTime && !command.relativeTime) ||
+      (command.relativeTime && commands.indexOf(command) === 0)
+    ) {
+      return commands;
+    }
+
+    // Keep track of the previously seen absolute time so we can convert the next relative timestamp we come across.
+    if (command.absoluteTime) {
```
dyst5422: These kind of auto-cast checks aren't great because they have a lot of loopholes that are hard to keep track of (0 is falsy, empty string is falsy, etc etc). We should do explicit comparisons to undefined. The explicit checks should make the unsafe non-null assertions unnecessary as well.

sequencing-server/src/defaultSeqBuilder.ts
```diff
@@ -42,3 +44,44 @@ export const defaultSeqBuilder: SeqBuilder = (
...
+        command.relativeTime!,
+      );
+    }
+  }
+
+  // The commands will either be absolute or relative at this point, so we just compare their times.
+  commands.sort((a, b) => {
+    const firstCommandTime = a.absoluteTime
+      ? a.absoluteTime
+      : relativeTimeTracker[a.stem + a.relativeTime?.toString()]!;
```
dyst5422: The non-null assertions here aren't safe. We should do an explicit checks for them and throw appropriate errors. We have to assume that users will supply "bad" sequences w.r.t. time values.

camargo: 🥷

```ts
const firstCommandTime = a.absoluteTime ?? relativeTimeTracker[a.stem + a.relativeTime?.toString()]!;
```

camargo (approved):
Tested locally, nice! LGTM

sequencing-server/src/defaultSeqBuilder.ts
```diff
@@ -42,3 +44,43 @@ export const defaultSeqBuilder: SeqBuilder = (
...
+// Order commands by time in expanded sequences.
+function sortCommandsByTime(commands: CommandStem<{} | []>[]): CommandStem<{} | []>[] {
+  const relativeTimeTracker: Record<string, Temporal.Instant> = {};
+  let previousAbsoluteTimestamp: Temporal.Instant | undefined = undefined;
+
+  for (const command of commands) {
+    // If the command is epoch-relative, complete, or the first command and relative then short circuit and don't try and sort.
+    if (
+      command.epochTime ||
+      (!command.absoluteTime && !command.epochTime && !command.relativeTime) ||
```
dyst5422: Additional casts to booleans here

sequencing-server/src/defaultSeqBuilder.ts
```diff
@@ -42,3 +44,43 @@ export const defaultSeqBuilder: SeqBuilder = (
...
+// Order commands by time in expanded sequences.
+function sortCommandsByTime(commands: CommandStem<{} | []>[]): CommandStem<{} | []>[] {
+  const relativeTimeTracker: Record<string, Temporal.Instant> = {};
+  let previousAbsoluteTimestamp: Temporal.Instant | null = null;
+
+  for (const command of commands) {
+    // If the command is epoch-relative, complete, or the first command and relative then short circuit and don't try and sort.
+    if (
+      command.epochTime !== null ||
+      (command.absoluteTime === null && command.epochTime === null && command.relativeTime === null) ||
```
camargo, line 57: Are these explicitly set to `null` if they do not exist?

cohansen: Yeah!
https://github.com/NASA-AMMOS/aerie/blob/feature/time-order-commands/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts#L614-L616

goetzrrGit (approved):
Looks good

PR #845, opened 2023-04-04
Update Aerie_DB_Migration

### LINKED ISSUE #579: Modify aerie_db_migration.py to use new migration schema
Currently `aerie_db_migration.py` uses the Hasura CLI to track migrations. Since we are changing to tracking which migrations have been applied in the DB itself, this script needs to be updated to reflect that change. Additionally, it needs to be able to offer and then execute `prepare.sql`s should they be included with a migration (example: `1_make_scheduling_goals_unique`  in #521)

Mythicaeda:
* **Tickets addressed:** Closes #579.
* **Review:** By commit  

## Description

After running the migrations but prior to applying them, the script now queries the `migration.schema_migrations` table directly using [`psycopg3`](https://www.psycopg.org/psycopg3/docs/index.html) to get the current version. It then marks that in Hasura. If the migrations table does not exist, then the script assumes that the database is on a version before 0.

The PGCMP workflow was updated to reflect the changes to `aerie_db_migration.py`.

## Verification

Ran some migrations up and down manually, as well as running "revert all down" and "apply all up" on a db from the latest schema.
Manually ran the PGCMP workflow.

## Documentation

Docs Ticket has been created [here](https://github.com/NASA-AMMOS/aerie-docs/issues/31).


deployment/aerie_db_migration.py
```diff
@@ -0,0 +1,301 @@
+#!/usr/bin/env python3
+"""Migrate an AERIE Database"""
+
+import os
+import argparse
+import sys
+import shutil
+import subprocess
+import psycopg
```
camargo, line 9: Since we have this dependency now can we include a [requirements.txt](https://pip.pypa.io/en/stable/cli/pip_freeze/#examples) with this script to make the dependencies easier to download? We can note in the docs how to install the dependencies with Pip using requirements.txt.

deployment/aerie_db_migration.py
```diff
@@ -0,0 +1,301 @@
...
+    sys.exit(f'Hasura CLI is not installed. Exiting...')
+  else:
+    os.system('hasura version')
+
+  # Get the Username/Password
+  username = ""
+  password = ""
+  usernameFound = False
+  passwordFound = False
+  with open(args.env_path) as envFile:
```
camargo, line 246: Not for this PR, but just noting we should probably use something like [python-dotenv](https://github.com/theskumar/python-dotenv) here so we don't have to parse the environment manually.

camargo (approved):
👍

deployment/aerie_db_migration.py
Mythicaeda: `requirements.txt` has been added and a needing to document how to install the dependencies has been added to the docs ticket.

PR #847, opened 2023-04-05
Allow for separation of activity matching from activity creation in scheduling eDSL

### LINKED ISSUE #751: Add activity matching on type only in the scheduling edsl
In the scheduling edsl, there is currently no distinction between activity matching (how an existing activity can satisfy the goal) and activity creation (what new activity we are creating to satisfy the goal). 

Ideally, we'd like to have these two distinct capabilities with advanced features such as search on parameter ranges (rather than exact values) or parameter subsets (matching on only a subset of the given parameters). 

As a first easy step, we can add a boolean parameter to the activity creation template stating that the scheduler should match on any activity of same given type without having to match any parameter value.

### LINKED ISSUE #782: Filter activity expressions based on parameter values
srschaff
https://jira.jpl.nasa.gov/browse/AERIEQS-346
Moderate
I would like to express a coexistence goal anchored not just on all instances of an anchor activity type, but only those that have also have specific parameter value.

Something like:
{color:#008080}ActivityExpression{color}{color:#000000}.ofType({color}{color:#008080}ActivityTypes{color}{color:#000000}.{color}{color:#008080}MyAct{color}{color:#000000}).withParameters([\\{ key,"value"}]){color}

adrienmaillard:
Resolves #751 
Resolves #782 

* **Review:** By commit  

## Description

In this PR, we provide an optional `activityFinder` for each eDSL goal. If passed, it is used for activity matching, replacing the activity template.

This `activityFinder` is an `ActivityExpression` that can match on activity type and also parameters. Users may only define a subset of the parameters they are looking for and the expression will match all activities that have a superset of these defined parameters. 

Let's say the parameters of an activity X looks like that: 
```
A : "1"
B : { C: "2" D: "3}
E : "4"
```
Then if the user specifies the following parameters in the activity finder :
`{ A:"1"}` will match X
`{A : "1", B : { C: "2"}}` will also match X 
`{A : "1", B : { C: "4"}}` will not match X

Edit:
The first commit also returns the full stacktrace to our backend if a compilation error happens when running user code in the ts-code-runner, in both constraints and scheduling edsl compilers.

## Verification

New tests 

## Documentation

- [x] I have to update the scheduling eDSL doc 

## Future work

More operators on `ActivityExpression` other than parameter equality

### DIFF
--- a/merlin-server/constraints-dsl-compiler/src/main.ts
+++ b/merlin-server/constraints-dsl-compiler/src/main.ts
@@ -39,6 +39,19 @@ interface AstNode {
   __astNode: object
 }
 
+function toJson(unwrappedErr: any){
+  var completeStackValue = "";
+  if ('error' in unwrappedErr && 'stack' in unwrappedErr.error){
+    completeStackValue = JSON.stringify(unwrappedErr.error.stack)
+  }
+  return {
+    message: unwrappedErr.message,
+    stack: unwrappedErr.stack,
+    location: unwrappedErr.location,
+    completeStack:  completeStackValue,
+  }
+}
+
 async function handleRequest(data: Buffer) {
   try {
     // Test the health of the service by responding to "ping" with "pong".
@@ -73,8 +86,9 @@ async function handleRequest(data: Buffer) {
     );
 
     if (result.isErr()) {
+      const secondLine = JSON.stringify(result.unwrapErr().map(err => toJson(err))) + '\n';
       process.stdout.write('error\n')
-      process.stdout.write(JSON.stringify(result.unwrapErr().map(err => err.toJSON())) + '\n');
+      process.stdout.write(secondLine);
       lineReader.once('line', handleRequest);
       return;
     }
--- a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
+++ b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
@@ -15,7 +15,7 @@ class TypescriptCodeGenerationServiceTest {
   @Test
   void testCodeGen() throws MissionModelService.NoSuchMissionModelException, NoSuchPlanException {
     final var codeGenService = new TypescriptCodeGenerationServiceAdapter(new StubMissionModelService(), new StubPlanService());
-
+    final var expected = codeGenService.generateTypescriptTypes("abc", Optional.of(new PlanId(1L)));
     assertEquals(
         """
             /** Start Codegen */
@@ -71,6 +71,18 @@ class TypescriptCodeGenerationServiceTest {
               [ActivityType.activity2]:ParameterTypeactivity2,
               [ActivityType.activity]:ParameterTypeactivity,
             };
+            export type ParameterTypeWithUndefinedactivity2 = {
+              Param?: (( | "hello" | "there") | Discrete<( | "hello" | "there")> | undefined),
+            }
+            export type ParameterTypeWithUndefinedactivity = {
+              Param?: (string | Discrete<string> | undefined),
+              AnotherParam?: (number | Real | undefined),
+              Duration?: (Temporal.Duration | Discrete<Temporal.Duration> | undefined),
+            }
+            export type ActivityTypeParameterMapWithUndefined = {
+              [ActivityType.activity2]:ParameterTypeWithUndefinedactivity2,
+              [ActivityType.activity]:ParameterTypeWithUndefinedactivity,
+            };
             declare global {
               enum ActivityType {
                 activity2 = "activity2",
@@ -81,7 +93,7 @@ class TypescriptCodeGenerationServiceTest {
               ActivityType
             });
             /** End Codegen */""",
-        codeGenService.generateTypescriptTypes("abc", Optional.of(new PlanId(1L)))
+        expected
     );
   }
 }
--- a/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
+++ b/merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
@@ -16,7 +16,7 @@ class TypescriptCodeGenerationServiceTest {
   void testCodeGen() throws MissionModelService.NoSuchMissionModelException, NoSuchPlanException {
     final var codeGenService = new TypescriptCodeGenerationServiceAdapter(new StubMissionModelService(), new StubPlanService());
     final var expected = codeGenService.generateTypescriptTypes("abc", Optional.of(new PlanId(1L)));
-    assertEquals(
+    assertEquals(expected,
         """
             /** Start Codegen */
             import * as AST from './constraints-ast.js';
@@ -92,8 +92,7 @@ class TypescriptCodeGenerationServiceTest {
             Object.assign(globalThis, {
               ActivityType
             });
-            /** End Codegen */""",
-        expected
+            /** End Codegen */"""
     );
   }
 }
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
@@ -19,7 +19,9 @@ import gov.nasa.jpl.aerie.scheduler.model.ActivityType;
 import gov.nasa.jpl.aerie.scheduler.NotNull;
 import gov.nasa.jpl.aerie.scheduler.Nullable;
 
+import java.math.BigDecimal;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -580,14 +582,7 @@ public class ActivityExpression implements Expression<Spans> {
       Map<String, SerializedValue> actInstanceArguments = act.parameters;
       final var instantiatedArguments = SchedulingActivityDirective
                 .instantiateArguments(arguments, act.interval.start, simulationResults, evaluationEnvironment, type);
-      for (var param : instantiatedArguments.entrySet()) {
-        if (actInstanceArguments.containsKey(param.getKey())) {
-          match = actInstanceArguments.get(param.getKey()).equals(param.getValue());
-        }
-        if (!match) {
-          break;
-        }
-      }
+      match = subsetOrEqual(SerializedValue.of(actInstanceArguments), SerializedValue.of(instantiatedArguments));
     }
     return match;
   }
@@ -613,5 +608,87 @@ public class ActivityExpression implements Expression<Spans> {
 
   @Override
   public void extractResources(final Set<String> names) { }
+
+
+  /**
+   * Evaluates whether a SerializedValue can be qualified as the subset of another SerializedValue or not
+    * @param superset the proposed superset
+   * @param subset the proposed subset
+   * @return true if subset is a subset of superset
+   */
+  public static boolean subsetOrEqual(SerializedValue superset, SerializedValue subset){
+    Objects.requireNonNull(superset);
+    Objects.requireNonNull(subset);
+    final var visitor = new SerializedValue.Visitor<Boolean>(){
+      @Override
+      public Boolean onNull() {
+        return true;
+      }
+
+      @Override
+      public Boolean onNumeric(final BigDecimal value) {
+        final var argumentsAsNumeric = superset.asNumeric();
+        if(argumentsAsNumeric.isEmpty()){
+          return false;
+        }
+        return argumentsAsNumeric.get().equals(value);
+      }
+
+      @Override
+      public Boolean onBoolean(final boolean value) {
+        final var argumentsAsBoolean = superset.asBoolean();
+        if(argumentsAsBoolean.isEmpty()){
+          return false;
+        }
+        return argumentsAsBoolean.get().equals(value);
+      }
+
+      @Override
+      public Boolean onString(final String value) {
+        final var argumentsAsString = superset.asString();
+        if(argumentsAsString.isEmpty()){
+          return false;
+        }
+        return argumentsAsString.get().equals(value);
+      }
+
+      @Override
+      public Boolean onMap(final Map<String, SerializedValue> value) {
+        final var argumentsAsMap = superset.asMap();
+        if(argumentsAsMap.isEmpty()){
+          return false;
+        }
+        for(final var elementInPattern: value.entrySet()){
+          final var elementInArguments = argumentsAsMap.get().get(elementInPattern.getKey());
+          if(elementInArguments != null){
+            if(!subsetOrEqual(elementInArguments, elementInPattern.getValue())){
+              return false;
+            }
+          } else {
+            return false;
+          }
+        }
+        return true;
+      }
+
+      @Override
+      public Boolean onList(final List<SerializedValue> value) {
+        final var argumentsAsListOptional = superset.asList();
+        if(argumentsAsListOptional.isEmpty()){
+          return false;
+        }
+        if(argumentsAsListOptional.get().size() < value.size()){
+          return false;
+        }
+        for(int i = 0; i < value.size(); i++){
+          if(!subsetOrEqual(argumentsAsListOptional.get().get(i), value.get(i))){
+            return false;
+          }
+        }
+        return true;
+      }
+    };
+    return subset.match(visitor);
+  }
 
 }
--- a/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
+++ b/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
@@ -1,0 +1,102 @@
+package gov.nasa.jpl.aerie.scheduler;
+
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import org.junit.jupiter.api.Test;
+
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityExpression.subsetOrEqual;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class SerializedValueSubsetTest {
+
+  @Test
+  public void testSubsetEqualityMap(){
+    final var reference = SerializedValue.of(
+        Map.of("A", SerializedValue.of(8.0),
+               "B", SerializedValue.of(
+                Map.of("C", SerializedValue.of(14), "D", SerializedValue.of(15))
+            )));
+
+    final var subsetMap = SerializedValue.of(
+        Map.of(
+            "B", SerializedValue.of(
+                Map.of("D", SerializedValue.of(15))
+            )));
+    final var subsetMapWithSerializedNull = SerializedValue.of(
+        Map.of(
+            "B", SerializedValue.of(
+                Map.of("D", SerializedValue.NULL)
+            )));
+    final var notSubsetMap = SerializedValue.of(
+        Map.of(
+            "B", SerializedValue.of(
+                Map.of("D", SerializedValue.of(7))
+            )));
+
+    assertTrue(subsetOrEqual(reference, subsetMap));
+    assertFalse(subsetOrEqual(reference, notSubsetMap));
+    assertTrue(subsetOrEqual(reference, subsetMapWithSerializedNull));
+
+  }
+
+  @Test
+  public void testSubsetEqualityList(){
+    final var reference = SerializedValue.of(List.of(SerializedValue.of(8.0), SerializedValue.of(6.0), SerializedValue.of(9.0)));
+    final var subsetList = SerializedValue.of(List.of(SerializedValue.of(8.0),SerializedValue.of(6.0)));
+    final var nullList = SerializedValue.of(List.of(SerializedValue.NULL));
+    final var notSubsetList = SerializedValue.of(List.of(SerializedValue.of(8.0),SerializedValue.of(7.0)));
+    final var wrongOrderList = SerializedValue.of(List.of(SerializedValue.of(9.0), SerializedValue.of(6.0), SerializedValue.of(8.0)));
+    assertTrue(subsetOrEqual(reference, subsetList));
+    assertTrue(subsetOrEqual(reference, nullList));
+    assertFalse(subsetOrEqual(reference, notSubsetList));
+    assertFalse(subsetOrEqual(reference, wrongOrderList));
+  }
+  @Test
+  public void testSubsetEqualityValuesInteger(){
+    final var reference = SerializedValue.of(1111);
+    final var sameValue = SerializedValue.of(1111);
+    final var differentValue = SerializedValue.of(1112);
+    assertTrue(subsetOrEqual(reference, sameValue));
+    assertFalse(subsetOrEqual(reference, differentValue));
+  }
+
+  @Test
+  public void testSubsetEqualityValuesBoolean(){
+    final var reference = SerializedValue.of(true);
+    final var sameValue = SerializedValue.of(true);
+    final var differentValue = SerializedValue.of(false);
+    final var differentType = SerializedValue.of(11);
+    assertTrue(subsetOrEqual(reference, sameValue));
+    assertFalse(subsetOrEqual(reference, differentValue));
+    assertFalse(subsetOrEqual(reference, differentType));
+  }
+
+  @Test
+  public void testSubsetEqualityValuesDecimal(){
+    final var reference = SerializedValue.of(BigDecimal.valueOf(122.56));
+    final var sameValue = SerializedValue.of(BigDecimal.valueOf(122.56));
+    final var differentValue = SerializedValue.of(BigDecimal.valueOf(122.57));
+    final var differentType = SerializedValue.of(11);
+    assertTrue(subsetOrEqual(reference, sameValue));
+    assertFalse(subsetOrEqual(reference, differentValue));
+    assertFalse(subsetOrEqual(reference, differentType));
+  }
+
+  @Test
+  public void testSubsetEqualityValuesString(){
+    final var reference = SerializedValue.of("AbCd");
+    final var sameValue = SerializedValue.of("AbCd");
+    final var differentValue = SerializedValue.of("bCd");
+    final var differentType = SerializedValue.of(11);
+    final var differentCase = SerializedValue.of("abcd");
+    assertTrue(subsetOrEqual(reference, sameValue));
+    assertFalse(subsetOrEqual(reference, differentValue));
+    assertFalse(subsetOrEqual(reference, differentCase));
+    assertFalse(subsetOrEqual(reference, differentType));
+  }
+}
--- a/scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -9,6 +9,7 @@ import * as AST from "./scheduler-ast.js";
 import  * as WindowsEDSL from "./constraints-edsl-fluent-api.js";
 import {ActivityInstance} from "./constraints-edsl-fluent-api.js";
 import * as ConstraintsAST from "./constraints-ast.js";
+import {makeArgumentsDiscreteProfiles} from "./scheduler-mission-model-generated-code";
 
 type WindowProperty = AST.WindowProperty
 type TimingConstraintOperator = AST.TimingConstraintOperator
@@ -268,10 +269,14 @@ export class Goal {
    *
    * @param opts an object containing the activity template and the interval at which the activities must be placed
    */
-  public static ActivityRecurrenceGoal <A extends WindowsEDSL.Gen.ActivityType>(opts: { activityTemplate: ActivityTemplate<A>, interval: Temporal.Duration }): Goal {
+  public static ActivityRecurrenceGoal <A extends WindowsEDSL.Gen.ActivityType, B extends WindowsEDSL.Gen.ActivityType>(opts: {
+    activityTemplate: ActivityTemplate<A>,
+    interval: Temporal.Duration,
+    activityFinder?: ActivityExpression<B>}): Goal {
     return Goal.new({
       kind: AST.NodeKind.ActivityRecurrenceGoal,
       activityTemplate: opts.activityTemplate,
+      activityFinder: opts.activityFinder?.__astNode,
       interval: opts.interval,
       shouldRollbackIfUnsatisfied: false
     });
@@ -346,12 +351,14 @@ export class Goal {
    * @param opts an object containing the activity template, a set of windows, and optionally temporal constraints.
    */
 
-  public static CoexistenceGoal<T extends WindowsEDSL.Gen.ActivityType, S extends WindowsEDSL.Gen.ActivityType>(opts: ({
+  public static CoexistenceGoal<T extends WindowsEDSL.Gen.ActivityType, S extends WindowsEDSL.Gen.ActivityType,  B extends WindowsEDSL.Gen.ActivityType>(opts: ({
     activityTemplate: (( interval: WindowsEDSL.Interval ) => ActivityTemplate<S>) | ActivityTemplate<S>,
     forEach:  WindowsEDSL.Windows,
+    activityFinder?: ActivityExpression<B>
   } | {
     activityTemplate: (( span: ActivityInstance<T> ) => ActivityTemplate<S>) | ActivityTemplate<S>,
     forEach:  ActivityExpression<T>,
+    activityFinder?: ActivityExpression<B>
   }) & CoexistenceGoalTimingConstraints): Goal {
 
     let alias: string;
@@ -380,8 +387,9 @@ export class Goal {
 
     return Goal.new({
       kind: AST.NodeKind.ActivityCoexistenceGoal,
-      alias,
-      activityTemplate,
+      alias: alias,
+      activityTemplate: activityTemplate,
+      activityFinder: opts.activityFinder?.__astNode,
       forEach: opts.forEach.__astNode,
       startConstraint: (("startsAt" in opts) ? opts.startsAt.__astNode : ("startsWithin" in opts) ? opts.startsWithin.__astNode : undefined),
       endConstraint: (("endsAt" in opts) ? opts.endsAt.__astNode : ("endsWithin" in opts) ? opts.endsWithin.__astNode : undefined),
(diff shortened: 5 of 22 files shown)


Mythicaeda:
> This activityFinder is an ActivityExpression that can match on activity type and also parameters. Users may only define a subset of the parameters they are looking for and the expression will match all activities that have a superset of these defined parameters.

Can a user define only a type and no parameters? Or do they always need to define parameters?

adrienmaillard:
> > This activityFinder is an ActivityExpression that can match on activity type and also parameters. Users may only define a subset of the parameters they are looking for and the expression will match all activities that have a superset of these defined parameters.
> 
> Can a user define only a type and no parameters? Or do they always need to define parameters?

Type-only expression works, yes.

merlin-server/constraints-dsl-compiler/src/main.ts
```diff
@@ -73,8 +86,9 @@ async function handleRequest(data: Buffer) {
     );
 
     if (result.isErr()) {
+      const secondLine = JSON.stringify(result.unwrapErr().map(err => toJson(err))) + '\n';
       process.stdout.write('error\n')
-      process.stdout.write(JSON.stringify(result.unwrapErr().map(err => err.toJSON())) + '\n');
+      process.stdout.write(secondLine);
```
Mythicaeda, line 91: Why was `JSON.stringify(result.unwrapErr().map(err => toJson(err))) + '\n'` pulled out into a variable?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
   @Override
   public void extractResources(final Set<String> names) { }
 
+  //true if second is a subset or equal to first
+  public static boolean subsetOrEqual(SerializedValue first, SerializedValue second){
```
Mythicaeda: The names `first` and `second` aren't clear. Are these from a specific naming convention? If not, I'd recommend replacing `first` with something like `arguments` or `input` and `second` with `pattern` or `template`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
   @Override
   public void extractResources(final Set<String> names) { }
 
+  //true if second is a subset or equal to first
```
Mythicaeda: Minor: Can this comment be replaced with a JavaDoc comment?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
...
+        return firstAsBoolean.get().equals(value);
+      }
+
+      @Override
+      public Boolean onString(final String value) {
+        final var firstAsString = first.asString();
+        if(firstAsString.isEmpty()){
+          return false;
+        }
+        return firstAsString.get().equals(value);
```
Mythicaeda: This will perform a case-sensitive comparison. Do we want that?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
...
+        }
+        for(final var elementInSecond: value.entrySet()){
+          final var elementInFirst = firstAsMap.get().get(elementInSecond.getKey());
+          if(elementInFirst != null){
+            if(!subsetOrEqual(elementInFirst, elementInSecond.getValue())){
+              return false;
+            }
+          }
+        }
+        return true;
```
Mythicaeda: If there's an element in `second` that's not in `first`, then first is not a subset of second and this method should return `false`, correct? Currently that case is ignored.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
...
+      }
+
+      @Override
+      public Boolean onList(final List<SerializedValue> value) {
+        final var firstAsList = first.asList();
+        if(firstAsList.isEmpty()){
```
Mythicaeda: Minor, feel free to ignore: Since `List` and `Optional` both have an `isEmpty()` method, it might be useful to either specify the type (`Optional<List<SerializedValue>>`) or modify the variable name `firstAsList` to make it clear that its type is an Optional.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
...
+          return false;
+        }
+        if(firstAsList.get().size() < value.size()){
+          return false;
+        }
+        for(int i = 0; i < value.size(); i++){
+          if(!subsetOrEqual(firstAsList.get().get(i), value.get(i))){
+            return false;
+          }
```
Mythicaeda: Are the values of the two Lists guaranteed to be ordered the same?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +616,76 @@ public String prettyPrint(final String prefix) {
...
+
+      @Override
+      public Boolean onMap(final Map<String, SerializedValue> value) {
+        final var firstAsMap = first.asMap();
+        if(firstAsMap.isEmpty()){
+          return false;
+        }
+        for(final var elementInSecond: value.entrySet()){
+          final var elementInFirst = firstAsMap.get().get(elementInSecond.getKey());
+          if(elementInFirst != null){
```
Mythicaeda: Is it possible for `elementInFirst` to be `null` even if the key is present?

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
```diff
@@ -0,0 +1,80 @@
...
+  public void testSubsetEqualityValues3(){
+    final var s1 = SerializedValue.of(BigDecimal.valueOf(122.56));
+    final var s2True = SerializedValue.of(BigDecimal.valueOf(122.56));
+    final var s2False = SerializedValue.of(BigDecimal.valueOf(122.57));
+    final var s2False2 = SerializedValue.of(11);
+    assertTrue(subsetOrEqual(s1, s2True));
+    assertFalse(subsetOrEqual(s1, s2False));
+    assertFalse(subsetOrEqual(s1, s2False2));
+  }
+}
```
Mythicaeda: Rather than being `testSubsetEqualityValues` 1, 2 and 3, can these tests be renamed to be more accurate (ie `testSubsetEqualityValuesInteger`, `testSubsetEqualityValuesBoolean`, `testSubsetEqualityValuesDecimal`?

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
Mythicaeda, line 1: Two things with these tests:
1. (Moderate) Can the variables please be more meaningfully named than s1, s2, s22, etc?
2. (Major) Tests for Strings are missing.

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -268,10 +269,14 @@ export class Goal {
    *
    * @param opts an object containing the activity template and the interval at which the activities must be placed
    */
-  public static ActivityRecurrenceGoal <A extends WindowsEDSL.Gen.ActivityType>(opts: { activityTemplate: ActivityTemplate<A>, interval: Temporal.Duration }): Goal {
+  public static ActivityRecurrenceGoal <A extends WindowsEDSL.Gen.ActivityType, B extends WindowsEDSL.Gen.ActivityType>(opts: {
+    activityTemplate: ActivityTemplate<A>,
+    interval: Temporal.Duration,
+    activityFinder?: ActivityExpression<B>}): Goal {
```
Mythicaeda, line 293: Why can't we reuse A's type when creating `activityFinder`?

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
Mythicaeda, line 1: Reminder that the docs in this file need to be updated.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -1144,7 +1339,7 @@ private SchedulingRunResults runScheduler(
   )
   {
     final var activities = new HashMap<ActivityDirectiveId, ActivityDirective>();
-    long id = 0L;
+    long id = 1;
```
Mythicaeda, line 1389: Why was the value of this variable changed?

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -305,6 +305,201 @@ export default () => Goal.CoexistenceGoal({
...
+    assertTrue(goalResult.satisfied());
+    assertEquals(0, goalResult.createdActivities().size());
+    assertEquals(1, goalResult.satisfyingActivities().size());
+    for (final var activity : goalResult.satisfyingActivities()) {
+      assertNotNull(activity);
+    }
+  }
+
+  @Test
+  void testCoexistencePartialAct2() {
```
Mythicaeda: Can this test be renamed so that it's clear what case this one is testing vs `testCoexistancePartialAct`?

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -305,6 +305,201 @@ export default () => Goal.CoexistenceGoal({
...
+      assertNotNull(activity);
+    }
+    assertEquals(2, goalResult.satisfyingActivities().size());
+    for (final var activity : goalResult.satisfyingActivities()) {
+      assertNotNull(activity);
+    }
+
+    final var planByActivityType = partitionByActivityType(results.updatedPlan());
+    final var growBananas = planByActivityType.get("GrowBanana");
+    assertEquals(3, growBananas.size());
```
Mythicaeda: I'm not certain that the assertions here are sufficient to prove that the behavior is correct. 

Correct Behavior: 
- Added Activity was a GrowBanana placed at 10 minutes with the arguments `(quantity: 1, growingDuration: 1 Minute)`
- Already Existing Satisfied Activity is the GrowBanana at 5 minutes with the arguments `(quantity: 1, growingDuration: 1 Minute)` [Side note: the Satisfied Activity should have a different growingDuration than the creationTemplate]

Incorrect Behavior that would pass this test:
- Added Activity was a GrowBanana placed at 5 minutes with any arguments
- Already Existing Satisfied Acitivity is the GrowBanana at 10 minutes with the arguments `(quantity: 2, growingDuration: 1 Minute)`

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -305,6 +305,201 @@ export default () => Goal.CoexistenceGoal({
...
+                interval: Temporal.Duration.from({ hours : 24 })
+              })""", true)
+        ),
+        PLANNING_HORIZON
+    );
+
+    final var goalResult = results.scheduleResults.goalResults().get(new GoalId(0L));
+    assertEquals(2, goalResult.createdActivities().size());
+    assertEquals(PLANNING_HORIZON.getEndAerie().dividedBy(24,HOURS), goalResult.satisfyingActivities().size());
+    assertTrue(goalResult.satisfied());
```
Mythicaeda, line 498: While this test proves that 2 activities were added and 2 were already considered as satisfying the goal, how does it prove that specifically the first two activity directives were correctly matched and not, for example, the first and the third?

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -177,6 +175,42 @@ export default function myGoal() {
...
+        false);
+    if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success<SchedulingDSL.GoalSpecifier> r) {
+      assertEquals(expectedGoalDefinition, r.value());
+    } else if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Error<SchedulingDSL.GoalSpecifier> r) {
+      fail(r.toString());
+    }
+  }
+
+  @Test
+  void  testSchedulingDSL_recurrence_activityFinder()
```
Mythicaeda:
```suggestion
  void testSchedulingDSL_recurrence_activityFinder()
```

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
```diff
@@ -15,7 +15,7 @@ class TypescriptCodeGenerationServiceTest {
   @Test
   void testCodeGen() throws MissionModelService.NoSuchMissionModelException, NoSuchPlanException {
     final var codeGenService = new TypescriptCodeGenerationServiceAdapter(new StubMissionModelService(), new StubPlanService());
-
+    final var expected = codeGenService.generateTypescriptTypes("abc", Optional.of(new PlanId(1L)));
     assertEquals(
```
Mythicaeda: Minor: `expected` should be the first parameter in this `assertEquals`.

Mythicaeda (commented):
Looks good so far!

merlin-server/constraints-dsl-compiler/src/main.ts
adrienmaillard: If an error happened during the unwrapping, it would print `error` and then go to the `panic` mode (in the catch error block below). Now it goes straight to `panic` and it is clearer in the output.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: Replaced with arguments and pattern.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: ok

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I think we do. Why not ?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I think you're right, but you mixed up which one should be the subset of the other. We are testing whether second is the subset of first. 

here's how it looks now 
```
for(final var elementInPattern: value.entrySet()){
          final var elementInArguments = argumentsAsMap.get().get(elementInPattern.getKey());
          if(elementInArguments != null){
            if(!subsetOrEqual(elementInArguments, elementInPattern.getValue())){
              return false;
            }
          } else {
            return false;
          }
        }
        return true;
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: done

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I think we can consider that a list is an ordered data structure.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: As you said, there might be elements in second that are not in first.

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
adrienmaillard: I guess we could have different types there. Validate with one and insert another... like supertypes/subtypes.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: No idea I'll revert.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Done

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: because if the input acts start at 0, their opposite id will be used in the scheduler and the opposite of 0 is 0, that is the same as the first id that will be used by the scheduler. I am surprised but this shows up only when running the tests individually somehow.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: OK I have updated to look specifically for these acts.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: updated to look specifically for these acts.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
adrienmaillard: ok

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/TypescriptCodeGenerationServiceTest.java
adrienmaillard: ok

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
Mythicaeda: I worry that it will lead to confusion/frustration, especially since the comparison is against user input. Ie: Let's say we that the pattern here is `{ producer: "Dole" }`, and the user has a `ChangeProducer` activity whose `producer` argument is `"dole"`. Would they expect their ChangeProducer to be considered as satisfying the pattern? If so, then we need to do a case-insensitive comparison.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
Mythicaeda: Could you briefly explain why that is the case here? I just wanna be confident that this method won't return `false` unexpectedly.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
Mythicaeda: No, what I'm asking is if there could be a key-value pair in the Map that is `someKey: null`. If so, then this check shouldn't be `if(elementInFirst != null)`, it should be `if(!firstAsMap.containsKey(elementInSecond.getKey()))`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
mattdailis: In general, all our equality checks to date are case-sensitive - resource names, activity types, parameter names, etc. We should keep that consistent - especially since activity parameters can make their way directly into spacecraft commands via command expansion, we should err on the side of strictness rather than permissiveness with respect to user input

mattdailis: You do bring up an interesting point - which is that there is a notion of "match" that may be more general than "is syntactically equal to". That may be an avenue worth exploring

Mythicaeda: > especially since activity parameters can make their way directly into spacecraft commands via command expansion, we should err on the side of strictness rather than permissiveness with respect to user input

That's a good point. I'm alright with us erring on the side of strictness as long as we have it documented. So in this case, if we're going with "String comparisons are case-sensitive", that should be in some user-accessible documentation related to this feature.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -177,6 +175,42 @@ export default function myGoal() {
...
+        false);
+    if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Success<SchedulingDSL.GoalSpecifier> r) {
+      assertEquals(expectedGoalDefinition, r.value());
+    } else if (result instanceof SchedulingDSLCompilationService.SchedulingDSLCompilationResult.Error<SchedulingDSL.GoalSpecifier> r) {
+      fail(r.toString());
+    }
+  }
+
+  @Test
+  void stestSchedulingDSL_recurrence_activityFinder()
```
Mythicaeda:
```suggestion
  void testSchedulingDSL_recurrence_activityFinder()
```

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Test
+  public void testSubsetEqualityValuesString(){
+    final var reference = SerializedValue.of("AbCd");
+    final var sameValue = SerializedValue.of("AbCd");
+    final var differentValue = SerializedValue.of("bCd");
+    final var differentType = SerializedValue.of(11);
+    assertTrue(subsetOrEqual(reference, sameValue));
+    assertFalse(subsetOrEqual(reference, differentValue));
+    assertFalse(subsetOrEqual(reference, differentType));
```
Mythicaeda: Can we add a `differentCase` here since we specifically want that to be `false`?
```suggestion
    final var reference = SerializedValue.of("AbCd");
    final var sameValue = SerializedValue.of("AbCd");
    final var differentValue = SerializedValue.of("bCd");
    final var differentCase = SeriealizeValue.of("abcd");
    final var differentType = SerializedValue.of(11);
    assertTrue(subsetOrEqual(reference, sameValue));
    assertFalse(subsetOrEqual(reference, differentValue));
    assertFalse(subsetOrEqual(reference, differentCase));
    assertFalse(subsetOrEqual(reference, differentType));
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: If we did not wanted order, we would use a set ? A list is an ordered data structure. Then comparison should include order.

mattdailis: This seems like another case of "syntactically equal" vs some other definition of match. Currently, all of our matches use syntactic equality - but there may be cases where users want a different definition of match (e.g. list contains, vs list equals exactly)

We are dabbling with this "looser" definition of match with this method `subsetOrEquals` - it expects strict equality on all types except for structs - for those it will match even if extra fields are present. That said, that's the same way typescript types work - a Typescript interface will match any object that has at least those fields. There's no way (that I'm aware of) to similarly define a type "sublist of X", or "contains same elements as X, independent of order". So, after considering the similarities with Typescript, I think it does make sense for our match to allow extra keys, but require strict equality on all other types

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: If there is a `someKey: null` element, the subsequent `subsetOrEqual` call will fail as the matching will call a `asSomeType` method on it. So I guess it needs to be protected.

mattdailis: I do agree that `containsKey` is a more direct way to communicate what is going on here. (The fact that `get` on `Map` returns null makes me sad every time...)

I'll leave it up to you what you think is more readable - here's what came out when I moved code around in my editor for a few minutes:
```java
     @Override
      public Boolean onMap(final Map<String, SerializedValue> value) {
        return arguments
            .asMap()
            .map(argumentsAsMap -> value
                .entrySet()
                .stream()
                .allMatch(patternElement -> {
                  if (!argumentsAsMap.containsKey(patternElement.getKey())) return false;
                  return subsetOrEqual(argumentsAsMap.get(patternElement.getKey()), patternElement.getValue());
                }))
            .orElse(false);
      }
```

mattdailis: Or, if we wanted to use `&&` instead of `if`:

```java
      @Override
      public Boolean onMap(final Map<String, SerializedValue> value) {
        return arguments
            .asMap()
            .map(argumentsAsMap -> value
                .entrySet()
                .stream()
                .allMatch(
                    patternElement ->
                        argumentsAsMap.containsKey(patternElement.getKey()) 
                        && subsetOrEqual(argumentsAsMap.get(patternElement.getKey()), patternElement.getValue())))
            .orElse(false);
      }
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -614,4 +618,84 @@ public String prettyPrint(final String prefix) {
...
+          } else {
+            return false;
+          }
+        }
+        return true;
+      }
+
+      @Override
+      public Boolean onList(final List<SerializedValue> value) {
+        final var optionalArgumentsAsList = arguments.asList();
```
Mythicaeda: This is minor, feel free to resolve, but `argumentsAsListOptional` is clearer than `optionalArgumentsAsList`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
Mythicaeda: Sure, although if we have to ensure that order equals, we need to be confident that neither of these Lists got their elements potentially reordered coming from Typescript/the DB. If we're confident in that, then I'm fine with resolving this.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -305,6 +308,234 @@ export default () => Goal.CoexistenceGoal({
...
+    var lookingFor = false;
+    final var expectedCreation = new SerializedActivity("GrowBanana",
+                                            Map.of("quantity", SerializedValue.of(1),
+                                                   "growingDuration", SerializedValue.of(MINUTES.in(MICROSECOND))));
+    for(final var actAtTime10: planByTime.get(MINUTE.times(10))){
+      if(actAtTime10.serializedActivity().equals(expectedCreation)){
+        lookingFor = true;
+      }
+    }
+    assertTrue(lookingFor);
```
Mythicaeda, line 437: Not this PR, but it's kinda irksome that the equality check has to be done like this because `MockMerlinService` drops ActivityIds

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I have added this at the beginning of `subsetOrEqual`: 
```
   if(pattern == null){
      return true;
    }
    if(arguments == null){
      return pattern.isNull();
    }
```
and a test in `SerializedValueSubsetTest`.

It is tricky, here I say that `SerializedValue.NULL` is a subset of `null` but that could be the subject of a discussion.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Test
+  public void testSubsetEqualityList(){
+    final var reference = SerializedValue.of(List.of(SerializedValue.of(8.0), SerializedValue.of(6.0), SerializedValue.of(9.0)));
+    final var subsetList = SerializedValue.of(List.of(SerializedValue.of(8.0),SerializedValue.of(6.0)));
+    final var nullList = SerializedValue.of(List.of(SerializedValue.NULL));
+    final var notSubsetList = SerializedValue.of(List.of(SerializedValue.of(8.0),SerializedValue.of(7.0)));
+    assertTrue(subsetOrEqual(reference, subsetList));
+    assertTrue(subsetOrEqual(reference, nullList));
+    assertFalse(subsetOrEqual(reference, notSubsetList));
```
Mythicaeda: Similar to with `testSubsetString`, can we have a case where the elements are in the wrong order?
```suggestion
    final var reference = SerializedValue.of(List.of(SerializedValue.of(8.0), SerializedValue.of(6.0), SerializedValue.of(9.0)));
    final var subsetList = SerializedValue.of(List.of(SerializedValue.of(8.0), SerializedValue.of(6.0)));
    final var nullList = SerializedValue.of(List.of(SerializedValue.NULL));
    final var notSubsetList = SerializedValue.of(List.of(SerializedValue.of(8.0), SerializedValue.of(7.0)));
    final var wrongOrderList = SerializedValue.of(List.of(SerializedValue.of(9.0), SerializedValue.of(6.0), SerializedValue.of(8.0)));
    assertTrue(subsetOrEqual(reference, subsetList));
    assertTrue(subsetOrEqual(reference, nullList));
    assertFalse(subsetOrEqual(reference, notSubsetList));
    assertFalse(subsetOrEqual(reference, wrongOrderList));
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I think we can be confident. If such a reordering happens, the problem lies elsewhere.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: We had a discussion with @mattdailis and the conclusion is that neither arguments should be null, so I check them with `Objects.requireNonNull`.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: True

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
adrienmaillard: done

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
adrienmaillard: done

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: done

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SerializedValueSubsetTest.java
adrienmaillard: done

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
adrienmaillard: done

camargo (approved):
Tested locally, also ran UI e2e tests against this. LGTM

PR #850, opened 2023-04-07
AutoValueMapper improvements

### LINKED ISSUE #816: Allow Record AutoValueMapper for any record type definition
The Record AutoValueMapper would be very useful if it was expanded to process annotations on any record type definition, instead of only working on Computed Attributes.

mattdailis:
* **Tickets addressed:** Closes #816 
* **Review:** By commit  

## Description

This PR improves the usefulness of the `@AutoValueMapper.Record` annotation by enabling its use outside of computed attributes.

The commits, in order:
1. [Fix getSupportedAnnotationTypes to refer to correct annotation](https://github.com/NASA-AMMOS/aerie/commit/2992bee1fdbe0d4b1bbd3c37271922fd5418519b) - due to an oversight when I originally wrote this code, `getSupportedAnnotationTypes` was looking for `@AutoValueMapper` rather than `@AutoValueMapper.Record`, which would sometimes cause it to split the files we're interested in across multiple rounds.
2. [Add box method to TypePattern](https://github.com/NASA-AMMOS/aerie/commit/e4adf6d4cd9d737aa4ff169a5cd8cb8f81cfbea0) - this change allows converting a Primitive type pattern into a Class type pattern, and was useful for [Handle parameterized record components](https://github.com/NASA-AMMOS/aerie/commit/7cb10693263dec06a89725a35480d42894cd7675).
3. [Move generateAutoValueMappers into a new class](https://github.com/NASA-AMMOS/aerie/commit/03249bce936f5f3a3606e7ebb42aba3504e40f09) - I think at least for AutoValueMappers (and perhaps for other aspects of annotation processing) - it seems more cohesive to group by feature rather than by layer. So - parsing and generation of AutoValueMappers is colocated in one file, rather than split across "MissionModelParser" and  "MissionModelGenerator". The layered approach would make more sense if we want to do any kind of transformations at the intermediate representation layer - which would mean that features wouldn't map one-to-one from parsing to generation.
4. [Refactor generateAutoValueMappers to use single expression](https://github.com/NASA-AMMOS/aerie/commit/0ada0662e101d50596bbd97a5c10554eae65e775) - this change helps with legibility of the generateAutoValueMappers method - the generating code now mirrors the generated code more closely.
5. [Split generateAutoValueMappers into two methods](https://github.com/NASA-AMMOS/aerie/commit/df61da404d10890f997105035ed68dd3004e3ef1) - this change enables looking up the type rules separately from code generation. This can be helpful for allowing more type rules to be known at generation time, allowing for more expressive "cyclic dependencies" between generated files.
6. [Allow AutoValueMapper outside of computed attributes](https://github.com/NASA-AMMOS/aerie/commit/d47dabd8ed1cf134ab08971d32db7e1fb5f49e47) - this commit removes the scoping of AutoValueMappers to computed attributes, and instead finds all `@AutoValueMapper.Record` annotations that are available in the same round as the `@MissionModel` annotation.
7. [Handle parameterized record components](https://github.com/NASA-AMMOS/aerie/commit/7cb10693263dec06a89725a35480d42894cd7675) - this was a bug reported to me by Josh while I was in the middle of working in this part of the code. Essentially records that look like `record Abc(List<String> foo) {}` would fail to generate an autovaluemapper due to this error: `class com.squareup.javapoet.ParameterizedTypeName cannot be cast to class com.squareup.javapoet.ClassName`. This commit removes the errant cast, and leverages TypePattern.from, which is more robust.
8. [Add support for generic records](https://github.com/NASA-AMMOS/aerie/commit/2d48216926a056294bdd9e44cdcab478bdbef9d4) - this commit makes sure that records that look like `record Pair<K, V>(K key, V value) {}` generate valid value mappers. The generated value mapper is parameterized by value mappers `KValueMapper` and `VValueMapper`, which will need to be provided by any generated code that wants to call the Pair value mapper.
9. [Enable use of AutoValueMapper in Configuration](https://github.com/NASA-AMMOS/aerie/commit/ab37ece6526fe57aa1aecb540cd2c078925a8e3a) - this commit moves the `typeRule` calls as early as possible, to allow for the broadest possible use of the AutoValueMapper annotation. Configuration is one of the first things to be generated, so I used that to test that the AutoValueMappers are indeed made available.

## Verification

I added examples to banananation and ran some manual tests in the UI for ParameterTestActivity, as well as BiteBanana's computed attributes. Every commit should successfully generate code and compile.

## Documentation

- [x] TODO: Open a PR to update AutoValueMapper docs before merging this PR

https://github.com/NASA-AMMOS/aerie-docs/pull/38

## Future work

This work is fairly self contained - no future work anticipated at the moment.


examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
```diff
@@ -23,6 +29,12 @@ public boolean validateInitialDataPath() {
...
   public static @Template Configuration defaultConfiguration() {
-    return new Configuration(DEFAULT_PLANT_COUNT, DEFAULT_PRODUCER, DEFAULT_DATA_PATH);
+    return new Configuration(DEFAULT_PLANT_COUNT, DEFAULT_PRODUCER, DEFAULT_DATA_PATH, DEFAULT_INITIAL_CONDITIONS);
   }
+
+  @AutoValueMapper.Record
+  public record InitialConditions(LinearResource fruit, double peel, Flag flag) {}
+
+  @AutoValueMapper.Record
+  public record LinearResource(double initial, double rate) {}
```
JoelCourtney: Would it be possible to use the `RealDynamics` class in Merlin Framework? I'd rather not have the modeller make a basic common structure like this if we can avoid it.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
```diff
@@ -33,9 +35,9 @@ public ComputedAttributes run(final Mission mission) {
     final var newFlag = bigBiteSize ? Flag.B : Flag.A;
     mission.flag.set(newFlag);
     mission.fruit.subtract(biteSize);
-    return new ComputedAttributes(bigBiteSize, newFlag);
+    return new ComputedAttributes(bigBiteSize, List.of(newFlag));
   }
 
   @AutoValueMapper.Record
-  public record ComputedAttributes(boolean biteSizeWasBig, Flag newFlag) {}
+  public record ComputedAttributes(boolean biteSizeWasBig, List<Flag> newFlag) {}
```
JoelCourtney: I don't understand the purpose of this change, it's always a one-element list.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -0,0 +1,166 @@
...
+        .skipJavaLangImports(true)
+        .build();
+  }
+
+  private static CodeBlock castIfGeneric(final Element record) {
+    final var typeParameters = ((Parameterizable) record).getTypeParameters();
+    if (typeParameters.isEmpty()) {
+      return CodeBlock.of("");
+    } else {
+      return CodeBlock.of("(Class<$T>) (Object) ", TypeName.get(record.asType()));
```
JoelCourtney, line 167: That looks sus. I don't know the use cases for double casts in Java, but in Typescript its usually bad news. Why is `(Object)` necessary?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -0,0 +1,166 @@
...
+        final var elementName = element.toString();
+        final var valueMapperIdentifier = typeMirror
+                                              .toString()
+                                              .replace(".", "_")
+                                              .replace("<", "_")
+                                              .replace(">", "_")
+                                              .replace("[", "_")
+                                              .replace("]", "_")
+                                              .replace(",", "_")
+                                          + "ValueMapper";
```
JoelCourtney:
```suggestion
        final var valueMapperIdentifier = typeMirror
                                              .toString()
                                              .replaceAll("[.><,\[\]]", "_")
                                          + "ValueMapper";
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Set;
+
+public class AutoValueMappers {
+  static TypeRule typeRule(final Element autoValueMapperElement, final ClassName generatedClassName) throws InvalidMissionModelException {
+    if (!autoValueMapperElement.getKind().equals(ElementKind.RECORD)) {
+      throw new InvalidMissionModelException(
+          "@%s.%s is only allowed on records".formatted(
+              AutoValueMapper.class.getSimpleName(),
+              AutoValueMapper.Record.class.getSimpleName()),
+          autoValueMapperElement);
```
JoelCourtney, line 42: You had to hard-code the names of `AutoValueMapper` and `Record` anyway, might as well inline it into the string.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelProcessor.java
```diff
@@ -183,26 +180,6 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
...
-
-        if (foundAutoValueMapperRequest.getKind() != ElementKind.RECORD) {
-          this.messager.printMessage(
-              Diagnostic.Kind.WARNING,
-              "@%s.%s is only allowed on records".formatted(
-                  AutoValueMapper.class.getSimpleName(),
-                  AutoValueMapper.Record.class.getSimpleName()),
-              foundAutoValueMapperRequest);
-        }
-      }
```
JoelCourtney, line 205: same thing about inline class names

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
JoelCourtney: [proof](https://regex101.com/r/M0YwaZ/1)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
Twisol: If we do an IDE-based refactor of `AutoValueMapper` to some other name, ideally we want the IDE to find all use-sites. There's no real guarantee it'll find references that are just sequences of characters in string literals, since those aren't true references at the language level.

If there are other places where these names are hardcoded into strings, we should probably fix those.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
Twisol: This method is used in this context:

```java
.add("$L$T.class,\n",
    castIfGeneric(record),
    ClassName.get((TypeElement) record))
```

So we can expect the expression at issue to be `(Class<T>) (Object) T.class`, where `T` is some parametric type. For instance, `(Class<List<Integer>>) (Object) List.class`.

The expression `List.class` alone has type `Class<List>`, not `Class<List<Integer>>`, and attempting to cast directly to `(Class<List<Integer>>)` throws a compile-time error. Via `jshell`:

```
jshell> (Class<List<Integer>>) List.class
|  Error:
|  incompatible types: java.lang.Class<java.util.List> cannot be converted to java.lang.Class<java.util.List<java.lang.Integer>>
|  (Class<List<Integer>>) List.class
|                         ^--------^
```

Hence the two-phase cast.

(EDIT: corrected the `jshell` example; it's still the same error, for the same reason, but now i'm actually comparing apples to apple&lt;Integer&gt;s)

JoelCourtney: Seems like if it would give a compile error, but you obfuscate the type enough that it can't check, it should just give a runtime cast error.

Twisol: It doesn't give a runtime error though :) This is perfectly safe, because of what `Class` represents. Because generics were added to the language late, there's no way to tell `List.class` what its type parameters are directly. This is "just" letting us get a Class instance that is restricted to use over a particular specialization of type parameters.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
mattdailis: Good point - the purpose was to reproduce a bug where generic record components would cause an exception to be thrown during annotation processing.

I think the RecordParameter example should cover this case, so I'll go ahead and revert BiteBanana 👍

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: Yes - here is the comment from the original PR where Jonathan asked me to keep the classname where the IDE can see it:

https://github.com/NASA-AMMOS/aerie/pull/188#discussion_r889611173

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: I stole this idea from here:
https://github.com/NASA-AMMOS/aerie/blob/26a7d5d5f426495bde8fe89a4b900fb1933d38d7/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/Resolver.java#L71-L73

The goal is to unify the record `.class` with the type of the "projection" functions in the components. The contract is that the RecordValueMapper takes a `Class<R>` (whose main purpose is to provide a constructor via reflection), as well as a list of components, each of which defines a "projection" `Function<R, T>`, where `T` is the type of that component.

https://github.com/NASA-AMMOS/aerie/blob/26a7d5d5f426495bde8fe89a4b900fb1933d38d7/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/serialization/mappers/RecordValueMapper.java#L22-L25

With no casts at all, we get this fun-to-decipher error that says that `R` cannot be simultaneously a `RecordParameter` (the type of the `.class`), and a `RecordParameter<T>` (the input type of the projection function)

```
incompatible types: cannot infer type arguments for RecordValueMapper<>
    return new RecordValueMapper<>(
                                ^
    reason: inference variable R#1 has incompatible equality constraints RecordParameter<T>,R#2,RecordParameter
  where R#1,T,R#2 are type-variables:
    R#1 extends Record declared in class RecordValueMapper
    T extends Object declared in method <T>gov_nasa_jpl_aerie_banananation_activities_ParameterTestActivity_RecordParameter(ValueMapper<T>)
    R#2 extends Object declared in record Component
```

Sounds like this pattern requires a code comment - since it does elicit a perfectly reasonable "what the..." from the reader. I'll see if I can wordsmith something together

mattdailis: > Seems like if it would give a compile error, but you obfuscate the type enough that it can't check, it should just give a runtime cast error.

In this particular case - there can be no runtime error because there is no observable difference at runtime between a `Class<MyRecord<T>>` and a `Class<MyRecord>` - both identically erase to `MyRecord.class`. This cast is merely to help us with some compile time bookkeeping: it makes sure that we use this particular `MyRecord.class` to represent `MyRecord<T>` and not simply `MyRecord`.

This does seem to be an artifact of Java's generics being added late, and trying to interoperate with non-generic types - it allows you to cast at the top level:
```java
(ArrayList<List<Integer>>) new ArrayList();
```
But not a level deeper:
```java
(ArrayList<List<Integer>>) new ArrayList<List>();
// Inconvertible types; cannot cast 'java.util.ArrayList<java.util.List>' to 'java.util.ArrayList<java.util.List<java.lang.Integer>>'
```
We could hack around this with an `(Object)` cast. We would know that it's safe because we can see syntactically that the arraylist is empty, and that no other references exist to it - we are merely restricting uses of this value to treat it as an `ArrayList<List<Integer>>`.
```java
(ArrayList<List<Integer>>) (Object) new ArrayList<List>();
```
In that case - casting to `(Object)` would be silly, because we have control over the generic on the right, and can simply adjust it to be what we want it to be. In the case of `MyRecord.class` - that comes out as a `Class<MyRecord>`, and we have no opportunity to restrict it to `Class<MyRecord<T>>` without resorting to this ugly cast.

TL;DR - I suspect that Typescript was designed with generics from its inception, and thus has a cleaner story around them. In Java, there are some rough edges we have to work around...

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
mattdailis: Absolutely - this was merely meant as an example of a nested record. I can come up with something else to demonstrate the nested record to avoid duplication/confusion with RealDynamics

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: Thoughts on https://github.com/NASA-AMMOS/aerie/pull/850/commits/ed0ae77c4765bb8502fdc1614ee46b0b01b0bbd7 ? IntelliJ suggested that I pre-compile the regex, so I let it do that, but let me know if you think that hurts readability.

mattdailis: Another thought on this - I was mildly concerned that replacing all of these with underscores _could_ result in a name collision. There isn't really any way around this in general, but I contemplated whether using different replacement characters would alleviate that risk. I think my latest stance is that the risk is low enough, and the examples contrived enough, that it's fine to replace everything with underscores

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Configuration.java
mattdailis: I've removed the nested record - this functionality is covered by `RecordParameter`.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
mattdailis: I've reverted all changes to BiteBanana - they were useful when I was experimenting, but don't make sense in this activity

mattdailis:
Adding a `DON'T MERGE` label to remind me to either squash or drop the fixup commit before merging

- [x] Squash or drop fixup commit

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: I'm adding a `SAFETY` comment:

```
SAFETY: This cast cannot fail - it merely exists to compensate for an inability
to express a MyRecord<T>.class. Instead, we use MyRecord.class, and cast it, via
Object, to a `Class<MyRecord<T>>`. We need to go via Object because Java generics
do not allow casting a Foo<Bar> to a Foo<Bar<Baz>> - since Bar may be a container
that already contains non-Baz objects. In this case, since MyRecord.class is not
a container, the Java type checker is being overly conservative
```

Twisol: I would add, what we really want is the `Class<F<T>>`, but there is literally no way to do so without a cast. And it's safe to do *this* cast because of what you said above.

`List.class` semantically describes *all* lists, but its default type is too restrictive to actually use it on any but raw lists.

camargo (approved):
Tested locally, LGTM!

PR #851, opened 2023-04-10
Remove Simulation Bounds from Simulation Templates

Mythicaeda:
* **Review:** By commit  

## Description

- Removes the ability for a simulation template to specify the simulation bounds.
- Forces the sim bounds in the sim config to be non null. 
- Initializes the sim bounds to the plan bounds on creation of a simulation row
- Enforces the sim end bound to be >= sim start bound. The "or equal to" was added so that plans with duration zero can be created, although I'm not sure what their use case would be.
- Removed Optional wrapper around return of `GetSimulationTemplateAction.get()`

## Verification

- Updated Temporal Subset Simulation e2eTests and removed those that no longer apply (ie those that tested whether bounds specified in the sim config or sim config template took precedence)
- Updated "Verify posting of simulation results" test in scheduler

## Documentation

- [ ] Docs issue [here](https://github.com/NASA-AMMOS/aerie-docs/issues/32).

### DIFF
--- a/e2e-tests/src/tests/scheduler.test.ts
+++ b/e2e-tests/src/tests/scheduler.test.ts
@@ -226,7 +226,7 @@ test.describe('Scheduling', () => {
     //expect one profile per resource in the banananation model
     expect(dataset.length).toEqual(7)
     for(let resource of dataset){
-      if(resource.name =="/fruit" || resource.name == "/peel"){
+      if(resource.name == "/fruit" || resource.name == "/peel"){
         let startOffsetOfResource = new Set<string>()
         for(let segment of resource.profile_segments){
           startOffsetOfResource.add(segment.start_offset)
@@ -244,8 +244,6 @@ test.describe('Scheduling', () => {
     let peelOutputIsThere = false
     let biteOutputIsThere = false
     let growOutputIsThere = false
-    //topic added in the mission model
-    let producerIsThere = false
     for(let topic of topics){
       switch (topic.name){
         case prefixInput+"BiteBanana":
@@ -272,9 +270,9 @@ test.describe('Scheduling', () => {
           peelOutputIsThere = true
           expect(topic.events.length).toEqual(12)
           break
-        case "/producer":
-          producerIsThere = true
-          expect(topic.events.length).toEqual(0)
+        default:
+          test.fail(topic.events.length !== 0, "Unexpected topic with events: "+topic.name)
+          break
       }
     }
     expect(peelInputIsThere)
@@ -283,7 +281,6 @@ test.describe('Scheduling', () => {
     expect(peelOutputIsThere)
     expect(growOutputIsThere)
     expect(biteOutputIsThere)
-    expect(producerIsThere)
   });
 
   test('Get Plan', async ({ request }) => {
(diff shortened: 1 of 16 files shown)


e2e-tests/src/tests/scheduler.test.ts
```diff
@@ -272,9 +270,9 @@ test.describe('Scheduling', () => {
           peelOutputIsThere = true
           expect(topic.events.length).toEqual(12)
           break
-        case "/producer":
-          producerIsThere = true
-          expect(topic.events.length).toEqual(0)
+        default:
+          test.fail(topic.events.length !== 0, "Unexpected topic with events: "+topic.name)
```
mattdailis, line 274: Is this change simply refactoring, or is it related to the changes to simulation templates made in this PR? If it's simply refactoring, could we pull it into its own commit?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationTemplateAction.java
```diff
@@ -28,13 +27,11 @@ public GetSimulationTemplateAction(final Connection connection) throws SQLExcept
     this.statement = connection.prepareStatement(sql);
   }
 
-  public Optional<SimulationTemplateRecord> get(final long simulationTemplateId)
+  public SimulationTemplateRecord get(final long simulationTemplateId)
```
mattdailis: If it's no longer safe to call this action on a non-existent simulation template id - can we add a javadoc comment documenting what it will do if it is provided a non-existent id? It is true that it is currently called in a safe fashion, but by changing the behavior from "returns Optional.empty()" to "throws an exception, probably?", we've made it more difficult for a future programmer to use this action safely.

mattdailis (commented):
SQL changes look good to me - just a couple comments about the e2e tests and the GetSimulationTemplate action

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationTemplateAction.java
Mythicaeda: Done 👍. Also added a similar documentation for GetSimulationAction, where the same change was made in a prior PR.

Adjacently, I'm not sure if this method is now more "unsafe". Obviously it throws an exception rather than returning an empty Optional if there are no results, but blindly using that Optional without checking if the value is present would already lead to an exception (with a slightly obfuscated root cause, depending on how far from the `get` the value is blindly used), and I'm not certain the safety differences between checking the Optional as soon as it's returned and throwing an exception if it's empty and throwing an exception in the method itself. Could you please clarify that part?

e2e-tests/src/tests/scheduler.test.ts
Mythicaeda: Done 👍

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationTemplateAction.java
Mythicaeda: Resolving as I've dropped the commit this conversation is about.

mattdailis (approved):
Looks good to me 👌

PR #858, opened 2023-04-12
Add @FixedDuration annotation

### LINKED ISSUE #831: Investigate fixing scheduling running too many simulations

JoelCourtney:
* **Tickets addressed:** closes #831 
* **Review:** By commit  

## Description

This allows modellers to differentiate uncontrollable duration activities into dynamic and static durations. If the activity has a fixed statically known duration, they can apply the `FixedDuration` annotation to a static field or static factory method in the activity class which tells the scheduler the duration of the activity.

Currently the scheduler resimulates $2n+1$ times per goal where $n$ is the number of uncontrollable duration activities placed. For `FixedDuration` activities this is reduced to just $1$, same as `ControllableDuration`.

## Verification

- I modified `BananaNap` to use a fixed duration field, and added a test that the scheduler can place it correctly.
- I added `RipenBanana` which is pretty much the same as `BananaNap`, but uses the annotation on a static method instead of a field, and added a similar test.
- Fixed another integration test for window references in coexistence goal that was slightly changed by this.

## Documentation

- [x] Doc comments on duration annoations
- [ ] Real docs

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
@@ -1,6 +1,5 @@
 package gov.nasa.jpl.aerie.scheduler.constraints.activities;
 
-import gov.nasa.jpl.aerie.constraints.model.DiscreteProfile;
 import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
 import gov.nasa.jpl.aerie.constraints.model.Profile;
 import gov.nasa.jpl.aerie.constraints.time.Interval;
@@ -11,7 +10,6 @@ import gov.nasa.jpl.aerie.constraints.tree.DurationLiteral;
 import gov.nasa.jpl.aerie.constraints.tree.Expression;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.protocol.types.DurationType;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.scheduler.EquationSolvingAlgorithms;
 import gov.nasa.jpl.aerie.scheduler.NotNull;
 import gov.nasa.jpl.aerie.scheduler.model.SchedulingActivityDirective;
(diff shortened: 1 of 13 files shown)


examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BananaNapActivity.java
```diff
@@ -13,8 +16,11 @@
  */
 @ActivityType("BananaNap")
 public final class BananaNapActivity {
+  @ActivityType.FixedDuration
```
mattdailis, line 19: I was wondering whether it would be possible to enable passing the duration into the annotation directly - a quick attempt seems to indicate that it's not possible - so I think the two ways provided in this PR are indeed the best we can do

Screenshot 2023-04-13 at 6 57 58 AM

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
```diff
@@ -369,7 +370,25 @@ public Duration valueAt(final Duration start) {
...
+              this.arguments,
+              earliestStart,
+              facade.getLatestConstraintSimulationResults(),
+              evaluationEnvironment,
+              type),
+          null,
+          null,
+          true));
+    } else {
      throw new UnsupportedOperationException("Duration type other than Uncontrollable and Controllable are not suppoerted");
```
mattdailis: The exception message here can use a little update:
```suggestion
     throw new UnsupportedOperationException("Duration types other than Uncontrollable, Controllable, and Fixed are not supported: " + this.type.getDurationType());
```

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/FixedDurationTest.java
```diff
@@ -0,0 +1,101 @@
...
+  PlanningHorizon planningHorizon;
+  Problem problem;
+  Plan plan;
+
+  @BeforeEach
+  void setUp(){
+    planningHorizon = new PlanningHorizon(TestUtility.timeFromEpochSeconds(0), TestUtility.timeFromEpochDays(3));
+    MissionModel<?> bananaMissionModel = SimulationUtility.getBananaMissionModel();
+    problem = new Problem(bananaMissionModel, planningHorizon, new SimulationFacade(planningHorizon, bananaMissionModel), SimulationUtility.getBananaSchedulerModel());
+    plan = makeEmptyPlan();
```
mattdailis: The `plan` field seems unused - and there are also some unused imports in this class

mattdailis (approved):
A few minor comments, but overall looks great to me 💯

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BananaNapActivity.java
JoelCourtney: Yeah, I tried. Annotation arguments must be [constant expressions](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.28)

PR #876, opened 2023-04-18
Expanded sequence sorting

### LINKED ISSUE #866: Add log that says sequence commands are not sorted
This is just to add a metadata field to sequence telling if the commands are sorted or not.

### LINKED ISSUE #879: Time sort expansion commands only if there is more than one activity
We need to update our time sorting and only sort if there is more than one activity.

- If there is a mix of absolute and relative times (where each command has an absolute first command), make all times for both activities absolute, then sort. Return all times as absolute.
- Remove all single activity -> sequencing sorting

cohansen:
Closes #866 
Closes https://github.com/NASA-AMMOS/aerie/issues/879

* **Review:** By commit

## Description
Adds a new `timeSorted` flag to the sequence metadata and adds sorting if every command is `EPOCH_RELATIVE`.

## Verification
Updated existing tests and added new tests.

## Documentation
NA

## Future work
Possibly more sorting behavior? Hopefully this catches all of the valid cases.

### DIFF
--- a/sequencing-server/src/defaultSeqBuilder.ts
+++ b/sequencing-server/src/defaultSeqBuilder.ts
@@ -7,36 +7,46 @@ export const defaultSeqBuilder: SeqBuilder = (
   seqMetadata,
   simulationDatasetId,
 ) => {
+  // A combined list of all the activity instance commands.
+  let allCommands: CommandStem<{} | []>[] = [];
+  let activityInstaceCount = 0;
   let planId;
   let timeSorted = false;
 
-  const commands = sortedActivityInstancesWithCommands.flatMap(ai => {
-    // No associated Expansion
-    if (ai.errors === null) {
-      return [];
+  for (const ai of sortedActivityInstancesWithCommands) {
+    // If errors, no associated Expansion
+    if (ai.errors !== null) {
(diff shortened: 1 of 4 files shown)


camargo (approved):
Tested locally, LGTM

sequencing-server/src/defaultSeqBuilder.ts
```diff
@@ -7,80 +7,95 @@ export const defaultSeqBuilder: SeqBuilder = (
...
+  let previousTime: Temporal.Instant | null = null;
+
+  // If we have mixed time commands then we only sort if they're all absolute and relative.
   for (const command of commands) {
     // If the command is epoch-relative, complete, or the first command and relative then short circuit and don't try and sort.
     if (
       command.epochTime !== null ||
       (command.absoluteTime === null && command.epochTime === null && command.relativeTime === null) ||
       (command.relativeTime !== null && commands.indexOf(command) === 0)
     ) {
```
cohansen: Okay I understand, you're correct in your assumption. I'll change it so it checks the first command of each activity.

cartermak (approved):
LGTM, thanks for taking the time to workshop the details!

camargo:
Thanks so much for taking a look @cartermak, it's super helpful

Issue #687, opened 2023-02-21
Mark simulation as failed if error occurs while claiming a simulation


mattdailis:
There's currently a narrow window between when the merlin-worker sets the simulation_dataset status to `incomplete`, and when it actually starts a simulation. If an error occurs in this window, the worker simply crashes, and does not update the simulation status.

This ticket would be satisfied by catching errors that occur in this window and marking the simulation as failed, and updating the reason as well.

camargo:
Possibly related? https://github.com/NASA-AMMOS/aerie/issues/613

Issue #685, opened 2023-02-17
Intermittent IllegalStateException while simulating


parkerabercrombie:
We've seen this error sometimes while simulating plans in Aerie 1.0.1. It only appears intermittently, and the simulation usually works if we run it again.

```
java.lang.IllegalStateException: Unhandled message: 2 @ com.impossibl.postgres.protocol.v30.PrepareRequest$Handler
	at com.impossibl.postgres.protocol.v30.MessageDispatchHandler.dispatch(MessageDispatchHandler.java:225)
	at com.impossibl.postgres.protocol.v30.MessageDispatchHandler.channelRead(MessageDispatchHandler.java:152)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:311)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:432)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
[...]

PR #885, opened 2023-04-24
Remove empty transaction from claimSimulationDataset

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

#723 removed the only remaining action wrapped in this transaction context - so now this transaction context is empty and can safely be deleted.

## Verification

I'm relying on end-to-end tests that run simulation to verify this change.

## Documentation

No existing documentation should be affected by this change.

## Future work

#687 is still desirable - although after #723 and this PR, there are fewer moving parts between claiming and starting a simulation


camargo (approved):

mattdailis:
I'm not sure why the test failed the first time - my second commit would include the error message

My best guess is that it was a recurrence of #685

PR #888, opened 2023-04-26
Fix constraints codegen when there are no real resources

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

@ewferg discovered a bug where a mission model with a single enum discrete resource would cause invalid typescript to be generated for the constraints library.

This manifested itself in the following clear-as-mud error message:
```
[{"errors":[{"location":{"column":329,"line":10},"message":"TypeError: TS1110 Type expected.","stack":"at (10:329)"}],"goal_id":1}]
```

Upon examining the generated typescript code, I observed the following:
```typescript
export type RealResourceName = ;
```

Eric's mission model did not, at the time, export any real resources. It exported just a single discrete resource, of type ENUM.

This PR adds `| never` the the end of `ResourceName` and `RealResourceName` to make sure that valid code is generated even in the empty case. I consulted [this stackoverflow answer](https://stackoverflow.com/a/64246079) to confirm that including `never` in a union type is a no-op.

## Verification

I added a new `minimal-mission-model` that exhibits this behavior, and wrote a scheduling integration test that loads this model. In the second commit of this PR, that test should fail. After the third, it should pass.

(I pushed the second commit first, so the test failure is apparent: https://github.com/NASA-AMMOS/aerie/actions/runs/4809268807/jobs/8560283793)

I'm open to suggestions if people think this level of test is too high up the stack - it was simply the easiest way I could think of to make a mission model, and witness a typescript compilation error. Perhaps something in ConstraintsDSLCompilationTests would be more appropriate.

## Documentation

This is a bugfix - no existing documentation should be invalidated.

## Future work

- The root cause of this error was awfully difficult to see, in part because the user code runner obscures the source of the error. I would like to see us improve the debuggability of our DSLs - it is unfortunately not reasonable to assume that the generated code is always perfect


JoelCourtney (approved):

JoelCourtney:
Technically this same behavior would happen on `DiscreteResource` if the model had no resources at all. But like...that seems unlikely.

camargo (approved):
LGTM, thanks for the quick fix!

PR #890, opened 2023-04-26
Added Local and Parameters with TS 5

### LINKED ISSUE #706: Support locals and parameters in the Sequencing EDSL
The [Seq JSON spec](https://github.com/NASA-AMMOS/seq-json-schema/) supports passing sequence [parameters](https://github.com/NASA-AMMOS/seq-json-schema/blob/develop/schema.json#L651), and declaring [local variables](https://github.com/NASA-AMMOS/seq-json-schema/blob/develop/schema.json#L640). We need a way to allow declaring both parameters and locals in the sequencing EDSL, use them in the sequence, and properly output them in `.seq.json` as argument [symbols](https://github.com/NASA-AMMOS/seq-json-schema/blob/develop/schema.json#L392).

For example here is a proposed idea for how to do this in the EDSL. Psyche provided us one of their [sequence generation scripts](https://github.jpl.nasa.gov/PSYCHE/psyche-reusable-sequences/blob/develop/ep_swirl_preload.py), and [.seq.json output](https://github.jpl.nasa.gov/PSYCHE/psyche-reusable-sequences/blob/develop/ep/ep_swirl_preload/rs.ep_swirl_preload.r18.seq.json) we can draw inspiration from:

```ts
export default () =>
  Sequence.new({
    seqId: '',
    locals: [{ name: 'duration', type: 'FLOAT' }],
    parameters: [{ name: 'timeout', type: 'UINT' }],
    metadata: {},
    steps: [
      // Notice we use magic globals 'locals' and 'parameters' that bind to
      // the declared values in 'locals' and 'parameters' above.
      // This is just one idea.
      C.FSW_CMD_0({ arg0: locals.duration, arg1: parameters.timeout  })
    ],
  });
```

And the corresponding `.seq.json`:

```json
{
  "id": "",
  "metadata": {},
  "locals": [
    {
      "name": "duration",
      "type": "FLOAT"
    }
  ],
  "parameters": [
    {
      "name": "timeout",
      "type": "UINT"
    }
  ],
  "steps": [
    {
      "args": [
        { "name": "arg0", "type": "symbol", "value": "duration" },
        { "name": "arg1", "type": "symbol", "value": "timeout" }
      ],
      "stem": "FSW_CMD_0",
      "time": { "type": "COMMAND_COMPLETE" },
      "type": "command"
    }
  ]
}
```

goetzrrGit:
* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/706
* **Review:** By commit  

## Description
Leveraging the new TS 5 feature that @jdylanstewart figured out `local` and `parameters` work a lot better. 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    locals: [
      UINT('temperature')
    ],
    parameters: [
      UINT('sugarAmount')
    ],
    steps: ({ locals, parameters }) => ([
      A`2025-358T12:01:59.000`.PREHEAT_OVEN({
        temperature: locals.temperature,
      }),
      C.PREPARE_LOAF({
        tb_sugar: parameters.sugarAmount,
        gluten_free: 'FALSE',
      }),
    ]),
  });
```

There is also type checking between defined `local` and `parameters` and when you use them as argument parameters.

```ts
locals : [FLOAT("temperature")]
...
C.PREPARE_LOAF({
        tb_sugar: 360,
        gluten_free: locals.temperature <----ERROR: Temperature isn't a boolean
      }),
```

I've included examples of errors that may occur when converting between SeqJson to eDSL, which should make it easier to identify and fix any issues. 

### **From SeqJSON to eDSL Error checking**

```json

  "steps": [
    {
      "args": [
        {
          "name": "tb_sugar",
          "type": "symbol",
          "value": "sugarrrrr"
        },
        {
          "name": "gluten_free",
          "type": "string",
          "value": "FALSE"
        }
      ],
      "stem": "PREPARE_LOAF",
      "time": {
        "type": "COMMAND_COMPLETE"
      },
      "type": "command"
    }
  ]

```

Errors caught and displayed in the eDSL
```ts
      C.PREPARE_LOAF({
        tb_sugar: UNKNOWN.sugarrrrr, //ERROR:  variable 'sugarrrrr' is not defined as a local or parameter"
        gluten_free: 'FALSE',
 
)



```

## Verification
Added new e2e test to check all the errors for `locals` and `parameters`. All pass

## Documentation
This is probably a good write-up to include in the wiki

## Future work
Globals are next

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -7,6 +7,91 @@ export enum TimingTypes {
   COMMAND_COMPLETE = 'COMMAND_COMPLETE',
 }
 
+export enum VariableType {
+  FLOAT = 'FLOAT',
+  INT = 'INT',
+  STRING = 'STRING',
+  UINT = 'UINT',
+  ENUM = 'ENUM'
+}
+
+export type VariableOptions = {
+  name: string;
+  type: VariableType;
+  enum_name?: string | undefined;
+  allowable_values?: unknown[] | undefined;
+  // @ts-ignore : 'VariableRange' found in JSON Spec
+  allowable_ranges?: VariableRange[] | undefined;
+  sc_name?: string | undefined;
+};
+// @ts-ignore : 'VariableDeclaration' found in JSON Spec
+export interface INT<N extends string> extends VariableDeclaration {
+  name: N;
+  type: VariableType.INT;
+  // @ts-ignore : 'VariableRange' found in JSON Spec
+  allowable_ranges?: VariableRange[] | undefined;
+  allowable_values?: unknown[] | undefined;
+  sc_name?: string | undefined;
+}
+
+// @ts-ignore : 'VariableDeclaration' found in JSON Spec
+export interface UINT<N extends string> extends VariableDeclaration {
+  name: N;
+  type: VariableType.UINT;
+  // @ts-ignore : 'VariableRange' found in JSON Spec
+  allowable_ranges?: VariableRange[] | undefined;
+  allowable_values?: unknown[] | undefined;
+  sc_name?: string | undefined;
+}
+
+// @ts-ignore : 'VariableDeclaration' found in JSON Spec
+export interface FLOAT<N extends string> extends VariableDeclaration {
+  name: N;
+  type: VariableType.FLOAT;
+  // @ts-ignore : 'VariableRange' found in JSON Spec
+  allowable_ranges?: VariableRange[] | undefined;
+  allowable_values?: unknown[] | undefined;
+  sc_name?: string | undefined;
+}
+
+// @ts-ignore : 'VariableDeclaration' found in JSON Spec
+export interface STRING<N extends string> extends VariableDeclaration {
+  name: N;
+  type: VariableType.STRING;
+  allowable_values?: unknown[] | undefined;
+  sc_name?: string | undefined;
+}
+
+// @ts-ignore : 'VariableDeclaration' found in JSON Spec
+export interface ENUM<N extends string, E extends string> extends VariableDeclaration {
+  name: N;
+  enum_name: E;
+  type: VariableType.ENUM;
+  // @ts-ignore : 'VariableRange' found in JSON Spec
+  allowable_ranges?: VariableRange[] | undefined;
+  allowable_values?: unknown[] | undefined;
+  sc_name?: string | undefined;
+}
+
+export type SequenceOptions = {
+  seqId: string;
+  // @ts-ignore : 'Metadata' found in JSON Spec
+  metadata: Metadata;
+
+  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
+  locals?: [VariableDeclaration, ...VariableDeclaration[]];
+  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
+  parameters?: [VariableDeclaration, ...VariableDeclaration[]];
+  // @ts-ignore : 'Step' found in JSON Spec
+  steps?: Step[];
+  // @ts-ignore : 'Request' found in JSON Spec
+  requests?: Request[];
+  // @ts-ignore : 'ImmediateCommand' found in JSON Spec
+  immediate_commands?: ImmediateCommand[];
+  // @ts-ignore : 'HardwareCommand' found in JSON Spec
+  hardware_commands?: HardwareCommand[];
+};
+
 // @ts-ignore : 'Args' found in JSON Spec
 export type CommandOptions<A extends Args[] | { [argName: string]: any } = [] | {}> = {
   stem: string;
@@ -75,25 +160,6 @@ export type GroundOptions = {
 
 export type Arrayable<T> = T | Arrayable<T>[];
 
-export interface SequenceOptions {
-  seqId: string;
-  // @ts-ignore : 'Metadata' found in JSON Spec
-  metadata: Metadata;
-
-  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
-  locals?: [VariableDeclaration, ...VariableDeclaration[]];
-  // @ts-ignore : 'VariableDeclaration' found in JSON Spec
-  parameters?: [VariableDeclaration, ...VariableDeclaration[]];
-  // @ts-ignore : 'Step' found in JSON Spec
-  steps?: Step[];
-  // @ts-ignore : 'Request' found in JSON Spec
-  requests?: Request[];
-  // @ts-ignore : 'ImmediateCommand' found in JSON Spec
-  immediate_commands?: ImmediateCommand[];
-  // @ts-ignore : 'HardwareCommand' found in JSON Spec
-  hardware_commands?: HardwareCommand[];
-}
-
 declare global {
   // @ts-ignore : 'SeqJson' found in JSON Spec
   class Sequence implements SeqJson {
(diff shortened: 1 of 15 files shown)


dyst5422:
Love this.

Thoughts on changing locals and parameters definitions from arrays to maps? So the above would look like

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    locals: {
      temperature: UINT,
    },
    parameters: {
      sugarAmount: UINT,
    },
    steps: ({ locals, parameters }) => ([
      A`2025-358T12:01:59.000`.PREHEAT_OVEN({
        temperature: locals.temperature,
      }),
      C.PREPARE_LOAF({
        tb_sugar: parameters.sugarAmount,
        gluten_free: 'FALSE',
      }),
    ]),
  });
```
Also, what does the type hint on them when using them in the steps look like?

goetzrrGit:
> Love this.
> 
> Thoughts on changing locals and parameters definitions from arrays to maps? So the above would look like
> 
> ```ts
> export default () =>
>   Sequence.new({
>     seqId: '',
>     metadata: {},
>     locals: {
>       temperature: UINT,
>     },
>     parameters: {
>       sugarAmount: UINT,
>     },
>     steps: ({ locals, parameters }) => ([
>       A`2025-358T12:01:59.000`.PREHEAT_OVEN({
>         temperature: locals.temperature,
>       }),
>       C.PREPARE_LOAF({
>         tb_sugar: parameters.sugarAmount,
>         gluten_free: 'FALSE',
>       }),
>     ]),
>   });
> ```
> 
> Also, what does the type hint on them when using them in the steps look like?

I am glad you love this. It is way cleaner with your suggestions using TS 5. We can't really change `local` and `parameter` to a map as the SeqJson Spec says those are supposed to be VariableDeclaration[] and the user might want to insert a full SeqJSON into those fields or a combination of SeqJSON and eDSL.

In this example `tb_sugar` can be either `int | unit`. `parameters.sugarAmount` is a `unit` all parameters now have this second option when applicable of FLOAT, INT, UINT, STRING, and ENUM.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -7,6 +7,85 @@ export enum TimingTypes {
   COMMAND_COMPLETE = 'COMMAND_COMPLETE',
 }
 
+const VariableType = ['FLOAT', 'INT', 'STRING', 'UINT', 'ENUM'] as const;
```
dyst5422: Does it make more sense to have this be a string enum?

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -238,6 +304,69 @@ declare global {
...
+      name: N,
+      optionals: { allowable_values?: unknown[]; sc_name?: string },
+  ): STRING<N>;
+  function STRING<const N extends string>(
+      name: N,
+      // @ts-ignore : 'VariableRange' found in JSON Spec
+      optionals?: { allowable_values?: unknown[]; sc_name?: string },
+  ): STRING<N>;
+
+  function ENUM<const N extends string, const E extends string>(name: N, enum_name: E): ENUM<N, E>;
```
dyst5422, line 366: Long term, we can probably type check the enum variant against the enum name, but that would be backwards compatible with _valid_ sequences.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -304,8 +434,91 @@ export class Sequence implements SeqJson {
...
+        return acc;
+      }, {} as { [Index in Parameters[number] as Index['name']]: Index['type'] });
+
+      const steps =
+          typeof opts.steps === 'function' ? opts.steps({ locals: localsMap, parameters: parametersMap }) : opts.steps;
+
+      return new Sequence({
+        seqId,
+        metadata,
+        locals: locals.length !== 0 ? [locals[0], ...locals.slice(1)] : undefined,
```
dyst5422, line 530: Interesting that we make these undefined instead of empty arrays.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -522,39 +765,81 @@ export class Sequence implements SeqJson {
...
+        `${requestString.length !== 0 ? `${indent(`requests: ${requestString}`, 2)},\n` : ''}` +
+        `${indent(`});`, 1)}`
     );
   }
 
   // @ts-ignore : 'Args' found in JSON Spec
   public static fromSeqJson(json: SeqJson): Sequence {
+    const localNames = json.locals
+        ? [
+          json.locals[0].name,
```
dyst5422: Why do we have to do this slice here? Something to do with preserving tuple order in the typechecker?

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -597,16 +882,278 @@ export class Sequence implements SeqJson {
...
+      ...(json.enum_name ? { enum_name: json.enum_name } : {}),
+      ...(json.allowable_ranges ? { allowable_ranges: json.allowable_ranges } : {}),
+      ...(json.allowable_values ? { allowable_values: json.allowable_values } : {}),
+      ...(json.sc_name ? { sc_name: json.sc_name } : {}),
     });
+    variable.setKind(kind);
+    return variable;
+  }
+
+  public toRefernceString(): string {
```
dyst5422: Typo

sequencing-server/test/seqjson-to-edsl.spec.ts
```diff
@@ -298,6 +296,92 @@ describe('getEdslForSeqJson', () => {
...
+    ],
+    parameters: [
+      INT('sugar')
+    ],
+    steps: ({ locals, parameters }) => ([
+      C.PREHEAT_OVEN({
+        temperature: locals.temp,
+      }),
+      C.PREPARE_LOAF({
+        tb_sugar: unknown.sugarrrrr //ERROR: Variable 'sugarrrrr' is not defined as a local or parameter,
```
dyst5422, line 379: I LOVE this. Awesome way to do partial success on the conversion here.

dyst5422 (approved):

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: The reason for the slice is that an empty array is not allowed for `locals/parameters` defined in the `SeqJSON` which is what `map` could generate. The same restriction applies to the `commands/steps` and `request` fields. If you are aware of an alternative method, I would be open to exploring it.

`locals : [VariableDeclaration, ...VariableDeclaration[]]`
`parameters : [VariableDeclaration, ...VariableDeclaration[]]`

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: do you preference with`[]` I guess I was trying to make it really clear that the value is not provided. :P

dyst5422: Not sure I agree with that choice in the seqJSON schema, but that's neither here nor there.

I think you should be able to just do 
`locals: locals.length === 0 ? undefined : locals,`

there might be something else I'm missing here though

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
dyst5422: You should also be able to do
```
const localNames = json.locals !== undefined ? json.locals.map(local => local.name) : [];
```

Again, I might be missing something as I don't have everything in vscode right now.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: It makes it clear and easier to read so I can make the change

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Do you mean something like this?

`ENUM('direction', ['fromStem', 'fromPeel'])`

`C.PEEL_BANANA({peelDirection : locals.direction})`

dyst5422: I think I misunderstood here. Is this is defining a _new_ enum, not declaring a use of an enum from the command dictionary? In that case, disregard.

I was thinking we were using an existing enum and basically saying "this enum name, with this enum variant", but in retrospect that makes no sense when declaring an input.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I am going to have to keep the original implementation as the suggested change is not working with the json spec interface.

dyst5422: Yeah, I wondered if it was going to affect the type checking. My guess is that the form we have specifies it as a non empty array since we essentially say the first element must exist.

cohansen (approved):
Looks good to me!

PR #893, opened 2023-04-27
FixedDuration handling bug

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  

## Description
fixes a bug where the synchronous scheduler agent didn't allow fixed duration activities in the initial plan.
Follow up from https://github.com/NASA-AMMOS/aerie/pull/858

## Verification
added a test


mattdailis (approved):
Looks good to me - I'm scanning the codebase to see if there's anywhere else we may have missed something like this - but I won't block this PR on that :+1:

I can't wait for pattern matching to get out of preview... it's going into it's fourth preview in Java 20: https://openjdk.org/jeps/433

PR #894, opened 2023-04-27
Run google autoformatter

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  

## Description
The formatting of our java code is often very inconsistent (or sometimes just bad). In an ideal world I'd like to include a formatter in our gradle setup and check for good formatting during CI, but I think that would be finicky and extremely annoying due to the size of the project. So instead I just ran [the google java autoformatter](https://github.com/google/google-java-format) on all the java files in the repo and called it good.

## Verification
No behavioral changes. Tests still pass.


JoelCourtney:
Cool! I found a bug where the autoformatter is not idempotent! uuuug

Mythicaeda:
Looking it over quickly, while there are some changes I like (ie [here](https://github.com/NASA-AMMOS/aerie/pull/894/files#diff-cfea0ceb64968dc532da9202570c6c95087eafbe04270e9b9cd54ad386bd4e86L32-R33)), I'm not a fan of all of the changes the autoformatter made (ie [here](https://github.com/NASA-AMMOS/aerie/pull/894/files#diff-949a54314d8aa7d591b51d07110ac3927693730c68c5549d677967b1392c37a8L37-L40), [here](https://github.com/NASA-AMMOS/aerie/pull/894/files#diff-d07bf6b8b9d93c46bb21a6b9fbc5840f9a40bc1eaec7457faa4ddf0a28631c09R8-R15), and [especially here](https://github.com/NASA-AMMOS/aerie/pull/894/files#diff-949a54314d8aa7d591b51d07110ac3927693730c68c5549d677967b1392c37a8L164-L174)).

mattdailis:
While I agree with the sentiment that consistent formatting is important (and that we don't have it), I think running an auto formatter on the whole codebase is a little too nuclear of an option.

If you have particular style preferences that you think Google's formatter gets right, could you start a style guide? That way, we can build consensus around a set of rules before we automate them. 

In the meantime, I have no problems with running an auto-formatter on individual files that you're working with and find problematic. What I'd like to avoid is letting an autoformatter change files that people have carefully polished.

When we do reformat files, it would help improve the utility of git blame if we make that change in a separate commit and include it in the "ignore revs" file https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

dyst5422:
I agree with the sentiment Matt.

On the other hand, anything not enforced by some mechanism is guaranteed to drift over time. So if there are any formatting choices deemed required, I think that needs to happen automatically.

JoelCourtney:
Aight it was worth a shot

Twisol:
FWIW, there is an `.editorconfig` file checked-in that we used in the past to standardize IntelliJ's auto-formatting behavior. Do people still use that?

PR #899, opened 2023-05-01
Store expanded sequences

### LINKED ISSUE #829: Save expanded sequences in the database
Currently expanded sequence seq JSON is not saved in the database and can only be generated on demand via an action. This task is to save the seq JSON in the database and store them with the expansion run.

### LINKED ISSUE #906: Store EDSL text as well in new expanded sequences table
Users indicated we want to show both EDSL text and Seq JSON text. So we should store both.

Part of https://github.com/NASA-AMMOS/aerie/issues/829

cohansen:
Closes #829 
Closes #906 

* **Review:** By commit

## Description
This PR adds a new table that we use to store expanded sequences and the EDSL string per expansion run. In order to generate the sequence we need an activity linked to a sequence ID, so there are some prior steps required before we generate and save the sequence.
1. Create a plan
2. Add activities
3. Simulate the plan
4. Add an expansion set
5. Define a sequence
6. Link an activity to that sequence
7. Run the expansion (this is when we save the expanded sequences)

## Verification
Ran existing tests and added a new one to check if the expanded sequence was stored.

## Documentation
NA

## Future work
Future UI work to show the stored expanded sequences and let the user select which one they'd like to view.

### DIFF
--- a/deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
+++ b/deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
@@ -2,17 +2,20 @@ table:
   name: expansion_run
   schema: public
 object_relationships:
-- name: expansion_set
-  using:
-    foreign_key_constraint_on: expansion_set_id
+  - name: expansion_set
+    using:
+      foreign_key_constraint_on: expansion_set_id
+  - name: expanded_sequences
+    using:
+      foreign_key_constraint_on: expansion_set_id
 array_relationships:
-- name: activity_instance_commands
-  using:
-    foreign_key_constraint_on:
-      column: expansion_run_id
-      table:
-        name: activity_instance_commands
-        schema: public
+  - name: activity_instance_commands
+    using:
+      foreign_key_constraint_on:
+        column: expansion_run_id
+        table:
+          name: activity_instance_commands
+          schema: public
 remote_relationships:
   - name: simulation_dataset
     definition:
--- a/deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
+++ b/deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
@@ -2,20 +2,17 @@ table:
   name: expansion_run
   schema: public
 object_relationships:
-  - name: expansion_set
-    using:
-      foreign_key_constraint_on: expansion_set_id
-  - name: expanded_sequences
-    using:
-      foreign_key_constraint_on: expansion_set_id
+- name: expansion_set
+  using:
+    foreign_key_constraint_on: expansion_set_id
 array_relationships:
-  - name: activity_instance_commands
-    using:
-      foreign_key_constraint_on:
-        column: expansion_run_id
-        table:
-          name: activity_instance_commands
-          schema: public
+- name: activity_instance_commands
+  using:
+    foreign_key_constraint_on:
+      column: expansion_run_id
+      table:
+        name: activity_instance_commands
+        schema: public
 remote_relationships:
   - name: simulation_dataset
     definition:
--- a/deployment/hasura/migrations/AerieSequencing/2_add_expanded_sequences/up.sql
+++ b/deployment/hasura/migrations/AerieSequencing/2_add_expanded_sequences/up.sql
@@ -23,4 +23,4 @@ create table expanded_sequences (
       on delete cascade
 );
 
-call migrations.mark_migration_applied('1');
+call migrations.mark_migration_applied('2');
--- a/sequencing-server/src/routes/command-expansion.ts
+++ b/sequencing-server/src/routes/command-expansion.ts
@@ -376,7 +376,7 @@ commandExpansionRouter.post('/expand-all-activity-instances', async (req, res, n
 
         return {
           ...ai,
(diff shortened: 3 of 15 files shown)


dyst5422 (approved):
Clean and readable

deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
```diff
@@ -2,17 +2,20 @@ table:
...
+      foreign_key_constraint_on: expansion_set_id
 array_relationships:
-- name: activity_instance_commands
-  using:
-    foreign_key_constraint_on:
-      column: expansion_run_id
-      table:
-        name: activity_instance_commands
-        schema: public
+  - name: activity_instance_commands
```
camargo: Looks like the auto-formatter hit this file. Can you remove the unchanged lines please?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -300,13 +303,117 @@ commandExpansionRouter.post('/expand-all-activity-instances', async (req, res, n
...
+          seqToSimulatedActivity.rows.map(row => row.seq_id),
+        )})
+        and sequence.simulation_dataset_id = $1;
+      `,
+      [simulationDatasetId],
+    );
+
+    // If the user has created a sequence, we can try to save the expanded sequences when an expansion runs.
+    if (seqRows.rows.length > 0) {
+      const seqId = seqRows.rows[0].seq_id;
```
camargo: So instead of just saving a single sequence, we need to account for multiple sequences. For example try adding two+ sequences and expanding. We should see both in the DB. Currently it is only grabbing the first.

deployment/hasura/metadata/databases/AerieSequencing/tables/public_expansion_run.yaml
cohansen: Done.

goetzrrGit (changes requested):
Looks great Cody, I have a question about the workflow. I am not sure If we had a meeting that I missed or was out on leave.

The proposed coded change will make every expansion run generate a sequence. From my mission planning experience on MSL, if the planning team needs to iterate over the plan multiple times, it might not be necessary to save sequences that are not going to be used. This approach could help conserve computational and database resources.

Therefore, I think we shouldn't make this a 2 operation process. Keep the /expand-all-activity-instances endpoint the same, and use the /bulk-get-seqjson-for-seqid-and-simulation-dataset to generate the sequence and save it. This way, if the planning team needs to make changes to a subset of activities, fix errors in expansion, or modify how an activity is expanded, they can do so without generating unnecessary sequences that will not be used. Once the team is satisfied with the changes made to the plan, the sequences can be generated as needed.

camargo:
@goetzrrGit We have a new requirement to save all sequences with each expansion run. This is so users can browse past runs and the sequences that were generated.

cohansen:
> Looks great Cody, I have a question about the workflow. I am not sure If we had a meeting that I missed or was out on leave.
> 
> The proposed coded change will make every expansion run generate a sequence. From my mission planning experience on MSL, if the planning team needs to iterate over the plan multiple times, it might not be necessary to save sequences that are not going to be used. This approach could help conserve computational and database resources.
> 
> Therefore, I think we shouldn't make this a 2 operation process. Keep the /expand-all-activity-instances endpoint the same, and use the /bulk-get-seqjson-for-seqid-and-simulation-dataset to generate the sequence and save it. This way, if the planning team needs to make changes to a subset of activities, fix errors in expansion, or modify how an activity is expanded, they can do so without generating unnecessary sequences that will not be used. Once the team is satisfied with the changes made to the plan, the sequences can be generated as needed.

I think this is more a process / design question for Chris? The original ticket was specifically to save the expanded sequences for every expansion run. It takes a series of actions that won't always be fulfilled when the user clicks the expand button, so there are cases where the user can still expand and not generate + save the expanded sequences.

goetzrrGit (approved):
Thanks for the information @camargo and @cohansen. I approve.

sequencing-server/src/routes/command-expansion.ts
dyst5422: 🤦 @camargo catching my oversite. Should be a simple enough map over seqRows.rows.

cohansen: This is implemented now, the only problem I had was getting `seq_id` setup as a foreign key, so right now the tables aren't officially linked. I tried just adding it like:
```
  foreign key (seq_id)
    references sequence (seq_id)
    on delete cascade
```
but that didn't work, any ideas?

deployment/hasura/migrations/AerieSequencing/2_add_expanded_sequences/up.sql
```diff
@@ -0,0 +1,26 @@
...
+      references expansion_run
+      on delete cascade,
+
+  constraint expanded_sequences_to_seq_id
+    foreign key (seq_id, simulation_dataset_id)
+      references sequence (seq_id, simulation_dataset_id)
+      on delete cascade
+);
+
+call migrations.mark_migration_applied('2');
```
Mythicaeda, line 26: You need to update `sequencing-server/sql/sequencing/applied_migrations.sql` to include this line as well

cohansen: Good catch, just added that!

PR #908, opened 2023-05-03
ParametricDuration annotation

JoelCourtney:
* **Tickets addressed:** closes #904 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

Adds a `ParametricDuration` annotation which is apply to a duration getter method. During root finding, the scheduler will create an instance of the original activity type class and call the annotated method, rather than simulating. This means that there are now uses of root finding that don't simulate.

## Verification
Added a `DownloadBanana` activity for testing, added a couple tests for timing constraints.

## Future work
- [ ] docs incoming

PR #911, opened 2023-05-04
Assign zero duration to activities with no effect model

JoelCourtney:
* **Tickets addressed:** inspired by #908 
* **Review:** By commit  

## Description
For activities that have no effect model, we should default to assigning `FixedDuration(ZERO)` instead of uncontrollable.

## Verification
Banananananation's ParameterTest activity's generated code was changed by this PR. Given that we have tests for FixedDuration, and tests that use ParameterTest, I'm satisfied not adding any new tests.

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
@@ -296,7 +296,7 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
                                                             else if ($.parametricDuration().isPresent()) return CodeBlock.of("parametric($$ -> (new $L().new InputMapper()).instantiate($$).$L())", activityTypeRecord.inputType().mapper().name, $.parametricDuration().get());
                                                             else return CodeBlock.of("uncontrollable()");
                                                           })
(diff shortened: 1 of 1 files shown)


camargo (approved):
Looks reasonable to me

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -303,7 +303,7 @@ public JavaFile generateSchedulerModel(final MissionModelRecord missionModel) {
                                                             else if ($.fixedDurationExpr().isPresent()) return CodeBlock.of("fixed($L.$L)", activityTypeRecord.fullyQualifiedClass(), $.fixedDurationExpr().get());
                                                             else return CodeBlock.of("uncontrollable()");
                                                           })
-                                                          .orElse(CodeBlock.of("uncontrollable()"))))
+                                                          .orElse(CodeBlock.of("fixed(gov.nasa.jpl.aerie.merlin.protocol.types.Duration.ZERO)"))))
```
mattdailis:
```suggestion
                                                          .orElse(CodeBlock.of("fixed($T.ZERO)", ClassName.get(Duration.class)))))
```

mattdailis (changes requested):
Just a slightly more idiomatic way to express this in JavaPoet - the change itself looks great, I'm glad it was this straightforward.

mattdailis (approved):
:rocket:

PR #915, opened 2023-05-04
Change severity scale for bug issue template

camargo:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

Suggested after speaking with @Mythicaeda about usefulness of that section

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug_report.yml
@@ -70,8 +70,9 @@ body:
       label: Severity
       description: Select the severity of this issue.
       options:
-        - Annoyance
-        - Blocking an upgrade
-        - Blocking all usage of Aerie
+        - Minor
+        - Moderate
+        - Major
+        - Critical
     validations:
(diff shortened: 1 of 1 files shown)


Mythicaeda (approved):

Twisol:
I liked the impact-based scale; I'm not sure what a "moderate bug" would be.

Any chance you could document the rationale you two came up with in this PR?

Mythicaeda:
An example of a moderate bug can be found [here](https://github.com/NASA-AMMOS/aerie/issues/916). My rationale for moderate in this case was: the feature not working is not a "minor inconvenience", but because it works in many cases, it's not a "major" bug.

Rationale for the change:
1. The original system wasn't complete and as we were discussing it, we realized we could get very far in the weeds trying to be complete. Ie, the severity of the bug I linked earlier did not fall on the scale at all. The original proposal was adding the severity "Blocking Use of Feature", but even that doesn't match the severity of the bug (which would be "Partially Blocking Use of Feature").
2. This new scale matches how we consider the criticality of bugs.

Issue #920, opened 2023-05-08
Make Mission Model statically declare ResourceTypes and ResourceSchemas 


Mythicaeda:
Currently, we're making an assumption that a model's resource types and resource schemas don't change. We should enforce that assumption by having these fields declared statically, similar to how ActivityTypes are statically declared.

Additionally, this will make getting the resource types easier when we upload them after mission model upload, since we will not have to instantiate the model to access them.

PR #921, opened 2023-05-08
Add resource types table

### LINKED ISSUE #605: Add resourceTypes table to database to make querying for them better
The `resourceTypes` action can be slow. The Clipper `eurc` model for version `G23.0-RC1a` has `2180` profiles. It takes `~2646 ms` to query in Hasura, and the UI. We need a way to paginate the results. Add support for a cursor, filter, and search to the action to allow for clients to query for subsets of the `resourceTypes` (sort of like Hasura supports for regular database queries). 

We should add resource types to the database and obviate this action. Similar to what we do for activity types.

```gql
query GetResourceTypes($missionModelId: ID!) {
  resourceTypes(missionModelId: $missionModelId) {
    name
    schema
  }
}
```
```json
{
  "missionModelId": 1
}
```

Mythicaeda:
* **Tickets addressed:** Closes #605
* **Review:** By commit  

## Description

- Added resource_types table
- Added a "refresh resource types" action that instantiates the mission model, then upserts the resource types from it.
- Updated CreateActivityTypeAction to perform one batch upsert rather than upserting each ActivityType in its own transaction

Important: The `getResourceTypes` GQL Action has been deprecated. Query the `resource_type` table instead.

## Verification

E2E tests validating that the resource and activity types are extracted on mission model insert were added.

## Future work

- [Make Mission Models statically declare ResourceTypes and ResourceSchemas](https://github.com/NASA-AMMOS/aerie/issues/920)
- Remove `getResourceTypes`


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -270,6 +270,17 @@ public void refreshActivityTypes(final String missionModelId)
     }
   }
 
+  @Override
+  public void refreshResourceTypes(final String missionModelId){
+    try {
+      //instantiate model
```
mattdailis: 🔵 Looks like a note-to-self comment that can be removed now

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertActivityTypesAction.java
```diff
@@ -0,0 +1,71 @@
...
+import java.sql.BatchUpdateException;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Collection;
+
+import static gov.nasa.jpl.aerie.merlin.driver.json.ValueSchemaJsonParser.valueSchemaP;
+
+/*package-local*/ final class InsertActivityTypesAction implements AutoCloseable {
```
mattdailis, line 15: Huge fan of the `Create -> Insert` rename as well as the batching functionality :100:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertActivityTypesAction.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public void apply(final int modelId, Collection<ActivityType> activityTypes)
+  throws SQLException, FailedInsertException
+  {
+    final var connection = statement.getConnection();
+    try {
+      // From the docs (https://docs.oracle.com/javase/tutorial/jdbc/basics/retrieving.html#batch_updates):
+      // "To allow for correct error handling, you should always disable auto-commit mode before beginning a batch update."
+      connection.setAutoCommit(false);
```
mattdailis, line 38: JFYI - this logic for toggling auto commit was the _raison d'être_ of the TransactionContext.

No change requested - what you have is just fine for this case 👍

.github/scripts/explanations_merlin_down
```diff
@@ -0,0 +1,16 @@
+function definition	hdb_catalog	hdb_catalog.hdb_catalog."notify_hasura_refreshResourceTypes_INSERT"()	missing in 1st DB	From_RefreshResourceTypes_Action	2
```
mattdailis, line 1: Not sure I quite understand what's going on here - I remember that pgcmp uses "explanations" to dismiss certain expected differences - and that we do have expected differences, since we're comparing a hasura-less environment to one with hasura running.

What I don't quite follow is what this new file's role is - and how it is or isn't different from refreshActivityTypes. Could you catch me up here?

Mythicaeda: Sure!

Contents of the file: 
In order to support the new Hasura Event (`refreshActivityTypes`), Hasura automatically creates new functions/triggers in its schema (`hdb_catalog`) on the relevant table in the relevant database (`mission_model` in `aerie_merlin`). I believe these objects should go away if we added checking out the v1.0.1 and restarting Hasura to update the metadata to the `dump-current` job. However, that would add at minimum 30 seconds to the job's run time, which is already ~3 mins. 

Reason why this file exists: 
When providing an explanations file, [pgcmp will consider an explanation a "bad explanation" if the explanation doesn't match the state of the DB](https://github.com/cbbrowne/pgcmp#pgcmp-1) (ie, if you say something is missing in the first DB, but it's actually present in both). The original `explanations.txt` were things that needed to be explained in all DBs for all the "down" comparisons (note [that `explanations.txt` isn't used in the "up" comparisons](https://github.com/NASA-AMMOS/aerie/blob/81deeb9fdf1c043178dd24ee4c7ca99c6b7509c4/.github/workflows/pgcmp.yml#L278-L283), since both DBs should have a `migrations` schema). These new explanations are only needed when comparing the Merlin Base and Merlin Migrated to Base DBs, so they had to be put in their own file. To avoid duplication, the file only contains the unique explanations, and the workflow concatenates them together to get the complete list of explanations for the Merlin DB in the `pgcmp-down` job.

For additional clarity:
> since we're comparing a hasura-less environment to one with hasura running

Both of the DBs had Hasura running. However, the metadata changes in this PR required Hasura to add Postgres objects (functions and triggers on the `mission model` table), which PGCMP notices. (In general, our metadata goes into a table the `aerie_hasura` database, which both isn't compared and maintains the same structure, pg-objects-wise)

> and how it is or isn't different from refreshActivityTypes

Hopefully the above explanation made it clear, but just in case: this isn't different from `refreshActivityTypes`, they're the Postgres Objects needed for it to work.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
Mythicaeda: Thanks! It's been removed.

.github/scripts/explanations_merlin_down
mattdailis: Thanks for the explanation :+1: I appreciate you framing the tradeoff between simplicity and the time it takes to run this test

mattdailis (approved):
Great work :rocket: I did a quick local test with banananation; looks good to me.

PR #931, opened 2023-05-11
Changed constraint violation api to return a list of violations rathe…

### LINKED ISSUE #897: Include constraint ID in check constraint action response
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

_No response_

Constraint violation checking response should identify constraints by ID and not just `<plan|model>/<constraint_name>`. The different naming convention between constraints and constraint violations is a headache for the UI and likely everyone else dealing with these violations via API. This may require a breaking change to the check constraints response data structure.

Potentially related to https://github.com/NASA-AMMOS/aerie/issues/892. If 892 is done then this check constraints response could return a single violation instead of an array of violations for each constraint.

cohansen:
…r than a map of constraint names and types to violations

* **Tickets addressed:** Closes #897 
* **Review:** By commit

## Description
Previously the constraint violation API was returning a list of violations keyed by constraint name and type (model or plan). This PR updates the API to return a list of violation objects where those objects have a constraint id and constraint type property.

Here's a sample return:
```
{
  "data": {
    "checkConstraintsResponse": {
      "violationsMap": [
        {
          "associations": {
            "activityInstanceIds": [],
            "resourceIds": [
              "/fruit"
            ]
          },
          "gaps": [
            {
              "end": 0,
              "start": "-9223372036854775808"
            },
            {
              "end": "9223372036854775807",
              "start": 604800000000
            }
          ],
          "id": 1,
          "type": "PLAN",
          "windows": [
            {
              "end": 604800000000,
              "start": 0
            }
          ]
        }
      ]
    }
  }
}
```

## Verification
Updated existing tests.

## Documentation
I'm not sure if this API is documented, but if it is I can change that.

## Future work
A UI ticket to consume the new API changes.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Violation.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Violation.java
@@ -9,6 +9,7 @@ import java.util.Objects;
 import java.util.stream.StreamSupport;
 
 public final class Violation {
+  public final String constraintName;
   public final Long constraintId;
   public final ConstraintType constraintType;
   public final List<Long> activityInstanceIds;
@@ -16,7 +17,8 @@ public final class Violation {
   public final List<Interval> violationWindows;
   public final List<Interval> gaps;
 
-  public Violation(final Long constraintId, final ConstraintType constraintType, final List<Long> activityInstanceIds, final List<String> resourceNames, final Windows violationWindows) {
+  public Violation(final String constraintName, final Long constraintId, final ConstraintType constraintType, final List<Long> activityInstanceIds, final List<String> resourceNames, final Windows violationWindows) {
+    this.constraintName = constraintName;
     this.constraintId = constraintId;
     this.constraintType = constraintType;
     this.activityInstanceIds = new ArrayList<>(activityInstanceIds);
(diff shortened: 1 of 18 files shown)


mattdailis:
Looks like this part of the API [isn't documented](https://nasa-ammos.github.io/aerie-docs/api/examples/constraints/) - could you add a brief example for how to initiate a constraints check via the API? You can ignore the old way and just show the new way introduced by this PR

cohansen:
> Looks like this part of the API [isn't documented](https://nasa-ammos.github.io/aerie-docs/api/examples/constraints/) - could you add a brief example for how to initiate a constraints check via the API? You can ignore the old way and just show the new way introduced by this PR

Yes! I'll add some documentation and link it here later today.

https://github.com/NASA-AMMOS/aerie-docs/pull/53

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Violation.java
```diff
@@ -32,18 +38,21 @@ public Violation(final List<Long> activityInstanceIds, final List<String> resour
   }
 
   public Violation(final Windows violationWindows) {
-    this(new ArrayList<>(), new ArrayList<>(), violationWindows);
+    this(null, null, null, new ArrayList<>(), new ArrayList<>(), violationWindows);
```
mattdailis, line 41: This here gives me the most pause - I'm torn on whether name, id, and type belong in `Violation` (You'll notice that that information previously was owned exclusively by the merlin-server - the constraints library didn't know that its constraints had names, or ids, or were associated with plans or mission models).

However, I do like how clean the rest of the code becomes after adding this information to `Violation`, and I don't have any better suggestions at the moment

mattdailis (approved):
Looks good to me - I left one musing comment, but I think it's more of a "future work" topic. I do think what you've done here improves on the status quo :+1: Nicely done 🚀

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/Violation.java
cohansen: I agree, I wasn't super happy with the solution either. Something to improve upon in the future!

PR #933, opened 2023-05-11
Enable usage of JWT environment variables for local e2e testing with Sequencing Server

goetzrrGit:
* **Review:** By commit  

## Description
This will enable the Sequencing Server and JEST to utilize the JWT environment variables configured within the Docker container, thereby allowing the execution of end-to-end tests on a local environment.

### DIFF
--- a/sequencing-server/jestGlobalSetup.js
+++ b/sequencing-server/jestGlobalSetup.js
@@ -4,8 +4,7 @@ config();
 
 process.env.MERLIN_GATEWAY_URL = process.env.MERLIN_GATEWAY_URL ?? 'http://localhost:9000';
 process.env.MERLIN_GRAPHQL_URL = process.env.MERLIN_GRAPHQL_URL ?? 'http://localhost:8080/v1/graphql';
-process.env.HASURA_GRAPHQL_JWT_SECRET = process.env.HASURA_GRAPHQL_JWT_SECRET ?? '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }'
-process.env.HASURA_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? 'aerie';
+process.env.HASURA_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? '';
 
 export default async () => {
 
--- a/sequencing-server/jestGlobalSetup.js
+++ b/sequencing-server/jestGlobalSetup.js
@@ -4,7 +4,8 @@ config();
 
 process.env.MERLIN_GATEWAY_URL = process.env.MERLIN_GATEWAY_URL ?? 'http://localhost:9000';
 process.env.MERLIN_GRAPHQL_URL = process.env.MERLIN_GRAPHQL_URL ?? 'http://localhost:8080/v1/graphql';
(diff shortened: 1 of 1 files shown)


sequencing-server/jestGlobalSetup.js
```diff
@@ -4,7 +4,8 @@ config();
 
 process.env.MERLIN_GATEWAY_URL = process.env.MERLIN_GATEWAY_URL ?? 'http://localhost:9000';
 process.env.MERLIN_GRAPHQL_URL = process.env.MERLIN_GRAPHQL_URL ?? 'http://localhost:8080/v1/graphql';
-process.env.HASURA_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? '';
+process.env.HASURA_GRAPHQL_JWT_SECRET = process.env.HASURA_GRAPHQL_JWT_SECRET ?? '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }'
+process.env.HASURA_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? 'aerie';
```
cohansen: I'm not sure if we want to be hardcoding the dev env variables, shouldn't we just require that the user set them?

sequencing-server/src/env.ts
```diff
@@ -13,7 +14,8 @@ export type Env = {
 };
 
 export const defaultEnv: Env = {
-  HASURA_GRAPHQL_ADMIN_SECRET: '',
+  HASURA_GRAPHQL_ADMIN_SECRET: 'aerie',
+  HASURA_GRAPHQL_JWT_SECRET : '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }',
```
cohansen: Same as my other comment.

sequencing-server/jestGlobalSetup.js
```diff
@@ -4,7 +4,8 @@ config();
 
 process.env.MERLIN_GATEWAY_URL = process.env.MERLIN_GATEWAY_URL ?? 'http://localhost:9000';
 process.env.MERLIN_GRAPHQL_URL = process.env.MERLIN_GRAPHQL_URL ?? 'http://localhost:8080/v1/graphql';
-process.env.HASURA_GRAPHQL_ADMIN_SECRET = process.env.HASURA_GRAPHQL_ADMIN_SECRET ?? '';
+process.env.HASURA_GRAPHQL_JWT_SECRET = process.env.HASURA_GRAPHQL_JWT_SECRET ?? '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }'
```
camargo: We should not pass the `HASURA_GRAPHQL_JWT_SECRET` to the sequencing server since as a service the sequencing server should not be allowed to mint authentication tokens. Using the `HASURA_GRAPHQL_ADMIN_SECRET` should be enough since (for now) it gives this service the highest privilege  to query Hasrua.

cohansen (changes requested):
@camargo ping to check my comments as well. I think we should expect that the dev has their env configured properly rather than hardcoding?

sequencing-server/jestGlobalSetup.js
goetzrrGit: Got it I will remove this option.

sequencing-server/jestGlobalSetup.js
camargo: Yeah we should not include any hardcoded secrets.

goetzrrGit:
No changes need. :P

PR #940, opened 2023-05-15
Default Roles for Hasura

### LINKED ISSUE #861: Add default roles and permissions to Hasura
Requires https://github.com/NASA-AMMOS/aerie/issues/848 to be complete.

This task is to include the default roles and permissions in the Hasrua metadata. See [this document](https://docs.google.com/document/d/1sCIH2AhNDf6LXsSr6il44l_E4QFk3-ykH5SRPqGbdlw/edit) for specifics.

Mythicaeda:
* **Tickets addressed:** Partially Closes #861
* **Review:** By commit  

## Description

**This PR:**
- adds default roles and permissions for all of our DBs based on [this document](https://docs.google.com/document/d/1sCIH2AhNDf6LXsSr6il44l_E4QFk3-ykH5SRPqGbdlw/edit#)
- adds the `owner` column to `activity_preset` and `simulation_template`
- removes metadata from the `sequence` and `command_dictionary` tables based on discussion with @cohansen and @goetzrrGit
- removes dead code in the forms of:
     - `.yaml` files that do not correspond to existing DB tables and were not used in `tables.yaml`
     - unused DB tables

**Deviations from reference document:**
- Hasura does not allow use of cross-database relationships to inform permissions. Therefore, the "if plan owner/collaborator" restrictions on scheduling tables are not enforced and they are open to everyone with the "user" role
- Hasura does not allow fine-grained restriction to functions. One possible workaround is to untrack the functions and have them triggered via an insert on a table, as described [here](https://mnesarco.github.io/blog/2020/02/25/hasura-postgresql-function-as-mutation), but for this PR they have been left as available to everyone with the `user` role. Another workaround would be to perform the permissions checks in the hasura functions, but this may cause difficulties with allowing future Aerie customers to create roles, since they'd need to update and maintain SQL functions rather than just metadata.

**Breaking Changes:**
- The `activity_presets` table's object relationships were pluralized, which was confusing. The extra 's's at the end have been dropped.
- The signature of `begin_merge` and `create_merge_request` no longer take a username parameter. The value is now retrieved from the Hasura session
- The `command_dictionary` and `sequence` tables no longer have an `updated_at` field

## Verification

Verified by accessing the Hasura Admin Console and setting the `x-hasura-role` header to `user`, `viewer`, and `admin`, then seeing what GQL mutations and queries were available in the role's root, as described in [this Hasura article](https://hasura.io/learn/graphql/hasura/authorization/4-try-permission-queries/). Additionally performed manual tests for permissions where only an owner or a collaborator could perform a certain action. 

**Changes to existing tests:**
- Deparameterized Scheduler DB Tests
- Updated PlanMerge tests to use new signature of `duplicate_plan()` 
- Removed tests associated with removed metadata of `sequence` table

## Documentation

We should mention in our docs that these roles can be added to for deployment-specific permissions, and link to the Hasura docs for syntax reference.

## Future work

The deviations need to be resolved in such a way that it does not prevent a future Aerie Customer from adding mission-specific roles.


camargo (approved):
Looks like a great start! Let's target to get this merged with the front-end PR tomorrow morning.

mattdailis (approved):
Thanks for splitting this out into separate commits - that helped me quickly orient myself in this PR :+1: I wasn't able to dig into details - but from a high level everything looks good to me. Thanks for putting in all this work 🚀

PR #946, opened 2023-05-22
Help gradle find node for testing

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

@jmdelfa ran `./gradlew test`, and ran into the following error:

```
Cannot run program "node" (in directory "/home/jmdelfa/dev/aerie/merlin-server/constraints-dsl-compiler"): error=2, No such file or directory
```

What this error message means is that the `node` executable was not found on the `PATH`.

We can tell from this message that the `NODE_PATH` environment variable had assumed its "fallback" value of `"node"`, in the final if statement below:

https://github.com/NASA-AMMOS/aerie/blob/06182bf6945d6e15352012edd567b3fb9a4feb5d/merlin-server/build.gradle#L57C1-L64

What I _think_ this means is that the `NODE_PATH` environment variable does not immediately get set by `environment "NODE_PATH", ...` - which means the check `System.getenv("NODE_PATH")` returns `null` regardless of whether the loop above is able to successfully find the nodejs directory.

This PR uses a local variable, `node_path`, to handle this logic, and delays calling `environment` until the end of the block. This removes the assumption that the environment variable is updated in flight.

## Verification

I removed the `node` executable from my path, and ran `./gradlew test`, and reproduced the error mentioned above. I then made the changes included in this PR, and ran `./gradlew test` again, and the tests passed.

## Documentation

Open to ideas for whether this needs to be documented.

## Future work

None, at the moment.

### DIFF
--- a/merlin-server/build.gradle
+++ b/merlin-server/build.gradle
@@ -49,19 +49,14 @@ assemble {
 }
 
 test {
-  dependsOn(assembleConstraintsDSLCompiler)
+  dependsOn nodeSetup
+  dependsOn assembleConstraintsDSLCompiler
   useJUnitPlatform {
     includeEngines 'jqwik', 'junit-jupiter'
   }
 
   // Add node bin directory to PATH, helps CI/CD services without node installed
-  projectDir.toPath().resolve('.gradle/nodejs').toFile().listFiles().each {
-    environment 'NODE_PATH', "$it/bin/node"
-  }
-
-  if (System.getenv("NODE_PATH") == null) {
-    environment 'NODE_PATH', "node"
-  }
+  environment 'NODE_PATH', nodeSetup.nodeDir.get().toString().concat("/bin/node")
 
   environment "CONSTRAINTS_DSL_COMPILER_ROOT", projectDir.toPath().resolve('constraints-dsl-compiler')
   environment "CONSTRAINTS_DSL_COMPILER_COMMAND", './build/main.js'
--- a/scheduler-worker/build.gradle
+++ b/scheduler-worker/build.gradle
@@ -70,7 +70,8 @@ assemble {
 }
 
 test {
-  dependsOn(assembleSchedulingDSLCompiler)
+  dependsOn nodeSetup
+  dependsOn assembleSchedulingDSLCompiler
   useJUnitPlatform()
 
   testLogging {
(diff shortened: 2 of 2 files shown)


skovati (dismissed):
Nice fix! Tested on my local machine with `node` hidden from my `$PATH` and gradle is able to download and execute `node`.

skovati:
Looks like this passes CI since github actions VMs have `node` pre-installed

skovati:
Okay this is fun... running `./gradlew merlin-server:test` the first time (i.e. deleting `./.gradle` from the repo) results in the following error
```
> Task :merlin-server:test

MerlinBindingsTest > initializationError FAILED
    java.lang.Error at MerlinBindingsTest.java:39
        Caused by: java.io.IOException at MerlinBindingsTest.java:37
            Caused by: java.io.IOException at MerlinBindingsTest.java:37

ConstraintsDSLCompilationServiceTests > initializationError FAILED
    java.io.IOException at ConstraintsDSLCompilationServiceTests.java:71
        Caused by: java.io.IOException at ConstraintsDSLCompilationServiceTests.java:71
```

But then running `./gradlew merlin-server:test` a second time results in passed tests. Seems like gradle isn't able to find its own downloaded node files during the first run, but finds them during the filesystem traversal in the second run.

mattdailis:
Thanks for the investigation, @skovati - sounds like a classic case of gradle configuration time vs execution time: https://docs.gradle.org/current/userguide/build_lifecycle.html#sec:build_phases

The node executable doesn't _yet_ exist during configuration time - so the `NODE_PATH` is set to `node`, even though it _will_ exist by execution time.

skovati:
I've pushed a commit that always sets `NODE_PATH` to `nodeSetup.nodeDir.get().toString().concat("/bin/node")`, so that we are always referencing the version of `node` that `nodeSetup` downloaded.

This allows a "fresh" copy of the `aerie` repo (i.e. `rm -rf .gradle merlin-server/.gradle`) to successfully complete `./gradlew test`, even if `node` isn't installed locally.

merlin-server/build.gradle
```diff
@@ -49,22 +49,18 @@ assemble {
...
-  }
-
-  if (System.getenv("NODE_PATH") == null) {
-    environment 'NODE_PATH', "node"
-  }
+  environment 'NODE_PATH', nodeSetup.nodeDir.get().toString().concat("/bin/node")
 
   environment "CONSTRAINTS_DSL_COMPILER_ROOT", projectDir.toPath().resolve('constraints-dsl-compiler')
   environment "CONSTRAINTS_DSL_COMPILER_COMMAND", './build/main.js'
+
```
mattdailis: Rogue whitespace, otherwise looks great to me

mattdailis (commented):
Screenshot 2023-05-22 at 2 40 26 PM

To the extent that Github will let me, I approve @skovati 😝

merlin-server/build.gradle
skovati: Change has been made

skovati (approved):
Yes Github, my own commit looks good to me

PR #947, opened 2023-05-23
Refactor rootfinding to simulate less and catch simulation exceptions

### LINKED ISSUE #855: Catch all exceptions coming from simulation when rootfinding in the scheduler
The simulation that we use in rootfinding may throw an exception because the activity placement puts the model in an illegal state (which is expected because we are trying). Right now, we only catch a `SimulationException` but not other exceptions. This will make the rootfinding fail overall when it should continue trying.

### LINKED ISSUE #860: Rootfinding: avoid removing activities from simulation before assessing success
Right now, every time we are trying an activity placement in the rootfinding method, we place then remove the activity, triggering a restart of the simulation even though the activity might be the one we are going to end up using. Stop this and make it smarter.

adrienmaillard:
* **Tickets addressed:** Resolves #855  Resolves #860  
* **Review:** By commit  

## Description

Major stuff: 
- Rootfinding removes activity from simulation only if the iteration was unsuccessful. Otherwise, it is kept and will the activity will not be resimulated...
- ...even in `PrioritySolver` as `SimulationFacade.simulateActivity` now checks whether an activity has been simulated before. 
- I have deleted all the rootfinding algorithms we are not using. 
- Operation to remove-then-insert activities in simulation (introduced in commit 3, used in commit 7)
Imagine you have a plan A->B. You want to insert a C in between A and B and you use rootfinding. 
First iteration you end up with A->C1->B. C1 is not satisfying for rootfinding, now we want to try another act placement that we call C2. Until now, we were doing two independant operations remove() which lead to 1 sim for getting from A->C1->B to A->B and one other sim to go from A->B to A->C2 -> B because C2 is still in between A and B. Now we make a difference between simple removal and removal immediately followed by insertion. 
- Catching exceptions out of simulation 

Minor stuff: 
- Some code factorization in `SimulationFacade` and `ResumableSimulationDriver` (commit 2 and 7)
- refactored  activity insertion in `PrioritySolver`  
- Added/removed some log messages. Clipper asked for a message every time the simulation is restarted 

## Verification

- A test has been added with an activity throwing an exception for testing the rootfinding behavior. 
- Two tests that these new behaviors made obsolete were updated.  

## Documentation

No change, this is refactoring/bugfixing.

## Future work

Fix interface between STN, solver, and simulation in terms of interval semantics: 
- STN assumes closed intervals
- Solver works with open-closed intervals  
Problem is that right now, we translate open-closed into closed-closed intervals for stn propagation which may lead to incorrectly solutions.

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
@@ -305,7 +305,12 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
             try {
               facade.simulateActivity(actToSim);
               computedDuration = facade.getActivityDuration(actToSim);
-              history.add(new EventWithActivity(start,  start.plus(computedDuration.get()), actToSim));
+              if(computedDuration.isPresent()) {
+                history.add(new EventWithActivity(start, start.plus(computedDuration.get()), actToSim));
+              } else{
+                logger.debug("No simulation error but activity duration could not be found in simulation, unfinished activity?");
+                history.add(new EventWithActivity(start,  null, actToSim));
+              }
             } catch (SimulationFacade.SimulationException e) {
               logger.debug("Simulation error while trying to simulate activities: " + e);
               history.add(new EventWithActivity(start,  null, actToSim));
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
@@ -23,6 +23,7 @@ import gov.nasa.jpl.aerie.scheduler.solver.stn.TaskNetworkAdapter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
@@ -189,6 +190,22 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
     }
   }
 
+  public record EventWithActivity(Duration start, Duration end, SchedulingActivityDirective activity){}
+
+  public static class HistoryWithActivity{
+    List<EventWithActivity> events;
+
+    public HistoryWithActivity(){
+      events = new ArrayList<>();
+    }
+    public void add(EventWithActivity event){
+      this.events.add(event);
+    }
+    public Optional<EventWithActivity> getLastEvent(){
+      if(!events.isEmpty()) return Optional.of(events.get(events.size()-1));
+      return Optional.empty();
+    }
+  }
   /**
    * create activity if possible
    *
@@ -235,7 +252,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
     }
     final var success = tnw.solveConstraints();
     if (!success) {
-      logger.warn("Inconsistent temporal constraints, returning Optional.empty() instead of activity");
+      logger.warn("Inconsistent temporal constraints, will try next opportunity for activity placement if it exists");
       return Optional.empty();
     }
     final var solved = tnw.getAllData(name);
@@ -244,7 +261,8 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
     //now it is time to find an assignment compatible
     //CASE 1: activity has an uncontrollable duration
     if(this.type.getDurationType() instanceof DurationType.Uncontrollable){
-      final var f = new EquationSolvingAlgorithms.Function<Duration>(){
+      final var history = new HistoryWithActivity();
+      final var f = new EquationSolvingAlgorithms.Function<Duration, HistoryWithActivity>(){
         //As simulation is called, this is not an approximation
         @Override
         public boolean isApproximation(){
@@ -252,7 +270,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
         }
 
         @Override
-        public Duration valueAt(final Duration start) {
+        public Duration valueAt(Duration start, HistoryWithActivity history) {
           final var actToSim = SchedulingActivityDirective.of(
               type,
               start,
@@ -266,18 +284,38 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
               null,
               null,
               true);
-          try {
-            facade.simulateActivity(actToSim);
-            final var dur = facade.getActivityDuration(actToSim);
-            facade.removeActivitiesFromSimulation(List.of(actToSim));
-            return dur.map(start::plus).orElse(Duration.MAX_VALUE);
-          } catch (SimulationFacade.SimulationException e) {
-            return Duration.MAX_VALUE;
+          final var lastInsertion = history.getLastEvent();
+          Optional<Duration> computedDuration = Optional.empty();
+          if(lastInsertion.isPresent()){
+            try {
+              //remove and insert at the same time to avoid unnecessary potential resimulation.
+              //
+              // Current sim: A -> B1 -> C
+              // Sim at end of next iteration: A -> B2 -> C
+              //If we would do remove() and then insert(), we would simulation A -> C and then again A -> B2 -> C
+              facade.removeAndInsertActivitiesFromSimulation(List.of(lastInsertion.get().activity()), List.of(actToSim));
+              computedDuration = facade.getActivityDuration(actToSim);
+              //record insertion: if successful, it will stay in the simulation, otherwise, it'll get deleted at the next iteration
+              history.add(new EventWithActivity(start, start.plus(computedDuration.get()), actToSim));
+            } catch (SimulationFacade.SimulationException e) {
+              //still need to record so we can remove the activity at the next iteration
+              history.add(new EventWithActivity(start, null, actToSim));
+            }
+          } else {
+            try {
+              facade.simulateActivity(actToSim);
+              computedDuration = facade.getActivityDuration(actToSim);
+              history.add(new EventWithActivity(start,  start.plus(computedDuration.get()), actToSim));
+            } catch (SimulationFacade.SimulationException e) {
+              logger.debug("Simulation error while trying to simulate activities: " + e);
+              history.add(new EventWithActivity(start,  null, actToSim));
+            }
           }
+          return computedDuration.map(start::plus).orElse(Duration.MAX_VALUE);
         }
 
       };
-      return rootFindingHelper(f, solved, facade, evaluationEnvironment);
+      return rootFindingHelper(f, history, solved);
       //CASE 2: activity has a controllable duration
     } else if (this.type.getDurationType() instanceof DurationType.Controllable dt) {
       //select earliest start time, STN guarantees satisfiability
@@ -344,14 +382,15 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
           null,
           true));
     } else if (this.type.getDurationType() instanceof DurationType.Parametric dt) {
-      final var f = new EquationSolvingAlgorithms.Function<Duration>() {
+      final var history = new HistoryWithActivity();
+      final var f = new EquationSolvingAlgorithms.Function<Duration, HistoryWithActivity>() {
         @Override
         public boolean isApproximation(){
           return false;
         }
 
         @Override
-        public Duration valueAt(final Duration start) {
+        public Duration valueAt(final Duration start, final HistoryWithActivity history) {
           final var instantiatedArgs = SchedulingActivityDirective.instantiateArguments(
               arguments,
               start,
@@ -359,8 +398,17 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
               evaluationEnvironment,
               type
           );
+
           try {
-            return dt.durationFunction().apply(instantiatedArgs).plus(start);
+            final var duration = dt.durationFunction().apply(instantiatedArgs);
+            final var activity = SchedulingActivityDirective.of(type,start,
+                                                                duration,
+                                                                instantiatedArgs,
+                                                                null,
+                                                                null,
+                                                                true);
+            history.add(new EventWithActivity(start, start.plus(duration), activity));
+            return duration.plus(start);
           } catch (InstantiationException e) {
             logger.error("Cannot instantiate parameterized duration activity type: " + type.getName());
             throw new RuntimeException(e);
@@ -368,7 +416,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
         }
       };
 
-      return rootFindingHelper(f, solved, facade, evaluationEnvironment);
+      return rootFindingHelper(f, history, solved);
     } else {
      throw new UnsupportedOperationException("Unsupported duration type found: " + this.type.getDurationType());
     }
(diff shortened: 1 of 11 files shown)


scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -182,4 +183,40 @@ public void testBug(){
...
   }
+
+  @Test
+  public void testScheduleExceptionThrowingTask(){
+    final var zeroDurationUncontrollableActivity = new ActivityCreationTemplate.Builder()
+        .ofType(problem.getActivityType("LateRiser"))
+        .withTimingPrecision(Duration.of(1, Duration.MICROSECONDS))
+        .build();
+
+    //this time expression produces an interval [TimeAnchor.END, TimeAnchor.END + 2 Ms]
```
Mythicaeda, line 191:
```suggestion
    //this time expression produces an interval [TimeAnchor.START, TimeAnchor.START + 2 Ms]
```

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -182,4 +183,40 @@ public void testBug(){
...
+        .aliasForAnchors("its a me")
+        .build();
+
+    problem.setGoals(List.of(coexistenceControllable));
+    final var initialPlan = new PlanInMemory();
+    problem.setInitialPlan(initialPlan);
+
+    final var solver = new PrioritySolver(problem);
+    final var plan = solver.getNextSolution().get();
+    solver.printEvaluation();
```
Mythicaeda: Is this print necessary?

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -182,4 +183,40 @@ public void testBug(){
...
+    //Activity can be started in [0, 2m] but this activity will throw an exception if ran in [0, 1m] so it is scheduled at 2m (as being the second bounds the rootfinding tries before search).
```
Mythicaeda, line 211: Why does the activity end up scheduled at 2mins instead of sometime between 1 and 2 minutes?

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -2026,15 +2028,15 @@ void testOptionalSimulationAfterGoal_unsimulatedActivities() {
               new SchedulingGoal(new GoalId(0L), """
                   export default () => Goal.CoexistenceGoal({
                     forEach: ActivityExpression.ofType(ActivityTypes.GrowBanana),
-                    activityTemplate: ActivityTemplates.PeelBanana({peelDirection: "fromStem"}),
```
Mythicaeda, line 2029: How was the behavior of this and the other updated test obsoleted by the changes in this PR?

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -182,4 +183,40 @@ public void testBug(){
...
+        .build();
+
+    //this time expression produces an interval [TimeAnchor.END, TimeAnchor.END + 2 Ms]
+    final var intervalStartTimeExpression = new TimeExpressionRelativeFixed(TimeAnchor.START, false);
+    intervalStartTimeExpression.addOperation(TimeUtility.Operator.PLUS, Duration.of(2, Duration.MINUTE));
+
+    final var horizonExpression =  new SpansFromWindows(new WindowsWrapperExpression(new Windows(false).set(planningHorizon.getHor(), true)));
+
+    final var coexistenceControllable = new CoexistenceGoal.Builder()
+        .thereExistsOne(zeroDurationUncontrollableActivity)
```
Mythicaeda, line 198: If we're going to try and schedule off of the timeTrackerDaemon (instead of just checking it's value at a time in Simulation),`minutesElapsed` should be put into a cell instead of breaking encapsulation like it currently does. Otherwise, this scheduling goal is affected by the order in which the daemon vs the activity is run in simulation.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
```diff
@@ -266,18 +284,38 @@ public Duration valueAt(final Duration start) {
...
+              history.add(new EventWithActivity(start, start.plus(computedDuration.get()), actToSim));
+            } catch (SimulationFacade.SimulationException e) {
+              //still need to record so we can remove the activity at the next iteration
+              history.add(new EventWithActivity(start, null, actToSim));
+            }
+          } else {
+            try {
+              facade.simulateActivity(actToSim);
+              computedDuration = facade.getActivityDuration(actToSim);
+              history.add(new EventWithActivity(start,  start.plus(computedDuration.get()), actToSim));
```
Mythicaeda: What happens if `getActivityDuration` returned `Optional.empty()`?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -197,19 +216,16 @@ public static class SimulationException extends Exception {
...
+    if(insertedActivities.containsKey(activity)) return;
+    simulateActivities(List.of(activity));
   }
 
   public void computeSimulationResultsUntil(final Duration endTime) {
     var endTimeWithMargin = endTime;
     if(endTime.noLongerThan(Duration.MAX_VALUE.minus(MARGIN))){
       endTimeWithMargin = endTime.plus(MARGIN);
     }
+    //Simulation might throw an exception but we do not catch it because there is no countermeasure
```
Mythicaeda: Could you please elaborate on this? Additionally, if there is an Exception here, why do we not want to catch and rethrow it as a SimulationException?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -197,19 +216,16 @@ public static class SimulationException extends Exception {
   }
 
   public void simulateActivity(final SchedulingActivityDirective activity) throws SimulationException {
-    final var activityIdSim = new ActivityDirectiveId(itSimActivityId++);
-    final var activityDirective = schedulingActToActivityDir(activity);
-
-    planActDirectiveIdToSimulationActivityDirectiveId.put(activity.getId(), activityIdSim);
-    driver.simulateActivity(activityDirective, activityIdSim);
-    insertedActivities.put(activity, activityDirective);
+    if(insertedActivities.containsKey(activity)) return;
```
Mythicaeda, line 223: Is it possible/worth it to do a similar check on `simulateActivities`? I.E.: We remove from the Collection of `activitiesToInsert` the ones that are already in the plan, and return if the list is empty.

Mythicaeda (commented):
I'm going to run the profiler against this branch so we can visually show that we are running fewer simulations.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: because the rootfinding algorithm first tests the bounds of the domain

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: When we call this function, it is not because we have just inserted an activity that we could remove. If there is an exception, we do not know what to do but to crash. We could catch it and rethrow it as a SimulationException though, you're right.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: Sure, will do.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: I can remove them.

Mythicaeda:
I've finished the performance comparisons, both when Rootfinding is run by the scheduler and when it isn't.

## Setup:
- Mission Model: the `foo` model on the branch being run. This was safe, as the only difference is the addition of an ActivityType, which I did not use.
- Plan: Empty plan, 7 weeks long (Start Date: `2023-155T00:00:00.000`, End Date: `2023-204T00:00:00.000`)
- Goal: A Recurrence Goal that places one activity every 7 days. The specific goal used will be provided with its respective results.

Of note when looking at the screenshots: 
- Red lines mark when Scheduling Started and Ended
- Orange lines mark when the ResumableSchedulingDriver (RSD) was reinitialized (I added the same logging message that was added in the first commit of this PR to `develop` in order to be accurate)
- Purple lines mark when the RSD was restarted twice in the same instant
- The magenta line marks when the RSD was restarted three times in the same instant.

## Results 1: Rootfinding
Goal Used:
```typescript
export default function recurrenceGoalExample() {
    activityTemplate: ActivityTemplates.ZeroDurationUncontrollableActivity(),
    interval: Temporal.Duration.from({ days: 7 }),
  });
}
```
Zip File containing the original JProfiler Snapshots and Full-Size Screenshots of Results: [With_Rootfinding.zip](https://github.com/NASA-AMMOS/aerie/files/11692175/With_Rootfinding.zip)

Screenshot of this branch:
This Branch - With Rootfinding

Screenshot of Develop:
Develop - With Rootfinding

### Comparison:
| | This Branch | Develop |
|-:| :-: | :-: |
| **Times RSD Reinitialized:** | 7 | 21 |
| **Time to Run:** | 67 minutes | 93 minutes |

These results are as expected.

## Results 2: Without Rootfinding

Goal Used:
```typescript
export default function recurrenceGoalExample() {
    activityTemplate: ActivityTemplates.BasicFooActivity({
      duration: Temporal.Duration.from({ days: 1 }),
    }),
    interval: Temporal.Duration.from({ days: 7 }),
  });
```

Zip File containing the original JProfiler Snapshots and Full-Size Screenshots of Results: [No_Rootfinding.zip](https://github.com/NASA-AMMOS/aerie/files/11692280/No_Rootfinding.zip)


Screenshot of this branch:
This Branch - Without Rootfinding

Screenshot of Develop:
Develop - Without Rootfinding

### Comparison:
| | This Branch | Develop |
|-:| :-: | :-: |
| **Times RSD Reinitialized:** | 7 | 7 |
| **Time to Run:** | 66 minutes | 18 minutes |

While the number of Reinitializations are as expected, the time to run is very surprising, with this branch being three times slower than develop.

mattdailis:
That is intriguing - thank you for the detailed test, @Mythicaeda. Does anyone have any ideas what could be causing the non rootfinding case to take longer?

Mythicaeda:
No idea @mattdailis. @adrienmaillard's currently rerunning with CPU Profiling (I wasn't expecting such a large time difference, so I only instrumented memory).

We have also discovered that analyzing running in the container vs running a unit test has a significant time difference (the goals only took ~2 mins total in @adrienmaillard's unit tests, but took ~3 mins to reach the first driver reset in the container without any instrumentation/debugger running).

adrienmaillard:
Yes I ran the goals via test cases and found these timings when instrumenting the code with JProfiler
| | This Branch | Develop |
|-:| :-: | :-: |
| **Controllable goal** | 1:47 mins | 1:52 mins |
| **Uncontrollable goal** | 1:44 mins | 2:47 mins |
 
Instrumented the scheduler-worker with JProfiler, goals inputted via the UI 
| | This Branch | Develop |
|-:| :-: | :-: |
| **Controllable goal** | ? | ? |
| **Uncontrollable goal** | 32 mins | ? |

I could not pursue on the controllable one as my local deployment went into this jetty frenzy which I believe  is what I have already described in a post on slack and that Parker described in a ticket  (https://github.com/NASA-AMMOS/aerie/issues/971).

adrienmaillard:
scheduler_
Here is what I am talking about. Will post in the ticket as well.

mattdailis:
If it is indeed the same issue as the one you linked, perhaps something like this could help?

```graphql
mutation CancelPendingSchedulingRequests {
  update_scheduling_request(_set: {status: "failed"}, where: {status: {_eq: "pending"}}) {
    affected_rows
  }
}
```

adrienmaillard:
@Mythicaeda has ran the two goals again and here are the timings in minutes.
Inputted via the UI, no instrumentation 
| | This Branch | Develop |
|-:| :-: | :-: |
| **Controllable goal** | 16:40 | 16:12 |
| **Uncontrollable goal** | 10:26 | 24:30  |

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: The objective of this test is to test the simulation toggle that deactivates the simulation of activities at insertion. 
PeelBanana is an uncontrollable activity. With the new changes, the created activity stays in the simulation after rootfinding and thus is found when the second goal is evaluating the current plan. This is not what was expected in the original test where it was expected that the activity was removed from simulation and then not found. 
So I have replaced it with a parametric activity, that will not be simulated, thus that will not stay in the simulation and then not found by the second goal.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
adrienmaillard: I am replacing this try block with the following one to account for the case you mention:
```java
try {
  facade.simulateActivity(actToSim);
  computedDuration = facade.getActivityDuration(actToSim);
  if(computedDuration.isPresent()) {
    history.add(new EventWithActivity(start, start.plus(computedDuration.get()), actToSim));
  } else{
    logger.debug("No simulation error but activity duration could not be found in simulation, unfinished activity?");
    history.add(new EventWithActivity(start,  null, actToSim));
  }
} catch (SimulationFacade.SimulationException e) {
  logger.debug("Simulation error while trying to simulate activities: " + e);
  history.add(new EventWithActivity(start,  null, actToSim));
}
}
```

adrienmaillard:
@Mythicaeda I have adressed all your comments (5 more commits).

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
adrienmaillard: I have put `minutesElapsed` into a `Counter`.

Mythicaeda (approved):
LGTM!

Regarding performance: Since the massive slowdown was not seen when not instrumenting memory, I'm comfortable saying that this branch is no less performant than develop when not running rootfinding, and much more performant when it is.

PR #955, opened 2023-05-30
Set seq server owner fields

### LINKED ISSUE #936: Pass owner headers to sequencing server
We need the actions for the sequencing server to grab the owner header and set it in the DB.

cohansen:
* **Tickets addressed:** Closes #936 
* **Review:** By commit

## Description
There was only 1 spot that needed updating, which was setting the owner in the `expansion_run` table. I updated the way we grab the logged in user, so now we look at the logged in session (JWT token) or at the auth headers if the user makes a request directly against the sequencing server backend.

I also updated our tests to standardize the way we create a GraphQL client because we are now including a logged in `Test User`. I also changed uploading a command dictionary to only happen once per test file rather than before each test which seemed unnecessary, we'll see if this impacts test performance.

## Verification
Tests updated to include the Hasura logged in user to prevent failure on expansion run insert.

## Documentation
NA

## Future work
NA

### DIFF
--- a/sequencing-server/package-lock.json
+++ b/sequencing-server/package-lock.json
@@ -17,6 +17,7 @@
         "dataloader": "~2.2.0",
         "express": "~5.0.0-beta.1",
         "graphql-request": "~5.1.0",
+        "jwt-decode": "^3.1.2",
         "pg": "~8.9.0",
         "pg-format": "~1.0.4",
         "piscina": "~3.2.0",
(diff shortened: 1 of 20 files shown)


goetzrrGit (changes requested):
what do you think about combining `authSessionVariables.ts` and `jwt.ts` types into one type file called `jwtauthentication.ts` or something similar

cohansen:
> what do you think about combining `authSessionVariables.ts` and `jwt.ts` types into one type file called `jwtauthentication.ts` or something similar

Yeah that sounds fine, I can make that change!

goetzrrGit (approved):

PR #956, opened 2023-05-30
Added ability to specify a rate for a SampledResource

ewferg:
* **Tickets addressed:** None
* **Review:** By commit  

## Description
The example SampledResource in the contrib package was a little too simple in that it did not provide the capability to sample at a user specified rate. This PR updates the SampledResource so that users can both initialize a resource to a custom sample rate and change the rate later if they wish.  

## Verification
No tests were added and the original resource constructor still exists, which defaults the rate to one second. 

## Documentation
The [Resource and Models](https://nasa-ammos.github.io/aerie-docs/mission-modeling/resources-and-models/) section of the docs should be updated, but my plan was to address this update as part of https://github.com/NASA-AMMOS/aerie-docs/issues/46

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
@@ -10,37 +10,36 @@ import java.util.function.UnaryOperator;
 import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.*;
 
 /**
- * Simple resource that samples arbitrarily many existing resources/values at a specified rate (default rate is once per
+ * Simple resource that samples arbitrarily many existing resources/values at a specified period (default period is once per
  * second).
  */
 public class SampledResource<T> implements DiscreteResource<T> {
   private final Register<T> result;
   private final Supplier<T> sampler;
-  private double rate;
+  private Register<Double> period = Register.forImmutable(1.0);
 
   /**
-   * Constructor that does not require caller to specify a rate and therefore assumes a sample
-   * rate of 1 sample per second.
+   * Constructor that does not require caller to specify a period and therefore assumes a sample
+   * period of 1 sample per second.
    */
   public SampledResource(final Supplier<T> sampler, final UnaryOperator<T> duplicator) {
     this.result = Register.create(sampler.get(), duplicator);
     this.sampler = Objects.requireNonNull(sampler);
-    this.rate = 1.0;
     spawn(this::takeSamples);
   }
 
   /**
-   * Constructor that requires caller to specify an initial sample rate
+   * Constructor that requires caller to specify an initial sample period
    */
-  public SampledResource(final Supplier<T> sampler, final UnaryOperator<T> duplicator, final double rate) {
+  public SampledResource(final Supplier<T> sampler, final UnaryOperator<T> duplicator, final double period) {
     this.result = Register.create(sampler.get(), duplicator);
     this.sampler = Objects.requireNonNull(sampler);
-    this.rate = rate;
+    this.period.set(period);
     spawn(this::takeSamples);
   }
 
   /**
-   * Method that samples the supplied resource at the current specified rate
+   * Method that samples the supplied resource at the current specified period
    */
   private void takeSamples() {
     while (true) {
@@ -48,20 +47,20 @@ public class SampledResource<T> implements DiscreteResource<T> {
       if (!result.get().equals(sample)) {
         result.set(sample);
       }
-      delay( Duration.roundNearest(rate, Duration.SECONDS) );
+      delay( Duration.roundNearest(period.get(), Duration.SECONDS) );
     }
   }
 
   /**
-   * Get current sample rate
+   * Get current sample period
    */
-  public double getRate() { return rate; }
+  public double getPeriod() { return period.get(); }
 
   /**
-   * Method to adjust the specified rate of sampling. Note if takeSamples() is currently waiting, the
-   * new rate will not take effect until after the current wait cycle.
+   * Method to adjust the specified period of sampling. Note if takeSamples() is currently waiting, the
+   * new period will not take effect until after the current wait cycle.
    */
-  public void setRate(final double newRate ) { rate = newRate; }
+  public void setPeriod(final double newPeriod ) { period.set( newPeriod ); }
 
   @Override
   public T getDynamics() {
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
@@ -1,34 +1,67 @@
 package gov.nasa.jpl.aerie.contrib.models;
 
 import gov.nasa.jpl.aerie.merlin.framework.resources.discrete.DiscreteResource;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 
 import java.util.Objects;
 import java.util.function.Supplier;
 import java.util.function.UnaryOperator;
 
 import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.*;
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
 
+/**
+ * Simple resource that samples arbitrarily many existing resources/values at a specified rate (default rate is once per
+ * second).
+ */
 public class SampledResource<T> implements DiscreteResource<T> {
   private final Register<T> result;
   private final Supplier<T> sampler;
+  private double rate;
 
+  /**
+   * Constructor that does not require caller to specify a rate and therefore assumes a sample
+   * rate of 1 sample per second.
+   */
   public SampledResource(final Supplier<T> sampler, final UnaryOperator<T> duplicator) {
     this.result = Register.create(sampler.get(), duplicator);
     this.sampler = Objects.requireNonNull(sampler);
-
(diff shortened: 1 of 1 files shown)


contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
```diff
@@ -1,35 +1,68 @@
...
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
 
+/**
+ * Simple resource that samples arbitrarily many existing resources/values at a specified rate (default rate is once per
+ * second).
+ */
 public class SampledResource<T> implements DiscreteResource<T> {
   private final Register<T> result;
   private final Supplier<T> sampler;
+  private double rate;
```
mattdailis: Strictly speaking, we're talking about a "period" rather than a "rate" - any thoughts on what word would cause the least confusion?

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
```diff
@@ -1,35 +1,68 @@
...
 
+  /**
+   * Constructor that does not require caller to specify a rate and therefore assumes a sample
+   * rate of 1 sample per second.
+   */
   public SampledResource(final Supplier<T> sampler, final UnaryOperator<T> duplicator) {
```
mattdailis, line 25: I like how you preserved backwards compatibility here :+1: That makes this update straightforward for anyone currently relying on this.

That said - your change makes this class more useful - I have my doubts that it was getting heavy use before.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
```diff
@@ -1,35 +1,68 @@
...
       if (!result.get().equals(sample)) {
         result.set(sample);
       }
-      delay(1, SECOND);
+      delay( Duration.roundNearest(rate, Duration.SECONDS) );
     }
   }
 
+  /**
+   * Get current sample rate
```
mattdailis: Since this is returning a double, we should specify the units in this comment

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
```diff
@@ -1,35 +1,68 @@
...
+   */
+  public double getRate() { return rate; }
+
+  /**
+   * Method to adjust the specified rate of sampling. Note if takeSamples() is currently waiting, the
+   * new rate will not take effect until after the current wait cycle.
+   */
+  public void setRate(final double newRate ) { rate = newRate; }
```
mattdailis: I'm pondering whether `rate` should be tracked in a Cell, or if it's okay as a field 🤔 To err on the safe side I would recommend wrapping `rate` in a `Register`. 

The hazard here is that if Task A and Task B are executing concurrently - and Task A calls `setRate`, while Task B calls `getRate`, we can accidentally create a causal dependency that is invisible to Aerie.

However, if by policy this SampledResource's rate is accessed by a single daemon task, you may be okay with it as is. Pinging @Twisol in case he'd like to weigh in

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: Good point, we could also do something as simple as "dt"

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: Pardon my ignorance, but what are the implications of this causal dependency on the Aerie system? I'll make the update nonetheless.

Twisol: Yes, in general all mutable state in a Merlin model needs to be mediated by a Cell managed by Merlin. There are a couple of reasons for this, but a big one is that various parts of a simulated task might be repeated for any number of reasons, and we need to idempotize any changes to state so that they only take effect the first time they occur.

Probably the most visible and accessible way this can arise in the current system is through the use of `ReplayingTask`. The default mode of executing tasks wraps each task in its own thread; this is called a `ThreadedTask`. However, the overhead of context-switching, especially for tasks that only execute for a brief amount of time before yielding control again, adds up very quickly. For many kinds of task, using a replaying strategy is faster: when a task yields, it actually throws an exception to bail out; and when the task resumes, we call into it from the top again and run it forward until it gets back to where it was before, skipping any effects it performed along the way. It might seem wasteful to re-run the same code, but for "most" tasks, the overhead of rerunning the logic is drastically outweighed by the overhead of context-switching.

There are more "correctness"-oriented reasons, too: any resource exposed from a model is polled by the simulation host only when we detect changes to the Cells it depends on. If a resource is computed from unmanaged state, either in whole or in part, we cannot record all of the changes that ought to be reported to planners. We could poll, instead, but the specter of performance raises its head again: polling is extremely expensive overall, and checking a resource only when we *know* we need to reduces simulation overhead considerably.

Our approach also means that for many resources, sampling rates don't need to be managed explicitly; and for many others, there is a design path involving the `Cell.getExpiry` function to tell the simulation host when the last read should be considered invalidated by default, such as for complex polynomial resources: since we can only currently report linear segments, at some point the last linear segment will fall too far out of line with the actual curve, so we can say that the segment "expires" after a certain amount of time. This is also more performant than either `ThreadedTask` or `ReplayingTask`: the former incurs a context switch, and the latter incurs at least the cost of idempotized effect history (which would grow with every sample!). Resource expiration can be handled directly from the simulation host, keeping the task execution machinery off the hot path altogether.

Lastly, and most future-oriented, proposed incremental resimulation strategies uniformly depend on us either (or both) having a complete view on the state used by a model (so that we can snapshot and restore it as needed), or be able to re-run only the tasks which depend on data that has changed -- if some model state is not affected by a change in a plan, then tasks that depend on it may not need to be re-run at all. If unmanaged mutable state exists, then it is very easy to end up in a situation where the behavior of the model doesn't line up with what it would have done in a fresh, non-incremental run. This kind of divergence can completely neuter the benefit of incremental resimulation: it doesn't matter how fast it is if it's wrong!

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
Twisol: > That said - your change makes this class more useful - I have my doubts that it was getting heavy use before.

Daemon tasks -- especially threaded ones -- will incur some pretty decent task overhead, especially at high sampling rates. Ideally, `SampledResource` should be a tool of last resort. There will always be some legitimate use-cases that require it; but I'm not in a hurry to market it as a first-line solution either.

(to be clear, I like the changes in this class; i just wanted to provide some commentary)

ewferg: While I don't disagree with you, I will also say that as a new "modeler" to Aerie, this resource was a tool of first resort because it is easy to understand and easy to use. Moreover, there aren't many other out of the box resources that I can grab and use that are simple to define. Without a nice collection of first-line solution resources at a modelers disposal, they will likely be lazy like me and use what they have available to them.

dyst5422: I agree with @ewferg as it comes to our API, we need to make the easy thing, the right thing.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: I went your suggested term, period.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: No immediate action here, so I will close this out

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: Done.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
```diff
@@ -1,35 +1,67 @@
...
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
 
+/**
+ * Simple resource that samples arbitrarily many existing resources/values at a specified period (default period is once
+ * per second).
+ */
 public class SampledResource<T> implements DiscreteResource<T> {
   private final Register<T> result;
   private final Supplier<T> sampler;
+  private Register<Double> period = Register.forImmutable(1.0);
```
mattdailis: I believe this field can now be `final`, since the mutation has been encapsulated in the Register

mattdailis (approved):
Looks good to me, thanks for making those updates. One minor update requested, but giving my approval now.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
ewferg: Thanks for the detailed explanation!

PR #960, opened 2023-05-31
Make `Description` optional across the DB

### LINKED ISSUE #950: Make constraint description field optional
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

1.7.0

### Describe the bug

The description field for Constraint https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-server/sql/merlin/tables/constraint.sql#L5 should be optional to be consistent with other description fields.

### Reproduction

N/A

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

Moderate

Mythicaeda:
* **Tickets addressed:** Closes #950
* **Review:** By commit  

## Description

As pointed out in #950, there were places where there was an optional `description` column that was specified as not optional (as in, not null, no default). These columns have been updated to have a default value of `''`. Additionally, the scheduler was representing this optionality as `NULL` in two places. Those locations have been updated to match with the rest of the DB.

## Verification

No existing tests would be affected by this change.

### DIFF
--- a/deployment/hasura/migrations/AerieMerlin/15_defintion_consistency/down.sql
+++ b/deployment/hasura/migrations/AerieMerlin/15_defintion_consistency/down.sql
@@ -1,0 +1,7 @@
+alter table public.simulation_template
+  alter column description drop default;
+
+alter table public."constraint"
+  alter column description drop default;
+
+call migrations.mark_migration_rolled_back('15');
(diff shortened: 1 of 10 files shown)


mattdailis (approved):
Looks good to me 👍

PR #961, opened 2023-05-31
Fixed the issue where constraint violations were not being calculated properly

### LINKED ISSUE #919: Extraneous constraint violations
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

1.5.0

### Describe the bug

While testing a simple constraint with a linear resource profile, the UI and Hasure console report erroneous violations.

In particular, the profile oscillates between 1000 and 1,000,000, and the constraint demands that the value be greater than or equal to 500. There should be no violations, but the UI flags the entirety of every increasing segment as a violation. This behavior is very sensitive to the threshold in the constraint. Setting the constraint threshold to 400 or 600 results in no violations, the expected result.

### Reproduction

Create a profile with the following segments, named "/geometry/altitude_europa_km":
```
          {
            "start_offset": "00:00:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 250750
            }
          },
          {
            "start_offset": "00:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "01:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "02:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "03:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "04:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "05:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "06:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "07:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "08:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "09:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "10:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "11:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "12:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "13:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "14:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "15:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "16:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "17:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "18:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "19:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "20:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "21:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          },
          {
            "start_offset": "22:45:00",
            "dynamics": {
              "rate": -277.5,
              "initial": 1000000
            }
          },
          {
            "start_offset": "23:45:00",
            "dynamics": {
              "rate": 277.5,
              "initial": 1000
            }
          }
```

Then, create a constraint like this:
```
  return Real.Resource('/geometry/altitude_europa_km').greaterThan(500);
}
```

Finally, check check constraints, and you should see a UI like this:
image

### Logs

_No response_

### System Info

```shell
Running on a local Aerie deployment, using the Clipper sched-point prototype project.
```

### Severity

Moderate

cohansen:
* **Tickets addressed:** Closes #919
* **Review:** By commit

## Description
@mattdailis and I walked through the code and tracked down the issue to be not handling negative durations properly. This was causing the segments to be off by 1 second causing the bug. We added a function to the `Duration` class that ensures the modulo is always positive which fixed the issue.

## Verification
Manual testing was done and we updated the unit tests to include negative number checking.

## Documentation
NA

## Future work
NA

### DIFF
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
@@ -181,9 +181,11 @@ public final class Duration implements Comparable<Duration> {
    * rounding to the nearest representable value above.
    */
   public static Duration roundUpward(final double quantity, final Duration unit) {
+    final var integerPart = Math.floor(quantity);
+    final var decimalPart = quantity - integerPart;
     return add(
-        unit.times((long) Math.floor(quantity)),
-        EPSILON.times((long) Math.ceil(positiveModulo(quantity, 1) * unit.dividedBy(EPSILON))));
+        unit.times((long) integerPart),
+        EPSILON.times((long) Math.ceil(decimalPart * unit.dividedBy(EPSILON))));
   }
 
   /**
@@ -191,9 +193,11 @@ public final class Duration implements Comparable<Duration> {
    * rounding to the nearest representable value below.
    */
   public static Duration roundDownward(final double quantity, final Duration unit) {
+    final var integerPart = Math.floor(quantity);
+    final var decimalPart = quantity - integerPart;
     return add(
-        unit.times((long) Math.floor(quantity)),
-        EPSILON.times((long) Math.floor(positiveModulo(quantity, 1) * unit.dividedBy(EPSILON))));
+        unit.times((long) integerPart),
+        EPSILON.times((long) Math.floor(decimalPart * unit.dividedBy(EPSILON))));
   }
 
   /**
@@ -201,9 +205,11 @@ public final class Duration implements Comparable<Duration> {
    * rounding to the nearest representable value.
    */
   public static Duration roundNearest(final double quantity, final Duration unit) {
+    final var integerPart = Math.floor(quantity);
+    final var decimalPart = quantity - integerPart;
     return add(
-        unit.times((long) Math.floor(quantity)),
-        EPSILON.times((long) Math.rint(positiveModulo(quantity, 1) * unit.dividedBy(EPSILON))));
+        unit.times((long) integerPart),
+        EPSILON.times((long) Math.rint(decimalPart * unit.dividedBy(EPSILON))));
   }
 
   /**
@@ -537,8 +543,4 @@ public final class Duration implements Comparable<Duration> {
     return Long.compare(this.durationInMicroseconds, other.durationInMicroseconds);
   }
 
-  /** Returns the remainder of dividend / divisor as a value between 0 and divisor. */
-  private static double positiveModulo(final double dividend, final double divisor) {
-    return ((dividend % divisor) + divisor) % divisor;
-  }
 }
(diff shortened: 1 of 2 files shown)


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -536,4 +536,9 @@ public String toString() {
   public int compareTo(final Duration other) {
     return Long.compare(this.durationInMicroseconds, other.durationInMicroseconds);
   }
+
+  /** Returns the remainder of dividend / divisor as a value between 0 and divisor. */
+  private static double positiveModulo(final double dividend, final double divisor) {
+    return ((dividend % divisor) + divisor) % divisor;
+  }
```
Twisol: why is java like this 💀

Alexander-Da-Great:
Tested this fix and it solved a bug I was experiencing with extraneous constraint failures in regions where they should not have been failing.

mattdailis (approved):

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -536,4 +542,5 @@ public String toString() {
   public int compareTo(final Duration other) {
     return Long.compare(this.durationInMicroseconds, other.durationInMicroseconds);
   }
+
```
Mythicaeda, line 545: Blank new line accidentally got inserted

Mythicaeda (approved):

mattdailis:
Here is the diagram we drew up during a quick PR review session. What this is trying to convey is how the behavior of `roundNearest` on negative numbers results in incorrect constraint violations.

1. The intersection time of the two lines is calculated as a double number of seconds
2. `roundNearest` is used to convert that to a `Duration`. Due to a bug, this returns an erroneous value
3. The LinearEquation comparison logic thinks it's testing the lines on different sides of the intersection - but in reality is testing both lines on the same side of the intersection
4. The windows expression returns a single window spanning all time - since all of the comparisons happened on the same side of the intersection and returned the same value

Screenshot 2023-06-06 at 12 16 26 PM

PR #880, opened 2023-04-19
Add Metadata Columns in DB

Mythicaeda:
* **Tickets addressed:** Closes #854, #857, and #795.
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

The first three commits adds metadata columns to tables throughout our databases. The fourth commit addresses #857.

Two things of note:
1. Plans have one owner and multiple collaborators.
2. `activity_types` gained the metadata `subsystem`, which is a nullable text column.

## Verification

No tests need to be updated, as no functionality changed.

Verification that #857 has been resolved was completed by manually checking via the steps defined [here](https://github.com/NASA-AMMOS/aerie/issues/857#issuecomment-1505964717).

## Documentation

No documentation should need updating.

## Future work

Adding the `subsystem` column should help with filtering activities in the UI (and potentially in API queries), although work would need to be done in the backend to allow mission models to specify activity type subsystems. Maybe another Annotation, `@Subsystem("subsystem")`?

PR #962, opened 2023-05-31
Add field to condition trigger of scheduler worker after insertion on scheduling request

adrienmaillard:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

PR #880 merged `analysis_id` into the `scheduling_request` table so now we are forced to create a request to have an analysis. This broke how the Clipper external scheduler works and I predict that's going to be the same for any external scheduler that uses the UI.

Specifically, when we update the plan, we push a scheduling analysis to the UI so the user can visualize it. Creating a `scheduling_request` as is does not work as it triggers automatically the scheduler workers. In this PR, I am adding a field to condition this triggering to the value of a new field in the request.

## Verification

No functionality has changed, the field has a default value equal to true.

## Documentation

None. 

## Future work

None.

### DIFF
--- a/deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/down.sql
+++ b/deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/down.sql
@@ -1,0 +1,14 @@
+-- Drop Triggers
+drop trigger notify_scheduler_workers on scheduling_request;
+
+alter table scheduling_request
+  drop column for_aerie_scheduler;
+
+comment on column scheduling_goal.for_aerie_scheduler is null;
+
+create trigger notify_scheduler_workers
+  after insert on scheduling_request
+  for each row
+  execute function notify_scheduler_workers();
+
+call migrations.mark_migration_rolled_back('8');
--- a/deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/up.sql
+++ b/deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/up.sql
@@ -1,0 +1,16 @@
+-- Drop Triggers
+drop trigger notify_scheduler_workers on scheduling_request;
+
+alter table scheduling_request
+  add column for_aerie_scheduler boolean not null default true;
+
+comment on column scheduling_request.for_aerie_scheduler is e''
+  'Whether the request is destined to the aerie scheduler or to an external scheduler;';
+
+create trigger notify_scheduler_workers
+  after insert on scheduling_request
+  for each row
+  when (NEW.for_aerie_scheduler)
+  execute function notify_scheduler_workers();
+
+call migrations.mark_migration_applied('8');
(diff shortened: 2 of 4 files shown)


deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/up.sql
```diff
@@ -0,0 +1,13 @@
+-- Drop Triggers
+drop trigger notify_scheduler_workers on scheduling_request;
+
+alter table scheduling_request
+  add column for_aerie_scheduler boolean default true;
```
Mythicaeda: This statement is missing a `not null`.

scheduler-server/sql/scheduler/tables/scheduling_request.sql
```diff
@@ -34,6 +35,8 @@ comment on column scheduling_request.status is e''
   'The state of the the scheduling request.';
 comment on column scheduling_request.reason is e''
   'The reason for failure when a scheduling request fails.';
+comment on column scheduling_request.for_aerie_scheduler is e''
+  'Whether the request is destined to the aerie scheduler or to an external scheduler;';
```
Mythicaeda, line 39:
```suggestion
  'Whether the request is destined to the aerie scheduler or to an external scheduler.';
```

deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/down.sql
```diff
@@ -0,0 +1,14 @@
+-- Drop Triggers
+drop trigger notify_scheduler_workers on scheduling_request;
+
+alter table scheduling_request
+  drop column for_aerie_scheduler;
+
+comment on column scheduling_goal.for_aerie_scheduler is null;
```
Mythicaeda, line 7: This statement needs to go before the column is dropped.

scheduler-server/sql/scheduler/tables/scheduling_request.sql
```diff
@@ -10,6 +10,7 @@ create table scheduling_request (
   reason jsonb null,
   canceled boolean not null default false,
   dataset_id integer default null,
+  for_aerie_scheduler boolean not null default true,
```
Mythicaeda, line 13: Proposal (ignorable):
```suggestion
  use_external_scheduler boolean not null default false,
```
```sql
comment on column scheduling_request.use_external_scheduler is e''
 'Whether the request will be (or has been) handled by an external scheduler.';
```

Mythicaeda (commented):
Heads up, the Hasura metadata needs to be updated so that this new column can be selected/inserted into.

Otherwise, the design looks good.

deployment/hasura/migrations/AerieScheduler/8_add_condition_trigger_scheduler/up.sql
adrienmaillard: yep, I noticed that in the meantime ;)

adrienmaillard:
A discussion with @Mythicaeda and @mattdailis lead to a workaround: we create scheduling requests with status equal to `success` which avoids for the scheduler to claim them and act on it. I have tested and confirmed the requests are not claimed, which is due to the `ClaimRequestAction` called in `PostgresResultsCellRepository` that will fail and throw a `UnclaimableRequestException` if the status of the request is not `pending`.

PR #963, opened 2023-06-01
Tweak sampled resource initialization

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This PR is a minor adjustment to #956 that slipped my mind when reviewing yesterday. Cell allocation and updates are not allowed to occur during the same phase of simulation (Allocation must occur during initialization, updates must occur during task execution).  This PR moves the allocation into the SampledResource constructor - using the user-provided period as the initial value for the register, removing the need for an update during initialization.

## Verification

I added a simple `SampledResourceTest` to exercise this code a little bit.

## Documentation

See documentation section of the original PR #956 

## Future work

Sampled resources, while a good example of a daemon tasks, are pretty inefficient. @Twisol has some ideas for some better tools we can provide out of the box to address some use cases that currently require daemon tasks.

### DIFF
--- a/contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
+++ b/contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
@@ -1,0 +1,31 @@
+package gov.nasa.jpl.aerie.contrib.models;
+
+import gov.nasa.jpl.aerie.merlin.framework.junit.MerlinExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.delay;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECONDS;
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith(MerlinExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class SampledResourceTest {
+  final Accumulator accumulator = new Accumulator(0.0, 1.0);
+  final SampledResource<Double> sampledResource = new SampledResource<>(() -> accumulator.get(), $ -> $, 10.0);
+
+  @Test
+  void testSampledResource() {
+    assertEquals(0.0, sampledResource.get());
+    delay(1, SECOND);
+    assertEquals(0.0, sampledResource.get());
+    delay(10, SECONDS);
+    assertEquals(10.0, sampledResource.get());
+    delay(5, SECONDS);
+    assertEquals(10.0, sampledResource.get());
+    delay(5, SECONDS);
+    assertEquals(20.0, sampledResource.get());
+  }
+}
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/models/SampledResource.java
@@ -16,16 +16,14 @@ import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.*;
 public class SampledResource<T> implements DiscreteResource<T> {
   private final Register<T> result;
   private final Supplier<T> sampler;
-  private final Register<Double> period = Register.forImmutable(1.0);
+  private final Register<Double> period;
 
   /**
    * Constructor that does not require caller to specify a period and therefore assumes a sample
    * period of 1 sample per second.
    */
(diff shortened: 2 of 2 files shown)


Twisol:
Just some random thoughts around this situation:

> Allocation must occur during initialization

Although it wouldn't solve things here (since the problem is with the event emission at initialization time), I'd like to make it possible to allocate cells during the simulation, too. I think @JoelCourtney may have had some concerns around other features this might complicate though 🤔

> updates must occur during task execution

Emitting events during initialization doesn't seem impossible to allow, and I don't think it would cause any new problems. The fact that we don't allow it right now comes down to the fact that we didn't know all the ramifications of the current design, and stayed somewhat conservative to keep our options open. But I think this is something that could be pursued. (We originally didn't allow cells to be queried during initialization, either -- we'd need to make sure those queries see any initialization-time events if we went in this direction.)

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.delay;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.SECONDS;
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith(MerlinExtension.class)
```
Mythicaeda, line 13: What does this `@ExtendWith` do?

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
```diff
@@ -0,0 +1,31 @@
...
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class SampledResourceTest {
+  final Accumulator accumulator = new Accumulator(0.0, 1.0);
+  final SampledResource<Double> sampledResource = new SampledResource<>(() -> accumulator.get(), $ -> $, 10.0);
```
Mythicaeda, line 17: What does `$->$` mean?

Adjacently, based on the test, this line appears to be making a `SampledResource` that samples an accumulator that accumulates at a rate of once per second every ten seconds. Is that correct?

ewferg:
Looks good! Thanks for having my back @mattdailis - next time I will test like a good dev should. :)

ewferg (approved):

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
mattdailis: Here are the relevant docs:
> `@ExtendWith` [...] is used to register [extensions](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/Extension.html) for the annotated test class

https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtendWith.html

`MerlinExtension` is a JUnit extension we provide that makes it possible to unit test mission models within a simulation environment. Every test method is wrapped in a call to `SimulationDriver.simulateTask` - which is how we're able to delay, emit events, and query cells in the test method below.

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
mattdailis: `$ -> $` is the identity function - the `SampledResource` (more specifically the underlying register), asks for a way to duplicate the contained value. If the type contained in the SampledResource were mutable, this function would be expected to make a new copy. Since it's a `Double`, the identity function suffices here

mattdailis: > Adjacently, based on the test, this line appears to be making a SampledResource that samples an accumulator that accumulates at a rate of once per second every ten seconds. Is that correct?

Your understanding is accurate :+1:

contrib/src/test/java/gov/nasa/jpl/aerie/contrib/models/SampledResourceTest.java
Mythicaeda: > MerlinExtension is a JUnit extension we provide that makes it possible to unit test mission models within a simulation environment. Every test method is wrapped in a call to SimulationDriver.simulateTask - which is how we're able to delay, emit events, and query cells in the test method below.

Thanks! I'd found the JUnit docs you linked but that hadn't helped since they assume you're already familiar with JUnit extensions. This explains pretty much every question I had with the test 👍

Mythicaeda (approved):

PR #964, opened 2023-06-01
Update postgres base image to 14.8

skovati:
* **Tickets addressed:** n/a
* **Review:** By commit  

## Description
The base image that `aerie-postgres` is based on is susceptible to several critical security vulnerabilities, which can be seen in the [results of our security scan](https://github.com/NASA-AMMOS/aerie/actions/runs/5146978225) in `publish`. Updating our base image to 14.8 fixes all of the detected vulnerabilities.

## Verification
Mainly relying on e2e tests passing, although according to the [postgres versioning policy](https://www.postgresql.org/support/versioning/), no breaking changes are introduced between minor versions, and all the release notes for versions 14.1 to 14.8 say a dump/restore is not needed when upgrading.

## Documentation
None

## Future work
Change `deployment/docker-compose.yml` to use `aerie-postgres` instead of bind mounting metadata, which would remove the need to clone aerie or download `deployment.zip` when doing a non-custom Aerie deployment.

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -167,7 +167,7 @@ services:
       POSTGRES_USER: "${POSTGRES_USER}"
       AERIE_USERNAME: "${AERIE_USERNAME}"
       AERIE_PASSWORD: "${AERIE_PASSWORD}"
-    image: postgres:14.1
+    image: postgres:14.8
     ports: ["5432:5432"]
     restart: always
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -244,7 +244,7 @@ services:
       POSTGRES_USER: "${POSTGRES_USER}"
       AERIE_USERNAME: "${AERIE_USERNAME}"
       AERIE_PASSWORD: "${AERIE_PASSWORD}"
-    image: postgres:14.1
+    image: postgres:14.8
     ports: ["5432:5432"]
     restart: always
     volumes:
--- a/docker/Dockerfile.postgres
+++ b/docker/Dockerfile.postgres
@@ -1,2 +1,2 @@
-FROM postgres:14.1
+FROM postgres:14.8
 COPY deployment/postgres-init-db /docker-entrypoint-initdb.d
--- a/e2e-tests/docker-compose-test.yml
+++ b/e2e-tests/docker-compose-test.yml
@@ -241,7 +241,7 @@ services:
       POSTGRES_USER: '${POSTGRES_USER}'
       AERIE_USERNAME: '${AERIE_USERNAME}'
       AERIE_PASSWORD: '${AERIE_PASSWORD}'
-    image: postgres:14.1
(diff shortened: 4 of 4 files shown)


mattdailis (commented):
I agree that passing tests should be sufficient for this PR :+1:

There are also some bug fixes between 14.1 and 14.8 - we haven't been in the habit of using `REINDEX CONCURRENTLY`, so I don't think we are affected by them - but it's nice to know they've been fixed.

skovati:
@mattdailis I also looked into upgrading our image to the next major version (15.3) for this PR, but it looks like there are a few breaking changes that we'd have to iron out.

Mythicaeda:
> I also looked into upgrading our image to the next major version (15.3) for this PR, but it looks like there are a few breaking changes that we'd have to iron out.

Oh, yeah, I looked into moving us onto 15 awhile back. The big change that needed to be made from AFAIR is that 15 [changed CREATE permissions on the public schema](https://www.postgresql.org/docs/15/release-15.html), which breaks our `init.db` script. Did you find something different?

Mythicaeda (approved):
I agree, tests passing is sufficient here. Have you tested upgrading the DB container (as in, spin up Aerie with 14.1, put some things in the DB, then replace the 14.1 container with a 14.8 container)?

skovati:
Ooh good idea, let me do some manual testing quick 👍

skovati:
@Mythicaeda I didn't see any other pertinent breaking changes besides the public schema default permissions, should we create a ticket to look into upgrading to 15?

On a somewhat related note, we could also look into upgrading to JDK 20 in a separate ticket, since JDK 19's official support ended a few months ago (each non-LTS release is only supported for 6 months)

Also, I did a few manual tests like you mentioned, creating models, plans, constraints, views, etc on a 14.1 volume, and made sure they worked fine when using that same volume on 14.8 👍

camargo:
Just to be safe, before merging can we also make sure the UI e2e tests pass?

skovati:
e2e UI tests were being flaky on my local machine, so I opened a draft PR to run the e2e tests pointing at a `aerie-postgres` image running 14.8. [Looks like everything is passing](https://github.com/NASA-AMMOS/aerie-ui/actions/runs/5148641074/jobs/9270609966) 👍

Twisol:
> On a somewhat related note, we could also look into upgrading to JDK 20 in a separate ticket, since JDK 19's official support ended a few months ago (each non-LTS release is only supported for 6 months)

I was looking into this near the start of the year, when the academic quarter wasn't quite fully underway yet, and never managed to come back around to it. We do want to keep up with the latest releases, not because of any official support from Oracle (which, as Ron Pressler notes, is morally only meaningful if you have an actual support contract and/or are unable to upgrade normally), but because every version brings new improvements, both to the JVM and to the language.

Java 20 is a fairly low-impact revision -- which means it should be even easier for us to keep up.

PR #966, opened 2023-06-06
Fix dynamics serialization in GraphQLMerlinService

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

@ewferg observed that in simulation results produced by the scheduler, profile segment dynamics are represented as stringified JSON objects (note the quotes):

Screenshot 2023-06-06 at 7 06 12 AM

Running simulation again (without scheduling) produces profiles with the correct dynamics:

Screenshot 2023-06-06 at 7 08 00 AM

The root cause is the `.add("dynamics", stringDynamics)` towards the bottom of the following snippet. All dynamics are being stringified, then represented as a string:

https://github.com/NASA-AMMOS/aerie/blob/692968ec7fb0585da2a9b2c6236b070b686cbdb0/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java#L979-L994

This PR removes the `toString` call from `serializeDynamics`, and inlines that method into the call site. The `dynamicsP.unparse` method produces a `JsonValue`, which can directly be added to the JsonObjectBuilder.

## Verification

I added an assertion to the existing scheduler e2e test. If you check out the first commit of this PR, redeploy Aerie, and run that test, you will see this assertion error:
```
Expected: {"initial": 3, "rate": 0}
Received: "{\"initial\":3.0,\"rate\":0.0}"
```

After the changes in this PR, this test should now pass.

## Documentation

This is a bug fix - no documentation was invalidated.

## Future work

We should talk about whether the UI should auto-switch to the simulation dataset produced by scheduling. Currently it takes some manual steps to pull it up.

### DIFF
--- a/e2e-tests/src/tests/scheduler.test.ts
+++ b/e2e-tests/src/tests/scheduler.test.ts
@@ -236,6 +236,8 @@ test.describe.serial('Scheduling', () => {
         expect(eqSet(startOffsetOfResource, startOffsetsActivityDirectives)).toEqual(true)
       }
     }
+    const fruit = dataset.filter($ => $.name === "/fruit")[0];
+    expect(fruit.profile_segments[0].dynamics).toEqual({"initial": 3, "rate": 0});
     let topics = await req.getTopicsEvents(request, dataset_id)
     let prefixInput = "ActivityType.Input."
     let prefixOutput = "ActivityType.Output."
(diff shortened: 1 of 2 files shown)


skovati (approved):
LGTM! I appreciate the in-lining of the single-use function 👍

PR #970, opened 2023-06-07
Allow filtering forEach ActivityExpressions on activity arguments

### LINKED ISSUE #925: Coexistence `foreach` doesn't completely filter Activity Directives when given an ActivityExpression with parameters
If the `forEach` specifies an ActivityType with a certain subset of parameters, we disregard the parameters part of the filter and only filter on ActivityType

Example Goal:
```typescript
export default () =>
  Goal.CoexistenceGoal({
    forEach: ActivityExpression.build(ActivityTypes.PeelBanana, { peelDirection: 'fromStem' }),
    activityTemplate: ActivityTemplates.GrowBanana({
      quantity: 1,
      growingDuration: Temporal.Duration.from({ hours: 1 }),
    }),
    startsWithin: TimingConstraint.range(WindowProperty.END, Operator.PLUS, Temporal.Duration.from({ minutes: 5 }))
  });
```

Example Plan:
Bananation Plan with 2 `PeelBanana` Activities with different start times. One has `peelDirection: fromStem`, the other had `peelDirection: fromTip`

Expected Output:
One `GrowBanana` is placed 5 minutes after the `PeelBanana` with `peelDirection: fromStem`

Actual Output:
Two `GrowBanana`s are placed, one after each `PeelBanana`.

JoelCourtney:
* **Tickets addressed:** #925 
* **Review:** By commit  

## Description

This does two main things:
- modify `ForEachActivitySpans` to use an arbitrary activity predicate for its filtering.
- Use this and the argument data provided to `forEach` activity expressions to allow CoexistenceGoal's to filter activities in the `forEach` field based on their arguments.

## Verification

Added an integration test.

Also, the transition to predicates in `ForEachActivitySpans` caused a bunch of tests that relied on equality checks between `ForEachActivitySpans` instances to fail. So I added a little helper record `ForEachActivitySpans.MatchType` which implements `TriFunction` and more importantly provides an equality check.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -462,7 +462,7 @@ public final class ConstraintParsers {
         .field("expression", spansExpressionP)
         .map(
             untuple((kind, actType, alias, expression) -> new ForEachActivitySpans(actType, alias, expression)),
-            $ -> tuple(Unit.UNIT, $.activityType, $.alias, $.expression));
+            $ -> tuple(Unit.UNIT, ((ForEachActivitySpans.MatchType) $.activityPredicate()).type(), $.alias(), $.expression()));
   }
 
   static JsonParser<ForEachActivityViolations> forEachActivityViolationsF(final JsonParser<Expression<List<Violation>>> violationListExpressionP) {
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
@@ -1,34 +1,35 @@
 package gov.nasa.jpl.aerie.constraints.tree;
 
+import gov.nasa.jpl.aerie.constraints.model.ActivityInstance;
 import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
 import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
 import gov.nasa.jpl.aerie.constraints.time.Interval;
 import gov.nasa.jpl.aerie.constraints.time.Spans;
+import org.apache.commons.lang3.function.TriFunction;
 
 import java.util.HashMap;
 import java.util.Objects;
 import java.util.Set;
 
-public final class ForEachActivitySpans implements Expression<Spans> {
-  public final String activityType;
-  public final String alias;
-  public final Expression<Spans> expression;
+public record ForEachActivitySpans(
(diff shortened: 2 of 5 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
```diff
@@ -1,34 +1,35 @@
...
 
 import java.util.HashMap;
 import java.util.Objects;
 import java.util.Set;
 
-public final class ForEachActivitySpans implements Expression<Spans> {
-  public final String activityType;
-  public final String alias;
-  public final Expression<Spans> expression;
+public record ForEachActivitySpans(
```
mattdailis, line 14: Big fan of using records for constraint AST nodes 🚀

As I write this, I do wonder what the implications are for equality of these `TriFunctions` 🤔

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -462,7 +462,7 @@ static JsonParser<ForEachActivitySpans> forEachActivitySpansF(final JsonParser<E
         .field("expression", spansExpressionP)
         .map(
             untuple((kind, actType, alias, expression) -> new ForEachActivitySpans(actType, alias, expression)),
```
mattdailis, line 464: Musing - would it make any sense to expand the constraints' `ForEachActivity` nodes to filter by parameters as well?

I don't suppose it adds any expressive power - since you can already refer to parameters in the body of a constraint.

mattdailis (approved):
Looks good to me - just a couple musings that shouldn't block merge

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/ForEachActivitySpans.java
JoelCourtney: This is the doc comment on `ForEachActivitySpans.MatchType`:

```
   * A helper class for activity predicates that only match on type.
   *
   * This exists to provide an equality check between predicates. You could make an anonymous function that implements
   * {@link TriFunction}, but tests will fail because all such objects are considered non-equal, even when in reality they
   * have identical behavior.
```

I can make a similar record for matching on type and arguments, but currently we don't do any equality check on those types of nodes, so it isn't necessary.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
JoelCourtney: Currently this filtering option isn't exposed to the constraints eDSL for either `Spans` or `Constraint`. It could be added and would dramatically simplify this pattern, but I'd say that's beyond the scope of this ticket. At a minimum we'd probably have to move `ActivityExpression` to the constraints package, and I would put money on that being harder than it sounds.

PR #976, opened 2023-06-12
Optionally associate external datasets with simulation datasets

### LINKED ISSUE #594: Associate external datasets with a simulation dataset instead of a plan
Clipper is using external datasets with simulation datasets instead of plans. We should update our data model to reflect this association.

skovati:
* **Tickets addressed:** Closes #594 
* **Review:** By commit  

## Description

A new nullable column, `associated_simulation_dataset_id`, was added to the `plan_dataset` table. This allows for `plan_datasets`, which are commonly used to store external datasets, to be associated to a single simulation dataset, instead of being implicitly associated with all simulation dataset runs for a given plan.

Simulation associated plan datasets can now be selectively queried using the new eloquently named `GetSimAssociatedPlanDatasetsAction` database action, which is called by `ProfileRepository.getAllPlanDatasetsForPlan` if it is passed a non-empty optional `SimulationDatasetId`.

This allows for existing modules to grab all external datasets pertaining to a plan (such as `TypescriptCodeGenerationService`) while `GetSimulationResultsAction.getViolations` can filter external datasets based on a specific simulation run.

The `addExternalDataset` hasura action has also updated to take in an optional `associatedSimulationDatasetId`. 

Additionally, the `constraintViolations` hasura action has also been updated to take this same (optional) `associatedSimulationDatasetId`.

Since the UI can submit a simulation dataset ID for either `constraintViolations` or in its `GET_PROFILES_EXTERNAL` GQL query, external datasets for a particular simulation can now be selectively queried, and shown as out of date if a new simulation has occurred.

## Verification

Added rudimentary integration tests to check `constraintViolations` Hasura action can optionally take a new `simulation_dataset_id` and only returned violations for the respective simulation dataset.

Additional tests (WIP) will validate `GET_PROFILES_EXTERNAL` being able to query external profiles associated with a single simulation.

## Documentation

Add optional association to external dataset documentation, as well as the general hasura action docs.

## Future work

Frontend will need to be updated to call `/constraintViolations` with a simulation dataset id when this filtering is desired. Additionally, the [GQL query for getting external profiles](https://github.com/NASA-AMMOS/aerie-ui/blob/develop/src/utilities/effects.ts#L1759) in the UI can be updated to include the latest fresh simulation dataset id.

### DIFF
--- a/e2e-tests/src/tests/external-datasets.test.ts
+++ b/e2e-tests/src/tests/external-datasets.test.ts
@@ -1,5 +1,5 @@
 import { expect, test } from '@playwright/test';
-import req from '../utilities/requests.js';
+import req, { awaitSimulation } from '../utilities/requests.js';
 import time from '../utilities/time.js';
 
 test.describe.serial('External Datasets', () => {
@@ -65,6 +65,7 @@ test.describe.serial('External Datasets', () => {
 
   const profile_set_result_1 = {
     plan_dataset_by_pk: {
+      simulation_dataset_id: null,
       offset_from_plan_start: "06:00:00",
       dataset: {
         profiles: [
@@ -99,6 +100,7 @@ test.describe.serial('External Datasets', () => {
 
   const profile_set_result_2 = {
     plan_dataset_by_pk: {
+      simulation_dataset_id: null,
       offset_from_plan_start: "06:00:00",
       dataset: {
         profiles: [
@@ -161,6 +163,7 @@ test.describe.serial('External Datasets', () => {
 
   const profile_set_result_3 = {
     plan_dataset_by_pk: {
+      simulation_dataset_id: null,
       offset_from_plan_start: "06:00:00",
       dataset: {
         profiles: [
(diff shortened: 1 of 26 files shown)


deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
```diff
@@ -0,0 +1,16 @@
+alter table plan_dataset
+  add column associated_simulation_dataset_id integer default null,
+  add constraint simulation_dataset_references_simulation_dataset
```
Mythicaeda: Proposing this name change as it's both clearer and shorter
```suggestion
  add constraint associated_sim_dataset_exists
```

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
```diff
@@ -0,0 +1,16 @@
+alter table plan_dataset
+  add column associated_simulation_dataset_id integer default null,
```
Mythicaeda: Minor, but you don't need to specify `default null` 🙂 
```suggestion
  add column associated_simulation_dataset_id integer,
```

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
```diff
@@ -0,0 +1,16 @@
...
+    references simulation_dataset
+    on update cascade
+    on delete cascade;
+
+comment on column plan_dataset.associated_simulation_dataset_id is e''
+  'The ID of the simulation dataset optionally associated with the plan dataset.'
+  '\n'
+  'This is so that external datasets can be associated with a single simulation run, '
+  'i.e. one simulation_dataset row. If null, this plan dataset is associated with all '
+  'simulation runs for the plan_id plan.';
```
Mythicaeda:
```suggestion
comment on column plan_dataset.associated_simulation_dataset_id is e''
  'The ID of the simulation dataset optionally associated with the dataset.'
  'If null, the dataset is associated with all simulation runs for the plan.';
```

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
```diff
@@ -0,0 +1,16 @@
+alter table plan_dataset
+  add column associated_simulation_dataset_id integer default null,
+  add constraint simulation_dataset_references_simulation_dataset
+    foreign key (associated_simulation_dataset_id)
+    references simulation_dataset
+    on update cascade
+    on delete cascade;
```
Mythicaeda: Do we expect external datasets to change which simulation they're associated with? If so, then `on delete set null` may make more sense here.

merlin-server/sql/merlin/tables/plan_dataset.sql
```diff
@@ -1,6 +1,7 @@
 create table plan_dataset (
   plan_id integer not null,
   dataset_id integer not null,
+  associated_simulation_dataset_id integer default null,
```
Mythicaeda: Minor, may be ignored: 
```suggestion
  simulation_dataset_id integer default null,
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/SimulationDatasetId.java
```diff
@@ -0,0 +1,3 @@
+package gov.nasa.jpl.aerie.merlin.server.models;
+
+public record SimulationDatasetId(long id) {}
```
Mythicaeda, line 3: Proposing as, while I know our existing `Id`s use `long`, the valid range of this value (and our other `Id`s) is actually the `int` range.

```suggestion
public record SimulationDatasetId(int id) {}
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreatePlanDatasetAction.java
```diff
@@ -33,12 +37,17 @@ public PlanDatasetRecord apply(
     this.statement.setLong(1, planId);
     PreparedStatements.setTimestamp(this.statement, 2, datasetStart);
     PreparedStatements.setTimestamp(this.statement, 3, planStart);
+    this.statement.setObject(
+        4,
+        associatedSimulationDatasetId.map(SimulationDatasetId::id).orElse(null),
+        Types.BIGINT
```
Mythicaeda:
```suggestion
        Types.INTEGER
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/CreatePlanDatasetAction.java
```diff
@@ -25,6 +28,7 @@ public CreatePlanDatasetAction(final Connection connection) throws SQLException
 
   public PlanDatasetRecord apply(
       final long planId,
+      final Optional<SimulationDatasetId> associatedSimulationDatasetId,
```
Mythicaeda: This name change proposal is associated with the one I made on the db column
```suggestion
      final Optional<SimulationDatasetId> simulationDatasetId,
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanDatasetsAction.java
```diff
@@ -35,7 +37,11 @@ public List<PlanDatasetRecord> get(final PlanId planId) throws SQLException {
     while (resultSet.next()) {
       final var datasetId = resultSet.getLong(1);
       final var offsetFromPlanStart = parseOffset(resultSet, 2);
-      records.add(new PlanDatasetRecord(planId.id(), datasetId, offsetFromPlanStart));
+      final var associatedSimulationDatasetId = Optional
+          .ofNullable(resultSet.getString(3))
+          .map(Long::valueOf)
+          .map(SimulationDatasetId::new);
```
Mythicaeda: Minor, but a suggestion based on how we get an `Optional<Long>` in [`GetSpanRecords`](https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSpanRecords.java#L49):
```suggestion
      final Optional<SimulationDatasetId> associatedSimulationDatasetId = 
          resultSet.getObject(3) == null ? 
              Optional.empty() : 
              Optional.of(new SimulationDatasetId(resultSet.getLong(3)));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java
```diff
@@ -13,12 +13,14 @@ public sealed interface Input { }
...
   public record MissionModelInput(String missionModelId) implements Input { }
   public record PlanInput(PlanId planId) implements Input { }
+  public record ConstraintViolationsInput(PlanId planId, Optional<SimulationDatasetId> simulationDatasetId) implements Input { }
   public record ActivityInput(String missionModelId,
                               String activityTypeName,
                               Map<String, SerializedValue> arguments) implements Input {}
   public record MissionModelArgumentsInput(String missionModelId,
                               Map<String, SerializedValue> arguments) implements Input {}
   public record UploadExternalDatasetInput(PlanId planId,
+                                           Optional<SimulationDatasetId> associatedSimulationDatasetId,
```
Mythicaeda: Based on the fact that if `associatedSimulationDatasetId` is defined, a disjoint set of logic appears to be run, I wonder if it would make more sense to create a new action with new functions (ie `getAllPlanDatasetsForSimulation`) rather than add an Optional parameter to our existing actions.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimAssociatedPlanDatasetsAction.java
```diff
@@ -0,0 +1,50 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.parseOffset;
+
+/*package-local*/ final class GetSimAssociatedPlanDatasetsAction implements AutoCloseable {
```
Mythicaeda, line 16: Minor and nonblocking, but I found this class name difficult to read, although I'm not sure how to improve it without making it even longer, ie `GetPlanDatasetsFromSimDatasetAction`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java
Mythicaeda: Related to this, in the Java methods, if the Optional argument is defined, then the method `getAllPlanDatasetsForPlan` does not do what its name implies it does.

e2e-tests/src/tests/external-datasets.test.ts
```diff
@@ -341,3 +344,231 @@ test.describe.serial('External Datasets', () => {
...
+
+  test("Check there is one violation when simulationDatasetId is provided", async ({ request }) => {
+    const violations: ConstraintViolation[] = await req.checkConstraints(request, plan_id, first_simulation_dataset_id);
+    violation = violations[0];
+
+    expect(violations).not.toBeNull();
+    expect(violations).toBeDefined();
+    expect(violations).toHaveLength(1);
+  });
+
```
Mythicaeda: It looks like the violation gotten in both of these tests is expected to be the same one. Can that assertion please be added?

e2e-tests/src/tests/external-datasets.test.ts
```diff
@@ -341,3 +344,231 @@ test.describe.serial('External Datasets', () => {
...
+  test("Check there are no violations when simulationDatasetId is provided", async ({ request }) => {
+    const violations: ConstraintViolation[] = await req.checkConstraints(request, plan_id, second_simulation_dataset_id);
+
+    expect(violations).not.toBeNull();
+    expect(violations).toBeDefined();
+    expect(violations).toHaveLength(0);
+  });
```
Mythicaeda: Minor, but what if the first simId is provided after simulating a second time? I expect it will give the same results as `Check there is one violation when simulationDatasetId is provided`

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
skovati: change has been made

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
skovati: good call 👍

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
skovati: much more concise, thanks!

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
skovati: I don't think the original use-case intended to change association, but I think supporting it is a good idea. However, wouldn't using `on delete set null` inadvertently associate any related external datasets with *all* simulation datasets whenever a sim dataset is deleted?

merlin-server/sql/merlin/tables/plan_dataset.sql
skovati: While wordy, I think the original naming scheme emphasizes the optional association aspect of this new column, vs it being required.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java
skovati: I agree and this was really my main concern with proposed implementation. My reasoning was that a relatively niche feature would be better suited as an optional parameter for an existing API endpoint vs defining an entirely new endpoint for both `uploadExternalDataset` and `constraintViolations`

skovati: Additionally, I modeled this after `ConstraintsInput`:

https://github.com/NASA-AMMOS/aerie/blob/8cfdc83f0dbc0d61621d489f068e529dc85fba50/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java#L27

where an optional is used to augment an API endpoint instead of creating two endpoints.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/SimulationDatasetId.java
skovati: Happy to make this change, but should we break this precedent without changing all the other `Id` records?

merlin-server/sql/merlin/tables/plan_dataset.sql
Mythicaeda: I'd argue that the nullability of the column is sufficient to indicate that it is optional. Additionally, the proposed name matches the precedent for other optional "associated" columns (`activity_directive.anchor_id`/`constraints.model_id`/`constraint.plan_id`).

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java
Mythicaeda: What about just splitting the `getAllPlanDatasetsForSimulation` part of `getAllPlanDatasetsForPlan` into its method then, as it would resolve the main issue of `getAllPlanDatasetsForPlan` being an inaccurate method name.

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
Mythicaeda: It would. The question I should've asked is: "Do we always want to delete a plan dataset when its currently associated sim dataset is deleted?"

skovati: Maybe we can just remove the `on delete` clause then. Since the end-user is responsible for creating, uploading, and associating these datasets, maybe they can also be responsible for updating the `simulation_dataset_id` column with a simple mutation instead of that behavior being hardcoded within the merlin db?

Mythicaeda: Excluding the `on delete` clause will cause exceptions due to constraint violations. Our options are `RESTRICT`, `CASCADE`, and `SET NULL`, and I'm fine if we want to say that an external dataset associated with a simulation dataset is part of that sim dataset's data and needs to be removed when that sim dataset is (which, rereading [the discussion on the ticket](https://github.com/NASA-AMMOS/aerie/issues/594), appears the case).

From the [Postgres Docs](https://www.postgresql.org/docs/14/ddl-constraints.html):
> Restricting and cascading deletes are the two most common options. RESTRICT prevents deletion of a referenced row. NO ACTION means that if any referencing rows still exist when the constraint is checked, an error is raised; this is the default behavior if you do not specify anything. (The essential difference between these two choices is that NO ACTION allows the check to be deferred until later in the transaction, whereas RESTRICT does not.) CASCADE specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well. There are two other options: SET NULL and SET DEFAULT. These cause the referencing column(s) in the referencing row(s) to be set to nulls or their default values, respectively, when the referenced row is deleted. Note that these do not excuse you from observing any constraints. For example, if an action specifies SET DEFAULT but the default value would not satisfy the foreign key constraint, the operation will fail.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/SimulationDatasetId.java
Mythicaeda: Yeah, this is a weird case. Given that, [I've created a ticket for the refactor](https://github.com/NASA-AMMOS/aerie/issues/999) so that the change can be done all at once if we decide to do it.

deployment/hasura/migrations/AerieMerlin/16_associate_external_datasets_with_simulation/up.sql
skovati: I think `on delete cascade` makes the most sense then. `set null` would result in the external dataset being wrongly associated with other sim datasets unless explicitly updated, which I think is less intuitive than deletion.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/HasuraAction.java
skovati: change has been made!

e2e-tests/src/tests/external-datasets.test.ts
skovati: Good call, an additional test has been added for this case 👍

Mythicaeda (approved):
LGTM! Thanks for the great work!

PR #980, opened 2023-06-14
Activate and load: remove addition properties

goetzrrGit:
## Description
A hotfix to remove these extra properties added to the SeqJSON. This isn't allowed in the SPEC

## Verification
Ran e2e test without any issues.


camargo (approved):
LGTM!

PR #984, opened 2023-06-20
Add `getEffectiveArgumentsBulk` Hasura Action

### LINKED ISSUE #666: Provide single GraphQL endpoint to get activity effective arguments
Source: https://jira.jpl.nasa.gov/browse/AERIEQS-399

If a downstream tool would like to query for the effective arguments of all N activity directives in a plan, it needs to make 1 + N queries: 1 query to get all the directives, and N queries to get the effective arguments of each one. This loads the mission model jar N times - although after #312 this is a cheaper operation, since we no longer instantiate the mission model in order to get effective arguments.

This ticket would be satisfied by adding a `getEffectiveArgumentsBulk` action that takes a list of activity directives. This would reduce the number of queries to 2.

We could additionally consider providing a single action that encompasses both queries - name TBD.

skovati:
* **Tickets addressed:** closes #666 
* **Review:** By commit  

## Description

Adds a new `getEffectiveArgumentsBulk` Hasura action that takes a list of `EffectiveArgumentsInput` (i.e. a list of activity types and arguments), loads the mission model only once, and retrieves the effective arguments for every element in the input list.

Additionally, this PR deprecates the existing `getActivityEffectiveArguments` action in favor of the new bulk query action.

## Verification

e2e tests were added for checking effective arguments in several different cases (single, bulk, malformed, etc) 

## Documentation

Existing hasura action docs need to be updated to reference this new API endpoint.

## Future work

N/A

### DIFF
--- a/deployment/hasura/metadata/actions.yaml
+++ b/deployment/hasura/metadata/actions.yaml
@@ -79,6 +79,14 @@ actions:
     permissions:
       - role: aerie_admin
       - role: user
+  - name: getActivityEffectiveArgumentsBulk
+    definition:
+      kind: ""
+      handler: "{{AERIE_MERLIN_URL}}/getActivityEffectiveArgumentsBulk"
+      timeout: 300
+    permissions:
+      - role: aerie_admin
+      - role: user
   - name: getActivityTypeScript
     definition:
       kind: ""
--- a/e2e-tests/src/types/parameter.d.ts
+++ b/e2e-tests/src/types/parameter.d.ts
@@ -1,9 +1,14 @@
 type EffectiveArguments = {
-  arguments: ArgumentsMap;
-  errors: ParametersErrorMap;
+  arguments?: ArgumentsMap;
+  errors?: ParametersErrorMap;
   success: boolean;
+  typeName?: string;
 };
 
+type EffectiveArgumentItem = {
+  activityTypeName: string;
+  activityArguments: ArgumentsMap;
+};
 
 type Argument = any;
 
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import static gov.nasa.jpl.aerie.merlin.server.http.HasuraParsers.hasuraActivityActionP;
+import static gov.nasa.jpl.aerie.merlin.server.http.HasuraParsers.hasuraActivityBulkActionP;
 import static gov.nasa.jpl.aerie.merlin.server.http.HasuraParsers.hasuraActivityDirectiveEventTriggerP;
 import static gov.nasa.jpl.aerie.merlin.server.http.HasuraParsers.hasuraConstraintsCodeAction;
 import static gov.nasa.jpl.aerie.merlin.server.http.HasuraParsers.hasuraConstraintsViolationsActionP;
@@ -89,6 +90,7 @@ public final class MerlinBindings implements Plugin {
       path("validatePlan", () -> post(this::validatePlan));
       path("getModelEffectiveArguments", () -> post(this::getModelEffectiveArguments));
       path("getActivityEffectiveArguments", () -> post(this::getActivityEffectiveArguments));
+      path("getActivityEffectiveArgumentsBulk", () -> post(this::getActivityEffectiveArgumentsBulk));
       path("addExternalDataset", () -> post(this::addExternalDataset));
       path("extendExternalDataset", () -> post(this::extendExternalDataset));
       path("constraintsDslTypescript", () -> post(this::getConstraintsDslTypescript));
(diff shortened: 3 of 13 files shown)


Twisol:
FWIW, if a query accepts a list `[T]`, GraphQL will allow a single `T` to be passed -- it automatically promotes `T` to `[T]` on input.

This could be useful for other/future queries, since we basically get both a singular and a plural query in one shot.

Twisol:
Sorry, I just spotted that this is a Draft -- I don't think GitHub is so kind as to mention that fact in its email notifications. I'll hold my feedback 👀

skovati:
@Twisol Ooh good point; I was thinking their had to be a more elegant way to do this. We'd likely have to make this a breaking change if we implemented it that way, since AFAIK you can't have something like
```graphql
type Query {
  getActivityEffectiveArguments(
    missionModelId: ID!
    [
      activityTypeName: String!
      activityArguments: ActivityArguments!
    ]!
  ): EffectiveArgumentsResponse
}
```
i.e. an inline object definition within a list.

skovati:
> I'll hold my feedback

Please don't hold back! This is only a draft since I'm working on e2e tests still.

Twisol:
> We'd likely have to make this a breaking change if we implemented it that way

Yes -- I wouldn't advocate for breaking compatibility here, necessarily. I just wanted to float that tidbit of GraphQL API design -- it comes in real handy.

Twisol:
> > We'd likely have to make this a breaking change if we implemented it that way
> 
> Yes -- I wouldn't advocate for breaking compatibility here, necessarily. I just wanted to float that tidbit of GraphQL API design -- it comes in real handy.

Point being that it leads you to think earlier about what parts of your query inputs are pluralizable -- which might have led, at least, to the original singular query grouping its two arguments in a separate input type.

(The output would also need to be pluralized, though, and I think that would still be a breaking change. *C'est la vie*.)

e2e-tests/src/types/parameter.d.ts
```diff
@@ -1,9 +1,17 @@
 type EffectiveArguments = {
-  arguments: ArgumentsMap;
-  errors: ParametersErrorMap;
+  arguments?: ArgumentsMap;
```
mattdailis, line 2: Could you help me understand why these properties are now optional? My brain function is too low right now to try to figure it out myself 🥴

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
```diff
@@ -50,6 +50,11 @@ Map<String, SerializedValue> getActivityEffectiveArguments(String missionModelId
          NoSuchActivityTypeException,
          InstantiationException;
 
+  List<BulkEffectiveArgumentResponse> getActivityEffectiveArgumentsBulk(
```
mattdailis, line 48: Have you considered replacing `getActivityEffectiveArguments` with `getActivityEffectiveArgumentsBulk` at the `MissionModelService` layer?

e2e-tests/src/types/parameter.d.ts
skovati: 😆 This particular change just updates the existing type within our e2e tests to match what `getActivityEffectiveArguments` already returns: optional `errors` and `arguments` (i.e. one or the other).

I then matched this same behavior for `getActivityEffectiveArgumentsBulk` to return a list of these same objects with optional fields.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
skovati: Oh I hadn't! I'll work on that change, because I discovered a small bug with the existing implementation of `getActivityEffectiveArguments` that I can fix.

Currently, on `develop` the following GQL query will result in a server error (500) since we aren't correctly handling the exception where the activity doesn't exist.

```GQL
query ThrowsServerError {
  getActivityEffectiveArguments(
    activityArguments: {}, 
    activityTypeName: "BiteBananaDOESNOTEXIST",
    missionModelId: 1
   ) {
    arguments
    errors
    success
  }
}
```

The new `getActivityEffectiveArgumentsBulk` action correctly serializes this response, so I'll update this PR to make use of that.

skovati: Additionally, maybe the existing `getActivityEffectiveArguments` endpoint could be marked as
```gql
@deprecated(reason: "Use bulk query getActivityEffectiveArgumentsBulk instead")
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -332,14 +335,29 @@ private void getActivityEffectiveArguments(final Context ctx) {
...
+          missionModelId,
+          List.of(serializedActivity));
 
-      ctx.result(ResponseSerializers.serializeEffectiveArgumentMap(arguments).toString());
-    } catch (final InstantiationException ex) {
-      ctx.status(200).result(ResponseSerializers.serializeInstantiationException(ex).toString());
-    } catch (final MissionModelService.NoSuchActivityTypeException ex) {
-      ctx.status(400)
-         .result(ResponseSerializers.serializeFailures(List.of(ex.getMessage())).toString());
+      ctx.result(ResponseSerializers.serializeBulkEffectiveArgumentResponseList(arguments).toString());
```
mattdailis: For backwards compatibility, I suspect we'll need to pull out the only element in this list, and serialize it alone. Does that track?

EDIT: Whoops, I thought I had submitted this comment. Submitting now, belatedly

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
skovati: The latest commit I've added deprecates the existing `getActivityEffectiveArguments` Hasura action in favor of the new bulk query action. The existing endpoint still exists, but will now return a list of `EffectiveArgumentResponse`, which is a small breaking change.

Additionally, this fixes the above bug where invalid activity type names were throwing an unserialized server error, as `getActivityEffectiveArguments` now uses the new serialization introduced with the new `getActivityEffectiveArgumentsBulk` action. I could make this a non-breaking change by adding some extra glue code, if that is desired.

mattdailis: Hmm, I was imagining we could get away without a breaking change 🤔 In theory there can be two separate endpoints in `MerlinBindings` that call the same method in `MerlinService`. Ah, I guess that's what you mean by "adding some extra glue code".

If possible, I'd like to make this a smooth deprecation - give users a release or two to use the new endpoint in their scripts, and then pull the rug

skovati: Good call. Turns out the only "glue code" I needed was calling a function I already wrote that serializes a single response 🧠. Change is no longer breaking, and we can remove the deprecated endpoint (possibly along with `/getSequenceSeqJson` and `/getEdslForSeqJson`) in a later release.

deployment/hasura/metadata/actions.yaml
```diff
@@ -79,6 +79,13 @@ actions:
     permissions:
       - role: aerie_admin
       - role: user
+  - name: getActivityEffectiveArgumentsBulk
+    definition:
+      kind: ""
+      handler: "{{AERIE_MERLIN_URL}}/getActivityEffectiveArgumentsBulk"
+      timeout: 300
+    permissions:
+      - role: user
```
camargo, line 89: Changing out from under you: https://github.com/NASA-AMMOS/aerie/pull/1028
Can we add `- role: aerie_admin` here as well?

camargo (approved):
LGTM! Thanks for the great e2e tests as well.

PR #985, opened 2023-06-20
Make Migration Script Return Non-Zero Exit Code 

### LINKED ISSUE #917: Make DB Migration Script return non-zero exit code if one step of Bulk Migration fails
During multidb bulk migrations, the script's exit code is not affected by whether the a step was able to run. This means that in our pgcmp workflow, migration failures aren't caught in the job running the migration, making it harder to debug migration errors.

Completion of this ticket will make the migration script return a non-zero exit value if any database cannot be bulk-migrated, and potentially update the `pgcmp` workflow's behavior based on this change.

Mythicaeda:
* **Tickets addressed:** Closes #917
* **Review:** By commit  

## Description

Rather than use the exit code of the dryrun, which always returns 0, the script now uses the exit code of running the migration to determine if migrating failed, and if so, to update the exit code of the program. The exit code now communicates how many and which DBs failed to migrate.

The dry-run command has been kept, as it prints which migrations will be run, which is not printed when the migration is actually run.

Additionally a bug that would allow failed migrations to be "skipped" in step-by-step mode has been patched.

## Verification

Locally tested by running modified migrations that failed and checking the exit codes.

## Documentation

No documentation updates are needed.

### DIFF
--- a/deployment/aerie_db_migration.py
+++ b/deployment/aerie_db_migration.py
@@ -65,52 +65,55 @@ def step_by_step_migration(database, apply):
     timestamp = step.split("_")[0]
 
     if apply:
-      exit_code = os.system(f'hasura migrate apply --version {timestamp} --database-name {database.db_name} --dry-run --log-level WARN')
+      os.system(f'hasura migrate apply --version {timestamp} --database-name {database.db_name} --dry-run --log-level WARN')
     else:
-      exit_code = os.system(f'hasura migrate apply --version {timestamp} --type down --database-name {database.db_name} --dry-run --log-level WARN')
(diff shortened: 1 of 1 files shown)


skovati (approved):
LGTM!

Does the hasura cli return nonbinary exit codes that we might want to propagate instead of just incrementing `exit_with`? (e.g. exit code of 2 means failed DB connection vs failed migration application or something).

Food for thought, but this change looks good as is!

camargo (approved):

Mythicaeda:
> Does the hasura cli return nonbinary exit codes that we might want to propagate instead of just incrementing exit_with? (e.g. exit code of 2 means failed DB connection vs failed migration application or something).

As far as I can tell, no. It gives an exit code of 256 for both failing to apply a migration and failing to connect the Hasura instance.

PR #996, opened 2023-06-23
Users table

### LINKED ISSUE #938: Add users table to Merlin database in a schema
Prereq of https://github.com/NASA-AMMOS/aerie/issues/862

This ticket is to add `users` table to the Merlin database that will store users and their roles in Aerie.

Mythicaeda:
* **Tickets addressed:** Closes #938 
* **Review:** By commit  

## Description

Three tables and a view have been added to Merlin: 
- `user_roles`, which contains all the valid roles and optionally a description. By default, it contains `user`, `admin`, and `viewer`
- `users`, which by default has the "users" `Mission Model` and `Aerie Legacy`. This table also specifies the default role for each user, however no one with the `user` role may select this column, as the current user may only see _their_ default_role, [but the `user` role more pressingly needed to be able to select the `id` and `username` of all users](https://github.com/hasura/graphql-engine/issues/3442).
- `user_allowed_roles`, which contains all the roles a user may use for querying Hasura
- `user_and_roles`, which is a view that collects the role information for each user. This is what I recommend get queried for getting Hasura role information

#### Breaking Change:

The type of all columns that stored user information has changed from text to integer.

## Verification

Tests have been updated to reflect the breaking change


merlin-server/sql/merlin/views/users_and_roles.sql
```diff
@@ -0,0 +1,28 @@
...
+$$;
+
+create view metadata.users_and_roles as
+(
+  select
+    u.id as id,
+    u.username as username,
+    -- Roles
+    u.default_role as hasura_default_role,
+    metadata.get_user_allowed_roles(u.id) as hasura_allowed_roles
```
mattdailis: Have you considered using a lateral join here? I think that would make this query easier for postgres to optimize

deployment/hasura/migrations/AerieMerlin/17_users/up.sql
```diff
@@ -0,0 +1,873 @@
...
+update mission_model
+  set owner_id = u.id
+  from metadata.users u
+  where owner = u.username;
+
+alter table public.mission_model drop column owner;
+alter table public.mission_model rename column owner_id to owner;
+
+comment on column mission_model.owner is e''
+  'A human-meaningful identifier for the user responsible for this model.';
```
mattdailis: Is the human-meaningful part still accurate?

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/users_allowed_roles.yaml
```diff
@@ -0,0 +1,16 @@
+table:
+  name: users_allowed_roles
+  schema: metadata
+configuration:
+  custom_name: "users_allowed_roles"
+select_permissions:
+  - role: user
+    permission:
+      columns: [user_id, allowed_role]
+      filter: {"user_id":{"_eq":"X-Hasura-User-Id"}}
```
mattdailis, line 10: I'm noticing that the yaml files capitalize session variable names in Hyphenated-Title-Case while the Gateway and hasura functions use kebab-case. Not a concern, just an observation - my guess is that the yaml version is case insensitive, while the SQL version is case sensitive

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/users_and_roles_view.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: users_and_roles
+  schema: metadata
+configuration:
+  custom_name: "users_and_roles"
+select_permissions:
+  - role: user
+    permission:
+      columns: [id, username, hasura_default_role, hasura_allowed_roles]
+      filter: {"id":{"_eq":"X-Hasura-User-Id"}}
+      allow_aggregations: false
```
mattdailis, line 11: I'm observing that we sometimes set `allow_aggregations` to true and sometimes to false. Do we have a general feel for when it should be one or the other?

mattdailis (approved):
Very nicely done - the change from string to integer user ids was a pervasive one, and you've handled it thoroughly and cleanly. I've left some questions and observations, but none that are blockers.

merlin-server/sql/merlin/views/users_and_roles.sql
Mythicaeda: I did, however when that made the `hasura_allowed_roles` column have a value of  `{NULL}` instead of `NULL` when a user had no allowed roles.

The SQL code I attempted was this:

```sql
create view metadata.users_and_roles as
(
  select
    u.id as id,
    u.username as username,
    -- Roles
    u.default_role as hasura_default_role,
    array_agg(r.allowed_role) as hasura_allowed_roles
  from metadata.users u
  left join metadata.users_allowed_roles r
  on r.user_id = u.id
  group by u.id
);
```

Were you thinking of a different statement?

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/users_and_roles_view.yaml
Mythicaeda: In general, I allow aggregations to be `true`. I've only set it to false where a role has a filter on their `SELECT` privileges (or in the case of `users_and_roles_view` where they are only allowed to SELECT one row). Since aggregates only run on the portion of the table they're allowed to select, I can set the `users_allowed_roles` to allowing aggregates, although I stand by them remaining disable for `users_and_roles_view`.

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/users_allowed_roles.yaml
Mythicaeda: Actually, it might be Hasura that's case-insensitive, as capitalization for the session variables in the yaml came straight out of the Hasura Permissions Builder UI

merlin-server/sql/merlin/views/users_and_roles.sql
mattdailis: I see - that makes sense. `select array_agg((null::text));` returns `{NULL}`. Just playing around, I found that adding a filter seemed to produce `NULL` (inspired by: https://stackoverflow.com/a/23050948)

```sql
create view metadata.users_and_roles as
(
  select
    u.id as id,
    u.username as username,
    -- Roles
    u.default_role as hasura_default_role,
    array_agg(r.allowed_role) filter (where r.allowed_role is not null) as hasura_allowed_roles
  from metadata.users u
  left join metadata.users_allowed_roles r
  on r.user_id = u.id
  group by u.id
);
```

Mythicaeda: Thanks! I've updated the view to use the `filter` expression.

deployment/hasura/migrations/AerieMerlin/17_users/up.sql
Mythicaeda: Unfortunately not. I've updated the comment to reflect that.

Mythicaeda:
Closing as we have decided to stick with the text usernames and require Aerie Deployments to manage cascading user updates/deletes

PR #1000, opened 2023-06-26
Allow user to update model_id, plan_id of Constraints

Mythicaeda:
* **Tickets addressed:**  Hotfix
* **Review:** By commit  

## Description

Allow users to update the `plan_id` and `model_id` of constraints they're allowed to update. A post-update check has been added to enforce that they are not assigning either value to one they are not allowed to (ie, they may only update the plan_id to that of another plan they own/collaborate on)


camargo (approved):
LGTM. Thanks for the quick turnaround on this one.

PR #1001, opened 2023-06-26
Declare default permissions for Hasura actions/functions

Mythicaeda:
* **Review:** By commit  

## Description

[This Slack](https://jpl.slack.com/archives/CJ07FGRPH/p1687814932955879?thread_ts=1687812777.519679&cid=CJ07FGRPH) conversation made me realize that, despite having `SELECT` permissions on the correct tables and appearing to have access in the console, none of our `volatile` functions were being exposed to those with the `user` role. It turns out that for volatile functions, permissions need to be explicitly granted. At the same time, I've also granted the `user` role permission to run all actions. Fine-tuning of these permissions will be implemented in a later PR.

Additionally, I have set `HASURA_GRAPHQL_INFER_FUNCTION_PERMISSIONS` to `false` in our Docker Compose. This allows us to see what the permission _actually_ is for a given function.

## Verification
Attempted to call various actions/functions in the Hasura Metadata

### DIFF
--- a/deployment/hasura/metadata/actions.yaml
+++ b/deployment/hasura/metadata/actions.yaml
@@ -4,131 +4,183 @@ actions:
       kind: synchronous
       handler: "{{AERIE_MERLIN_URL}}/addExternalDataset"
       timeout: 300
+    permissions:
+      - role: user
   - name: extendExternalDataset
     definition:
       kind: synchronous
       handler: "{{AERIE_MERLIN_URL}}/extendExternalDataset"
       timeout: 300
+    permissions:
+      - role: user
   - name: uploadDictionary
     definition:
       kind: synchronous
       handler: "{{AERIE_SEQUENCING_URL}}/put-dictionary"
       timeout: 300
+    permissions:
+      - role: user
   - name: addCommandExpansionTypeScript
     definition:
       kind: synchronous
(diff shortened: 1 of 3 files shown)


camargo (approved):
LGTM

deployment/hasura/metadata/actions.yaml
```diff
@@ -4,131 +4,183 @@ actions:
       kind: synchronous
       handler: "{{AERIE_MERLIN_URL}}/addExternalDataset"
       timeout: 300
+    permissions:
+      - role: user
```
mattdailis, line 8: Just to wrap my head around this - does this mean that `admin` and `viewer` cannot run these actions?

Mythicaeda: `admin` can run everything. This means that `viewer` cannot run these actions.

mattdailis: Thanks, that makes sense - where do we specify that `admin` can run everything?

mattdailis: Ah, it seems that `admin` is a special role https://hasura.io/docs/latest/auth/authentication/admin-secret-access/#admin-role

mattdailis (approved):

Issue #934, opened 2023-05-11
Extract Subsystem tags from Mission Model


Mythicaeda:
Prerequisite Issue: #928

When a mission model is uploaded, we need to extract the subsystems and put them in the `tags` table, as well as associate ActivityTypes with their subsystems.

Subsystems may be declared in one of two places:

1. As an `@subsystem` in the JavaDocs of an Activity (see [Matt's work](https://jpl.slack.com/archives/CJ07FGRPH/p1681314813727559) for extracting this information)
2. Statically, similar to how ActivityTypes are declared (not yet implemented, focus on 1 first)

PR #1003, opened 2023-06-28
Users Table

### LINKED ISSUE #938: Add users table to Merlin database in a schema

Mythicaeda:
* **Tickets addressed:** Closes #938
* **Review:** By commit  

## Description

Three tables and a view have been added to Merlin: 
- `user_roles`, which contains all the valid roles and optionally a description. By default, it contains `user`, `admin`, and `viewer`
- `users`, which by default has the "users" `Mission Model` and `Aerie Legacy`. This table also specifies the default role for each user, however no one with the `user` role may select this column, as the current user may only see _their_ default_role, [but the `user` role more pressingly needed to be able to select the `id` and `username` of all users](https://github.com/hasura/graphql-engine/issues/3442).
- `user_allowed_roles`, which contains all the roles a user may use for querying Hasura
- `user_and_roles`, which is a view that collects the role information for each user. This is what I recommend get queried for getting Hasura role information

Foreign key relationships have been established between Merlin tables and the the `users` table where applicable. Additionally, all instances along the lines of `owner text not null default ''` have been replaced with `owner text` columns, with existing `default` values getting replaced with `null`.

## Verification

The DBTests now insert the users that they will need as part of their `beforeAll` tasks. The users that are used in `MerlinDatabaseTestHelper` methods are now inserted as part of the class's constructor.

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
@@ -80,6 +80,10 @@ public class PlanCollaborationTests {
     helper.startDatabase();
     connection = helper.connection();
     merlinHelper = new MerlinDatabaseTestHelper(connection);
+    merlinHelper.insertUser("PlanCollaborationTests");
+    merlinHelper.insertUser("PlanCollaborationTests Reviewer");
+    merlinHelper.insertUser("PlanCollaborationTests Requester");
+    merlinHelper.insertUser("TagsTest");
   }
 
   @AfterAll
@@ -104,7 +108,7 @@ public class PlanCollaborationTests {
   int duplicatePlan(final int planId, final String newPlanName) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement.executeQuery("""
-        select duplicate_plan(%s, '%s', 'DBTests') as id;
+        select duplicate_plan(%s, '%s', 'PlanCollaborationTests') as id;
       """.formatted(planId, newPlanName));
       res.next();
       return res.getInt("id");
--- a/merlin-server/sql/merlin/tables/metadata/users.sql
+++ b/merlin-server/sql/merlin/tables/metadata/users.sql
@@ -1,0 +1,18 @@
+create table metadata.users(
+  username text not null primary key,
+  default_role text not null references metadata.user_roles
+    on update cascade
+    on delete restrict
+);
+-- Insert the default roles into the table, then change the generated status to "Always"
+-- This can be changed back if we need to add more default users in the future
+insert into metadata.users(username, default_role)
+  values ('Mission Model', 'viewer'),
+         ('Aerie Legacy', 'viewer');
+
+comment on table metadata.users is e''
+'All users recognized by this deployment.';
+comment on column metadata.users.username is e''
+'The user''s username. A unique identifier for this user.';
+comment on column metadata.users.default_role is e''
+'The user''s default role for making Hasura requests.';
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
@@ -7,8 +7,24 @@ import java.util.UUID;
 final class MerlinDatabaseTestHelper {
   private final Connection connection;
 
-  MerlinDatabaseTestHelper(Connection connection) {
+  MerlinDatabaseTestHelper(Connection connection) throws SQLException {
     this.connection = connection;
+    insertUser("Merlin DB Tests");
+  }
+
+  void insertUser(final String username) throws SQLException {
+    insertUser(username, "admin");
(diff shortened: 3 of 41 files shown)


merlin-server/sql/merlin/tables/metadata/users.sql
```diff
@@ -0,0 +1,18 @@
+create table metadata.users(
+  username text not null primary key,
+  default_role text not null references metadata.user_roles
+    on update cascade
+    on delete restrict
+);
+-- Insert the default roles into the table, then change the generated status to "Always"
+-- This can be changed back if we need to add more default users in the future
+insert into metadata.users(username, default_role)
+  values ('Mission Model', 'viewer'),
```
camargo, line 10: What is the meaning of these default usernames?

Mythicaeda: [They're usernames that the Tags table reserved](https://github.com/NASA-AMMOS/aerie/blob/develop/merlin-server/sql/merlin/tables/metadata/tags.sql#L21-L24).

`Mission Model` is used for tags originating from an uploaded mission model (reserved in preparation for #934)
`Aerie Legacy` is used for tags originating from a version of Aerie prior to the tags table's creation.

`Mission Model` was reserved as it will be part of a user-agnostic DB Event. 
Taking another look, `Aerie Legacy` did not _need_ to be reserved, since all it does is make clear what is an old tag and what is a tag without an owner.

camargo (approved):
LGTM! Thanks

merlin-server/sql/merlin/tables/metadata/users.sql
mattdailis: Does the `viewer` part have any significance?

Mythicaeda: Somewhat. `default_role` is a non-nullable column, and `viewer` is the most restrictive role. However, even if a user managed to log in as a "reserved" user, they wouldn't be able to perform any queries/mutations, since they have no `allowed_roles`.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -80,6 +80,10 @@ void beforeAll() throws SQLException, IOException, InterruptedException {
     helper.startDatabase();
     connection = helper.connection();
     merlinHelper = new MerlinDatabaseTestHelper(connection);
+    merlinHelper.insertUser("PlanCollaborationTests");
+    merlinHelper.insertUser("PlanCollaborationTests Reviewer");
```
mattdailis, line 84: I'm curious about the intent behind inserting these users - e.g. "PlanCollaborationTests Reviewer" that seem unused at first glance

mattdailis (approved):
Looks good to me - no blocking concerns

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: All of these users are used during tests in this class:
- [PlanCollaborationTests](https://github.com/NASA-AMMOS/aerie/blob/5a9ffa15010924cad1e62ef206b08dfe3852aac3/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java#L111)
- [PlanCollaborationTests Requester](https://github.com/NASA-AMMOS/aerie/blob/5a9ffa15010924cad1e62ef206b08dfe3852aac3/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java#L190)
- [PlanCollaborationTests Reviewer](https://github.com/NASA-AMMOS/aerie/blob/5a9ffa15010924cad1e62ef206b08dfe3852aac3/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java#L202)
- TagsTest is implicitly used in the calls to [`tagsHelper.insertTag()`](https://github.com/NASA-AMMOS/aerie/blob/5a9ffa15010924cad1e62ef206b08dfe3852aac3/db-tests/src/test/java/gov/nasa/jpl/aerie/database/TagsTests.java#L89-L101) in [`PlanCollaboration.TagsTests`](https://github.com/NASA-AMMOS/aerie/blob/5a9ffa15010924cad1e62ef206b08dfe3852aac3/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java#L2716)

Since I needed to add an `insertUser` statement for the user used in `duplicatePlan()`, I chose to rename that user from the generic `DBTests` to the more specific `PlanCollaborationTests`. The same reasoning was used when renaming the user in `insertMissionModel` from `tester` to `Merlin DB Tests`

mattdailis: Ah, apologies - I was looking at the diff and didn't realize that these usernames predate this PR

PR #1005, opened 2023-06-28
Add missing columns to hasura permissions

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

When manually testing simulation permissions with @Mythicaeda and @skovati , we noticed that the "user" role was missing select permissions on the span "duration" column.

We did a survey of the other permissions and added any missing columns we found.

We included `metadata` in sequence insert permissions because otherwise it would always be null.

## Verification

Restarting hasura showed no inconsistent metadata.

## Documentation

No documentation was updated - these permissions are consistent with the default permissions laid out in #983 (pending more up-to-date permissions from https://docs.google.com/document/d/1sCIH2AhNDf6LXsSr6il44l_E4QFk3-ykH5SRPqGbdlw/edit)

## Future work

None

### DIFF
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_dataset.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_dataset.yaml
@@ -11,12 +11,12 @@ object_relationships:
 select_permissions:
   - role: user
     permission:
-      columns: [plan_id, dataset_id, offset_from_plan_start]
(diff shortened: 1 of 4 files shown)


skovati (commented):
Thinking out loud here, but it would be nice to have some sort of check for this within our CI, maybe grepping for column names in `metadata/**/*.yaml` files or, as @Mythicaeda suggested, having a bot comment a reminder when changes to the database are detected.

Mythicaeda (approved):

PR #1007, opened 2023-06-28
Shift operation for profiles

### LINKED ISSUE #830: Add shift operation for profiles to EDSL
In order to do rolling window constraints (i.e. no more than X minutes of this activity in any Y hour window) we need to be able to shift real profiles in time. The pattern will be:

```ts
let duration = Spans.ForEachActivity("my activity").accummulatedDuration();
return duration.plus(duration.shiftBy(X).negate()).lessThan(Y);
```

While we're at it, it can't hurt to allow shifting of discrete profiles too. It would be less used, but still could be helpful and not hard to implement.

JoelCourtney:
* **Tickets addressed:** closes #830 
* **Review:** By commit  

## Description

- This operation shifts profiles forward or backward in time. Both this and the original `Windows` shift operation are called `shiftBy` in the eDSL.
  - `real.shiftBy(Duration)`, `discrete.shiftBy(Duration)`, and `windows.shiftBy(Duration)` shifts the whole profile uniformly, and corresponds to the NEW `ShiftBy` AST node.
  - `windows.shiftBy(Duration, Duration)` is the old operation, which shift rising and falling edges separately. This now corresponds to the `ShiftWindowsEdges` AST node, which used to be called `ShiftBy`.
- The fourth commit adds behavior to the two shift nodes which shifts the bounds before giving them to subexpressions. Without this, shifting usually results in an unnecessary gap that could have been avoided and had to be manually removed with `assignGaps`.

## Verification

- [x] added tests for new shiftBy
- [x] added tests for new bounds shifting behavior

## Documentation

- [x] doc comments for `shiftBy` in eDSL

### DIFF
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -226,13 +226,21 @@ export class Windows {
    * @param fromStart duration to add from the start of each true segment
    * @param fromEnd duration to add from the end of each true segment
    */
-  public shiftBy(fromStart: AST.Duration, fromEnd: AST.Duration) : Windows {
-    return new Windows({
-      kind: AST.NodeKind.WindowsExpressionShiftBy,
-      windowExpression: this.__astNode,
-      fromStart,
-      fromEnd
-    })
+  public shiftBy(fromStart: AST.Duration, fromEnd?: AST.Duration | undefined) : Windows {
+    if (fromEnd === undefined) {
+      return new Windows({
+        kind: AST.NodeKind.ProfileExpressionShiftBy,
+        expression: this.__astNode,
+        duration: fromStart
+      });
+    } else {
+      return new Windows({
+        kind: AST.NodeKind.WindowsExpressionShiftBy,
+        windowExpression: this.__astNode,
+        fromStart,
+        fromEnd
+      })
+    }
   }
 
   /**
@@ -707,6 +715,14 @@ export class Real {
       timepoint : timepoint.__astNode
     });
   }
+
+  public shiftBy(duration: Temporal.Duration): Real {
+    return new Real({
+      kind: AST.NodeKind.ProfileExpressionShiftBy,
+      expression: this.__astNode,
+      duration
+    })
+  }
 }
 
 /**
@@ -867,6 +883,14 @@ export class Discrete<Schema> {
       originalProfile: this.__astNode,
       defaultProfile: defaultProfile.__astNode
     });
+  }
+
+  public shiftBy(duration: Temporal.Duration): Discrete<Schema> {
+    return new Discrete<Schema>({
+      kind: AST.NodeKind.ProfileExpressionShiftBy,
+      expression: this.__astNode,
+      duration
+    })
   }
 }
 
@@ -1071,7 +1095,7 @@ declare global {
      * @param fromStart duration to add from the start of each true segment
      * @param fromEnd duration to add from the end of each true segment
      */
-    public shiftBy(fromStart: AST.Duration, fromEnd: AST.Duration): Windows;
+    public shiftBy(fromStart: AST.Duration, fromEnd?: AST.Duration | undefined): Windows;
 
     /**
      * Returns a new windows object, with all true segments shorter than or equal to the given
@@ -1334,6 +1358,8 @@ declare global {
      * @param timepoint the timepoint, represented by a Spans (must be reduced to a single point)
      */
     public valueAt(timepoint: Spans): Discrete<number>;
+
+    public shiftBy(duration: Temporal.Duration): Real;
   }
 
   /**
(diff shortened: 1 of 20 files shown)


merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -226,13 +226,21 @@ export class Windows {
...
+        duration: fromStart
+      });
+    } else {
+      return new Windows({
+        kind: AST.NodeKind.WindowsExpressionShiftBy,
+        windowExpression: this.__astNode,
+        fromStart,
+        fromEnd
+      })
+    }
```
mattdailis, line 246: Wrapping my head around this... to rationalize to myself why both versions of this method are on the `Windows` class: profile expressions are a special case of windows expressions, so both of these operations make sense for profiles. Am I close?

~As far as the semantics of the operations - `shiftBy(singleDuration)` takes the entire profile and moves it over by `singleDuration`. `shiftBy(deltaStart, deltaEnd)` takes every window individually and adjusts its start and end time accordingly. Does that track?~ (EDIT: wow I should have read the PR description more carefully - you've laid this out there already)

I'm wondering if these two operations are similar enough to share a name, or are they different enough such that this would cause confusion? I can be convinced either way here

mattdailis: Perhaps a slightly different question would help me get a better feel for this - is `shiftBy(x)` equivalent to `shiftBy(x, x)`, or is there a subtle difference?

mattdailis: Musing further - one could frame this PR as generalizing the `shiftBy` operation beyond boolean profiles. The part that's tricky is what to do if the operation causes two intervals to overlap - we'd need a way to collapse that multi-valued profile back to a single-valued profile. `Windows.shiftBy` effectively chooses "OR" as the merge operation. By restricting `shiftBy` on real and discrete profiles to `shiftBy(x, x)`, we dodge that issue, because no overlaps can happen. We also dodge tricky questions like what does it mean to stretch out a linear profile - does it reach a higher final value, or does it change its slope?

JoelCourtney: yes, `windows.shiftBy(x)` is equivalent to `windows.shiftBy(x, x)`.

- [x] I should add that to the doc comment.

> if the operation causes two intervals to overlap - we'd need a way to collapse that multi-valued profile back to a single-valued profile. *Windows.shiftBy effectively chooses "OR" as the merge operation.*

Not quite. `windows.shiftBy(fromStart, fromEnd)` can be thought of in two ways:
1. shift each true interval's bounds by `fromStart` and `fromEnd`, and apply the *converse* to each false interval (shift the bounds by `fromEnd` and `fromStart`).
2. or think of it as `windows.shiftBy(shiftRising, shiftFalling)` shifting the rising and falling edges. I've come to prefer this interpretation but it doesn't really matter.

The important thing is that both of those interpretations is that for any given segment/edge, the shift depends on whether it is true or false (or rising or falling). This makes it very difficult to extend to any other datatype: for real profiles, the shift would depend on whether the value/edge is....what? Also, the above interpretations don't produce an intermediate multi-valued boolean profile.

If we instead applied the `(fromStart, fromEnd)` shift to all windows segments in the same manner, we could get some undesired gaps:

```
input:
(-inf, 0) -> true
[0, inf) -> false

shiftBy(1, 0):
(-inf, 0) -> true
[0, 1) -> gap
[1, inf) -> false
```

mattdailis:
Tested locally with a simple GrowBanana plan and the following constraint:

```typescript
  return Real.Resource("/plant").shiftBy(Temporal.Duration.from({ hours: -24 })).lessThan(200.6);
}
```

Screenshot 2023-07-05 at 2 50 36 PM

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
mattdailis: Awesome, thanks for the explanation. The rising/falling edges mental model helps a lot

mattdailis (approved):

parkerabercrombie:
@JoelCourtney is there a example of how to use this in the docs?

JoelCourtney:
No, I didn't think it needed an example. Originally this was intended to enable rolling-window-style constraints, and strictly speaking it does. This would have needed an example, but it turned out to be impossible to represent the results in an intuitive way using our current machinery. So I'm working on a dedicated rolling window constraint node that returns more helpful results, which will get its own example in the docs when it is finished.

JoelCourtney:
@parkerabercrombie If it helps, I can post my attempt at eDSL code that uses `shiftBy` to create a rolling window constraint. It has one minor edge case problem I could not fix, and will not be the long-term intended way to create that kind of constraint; but it is technically not incorrect.

parkerabercrombie:
Thanks @JoelCourtney. @camargo mentioned earlier that you were still working on the "final" version of this. We'll wait for that.

camargo:
@parkerabercrombie Here is the issue if you want to follow along https://github.com/NASA-AMMOS/aerie/issues/1015

PR #1009, opened 2023-06-28
Add `minus` helper function

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  

## Description

Adds a `minus(number | Real)` method to `Real` in the constraints eDSL that delegates to `plus` after negating the argument.

## Verification

Added a compilation test.


Mythicaeda (approved):
🎉

PR #1012, opened 2023-06-29
Merlin server action permissions

### LINKED ISSUE #952: Implement permission checking for merlin java actions
### Checked for duplicates

No - I haven't checked

### Alternatives considered

No - I haven't considered

### Related problems

I'm frustrated when [...] happens as documented in issue-XYZ

I need or want [...]

mattdailis:
* **Tickets addressed:** Closes #952 
* **Review:** By commit  

## Description

This PR implements permissions for `getSimulationResults`, `getResourceSamples`, `getConstraintViolations`, `addExternalDataset` and `extendExternalDataset` based on the schema defined [here](https://github.com/NASA-AMMOS/aerie/discussions/983#discussioncomment-6257146).

Open question: which of the above actions does `MISSION_MODEL_OWNER` make sense for? What exactly would its semantics be?

For the time being, this PR rejects `MISSION_MODEL_OWNER` for all of the above actions.

## Verification

@Mythicaeda and I did some manual testing yesterday. To do this, create the `metadata.user_role_permissions` table:

```sql
create table metadata.user_role_permissions(
  role user_roles not null primary key,
  action_permissions jsonb not null default '{}',
  function_permissions jsonb not null default '{}'
);
```

Add some permissions for the user role:

```sql
insert into metadata.user_role_permissions (role, action_permissions, function_permissions) values ('user', '{"simulate": "PLAN_OWNER"}'::jsonb, '{}'::jsonb);
```

Then create a plan, and attempt simulation. Verify that you can only simulate if you are a plan owner.

## Documentation

The only documentation is currently here: https://github.com/NASA-AMMOS/aerie/discussions/983#discussioncomment-6257146

We need to write analogous user-facing docs for this schema.

## Future work

- Do the same for the scheduler and sequencing servers
- Consider whether this logic would be worth "librarifying"

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
@@ -7,6 +7,9 @@ import gov.nasa.jpl.aerie.merlin.server.exceptions.NoSuchPlanDatasetException;
 import gov.nasa.jpl.aerie.merlin.server.exceptions.NoSuchPlanException;
 import gov.nasa.jpl.aerie.merlin.server.models.ActivityDirectiveForValidation;
 import gov.nasa.jpl.aerie.merlin.server.services.ConstraintAction;
+import gov.nasa.jpl.aerie.merlin.server.models.DatasetId;
+import gov.nasa.jpl.aerie.merlin.server.models.HasuraAction;
+import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
 import gov.nasa.jpl.aerie.merlin.server.services.GenerateConstraintsLibAction;
 import gov.nasa.jpl.aerie.merlin.server.services.GetSimulationResultsAction;
 import gov.nasa.jpl.aerie.merlin.server.services.MissionModelService;
@@ -56,19 +59,22 @@ public final class MerlinBindings implements Plugin {
   private final GetSimulationResultsAction simulationAction;
   private final GenerateConstraintsLibAction generateConstraintsLibAction;
   private final ConstraintAction constraintAction;
+  private final PermissionsService permissionsService;
 
   public MerlinBindings(
       final MissionModelService missionModelService,
       final PlanService planService,
       final GetSimulationResultsAction simulationAction,
       final GenerateConstraintsLibAction generateConstraintsLibAction,
-      final ConstraintAction constraintAction
+      final ConstraintAction constraintAction,
+      final PermissionsService permissionsService
   ) {
     this.missionModelService = missionModelService;
     this.planService = planService;
     this.simulationAction = simulationAction;
     this.generateConstraintsLibAction = generateConstraintsLibAction;
     this.constraintAction = constraintAction;
+    this.permissionsService = permissionsService;
   }
 
   @Override
@@ -185,14 +191,15 @@ public final class MerlinBindings implements Plugin {
     }
   }
 
-  private void getSimulationResults(final Context ctx) {
+  private void getSimulationResults(final Context ctx) throws Unauthorized {
     try {
       final var body = parseJson(ctx.body(), hasuraPlanActionP);
       final var planId = body.input().planId();
 
+      this.checkPlanPermissions("simulate", body.session(), planId);
+
       final var response = this.simulationAction.run(planId);
       ctx.result(ResponseSerializers.serializeSimulationResultsResponse(response).toString());
-
     } catch (final InvalidEntityException ex) {
       ctx.status(400).result(ResponseSerializers.serializeInvalidEntityException(ex).toString());
     } catch(final InvalidJsonException ex) {
@@ -204,12 +211,14 @@ public final class MerlinBindings implements Plugin {
     }
   }
 
-  private void getResourceSamples(final Context ctx) {
+  private void getResourceSamples(final Context ctx) throws Unauthorized {
     try {
-      final var planId = parseJson(ctx.body(), hasuraPlanActionP).input().planId();
+      final var body = parseJson(ctx.body(), hasuraPlanActionP);
+      final var planId = body.input().planId();
+
+      this.checkPlanPermissions("resource_samples", body.session(), planId);
 
       final var resourceSamples = this.simulationAction.getResourceSamples(planId);
-
       ctx.result(ResponseSerializers.serializeResourceSamples(resourceSamples).toString());
     } catch (final InvalidJsonException ex) {
       ctx.status(400).result(ResponseSerializers.serializeInvalidJsonException(ex).toString());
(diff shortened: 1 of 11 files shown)


Mythicaeda:
> @Mythicaeda and I did some manual testing yesterday.

Once the DB tables are in, would it be possible to convert some of the manual tests to E2ETests?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PermissionsRepository.java
```diff
@@ -0,0 +1,84 @@
+package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
+
+import gov.nasa.jpl.aerie.merlin.server.models.DatasetId;
+import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.merlin.server.remotes.PermissionType;
+
+import javax.sql.DataSource;
+import java.sql.SQLException;
+
+public final class PermissionsRepository {
```
skovati, line 10: Is this `PermissionsRepository` setting a new precedent that we should "flatten" interfaces like `ResultsCellRepository` to only the postgres implementation since `InMemory*` and `Stub*` aren't being effectively utilized?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -181,14 +187,15 @@ private void getResourceTypes(final Context ctx) {
     }
   }
 
-  private void getSimulationResults(final Context ctx) {
+  private void getSimulationResults(final Context ctx) throws Unauthorized {
```
skovati, line 194: What is the reasoning for throwing this `Unauthorized` exception at this top level instead of catching it below the try block in each handler function?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -200,12 +207,14 @@ private void getSimulationResults(final Context ctx) {
...
 
-  private void getResourceSamples(final Context ctx) {
+  private void getResourceSamples(final Context ctx) throws Unauthorized {
     try {
-      final var planId = parseJson(ctx.body(), hasuraPlanActionP).input().planId();
+      final var body = parseJson(ctx.body(), hasuraPlanActionP);
+      final var planId = body.input().planId();
 
-      final var resourceSamples = this.simulationAction.getResourceSamples(planId);
+      this.checkPlanPermissions("resource_samples", body.session(), planId);
```
skovati, line 219: Is it possible to encode these `action_permission` keys like `"resource_samples"` as an enum to better capture the possible options in our Java code? Otherwise, it seems the only place these are listed are this [github discussion](https://github.com/NASA-AMMOS/aerie/discussions/983#discussioncomment-6257146)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -448,4 +464,20 @@ private <T> T parseJson(final String subject, final JsonParser<T> parser)
       throw new InvalidJsonException(e);
     }
   }
+
+  private void checkPlanPermissions(
```
skovati, line 473: Will these functions eventually contain more logic? Otherwise our action handlers could just directly call `permissionsService.check`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/Unauthorized.java
```diff
@@ -0,0 +1,39 @@
+package gov.nasa.jpl.aerie.merlin.server.http;
+
+import gov.nasa.jpl.aerie.merlin.server.models.DatasetId;
+import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.merlin.server.remotes.PermissionType;
+
+public class Unauthorized extends Exception {
+  public Unauthorized(final String role, final String action) {
+    super("Role \"" + role + "\" is not allowed to perform action \"" + action + "\"");
```
skovati, line 9:
```suggestion
    super("Role \"%s\" is not allowed to perform action \"%s\"".formatted(role, action));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/Unauthorized.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public Unauthorized(
+      final String action,
+      final String role,
+      final String username,
+      final PermissionType permissionType,
+      final PlanId planId) {
+    super("User \"%s\" with role \"%s\" cannot perform \"%s\" because they are not a \"%s\" for plan with id \"%d\"".formatted(
```
skovati, line 18: We might be able to leverage Java string blocks (e.g. `"""User "%s" with role..."""`) to avoid the need to escape these double quotes.

skovati (commented):
Looks really good! Just a few non-blocking comments, but I really like this overall design. I'll submit another final review once the `MerlinBindingTests` issue is sorted out (in case that requires significant changes)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PermissionsRepository.java
```diff
@@ -0,0 +1,84 @@
...
+      final String username,
+      final String action,
+      final DatasetId datasetId) {
+    return switch (permissionType) {
+      case NO_CHECK -> true;
+      case OWNER -> throw new RuntimeException("OWNER permission cannot be applied to action \"" + action + "\"");
+      case MISSION_MODEL_OWNER -> throw new RuntimeException("MISSION_MODEL_OWNER permission cannot be applied to action \"" + action + "\"");
+      case PLAN_OWNER -> getPlanDatasetPermissions(username, datasetId).isPlanOwner();
+      case PLAN_COLLABORATOR -> getPlanDatasetPermissions(username, datasetId).isPlanCollaborator();
+      case PLAN_OWNER_COLLABORATOR -> getPlanDatasetPermissions(username, datasetId).isPlanOwnerOrCollaborator();
```
Mythicaeda, line 58: As far as I'm aware, this is function is solely called by `extendExternalDataset` action. If so, given that a plan dataset can belong by multiple plans, I don't think Plan Ownership permissions make sense as allowed permissions.

mattdailis:
This PR is subsumed by #1032 CC: @skovati

PR #1013, opened 2023-06-29
Track constraint checking

### LINKED ISSUE #514: Workflow for Constraints checking request with no simulation data

### LINKED ISSUE #637: Track constraint checking status in the database
We need to reflect the status of constraint checking the database. This should be similar to how we store simulation and scheduling statuses.

### LINKED ISSUE #997: Store constraint violations
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

Currently Aerie does not persist the result of a constraint check, the constraints can only be checked on-demand. For a mission that uses constraints to enforce flight rule checks (e.g. Europa Clipper) the constraint check result be should recorded so that operators can refer back to it, and so that mission tools can query for the constraint check data to include in reports. Without a persistent record there is no way to reliably recreate the run-for-record flight rule check in the case that a constraint definition is modified. For something as fundamental to mission safety as flight rule checks, I don't think we can rely on process alone to prevent changes to the constraint definitions, the result of the constraint check should be recorded.

I want Aerie to record constraint checks in the database. I want to be able to query the API for all constraint checks for a given simulation dataset and constraint run, and get back all constraints that were checked (including those with no violations) and all violations found (including the window of time the constraint was in violation).

### LINKED ISSUE #1006: Check constraint violations on a specific `simulationDatasetId` if present
### Related problems

After merging #976, the `constraintViolations` Hasura action now accepts an optional `simulationDatasetId`. However, this ID only filters the external datasets included in constraint checking, not the simulation results used for checking (which still defaults to the latest simulation).

If the sim dataset ID is provided, it would be beneficial to check constraints for that specific simulation dataset ID. This would enable constraint checking on historical simulation results, rather than solely relying on the latest one.
 
https://github.com/NASA-AMMOS/aerie/blob/4de9ebd1c26976aefe86dfdbf7a02da9dcb0ba9a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/GetSimulationResultsAction.java#L124 should be changed to retrieve the passed `simulationDatasetId` if it's nonempty; otherwise, retain the current behavior of defaulting to the latest simulation.

cohansen:
Closes #997
Closes #514 
Closes #637 

* **Review:** By file

## Description
This PR adds constraint run tracking. The way it's currently working is each `constraint_run` record is tied to a constraint that was checked. Regardless on if that constraint had violations or not we save a record to the `constraint_run` table that includes basic constraint information and the violations that were encountered. This allows us to reuse those violations in subsequent constraint violation checks.

There is also staleness checking for the runs through a field called `definition_outdated`. This field tracks if the constraint definition has changed since the violations were checked last.

~~There is also staleness checking for the runs, each record gets the status `resolved` when it has been saved to the db. If the user makes a change to a constraint with any `resolved` runs, those affected runs get a status change to `constraint-outdated`. This works the same way for simulations, if a violation check happened and then the user runs a new simulation all constraint runs for that previous dataset are marked as `simulation-outdated`.~~

This is my first big PR for `aerie-merlin` so please hit me with some feedback!

## Verification
Manual tests, I need to work on adding some real tests but I wanted to get the PR up for review first. Expect another commit with tests.

To verify everything is working properly you can do the following:
1. Create a model and a plan.
Add a constraint to the plan (I used `return Real.Resource('/peel').greaterThan(x);` for a lot of my testing).
3. Add an activity to the plan and simulate.
4. Do a constraint violation check, the `constraint_run` table now should be populated with a single entry and it shouldn't be marked as outdated.
5. Go modify your constraint, that entry in the `definition_outdated` field should now be `true`.
6. Do another constraint violation check, the entry should be marked as not outdated again.

~~1. Create a model and a plan.
2. Add a constraint to the plan (I used `return Real.Resource('/peel').greaterThan(x);` for a lot of my testing).
3. Add an activity to the plan and simulate.
4. Do a constraint violation check, the `constraint_run` table now should be populated with a single entry with a status `resolved`.
7. Go modify your constraint, that entry in the db should be now marked as `constraint-outdated`.
8. Do another constraint violation check, a new entry should exist in the table with a status `resolved`.
9. Add another activity and re-simulate.
10. The newest entry in the `constraint_run` table should now be marked `simulation-outdated`.~~

You can also run the constraint violation check in succession to test the caching behavior.

## Documentation
I'll update aerie docs to add information about constraint run tracking after this PR is created. Link will be added here.

## Future work
Going to work on #514 as a separate ticket.


merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -82,3 +82,20 @@ create trigger set_timestamp
 before update on "constraint"
 for each row
 execute function constraint_set_updated_at();
+
+create or replace function constraint_check_constraint_run()
```
Mythicaeda: With the SQL code, we use declarative `create`s outside of migrating function definitions, since we *know* that the function does not exist.
```suggestion
create function constraint_check_constraint_run()
```

merlin-server/sql/merlin/tables/simulation_dataset.sql
```diff
@@ -301,3 +301,21 @@ create trigger update_offset_from_plan_start_trigger
 before insert or update on simulation_dataset
 for each row
 execute function update_offset_from_plan_start();
+
+create or replace function simulation_dataset_check_constraint_run()
```
Mythicaeda:
```suggestion
create function simulation_dataset_check_constraint_run()
```

merlin-server/sql/merlin/tables/constraint_run.sql
```diff
@@ -0,0 +1,49 @@
+create type constraint_status as enum('resolved', 'constraint-outdated', 'simulation-outdated');
```
Mythicaeda: I believe you mentioned extracting this statement out to a file in `sql/merlin/domain-types`, alongside some of the other Merlin "run status" enums.

deployment/hasura/migrations/AerieMerlin/20_constraint_run/up.sql
```diff
@@ -0,0 +1,86 @@
...
+  returns trigger
+  security definer
+  language plpgsql as $$begin
+  update constraint_run
+  set status = 'constraint-outdated'
+  where constraint_id = new.id
+    and constraint_definition != new.definition
+    and status = 'resolved';
+  return new;
+end$$;
```
Mythicaeda: There is an indention difference between here and the function declared in `constraint.sql`. Because the function definition (the part between the `$$`s) is stored as a string, indention differences are considered definition mismatches.

deployment/hasura/migrations/AerieMerlin/20_constraint_run/up.sql
```diff
@@ -0,0 +1,86 @@
...
+create or replace function simulation_dataset_check_constraint_run()
+  returns trigger
+  security definer
+  language plpgsql as $$begin
+  update constraint_run cr
+  set status = 'simulation-outdated'
+  from simulation s
+  where cr.status = 'resolved'
+    and s.plan_id = cr.plan_id
+    and s.id = new.simulation_id;
```
Mythicaeda: These two lines are indented differently in `simulation_dataset`. This is the preferred indentation style.

merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -82,3 +82,20 @@ create trigger set_timestamp
...
+  set status = 'constraint-outdated'
+  where constraint_id = new.id
+  and constraint_definition != new.definition
+  and status = 'resolved';
+  return new;
+end$$;
+
+create trigger constraint_check_constraint_run_trigger
+after update on "constraint"
+for each row
```
Mythicaeda: You can add the following condition on this trigger to avoid running it when the definition doesn't change:
```sql
when (new.definition != old.definition`)
```

merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -82,3 +82,20 @@ create trigger set_timestamp
...
+
+create or replace function constraint_check_constraint_run()
+returns trigger
+security definer
+language plpgsql as $$begin
+  update constraint_run
+  set status = 'constraint-outdated'
+  where constraint_id = new.id
+  and constraint_definition != new.definition
+  and status = 'resolved';
```
Mythicaeda: Should this trigger also have the opposite update statement? As in, if the status _was_ `constraint-outdated`, but the definitions now match again, it is returned to `resolved`.

merlin-server/sql/merlin/tables/constraint.sql
cohansen: 👍

merlin-server/sql/merlin/tables/constraint.sql
cohansen: That's a good point, I forgot that we spoke about this in our initial meeting. I guess that's a larger design question, do we like the idea of resolving past constraint runs if the conditions are met?

Mythicaeda: My gut reaction is yes, because the fact that the status is `constraint-outdated` means that that cached result *would be valid* if the current constraint definition matched the one in the saved run. Although revisiting the trigger on `simulation_dataset`, it looks like only the currently-resolved constraints get marked as `simulation_outdated`, not all constraints from that dataset run.

merlin-server/sql/merlin/tables/constraint_run.sql
```diff
@@ -0,0 +1,49 @@
...
+  constraint constraint_run_to_simulation_dataset
+    foreign key (simulation_dataset_id)
+      references simulation_dataset
+      on delete cascade,
+  constraint constraint_run_requested_by
+    foreign key (requested_by)
+      references metadata.users
+      on update cascade
+      on delete set null
+);
```
Mythicaeda: This table needs a Primary Key. `(constraint_id, constraint_definition, simulation_dataset_id)` seems to be the smallest tuple that will uniquely identify a row.

merlin-server/sql/merlin/tables/constraint.sql
Mythicaeda: Actually, do we need to track `simulation_outdated` as a status? If the sim_dataset_id differs, we _should_ get a cache miss when we try to check for if that constraint definition has been run on the current sim_dataset.

cohansen: I like `simulation-outdated` because from the db perspective you can easily see which runs are outdated, if we rely on just checking with the `sim_dataset_id` we don't have that immediate visibility.

Mythicaeda: But if you can run constraints on against an older sim dataset (which, if it's not part of this PR, is a desired next step that we've discussed with @skovati), then `simulation_outdated` doesn't *mean* anything.

merlin-server/sql/merlin/tables/constraint_run.sql
Mythicaeda: Additionally, we may want to add an `index` on this table on `simulation_dataset_id`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/InMemoryResultsCellRepository.java
```diff
@@ -154,6 +155,12 @@ public InMemorySimulationResultsHandle(final SimulationResults simulationResults
       this.simulationResults = simulationResults;
     }
 
+    @Override
+    public SimulationDatasetId getSimulationDatasetId() {
+      // TODO: What should it return here?
+      return new SimulationDatasetId(1L);
```
Mythicaeda: I believe throwing an `UnsupportedOperationException` would make sense here, since Simulation Dataset Ids _are_ unsupported by the In Memory implementation.

merlin-server/sql/merlin/tables/constraint_run.sql
```diff
@@ -0,0 +1,49 @@
+create table constraint_run (
+  constraint_id integer not null,
+  constraint_definition text not null,
+  plan_id integer not null,
+  simulation_dataset_id integer not null,
+
+  status constraint_status not null default 'resolved',
+  violations jsonb null,
```
Mythicaeda: Why `null` rather than `not null default '{}'`? [The Java code expects the latter](https://github.com/NASA-AMMOS/aerie/blob/72aa6169ba37662a194cd86bb7f74fa2ded659fc/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java#L44-L47). Additionally, that is usually our standard for `jsonb` columns.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
```diff
@@ -0,0 +1,66 @@
...
+
+        // The constraint run didn't have any violations
+        if (violationString.equals("{}")) {
+          constraintRuns.add(new ConstraintRunRecord(constraintId, status, null));
+        } else {
+          constraintRuns.add(new ConstraintRunRecord(
+              constraintId,
+              status,
+              getJsonColumn(results, "violations", violationP)
+                  .getSuccessOrThrow($ -> new Error("Corrupt violations cannot be parsed: " + $.reason()))));
```
Mythicaeda: Can the parser not handle an empty JSON?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
```diff
@@ -0,0 +1,66 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static gov.nasa.jpl.aerie.constraints.json.ConstraintParsers.violationP;
+import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.getJsonColumn;
+
+final class GetSuccessfulConstraintRunsAction implements AutoCloseable {
```
Mythicaeda: `GetValidConstraintRunsAction` or `GetLatestContstraintRunsAction` would be more accurate names, since this action will return constraints with violations.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final @Language("SQL") String sql = """
+    select
+      cr.constraint_id,
+      cr.simulation_dataset_id,
+      cr.status,
+      cr.violations
+    from constraint_run as cr
+    where cr.status = 'resolved'
+    and cr.constraint_id = any(?)
+    and cr.simulation_dataset_id in (select distinct on (simulation_id) id from simulation_dataset order by simulation_id, id desc)
```
Mythicaeda: Could you explain what exactly this part of the `where` condition means?

merlin-server/sql/merlin/tables/constraint.sql
cohansen: That's a good point, I can remove it.

Mythicaeda: When you do, then you no longer actually need a new status type, just a boolean that tracks `definition_outdated`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintService.java
```diff
@@ -0,0 +1,15 @@
...
+import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.merlin.server.models.SimulationDatasetId;
+import gov.nasa.jpl.aerie.merlin.server.remotes.postgres.ConstraintRunRecord;
+
+import java.util.List;
+import java.util.Map;
+
+public interface ConstraintService {
+  void createConstraintRuns(Map<Long, Constraint> constraintMap, Map<Long, Violation> violations, PlanId planId, SimulationDatasetId simulationDatasetId);
+  Map<Long, ConstraintRunRecord> getPreviouslyResolvedConstraints(List<Constraint> constraints);
```
Mythicaeda: When the `simulation_outdated` status is removed, this method will likely need to include `simulationDatasetId`, otherwise you will end up fetching the resolved constraint for _all_ simulation datasets.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -0,0 +1,244 @@
...
+
+      for (final var entry : constraintCode.entrySet()) {
+
+        // Pipeline switch
+        // To remove the old constraints pipeline, delete the `useNewConstraintPipeline` variable
+        // and the else branch of this if statement.
+        final var constraint = entry.getValue();
+        final Expression<List<Violation>> expression;
+
+        // TODO: cache these results
```
Mythicaeda: Are the TODOs in this file still accurate?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalConstraintService.java
```diff
@@ -0,0 +1,41 @@
...
+  @Override
+  public void createConstraintRuns(final Map<Long, Constraint> constraintMap, final Map<Long, Violation> violations, final PlanId planId, final SimulationDatasetId simulationDatasetId) {
+    this.constraintRepository.insertConstraintRuns(constraintMap, violations, planId, simulationDatasetId.id());
+  }
+
+  @Override
+  public Map<Long, ConstraintRunRecord> getPreviouslyResolvedConstraints(List<Constraint> constraints) {
+    final var resolvedConstraintRuns = new HashMap<Long, ConstraintRunRecord>();
+    final var constraintIds = constraints.stream().map(Constraint::id).collect(Collectors.toList());
+    final var constraintRuns = constraintRepository.getSuccessfulConstraintRuns(constraintIds);
```
Mythicaeda: Why does `constraintRepository.getSuccessfulConstraintRuns` return a List rather than a Map?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubConstraintService.java
Mythicaeda, line 1: Why do we need a `Stub` implementation of this service? The methods are, well, stubs, and the `Local` version only delegates to the `ConstraintRepository` interface, which makes it look like the distinction between `Local` vs `Networked` vs any other type of `ConstraintService` would happen there.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
cohansen: Correct, it was throwing errors when it was empty.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
cohansen: Yeah!
1. Only look for `resolved` constraint runs
2. Only get the `constraint_run`s for the passed ids
3. Make sure we only grab the corresponding `constraint_run` from the most recent `simulation_dataset`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalConstraintService.java
cohansen: I was trying to limit any logic other than pulling from the db from being in the repository code, if it makes more sense to map it there rather than inside the service I can move that code and change return types.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
cohansen: I'm not sure if this TODO originally was talking about just caching the DSL compilation result, or the constraint check like I do in this PR. Any insight @JoelCourtney ?

The later 2 TODOs are still relevant though.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalConstraintService.java
Mythicaeda: I think it would make more sense to map it there, yeah.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
Mythicaeda: Thanks for explaining! Regarding (3), this query returns incorrect data when the most recent `simulation_dataset` does not belong to the plan that we're checking constraints for. The `simulation_dataset_id` ought to be passed in instead.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public List<ConstraintRunRecord> get() throws SQLException, ConstraintRunRecord.Status.InvalidRequestStatusException {
+    this.statement.setArray(1, this.statement.getConnection().createArrayOf("integer", constraintIds.toArray()));
+
+    try (final var results = this.statement.executeQuery()) {
+      final var constraintRuns = new ArrayList<ConstraintRunRecord>();
+
+      while (results.next()) {
+        final var constraintId = results.getLong("constraint_id");
+        final var status = ConstraintRunRecord.Status.fromString(results.getString("status"));
```
Mythicaeda: This line needs to be updated to `results.getBoolean("status")`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
```diff
@@ -0,0 +1,55 @@
...
+import java.sql.SQLException;
+import java.util.Map;
+
+import static gov.nasa.jpl.aerie.constraints.json.ConstraintParsers.violationP;
+
+/* package local */ class InsertConstraintRunsAction implements AutoCloseable {
+  private static final @Language("SQL") String sql = """
+    insert into constraint_run (constraint_id, constraint_definition, plan_id, simulation_dataset_id, status, violations)
+    values (?, ?, ?, ?, ?::constraint_status, ?::json)
+  """;
```
Mythicaeda:
```suggestion
  private static final @Language("SQL") String sql = """
    insert into constraint_run (constraint_id, constraint_definition, plan_id, simulation_dataset_id, violations)
    values (?, ?, ?, ?, ?::json)
  """;
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -274,6 +275,31 @@ static JsonParser<Transition> transitionP(JsonParser<ProfileExpression<?>> profi
...
+          .field("start", durationP)
+          .field("end", durationP)
+          .field("startInclusivity", inclusivityP)
+          .field("endInclusivity", inclusivityP)
+          .map(
+              untuple((start, end, startInclusivity, endInclusivity) -> Interval.between(start, startInclusivity, end, endInclusivity)),
+              $ -> tuple($.start, $.end, $.startInclusivity, $.endInclusivity)
+          );
+
+  public static final JsonParser<Violation> violationP =
```
Mythicaeda: Minor: Why are these both `public` rather than `package-private` like the other parsers?

merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -82,3 +82,28 @@ create trigger set_timestamp
...
+    where constraint_id = new.id
+      and constraint_definition == new.definition
+      and definition_outdated = true;
+  end if;
+  return new;
+end$$;
+
+create trigger constraint_check_constraint_run_trigger
+  after update on "constraint"
+  for each row
```
Mythicaeda, line 108: The `when (new.definition != old.definition)` condition I mentioned on the original trigger will still be useful here.

merlin-server/sql/merlin/tables/constraint_run.sql
```diff
@@ -0,0 +1,46 @@
...
+
+comment on table constraint_run is e''
+  'A single constraint run, used to cache violation results to be reused if the constraint and simulation are not stale.';
+
+comment on column constraint_run.constraint_id is e''
+  'The constraint that we are evaluating during the run.';
+comment on column constraint_run.constraint_definition is e''
+  'The definition of the constraint that is being checked, used to determine staleness.';
+comment on column constraint_run.simulation_dataset_id is e''
+  'The simulation dataset id from when the constraint was checked, used to determine staleness.';
```
Mythicaeda:
```suggestion
comment on column constraint_run.constraint_definition is e''
  'The definition of the constraint when it was checked, used to determine staleness.';
comment on column constraint_run.simulation_dataset_id is e''
  'The simulation dataset id from when the constraint was checked.';
```

merlin-server/sql/merlin/tables/constraint_run.sql
```diff
@@ -0,0 +1,46 @@
...
+      on delete cascade,
+  constraint constraint_run_requested_by
+    foreign key (requested_by)
+      references metadata.users
+      on update cascade
+      on delete set null
+);
+
+comment on table constraint_run is e''
+  'A single constraint run, used to cache violation results to be reused if the constraint and simulation are not stale.';
```
Mythicaeda:
```suggestion
comment on table constraint_run is e''
  'A single constraint run, used to cache violation results to be reused if the constraint definition is not stale.';
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -0,0 +1,243 @@
...
+    for (ConstraintRunRecord constraintRun : validConstraintRuns.values()) {
+      constraintCode.remove(constraintRun.constraintId());
+
+      if (constraintRun.violation() != null) {
+        violations.put(constraintRun.constraintId(), constraintRun.violation());
+      }
+    }
+
+    // If the lengths don't match we need check the left-over constraints.
+    if (validConstraintRuns.size() != constraintSize) {
```
Mythicaeda: Minor: the check `constraintCode.isEmpty()` would be cleaner.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
cohansen: I don't remember why `intervalP` was switched, but I changed it back and it seems fine.

`violationP` is public now since we're using it outside of the `constraints` module.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
cohansen: Yeah I missed this, it's all fixed up now!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSuccessfulConstraintRunsAction.java
cohansen: I think you were seeing a weird state, that line of code is gone now!

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubConstraintService.java
cohansen: The `StubConstraintService` just gives us something to provide inside of `MerlinBindingsTest`. I know we spoke of removing it since it doesn't test anyways, but this was my path to getting code that built.

camargo (approved):
Works great! Tested with a couple constraints in the demo mission model. Created https://github.com/NASA-AMMOS/aerie-ui/issues/770 to update the UI to use the new table. Good to go.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: ~~Did you push this change? I'm not seeing it.~~  I had to hard refresh GitHub. I see it now 👍

cohansen: Yeah I did!
https://github.com/NASA-AMMOS/aerie/blob/ecdb34191f4e1093ff3c662cd65da36412381923/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java#L15-L19

deployment/hasura/metadata/databases/AerieMerlin/tables/public_constraint_run.yaml
```diff
@@ -0,0 +1,21 @@
+table:
+  name: constraint_run
+  schema: public
+object_relationships:
+  - name: constraint
+    using:
+      foreign_key_constraint_on: constraint_id
+  - name: simulation_dataset
+    using:
+      foreign_key_constraint_on: simulation_dataset_id
```
Mythicaeda, line 10: Is there any value in the reverse relationships?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -221,7 +225,7 @@ private void getConstraintViolations(final Context ctx) {
       final var planId = input.planId();
       final var simulationDatasetId = input.simulationDatasetId();
 
-      final var constraintViolations = this.simulationAction.getViolations(planId, simulationDatasetId);
```
Mythicaeda, line 224: `getViolations` needs to be removed from `GetSimulationResultsAction`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -0,0 +1,244 @@
...
+      final var environment = new EvaluationEnvironment(realExternalProfiles, discreteExternalProfiles);
+
+      final var realProfiles = new HashMap<String, LinearProfile>();
+      final var discreteProfiles = new HashMap<String, DiscreteProfile>();
+
+      for (final var entry : constraintCode.entrySet()) {
+
+        // Pipeline switch
+        // To remove the old constraints pipeline, delete the `useNewConstraintPipeline` variable
+        // and the else branch of this if statement.
```
Mythicaeda: What `if` statement is this comment referring to?

cohansen: I'm not exactly sure, more comments that were there when I got to the code.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
cohansen: Forgot to do this as a cleanup step!

deployment/hasura/metadata/databases/AerieMerlin/tables/public_constraint_run.yaml
cohansen: I could see an argument for wanted to see the constraint runs for a given simulation dataset so maybe. I could go either way on the constraints reverse relationship.

cohansen:
> Works great! Tested with a couple constraints in the demo mission model. Created [NASA-AMMOS/aerie-ui#770](https://github.com/NASA-AMMOS/aerie-ui/issues/770) to update the UI to use the new table. Good to go.

Thanks Chris!

deployment/hasura/metadata/databases/AerieMerlin/tables/public_constraint_run.yaml
Mythicaeda: Yeah, I'd say do the sim dataset reverse relationship. The constraints reverse relationship can be added at a future time if it's desired.

Mythicaeda (approved):
LGTM! Thanks Cody!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: I'd argue to remove the comment then, since the code it's on doesn't relate to the comment.

PR #954, opened 2023-05-30
Parallelize CI builds/tests

skovati:
* **Tickets addressed:** Closes #943 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description
The `publish` workflow has been rewritten using a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs), which builds, pushes, and scans each image in parallel instead of sequentially, as seen in the figure below.

Screenshot 2023-05-30 at 3 31 03 PM

This reduces the `publish` workflow time significantly, from ~18 minutes to ~6.

Additionally, the `scan` step has been extracted to a separate matrix of jobs, allowing us to fail the each scan individually if critical vulnerabilities are detected in the respective container image. The `scan` results are uploaded as HTML documents which can be checked for found critical vulnerabilities and their fixes, if any exist.

## Verification
Extensive testing has happened in a clone of the `aerie` repo called [copie](https://github.com/skovati/copie), which allowed me to iterate on workflows without failed actions cluttering the main repo. However, CI tends to be quite brittle, so further hotfixes may be necessary if the new workflows don't perfectly integrate with `aerie` proper.

## Documentation
None

## Future work
None

PR #1014, opened 2023-07-06
Restore docker-compose compatibility check

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

For those who use `docker-compose` instead of `docker compose` (note hyphen vs space), there are more stringent requirements on what constitutes a legal `docker-compose.yml` file. In particular, the literal `false` is disallowed, in favor of `"false"`. When I run `docker-compose up -d` on the `develop` branch, I see the following error message:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.hasura.environment.HASURA_GRAPHQL_INFER_FUNCTION_PERMISSIONS contains false, which is an invalid type, it should be a string, number, or a null
```

This change was introduced in https://github.com/NASA-AMMOS/aerie/pull/1001, and, in theory, would have been caught by the workflow check introduced in https://github.com/NASA-AMMOS/aerie/pull/139 had it not been removed (inadvertently?) in https://github.com/NASA-AMMOS/aerie/pull/954. I would like @skovati's input in particular, if there is any reason that this workflow step is detrimental.

This PR is merely a cherry-pick of 96e38505a4427e08bd0a04348a2f095abd5dbfcb, originally committed in https://github.com/NASA-AMMOS/aerie/pull/139

## Verification

After I open this PR, I will wait to see if the check fails. If it does, I will add a follow-up commit to add quotes to `false`.

EDIT: It took a couple tries - but adding `|| exit` did the trick :+1:. The trouble was that the previous version of the code exited with the status of the last file checked - earlier failures were masked by later successes. 

## Documentation

We do not have developer docs for this aspect of CI - the Github trail is the documentation.

## Future work

- None


skovati:
I had removed this check in #954 due to the fact that compose v1 (i.e. `docker-compose` vs `docker compose`) is [officially deprecated](https://docs.docker.com/compose/migrate/) and hasn't received security updates for a few years.

PR #1018, opened 2023-07-07
EPOCH HMS String Parsing should support Negative Duration

goetzrrGit:
* **Tickets addressed:** Closes [#647](https://github.com/NASA-AMMOS/aerie-ui/issues/647)
* **Review:** By commit  

## Description
The epoch duration should allow both positive and negative durations, while the relative duration should only accept positive durations. We have made updates to the custom worker on the user interface (UI) to inform users about invalid time formats. Additionally, the eDSL will now alert users if a negative time value is not supported, preventing the creation of an invalid SeqJson.

```ts
E`-12:30:00`.SEQ_VAR_CMD
E`+12:30:00`.SEQ_VAR_CMD
E`-003T12:30:00`.SEQ_VAR_CMD

R`12:30:00`.SEQ_VAR_CMD
R`-12:30:00`.SEQ_VAR_CMD <-- invalid
R`003T12:30:00`.SEQ_VAR_CMD <-- invalid
```
## Verification
Updated e2e test and did some manual testing

## Future work
Address time values that are 90s as they should be converted to 1:30

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.spec.ts
@@ -575,3 +575,94 @@ describe('Sequence', () => {
     });
   });
 });
+
+describe('Time Validation', () => {
+  it('should have valid Epoch and Relative Times', () => {
+    let epoch = CommandStem.new({
+      stem: 'TEST',
+      arguments: {
+      },
+      epochTime : hmsToDuration('001T00:01:34.284' as HMS_STRING,true),
+    })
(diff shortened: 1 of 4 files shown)


cohansen (changes requested):
Looks good, my one request would be to add tests with signed time for relative and DOY for relative as well expecting errors!

goetzrrGit:
@cohansen I have updated the test to ensure that errors are correctly checked for Day of Year (DOY) and negative signs for relative times.

cohansen (approved):
Looks good!

PR #1019, opened 2023-07-07
Permissions Table and Hasura Function Permissions

### LINKED ISSUE #944: Implement permission checking for hasura_functions
As mentioned in the [original Hasura Permissions issue](#861) and in #940, due to restrictions on what permissions Hasura allows us to set, some deviations from the planned permissions were made. See #940 for the description of these deviations. This ticket is for resolving those discrepancies.

### LINKED ISSUE #1020: Add an "aerie_admin" role to Hasura
This ticket is to add an `aerie_admin` role to Hasura that mirrors the permissions of `admin`. This will allow us to set custom permissions on `aerie_admin` so we can do things like automatically set the owner when you are the `aerie_admin` role (Hasura restricts setting custom permissions for their default `admin` role). This will give us more consistent permission behavior across Aerie for the core roles we support.

We also need to update Hasura metadata permissions that automatically set fields related to the username when you are "aerie_admin".

See this thread: https://jpl.slack.com/archives/CJ07FGRPH/p1688763689502879

Mythicaeda:
* **Tickets addressed:** 
    *  Closes #944.
    *  Closes #1020.
* **Review:** By commit  

## Description

- New Table `user_role_permission` was added that stores the action and function permissions for a role. 
- Functions were added for getting and checking permissions.
- All Hasura functions except for `get_resources_at_start_offset` were updated to take in the current session.
- Default Hasura 'admin' role has been replaced with 'aerie_admin'. Permissions have been added for new role.

## Verification

- Existing DB Tests were updated to include a mock Hasura session JSON when calling Hasura functions and to reflect the 'admin' -> 'aerie_admin' role name change.
- PermissionsTest were added

## Documentation

Docs ticket is [here](https://github.com/NASA-AMMOS/aerie-docs/issues/61).

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
@@ -877,16 +877,16 @@ public class AnchorTests {
         // Reanchor to Plan Start
         var results = statement.executeQuery(
             """
-                select hasura_functions.delete_activity_by_pk_reanchor_plan_start(%d, null)
-                """.formatted(activityId));
+                select hasura_functions.delete_activity_by_pk_reanchor_plan_start(%d, null, '%s'::json)
+                """.formatted(activityId, merlinHelper.admin.session()));
         if (results.next()) {
           fail();
         }
 
         results = statement.executeQuery(
             """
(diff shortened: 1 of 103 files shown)


cohansen (approved):
I think you answered all of my questions during the walkthrough, otherwise I may have some more specific feedback when I start using this for sequencing!

PR #1027, opened 2023-07-14
Implement ShiftEdges for Spans

JoelCourtney:
* **Review:** By commit  

## Description
This renames `ShiftWindowsEdges` to `ShiftEdges` and generifies it to work on Spans too. Now we can call `spans.shiftBy(dur1, dur2)` in the eDSL.

## Verification
- [x] AST test
- [x] compilation test

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
@@ -217,7 +217,8 @@ public class Spans implements IntervalContainer<Spans>, Iterable<Segment<Optiona
 
   @Override
   public Spans shiftEdges(final Duration fromStart, final Duration fromEnd) {
-    return this.map($ -> Interval.between($.start.plus(fromStart), $.startInclusivity, $.end.plus(fromEnd), $.endInclusivity));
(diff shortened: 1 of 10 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/Spans.java
```diff
@@ -215,6 +215,16 @@ public Spans ends() {
     return this.map($ -> Interval.at($.end));
   }
 
+  @Override
+  public Spans shiftEdges(final Duration fromStart, final Duration fromEnd) {
+    return this.map($ -> Interval.between($.start.plus(fromStart), $.startInclusivity, $.end.plus(fromEnd), $.endInclusivity));
+  }
+
+  @Override
+  public Spans select(final Interval... intervals) {
```
mattdailis, line 225: Just wrapping my head around `Spans::select` - this may split a span into multiple spans, if it spans two selection intervals - right?

JoelCourtney: That is correct. I'll admit it does feel a little weird, but I couldn't think of a reason why that would be a bad way to do it.

JoelCourtney:
In order to implement `ShiftEdges`, I had to also implement a `select` function on spans. So I also exposed that one to the eDSL. Both are tested and ready for review.

mattdailis (approved):

PR #1029, opened 2023-07-17
Fix Merge Function Volatility and Add Missing Select Permissions

Mythicaeda:
* **Review:** By commit  

## Description

- Adds missing SELECT permissions to `plan_snapshot`. Additionally, `aerie_admin` can DELETE snapshots
- Adjusts the volatility of `get_conflicting_activities` and `get_non_conflicting_activities` from `STABLE` to `VOLATILE`. This was needed, as `check_merge_permissions` is a procedure, which cannot be called in `STABLE` functions

## Verification

@camargo will validate against the UI


camargo (approved):
UI e2e tests pass! LGTM

PR #1030, opened 2023-07-17
Added permission checking to the sequencing server endpoints

### LINKED ISSUE #989: Implement permission checking for sequencing actions
This ticket is to add permission checks to the sequencing server action endpoints. See #944 for some nice utility functions that check permissions.

cohansen:
* **Tickets addressed:** Closes #989.
* **Review:** By file

## Description
This adds permission checking for the sequencing server. We keep a map that holds endpoints to the action keys stored in the database to check against, if the user calls an endpoint in that map (or it gets triggered through a Hasura action) we make a GQL query to check and see if that user can perform the action.

## Verification
Manual testing performed using the role selector in the UI.

## Documentation
NA

## Future work
If we go the route of merging the database the GQL call should be changed to calling the db function instead.


sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,101 @@
...
+        key: ENDPOINTS_TO_ACTION_KEY[url],
+      },
+    );
+
+    if (permissionCheckQuery.user_role_permission_by_pk.action_permissions === null) {
+      return false;
+    }
+  }
+
+  return true;
```
dyst5422: I highly recommend inverting this logic to deny by default/fall-through rather than allow by default/fall-through. Harder to accidentally allow unintentional access that way.

dyst5422 (changes requested):

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,101 @@
...
+      gql`
+        query checkUserPermissions($role: user_roles_enum!, $key: String!) {
+          user_role_permission_by_pk(role: $role) {
+            action_permissions(path: $key)
+          }
+        }
+      `,
+      {
+        role: hasuraSession['x-hasura-role'],
+        key: ENDPOINTS_TO_ACTION_KEY[url],
```
Mythicaeda: I thought we found that this needed to be formatted as `$.ENDPOINTS_TO_ACTION_KEY[url]`

sequencing-server/src/utils/hasura.ts
cohansen: The problem is we have some endpoints that are unprotected, I could be exhaustive in the mapping but currently `health`, `get-command-typescript`, and `get-activity-typescript` don't do any permission checking.

Mythicaeda: You could rewrite the functions as:
```ts
if (!(url in ENDPOINTS_TO_ACTION_KEY)) return true;
// The rest of the logic
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,108 @@
...
+      },
+    );
+
+    if (permissionCheckQuery.user_role_permission_by_pk.action_permissions === null) {
+      return false;
+    }
+  }
+
+  return true;
```
Mythicaeda: Just a reminder that this doesn't yet sufficiently check if the user has permission to run the function (although it seems like you're still working on this part)

sequencing-server/src/utils/hasura.ts
dyst5422: Could resolve that with an explicit whitelist of unprotected endpoints.

I'm just considering the case where another endpoint gets added and the map doesn't get updated, accidentally exposing endpoints that should be protected.

cohansen: I'm going to add a whitelist set, I think that's more readable and then we won't automatically allow usage on new endpoints.

sequencing-server/src/utils/hasura.ts
cohansen: This was working through my testing, I was able to get values out of the table for the given `key`.

sequencing-server/src/utils/hasura.ts
cohansen: I didn't see any keys (for the seq endpoints) that indicated how I should do the checking so I assumed that a `null` corresponded to not enough permissions, is there something else I should be checking here?

Mythicaeda: While `null` does correspond to "no permission", the general keys are usable on **all** endpoints, **unless there is a reason that a specific key does not make sense on a specific endpoint**.
If that is the case for any key, then specify *why* it cannot be applied so that it can be documented. Additionally, the endpoint needs to _reject_ the invalid permissions and _check_ that user satisfies the valid permissions.

Under the current implementation, the following unauthorized transactions would not be rejected:
1. The role `user` has the permission `{expand_all_activities: PLAN_OWNER_SOURCE}` and a User with that role attempts to run `expand_all_activities`. Expected behavior: the request is rejected because the permission rule to check is invalid (PLAN_OWNER_SOURCE is a merge permission and `expand_all_activities` is not a merge action)
2. The role `user` has the permission `{expand_all_activities: PLAN_OWNER}` and a User with that role _who does not owner the plan that was simulated_ attempts to run `expand_all_activities`. Expected behavior: the requested is rejected because the User does not satisfy the permission rule (they do not own the Plan).

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,207 @@
...
+      }
+    `,
+    { planId },
+  );
+
+  for (const planCollaborator of planCollaboratorRes.plan_collaborators) {
+    if (user === planCollaborator.collaborator) {
+      return true;
+    }
+  }
```
Mythicaeda: This query will simplify the function, as it will either return `null` or `$username`
```gql
query planCollaborator($planId: Int!, $username: String!) {
  plan_collaborators_by_pk(plan_id: $planId, collaborator: $username) {
    collaborator
  }
}
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,207 @@
...
+
+async function isPlanOwner(graphqlClient: GraphQLClient, planId: number, user: string): Promise<boolean> {
+  const planOwner = await graphqlClient.request<{
+    plan: { owner: string }[];
+  }>(
+    gql`
+      query planOwner($planId: Int!) {
+        plan(where: { id: { _eq: $planId } }) {
+          owner
+        }
```
Mythicaeda:
```suggestion
        plan_by_pk(id: $planId) {
          owner
        }
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,207 @@
...
+): Promise<boolean> {
+  const missionModelOwner = await graphqlClient.request<{
+    mission_model: { owner: string }[];
+  }>(
+    gql`
+      query missionModelOwner($missionModelId: Int!) {
+        mission_model(where: { id: { _eq: $missionModelId } }) {
+          owner
+        }
+      }
```
Mythicaeda:
```suggestion
      query missionModelOwner($missionModelId: Int!) {
        mission_model_by_pk(id: $missionModelId) {
          owner
        }
      }
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,230 @@
...
+    case HasuraPermissions.PLAN_OWNER:
+      return planId !== null && (await isPlanOwner(graphqlClient, username, planId));
+    case HasuraPermissions.PLAN_COLLABORATOR:
+      return planId !== null && (await isPlanCollaborator(graphqlClient, username, planId));
+    case HasuraPermissions.PLAN_OWNER_COLLABORATOR:
+      return (
+        (planId !== null && (await isPlanOwner(graphqlClient, username, planId))) ||
+        (await isPlanCollaborator(graphqlClient, username, planId))
+      );
+  }
```
Mythicaeda: The current switch statement is insufficient. Something roughly like this will satisfy the valid keys for all endpoints.
```ts
  switch (permission) {
    case HasuraPermissions.NO_CHECK:
      return true;
    case HasuraPermissions.MISSION_MODEL_OWNER:
      if (missionModelId !== null) {
         return await isMissionModelOwner(graphqlClient, username, missionModelId);
      }
      return await isMissionModelOwnerFromPlanId(graphqlClient, username, planId);
    case HasuraPermissions.OWNER:
      if ( ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_rule 
           || ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_set ) {
            return await isMissionModelOwner(graphqlClient, username, missionModelId);
      }
      return planId !== null && (await isPlanOwner(graphqlClient, username, planId));
    case HasuraPermissions.PLAN_OWNER:
      if ( ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_rule 
           || ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_set ) {
            return await isAnyPlanOwner(graphqlClient, username, missionModelId);
      }
      return planId !== null && (await isPlanOwner(graphqlClient, username, planId));
    case HasuraPermissions.PLAN_COLLABORATOR:
      if ( ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_rule 
           || ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_set ) {
            return await isAnyPlanCollaborator(graphqlClient, username, missionModelId);
      }
      return planId !== null && (await isPlanCollaborator(graphqlClient, username, planId));
    case HasuraPermissions.PLAN_OWNER_COLLABORATOR:
       if ( ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_rule 
           || ENDPOINTS_TO_ACTION_KEY[url] == create_expansion_set ) {
            return (await isAnyPlanOwner(graphqlClient, username, missionModelId)) 
                    ||  (await isAnyPlanCollaborator(graphqlClient, username, missionModelId));
      }
      return (
        (planId !== null && (await isPlanOwner(graphqlClient, username, planId))) ||
        (await isPlanCollaborator(graphqlClient, username, planId))
      );
  }
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,230 @@
...
+        }
+      }
+    `,
+    {
+      role,
+      key: ENDPOINTS_TO_ACTION_KEY[url],
+    },
+  );
+
+  const permission = permissionCheckQuery.user_role_permission_by_pk.action_permissions as HasuraPermissions | null;
```
Mythicaeda, line 127: What happens if `permission` is not a `HasuraPermission`?

Mythicaeda (changes requested):

sequencing-server/src/utils/hasura.ts
cohansen: It'll be null and then short-circuit on the next line denying the API call with an invalid permission error. I think this would only happen in the case that an invalid permission was configured in the DB or a new permission was added in the DB but not inside the `HasuraPermission` enum. This will be a lot better if (when?) we merge the databases and this can all just be a function call :)

sequencing-server/src/utils/hasura.ts
cohansen: I implemented this in a slightly different way which is hopefully more clear and still gets each case.

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+    planId = (
+      await graphqlClient.request<{
+        plan: { id: number }[];
+      }>(
+        gql`
+          query plan($simulationDatasetId: Int!) {
+            plan(where: { simulations: { simulation_dataset: { id: { _eq: $simulationDatasetId } } } }) {
+              id
+            }
+          }
```
Mythicaeda: This query joins on fewer tables:
```gql
query plan($simulationDatasetId: Int!) {
  simulation_dataset_by_pk(id: $simulationDatasetId){
    simulation {
      plan_id
    }
  }
}
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+    planId = (
+      await graphqlClient.request<{
+        mission_model: { id: number }[];
+      }>(
+        gql`
+          query mission_model($planId: Int!) {
+            plan(where: { plans: { id: { _eq: $planId } } }) {
+              id
+            }
+          }
```
Mythicaeda: The query you have is incorrect (it's returning the plan's id). This query will return the mission model id:
```gql
query mission_model($planId: Int!) {
  plan_by_pk(id: $planId) {
    model_id
  }
}
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+  return planId;
+}
+
+async function getMissionModelId(
+  graphqlClient: GraphQLClient,
+  planId?: number,
+  simulationDatasetId?: number,
+): Promise<number> {
+  if (planId) {
+    planId = (
```
Mythicaeda: I believe this should be `missionModelId`? In general, it looks like this method is calling `missionModelId` `planId`

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+          }
+        `,
+        { simulationDatasetId },
+      )
+    ).mission_model[0]?.id;
+  }
+
+  if (planId === null || planId === undefined) {
+    throw new Error('Could not determine the plan based on the provided missionModelId or simulationDatasetId');
+  }
```
Mythicaeda:
```suggestion
  if (missionModelId === null || missionModelId === undefined) {
    throw new Error('Could not determine the plan based on the provided planId');
  }
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  const simulationDatasetId = body.input.simulationDatasetId as number;
+  let missionModelId = body.input.missionModelId as number;
+  let planId = body.input.planId as number;
+
+  // If we have a missionModelId and we need to check a plan permission, get the planId.
+  if (missionModelId !== null && permission in HASURA_PLAN_PERMISSIONS) {
+    planId = await getPlanId(graphqlClient, missionModelId);
+  }
```
Mythicaeda: This will not work, as one mission model is associated with many plans, and this call to getPlanId only returns one of those planIds.

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  // If we have a simulationDatasetId and we need to check a plan permission, get the planId.
+  if (simulationDatasetId !== null && permission in HASURA_PLAN_PERMISSIONS) {
+    planId = await getPlanId(graphqlClient, undefined, simulationDatasetId);
+  }
```
Mythicaeda: Why only do this if `permission in HASURA_PLAN_PERMISSIONS`? The code would be simpler if you always converted `simulationDatasetId`s to `planId`s.

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,314 @@
...
+        mission_model: { id: number }[];
+      }>(
+        gql`
+          query mission_model($simulationDatasetId: Int!) {
+            mission_model(
+              where: { plans: { simulations: { simulation_dataset: { id: { _eq: $simulationDatasetId } } } } }
+            ) {
+              id
+            }
+          }
```
Mythicaeda: This query joins on fewer tables:
```gql
query mission_model($simulationDatasetId: Int!) {
  simulation_dataset_by_pk(id: $simulationDatasetId) {
    simulation {
      plan {
        model_id
      }
    }
  }
}
```

sequencing-server/src/utils/hasura.ts
cohansen: I'm trying to avoid extraneous calls, if an action gets called with a `simulationDatasetId` I only try and fetch the plan id for that simulation dataset if the user is actually checking one of the plan permissions.
`NO_CHECK` won't do another call
`MISSION_MODEL_OWNER` will get the mission model.
The rest will fallback to using the `planId`.

I missed adding `OWNER` to `HASURA_PLAN_PERMISSIONS` though so I'll fix that.

Mythicaeda: Regarding the `MISSION_MODEL_OWNER` case: I have mixed feelings. While I agree that saving a query is a good thing, I think it does so at the detriment to the readability of the function `getMissionModelId`, which is really two disjoint functions stitched together via optional arguments.

sequencing-server/src/utils/hasura.ts
Mythicaeda: The following queries will properly check if a user is a Plan Owner/Collaborator for any plan belonging to a mission model. They return `null` if the mission model id is invalid. Otherwise, they return an object containing an array that's either empty or has a single element.

```gql
query planOwner($id: Int!, $username: String!) {
  mission_model_by_pk(id: $id) {
    plans(where: {owner: {_eq: $username}}, limit: 1) {
      name
    }
  }
}

query planCollaborator($id: Int!, $username: String!) {
  mission_model_by_pk(id: $id) {
    plans(where: {collaborators: {collaborator: {_eq: $username}}}, limit: 1) {
      name
    }
  }
}
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,367 @@
...
+      gql`
+        query planCollaborator($id: Int!, $username: String!) {
+          mission_model_by_pk(id: $id) {
+            plans(where: { collaborators: { collaborator: { _eq: $username } } }, limit: 1) {
+              name
+            }
+          }
+        }
+      `,
+      { planId, username },
```
Mythicaeda:
```suggestion
      { missionModelId, username },
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,367 @@
...
+  planId?: number,
+  missionModelId?: number,
+): Promise<boolean> {
+  if (planId !== null) {
+    const planOwner = await graphqlClient.request<{
+      plan_by_pk: { owner: string };
+    }>(
+      gql`
+      query planOwner($planId: Int!) {
+        plan_by_pk(id $planId) {
```
Mythicaeda:
```suggestion
        plan_by_pk(id: $planId) {
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,367 @@
...
+      gql`
+        query planOwner($id: Int!, $username: String!) {
+          mission_model_by_pk(id: $id) {
+            plans(where: { owner: { _eq: $username } }, limit: 1) {
+              name
+            }
+          }
+        }
+      `,
+      { planId },
```
Mythicaeda:
```suggestion
      { missionModelId, username },
```

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  const permission = permissionCheckQuery.user_role_permission_by_pk.action_permissions as HasuraPermissions | null;
+
+  if (permission === null) {
+    return false;
+  }
+
+  const simulationDatasetId = body.input.simulationDatasetId as number;
+  let missionModelId = body.input.missionModelId as number;
+  let planId = body.input.planId as number;
```
Mythicaeda: When these arguments aren't provided, they're actually `undefined`, not `null`. This is causing issues throughout the file since `undefined !== null` is true

cohansen: Ahh, okay let me fix this.

sequencing-server/src/utils/hasura.ts
```diff
@@ -0,0 +1,367 @@
...
+  if (permission === null) {
+    return false;
+  }
+
+  const simulationDatasetId = body.input.simulationDatasetId as number | undefined;
+  let missionModelId = body.input.missionModelId as number | undefined;
+  let planId = body.input.planId as number | undefined;
+
+  // If we have a simulationDatasetId and we need to check a plan permission, get the planId.
+  if (simulationDatasetId !== undefined && permission in HASURA_PLAN_PERMISSIONS) {
```
Mythicaeda: The correct check is:
```suggestion
  if (simulationDatasetId !== undefined && HASURA_PLAN_PERMISSIONS.includes(permission)) {
```
`permission in HASURA_PLAN_PERMISSIONS` returns `false` even when `permission` is an element of `HASURA_PLAN_PERMISSIONS`

Mythicaeda (approved):
LGTM!

PR #1036, opened 2023-07-19
Support Pass By Position for Sequencing

### LINKED ISSUE #702: Support List Notation in the eDSL
We used to allow positional based commands, go back to supporting that as well as supporting the more verbose object notation.

goetzrrGit:
* **Tickets addressed:** Closes #702
* **Review:** By commit  

## Description
The Sequence eDSL now offers support for passing arguments by position, while retaining backward compatibility with object notation. This allows users to utilize both options based on their preferences. By default, the sequence editor will suggest passing arguments by position. Additionally, when importing a SeqJson, the generated eDSL will be in position notation.

```ts
export default () =>
  Sequence.new({
    seqId: '',
    metadata: {},
    steps: [
      //Regular Args
      C.PREPARE_LOAF({tb_sugar : 10, gluten_free : "FALSE"}),
      C.PREPARE_LOAF(10,"FALSE"),

      //Repeate Args
      C.PACKAGE_BANANA({lot_number : 10, bundle : [{bundle_name : "Dole", number_of_bananas : 10},{bundle_name : "Blue", number_of_bananas : 100}]}),
      C.PACKAGE_BANANA(1394, [['Dole',100],['Blue',10]])
    ]
  });

```

## Verification
Extensive testing has been conducted using both bananantion and Clipper. Real clipper sequences were executed to ensure that there are no regressions with their command dictionary and sequences. Furthermore, the e2e tests have been updated accordingly.

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -1412,7 +1412,10 @@ export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}
                 ? `R\`${durationToHms(this.relativeTime)}\``
                 : 'C';
 
-    const argsString = Object.keys(this.arguments).length === 0 ? '' : `(${argumentsToString(this.arguments)})`;
+    const argsString =
+        Object.keys(this.arguments).length === 0
+            ? ''
+            : `(${argumentsToPositionString(this.arguments)})`;
 
     const metadata =
         this._metadata && Object.keys(this._metadata).length !== 0
@@ -1502,7 +1505,11 @@ export class ImmediateStem<A extends Args[] | { [argName: string]: any } = [] | 
   }
 
   public toEDSLString(): string {
-    const argsString = Object.keys(this.arguments).length === 0 ? '' : `(${argumentsToString(this.arguments)})`;
+    const argsString =
+        Object.keys(this.arguments).length === 0
+            ? ''
+            : `(${argumentsToPositionString(this.arguments)})`;
+
 
     const metadata =
         this._metadata && Object.keys(this._metadata).length !== 0
@@ -1648,11 +1655,11 @@ export class Ground_Block implements GroundBlock {
   }
 
   // @ts-ignore : 'Description' found in JSON Spec
-  public ARGUMENTS(args: Args): Ground_Block {
+  public ARGUMENTS(...args: [Args] | [A, ...A[]]): Ground_Block {
     return Ground_Block.new({
       name: this.name,
       ...(this._models && { models: this._models }),
-      args: args,
+      args: typeof args[0] === 'object' ? args[0] : convertArgsToInterfaces(commandArraysToObj(args, [])),
       ...(this._description && { description: this._description }),
       ...(this._metadata && { metadata: this._metadata }),
       ...(this._absoluteTime && { absoluteTime: this._absoluteTime }),
@@ -1720,7 +1727,7 @@ export class Ground_Block implements GroundBlock {
 
     const args =
         this._args && Object.keys(this._args).length !== 0
-            ? '\n' + indent(`.ARGUMENTS(${argumentsToString(this._args)})`, 1)
+            ? '\n' + indent(`.ARGUMENTS(${argumentsToPositionString(convertInterfacesToArgs(this._args))})`, 1)
             : '';
 
     const metadata =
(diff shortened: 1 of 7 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -1881,11 +1888,12 @@ export class Ground_Event implements GroundEvent {
   }
 
   // @ts-ignore : 'Description' found in JSON Spec
-  public ARGUMENTS(args: Args): Ground_Event {
+  public ARGUMENTS(...args: [Args] | [A, ...A[]]): Ground_Event {
+    debugger;
```
cohansen: Hanging `debugger`.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -3293,6 +3301,50 @@ function commandsWithTimeValue<T extends TimingTypes>(
...
+    }
+    return nestedObjs;
+  }
+
+  for (let i = 0; i < args.length; i++) {
+    const key = keys[i] || `arg_${i}`; // Use `arg_${i}` if key is undefined
+    const value = args[i];
+
+    if (value === undefined) {
+      continue;
```
cohansen: When does this `if` get triggered? I think this function in general could use a few comments to explain what each block is doing.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -3595,4 +3647,61 @@ function argumentsToString<A extends Args[] | { [argName: string]: any } = [] |
...
+    } else {
+      output += `${typeof value === 'string' ? `'${value}'` : value},` ;
+    }
+  }
+  output += '['
+  printValue(args);
+  output += ']'
+
+  // convert [10,[["dole",1483,],]] -> 10,[["dole",1483]]
+  return output.replace(/,(?=\])/g, '').replace(/^\[|\]$/g, '');
```
cohansen: Is there a way we can do this without a regex? If `printValue` and `printArray` knew about the position of the elements we wouldn't have to strip off extraneous commas and square brackets?

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
```diff
@@ -133,38 +133,73 @@ ${doc}
...
+      case '}':
+        return ']';
+      case ';':
+        return ',';
+      default:
+        return match;
+    }
+  });
+
+  type = type.replace(/'([^']+)':/g, '$1:')
```
cohansen, line 248: Can you add comments explaining what this regex and the above 2 are accomplishing?

cohansen (changes requested):
Still need to test this manually, but I added a few initial comments just parsing through the code quickly!

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
camargo: We need to add ESLint to the repo: https://eslint.org/docs/latest/rules/no-debugger

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
Mythicaeda: Can you explain this block? Based on the comment, it looks like `args` is surrounded by a `[]`, which is then stripped by the regex.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: whoops this for loop was going off the keys before I changed it to loop off args. Yep, this isn't needed anymore. I will remove it.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: Removed

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: whoops, I was originally looping through keys instead of args but I changed it towards the end and forgot to remove this. It is gone now

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I reworked the function to exactly what I needed instead of using the regex at the end

sequencing-server/src/lib/codegen/CommandTypeCodegen.ts
goetzrrGit: Sure, I added some js doc to the function and some examples.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
goetzrrGit: I reworked the function and removed the regex.

cohansen (approved):
Looks good, thank you!

PR #1040, opened 2023-07-25
Track simulation extent during simulation

### LINKED ISSUE #771: Show simulation progress
abercrom
https://jira.jpl.nasa.gov/browse/AERIEQS-398
Major
It would be useful for the user to see some indication of simulation progress as a long simulation runs. For Clipper we have use cases for simulations that will run for multiple hours. I know it may be difficult to provide a completely accurate progress, but any indication of what's happening would be useful (e.g. current simulation time, percent complete, etc.)

mattdailis:
* **Tickets addressed:** Closes #771
* **Review:** By commit  

## Description

Inspired by some user observation, I felt compelled to implement simulation extent tracking. This PR adds a new `simulation_extent` table to track the elapsed simulation time, and wires up the simulation driver to report its `elapsedTime` variable to a listener.

The `SimulationDriver::simulate` method has a new overload that takes a `SimulationExtentReporter`. During the simulation, the driver will call the listener's `report` method with the current elapsed time.

One design driver was to limit the overhead of this `report` method - it will be called in the simulation's hot loop, and we don't want reporting to slow down the simulation. For this reason, the `report` method on the `ThreadedSimulationExtentReporter` merely sets an `AtomicReference` and returns immediately:

https://github.com/NASA-AMMOS/aerie/blob/64caed7372d32927c3dfa5689a6a56d243bff31a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java#L30-L32

A separate worker thread polls this value at a configurable cadence and uses the `ResultsCell.WriterRole` to report it to the database.

https://github.com/NASA-AMMOS/aerie/blob/64caed7372d32927c3dfa5689a6a56d243bff31a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java#L59-L81

(Pasting from slack):
Another consideration is that this is a very update-heavy workload, which isn’t postgres’s strong suit. Originally, I wanted to make the extent a new column on the simulation_dataset table - but that means that the entire row will be copied on every update. Since those rows contain the simulation arguments, I was concerned those could spill over page boundaries and hinder postgres’s ability to use Heap-Only-Tuple updates. This drove the decision to make a new table, with a foreign key to the simulation_dataset table

I made all the changes in one commit - if anyone would like them to be split, I will be happy to oblige.

## Verification

All testing has been done manually thus far, using the UI changes in the aerie-ui PR I will open soon.

https://github.com/NASA-AMMOS/aerie/assets/1189602/82291d4a-daa0-42f8-af41-4eb8f767a81f

Additional manual testing encouraged, if reviewers have time.

## Documentation

- [ ] TODO: Consider what documentation, if any, this PR needs

## Future work

- A natural next step is to think about graceful cancellation of simulation runs

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -52,6 +52,7 @@ services:
       MERLIN_WORKER_DB_SERVER: postgres
       MERLIN_WORKER_DB_USER: "${AERIE_USERNAME}"
       MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      SIMULATION_PROGRESS_POLL_PERIOD_MILLIS: 2000
       JAVA_OPTS: >
         -Dorg.slf4j.simpleLogger.defaultLogLevel=INFO
         -Dorg.slf4j.simpleLogger.log.com.zaxxer.hikari=WARN
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulation_extent.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulation_extent.yaml
@@ -1,0 +1,23 @@
+table:
+  name: simulation_extent
+  schema: public
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [simulation_dataset_id, extent]
+      filter: {}
+      allow_aggregations: true
+  - role: user
+    permission:
+      columns: [simulation_dataset_id, extent]
+      filter: {}
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: [simulation_dataset_id, extent]
+      filter: {}
+      allow_aggregations: true
+delete_permissions:
+  - role: aerie_admin
+    permission:
+      filter: {}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -126,6 +126,7 @@ services:
       MERLIN_WORKER_DB_SERVER: postgres
       MERLIN_WORKER_DB_USER: "${AERIE_USERNAME}"
       MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      SIMULATION_PROGRESS_POLL_PERIOD_MILLIS: 2000
       JAVA_OPTS: >
         -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
@@ -150,6 +151,7 @@ services:
       MERLIN_WORKER_DB_SERVER: postgres
       MERLIN_WORKER_DB_USER: "${AERIE_USERNAME}"
       MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      SIMULATION_PROGRESS_POLL_PERIOD_MILLIS: 2000
       JAVA_OPTS: >
         -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
         -Dorg.slf4j.simpleLogger.defaultLogLevel=DEBUG
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -16,6 +16,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 
 public final class SimulationDriver {
   public static <Model>
@@ -26,6 +27,20 @@ public final class SimulationDriver {
       final Duration simulationDuration,
       final Instant planStartTime,
       final Duration planDuration
+  )
+  {
+    return simulate(missionModel, schedule, simulationStartTime, simulationDuration, planStartTime, planDuration, $ -> {});
+  }
+
+  public static <Model>
+  SimulationResults simulate(
+      final MissionModel<Model> missionModel,
+      final Map<ActivityDirectiveId, ActivityDirective> schedule,
+      final Instant simulationStartTime,
+      final Duration simulationDuration,
+      final Instant planStartTime,
+      final Duration planDuration,
+      final Consumer<Duration> simulationExtentConsumer
   ) {
     try (final var engine = new SimulationEngine()) {
       /* The top-level simulation timeline. */
@@ -33,6 +48,8 @@ public final class SimulationDriver {
       var cells = new LiveCells(timeline, missionModel.getInitialCells());
       /* The current real time. */
       var elapsedTime = Duration.ZERO;
+
+      simulationExtentConsumer.accept(elapsedTime);
 
       // Begin tracking all resources.
       for (final var entry : missionModel.getResources().entrySet()) {
@@ -90,6 +107,8 @@ public final class SimulationDriver {
           timeline.add(delta);
           // TODO: Advance a dense time counter so that future tasks are strictly ordered relative to these,
           //   even if they occur at the same real time.
+
+          simulationExtentConsumer.accept(elapsedTime);
 
           if (batch.jobs().isEmpty() && batch.offsetFromStart().isEqualTo(simulationDuration)) {
             break;
--- a/merlin-server/sql/merlin/tables/simulation_extent.sql
+++ b/merlin-server/sql/merlin/tables/simulation_extent.sql
@@ -1,0 +1,20 @@
+create table simulation_extent (
+  simulation_dataset_id integer not null primary key,
+  extent interval not null,
+  constraint simulation_dataset_exists
+    foreign key (simulation_dataset_id)
+    references simulation_dataset
+    on update cascade
+    on delete cascade
+);
+
+comment on table simulation_extent is e''
+  'Tracks the progress of a simulation as the latest achieved offset from the simulation start time. \n'
+  'This is expected to be an update-heavy table, so it is to be kept compact to maximize the likelihood of HOT updates and minimize bloat \n'
+  'The data in this table is not particularly valuable once a simulation has completed, and can be cleared out periodically';
+
+comment on column simulation_extent.simulation_dataset_id is e''
+  'The simulation dataset to which this extent pertains';
+
+comment on column simulation_extent.extent is e''
+  'The latest achieved offset from the simulation start time';
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/InMemoryResultsCellRepository.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/InMemoryResultsCellRepository.java
@@ -129,6 +129,11 @@ public final class InMemoryResultsCellRepository implements ResultsCellRepositor
       this.state = new ResultsProtocol.State.Failed(0, reason);
     }
 
+    @Override
+    public void reportSimulationExtent(final Duration extent) {
+      System.out.println("Simulation extent: " + extent);
+    }
+
     public boolean isEqualTo(final InMemoryCell other) {
       if (this.canceled != other.canceled) return false;
       return Objects.equals(this.state, other.state);
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/DeleteSimulationExtentAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/DeleteSimulationExtentAction.java
@@ -1,0 +1,34 @@
+package gov.nasa.jpl.aerie.merlin.server.remotes.postgres;
+
+import org.intellij.lang.annotations.Language;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+/*package-local*/ final class DeleteSimulationExtentAction implements AutoCloseable {
+  private final @Language("SQL") String sql = """
+        delete from simulation_extent
+        using simulation_dataset
+        where simulation_dataset.id = simulation_extent.simulation_dataset_id
+          and simulation_dataset.dataset_id = ?;
+        """;
+
+  private final PreparedStatement statement;
+
+  public DeleteSimulationExtentAction(final Connection connection) throws SQLException {
+    this.statement = connection.prepareStatement(sql);
+  }
+
+  public void apply(final long datasetId)
+  throws SQLException, NoSuchSimulationDatasetException
+  {
+    this.statement.setLong(1, datasetId);
+    this.statement.executeUpdate();
+  }
+
+  @Override
+  public void close() throws SQLException {
+    this.statement.close();
+  }
+}
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/FixedRateListener.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/FixedRateListener.java
@@ -1,0 +1,46 @@
+package gov.nasa.jpl.aerie.merlin.server.services;
+
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+
+public interface FixedRateListener<T> extends AutoCloseable {
+  void updateValue(T t);
+
+  @Override
+  void close();
+
+  /**
+   * Call reporter.accept at fixed intervals using the given period. The value passed to the consumer is the
+   * most recent argument to `updateValue`.
+   *
+   * @param reporter The Consumer to call periodically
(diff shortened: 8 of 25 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -26,6 +26,24 @@ SimulationResults simulate(
...
       final Instant planStartTime,
       final Duration planDuration
+  )
+  {
+    return simulate(missionModel, schedule, simulationStartTime, simulationDuration, planStartTime, planDuration, $ -> {});
+  }
+
+  public interface SimulationExtentReporter {
+    void report(Duration extent);
+  }
```
Mythicaeda: It's a little odd to have an interface declared here. Could this please be moved above the other `simulate`?

merlin-server/sql/merlin/tables/simulation_extent.sql
```diff
@@ -0,0 +1,20 @@
...
+    foreign key (simulation_dataset_id)
+    references simulation_dataset
+    on update cascade
+    on delete cascade
+);
+
+comment on table simulation_extent is e''
+  'Tracks the progress of a simulation as the latest achieved offset from the simulation start time. \n'
+  'This is expected to be an update-heavy table, so it is to be kept compact to maximize the likelihood of HOT updates and minimize bloat \n'
+  'The data in this table is not particularly valuable once a simulation has completed, and can be cleared out periodically';
```
Mythicaeda, line 14: We briefly discussed this in the thread, but is having the worker remove this row when it finishes simulating a future task?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/InMemoryResultsCellRepository.java
```diff
@@ -129,6 +129,11 @@ public void failWith(final SimulationFailure reason) {
       this.state = new ResultsProtocol.State.Failed(0, reason);
     }
 
+    @Override
+    public void reportSimulationExtent(final Duration extent) {
+
+    }
```
Mythicaeda: I think either throwing an `UnsupportedOperationException` or logging/printing the value of `extent` would make sense here.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -244,14 +252,26 @@ public SimulationResults runSimulation(final CreateSimulationMessage message)
...
-        message.simulationDuration(),
-        message.planStartTime(),
-        message.planDuration());
+    try (final var simulationExtentListener =
+        ThreadedSimulationExtentReporter
+            .spawn(
+                "ThreadedSimulationExtentReporter",
+                writer::reportSimulationExtent,
+                simulationExtentMinimumPollPeriodMillis
+            )) {
```
Mythicaeda: Can you clarify the naming of the variable (`simulationExtentListener`) vs the thread (`ThreadedSimulationExtentReporter`)? I'm specifically confused as to why the worker thread is called "Reporter".

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
```diff
@@ -0,0 +1,83 @@
...
+    try {
+      this.worker.close();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static class Worker implements Runnable {
+    private final AtomicReference<Duration> latestExtent;
+    private final SimulationDriver.SimulationExtentReporter listener;
```
Mythicaeda: Why is this field called `listener`?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
```diff
@@ -0,0 +1,83 @@
...
+        e.printStackTrace(System.err);
+      }
+
+      var lastPollTime = System.nanoTime();
+      while (!Thread.currentThread().isInterrupted()) {
+        try {
+          final var millisSinceLastPoll = (System.nanoTime() - lastPollTime) / 1_000_000;
+          if (millisSinceLastPoll < minimumPollPeriodMillis) {
+            Thread.sleep(minimumPollPeriodMillis - millisSinceLastPoll);
+          }
```
Mythicaeda: Technically, this ought to be a `while` loop to ensure that a spurious wakeup doesn't cause it to run early.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
```diff
@@ -0,0 +1,83 @@
+package gov.nasa.jpl.aerie.merlin.server.services;
+
+import gov.nasa.jpl.aerie.merlin.driver.SimulationDriver;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+public class ThreadedSimulationExtentReporter implements SimulationDriver.SimulationExtentReporter, AutoCloseable {
```
Mythicaeda: Can you add a comment on this class/its subclasses explaining what the class does? It's difficult to keep straight the different parts of the architecture.

deployment/docker-compose.yml
```diff
@@ -52,6 +52,7 @@ services:
       MERLIN_WORKER_DB_SERVER: postgres
       MERLIN_WORKER_DB_USER: "${AERIE_USERNAME}"
       MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      SIMULATION_EXTENT_MINIMUM_POLL_PERIOD_MILLIS: 2000
```
camargo: Can we add this new env var to the environment var document? https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/Environment.md

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     container_name: aerie_gateway
     depends_on: ["postgres"]
     environment:
-      AUTH_TYPE: none
+      AUTH_TYPE: cam
```
camargo: Can we keep the default `none` here? Just to make the default config work if you are not on the JPL network.

mattdailis: Oops, this was for my own testing. Reverting now.

deployment/docker-compose.yml
mattdailis: Absolutely: https://github.com/NASA-AMMOS/aerie/blob/d86e6b084707b1e7ae2e53384c1d5ce5cf0b8658/deployment/Environment.md

docker-compose.yml
```diff
@@ -126,6 +126,7 @@ services:
       MERLIN_WORKER_DB_SERVER: postgres
       MERLIN_WORKER_DB_USER: "${AERIE_USERNAME}"
       MERLIN_WORKER_LOCAL_STORE: /usr/src/app/merlin_file_store
+      SIMULATION_EXTENT_MINIMUM_POLL_PERIOD_MILLIS: 2000
```
Mythicaeda: Maybe it's because the word `EXTENT` has now lost all meaning, but `SIMULATION_PROGRESS_REPORTING_FREQUENCY_MILLIS` better explains what the variable is doing from a user-perspective

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulation_extent.yaml
```diff
@@ -0,0 +1,19 @@
+table:
+  name: simulation_extent
+  schema: public
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [simulation_dataset_id, extent]
+      filter: {}
+      allow_aggregations: true
```
Mythicaeda, line 9: `aerie_admin` needs delete permissions on this table.

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulation_extent.yaml
```diff
@@ -0,0 +1,19 @@
+table:
+  name: simulation_extent
+  schema: public
+select_permissions:
```
Mythicaeda, line 4: Would the reverse object relationship have any value (sim_extent -> sim_dataset)?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: I had originally called it a listener, and then doubted myself and called it a reporter - open to suggestions for a better name :+1: Listener is a more common pattern, so I'll try going back to that

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
mattdailis: Agreed. In fact, I'll go ahead and remove this interface for now - `Consumer<Duration>` should be sufficient for the time being

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
mattdailis: Could you elaborate on what you mean by a spurious wakeup?

docker-compose.yml
mattdailis: The word progress would work well here :+1: I can make the change.

Frequency and period are not synonyms, however - one is measured in Hertz, and the other in seconds. I think period is a more natural metric to specify here

deployment/hasura/metadata/databases/AerieMerlin/tables/public_simulation_extent.yaml
mattdailis: I can't think of a use case for that relationship - let me know if you have one in mind

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
mattdailis: Yes - this class is mirroring the ThreadedSimulationAgent pattern, with a few differences. Its responsibility is to manage a Worker thread that "debounces" calls to its `report` method from calls to its delegate's report method.

This may be clearer if I find some better names for these classes - in particular, this class isn't specific to simulation, maybe making it more general would make it easier to understand.

Its job is to take a `Consumer<T>` and call it at a set rate. It itself is also a `Consumer<T>` that can be called at a different rate. I'll brainstorm a better name - "Debouncer" is the best I've got for now, but it's not 100% accurate, since a debouncer usually waits for inputs to "settle" before firing off a request, while this one fires off requests at a set cadence

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
Mythicaeda: Sure! A [spurious wakeup](https://en.wikipedia.org/wiki/Spurious_wakeup) is when a thread is woken up but its condition is not satisfied. This is a concern that is explicitly called out in [the docs for `Object.wait()`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Object.html#wait(long,int)). With `Thread.sleep`, this can happen if the OS decides to schedule the sleeping thread early.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -244,14 +252,26 @@ public SimulationResults runSimulation(final CreateSimulationMessage message)
...
-        message.simulationStartTime(),
-        message.simulationDuration(),
-        message.planStartTime(),
-        message.planDuration());
+    try (final var simulationExtentListener =
+        ThreadedSimulationExtentReporter
+            .spawn(
+                "ThreadedSimulationExtentReporter",
+                writer::reportSimulationExtent,
+                simulationExtentMinimumPollPeriodMillis
```
Mythicaeda: Why is this specifically called the _minimum_ poll period instead of just the poll period? What would be lost if the word `minimum` is removed? Asking as I am concerned about the length of name and am not sure if that word is meaningfully contributing.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ThreadedSimulationExtentReporter.java
mattdailis: I see, thanks

> But spurious wakeup don't happen for no reason: they usually happen because, in between the time when the condition variable was signaled and when the waiting thread finally ran, another thread ran and changed the condition

Since the condition we're waiting for is "time progressing", it does not appear to me that we are vulnerable to spurious wakeups. There is no way for our condition to be triggered, and then negated by another thread.

https://stackoverflow.com/a/2389016

That said, I'm not opposed to making it a while loop - I'll give it a shot

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
mattdailis: I think we can remove it without adding too much confusion :+1: It is, strictly speaking, a minimum, because it is possible that the time it takes to write to the database could exceed the poll period. In that scenario, it would issue the next request as soon as the previous completes.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/FixedRateListener.java
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Call reporter.accept at fixed intervals using the given period. The value passed to the consumer is the
+   * most recent argument to `updateValue`.
+   *
+   * @param reporter The consumer to call periodically
+   * @param initialValue The initial value to pass to the given consumer
+   * @param pollPeriodMillis The interval at which to call the given consumer
+   * @return a consumer with an additional close method
+   */
+  static <T> FixedRateListener<T> callAtFixedRate(final Consumer<T> reporter, final T initialValue, final long pollPeriodMillis) {
```
Mythicaeda: Very minor suggestion, can be ignored aside from `pollPeriodMillis`, which needs to be renamed and the `@return`, which seems incorrect since the consumer is inaccessible in the `FixedRateListener` object returned.

```suggestion
  /**
   * Call reporter.accept at fixed intervals using the given period. The value passed to reporter.accept is the
   * most recent argument to `updateValue`.
   *
   * @param reporter The Consumer to call periodically
   * @param initialValue The initial value to pass to reporter
   * @param reportPeriodMillis The interval at which to call reporter.accept
   * @return a FixedRateListener with an additional close method
   */
  static <T> FixedIntervalListener<T> callAtFixedInterval(final Consumer<T> reporter, final T initialValue, final long reportPeriodMillis) {
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/DeleteSimulationExtentAction.java
```diff
@@ -0,0 +1,33 @@
...
+  private final PreparedStatement statement;
+
+  public DeleteSimulationExtentAction(final Connection connection) throws SQLException {
+    this.statement = connection.prepareStatement(sql);
+  }
+
+  public void apply(final long datasetId)
+  throws SQLException, NoSuchSimulationDatasetException
+  {
+    this.statement.setLong(1, datasetId);
```
Mythicaeda, line 26: I believe a `this.statement.execute()` is missing here, as this statement isn't being run.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/FixedRateListener.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  void close();
+
+  /**
+   * Call reporter.accept at fixed intervals using the given period. The value passed to the consumer is the
+   * most recent argument to `updateValue`.
+   *
+   * @param reporter The Consumer to call periodically
+   * @param initialValue The initial value to pass to the given Consumer
+   * @param periodMillis The interval at which to call the given reporter.accept
```
Mythicaeda:
```suggestion
   * @param periodMillis The interval at which to call reporter.accept
```

Mythicaeda (approved):
LGTM!

PR #1041, opened 2023-07-25
Activity Directive Changelog

### LINKED ISSUE #1016: Implement change log for activity directives
See designs in https://github.com/NASA-AMMOS/aerie-ui/issues/738

skovati:
* **Tickets addressed:** closes #1016 
* **Review:** By commit  

## Description

As a PR reviewer, I would start looking at the main changes in `merlin-server/sql/merlin/tables/activity_directive_changelog.sql`, which is a new table that stores revisions of `activity_directive`'s. The other ~2000 or so lines of this PR are to make migrations happy, and a handful of basic database trigger tests.

This new changelog table is automatically populated by inserts / updates to the `activity_directive` table. Each revision in uniquely identified by the tuple `(plan_id, activity_directive_id, revision)`. The largest revision number of this table (for a given `(plan_id, activity_directive_id)`) is the current activity_directive state. This means users can revert to the second largest revision for a given `activity_directive` to achieve a traditional "undo".

This new file also adds a handful of `plpgsql` triggers:

- `store_activity_directive_change`, which inserts a new changelog entry whenever `activity_directive` is updated.
- `delete_activity_directive_revisions_past_limit`, which deletes all entries from the changelog older than the latest 10 revisions.

Finally, it exposes a new function `restore_activity_changelog`, exposed by Hasura as `restoreActivityFromChangelog`, which gives Aerie clients a simple interface to revert some activity directive to a certain revision by providing `(plan_id, activity_directive_id, revision)`. 

Additionally, the `activity_directive` table was updated to add a new column, `last_modifed_by`. This new column allows `activity_directive_changlog` entries to track who made which changes, which will be useful for the frontend to display this changelog (see https://github.com/NASA-AMMOS/aerie-ui/issues/738).

Unfortunately, this new column required one line changes to many existing stored functions for the anchor logic and plan merge, and the only way to update a function in postgres is to copy and paste the entire function definition into a `create or replace function` statement, so the database migrations for this PR cause an extra ~1500 lines of changes that aren't actually changes.

### Using this new feature
Get a list of previous revisions for a given activity directive and plan:

```graphql
query revisions {
  activity_directive_changelog(where: {
    plan_id: {_eq: 1}, 
    _and: {activity_directive_id: {_eq: 1}}}, 
    order_by: {revision: desc}) {
    revision
    start_offset
    name
    arguments
    changed_by
    changed_at
    ...
  }
}
```

Restore activity directives to an existing revision in the changelog using the following mutation:
```graphql
mutation restore {
  restoreActivityFromChangelog(args: {
    _activity_directive_id: 1, 
    _plan_id: 1, 
    _revision: 1}) {
    id
    ...
  }
}
```

## Verification

Added a handful of basic tests (check insert on insert, check insert on update, check revert, etc) in `ActivityDirectiveChangelogTests.java`

## Documentation

None, although we can add new documentation for the new exposed SQL function.

## Future work

Maybe add a changelog for more things (e.g. plan)

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
@@ -1,0 +1,290 @@
+package gov.nasa.jpl.aerie.database;
+
+import gov.nasa.jpl.aerie.database.PlanCollaborationTests.Activity;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import java.io.File;
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class ActivityDirectiveChangelogTests {
+  private static final File initSqlScriptFile = new File("../merlin-server/sql/merlin/init.sql");
+  private DatabaseTestHelper helper;
+  private MerlinDatabaseTestHelper merlinHelper;
+  private Connection connection;
+  private int planId;
+
+  @BeforeEach
+  void beforeEach() throws SQLException {
+    int fileId = merlinHelper.insertFileUpload();
+    int missionModelId = merlinHelper.insertMissionModel(fileId);
+    planId = merlinHelper.insertPlan(missionModelId);
+  }
+
+  @AfterEach
+  void afterEach() throws SQLException {
+    helper.clearTable("uploaded_file");
+    helper.clearTable("mission_model");
+    helper.clearTable("plan");
+    helper.clearTable("activity_directive_changelog");
+    helper.clearTable("activity_directive");
+  }
+
+  @BeforeAll
+  void beforeAll() throws SQLException, IOException, InterruptedException {
+    helper = new DatabaseTestHelper(
+        "aerie_merlin_test",
+        "Merlin Database Tests",
+        initSqlScriptFile
+    );
+    helper.startDatabase();
+    connection = helper.connection();
+    merlinHelper = new MerlinDatabaseTestHelper(connection);
+  }
+
+  @AfterAll
+  void afterAll() throws SQLException, IOException, InterruptedException {
+    helper.stopDatabase();
+    connection = null;
+    helper = null;
+  }
+
+  //region Helper Methods
+  private Activity getActivity(final int planId, final int activityId) throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      final var res = statement.executeQuery(
+          // language=sql
+          """
+          SELECT *
+          FROM activity_directive
+          WHERE id = %d
+          AND plan_id = %d;
+          """.formatted(activityId, planId));
+
+      res.next();
+
+      return new Activity(
+          res.getInt("id"),
+          res.getInt("plan_id"),
+          res.getString("name"),
+          res.getInt("source_scheduling_goal_id"),
+          res.getString("created_at"),
+          res.getString("last_modified_at"),
+          res.getString("last_modified_by"),
+          res.getString("start_offset"),
+          res.getString("type"),
+          res.getString("arguments"),
+          res.getString("last_modified_arguments_at"),
+          res.getString("metadata"),
+          res.getString("anchor_id"),
+          res.getBoolean("anchored_to_start")
+      );
+    }
+  }
+
+  private void revertActivityDirectiveToChangelog(int planId, int activityDirectiveId, int revision) throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      statement.executeQuery(
+          // language=sql
+          """
+          SELECT hasura_functions.restore_activity_changelog(
+            _plan_id => %d,
+            _activity_directive_id => %d,
+            _revision => %d,
+            hasura_session => '%s'::json);
+          """.formatted(planId, activityDirectiveId, revision, merlinHelper.admin.session())
+      );
+    }
+  }
+  private int getChangelogRevisionCount(int planId, int activityDirectiveId) throws SQLException {
+    try (final var statement = connection.createStatement()) {
+      final var res = statement.executeQuery(
+          // language=sql
+          """
+          SELECT count(revision)
+          FROM activity_directive_changelog
+          WHERE plan_id = %d and activity_directive_id = %d;
+          """.formatted(planId, activityDirectiveId)
+      );
+      res.next();
+      return res.getInt(1);
+    }
+  }
+  //endregion
+
+  @Test
+  void shouldHaveNoChangelogsForEmptyPlan() throws SQLException {
+    try (final var statement = connection.createStatement()) {
+       final var res = statement.executeQuery(
+          // language=sql
+          """
+          SELECT count(revision)
+          FROM activity_directive_changelog
+          WHERE plan_id = %d;
+          """.formatted(planId)
+      );
+      res.next();
+      final var count = res.getInt(1);
+      assertEquals(0, count);
+    }
+  }
+
+  @Test
+  void shouldCreateChangelogForInsertedActDir() throws SQLException {
+    final var activityId = merlinHelper.insertActivity(planId);
+    assertEquals(1, getChangelogRevisionCount(planId, activityId));
+  }
+
+  @Test
+  void shouldCreateChangelogForUpdatedActDir() throws SQLException {
+    final var activityId = merlinHelper.insertActivity(planId);
+
+    assertEquals(1, getChangelogRevisionCount(planId, activityId));
+
+    try (final var statement = connection.createStatement()) {
+      final var updatedRows = statement.executeQuery(
+          // language=sql
+          """
+          UPDATE activity_directive
+          SET start_offset = '%s'
+          WHERE plan_id = %d and id = %d
+          RETURNING id;
+          """.formatted("01:01:01", planId, activityId));
+      updatedRows.next();
+      assertEquals(activityId, updatedRows.getInt(1));
+    }
+
+    assertEquals(2, getChangelogRevisionCount(planId, activityId));
+}
+    @Test
+    void changelogRevisionHasCorrectValues() throws SQLException {
+      final var activityId = merlinHelper.insertActivity(planId);
+
+      try (final var statement = connection.createStatement()) {
+        final var res = statement.executeQuery(
+            // language=sql
+            """
+            SELECT *
+            FROM activity_directive_changelog
+            WHERE plan_id = %d and activity_directive_id = %d;
+            """.formatted(planId, activityId));
+
+        res.next();
+        final var current = getActivity(planId, activityId);
+
+        assertEquals(current.activityId(), res.getInt("activity_directive_id"));
+        assertEquals(current.planId(), res.getInt("plan_id"));
+        assertEquals(current.name(), res.getString("name"));
+        assertEquals(current.sourceSchedulingGoalId(), res.getInt("source_scheduling_goal_id"));
+        assertEquals(current.lastModifiedAt(), res.getString("changed_at"));
+        assertEquals(current.lastModifiedBy(), res.getString("changed_by"));
+        assertEquals(current.startOffset(), res.getString("start_offset"));
+        assertEquals(current.type(), res.getString("type"));
+        assertEquals(current.arguments(), res.getString("arguments"));
+        assertEquals(current.lastModifiedArgumentsAt(), res.getString("changed_arguments_at"));
+        assertEquals(current.metadata(), res.getString("metadata"));
+        assertEquals(current.anchorId(), res.getString("anchor_id"));
+        assertEquals(current.anchoredToStart(), res.getBoolean("anchored_to_start"));
+    }
+  }
+
+  @Test
+  void shouldDeleteChangelogsOverRevisionLimit() throws SQLException {
+    final var maxRevisionsLimit = 11;
+    final var activityId = merlinHelper.insertActivity(planId);
+
+    // randomly update activity directive > maxRevisionsLimit times
+    for (int i = 0; i < maxRevisionsLimit * 2; i++) {
+      connection.createStatement()
+                .executeQuery(
+                    // language=sql
+                    """
+                    UPDATE activity_directive
+                    SET start_offset = '%02d'
+                    WHERE plan_id = %d and id = %d
+                    RETURNING id;
+                    """.formatted(i, planId, activityId)
+                )
+                .close();
+    }
+
+    assertEquals(maxRevisionsLimit, getChangelogRevisionCount(planId, activityId));
+  }
+
+  @Test
+  void revertNonExistentPlanThrowsError() throws SQLException {
+    try {
+      revertActivityDirectiveToChangelog(-1, -1, -1);
+    } catch (SQLException ex) {
+      if (!ex.getMessage().contains("Plan %d does not exist".formatted(-1))) {
+        throw ex;
+      }
+    }
+  }
+
+  @Test
+  void revertNonExistentDirectiveThrowsError() throws SQLException {
+    try {
+      revertActivityDirectiveToChangelog(planId, -1, -1);
+    } catch (SQLException ex) {
+      if (!ex.getMessage().contains("Activity Directive %d does not exist in Plan %d".formatted(-1, planId))) {
+        throw ex;
+      }
+    }
+  }
+
+  @Test
+  void revertNonExistentRevisionThrowsError() throws SQLException {
+    final var activityId = merlinHelper.insertActivity(planId);
+    try {
+      revertActivityDirectiveToChangelog(planId, activityId, -1);
+    } catch (SQLException ex) {
+      if (!ex.getMessage().contains("Changelog Revision %d does not exist for Plan %d and Activity Directive %d".formatted(-1, planId, activityId))) {
+        throw ex;
+      }
+    }
+  }
+
+  @Test
+  void shouldRevertActDirToChangelogEntry() throws SQLException {
+    final var activityId = merlinHelper.insertActivity(planId);
+
+    final var actDirBefore = getActivity(planId, activityId);
+
+    try (final var statement = connection.createStatement()) {
+      statement.executeQuery(
+          // language=sql
+          """
+          UPDATE activity_directive
+          SET
+            name = 'changed',
+            start_offset = '01:01:01',
+            arguments = '{"biteSize": 2}'
+          WHERE plan_id = %d and id = %d
+          RETURNING *;
+          """.formatted(planId, activityId));
+    }
+    final var actDirMid = getActivity(planId, activityId);
+
+    revertActivityDirectiveToChangelog(planId, activityId, 0);
+    final var actDirAfter = getActivity(planId, activityId);
+
+    assertNotEquals(actDirBefore.name(), actDirMid.name());
+    assertNotEquals(actDirBefore.startOffset(), actDirMid.startOffset());
+    assertNotEquals(actDirBefore.arguments(), actDirMid.arguments());
+
+    assertEquals(actDirBefore.name(), actDirAfter.name());
+    assertEquals(actDirBefore.startOffset(), actDirAfter.startOffset());
+    assertEquals(actDirBefore.arguments(), actDirAfter.arguments());
+  }
+}
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
@@ -120,14 +120,17 @@ final class MerlinDatabaseTestHelper {
   }
 
   int insertActivity(final int planId, final String startOffset, final String arguments) throws SQLException {
+    return insertActivity(planId, startOffset, arguments, admin);
+  }
+  int insertActivity(final int planId, final String startOffset, final String arguments, User user) throws SQLException {
     try (final var statement = connection.createStatement()) {
       final var res = statement
           .executeQuery(
               """
-                  INSERT INTO activity_directive (type, plan_id, start_offset, arguments)
-                  VALUES ('test-activity', '%s', '%s', '%s')
+                  INSERT INTO activity_directive (type, plan_id, start_offset, arguments, last_modified_by)
+                  VALUES ('test-activity', '%s', '%s', '%s', '%s')
                   RETURNING id;"""
-                  .formatted(planId, startOffset, arguments)
+                  .formatted(planId, startOffset, arguments, user.name)
           );
 
       res.next();
--- a/deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
@@ -201,3 +201,12 @@
     - role: aerie_admin
     - role: user
     - role: viewer
+- function:
+    name: restore_activity_changelog
+    schema: hasura_functions
+  configuration:
+    custom_name: restoreActivityFromChangelog
+    session_argument: hasura_session
+  permissions:
+    - role: aerie_admin
+    - role: user
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
@@ -1,0 +1,81 @@
+table:
+  name: activity_directive_changelog
+  schema: public
+object_relationships:
+- name: activity_directive
+  using:
+    manual_configuration:
+      remote_table:
+        name: activity_directive
+        schema: public
+      column_mapping:
+        plan_id: plan_id
+        activity_directive_id: id
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns:
+       - revision
+       - plan_id
+       - activity_directive_id
+       - name
+       - source_scheduling_goal_id
+       - changed_at
+       - changed_by
+       - start_offset
+       - type
+       - arguments
+       - changed_arguments_at
+       - metadata
+       - anchor_id
+       - anchored_to_start
+      filter: {}
+      allow_aggregations: true
+  - role: user
+    permission:
+      columns:
+       - revision
+       - plan_id
+       - activity_directive_id
+       - name
+       - source_scheduling_goal_id
+       - changed_at
+       - changed_by
+       - start_offset
+       - type
+       - arguments
+       - changed_arguments_at
+       - metadata
+       - anchor_id
+       - anchored_to_start
+      filter: {}
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns:
+       - revision
+       - plan_id
+       - activity_directive_id
+       - name
+       - source_scheduling_goal_id
+       - changed_at
+       - changed_by
+       - start_offset
+       - type
+       - arguments
+       - changed_arguments_at
+       - metadata
+       - anchor_id
+       - anchored_to_start
+      filter: {}
+      allow_aggregations: true
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns:
+       - changed_by
+      filter: {}
+delete_permissions:
+  - role: aerie_admin
+    permission:
+      filter: {}
--- a/deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
+++ b/deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
@@ -1,0 +1,1013 @@
+-- remove restore_activity_changelog function permissions key enum value
+update metadata.user_role_permission
+  set function_permissions = function_permissions  - 'restore_activity_changelog';
+
+drop procedure metadata.check_merge_permissions(_function metadata.function_permission_key, _permission metadata.permission, _plan_id_receiving integer, _plan_id_supplying integer, _user text);
+drop procedure metadata.check_merge_permissions(_function metadata.function_permission_key, _merge_request_id integer, hasura_session json);
+drop function metadata.raise_if_plan_merge_permission(_function metadata.function_permission_key, _permission metadata.permission);
+drop procedure metadata.check_general_permissions(_function metadata.function_permission_key, _permission metadata.permission, _plan_id integer, _user text);
+drop function metadata.get_function_permissions(_function metadata.function_permission_key, hasura_session json);
+
+drop type metadata.function_permission_key;
+create type metadata.function_permission_key
+  as enum ('apply_preset', 'branch_plan', 'create_merge_rq', 'withdraw_merge_rq', 'begin_merge', 'cancel_merge',
+    'commit_merge', 'deny_merge', 'get_conflicting_activities', 'get_non_conflicting_activities', 'set_resolution',
+    'set_resolution_bulk', 'delete_activity_subtree', 'delete_activity_subtree_bulk', 'delete_activity_reanchor_plan',
+    'delete_activity_reanchor_plan_bulk', 'delete_activity_reanchor', 'delete_activity_reanchor_bulk', 'get_plan_history');
+
+create function metadata.get_function_permissions(_function metadata.function_permission_key, hasura_session json)
+  returns metadata.permission
+  stable
+  language plpgsql as $$
+  declare
+    _role text;
+    _function_permission metadata.permission;
+  begin
+    _role := metadata.get_role(hasura_session);
+    -- The aerie_admin role is always treated as having NO_CHECK permissions on all functions.
+    if _role = 'aerie_admin' then return 'NO_CHECK'; end if;
+
+    select (function_permissions ->> _function::text)::metadata.permission
+    from metadata.user_role_permission urp
+    where urp.role = _role
+    into _function_permission;
+
+    -- The absence of the function key means that the role does not have permission to perform the function.
+    if _function_permission is null then
+      raise insufficient_privilege
+        using message = 'User with role '''|| _role ||''' is not permitted to run '''|| _function ||'''';
+    end if;
+
+    return _function_permission::metadata.permission;
+  end
+  $$;
+
+create procedure metadata.check_general_permissions(_function metadata.function_permission_key, _permission metadata.permission, _plan_id integer, _user text)
+language plpgsql as $$
+declare
+  _mission_model_id integer;
+  _plan_name text;
+begin
+  select name from public.plan where id = _plan_id into _plan_name;
+
+  -- MISSION_MODEL_OWNER: The user must own the relevant Mission Model
+  if _permission = 'MISSION_MODEL_OWNER' then
+    select id from public.mission_model mm
+    where mm.id = (select model_id from plan p where p.id = _plan_id)
+    into _mission_model_id;
+
+    if not exists(select * from public.mission_model mm where mm.id = _mission_model_id and mm.owner =_user) then
+        raise insufficient_privilege
+        using message = 'Cannot run '''|| _function ||''': '''|| _user ||''' is not MISSION_MODEL_OWNER on Model ' || _mission_model_id ||'.';
+    end if;
+
+  -- OWNER: The user must be the owner of all relevant objects directly used by the KEY
+  -- In most cases, OWNER is equivalent to PLAN_OWNER. Use a custom solution when that is not true.
+  elseif _permission = 'OWNER' then
+		if not exists(select * from public.plan p where p.id = _plan_id and p.owner = _user) then
+      raise insufficient_privilege
+        using message = 'Cannot run '''|| _function ||''': '''|| _user ||''' is not OWNER on Plan ' || _plan_id ||' ('|| _plan_name ||').';
+    end if;
+
+  -- PLAN_OWNER: The user must be the Owner of the relevant Plan
+  elseif _permission = 'PLAN_OWNER' then
+    if not exists(select * from public.plan p where p.id = _plan_id and p.owner = _user) then
+      raise insufficient_privilege
+        using message = 'Cannot run '''|| _function ||''': '''|| _user ||''' is not PLAN_OWNER on Plan '|| _plan_id ||' ('|| _plan_name ||').';
+    end if;
+
+  -- PLAN_COLLABORATOR:	The user must be a Collaborator of the relevant Plan. The Plan Owner is NOT considered a Collaborator of the Plan
+  elseif _permission = 'PLAN_COLLABORATOR' then
+    if not exists(select * from public.plan_collaborators pc where pc.plan_id = _plan_id and pc.collaborator = _user) then
+      raise insufficient_privilege
+        using message = 'Cannot run '''|| _function ||''': '''|| _user ||''' is not PLAN_COLLABORATOR on Plan '|| _plan_id ||' ('|| _plan_name ||').';
+    end if;
+
+  -- PLAN_OWNER_COLLABORATOR:	The user must be either the Owner or a Collaborator of the relevant Plan
+  elseif _permission = 'PLAN_OWNER_COLLABORATOR' then
+    if not exists(select * from public.plan p where p.id = _plan_id and p.owner = _user) then
+      if not exists(select * from public.plan_collaborators pc where pc.plan_id = _plan_id and pc.collaborator = _user) then
+        raise insufficient_privilege
+          using message = 'Cannot run '''|| _function ||''': '''|| _user ||''' is not PLAN_OWNER_COLLABORATOR on Plan '|| _plan_id ||' ('|| _plan_name ||').';
+      end if;
+    end if;
+  end if;
+end
+$$;
+
+create function metadata.raise_if_plan_merge_permission(_function metadata.function_permission_key, _permission metadata.permission)
+returns void
+immutable
+language plpgsql as $$
+begin
+  if _permission::text = any(array['PLAN_OWNER_SOURCE', 'PLAN_COLLABORATOR_SOURCE', 'PLAN_OWNER_COLLABORATOR_SOURCE',
+    'PLAN_OWNER_TARGET', 'PLAN_COLLABORATOR_TARGET', 'PLAN_OWNER_COLLABORATOR_TARGET'])
+  then
+    raise 'Invalid Permission: The Permission ''%'' may not be applied to function ''%''', _permission, _function;
+  end if;
+end
+$$;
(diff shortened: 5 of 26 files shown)


skovati:
Comments after first pass PR review
- [x] `revision` column can be auto-incrementing and unique
- [x] `activity_changelog_directive` should explicitly copy columns from `activity_directive`
- [x] stored procedure needs key and permission check
- [x] stored procedure needs to be in `hasura_functions` schema
- [x] edit plan merge to add `last_modified_by` column

lklyne:
UI approach: https://www.figma.com/file/drLOM5EezMPWRSIuow3UwS/AERIE-Undo-%2F-Redo-or-History-%2F-Changelog-%23738?type=design&node-id=66-23265&mode=design&t=JSHRQkmJt2dBiqbl-4

deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
```diff
@@ -0,0 +1,32 @@
...
+object_relationships:
+- name: activity_directive
+  using:
+    manual_configuration:
+      remote_table:
+        name: activity_directive
+        schema: public
+      column_mapping:
+        plan_id: plan_id
+        activity_directive_id: id
```
Mythicaeda, line 13: Would the reverse relationship (activity_directivy -> changelog) be useful?

merlin-server/sql/merlin/domain-types/permissions.sql
```diff
@@ -11,4 +11,5 @@ create type metadata.function_permission_key
   as enum ('apply_preset', 'branch_plan', 'create_merge_rq', 'withdraw_merge_rq', 'begin_merge', 'cancel_merge',
     'commit_merge', 'deny_merge', 'get_conflicting_activities', 'get_non_conflicting_activities', 'set_resolution',
     'set_resolution_bulk', 'delete_activity_subtree', 'delete_activity_subtree_bulk', 'delete_activity_reanchor_plan',
-    'delete_activity_reanchor_plan_bulk', 'delete_activity_reanchor', 'delete_activity_reanchor_bulk', 'get_plan_history');
+    'delete_activity_reanchor_plan_bulk', 'delete_activity_reanchor', 'delete_activity_reanchor_bulk', 'get_plan_history',
+    'restore_activity_directive_from_changelog');
```
Mythicaeda: This isn't migrated. See [here](https://github.com/NASA-AMMOS/aerie/pull/1088/files#diff-9569c30dc5119df102146af6395790632832920be6dc39cbd6f3e7da87c169cbR199-R202) and [here](https://github.com/NASA-AMMOS/aerie/pull/1088/files#diff-9569c30dc5119df102146af6395790632832920be6dc39cbd6f3e7da87c169cbR504-R509) for how to perform the up and [here](https://github.com/NASA-AMMOS/aerie/pull/1088/files#diff-3d3016014cbc87e44e66b21e0f899d9f3cc35b73b2c673dac18cfc0196269e54R6-R8) and [here](https://github.com/NASA-AMMOS/aerie/pull/1088/files#diff-3d3016014cbc87e44e66b21e0f899d9f3cc35b73b2c673dac18cfc0196269e54R16-R282) for how to do the down

Mythicaeda: This is a really long key and can be shortened to `restore_activity_changelog`

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
Mythicaeda, line 1: `Up` and `down` migrations should always occur in opposite orders (ie, if the up is `Alter Table A -> Add Function -> Add Trigger -> Create Table B`, the down should be `Drop Table B -> Drop Trigger -> Drop Function -> Alter Table A`). This 1) makes the `down` a true inversion of the `up` (and vice versa) and 2) helps avoids errors in the migrations

merlin-server/sql/merlin/functions/public/begin_merge.sql
```diff
@@ -266,18 +266,18 @@ begin
...
+           source_scheduling_goal_id, last_modified_by, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
       from plan_snapshot_activities psa
       join diff_diff dd
         on dd.activity_id = psa.id
       where psa.snapshot_id = snapshot_id_supplying
         and (dd.change_type_receiving = 'modify' and dd.change_type_supplying = 'modify')
     intersect
     select activity_id, name, metadata.tag_ids_activity_directive(dd.activity_id, ad.plan_id) as tags,
-           source_scheduling_goal_id, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
+           source_scheduling_goal_id, last_modified_by, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
```
Mythicaeda: This change is incorrect, as the code is currently checking for effective equality (See comment on L267) and we _do not_ want to include `last_modified_by` as part of that check. A better way may be to leave this column blank during this insert statement and then add an update statement afterwards to fill in last_modified_by. That way the information is still tracked but not included in the diffing process

Alternatively, we don't track `last_modified_by` on the activity_directive table at all and leave it as only accessible via `activity_directive_changelog`/`activity_directive_extended` (which needs to be updated regardless), although that would complicate how the changelog gets updated so :-/.

deployment/hasura/metadata/databases/AerieMerlin/functions/functions.yaml
```diff
@@ -201,3 +201,12 @@
     - role: aerie_admin
     - role: user
     - role: viewer
+- function:
+    name: restore_activity_directive_from_changelog
+    schema: hasura_functions
+  configuration:
+    custom_name: restoreActivityDirectiveFromChangelog
```
Mythicaeda: Name proposal: `restoreActivityFromChangelog` (Although we need to settle on if we're using _s or camelCase for Hasura Functions)

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
```diff
@@ -0,0 +1,101 @@
...
+  changed_at timestamptz not null default now(),
+  changed_by text,
+  start_offset interval not null,
+  type text not null,
+  arguments merlin_argument_set not null,
+  changed_arguments_at timestamptz not null default now(),
+  metadata merlin_activity_directive_metadata_set default '{}'::jsonb,
+
+  anchor_id integer default null,
+  anchored_to_start boolean default true not null,
```
Mythicaeda, line 17: Are any columns here meant to be unique metadata for the changelog (aside from revision)?

merlin-server/sql/merlin/functions/public/begin_merge.sql
Mythicaeda: Thinking about it more, with the addition of the changelog, we may be able to move the `last_modified_at`/`last_modified_args_at` columns off of activity_directive entirely, since that information would already be stored in the changelog and doesn't rely on any user information getting threaded through from Hasura (although I'm not sure why we need to store both, @mattdailis or @camargo, do either of you two recall?)

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
Mythicaeda, line 1: Minor, nonblocking: the indentation after the `create table` statement is weird here (specifically the initial indentation is too high, it should be at the same indentation level that the columns are at in the `create table` statement).

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
```diff
@@ -0,0 +1,101 @@
...
+                          where activity_directive_id = new.activity_directive_id
+                            and plan_id = new.plan_id);
+      return new;
+    end$$;
+
+create trigger delete_activity_directive_revisions_past_limit_trigger
+    after insert
+    on activity_directive_changelog
+    for each row
+    execute function delete_activity_directive_revisions_past_limit();
```
Mythicaeda: Since revision is tracked per activity, this can be simplified into:

```sql
create function delete_activity_directive_revisions_past_limit()
  returns trigger
  language plpgsql as $$
begin
  delete from activity_directive_changelog
  where activity_directive_id = new.activity_directive_id
    and plan_id = new.plan_id
    and revision = (select min(revision)
                    from activity_directive_changelog
                    where activity_directive_id = new.activity_directive_id
                      and plan_id = new.plan_id);
  return new;
end
$$;

create trigger delete_activity_directive_revisions_past_limit_trigger
  after insert on activity_directive_changelog
  for each row
  when ( new.revision > 10 )
  execute function delete_activity_directive_revisions_past_limit();
```

The `<` isn't necessary because you're deleting one row per row added (post threshold).

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
```diff
@@ -0,0 +1,101 @@
...
+          new.anchor_id,
+          new.anchored_to_start
+        );
+
+        return new;
+    end$$;
+
+create trigger store_activity_directive_change_trigger
+    after update or insert
+    on activity_directive
```
Mythicaeda: Minor, style change: 
```suggestion
    after update or insert on activity_directive
```

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
```diff
@@ -0,0 +1,101 @@
...
+          type,
+          arguments,
+          changed_arguments_at,
+          metadata,
+          changed_by,
+          anchor_id,
+          anchored_to_start
+        )
+        values (
+          (select coalesce(max(revision), 0) + 1
```
Mythicaeda: The revision counter should start at `0` to match the semantics used elsewhere in the DB
```suggestion
          (select coalesce(max(revision), -1) + 1
```

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
```diff
@@ -27,3 +27,58 @@ begin
...
 $$;
+
+create function hasura_functions.restore_activity_directive_from_changelog(
+  _plan_id integer,
+  _activity_directive_id integer,
+  _revision integer,
+  hasura_session json)
+  returns setof activity_directive
+  language plpgsql
+  volatile
```
Mythicaeda: The indentation here makes it difficult to read when the function signature ends and when the `returns` etc begins

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
```diff
@@ -27,3 +27,58 @@ begin
 	  order by p.name, ps.start_offset desc;
 end
 $$;
+
+create function hasura_functions.restore_activity_directive_from_changelog(
+  _plan_id integer,
+  _activity_directive_id integer,
+  _revision integer,
+  hasura_session json)
+  returns setof activity_directive
```
Mythicaeda, line 37: Hm, I wonder if this can be changed from `returns setof` to just `returns`.

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
```diff
@@ -27,3 +27,58 @@ begin
...
+      );
+  end if;
+
+  if not exists(select revision
+                from public.activity_directive_changelog
+                where (plan_id, activity_directive_id, revision) =
+                      (_plan_id, _activity_directive_id, _revision))
+  then
+    raise exception 'Changelog Revision % does not exist for Plan % and Activity %', _revision, _plan_id, _activity_directive_id;
+  end if;
```
Mythicaeda: You may want to check that the plan and activity directive exist so this method can throw a more understandable error if either don't.

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
```diff
@@ -27,3 +27,58 @@ begin
...
+  from activity_directive_changelog as c
+  where id = _activity_directive_id
+    and c.plan_id = _plan_id
+    and c.revision = _revision;
+
+  return query select *
+               from activity_directive
+               where id = _activity_directive_id
+                 and plan_id = _plan_id;
+
```
Mythicaeda: This can be simplified into just:
```sql
return query 
update activity_directive
  set name                       = c.name,
      source_scheduling_goal_id  = c.source_scheduling_goal_id,
      start_offset               = c.start_offset,
      type                       = c.type,
      arguments                  = c.arguments,
      last_modified_arguments_at = c.changed_arguments_at,
      metadata                   = c.metadata,
      anchor_id                  = c.anchor_id,
      anchored_to_start          = c.anchored_to_start,
      last_modified_at           = c.changed_at,
      last_modified_by           = c.changed_by
  from activity_directive_changelog as c
  where id = _activity_directive_id
    and c.plan_id = _plan_id
    and c.revision = _revision
  returning *;

```

deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
```diff
@@ -0,0 +1,32 @@
...
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [ plan_id, name, source_scheduling_goal_id, changed_at, changed_by, start_offset, type,
+                 arguments, changed_arguments_at, activity_directive_id, revision, metadata, anchor_id, anchored_to_start]
```
Mythicaeda: Minor, nonblocking: could you reorder the columns here so they're in the same order as on the table? It makes it easier to check that they match

deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
```diff
@@ -0,0 +1,32 @@
...
+  using:
+    manual_configuration:
+      remote_table:
+        name: activity_directive
+        schema: public
+      column_mapping:
+        plan_id: plan_id
+        activity_directive_id: id
+select_permissions:
+  - role: aerie_admin
```
Mythicaeda, line 15: `aerie_admin` needs delete permissions on this table, and should probably be allowed to update the `changed_by` column

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+  @AfterAll
+  void afterAll() throws SQLException, IOException, InterruptedException {
+    helper.stopDatabase();
+    connection = null;
+    helper = null;
+  }
+
+  @Test
+  void shouldHaveNoChangelogsForEmptyPlan() throws SQLException {
+    final var expectedCount = 0;
```
Mythicaeda: Since this variable is only used once, you can just say `assertEquals(0, count);` The name of the test/syntax of `assertEquals` implies that `0` is the "expected value of count". Additionally, by having `0` inlined here, the test is easier to understand, as you don't have to look away from the assert statement to know what we're asserting `count` is equal to.

In general, you don't need to declare these "expected" variables if they're primitives or primative-like (ie Strings, Durations, ints, longs, booleans, etc).

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+        "1970-1-1 00:00:00+00");
+
+    final var res = connection.createStatement()
+                              .executeQuery(
+                                  // language=sql
+                                  """
+                                      SELECT count(revision)
+                                      FROM activity_directive_changelog
+                                      WHERE plan_id = %d;"""
+                                      .formatted(planId)
```
Mythicaeda: (General Feedback for the file) Please use try-with-resources blocks instead.

```suggestion
try(final var statement = connection.createStatement()){
      final var res =  statement.executeQuery(
          """
          SELECT count(revision) FROM activity_directive_changelog
          WHERE plan_id = %d;
          """.formatted(planId)
      );
 }
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+
+  @Test
+  void shouldHaveNoChangelogsForEmptyPlan() throws SQLException {
+    final var expectedCount = 0;
+
+    final var planId = merlinHelper.insertPlan(
+        missionModelId,
+        merlinHelper.admin.name(),
+        "changelog plan" + UUID.randomUUID(),
+        "1970-1-1 00:00:00+00");
```
Mythicaeda: Minor: If the plan's name and start time are not relevant for the test and you were going to use the admin user to insert the plan anyway, use the `insertPlan(missionModelId)` override. Using the full override implies that explicitly setting those values is important for this test (or at least that the start_time is important).

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+
+    final var resBefore = connection.createStatement()
+                                    .executeQuery(
+                                        // language=sql
+                                        """
+                                            SELECT count(revision)
+                                            FROM activity_directive_changelog
+                                            WHERE plan_id = %d and activity_directive_id = %d;"""
+                                            .formatted(planId, activityId)
+                                    );
```
Mythicaeda: This block of code should be pulled out into a function.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+
+    final var updatedRows = connection.createStatement()
+                                      .executeQuery(
+                                          // language=sql
+                                          """
+                                              UPDATE activity_directive
+                                              SET start_offset = '%s'
+                                              WHERE plan_id = %d and id = %d
+                                              RETURNING id;"""
+                                              .formatted(expectedOffsetCurrent, planId, activityId)
```
Mythicaeda: Rather than repeat this code, feel free to pull either `updateOffsetFromAnchor` or `updateActivityName` out into the `MerlinDBHelper` and call it here

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,310 @@
...
+
+    resAfter.next();
+    final var startNew = resAfter.getString("start_offset");
+    final var revNew = resAfter.getInt("revision");
+    resAfter.next();
+    final var startOld = resAfter.getString("start_offset");
+    final var revOld = resAfter.getInt("revision");
+    resAfter.close();
+
+    assertTrue(revOld < revNew);
```
Mythicaeda: We know for sure that `revOld == 1` and `revNew == 2`, right? If so, assert that more specific fact instead of `revOld < revNew`. Otherwise revOld = -1 and revNew = 1000 would pass this check, even though we know that's not what we expect them to be.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
Mythicaeda, line 1: Missing Test: Is the value in the changelog correct? (see `PlanCollaborationTests::snapshotCapturesAllActivities`)

Additionally, `PlanCollaborationTests`'s Activity and Snapshot records need to be updated to reflect the new columns on the tables.

Mythicaeda (changes requested):

skovati:
Hey @Mythicaeda, I've implemented most of your recommended changes (i.e. changes I've replied with 👍) for this PR, which can be seen in [this diff](https://github.com/NASA-AMMOS/aerie/compare/0db25034f9a55c917fe54e3ea3fbd4ccd4a79980..bfb95046145d60a0f4cc8d2bb5399ff70de3ce96). I'm still working on updating the tests with your comments.

merlin-server/sql/merlin/tables/activity_directive_changelog.sql
skovati: I don't think so, unless you have any suggestions.

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
skovati: Looks like this does need to be `returns setof` when using `return query`

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
skovati: Good point, that check was left behind from when `revision` was unique within the whole DB, so only the ordering (not specific values) was known.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
skovati: Re: `PlanCollaborationTests::Activity`, should we export this into a shared module so that `ActivityDirectiveChangelogTests` can use the exact same record?

merlin-server/sql/merlin/functions/public/begin_merge.sql
skovati: I'm a bit confused on how the merge equality is actually being calculated here. I originally added `last_modified_by` here since we needed the same number of columns for unions / intersections. Where should I update the column otherwise?

skovati: Change has been made. We decided to add `last_modified_by` into the MSA by grabbing it from the plan snapshot table, so we apply changes to this column during merge commit, but it doesn't count as a "real change" in the diff.

merlin-server/sql/merlin/functions/public/begin_merge.sql
```diff
@@ -266,26 +266,26 @@ begin
...
+           source_scheduling_goal_id, created_at, last_modified_by, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
       from diff_diff dd
       join activity_directive ad
         on dd.activity_id = ad.id
       where ad.plan_id = plan_id_receiving
         and (dd.change_type_supplying = 'modify' and dd.change_type_receiving = 'modify')
   ) a;
 
-  -- 'modify' against 'delete' and inequal 'modify' against 'modify' goes into conflict table (aka everything left in diff_diff)
+  -- 'modify' against 'delete' and 'modify' against 'modify' goes into conflict table (aka everything left in diff_diff)
```
Mythicaeda: This comment update needs to be removed

merlin-server/sql/merlin/functions/public/begin_merge.sql
```diff
@@ -266,26 +266,26 @@ begin
...
     intersect
     select activity_id, name, metadata.tag_ids_activity_directive(dd.activity_id, ad.plan_id) as tags,
-           source_scheduling_goal_id, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
+           source_scheduling_goal_id, created_at, last_modified_by, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
       from diff_diff dd
       join activity_directive ad
         on dd.activity_id = ad.id
       where ad.plan_id = plan_id_receiving
         and (dd.change_type_supplying = 'modify' and dd.change_type_receiving = 'modify')
   ) a;
```
Mythicaeda: This is incorrect, as it is comparing the `last_modified_by` column. The code at the bottom of this comment is what I was referring to when I said "do the comparison, then join on `activity_directive` to get the `last_modified_by` column". 

Additionally, I'd like a new test in the `PlanCollaborationTests` that has a different user modify the activity on the plan and the branch, so that this bug cannot be accidentally reintroduced. Alternatively, updating `identicalModifyModifyResolvesAsNone` would be sufficient.

(Note that this has been split into its own `INSERT` statement for readability and should replace the entire highlighted block. I'd've put it as a suggestion, but GH seems to think you deleted lines in here somewhere and won't let me.)

```sql
  -- 'modify' against a 'modify' must be checked for equality first.
  with false_modify as (
  	select activity_id, name, metadata.tag_ids_activity_directive(dd.activity_id, psa.snapshot_id) as tags,
           source_scheduling_goal_id, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
      from plan_snapshot_activities psa
      join diff_diff dd
        on dd.activity_id = psa.id
      where psa.snapshot_id = snapshot_id_supplying
        and (dd.change_type_receiving = 'modify' and dd.change_type_supplying = 'modify')
    intersect
    select activity_id, name, metadata.tag_ids_activity_directive(dd.activity_id, ad.plan_id) as tags,
           source_scheduling_goal_id, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
      from diff_diff dd
      join activity_directive ad
        on dd.activity_id = ad.id
      where ad.plan_id = plan_id_receiving
        and (dd.change_type_supplying = 'modify' and dd.change_type_receiving = 'modify'))
  insert into merge_staging_area (
    merge_request_id, activity_id, name, tags, source_scheduling_goal_id, created_at, last_modified_by,
    start_offset, type, arguments, metadata, anchor_id, anchored_to_start, change_type)
  select _merge_request_id, ad.id, ad.name, tags,  ad.source_scheduling_goal_id, ad.created_at,
         ad.last_modified_by, ad.start_offset, ad.type, ad.arguments, ad.metadata, ad.anchor_id, ad.anchored_to_start, 'none'
  from false_modify fm left join activity_directive ad on (ad.plan_id, ad.id) = (plan_id_receiving, fm.activity_id);
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
Mythicaeda: Sure!

merlin-server/sql/merlin/domain-types/permissions.sql
Mythicaeda: Regarding the updated name: please remove the `from` so that it matches with `restore_snapshot` from the User Snapshots PR (The GQL API function name can stay the same)

deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
```diff
@@ -0,0 +1,81 @@
...
+    permission:
+      columns:
```
Mythicaeda, line 17: Minor, nonblocking: This formatting is inconsistent with our other `.yaml`s

merlin-server/sql/merlin/functions/public/begin_merge.sql
Mythicaeda: > Change has been made. We decided to add `last_modified_by` into the MSA by grabbing it from the plan snapshot table

We want to grab `last_modified_by` from the same place we're loading the activity in from, _not_ exclusively from the plan snapshots table. (The code is doing this correctly, just correcting the mental model for if we need to look at this discussion again in a few months)

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
Mythicaeda: The change made does not check if either the plan or activity directive exist. It in fact assumes both exist.

What I was asking for were the following checks:
```sql
if not exists(select id from public.plan where id = _plan_id) then
  raise exception 'Plan % does not exist', _plan_id;
end if;
if not exists(select id from public.activity_directive where (id, plan_id) = (_activity_id, _plan_id)) then
  raise exception 'Activity Directive % does not exist in Plan %', _activity_id, _plan_id;
end if;
```

For future context, I was asking for the input validation here because this function's logic exists entirely inside of it (like in `apply_preset` and the anchor delete functions). If this function had been a wrapper on an internal function (like with `duplicate_plan`), then the input validation only needs to be on the internal function.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,266 @@
...
+          // language=sql
+          """
+          SELECT count(revision)
+          FROM activity_directive_changelog
+          WHERE plan_id = %d and activity_directive_id = %d;
+          """.formatted(planId, activityDirectiveId)
+      );
+      res.next();
+      return res.getInt(1);
+    }
```
Mythicaeda: Two things:

1) Please move helper functions to be between the `Before` and `After` functions and the start of the tests, surrounded by the comment markers `//region Helper Functions` `//endregion`to match the style in the other DB test files
2) Since the `getActivities` function appears to be copied from PlanCollaborationTests, it and the `Activity` record both ought to be pulled out into `MerlinDBHelper`. If doing so would require significant refactors to `PlanCollaborationTests`, then this can be left for a later PR

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
Mythicaeda, line 1: Missing Tests:

- Try to restore a directive on a plan that doesn't exist
- Try to restore a directive that doesn't exist
- Try to restore to a revision that doesn't exist

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,266 @@
...
+      final var res = statement.executeQuery(
+          // language=sql
+          """
+          SELECT count(revision)
+          FROM activity_directive_changelog
+          WHERE plan_id = %d and activity_directive_id = %d;
+          """.formatted(planId, activityId));
+
+      res.next();
+      final var count = res.getInt(1);
```
Mythicaeda: This should be replaced with `getChangelogRevisionCount`

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
```diff
@@ -120,14 +120,17 @@ int insertActivity(final int planId, final String startOffset) throws SQLExcepti
   }
 
   int insertActivity(final int planId, final String startOffset, final String arguments) throws SQLException {
+    return insertActivity(planId, startOffset, arguments, user);
+  }
+  int insertActivity(final int planId, final String startOffset, final String arguments, User modifier) throws SQLException {
```
Mythicaeda: Suggestion, as `modifier` doesn't make sense given that this is an insert statement:
```suggestion
  int insertActivity(final int planId, final String startOffset, final String arguments, User user) throws SQLException {
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/MerlinDatabaseTestHelper.java
```diff
@@ -120,14 +120,17 @@ int insertActivity(final int planId, final String startOffset) throws SQLExcepti
   }
 
   int insertActivity(final int planId, final String startOffset, final String arguments) throws SQLException {
+    return insertActivity(planId, startOffset, arguments, user);
```
Mythicaeda: The default user used in tests is `admin`, not `user`.

merlin-server/sql/merlin/domain-types/permissions.sql
Mythicaeda: Ah, I should've been clear: the Hasura function name was also fine keeping the `from`! _There is no change necessary, this is just for clarification._

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
skovati: 1. I was following the style of `MerlinDatabaseTests.java`, although it looks like that style is somewhat outdated.
2. I figured DB test refactoring could be a separate unit of work, since there is also `Activity` record overlap in `AnchorTests` and `PresetTests`

merlin-server/sql/merlin/functions/public/begin_merge.sql
```diff
@@ -257,33 +257,36 @@ begin
...
     select activity_id, name, metadata.tag_ids_activity_directive(dd.activity_id, psa.snapshot_id) as tags,
            source_scheduling_goal_id, created_at, start_offset, type, arguments, metadata, anchor_id, anchored_to_start
-      from plan_snapshot_activities psa
-      join diff_diff dd
-        on dd.activity_id = psa.id
-      where psa.snapshot_id = snapshot_id_supplying
-        and (dd.change_type_receiving = 'modify' and dd.change_type_supplying = 'modify')
+    from plan_snapshot_activities psa
+           join diff_diff dd
+                on dd.activity_id = psa.id
```
Mythicaeda: Minor: the indentation got messed up here and on L280-281

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
Mythicaeda: 1. Yeah, MerlinDatabaseTests needs a refactor.
2. Fair. Made #1099 to cover the refactor.

deployment/hasura/metadata/databases/AerieMerlin/tables/public_activity_directive_changelog.yaml
skovati: I figured this formatting more closely mirrors the `sql` schema definitions, and is easier to view diffs of.

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
skovati: I figured this was logically equivalent to just checking the `activity_directive_changelog` table, due to foreign key constraints.

Mythicaeda: For Plan, it is not, as proven by the test `shouldHaveNoChangelogsForEmptyPlan`.

For Activity Directive, while the test `shouldCreateChangelogForInsertedActDir` supports the idea that there should be at least one row in the changelog per revision, no rows in the table does not imply that the directive does not exist, as the table could've had rows deleted by something other than the insert trigger.

Additionally, the given error messages do not communicate to the user that the plan or activity directive do not exist.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
```diff
@@ -0,0 +1,291 @@
...
+                .close();
+    }
+
+    assertEquals(maxRevisionsLimit, getChangelogRevisionCount(planId, activityId));
+  }
+
+  @Test
+  void revertNonExistentPlanThrowsError() throws SQLException {
+    try {
+      revertActivityDirectiveToChangelog(-1, -1, -1);
```
Mythicaeda, line 227: Why was the choice made to have all the parameters of the function be invalid rather than just the one under examination (`planId` in this case)?

skovati: To check that the plan exception is thrown first, even if the other params are invalid as well.

merlin-server/sql/merlin/functions/hasura/hasura_functions.sql
skovati: Change has been made. By "logically equivalent" I meant that my original checks would still prevent the user from calling this function with invalid params.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/ActivityDirectiveChangelogTests.java
skovati: This snippet is now only used once

Mythicaeda: By "repeat this code" I meant that there are already functions in the DBTests to update activities, not that this snippet appears multiple times in this file (that was a different conversation that I have already resolved). 

Since this test is that an entry in the changelog is made when the activity is updated, it would be cleaner to use an existing update activity method.

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/up.sql
```diff
@@ -0,0 +1,904 @@
+-- add last_modified_by to activity_directive
+alter table activity_directive
+  add column last_modified_by text,
+
```
Mythicaeda, line 4: Minor: Extra newline

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
```diff
@@ -0,0 +1,1006 @@
...
+    select inserted_snapshot_id, snapshot_id
+    from plan_latest_snapshot where plan_latest_snapshot.plan_id = _plan_id;
+
+  --remove all of those entries from plan_latest_snapshot and add this new snapshot.
+  delete from plan_latest_snapshot where plan_latest_snapshot.plan_id = _plan_id;
+  insert into plan_latest_snapshot(plan_id, snapshot_id) values (_plan_id, inserted_snapshot_id);
+
+  return inserted_snapshot_id;
+  end;
+$$;
```
Mythicaeda: This section of the migration is not inverted compared to the up

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
```diff
@@ -0,0 +1,1006 @@
...
+drop table activity_directive_changelog;
+
+drop trigger store_activity_directive_change_trigger on activity_directive;
+drop function store_activity_directive_change;
+drop function delete_min_activity_directive_revision;
+drop function hasura_functions.restore_activity_changelog(
+  _plan_id integer,
+  _activity_directive_id integer,
+  _revision integer,
+  hasura_session json);
```
Mythicaeda: The correct order for this section is:
```suggestion
drop function hasura_functions.restore_activity_changelog(
  _plan_id integer,
  _activity_directive_id integer,
  _revision integer,
  hasura_session json);
drop trigger delete_min_activity_directive_revision_trigger on activity_directive_changelog;
drop function delete_min_activity_directive_revision;
drop trigger store_activity_directive_change_trigger on activity_directive;
drop function store_activity_directive_change;

comment on table activity_directive_changelog is null;
drop table activity_directive_changelog;
```

Additionally, the trigger `delete_min_activity_directive_revision_trigger` isn't being dropped here (the suggestion includes it).

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/up.sql
```diff
@@ -0,0 +1,904 @@
...
+    and c.activity_directive_id  = _activity_directive_id
+    and ad.plan_id               = _plan_id
+    and c.plan_id                = _plan_id
+    and c.revision               = _revision
+  returning ad.*;
+end
+$$;
+
+-- add new function key for restore_activity_changelog
+alter type metadata.function_permission_key add value 'restore_activity_changelog';
```
Mythicaeda, line 902: This key needs to be added to the `user` role. [Here](https://github.com/NASA-AMMOS/aerie/blob/6f046fb4c10b31f4210ba30a4b44eca0b61ea0ac/deployment/hasura/migrations/AerieMerlin/23_user_plan_snapshots/up.sql#L504-L509) is how to do so.

merlin-server/sql/merlin/domain-types/permissions.sql
Mythicaeda: Unresolving as this comment has not yet been satisfied.

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
Mythicaeda: Unresolving as this comment has not yet been satisfied.

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
```diff
@@ -0,0 +1,1006 @@
...
+drop function metadata.raise_if_plan_merge_permission(_function metadata.function_permission_key, _permission metadata.permission);
+drop procedure metadata.check_general_permissions(_function metadata.function_permission_key, _permission metadata.permission, _plan_id integer, _user text);
+drop function metadata.get_function_permissions(_function metadata.function_permission_key, hasura_session json);
+
+drop type metadata.function_permission_key;
+create type metadata.function_permission_key
+  as enum ('apply_preset', 'branch_plan', 'create_merge_rq', 'withdraw_merge_rq', 'begin_merge', 'cancel_merge',
+    'commit_merge', 'deny_merge', 'get_conflicting_activities', 'get_non_conflicting_activities', 'set_resolution',
+    'set_resolution_bulk', 'delete_activity_subtree', 'delete_activity_subtree_bulk', 'delete_activity_reanchor_plan',
+    'delete_activity_reanchor_plan_bulk', 'delete_activity_reanchor', 'delete_activity_reanchor_bulk', 'get_plan_history');
```
Mythicaeda, line 16: This key isn't being removed from all roles that included it. [Here](https://github.com/NASA-AMMOS/aerie/blob/6f046fb4c10b31f4210ba30a4b44eca0b61ea0ac/deployment/hasura/migrations/AerieMerlin/23_user_plan_snapshots/down.sql#L6-L8) is an example of how to do this;

Mythicaeda (changes requested):

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
skovati: Dropping the table automatically drops all associated triggers as well as comments.

merlin-server/sql/merlin/views/activity_directive_extended.sql
```diff
@@ -77,6 +77,7 @@ create view activity_directive_extended as
     ad.source_scheduling_goal_id as source_scheduling_goal_id,
     ad.created_at as created_at,
     ad.last_modified_at as last_modified_at,
+    ad.last_modified_by as last_modified_by,
```
Mythicaeda, line 80: Metadata needs to be adjusted so that everyone can select the new column

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/up.sql
skovati: This was an intentional delineation between columns and constraints, as is done throughout many Aerie existing `sql` files.

deployment/hasura/migrations/AerieMerlin/23_activity_directive_changelog/down.sql
skovati: Reordered most of the `down.sql` file and it _should_ be good now 👍

Mythicaeda: It's perfect now, thanks!

Mythicaeda (approved):
It's strange not having more feedback after how iterative this review's been, but this looks good to me! Thanks for the great work!

PR #1043, opened 2023-07-26
Make claim actions respect `canceled` status

Mythicaeda:
* **Review:** By commit  

## Description

Currently, neither the Scheduler nor the Simulation Workers would respect that an incoming request had been canceled.

## Verification

Manual testing via:
1. Freezing the worker with a debugger,
2. Setting up the DB with multiple requests, some canceled, others not,
3. Resuming the worker
4. Validating that it received all of the requests but only claimed the ones that weren't canceled.

## Documentation

No documentation updates are needed.


mattdailis:
Good idea! Testing locally now

mattdailis (approved):
Looks good to me

Issue #1045, opened 2023-07-27
Resolve inconsistencies when setting Postgres intervals in DB


camargo:
We have some inconsistencies in how Postgres interval strings are stored in the database.
For example `1 day` vs `24:00:00`. How can we configure the DB to store a single format (i.e. `hh:mm:ss.mmm`)?

There are a couple tasks associated with this ticket:

1. Update [spans action](https://github.com/NASA-AMMOS/aerie/blob/c7b69372fa722854b26159b9a72a37da111d5678/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java#L24) so we just set the interval (no math because the DB gives inconsistent behavior) (easier)
2. How do we enforce a consistent interval format across the DB (harder) 
    - Break this out into another ticket if needed 

See: https://jpl.slack.com/archives/CJ07FGRPH/p1690482612292829

Mythicaeda:
The `addExternalProfile` action inserts the `offset_from_plan_start` column using the `days HH:MM:SS` format

PR #1047, opened 2023-07-31
Avoid timestamptz subtraction for span and event start offsets

mattdailis:
* **Tickets addressed:** #1045 
* **Review:** By commit  

## Description

Postgres stores intervals in tuples of (months, days, microseconds)  ([docs](https://www.postgresql.org/docs/current/datatype-datetime.html))
The first two are primarily needed for calendrical durations. For example, January 1st -> February 1st -> March 1st is 2 months. Noon on Sunday + 1 day is Noon on Monday, even if that crosses a daylight savings boundary (so a day can be 23, 24, or 25 hours, depending on when it is)

On Aerie, to the extent possible, we want to use the microseconds part of that tuple. This is primarily to canonicalize the representation of intervals in Aerie, to avoid the multiple representations `24:00:00.000` vs `24 days`.

This PR updates the `PostSpansAction` (renamed to `InsertSpansAction` to mirror database parlance) to write durations using `PreparedStatements.setDuration` rather than by doing subtraction between two timestamps. The primary purpose of this change is to avoid `start_offsets` that mix the days and microseconds portions of the `interval` type.

> Aside:
> I found [the code](https://github.com/postgres/postgres/blob/341996248e4d720556689e5fb3da7a408cf94228/src/backend/utils/adt/timestamp.c#L2649-L2693) that implements subtraction between two timestamptzs (it happens to be the same code to subtract regular timestamps). At the bottom, it calls interval_justify_hours ([code here](https://github.com/postgres/postgres/blob/341996248e4d720556689e5fb3da7a408cf94228/src/backend/utils/adt/timestamp.c#L2757-L2789)).
> 
> As far as I can tell, it:
> 1. Does the subtraction at the microsecond level, and gets the exact difference in microseconds
> 2. “Justifies” it so that the microsecond portion is less than 24 hours, by incrementing the day portion as much as possible

> Seconds aside:
> Another reason is to avoid calendrical durations, which could cause issues if we were to apply Aerie durations to calendrical time systems. In Aerie, we generally stick to UTC, and do not have a good story for leap seconds, so we can make the claim that days are always 24 hours long. However, it is not out of the question that one might overlay activity times on a more human time scale, e.g. operational meeting times. Using the microsecond representation guarantees that the offsets are correct even in the presence of daylight savings time or leap seconds.

## Verification

I built Aerie on `develop`, ran simulation and saw that it inserted mixed days/microseconds intervals into the `span.start_offset` column. I then switched to this branch, rebuilt, and saw that it inserted only microsecond-based intervals.

## Documentation

We definitely need a document on "Time in Aerie", since this is a topic that comes up often. I'll make a documentation ticket.

https://github.com/NASA-AMMOS/aerie-docs/issues/66

## Future work

- Consider how to avoid the non-microseconds portions of the postgres `interval` type. Perhaps a domain type, or check constraints could help here, or moving to a different representation of offsets altogether.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSimulationEventsAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSimulationEventsAction.java
@@ -15,12 +15,13 @@ import java.util.Map;
 
 import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
+import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatements.setDuration;
 import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatements.setTimestamp;
 
 /*package-local*/ final class InsertSimulationEventsAction implements AutoCloseable {
   @Language("SQL") private static final String sql = """
       insert into event (dataset_id, real_time, transaction_index, causal_time, topic_index, value)
-      values (?, ?::timestamptz - ?::timestamptz, ?, ?, ?, ?::jsonb)
+      values (?, ?::interval, ?, ?, ?, ?::jsonb)
     """;
 
   private final PreparedStatement statement;
@@ -31,8 +32,7 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatemen
 
   public void apply(
       final long datasetId,
-      final Map<Duration, List<EventGraph<Pair<Integer, SerializedValue>>>> eventPoints,
-      final Timestamp simulationStart
+      final Map<Duration, List<EventGraph<Pair<Integer, SerializedValue>>>> eventPoints
   ) throws SQLException {
     for (final var eventPoint : eventPoints.entrySet()) {
       final var time = eventPoint.getKey();
@@ -40,7 +40,7 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PreparedStatemen
       for (int transactionIndex = 0; transactionIndex < transactions.size(); transactionIndex++) {
         final var eventGraph = transactions.get(transactionIndex);
         final var flattenedEventGraph = EventGraphFlattener.flatten(eventGraph);
-        batchInsertEventGraph(datasetId, time, transactionIndex, simulationStart, flattenedEventGraph, this.statement);
+        batchInsertEventGraph(datasetId, time, transactionIndex, flattenedEventGraph, this.statement);
       }
     }
     this.statement.executeBatch();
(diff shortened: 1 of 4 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSimulationEventsAction.java
```diff
@@ -31,16 +32,15 @@ public InsertSimulationEventsAction(final Connection connection) throws SQLExcep
...
+      final Map<Duration, List<EventGraph<Pair<Integer, SerializedValue>>>> eventPoints
   ) throws SQLException {
     for (final var eventPoint : eventPoints.entrySet()) {
       final var time = eventPoint.getKey();
       final var transactions = eventPoint.getValue();
       for (int transactionIndex = 0; transactionIndex < transactions.size(); transactionIndex++) {
         final var eventGraph = transactions.get(transactionIndex);
         final var flattenedEventGraph = EventGraphFlattener.flatten(eventGraph);
-        batchInsertEventGraph(datasetId, time, transactionIndex, simulationStart, flattenedEventGraph, this.statement);
+        batchInsertEventGraph(datasetId, time, transactionIndex, flattenedEventGraph, this.statement);
```
Mythicaeda, line 43: Why was this done via a static method rather than inlined in the `apply` like our other batch inserts?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSimulationEventsAction.java
```diff
@@ -50,21 +50,19 @@ private static void batchInsertEventGraph(
...
       final int transactionIndex,
-      final Timestamp simulationStart,
       final List<Pair<String, Pair<Integer, SerializedValue>>> flattenedEventGraph,
       final PreparedStatement statement
   ) throws SQLException {
-    for (final Pair<String, Pair<Integer, SerializedValue>> entry : flattenedEventGraph) {
+    for (final var entry : flattenedEventGraph) {
       final var causalTime = entry.getLeft();
-      final Pair<Integer, SerializedValue> event = entry.getRight();
+      final var event = entry.getRight();
```
Mythicaeda, line 58: Idle question: why are these getting converted from their specific types to `var`s?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -338,7 +338,7 @@ private static void postActivities(
       final Timestamp simulationStart
   ) throws SQLException {
     try (
-        final var postActivitiesAction = new PostSpansAction(connection);
+        final var insertSpansAction = new InsertSpansAction(connection);
```
Mythicaeda, line 341: Minor: Looking at this with the rename, I wonder if something like `InsertActivitySpansAction` rather than just `InsertSpansAction` (or renaming the function from `postActivities` to `postActivitySpans`) would help with clarity, since Aerie does have the concept of non-Activity Spans

Mythicaeda (commented):
General question with this PR: Are we trying to avoid using days and micros for an Aerie reason? The PR mentions accuracy concerns at the start, but based on how the math is done those accuracy concerns don't seem like they'd appear (since it tries to fill the `days` part of the tuple, not the `months` part)

mattdailis:
The original driver for this PR was to have consistency in time representation. Spans inserted by the scheduler have the form `HH:MM:SS.XXX`, while spans inserted by the simulator have the form `N days HH:MM:SS.XXX`.

If we were to choose one of these representations to standardize on, I would prefer the former. It is true that under controlled circumstances, we shouldn't see issues with precision, but the former form has a different semantic meaning from the latter form, one that conforms better to our notion of duration in Aerie.

> Fun note from the postgres code about the accuracy of this subtraction. They're referring to the case where the two instances fall on different sides of a daylight savings shift - the subtraction will incorrectly assume that the length of a day is 24 hours.
> "This is wrong, but removing it breaks a lot of regression tests."
https://github.com/postgres/postgres/blob/341996248e4d720556689e5fb3da7a408cf94228/src/backend/utils/adt/timestamp.c#L2668

Mythicaeda:
> The original driver for this PR was to have consistency in time representation. Spans inserted by the scheduler have the form HH:MM:SS.XXX, while spans inserted by the simulator have the form N days HH:MM:SS.XXX.
>
> If we were to choose one of these representations to standardize on, I would prefer the former.

Ah, yeah, the consistency argument makes sense to me :+1:. Was that implied by the line `On Aerie, to the extent possible, we want to use the microseconds part of that tuple.`? If not, you ought to mention that in the PR body, as standardizing how we insert into these columns is very compelling.

Mythicaeda:
Thought about this PR: is it possible to encode this conversion from (months, days, microseconds) -> (microseconds) via DB triggers? I'm thinking of [this bug in the UI](https://github.com/NASA-AMMOS/aerie-ui/issues/853), which wouldn't have occurred if the DB enforced the standardization

mattdailis:
Closing to reduce clutter - we can revisit this if the need arises

PR #1051, opened 2023-08-01
Load simulation results from DB before scheduling + performance improvements 

### LINKED ISSUE #748: Load simulation results from DB before scheduling
This ticket builds towards the effort of reducing the number of simulations during scheduling.

Before scheduling, the solver simulates the whole plan to be able to execute expressions. As there is a high chance that the user has already ran a simulation corresponding to the current plan, we could load these simulation results before scheduling and avoid one full simulation.

### LINKED ISSUE #913: Add `iterative` option for scheduling goals
To avoid malformed goals that could cause infinite loops in the scheduler, the scheduler should not iterate on goals by default. Turn that behavior off, and reenable it under a flag in the eDSL.

adrienmaillard:
* **Tickets addressed:** #748 #913 
* **Review:** By commit  

## Description

As usual, on my way to the feature, I get distracted by other improvements and fixes. Some of those are really worth it, in commits (4) and (5) especially. The last commit shows a performance improvement in 20+ tests.  About 30-50% less simulations in these test cases.

In (1), I removed the need for an ugly artificial time buffer in the scheduler simulation by fixing an issue in `Windows`: if the last interval of a profile was closed-open, `Windows` considered that there was a gap because it only used `Interval.contains` and that contains does not care about the open/closed property. The detected gap prevented from turning these windows into spans and so on. Hopefully, that is now squared. 

In (2), I have added a small optimization that may spare us a simulation if not needed.

In (3), a small fix regarding the domain of converted simulation results.

In (4), a significant change. I removed the conflict detection _after_ goal solving. After solving a goal, we ran conflict detection to check that we effectively solved it. This resulted in a simulation runs that could be avoided. The only goal which was taking advantage of this mechanism was the cardinality goal in the case of activities with uncontrollable duration. It was posting one conflict at a time and kept posting conflict until the duration and cardinality conditions were not satisfied. The `MissingActivityTemplateConflict` (and its processing by `PrioritySolver`) has been generalized to handle duration and cardinality thus removing the need for the second conflict detection after goal solving. 

In (5), I fixed a boundary reset condition that were causing more simulations than needed. This bound had been set conservatively because at some point, the driver was dropping some tasks. But this has been fixed a few month ago. 

In (6), I have modified the scheduler so that it is able to use initial simulation results coming from the DB. 
- the solver keeps the initial simulation results and returns them when needed if the associated initial plan has not been modified. 
- a field has been added to the `SimulationFacade` to hold an initial plan until a simulation is needed. This is to avoid simulating a plan that will not be needed. Rootfinding needs the facade to have the current plan "loaded" but if external simulation results have been loaded, there is no need for a first simulation before the first rootfinding try. 

In (7) I pull the simulation results from the DB and pass them to scheduling. DB simulation results are considered useful if they correspond to the current revision of the plan and if they cover the entire plan horizon. We do not pull events and topics to save time (and they are useless to the scheduler right now).  

In (8), I have fixed some tests. The results are the same but the data path as well as the in-memory state was wrong. 

In (9) I have added a test to show how loading initial simulation results allows to remove the first simulation. The test is the same as its neighbor and displays a lower number of simulations (-1) for the same results. 

In (10), I have updated the simulation performance of the scheduler tests. These improvements are mostly due to (4) (and a little bit (5)). 
 
## Verification

Existing tests are passing.  Only one new test. Testing the graphql requests/scheduler-worker is not easy, I am open to suggestions. 

## Documentation

None. 

## Future work

None.

### DIFF
--- a/e2e-tests/src/tests/scheduler-with-sim.test.ts
+++ b/e2e-tests/src/tests/scheduler-with-sim.test.ts
@@ -1,0 +1,319 @@
+
+import {expect, test} from "@playwright/test";
+import req, {awaitScheduling, awaitSimulation} from "../utilities/requests.js";
+import time from "../utilities/time.js";
+
+/*
+This is testing the check on Plan Revision before loading initial simulation results. We inject results associated with an old plan revision.
+In these results, there is only one GrowBanana activity instead of the actual 2 present in the latest plan revision.
+A coexistence goal attaching to GrowBanana activities shows that the scheduler did not use the stale sim results.
+*/
+test.describe.serial('Scheduling with initial sim results', () => {
+  const rd = Math.random() * 100;
+  const plan_start_timestamp = "2021-001T00:00:00.000";
+  const plan_end_timestamp = "2021-001T12:00:00.000";
+
+  test('Main', async ({ request }) => {
+    //upload bananation jar
+    const jar_id = await req.uploadJarFile(request);
+
+    const model: MissionModelInsertInput = {
+      jar_id,
+      mission: 'aerie_e2e_tests' + rd,
+      name: 'Banananation (e2e tests)' + rd,
+      version: '0.0.0' + rd,
+    };
+    const mission_model_id = await req.createMissionModel(request, model);
+    //delay for generation
+    await delay(2000);
+    const plan_input : CreatePlanInput = {
+      model_id : mission_model_id,
+      name : 'test_plan' + rd,
+      start_time : plan_start_timestamp,
+      duration : time.getIntervalFromDoyRange(plan_start_timestamp, plan_end_timestamp)
+    };
+    const plan_id = await req.createPlan(request, plan_input);
+
+    const firstGrowBananaToInsert : ActivityInsertInput =
+        {
+          //no arguments to ensure that the scheduler is getting effective arguments
+          arguments : {},
+          plan_id: plan_id,
+          type : "GrowBanana",
+          start_offset : "1h"
+        };
+
+    await req.insertActivity(request, firstGrowBananaToInsert);
+
+    await awaitSimulation(request, plan_id);
+
+    const secondGrowBananaToInsert : ActivityInsertInput =
+        {
+          //no arguments to ensure that the scheduler is getting effective arguments
+          arguments : {},
+          plan_id: plan_id,
+          type : "GrowBanana",
+          start_offset : "2h"
+        };
+
+    await req.insertActivity(request, secondGrowBananaToInsert);
+
+    const schedulingGoal1 : SchedulingGoalInsertInput =
+        {
+          description: "Test goal",
+          model_id: mission_model_id,
+          name: "ForEachGrowPeel"+rd,
+          definition: `export default function myGoal() {
+                  return Goal.CoexistenceGoal({
+                    forEach: ActivityExpression.ofType(ActivityType.GrowBanana),
+                    activityTemplate: ActivityTemplates.BiteBanana({
+                      biteSize: 1,
+                    }),
+                    startsAt:TimingConstraint.singleton(WindowProperty.END)
+                  })
+                }`
+        };
+
+    const first_goal_id = await req.insertSchedulingGoal(request, schedulingGoal1);
+
+    let plan_revision = await req.getPlanRevision(request, plan_id);
+
+    const schedulingSpecification : SchedulingSpecInsertInput = {
+      // @ts-ignore
+      horizon_end: plan_end_timestamp,
+      horizon_start: plan_start_timestamp,
+      plan_id : plan_id,
+      plan_revision : plan_revision,
+      simulation_arguments : {},
+      analysis_only: false
+    }
+    const scheduling_specification_id = await req.insertSchedulingSpecification(request, schedulingSpecification);
+
+    const priority = 0;
+    const specGoal: SchedulingSpecGoalInsertInput = {
+      goal_id: first_goal_id,
+      priority: priority,
+      specification_id: scheduling_specification_id,
+    };
+    await req.createSchedulingSpecGoal(request, specGoal);
+
+    const { status, datasetId } = await awaitScheduling(request, scheduling_specification_id);
+
+    expect(status).toEqual("complete")
+    expect(datasetId).not.toBeNull();
+
+    const plan = await req.getPlan(request, plan_id)
+    expect(plan.activity_directives.length).toEqual(4);
+
+    //delete plan
+    const deleted_plan_id = await req.deletePlan(request, plan_id);
+    expect(deleted_plan_id).not.toBeNull();
+    expect(deleted_plan_id).toBeDefined();
+    expect(deleted_plan_id).toEqual(plan_id);
+
+    //delete mission model
+    const deleted_mission_model_id = await req.deleteMissionModel(request, mission_model_id)
+    expect(deleted_mission_model_id).not.toBeNull();
+    expect(deleted_mission_model_id).toBeDefined();
+    expect(deleted_mission_model_id).toEqual(mission_model_id);
+  });
+
+
+  /* In this test, we load simulation results with the current plan revision but with a different sim config. If the
+   injected results are picked up, the goal will not be satisfied and the number of activities will stay at its original value.
+   */
+  test('Scheduling sim results 2', async ({ request }) => {
+    const rd = Math.random() * 100;
+    const plan_start_timestamp = "2021-001T00:00:00.000";
+    const plan_end_timestamp = "2021-001T12:00:00.000";
+
+    //upload bananation jar
+    const jar_id = await req.uploadJarFile(request);
+
+    const model: MissionModelInsertInput = {
+      jar_id,
+      mission: 'aerie_e2e_tests' + rd,
+      name: 'Banananation (e2e tests)' + rd,
+      version: '0.0.0' + rd,
+    };
+    const mission_model_id = await req.createMissionModel(request, model);
+    //delay for generation
+    await delay(2000);
+    const plan_input : CreatePlanInput = {
+      model_id : mission_model_id,
+      name : 'test_plan' + rd,
+      start_time : plan_start_timestamp,
+      duration : time.getIntervalFromDoyRange(plan_start_timestamp, plan_end_timestamp)
+    };
+    const plan_id = await req.createPlan(request, plan_input);
+
+    const simulation_id = await req.getSimulationId(request, plan_id);
+
+    const simulation_template : InsertSimulationTemplateInput = {
+      model_id: mission_model_id,
+      arguments: {
+        "initialPlantCount": 400,
+      },
+      description: 'Template for Plan ' +plan_id
+    };
+
+    await req.insertAndAssociateSimulationTemplate(request, simulation_template, simulation_id);
+
+    await awaitSimulation(request, plan_id);
+
+    const empty_simulation_template : InsertSimulationTemplateInput = {
+      model_id: mission_model_id,
+      arguments: {
+      },
+      description: 'Template for Plan ' +plan_id
+    };
+
+    await req.insertAndAssociateSimulationTemplate(request, empty_simulation_template, simulation_id);
+
+    const schedulingGoal1 : SchedulingGoalInsertInput =
+        {
+          description: "Test goal",
+          model_id: mission_model_id,
+          name: "ForEachPlanLessThan300"+rd,
+          definition: `export default () => Goal.CoexistenceGoal({
+            forEach: Real.Resource("/plant").lessThan(300),
+            activityTemplate: ActivityTemplates.GrowBanana({quantity: 10, growingDuration: Temporal.Duration.from({minutes:1}) }),
+            startsAt: TimingConstraint.singleton(WindowProperty.START)
+          })`
+        };
+
+    const first_goal_id = await req.insertSchedulingGoal(request, schedulingGoal1);
+
+    const plan_revision = await req.getPlanRevision(request, plan_id);
+
+    const schedulingSpecification : SchedulingSpecInsertInput = {
+      // @ts-ignore
+      horizon_end: plan_end_timestamp,
+      horizon_start: plan_start_timestamp,
+      plan_id : plan_id,
+      plan_revision : plan_revision,
+      simulation_arguments : {},
+      analysis_only: false
+    }
+    const specification_id = await req.insertSchedulingSpecification(request, schedulingSpecification);
+
+    const priority = 0;
+    const specGoal: SchedulingSpecGoalInsertInput = {
+      goal_id: first_goal_id,
+      priority: priority,
+      specification_id: specification_id,
+    };
+    await req.createSchedulingSpecGoal(request, specGoal);
+
+    await awaitScheduling(request, specification_id);
+
+    const plan = await req.getPlan(request, plan_id)
+    expect(plan.activity_directives.length).toEqual(1);
+
+    //delete plan
+    const deleted_plan_id = await req.deletePlan(request, plan_id);
+    expect(deleted_plan_id).not.toBeNull();
+    expect(deleted_plan_id).toBeDefined();
+    expect(deleted_plan_id).toEqual(plan_id);
+
+    //delete mission model
+    const deleted_mission_model_id = await req.deleteMissionModel(request, mission_model_id)
+    expect(deleted_mission_model_id).not.toBeNull();
+    expect(deleted_mission_model_id).toBeDefined();
+    expect(deleted_mission_model_id).toEqual(mission_model_id);
+  });
+
+  /* In this test, we inject simulation results to test that the scheduler is loading them properly. If they are picked up,
+  there should be no activity in the plan (plant>300). Otherwise, there should be one activity.
+   */
+  test('Scheduling sim results 3', async ({ request }) => {
+    const rd = Math.random() * 100;
+    const plan_start_timestamp = "2021-001T00:00:00.000";
+    const plan_end_timestamp = "2021-001T12:00:00.000";
+    //upload bananation jar
+    const jar_id = await req.uploadJarFile(request);
+
+    const model: MissionModelInsertInput = {
+      jar_id,
+      mission: 'aerie_e2e_tests' + rd,
+      name: 'Banananation (e2e tests)' + rd,
+      version: '0.0.0' + rd,
+    };
+    const mission_model_id = await req.createMissionModel(request, model);
+    //delay for generation
+    await delay(2000);
+    const plan_input : CreatePlanInput = {
+      model_id : mission_model_id,
+      name : 'test_plan' + rd,
+      start_time : plan_start_timestamp,
+      duration : time.getIntervalFromDoyRange(plan_start_timestamp, plan_end_timestamp)
+    };
+    const plan_id = await req.createPlan(request, plan_input);
+
+    const simId = await req.getSimulationId(request, plan_id);
+    let plan_revision = await req.getPlanRevision(request, plan_id);
+
+    const datasetId = await req.insertSimulationDataset(
+        request,
+        simId,
+        plan_start_timestamp,
+        plan_end_timestamp,
+        "success",
+        {},
+        plan_revision);
+
+    const profileId = await req.insertProfile(request, datasetId, "12h", "/plant", { "type": "discrete", "schema": { "type": "int" } });
+    await req.insertProfileSegment(request, datasetId, 400, false, profileId, "0h");
+
+    const schedulingGoal1 : SchedulingGoalInsertInput =
+        {
+          description: "Test goal",
+          model_id: mission_model_id,
+          name: "ForEachPlanLessThan300"+rd,
+          definition: `export default () => Goal.CoexistenceGoal({
+            forEach: Real.Resource("/plant").lessThan(300),
+            activityTemplate: ActivityTemplates.GrowBanana({quantity: 10, growingDuration: Temporal.Duration.from({minutes:1}) }),
+            startsAt: TimingConstraint.singleton(WindowProperty.START)
+          })`
+        };
+
+    const first_goal_id = await req.insertSchedulingGoal(request, schedulingGoal1);
+
+    plan_revision = await req.getPlanRevision(request, plan_id);
+
+    const schedulingSpecification : SchedulingSpecInsertInput = {
+      // @ts-ignore
+      horizon_end: plan_end_timestamp,
+      horizon_start: plan_start_timestamp,
+      plan_id : plan_id,
+      plan_revision : plan_revision,
+      simulation_arguments : {},
+      analysis_only: false
+    }
+    const specification_id = await req.insertSchedulingSpecification(request, schedulingSpecification);
+
+    const priority = 0;
+    const specGoal: SchedulingSpecGoalInsertInput = {
+      goal_id: first_goal_id,
+      priority: priority,
+      specification_id: specification_id,
+    };
+    await req.createSchedulingSpecGoal(request, specGoal);
+
+    await awaitScheduling(request, specification_id);
+
+    const plan = await req.getPlan(request, plan_id)
+    expect(plan.activity_directives.length).toEqual(0);
+
+    //delete plan
+    await req.deletePlan(request, plan_id);
+
+    //delete mission model
+    await req.deleteMissionModel(request, mission_model_id)
+  });
+});
+
+
+function delay(ms: number) {
+  return new Promise( resolve => setTimeout(resolve, ms) );
+}
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityTemplateConflict.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityTemplateConflict.java
@@ -2,12 +2,14 @@ package gov.nasa.jpl.aerie.scheduler.conflicts;
 
 import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
 import gov.nasa.jpl.aerie.constraints.time.Windows;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.scheduler.constraints.activities.ActivityCreationTemplate;
 import gov.nasa.jpl.aerie.scheduler.goals.ActivityTemplateGoal;
 
+import java.util.Optional;
+
 /**
  * describes plan problem due to lack of a matching instance for a template
- *
  * such conflicts are typically addressed by scheduling additional activities
  * using the corresponding creation template
  */
@@ -16,16 +18,20 @@ public class MissingActivityTemplateConflict extends MissingActivityConflict {
   /**
    * ctor creates a new conflict regarding a missing activity
    *
-   * @param goal IN STORED the dissatisfied goal that issued the conflict
-   * @param temporalContext IN STORED the times in the plan when the goal was
-   *     disatisfied enough to induce this conflict (including just the
-   *     desired start times of the activity, not necessarily the end time)
+   * @param goal  the dissatisfied goal that issued the conflict
+   * @param temporalContext  the times in the plan when the goal was
+   * @param template  desired activity template
+   * @param evaluationEnvironment the evaluation environment at the time of creation so variables can be retrieved later at instantiation
+   * @param cardinality the desired number of times the activity template should be inserted
+   * @param totalDuration the desired total duration
    */
   public MissingActivityTemplateConflict(
       ActivityTemplateGoal goal,
       Windows temporalContext,
       ActivityCreationTemplate template,
-      EvaluationEnvironment evaluationEnvironment)
+      EvaluationEnvironment evaluationEnvironment,
+      int cardinality,
+      Optional<Duration> totalDuration)
   {
     super(goal, evaluationEnvironment);
 
@@ -35,6 +41,22 @@ public class MissingActivityTemplateConflict extends MissingActivityConflict {
     }
     this.temporalContext = temporalContext;
     this.template = template;
+    this.cardinality = cardinality;
+    this.totalDuration = totalDuration;
+  }
+
+  //the number of times the activity needs to be inserted
+  int cardinality;
+
+  //the desired total duration over the number of activities needed
+  Optional<Duration> totalDuration;
+
+  public int getCardinality(){
+    return cardinality;
+  }
+
+  public Optional<Duration> getTotalDuration(){
+    return totalDuration;
   }
 
   /**
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
@@ -2,6 +2,7 @@ package gov.nasa.jpl.aerie.scheduler.constraints.activities;
 
 import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
 import gov.nasa.jpl.aerie.constraints.model.Profile;
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
 import gov.nasa.jpl.aerie.constraints.time.Interval;
 import gov.nasa.jpl.aerie.constraints.time.Spans;
 import gov.nasa.jpl.aerie.constraints.time.Windows;
@@ -271,6 +272,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
 
         @Override
         public Duration valueAt(Duration start, HistoryWithActivity history) {
+          final var latestConstraintsSimulationResults = getLatestSimulationResults(facade, start);
           final var actToSim = SchedulingActivityDirective.of(
               type,
               start,
@@ -278,7 +280,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
               SchedulingActivityDirective.instantiateArguments(
                   arguments,
                   start,
-                  facade.getLatestConstraintSimulationResults(),
+                  latestConstraintsSimulationResults,
                   evaluationEnvironment,
                   type),
               null,
@@ -313,7 +315,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
       final var instantiatedArguments = SchedulingActivityDirective.instantiateArguments(
           this.arguments,
           earliestStart,
-          facade.getLatestConstraintSimulationResults(),
+          getLatestSimulationResults(facade, earliestStart),
           evaluationEnvironment,
           type);
 
@@ -343,7 +345,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
           SchedulingActivityDirective.instantiateArguments(
               this.arguments,
               earliestStart,
-              facade.getLatestConstraintSimulationResults(),
+              getLatestSimulationResults(facade, earliestStart),
               evaluationEnvironment,
               type),
           null,
@@ -365,7 +367,7 @@ public class ActivityCreationTemplate extends ActivityExpression implements Expr
           SchedulingActivityDirective.instantiateArguments(
               this.arguments,
               earliestStart,
-              facade.getLatestConstraintSimulationResults(),
+              getLatestSimulationResults(facade, earliestStart),
               evaluationEnvironment,
               type),
           null,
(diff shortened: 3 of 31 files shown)


adrienmaillard:
Here are some actions I am going to take following the walkthrough this morning: 
- check that the documentation mentions that the sim configuration is loaded by the scheduler 
- when looking for a suitable simulation dataset, add a condition checking that the current sim config is equal to the sim config used for the potential dataset
- break down the big graphql request into similar smaller requests (and maybe multi-thread them)
- add an e2e test with stale results

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
```diff
@@ -165,7 +165,7 @@ public void simulateActivities(@NotNull Map<ActivityDirectiveId, ActivityDirecti
     resolved.get(null).sort(Comparator.comparing(Pair::getRight));
     final var earliestStartOffset = resolved.get(null).get(0);
 
-    if(earliestStartOffset.getRight().noLongerThan(curTime)){
+    if(earliestStartOffset.getRight().shorterThan(curTime)){
```
mattdailis: I'm not sure I have a solid enough grasp of the semantics here to feel confident that this is a correct change. Could you help me understand better?

This seems like it would be a safe change if the simulation has run everything before, but not including the `curTime` - in that case adding a new activity at `curTime` will execute in parallel with everything else that was scheduled at that time.

However, if the simulation has executed anything at `curTime`, continuing the simulation without re-initializing would cause the new activity to execute causally after the other events at `curTime`, rather than concurrently.

So my question is: what does `curTime` mean with respect to the simulation engine? Have we finished executing all jobs before and including the jobs at `curTime`? Or excluding the jobs at `curTime`? I'll try to figure this out myself, but if anyone knows off-hand I'd love to hear from you.

Mythicaeda: TL;DR: Yes, simulation has always run everything up to but not including `curTime`

`curTime` is the time of the batch that the engine will resume at, which is always the first batch at a time in the future of when the engine last stopped. In `simulateSchedule`, this is done by continuing the while loop if the tasks we're interested in are finished but there are still batches at the same time (`while (!allTaskFinished || delta.isZero())`). In `simulateUntil`, this is natural shakeout of the loop condition `!batch.offsetFromStart().longerThan(endTime)`, since all batches at a time will all obviously pass that check.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityTemplateConflict.java
```diff
@@ -16,16 +18,20 @@ public class MissingActivityTemplateConflict extends MissingActivityConflict {
   /**
    * ctor creates a new conflict regarding a missing activity
```
Mythicaeda, line 19: What does `ctor` mean?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
```diff
@@ -194,17 +195,15 @@ else if (this.initiallyEvaluatedTemporalContext == null) {
...
-        conflicts.add(new MissingActivityTemplateConflict(this, subIntervalWindows, this.desiredActTemplate, new EvaluationEnvironment()));
+
+      if(nbToSchedule>0 || durToSchedule.isPositive()) {
+        conflicts.add(new MissingActivityTemplateConflict(
+            this,
+            subIntervalWindows,
+            this.desiredActTemplate,
+            new EvaluationEnvironment(),
+            nbToSchedule,
+            durToSchedule.isPositive() ? Optional.of(durToSchedule) : Optional.empty()));
```
Mythicaeda, line 206: Does this mean that you can now have a cardinality goal that is concerned with both the number of occurrences _and_ the total duration of all occurrences?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -523,59 +525,88 @@ private void satisfyGoalGeneral(Goal goal) {
...
+    while (itConflicts.hasNext()) {
+      final var missing = itConflicts.next();
+      assert missing != null;
 
-    while (!missingConflicts.isEmpty() && madeProgress) {
-      madeProgress = false;
+      //determine the best activities to satisfy the conflict
+      if (!analysisOnly && (missing instanceof MissingActivityInstanceConflict missingActivityInstanceConflict)) {
+        final var acts = getBestNewActivities(missingActivityInstanceConflict);
+        assert acts != null;
```
Mythicaeda: This assertion will never fail. What is it meant to be checking? (This also applies to L563)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -523,59 +525,88 @@ private void satisfyGoalGeneral(Goal goal) {
...
-        assert missing != null;
+            evaluation.forGoal(goal).associate(insertionResult.activitiesInserted(), true);
+            itConflicts.remove();
+            //REVIEW: really association should be via the goal's own query...
 
-        //determine the best activities to satisfy the conflict
-        if (!analysisOnly && (missing instanceof MissingActivityInstanceConflict || missing instanceof MissingActivityTemplateConflict)) {
-          final var acts = getBestNewActivities((MissingActivityConflict) missing);
+            //NB: repropagation of new activity effects occurs on demand
+            //    at next constraint query, if relevant
```
Mythicaeda: What does this comment mean? (This also applies to L572-573)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
```diff
@@ -482,5 +484,18 @@ private  Optional<SchedulingActivityDirective> rootFindingHelper(
...
   }
 
+  private SimulationResults getLatestSimulationResults(final SimulationFacade facade, final Duration until){
+    final var latestConstraintsSimulationResults = facade.getLatestConstraintSimulationResults();
+    if(latestConstraintsSimulationResults.isEmpty()){
+      try {
+        facade.computeSimulationResultsUntil(until);
+        return facade.getLatestConstraintSimulationResults().get();
+      } catch(SimulationFacade.SimulationException e){
+        throw new RuntimeException("Simulation is in irreparable state");
```
Mythicaeda, line 494: Why would a `SimulationException` imply that simulation is in an `irreparable` state specifically? If there was an error in the plan, wouldn't the initial sim throw this exception but be reparable?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/Problem.java
```diff
@@ -123,10 +128,26 @@ public Plan getInitialPlan() {
    *
    * @param plan the initial seed plan that schedulers may start from
```
Mythicaeda, line 129: `@param initialSimulationResults` needs to be added to the javadocs here

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -143,11 +157,15 @@ public void removeAndInsertActivitiesFromSimulation(
         insertedActivities.remove(act);
       }
     }
+    if(!initialPlan.isEmpty()) activitiesToAdd.addAll(this.initialPlan);
+    this.initialPlan.clear();
```
Mythicaeda: Could you clarify why we need to add the initial plan to `activitiesToAdd` and clear `initialPlan` if there are activities in the plan at this point?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -49,6 +50,11 @@ public class PrioritySolver implements Solver {
   boolean checkSimBeforeInsertingActivities;
   boolean checkSimBeforeEvaluatingGoal;
 
+  /**
+   * state boolean stating whether the initial plan has been modified to allow initial simulation results to be used
+   */
+  boolean initialPlanHasBeenModified = false;
```
Mythicaeda: Can this boolean be private?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsers.java
```diff
@@ -2,32 +2,33 @@
...
 import static gov.nasa.jpl.aerie.json.BasicParsers.mapP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.productP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
 import static gov.nasa.jpl.aerie.json.Uncurry.tuple;
 import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
 import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
 import static gov.nasa.jpl.aerie.merlin.driver.json.ValueSchemaJsonParser.valueSchemaP;
-
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECOND;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
```
Mythicaeda: Minor: Could you standardize this file onto either `Duration.MICROSECONDS` or `Duration.MICROSECOND`?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsers.java
```diff
@@ -130,4 +92,50 @@ public static Duration parseGraphQLInterval(final String in) {
...
+                                               .plusSeconds(asInterval.getWholeSeconds())
+                                               .plusNanos(asInterval.getMicroSeconds()*1000);
+      return Duration.of(asDuration.toNanos()/1000, MICROSECONDS);
+    }catch(SQLException e){
+      throw new RuntimeException(e);
+    }
+  }
+  public static String stringPGIntervalFromDuration(final Duration duration) {
+      return graphQLIntervalFromDuration(duration).getValue();
+  }
```
Mythicaeda: Given that this method is only used in one place, could it be inlined there?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
mattdailis: Thanks - that helps point me in the right direction. Reading your explanation, and looking at the code, I'm not sure I reach the same conclusion. The code updates `curTime` only when the batch is about to be executed - if the `while` loop deems that the next batch is too far in the future, it leaves the value of `curTime` as the time of the last executed batch.

https://github.com/NASA-AMMOS/aerie/blob/c22b56cd74cf1a53917a071f0824978d6409103f/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java#L123-L125

https://github.com/NASA-AMMOS/aerie/blob/c22b56cd74cf1a53917a071f0824978d6409103f/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java#L247-L248

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -766,6 +770,211 @@ public DatasetId storeSimulationResults(final PlanMetadata planMetadata,
...
+              }
+              }
+
+            }
+          }
+        }""".formatted(planMetadata.planId().id(), planMetadata.planRev(), simulationDatasetId.get().id());
+
+    final JsonObject response;
+    response = postRequest(request).get();
+    final var data = response.getJsonObject("data");
```
Mythicaeda: Reminder to split this into several requests (optionally made and parsed in parallel) to significantly improve performance.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -766,6 +770,211 @@ public DatasetId storeSimulationResults(final PlanMetadata planMetadata,
...
+          simulation(where: {plan_id: {_eq: %d}}) {
+            id
+            simulation_datasets(where: {plan_revision: {_eq: %d}, canceled: {_eq: false}}) {
+              id
+              simulation_start_time
+              simulation_end_time
+            }
+          }
+        }""".formatted(planMetadata.planId().id(),
+                       planMetadata.planRev());
```
Mythicaeda: Proposal: The following query more directly accesses the desired simulation dataset. It removes the unused `id` from `simulation` that's fetched, replacing it with `arguments`, which can be used to ensure that the args that the sim dataset used matches the one that the scheduler will use (although if that information's already accessible in this method, that part of the query can be removed). It additionally moves the sim bounds check into the `where` clause of the query and only queries successful sim runs to avoid fetching unnecessary rows.

```java
    final var request =
      """
      {
        simulation_dataset(
          where: {
            canceled: {_eq: false},
            status: {_eq: "success"},
            plan_revision: {_eq: %d},
            simulation_start_time: {_eq: "%s"},
            simulation_end_time: {_eq: "%s"},
            simulation: {plan_id: {_eq: %d}}
          }) {
            id
            arguments
            simulation {
              arguments
            }
          }
      }""".formatted(
              planMetadata.planRev(),
              planMetadata.horizon().getStartInstant(), // this may need a parse
              planMetadata.horizon().getEndInstant(), // this may need a parse 
              planMetadata.planId().id());
```

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsersTest.java
```diff
@@ -28,15 +28,7 @@ public static Stream<Arguments> parseGraphQLInterval() {
...
         Arguments.of("+322:21:15", parseDurationISO8601("PT322H21M15S")),
-        Arguments.of("-322:21:15", parseDurationISO8601("PT-322H-21M-15S")),
-        Arguments.of("21:15.111", parseDurationISO8601("PT21M15.111S")),
-        Arguments.of("+21:15.111", parseDurationISO8601("PT21M15.111S")),
-        Arguments.of("-21:15.111", parseDurationISO8601("PT-21M-15.111S")),
-        Arguments.of("15.111", parseDurationISO8601("PT15.111S")),
-        Arguments.of("15.", parseDurationISO8601("PT15S")),
-        Arguments.of("15", parseDurationISO8601("PT15S")),
-        Arguments.of("-15", parseDurationISO8601("PT-15S")),
-        Arguments.of("+15", parseDurationISO8601("PT15S"))
```
Mythicaeda, line 39: Why were these parse tests removed?

Mythicaeda (commented):
Looking good so far! As mentioned during the walkthrough, I'd like to see two E2E tests to prove that "stale" simulation data will not be used.

The first would test the check on Plan Revision, by having a coexistence goal between two activities (ie for each GrowBanana, PeelBanana). The stale sim run will contain fewer GrowBananas than the "current" Plan at the time of scheduling, allowing the number of PeelBananas to indicate whether the stale data was used.

The second would test that the scheduler will not load a dataset that used a different sim config than it will. This can be done by have a coexistence goal on a resource (ie for each plant count < 5, GrowBanana) and then having that resource differ between when sim was run and when scheduling is run.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
Mythicaeda: Oh! Thanks for the catch! @adrienmaillard, how does the scheduler avoid any issues that may arise if a new task was placed between `curTime` and the start time of the batch the engine will resume on?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CardinalityGoal.java
adrienmaillard: This was always the case.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityTemplateConflict.java
adrienmaillard: constructor

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: That's a good question, I need to build a specific test to see what happens in that case.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: It will be.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: I do not know... to be sure we enforce the contract?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: Not sure it means something anymore, I'll remove these comments.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
Mythicaeda: I'm not sure I understand. Could you clarify what you mean by "to be sure we enforce the contract"? If you look at `getBestNewActivities`, on the second line we declare a linked list, which is the object that is returned, potentially with some activities added to it. `acts` can be empty, but there is no way for it to be `null`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityCreationTemplate.java
adrienmaillard: Because the only time we can do something about it is when we insert a new activity in simulation. This is done at line 294 with `removeAndInsertActivitiesFromSimulation` and you can see that we catch the simulation exception because we can actually remove the activity. In other cases, we use the results to instantiate a new instance (that is not yet inserted in the plan). If it fails, it means the error happened somewhere else, that an activity should not have been inserted and we cannot do anything about it other than scream. Basically, if this happens, it means that there is something wrong with the scheduler logic.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
adrienmaillard: Because `initialPlan` is just a temporary buffer before the first simulation.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: I will use that, thank you.

scheduler-server/src/test/java/gov/nasa/jpl/aerie/scheduler/server/graphql/GraphQLParsersTest.java
adrienmaillard: Because they are not valid wrt to the `PGInterval` class and do not get parsed correctly. 
Here is a link to the discussion https://jpl.slack.com/archives/CJ07FGRPH/p1690482612292829 for more details.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: I guess it's defensive programming, you protect against potential changes in other ways than tests. We can remove it if you prefer.

Mythicaeda: Yeah, let's cut it 👍

mattdailis: Sounds good to me - we started out with a lot of null-related bugs in the scheduler prototype because very many things were very often null. Although we've since integrated it into Aerie, we still haven't quite finished adapting the scheduler to the Aerie style of avoiding null altogether

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: I think it can't happen because of how we maintain the current state of the simulation facade. We never have pending activities in the future because everytime we add an activity, we simulate until the end of all present activities (with or without restart). The `simulateUntil(t)` is only acting from the end of the last activity to `t`.

Mythicaeda: What about with daemon tasks? Ie: Let's say that `curTime` is 10hrs, and that `batch` is a daemon task that's going to run at 12hrs. The next activity that is placed is at 11hrs. That would give us a task that's been placed between `curTime` and `batch.offsetFromStart`.

mattdailis: > I think it can't happen

I'm not sure I quite follow the question we're answering here - I'll clarify what I'm concerned about. 

Let's say we previous simulated activity A, and have just been asked to simulate activity B which starts at the same time that activity A ended (note: B is not anchored to A, but merely happens to start at the end time of A).

Suppose that the final effect performed by activity A was to change a value from `Old Value` to `New Value`, and the first action taken by activity B is to read this value.

Screenshot 2023-08-07 at 5 48 02 PM

What value does activity B see?

Aerie simulation semantics dictate that an activity in the plan that is scheduled to start at time `t` will not start causally after any effects that occur at time `t` - i.e. its first step will be concurrent with all other tasks that were scheduled to resume at time `t`. My concern is that the scheduler's simulation semantics will differ from Aerie's simulation semantics - possibly leading to subtle bugs

adrienmaillard: This is the issue we were discussing with @Mythicaeda recently. Even though the semantics dictates something, there must be something in the implementation that make B execute after A because otherwise, a lot of activities placed by coexistence goals would not work as expected in our tests or Clipper.

adrienmaillard: To address your point, I think that indeed, we are following different semantics in the scheduler. Maybe we should roll back on this change... We should probably open a ticket to modify how `CoexistenceGoal` works + documentation to tell people about the semantics of using `startsAt` (and the cost in simulation time). Maybe make some syntactic sugar with 'startsJustAfter' that would make the activity starts 1 microsecond after...

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -766,6 +773,244 @@ public DatasetId storeSimulationResults(final PlanMetadata planMetadata,
...
+    final JsonObject response;
+    response = postRequest(request).get();
+    final var data = response.getJsonObject("data").getJsonArray("profile");
+    return parseProfiles(data);
+}
+
+private Map<SimulatedActivityId, UnfinishedActivity> getSpans(SimulationDatasetId datasetId, Instant startTime) throws PlanServiceException, IOException {
+    final var request = """
+       query{
+       span(where: {duration: {_is_null: true}, dataset_id: {_eq: %d}}) {
```
Mythicaeda, line 829: Why do we specifically want spans with `null` durations here?

adrienmaillard: Spans with the null durations are the unfinished activities (?)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: I have created an issue #1087

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
Mythicaeda: Oh! The method name confused me.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: Reverted this change btw.

Mythicaeda: I don't think reverting this change/the created ticket covers the other issue, which is that the scheduler can place activities in the gap between `curTime` and `batch.startOffset`, essentially placing work in the past of the engine (see [this comment](https://github.com/NASA-AMMOS/aerie/pull/1051#discussion_r1286488261) for an example case).

adrienmaillard: @Mythicaeda this is probably worth another ticket.

adrienmaillard: https://github.com/NASA-AMMOS/aerie/issues/1097

Mythicaeda: Thank you!

adrienmaillard:
@Mythicaeda I have added the 2 e2e-tests.

e2e-tests/src/tests/scheduler-with-sim.test.ts
```diff
@@ -0,0 +1,312 @@
+
+import {expect, test} from "@playwright/test";
+import req from "../utilities/requests.js";
+import time from "../utilities/time.js";
+
+/*
+This is testing the check on Plan Revision before loading initial simulation results. We inject results associated with an old plan revision.
+In these results, there is only one GrowBanana activity instead of the actual 2 present in the latest plan revision.
+A coexistence goal attaching to GrowBanana activities allows to show whether the scheduler has picked up the stale sim results or not.
```
Mythicaeda:
```suggestion
A coexistence goal attaching to GrowBanana activities allows shows that the scheduler did not use the stale sim results.
```

e2e-tests/src/tests/scheduler-with-sim.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      plan_id : plan_id,
+      plan_revision : plan_revision,
+      simulation_arguments : {},
+      analysis_only: false
+    }
+    specification_id = await req.insertSchedulingSpecification(request, schedulingSpecification);
+
+    const priority = 0;
+    const specGoal: SchedulingSpecGoalInsertInput = {
+      // @ts-ignore
```
Mythicaeda: What is being ignored here?

e2e-tests/src/tests/scheduler-with-sim.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      dataset_id = datasetId
+      await delay(1000);
+    }
+    if (status_local == "failed") {
+      console.error(reason_local)
+      throw new Error(reason_local);
+    }
+    expect(status_local).toEqual("complete")
+    expect(analysisId_local).toEqual(analysisId)
+    expect(dataset_id).not.toBeNull();
```
Mythicaeda: Minor: this should probably get pulled out into an `awaitScheduling` function, similar to what was done with `awaitSimulation` in `requests.ts`

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationFacade.java
```diff
@@ -41,22 +39,30 @@ public class SimulationFacade implements AutoCloseable{
...
 
-  public gov.nasa.jpl.aerie.constraints.model.SimulationResults getLatestConstraintSimulationResults(){
-    return lastSimConstraintResults;
+  public SimulationData lastSimulationData;
+
+  /**
+   * The set of activities to be added to the first simulation.
+   * Used to potentially delay the first simulation until the loaded results are stale.
+   * The only way to add activities to the facade is to simulate them. But sometimes, we have initial sim results and we
+   * do not need to simulate before the first activity insertion. This initial plan allows to "load" the activities in simulation
```
Mythicaeda: Minor: typo (I'm assuming `the facade` is the dropped subject?)
```suggestion
   * do not need to simulate before the first activity insertion. This initial plan allows the facade to "load" the activities in simulation
```

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/PrioritySolverTest.java
```diff
@@ -149,7 +150,7 @@ public void getNextSolution_proceduralGoalCreatesActivities() {
     assertThat(plan.getActivitiesByTime())
         .comparingElementsUsing(equalExceptInName)
         .containsExactlyElementsIn(expectedPlan.getActivitiesByTime());
-    assertEquals(4, problem.getSimulationFacade().countSimulationRestarts());
+    assertEquals(7, problem.getSimulationFacade().countSimulationRestarts());
```
Mythicaeda: Why does this goal now require more simulations? Same is true for other goals in this PR where the number of sims run has increased relative to develop.

adrienmaillard: This has been fixed with the work we did together.

e2e-tests/src/tests/scheduler-with-sim.test.ts
adrienmaillard: Yes, I  figured the same thing as you and did it.

e2e-tests/src/tests/scheduler-with-sim.test.ts
```diff
@@ -0,0 +1,319 @@
+
+import {expect, test} from "@playwright/test";
+import req, {awaitScheduling, awaitSimulation} from "../utilities/requests.js";
+import time from "../utilities/time.js";
+
+/*
+This is testing the check on Plan Revision before loading initial simulation results. We inject results associated with an old plan revision.
+In these results, there is only one GrowBanana activity instead of the actual 2 present in the latest plan revision.
+A coexistence goal attaching to GrowBanana activities allows shows that the scheduler did not use the stale sim results.
```
Mythicaeda: Very Minor
```suggestion
A coexistence goal attaching to GrowBanana activities shows that the scheduler did not use the stale sim results.
```

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -766,6 +775,274 @@ public DatasetId storeSimulationResults(final PlanMetadata planMetadata,
...
+        realProfiles.put(name, realProfile);
+      } else {
+        final var discreteProfile = parseProfile(profile.asJsonObject(), duration, serializedValueP);
+        discreteProfiles.put(name, discreteProfile);
+      }
+    }
+    return new Profiles(realProfiles, discreteProfiles);
+  }
+
+  public <Dynamics> Pair<ValueSchema, List<ProfileSegment<Dynamics>>> parseProfile(JsonObject profile, Duration profileExtent, JsonParser<Dynamics> dynamicsParser){
```
Mythicaeda: How does what this function does differ from `realProfileP.parse(profile).getSuccessOrThrow()`/`discreteProfileP.parse(profile).getSuccessOrThrow()`?

adrienmaillard: It correctly offsets the times of the segments with regard to the total duration of the profile.

e2e-tests/src/tests/scheduler-with-sim.test.ts
adrienmaillard: Done

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: Very similarly to what's done in `GetProfileSegmentsAction`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
```diff
@@ -766,6 +775,274 @@ public DatasetId storeSimulationResults(final PlanMetadata planMetadata,
...
+    final var name = stringP.parse(profile.getJsonString("name"));
```
Mythicaeda: Very small method signature refinement:
```suggestion
  public <Dynamics> Pair<ValueSchema, List<ProfileSegment<Dynamics>>> parseProfile(JsonObject profile, JsonParser<Dynamics> dynamicsParser){
    final var profileExtent = durationFromPGInterval(profile.asJsonObject().getString("duration"));
```
With this we also don't need to grab the `duration` in `parseProfiles` (L916)

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
Mythicaeda: Understood. Can you add a brief comment clarifying that this method is necessary in order to get the duration, since the `duration` column in the DB is not equal to the local "extent" (what is needed in the Java code)?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: done.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinService.java
adrienmaillard: done

Mythicaeda (approved):
LGTM! Excited to see this finally get in!

PR #1052, opened 2023-08-01
Fix Merlin framework Condition negation

DavidLegg:
Fixes two issues with negating Conditions in the Merlin framework:

1) Negation for literal TRUE and FALSE Conditions wasn't implemented.
2) Negation for "and" and "or" Conditions needs to flip the "positive" flag after applying DeMorgan's rule. Without this, there's an infinite recursion.

* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
(see above)

## Verification
Added `ConditionTest` to exercise the literal conditions as well as `Condition.not`, `Condition.or`, and `Condition.and`. Also verified that all tests pass with proposed code, and some of these tests fail consistent with the description above using the code currently on develop.

## Documentation
N/A

## Future work
N/A

### DIFF
--- a/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Condition.java
+++ b/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Condition.java
@@ -28,8 +28,8 @@ public interface Condition {
       if (atLatest.shorterThan(atEarliest)) return Optional.empty();
       if (!positive) return and(not(left), not(right)).nextSatisfied(true, atEarliest, atLatest);
 
-      final var left$ = left.nextSatisfied(positive, atEarliest, atLatest);
-      final var right$ = right.nextSatisfied(positive, atEarliest, left$.orElse(atLatest));
+      final var left$ = left.nextSatisfied(true, atEarliest, atLatest);
+      final var right$ = right.nextSatisfied(true, atEarliest, left$.orElse(atLatest));
 
       if (left$.isEmpty()) return right$;
       if (right$.isEmpty()) return left$;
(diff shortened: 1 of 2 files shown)


merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Condition.java
```diff
@@ -20,13 +20,13 @@ default Condition not() {
...
+  Condition FALSE = not(TRUE);
 
   static Condition or(final Condition left, final Condition right) {
     return (positive, atEarliest, atLatest) -> {
       if (atLatest.shorterThan(atEarliest)) return Optional.empty();
-      if (!positive) return and(not(left), not(right)).nextSatisfied(positive, atEarliest, atLatest);
+      if (!positive) return and(not(left), not(right)).nextSatisfied(true, atEarliest, atLatest);
 
       final var left$ = left.nextSatisfied(positive, atEarliest, atLatest);
       final var right$ = right.nextSatisfied(positive, atEarliest, left$.orElse(atLatest));
```
mattdailis: Thanks to the `!positive` guard, all subsequent references to `positive` are `true`. Would it help readability to use the literal `true` there as well?

Same question applies to the `and` method below.

mattdailis (commented):
Great catch, and thanks for the thorough test cases. I'm stumped why the e2e tests are failing - it doesn't seem related to the changes in this PR

mattdailis (approved):
Looks good to me 🚀

I played around with it locally, and verified that reverting the change in `Condition.java` causes the tests to fail.

PR #1055, opened 2023-08-02
Parse Java docs to pull useful information from mission model

### LINKED ISSUE #1034: Pull Units from mission model annotations and store them in DB
This task is to pull useful information (especially units) from the mission model Java docs and store them in the database. For example activity type parameters and resource types might have units. Activity types might have a color or subsystem. There may be others. More generally how do we pull info out of mission model docs into our db?
Should we store them in columns or just as a JSON blob? How do we specify these special things in Java Doc?

cohansen:
* **Tickets addressed:** Closes #1034
* **Review:** By file

## Description
This PR works off @mattdailis's prototype javadoc parsing code. I added a `JavadocParser` class that handles specific parsing for `@aerie.unit`, `@aerie.computedAttribute`, and `@aerie.resourceName`.

1. For general parameter parsing we look in 2 different spots:
    * They can be annotated at the record level, this requires a `@param` tag followed by a `@aerie.unit` tag on a new line. [Example](https://github.com/NASA-AMMOS/aerie/compare/prototype/parse-javadoc?expand=1#diff-ee3c8857c9e4618c263c5a6422deb026b3ad541d86f3f743c8a6105767631e97R26)
    * Or they can be annotated as a class property, this doesn't require an explicit `@param` tag. [Example](https://github.com/NASA-AMMOS/aerie/compare/prototype/parse-javadoc?expand=1#diff-058555b56d89eadc1cd3fc8269049082dfcdc003e5ed05df6abd0bfdc06abaf3R27)
2. Resource types are annotated in the `@MissionModel` in 2 different spots as well, they are annotated with a `@aerie.resourceName` followed by a `@aerie.unit` tag on a new line:
    * They can exist as a block comment on the `MissionModel` class itself. [Example](https://github.com/NASA-AMMOS/aerie/compare/prototype/parse-javadoc?expand=1#diff-6a577bb2e19b62cea7260e9a3ad9f99f7ea72d8d8dbb05391b0329cb8d8defe5R24)
    * Or they can be on a property of the `MissionModel`. [Example](https://github.com/NASA-AMMOS/aerie/compare/prototype/parse-javadoc?expand=1#diff-6a577bb2e19b62cea7260e9a3ad9f99f7ea72d8d8dbb05391b0329cb8d8defe5R31)
3. The last new tag is `@aerie.computedAttribute`, which must be annotated on a `Record` type returned from the `@EffectModel`. [Example](https://github.com/NASA-AMMOS/aerie/compare/prototype/parse-javadoc?expand=1#diff-058555b56d89eadc1cd3fc8269049082dfcdc003e5ed05df6abd0bfdc06abaf3R52)

~One thing to discuss as part of this PR is if we should prefix all of our custom tags with `@Aerie` or something similar. This would be more verbose but would avoid any conflicts if we pull in a library that uses the same tags as we do or they become official tags as part of the Javadoc spec.~

This PR introduces breaking changes for the UI because we have renamed the following fields:
- `activity_type.parameters` -> `activity_type.parameter_definitions`
- `activity_type.computed_attributes_value_schema` -> `activity_type.computed_attribute_definitions`
- `mission_model_parameters.parameters` -> `mission_model_parameters.parameter_definitions`
- `resource_type.schema` -> `resource_type.definition`

## Verification
Tests were added for some of the parsing, these could be improved as right now we don't have coverage for resource types and computed attributes. If the best place to add this is inside of `FooMissionModel`, I can do that as part of this PR.

## Documentation
https://github.com/NASA-AMMOS/aerie-docs/pull/71

## Future work
More javadoc tag parsing, up next is `@subsystem` and `@color` (#934) which will change some of this code.


merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
```diff
@@ -0,0 +1,157 @@
...
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.util.Elements;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class JavadocParser {
+  private static final String COMPUTED_ATTRIBUTE_TAG = "@computedAttribute";
+  private static final String RESOURCE_TYPE_TAG = "@registeredState";
```
mattdailis: The word state doesn't fit with the Aerie vocabulary, at least not in the sense used here. Let's find a way to work the word "resource" in, perhaps `@resourceName`?

cohansen: > The word state doesn't fit with the Aerie vocabulary, at least not in the sense used here. Let's find a way to work the word "resource" in, perhaps `@resourceName`?

Sounds good, I was pulling from Clipper's examples.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
```diff
@@ -0,0 +1,175 @@
...
+    Optional.ofNullable(elementUtils.getDocComment(element))
+        .map(JavadocParser::removeSingleLeadingSpaceFromEachLine)
+        .map(comment -> {
+          parseComment(comment, parsedUnits, RESOURCE_TYPE_TAG);
+
+          return comment;
+        });
+
+    element
+        .getEnclosedElements()
```
mattdailis, line 41: From reading this code, it appears that all `@resourceName` annotations are expected to occur in the top-level Mission class. Is this an accurate reading?

I'm wondering whether that would cause resource unit declarations to be written far from the registration of those resources

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
```diff
@@ -0,0 +1,175 @@
...
+        .flatMap(e -> Optional
+            .ofNullable(elementUtils.getDocComment(e))
+            .map(JavadocParser::removeSingleLeadingSpaceFromEachLine)
+            .map(comment -> Pair.of(e.getSimpleName().toString(), comment))
+            .stream())
+        .forEach($ -> {
+          final var value = $.getValue();
+          var property = $.getKey();
+
+          System.out.println(value);
```
mattdailis: Some lingering debug statements - probably a good idea to remove them to avoid cluttering the compile output

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java
```diff
@@ -304,7 +309,24 @@ public void refreshResourceTypes(final String missionModelId)
...
       final var model = this.loadAndInstantiateMissionModel(missionModelId);
-      this.missionModelRepository.updateResourceTypes(missionModelId, model.getResources());
+      final var unitMap = new HashMap<String, String>();
+
+      // Resource type units can be regex so go through our resources and apply the matching units.
+      for (final var resource : model.getResources().keySet()) {
+        for (final var unitString : model.getResourceTypeUnits().keySet()) {
+          // Check validity of the regex.
+          try {
+            Pattern.compile(unitString);
```
mattdailis: I think of regex compilation as a somewhat expensive operation - we may want to consider swapping the inner and outer loop

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
```diff
@@ -18,6 +18,11 @@
  */
 @ActivityType("BiteBanana")
 public final class BiteBananaActivity {
+
+  /**
+   * The amount by which to reduce /fruit
+   * @aerie.unit number
```
mattdailis, line 24: Just playing around a bit - any thoughts on whether I'm "holding it wrong", or if it's worth altering the behavior here?

```java
/**
 * The amount by which to reduce /fruit
 * @aerie.unit number
 *
 * Other information that I wouldn't expect to be captured in the unit.
 */
@Parameter
public double biteSize = 1.0;
```

```java
@Override
public Map<String, String> getParameterUnits() {
  return Map.ofEntries(
          Map.entry("biteSize", "number\n"
              + "\n"
              + "Other information, that I wouldn't expect to be captured in the unit."));
}
```

deployment/hasura/migrations/AerieMerlin/23_add_units_to_activity_types/down.sql
```diff
@@ -0,0 +1,11 @@
+comment on column activity_type.parameter_units is null;
+comment on column activity_type.computed_attribute_units is null;
```
camargo: When talking with @duranb he brought up this new column and asked what was the reason for it? Could we instead just add all this data to the existing `parameter`? It would make the API much easier to use.

camargo:
Before merging, we will need to update the `activity_type` permissions for the different roles for the new columns.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
cohansen: You can declare them as a javadoc on the Mission class, or a property of the class. Here's a small example:
```
/**
 * @aerie.resourceName \/p.*
 * @aerie.unit Things that start with p
 */
public final class Mission {
  public final Accumulator fruit;
  public final AdditiveRegister peel;
  /**
   * @aerie.resourceName /flag
   * @aerie.unit flag
   */
  public final Register<Flag> flag;
}
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
cohansen: Good catch, thought I removed everything!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
cohansen: This was just an oversight in implementation, I didn't expect that users would add text after the unit annotation but that's perfectly reasonable. I think the expected behavior here is that we'd only capture `@aerie.unit number`. Do you agree?

deployment/hasura/migrations/AerieMerlin/23_add_units_to_activity_types/down.sql
cohansen: It's added in this way to keep it consistent with computed attributes and resource types.

We don't store computed attributes in the same way, only their schema so I don't think it makes sense to alter that object, same with resource types.

Maybe we should huddle to discuss further?

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
mattdailis: I concur :+1:

deployment/hasura/migrations/AerieMerlin/23_add_units_to_activity_types/down.sql
camargo: That reasoning makes sense. Sure ping me and we can huddle!

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/utils/JavadocParser.java
mattdailis: Gotcha - for more complex models, resources tend to be defined in sub-models - e.g:

```java
// Mission.java
public final class Mission {
  public final PowerModel power;
}
```

```java
// PowerModel.java
public final class PowerModel {
  /**
   * @aerie.resourceName /power/soc
   * @aerie.unit percentage
   */
  public final Accumulator batteryStateOfCharge;
}
```

Is this something you feel like we could/should support?

cohansen: I can look into this yes, I think if it's a common use-case we should try and support it!

sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts
```diff
@@ -199,6 +199,8 @@ export function mapGraphQLActivityInstance(
   activityInstance: GraphQLSimulatedActivityInstance<any, any>,
   activitySchema: GraphQLActivitySchema,
 ): SimulatedActivity {
+  console.log(activitySchema);
+
```
Mythicaeda, line 203: Minor: `console.log` needs to be removed

deployment/hasura/migrations/AerieMerlin/26_add_units_to_activity_types/down.sql
Mythicaeda, line 1: Doesn't this file need to be modifying the entries in these columns, as the entries have additional keys?

PR #1062, opened 2023-08-03
Add TrampoliningTask

DavidLegg:
Adds a new kind of Task, called TrampoliningTask.
This task directly uses the continuation pattern in the Task interface, without using a different thread like ThreadedTask or exceptions like ReplayingTask. Instead, TrampoliningTask doesn't support operations that would break the normal flow of control.

The main use case for trampolining tasks are daemon tasks spawned by a mission model which do very little work per iteration but run for many iterations. Since they do very little work per iteration, using a separate thread incurs large overhead. Since they run for many iterations, using a replaying task requires time quadratic in the plan length.

However, these daemon tasks also tend not to require arbitrary interruptions to the flow of control. Instead, they tend to follow a simple pattern of doing a small amount of work and then repeating pending a delay or condition. This lends itself well to trampolining.

It is possible to implement trampolining with replaying tasks, without modifying Merlin. However, this requires registering a brand new task with every iteration. As such, the SimulationEngine consumes memory linear in the length of the plan to track these tasks. By implementing the trampolining tasks directly in Merlin, we avoid this memory creep.

Finally, we demonstrate the use of trampolining tasks in the Foo mission model. Performance testing suggests that this change increases the simulation-to-real-time ratio for this model from ~10,000 when using threaded tasks, to ~1.6 million using trampolining tasks, without significantly changing the memory usage.

* **Tickets addressed:** N/A
* **Review:** By file  

## Description
(see above)

## Verification
The Foo mission model was changed to use trampolining repeating tasks for its daemon tasks. Locally, I also exposed the timeTrackerDaemon's minutesElapsed resource and verified that it was updated correctly, though this change wasn't committed. I also measured the time the model took to run on 1-year plans, and used a profiler to ensure that the engine wasn't storing large numbers of task-related objects.

Additionally, I copied the trampolining strategy I use in the Clipper model over to the Foo model for comparison. In this case, memory usage was so high that 1-year plans ran out of memory before being completed. Profiling confirmed that large amounts of memory were used to track tasks, replicating the behavior from the Clipper model. This code was not committed.

I'd like to have more careful unit testing around the trampolining tasks, but I'm not sure how to implement that testing within the Aerie repo; any suggestions here would be appreciated.

## Documentation
No documentation is included yet, beyond comments in the code itself. However, it may be helpful to have some additional discussion for modelers around the pros and cons of each kind of task, and when to use each type. Suggestions on where to put such a discussion would be appreciated.

## Future work
N/A - This feature is self-contained.

### DIFF
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/Mission.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/Mission.java
@@ -11,7 +11,6 @@ import gov.nasa.jpl.aerie.foomissionmodel.models.Imager;
 import gov.nasa.jpl.aerie.foomissionmodel.models.ImagerMode;
 import gov.nasa.jpl.aerie.foomissionmodel.models.SimpleData;
 import gov.nasa.jpl.aerie.foomissionmodel.models.TimeTrackerDaemon;
-import gov.nasa.jpl.aerie.merlin.framework.ModelActions;
 import gov.nasa.jpl.aerie.merlin.framework.Registrar;
 import gov.nasa.jpl.aerie.merlin.framework.resources.real.RealResource;
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
@@ -19,6 +18,7 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import java.time.Instant;
 
 import static gov.nasa.jpl.aerie.merlin.framework.ModelActions.*;
+import static gov.nasa.jpl.aerie.merlin.framework.TrampoliningTask.RepeatingTaskStatus.delayed;
 
 public final class Mission {
   // Need a way to pose constraints against activities, and generally modeling activity behavior with resources.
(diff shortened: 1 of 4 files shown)


camargo:
What is the status on this one?

Twisol:
> What is the status on this one?

@camargo, I've been chatting with Matt and David about this PR -- in particular this paragraph:

> It is possible to implement trampolining with replaying tasks, without modifying Merlin. However, this requires registering a brand new task with every iteration. As such, the SimulationEngine consumes memory linear in the length of the plan to track these tasks. By implementing the trampolining tasks directly in Merlin, we avoid this memory creep.

I have a tentative alternative approach implemented [on this branch](https://github.com/Twisol/aerie/tree/feature/castello--engine-efficiency) that follows this idea, improving the efficiency of tasks to leave behind less garbage (including the linear consumption due to parent-child relationships). This approach also led to a (IMHO) reduction in complexity in the engine, which is always welcome.

One of the drawbacks to introducing a new type of task is the lack of support for [control flow effects](https://github.com/NASA-AMMOS/aerie/pull/1062/files#diff-af531dd2d3da14c126411e5461a7979745355c170054d85f83a256e865f0c1c1R129-R145). This means that you have to be very careful, when using a trampolining task, to avoid calling existing model methods that use e.g. `delay` or `call`. By improving engine support for spawn-based trampolining, we can avoid this schism between trampolined code and regular model code.

I'm hoping to chat with @mattdailis and @DavidLegg sometime soon (TBD?) to see if these changes are viable for inclusion upstream.

Twisol:
(To be clear, I think that the lack of support for control flow effects in this PR could be mitigated -- for instance, by using a replaying-style task whose breadcrumbs are cleared on every iteration. I'm actually quite a fan of the idea of using novel task description styles for efficient and (ideally) idiomatic solution of specialized domain problems. However, I think that in this *particular* instance, we benefit from improving the engine's general efficiency and reusing existing idioms.)

DavidLegg:
Closing this PR in favor of https://github.com/NASA-AMMOS/aerie/pull/1174

PR #1064, opened 2023-08-03
Let `viewer` role get effective args

Mythicaeda:
* **Review:** By commit  

## Description

Noticed while using the UI that users using the `viewer` role would see no arguments for activities/simulation configuration, with the console having a "cannot find in query_root" error. This PR tweaks the metadata so that `viewer`s can now _view_ these arguments.

## Verification

Before:
image

After
image

## Future work

Are there any other Actions that the UI expects all roles to have access to?

### DIFF
--- a/deployment/hasura/metadata/actions.yaml
+++ b/deployment/hasura/metadata/actions.yaml
@@ -71,6 +71,7 @@ actions:
     permissions:
       - role: aerie_admin
       - role: user
+      - role: viewer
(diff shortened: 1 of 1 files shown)


cohansen (approved):
Echoing your question in the followup, maybe it's worth going through the actions and double checking everything with a few team members?

Approved pending the checks!

mattdailis (approved):

PR #1076, opened 2023-08-08
Reduce Worker Pool Size

### LINKED ISSUE #631: Postgres container unresponsive: remaining connection slots are reserved for non-replication superuser connections
We got into a situation where postgres became completely unresponsive for about 10 min. During this time the UI would not load any models, plans, etc, and CPU usage on the Postgres container was pegged at 100%. After about 10 min CPU dropped and the service became responsive again. Around this time I was trying to delete some plans, and experience very long query times. Don't know if that was a cause of the CPU spike or not. Postgres log attached.

[postgres.log.txt](https://github.com/NASA-AMMOS/aerie/files/10551705/postgres.log.txt)

### LINKED ISSUE #1072: Do a pass and try to minimize database connections
Currently, all of the Merlin/Scheduler Workers have a pool of 10 connections. Each of those only needs 2 connections: 1 for posting/querying data from the DB and 1 for listening to notifications.

Mythicaeda:
* **Tickets addressed:** Closes #1072, Closes #631
* **Review:** By commit  

## Description

Reduces the size of the DB Connection pools in our workers from 10 down to 2. Server connection pool sizes were not adjusted as they can be handling multiple user requests at once.

## Verification

`docker-compose-many-workers.yml` was created. It spins up an Aerie instance with 6 Merlin Workers and 6 Scheduler Workers. 

If this compose is used on `develop`, the postgres container will output the error "max connections reach, remaining connections reserved for super user" ([since PG max_connections is per PG instance, not per DB](https://www.postgresql.org/docs/current/runtime-config-connection.html)). 

If this compose is run against this branch, however, no such error occurs.

Additionally, if you run the SQL statement `SELECT sum(numbackends) FROM pg_stat_database;`, only ~50 of the 100 available connections should be in use (I've seen this vary from 48-58).

## Documentation

No documentation is needed

### DIFF
--- a/e2e-tests/docker-compose-many-workers.yml
+++ b/e2e-tests/docker-compose-many-workers.yml
@@ -1,0 +1,209 @@
+version: "3.7"
+services:
+  aerie_gateway:
+    container_name: aerie_gateway
+    depends_on: ["postgres"]
+    environment:
+      AUTH_TYPE: none
+      AUTH_URL: https://atb-ocio-12b.jpl.nasa.gov:8443/cam-api
+      GQL_API_URL: http://localhost:8080/v1/graphql
+      HASURA_GRAPHQL_JWT_SECRET: "${HASURA_GRAPHQL_JWT_SECRET}"
+      LOG_FILE: console
+      LOG_LEVEL: debug
+      NODE_TLS_REJECT_UNAUTHORIZED: "0"
+      PORT: 9000
+      POSTGRES_AERIE_MERLIN_DB: aerie_merlin
+      POSTGRES_HOST: postgres
+      POSTGRES_PASSWORD: "${AERIE_PASSWORD}"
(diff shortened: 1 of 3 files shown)


skovati (commented):
This looks like a good opportunity to leverage `docker compose`'s ability to [include](https://docs.docker.com/compose/multiple-compose-files/include/) other `docker-compose.yml` files, so that we don't have extraneous copies that all need to be updated if we e.g. add an environment variable.

skovati (approved):
Looks good! How did you get around workers needing unique ports?

Mythicaeda:
I removed the port mappings from the container to the host machine. `many-workers` proves that we don't run out of db connections spinning up, well, many workers (It's up to 16 workers total and hypothetically maxes out at 25 workers), it doesn't need exposed health/debugger ports.

PR #1086, opened 2023-08-09
Remove unneeded test task in db-tests build.gradle

camargo:
* **Tickets addressed:** None
* **Review:** By commit  

## Description

* Test task is not needed because the database is required for the DB tests to run
* Update the test workflow to add db-tests output to artifacts

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,6 +87,12 @@ jobs:
         with:
           name: Test Results
           path: "**/e2e-tests/test-results"
+      - name: Upload DB Test Results as HTML
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: Test Results
+          path: "**/db-tests/build/reports/tests/e2eTest"
       - name: Upload Sequencing Server Test Results
         if: always()
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -87,6 +87,12 @@ jobs:
         with:
           name: Test Results
           path: "**/e2e-tests/test-results"
```
skovati, line 89: I think db-test results are already uploaded here, since `./gradlew e2eTest` calls `:db-tests:e2eTest`, but you might want to double check i.e. remove this additional `upload-artifact` call you added and see if db-test results are still there.

skovati (approved):
Looks good to me either way!

.github/workflows/test.yml
camargo: I'm not seeing the `db-tests` results in the `e2e-tests` directory

camargo: Will verify without just to see

camargo: Yeah the additional upload artifact is needed to grab the nested `e2eTest` folder from the `db-tests`

skovati: Yep you are correct!

PR #1090, opened 2023-08-11
Addition of two examples to SchedulingIntegrationTests to illustrate …

### LINKED ISSUE #994: Define a goal type that instantiates one single activity
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

_No response_

I want to add the capability to express a goal that results in the instantiation of one single activity. 
The use should be able to add logic or temporal conditions to the goal such as indicating the specific interval at where the activity should be place (for example a direct to earth communication from the rover at the end of a sol)

jmdelfa:
…how to model a unitary goal.

* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/994
* **Review:** By commit  

### DIFF
--- a/scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
+++ b/scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
@@ -319,7 +319,7 @@ export class Goal {
    * The Coexistence Goal specifies that a certain activity should occur once **for each** occurrence of some condition.
    *
    * #### Inputs
-   * - **forEach**: a set of time windows (`Windows`, see [documentation](../../constraints-edsl-api/classes/Windows) on how to produce such an expression) or a set of activities (`ActivityExpression`)
+   * - **forEach**: a set of time windows (`Windows`, see [documentation](../../constraints-edsl-api/classes/Windows) on how to produce such an expression) or a set of activities (`ActivityExpression`) or an Interval or a Temporal.Instant
    * - **activityTemplate**: the description of the activity to insert after each activity identified by `forEach`. If activityFinder is not defined, used for both matching against existing activities and creating new ones.
    * - **activityFinder: an optional activity expression. If present, it will be used as replacement of activityTemplate to match against existing activities in the plan.
    * - **startsAt**: optionally specify a specific time when the activity should start relative to the window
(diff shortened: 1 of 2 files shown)


adrienmaillard:
@jmdelfa I was reading these tests and I felt bad for our users having to do all the boilerplate with `Windows` and `Interval` so I made another commit to your branch adding the possibility for the user to give `Interval` or `Temporal.Instant` directly to `CoexistenceGoal` and then we do the conversion ourselves. I have also modified the examples you gave.

adrienmaillard:
The only downside for you is that you would have to write more documentation to show these use-cases to the user :)

jmdelfa:
I agree with the modifications proposed by @adrienmaillard

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
```diff
@@ -393,7 +393,16 @@ export class Goal {
...
 
-    if (opts.forEach instanceof WindowsEDSL.Windows) {
+    let localForEach;
+    if(opts.forEach instanceof WindowsEDSL.Interval){
+      localForEach = Windows.Value(true, opts.forEach).assignGaps(false);
+    } else if(opts.forEach instanceof Temporal.Instant){
+      localForEach = Windows.Value(true,Interval.At(opts.forEach)).assignGaps(false);
+    }else{
+      localForEach = opts.forEach;
+    }
```
mattdailis: Here are the spaces you and I discussed adding - it doesn't look to me like they made it into the final version

```suggestion
    if (opts.forEach instanceof WindowsEDSL.Interval) {
      localForEach = Windows.Value(true, opts.forEach).assignGaps(false);
    } else if (opts.forEach instanceof Temporal.Instant) {
      localForEach = Windows.Value(true,Interval.At(opts.forEach)).assignGaps(false);
    } else {
      localForEach = opts.forEach;
    }
```

mattdailis (changes requested):

scheduler-worker/scheduling-dsl-compiler/src/libs/scheduler-edsl-fluent-api.ts
jmdelfa: Updated code with spacing included

mattdailis (approved):

mattdailis (approved):

mattdailis (approved):

jmdelfa:
All comments resolved. PR ready to be merged

adrienmaillard:
WOW, triple approval, this must mean something :-) 
Screenshot 2023-08-30 at 3 20 18 PM

PR #1093, opened 2023-08-16
Support non-commands expansions

### LINKED ISSUE #853: Enable non-command step types in Command Expansion
Currently, `Arrayable<CommandStem>` is the only allowable return value for command expansion. It will be necessary to activate/load/execute sequences via command expansion, so the `ExpansionReturn` type should allow these step types (which, I believe, are already added/being added to the EDSL).

goetzrrGit:
* **Tickets addressed:** Closes #853 
* **Review:** By commit  

## Description
There are five **Step** that a user can define within a sequence **(Command, Load, Activate, Ground_Event, and Ground_Block)**. Initially, only the **Command** step was eligible for expansion during both the 'command expansion' and 'sequence generation' processes. In order to execute background sequences, users will utilize the **Load** and **Activate** steps to initiate a background sequence within the sequence engine. Furthermore, users are interested in being able to formulate expansion logic using both **Load** and **Activate** steps. 

Provided below is an illustrative example that demonstrates the utilization of Load, Activate, and Command within an Authored Expansion Logic.

```ts
export default function MyExpansion(props: {
  activityInstance: ActivityType
}): ExpansionReturn {
  const { activityInstance } = props;
  return [
          A("2022-203T00:00:00").LOAD("BACKGROUND-A").ARGUMENTS(props.activityInstance.attributes.arguments.temperature),
          A("2022-204T00:00:00").ACTIVATE("BACKGROUND-B"),
          R("00:00:90").ADD_WATER
          ];
}

```

The final sequence generated by expansion

```json
{
  "id": "banana00001",
  "metadata": {
    "planId": 1,
    "simulationDatasetId": 1,
    "timeSorted": false
  },
  "steps": [
    {
      "args": [
        {
          "name": "arg_0",
          "type": "number",
          "value": 350
        }
      ],
      "metadata": {
        "simulatedActivityId": 1
      },
      "sequence": "BACKGROUND-A",
      "time": {
        "tag": "2022-203T00:00:00.000",
        "type": "ABSOLUTE"
      },
      "type": "load"
    },
    {
      "metadata": {
        "simulatedActivityId": 1
      },
      "sequence": "BACKGROUND-B",
      "time": {
        "tag": "2022-204T00:00:00.000",
        "type": "ABSOLUTE"
      },
      "type": "activate"
    },
    {
      "args": [],
      "metadata": {
        "simulatedActivityId": 1
      },
      "stem": "ADD_WATER",
      "time": {
        "tag": "00:01:30.000",
        "type": "COMMAND_RELATIVE"
      },
      "type": "command"
    }
  ]
}
```

## Verification
Added and manually ran through command expansion and sequence generation with a plan.

## Documentation
none 

## Future work
none

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -246,8 +246,6 @@ export type RequestOptions =
   metadata?: Metadata;
 };
 
-export type Arrayable<T> = T | Arrayable<T>[];
-
 /**-----------------------------
  *      GLOBAL eDSL Declarations
  * -----------------------------
@@ -385,7 +383,8 @@ declare global {
   }
 
   type Context = {};
-  type ExpansionReturn = Arrayable<CommandStem>;
+  // @ts-ignore : 'Description' found in JSON Spec
+  type ExpansionReturn = (Command | Load | Activate)[];
 
   type U<BitLength extends 8 | 16 | 32 | 64> = number;
   type U8 = U<8>;
@@ -1224,9 +1223,10 @@ export function ENUM<const N extends string, const E extends string>(
 // @ts-ignore : 'Args' found in JSON Spec
 export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}> implements Command {
   public readonly arguments: A;
-  public readonly absoluteTime: Temporal.Instant | null = null;
-  public readonly epochTime: Temporal.Duration | null = null;
-  public readonly relativeTime: Temporal.Duration | null = null;
+  private readonly _absoluteTime: Temporal.Instant | undefined = undefined;
+  private readonly _epochTime: Temporal.Duration | undefined = undefined;
+  private readonly _relativeTime: Temporal.Duration | undefined = undefined;
+
 
   public readonly stem: string;
   // @ts-ignore : 'Args' found in JSON Spec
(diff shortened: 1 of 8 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -1324,29 +1324,41 @@ export class CommandStem<A extends Args[] | { [argName: string]: any } = [] | {}
...
+      ...(this._epochTime && { epochTime: this._epochTime }),
+      ...(this._relativeTime && { relativeTime: this._relativeTime }),
     });
   }
   // @ts-ignore : 'Description' found in JSON Spec
   public GET_DESCRIPTION(): Description | undefined {
     return this._description;
   }
 
+  public GET_ABSOLUTE_TIME(): Temporal.Instant | undefined {
```
dyst5422, line 1337: Why the switch to a getter here?

Could you use an actual getter here? ie
```ts
public get absoluteTime(): Temporal.Instant | undefined {
  return this._absoluteTime;
}
```

it allows you to maintain the same code elsewhere, ie step.absoluteTime

goetzrrGit: I wanted to maintain consistency with the eDSL's design. My approach involves exposing a limited set of methods during chaining so the user can just focus on the CAPITALIZE methods and ignore the other values that show up in the content assist. ex:

BAKE_BANANA.<a subset of methods>

Content assisted box:
...
MODEL(model)
GET_MODEL()
METADATA(metadata)
GET_METADAT()
DESCRIPTION(string)
GET_DESCRIPTION()
GET_ABSOLUTE_TIME()
GET_EPOCH_TIME()
GET_RELATIVE_TIME()
arguments
name
type
...

goetzrrGit: As I move forward with the custom worker, I'm considering implementing a filter to refine the content assist feature. I want to only show those methods that are CAPITALIZED while concealing the others while the user is typing.

dyst5422: Fair enough. I suppose if modifying the autocomplete was easier for us, we might have gone a different route, but I like the ingenuity in working with what we have!

cartermak:
Thank you! I tested this and it looks good to me; I'm still waiting on feedback from a few stakeholders but I think Clipper would be happy to see this in Monday's release.

camargo (approved):
LGTM

PR #1105, opened 2023-08-23
Scheduling edsl real resource tolerance

### LINKED ISSUE #846: Add toleranced equality check to Real objects (eDSL)
It would be nice to have a way to check if a real resource is equal to some value within a specified tolerance in the scheduling eDSL. Right now the way of doing this would look  something like

```ts
var value = 4.0;
var tol = 0.001;
Real.Resource('/fruit').lessThan(value + tol).and(Real.Resource('/fruit').greaterThan(value - tol))
```

which is unnecessarily clunky.

A preferable syntax might be:

 ```ts
 Real.Resource('/fruit').isWithin(tol).of(value)
 ``` 
 
 or 
 
 ```ts
 Real.Resource('/fruit').isWithin(tol, value)
 ```

jmdelfa:
Closes https://github.com/NASA-AMMOS/aerie/issues/846

Adding a new EDSL fluent to express a real resource that has a value and a tolerance that acts as an error margin for the value.

The syntax is: 

```ts
Real.Resource("name").isWithin(value, tolerance)
```

The syntax uses already existing edsl constraints to represent the tolerance. The constraint is evaluated as follows:
- the resource value is less than or equal the value + tolerance
- the resource value is greater than or equal the value - tolerance

A test has been added to ConstraintsDSLCompilationServiceTests to demonstrate the correctness

### DIFF
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
@@ -35,6 +35,7 @@ export enum NodeKind {
   RealProfileLessThanOrEqual = 'RealProfileLessThanOrEqual',
   RealProfileGreaterThan = 'RealProfileGreaterThan',
   RealProfileGreaterThanOrEqual = 'RealProfileGreaterThanOrEqual',
+  RealProfileIsWithin = 'RealProfileIsWithin',
   WindowsExpressionAnd = 'WindowsExpressionAnd',
   WindowsExpressionOr = 'WindowsExpressionOr',
   WindowsExpressionNot = 'WindowsExpressionNot',
@@ -88,6 +89,7 @@ export type WindowsExpression =
   | RealProfileLessThanOrEqual
   | RealProfileGreaterThan
   | RealProfileGreaterThanOrEqual
+  | RealProfileIsWithin
   | DiscreteProfileTransition
   | ExpressionEqual<RealProfileExpression>
   | ExpressionEqual<DiscreteProfileExpression>
@@ -160,6 +162,13 @@ export interface RealProfileGreaterThanOrEqual {
   kind: NodeKind.RealProfileGreaterThanOrEqual;
   left: RealProfileExpression;
   right: RealProfileExpression;
+}
+
+export interface RealProfileIsWithin {
+  kind: NodeKind.RealProfileIsWithin;
+  left: RealProfileExpression;
+  right: RealProfileExpression;
+  margin: RealProfileExpression;
 }
 
 export interface RealProfileGreaterThan {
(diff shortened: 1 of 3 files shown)


merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -162,6 +164,13 @@ export interface RealProfileGreaterThanOrEqual {
   right: RealProfileExpression;
 }
 
+export interface RealProfileIsWithin {
+  kind: NodeKind.RealProfileIsWithin;
+  left: RealProfileExpression;
+  right: RealProfileExpression;
+  margin: RealProfileExpression;
+}
+
```
JoelCourtney: For edsl-only wrappers like this, you don't need to create a new AST node. AST changes are only needed when creating a new node in the Java backend.

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -88,6 +89,7 @@ export type WindowsExpression =
   | RealProfileLessThanOrEqual
   | RealProfileGreaterThan
   | RealProfileGreaterThanOrEqual
+  | RealProfileIsWithin
```
JoelCourtney: same comment

merlin-server/constraints-dsl-compiler/src/libs/constraints-ast.ts
```diff
@@ -35,6 +35,7 @@ export enum NodeKind {
   RealProfileLessThanOrEqual = 'RealProfileLessThanOrEqual',
   RealProfileGreaterThan = 'RealProfileGreaterThan',
   RealProfileGreaterThanOrEqual = 'RealProfileGreaterThanOrEqual',
+  RealProfileIsWithin = 'RealProfileIsWithin',
```
JoelCourtney: same comment

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintsDSLCompilationServiceTests.java
```diff
@@ -633,7 +633,26 @@ export default () => {
...
+                    new LessThanOrEqual(new RealResource("an integer"), new Plus(new RealValue(10), new RealValue(1))),
+                    new GreaterThanOrEqual(new RealResource("an integer"), new Plus(new RealValue(10), new Times(new RealValue(1), -1)))
+                )
+            )
+        )
+    );
+  }
+
+  @Test
+  void testReChanges() {
```
JoelCourtney: looks like the name of the next test got mangled (`testReChanges`)

JoelCourtney (changes requested):
cool!

JoelCourtney (approved):

PR #1106, opened 2023-08-24
Add initial basic load testing

### LINKED ISSUE #1095: Add Load Testing

skovati:
* **Tickets addressed:** closes #1095 
* **Review:** By commit  

## Description
Adds basic load tests using [k6](https://k6.io/), an industry standard load testing tool. This PR is meant as a starting point, i.e. setting up the initial testing infrastructure and automation, so that more specific load tests can be easily added.

Currently spins up 10-20 users who create plans and activities and run simulations. Results are uploaded as JSON for further (trending) analysis, as well as visualized as HTML using [xk6-dashboard](https://github.com/grafana/xk6-dashboard).

The following "scenarios" are run in parallel, with ~20 (virtual) users running each scenario:
 - `query_missions`, simple GraphQL query getting existing missions
 - `insert_plans`, randomly inserts new plans
 - `insert_random_activities`, randomly inserts new activities
 - `insert_targeted_activities`, inserts activities into `plan_id` 1, to generate a large plan
 - `run_targeted_simulations`, runs simulations on random (mostly sparse) plans
 - `run_targeted_simulations`, runs a handful of simulations on our artifically large plan 

Additional scenarios can be added as we expand our load-testing of `aerie`.

The main quirk of k6 is that it uses [goja](https://github.com/dop251/goja) to execute javascript tests, which means it is not compatible with our existing playwright-based e2e tests.

## Verification
The new load testing workflow will fail if thresholds aren't met, as defined here:
https://github.com/NASA-AMMOS/aerie/blob/90325e75518d9741113d9ba43ca77e766b80f4b3/load-tests/src/load-test.ts#L100-L104

## Documentation
Added a `README.md` as well as inline docs to help users run and write tests.

## Future work
- Long-term, trending analysis of load testing results, e.g. "are simulations taking longer with this new feature"
- Track custom metrics, e.g. "simulation time"
- Storing load test run info in a long-term datastore (e.g. [prometheus](https://prometheus.io/)) and [grafana](https://grafana.com/)

### DIFF
--- a/.github/workflows/load-test.yml
+++ b/.github/workflows/load-test.yml
@@ -1,0 +1,58 @@
+name: Load Test
+
+on:
+  workflow_dispatch:
+
+env:
+  AERIE_USERNAME: "aerie"
+  AERIE_PASSWORD: "aerie"
+  HASURA_GRAPHQL_ADMIN_SECRET: "aerie"
+  HASURA_GRAPHQL_JWT_SECRET: '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }'
+  POSTGRES_USER: "postgres"
+  POSTGRES_PASSWORD: "postgres"
+
+jobs:
+  load-test:
+    runs-on: ubuntu-latest
+    environment: load-test
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "temurin"
+          java-version: "19"
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: Assemble
+        run: ./gradlew assemble --parallel
+      - name: Start Services
+        run: |
+          docker compose up -d --build --quiet-pull
+          docker images
+          docker ps -a
+      - name: Sleep for 1 min
+        shell: bash
+        run: sleep 1m
+      - name: Run load test
+        run: |
+          cd load-tests
+          ./load-test.sh
+      - name: Upload Load Test Results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: Load Test Results
+          path: "**/load-tests/load-report.*"
+      - name: Stop Services
+        if: always()
+        run: |
+          docker ps -a
+          docker compose down
+          docker ps -a
+      - name: Prune
+        if: always()
+        run: docker volume prune -fa
--- a/load-tests/.dashboard.js
+++ b/load-tests/.dashboard.js
@@ -1,0 +1,28 @@
+function chartTimings(metric, title) {
+  return {
+    id: metric,
+    title: title,
+    series: {
+      [`${metric}_trend_avg`]: { label: 'avg', width: 2, format: 'duration' },
+      [`${metric}_trend_p(90)`]: { label: 'p(90)', format: 'duration' },
+      [`${metric}_trend_p(95)`]: { label: 'p(95)', format: 'duration' }
+    },
+    axes: [{}, {format: 'duration'}, { side: 1, format: 'duration' }],
+    height: 224
+  }
+}
+
+const customTab = {
+    id: `custom_snapshot`,
+    title: `Custom Metrics`,
+    event: "snapshot",
+    charts: [
+      chartTimings('effective_arg_duration', 'Effective Argument Request Duration'),
+    ],
+    report: true,
+    description: 'Example tab displaying custom metrics'
+};
+
+defaultConfig.tabs.push(customTab);
+
+export default defaultConfig;
--- a/load-tests/src/load-test.ts
+++ b/load-tests/src/load-test.ts
@@ -1,0 +1,285 @@
+import http from 'k6/http';
+import { sleep, check } from 'k6';
+import { Trend } from 'k6/metrics';
+import { Options } from 'k6/options';
+import { req } from './requests';
+import gql from '../../e2e-tests/src/utilities/gql';
+import * as urls from '../../e2e-tests/src/utilities/urls';
+
+// Besides the default metrics k6 records, we can define our own statistics and manually add data points to them
+const effective_args_duration = new Trend('effective_arg_duration', true);
+
+type VUSharedData = {
+  token: string
+};
+
+// The options object configures our overall load test
+// https://k6.io/docs/using-k6/k6-options/reference
+export const options: Options = {
+  // The following scenarios are defined, which all run in parallel
+  scenarios: {
+    insert_plans: {
+      // This defines the function that this scenario will run
+      exec: 'insert_plans',
+
+      // "VUs" (virtual users) are threads that run the function we referenced above in a loop, simulating real users.
+      // Pre-allocating VUs simply saves on run-time resources
+      preAllocatedVUs: 50,
+
+      // When the executor is `ramping-arrival-rate`, the stages define the number of requests
+      // per second we want to target. The k6 scheduler will then automatically scale the number of VUs
+      // in order to hit that desired rate
+
+      // https://k6.io/docs/using-k6/scenarios/executors/
+      executor: 'ramping-arrival-rate',
+      stages: [
+        // Start 10 iterations per second for the first 30 seconds,
+        { target: 10, duration: '30s' },
+        // Then, linearly scale up to 50 iterations/second for a minute,
+        { target: 50, duration: '1m' },
+        // Then scale down again over 30 seconds.
+        { target: 10, duration: '30s' },
+      ],
+    },
+
+    insert_random_activities: {
+      exec: 'insert_random_activities',
+      executor: 'ramping-arrival-rate',
+      // This scenario will only start after 5 seconds, allowing plans to populate first
+      startTime: "5s",
+      preAllocatedVUs: 50,
+      stages: [
+        { target: 10, duration: '30s' },
+        { target: 50, duration: '1m' },
+        { target: 10, duration: '30s' },
+      ],
+    },
+
+    insert_targeted_activities: {
+      exec: 'insert_targeted_activities',
+      // Here we use a different executor, which keeps VUs constant, running as many iterations as
+      // possible for a given duration
+      executor: 'constant-vus',
+      startTime: '5s',
+      vus: 50,
+      duration: '30s'
+    },
+
+    run_targeted_simulations: {
+      exec: 'run_targeted_simulations',
+      startTime: '1m',
+      executor: 'constant-vus',
+      vus: 2,
+      duration: "30s"
+    },
+
+    run_random_simulations: {
+      exec: 'run_random_simulations',
+      startTime: '10s',
+      executor: 'ramping-arrival-rate',
+      preAllocatedVUs: 10,
+      stages: [
+        { target: 5, duration: '30s' },
+        { target: 10, duration: '1m' },
+        { target: 5, duration: '30s' },
+      ],
+    },
+
+    get_effective_args: {
+      exec: 'get_effective_args',
+      executor: 'ramping-arrival-rate',
+      preAllocatedVUs: 10,
+      stages: [
+        { target: 5, duration: '30s' },
+        { target: 10, duration: '1m' },
+        { target: 5, duration: '30s' },
+      ],
+    },
+  },
+
+  // If the statistics of the load test fall below the following threshold, it with exit with a nonzero exit code
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(50)<200', 'p(95)<1000'], // 50% below 200ms, 95% of requests should be below 1000ms
+  },
+};
+
+const plan_start_timestamp = '2021-001T00:00:00.000';
+const jar = open('./banananation.jar', 'b');
+const username = "load-tester";
+// this randomness is run once per VU
+const rd = Math.random() * 10000;
+
+// `setup()` runs once per k6 instance, and sets up a shared mission model for all VUs
+// VUSharedData is a custom type we return from the `setup()` function
+// This data is then accessible as an argument for other load testing functions
+export function setup(): VUSharedData {
+
+  const user = req.login(username);
+  const token = user.token;
+
+  check(user, {
+    'user has correct name': (user) => user.id === username
+  });
+
+  // upload banananation jar
+  const file = {
+    type: "application/java-archive",
+    file: http.file(jar)
+  };
+
+  const res = http.post(`${urls.GATEWAY_URL}/file`, file);
+
+  check(res, {
+    'is status 200': (r) => r.status === 200,
+  });
+
+  const jar_data = res.json();
+  const jar_id  = jar_data.id;
+
+  check(jar_id, {
+    "valid id": (id) => Number.isInteger(id)
+  });
+
+  // create mission model
+  const modelInput: MissionModelInsertInput = {
+    jar_id,
+    mission: "aerie-load-test" + rd,
+    name: "Banananation (load-test)"+rd,
+    version: "0.0.0"+ rd,
+  };
+
+  const model_resp = req.hasura(gql.CREATE_MISSION_MODEL, {model: modelInput}, token, username);
+  const model_data = model_resp.json("data");
+  check(model_data, {
+    "model id is valid": (data) => Number.isInteger(data.insert_mission_model_one.id)
+  });
+
+  sleep(2);
+
+  return {
+    token
+  };
+}
+
+export function insert_plans(data: VUSharedData) {
+  const { token } = data;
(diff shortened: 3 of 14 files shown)


JoelCourtney:
👀

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,76 @@
...
+      - v*
+  workflow_dispatch:
+
+env:
+  AERIE_USERNAME: "aerie"
+  AERIE_PASSWORD: "aerie"
+  HASURA_GRAPHQL_ADMIN_SECRET: "aerie"
+  HASURA_GRAPHQL_JWT_SECRET: '{ "type": "HS256", "key": "oursupersecretsupersecurekey1234567890" }'
+  POSTGRES_USER: "postgres"
+  POSTGRES_PASSWORD: "postgres"
```
Mythicaeda, line 12: Why isn't this using the e2e-test env?

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Run load test
+        run: |
+          cd load-tests
+          ./load-test.sh
+      - name: Upload Load Test Results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: Load Test Results
+          path: "**/load-tests/load-report.*"
```
Mythicaeda, line 49: Minor: Does `./load-test.sh` return a non-0 exit code and produce output if the tests don't pass? If no, `if: success()` would make more sense, since we would only have test results to upload in the `success()` case.

load-tests/src/requests.ts
```diff
@@ -0,0 +1,32 @@
...
+  hasura(
+    query: string,
+    variables: JSONObject,
+    token: string,
+    username: string): JSONValue {
+
+    const headers = {
+      'Authorization': `Bearer ${token}`,
+      'Content-Type': 'application/json',
+      'x-hasura-admin-secret': 'aerie',
```
Mythicaeda: Major: the inclusion of this header means that the `Authorization` header will be ignored.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,284 @@
...
+      plan {
+        id
+      }
+    }
+  `;
+
+  const data = req.hasura(query, {}, token, username);
+  const { plan } = data;
+  const ids: number[] = plan;
+  const random_id = ids[Math.floor(Math.random() * ids.length)];
```
Mythicaeda: This line seems dangerous in the case that `plan` is empty (ie: there are no plans), as the code will try to get index 0 of `ids`, which is empty.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,284 @@
...
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(90)<200', 'p(95)<1000'], // 90% below 200ms, 95% of requests should be below 1000ms
+  },
+};
+
+const plan_start_timestamp = '2021-001T00:00:00.000';
+const jar = open('./banananation.jar', 'b');
+
+const rd = Math.random() * 10000;
+const username = "q";
```
Mythicaeda: Minor: this user needs a more meaningful name than "q", ie: "loadTester", "aerieTester", etc. 

Additionally, we discussed potentially making the VUsers log in when theyre created rather than once as a setup step.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,284 @@
...
+  `;
+
+  const data = req.hasura(query, {}, token, username);
+  const { plan } = data;
+  const ids: number[] = plan;
+  const random_id = ids[Math.floor(Math.random() * ids.length)];
+
+  if (random_id) {
+    const id = random_id.id;
+    const resp = req.hasura(gql.SIMULATE, {plan_id: id}, token, username);
```
Mythicaeda: This oughta be updated to check that `resp` contains a `simulationDatasetId` to assert that the rq was made.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,282 @@
+import http from 'k6/http';
+import { sleep, check, group } from 'k6';
+import { Options } from 'k6/options';
+import { req } from './requests';
+import gql from '../../e2e-tests/src/utilities/gql';
+import * as urls from '../../e2e-tests/src/utilities/urls';
+
+export const options: Options = {
+  scenarios: {
```
Mythicaeda: Reminder: Docs were requested, most importantly here a note that these scenarios run in parallel.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,282 @@
...
+      executor: 'ramping-arrival-rate',
+      exec: 'run_random_simulations',
+      maxVUs: 1000,
+      preAllocatedVUs: 50,
+      stages: [
+        { target: 5, duration: '30s' },
+        { target: 20, duration: '1m' },
+        { target: 5, duration: '30s' },
+      ],
+    }
```
Mythicaeda: These tests need to be staggered to start after `upload_plans` has started, so that they have plans to work with.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,282 @@
...
+      executor: 'ramping-arrival-rate',
+      exec: 'query_missions',
+      maxVUs: 1000,
+      preAllocatedVUs: 50,
+      stages: [
+        { target: 5, duration: '30s' },
+        { target: 10, duration: '1m' },
+        { target: 5, duration: '30s' },
+      ],
+    },
```
Mythicaeda: Reminder: This test is not meaningful, as it tests Hasura's responsiveness, which we're not responsible for.

load-tests/src/load-test.ts
Mythicaeda, line 20: Reminder: Additional Scenarios hitting the `ValidateArgs` and `GetEffectiveArgs` Endpoints were requested

Mythicaeda (changes requested):

.github/workflows/load-test.yml
skovati: To avoid collision of test runs / results, load testing was moved to a new namespaced environment. This required redefining environment variables here or creating new secrets in the repo settings, since env vars are separated per environment in GH Actions.

.github/workflows/load-test.yml
skovati: Yes it does! If the statistics fail the thresholds defined in the test options, the load test will exit nonzero, but test results should still be uploaded.

.github/workflows/load-test.yml
Mythicaeda: > To avoid collision of test runs / results, load testing was moved to a new namespaced environment.

Can you clarify this? I'm not very familiar with GH environments.

load-tests/src/load-test.ts
skovati: In that case, `random_id` will be `undefined` which we check for in the next line.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,282 @@
...
+        biteSize: 1,
+      },
+      plan_id: 1,
+      type: 'BiteBanana',
+      start_offset: random_seconds_offset + 's',
+  };
+
+  const resp = req.hasura(gql.CREATE_ACTIVITY_DIRECTIVE, {activityDirectiveInsertInput: input}, token, username);
+
+  sleep(Math.random() * 5);
```
Mythicaeda: What are these `sleep` statements for?

load-tests/src/load-test.ts
Mythicaeda: Can you alter the if statement so that if `ids` is empty, the method returns prior to setting `random_id`? That is actually what's being checked with `if(random_id)`

.github/workflows/load-test.yml
skovati: Our e2e-test results currently show up under the [`e2e-test` deployment environment results](https://github.com/NASA-AMMOS/aerie/deployments/e2e-test), since we use the `e2e-test` environment in `test.yml`. I figured namespacing the load tests into a new environment would make sense to avoid the confusion of different test results showing up in the same section.

Unfortunately, the secret env vars we defined for `e2e-test` are scoped to that specific GH environment, so I had to redefine them here.

load-tests/src/requests.ts
skovati: Good catch, this has been removed

load-tests/src/requests.ts
```diff
@@ -0,0 +1,87 @@
...
+        variables
+    });
+
+    const response = http.post(`${urls.HASURA_URL}/v1/graphql`, body, { headers });
+
+    if (response.status !== 200) {
+      console.log(response);
+    }
+
+    return response;
```
Mythicaeda: [By specification, Hasura will always return a 200, even if the request failed.](https://github.com/hasura/graphql-engine/issues/3571) I think we should be able to check the response for an `errors` field, however.

load-tests/src/load-test.ts
skovati: I've refactored this logic a bit, and I think I address your issue now.

load-tests/src/load-test.ts
skovati: Username change has been made.

load-tests/src/load-test.ts
skovati: done!

load-tests/src/load-test.ts
skovati: Added a sprinkling of inline docs, I'll write up a full README additionally.

load-tests/src/load-test.ts
skovati: I've added start time offsets to these scenarios 👍

load-tests/src/load-test.ts
skovati: I've removed this test, and replaced it with one hitting `/getActivityEffectiveArgumentsBulk`

load-tests/src/load-test.ts
skovati: I've added a `GetEffectiveArgs` test.

load-tests/src/load-test.ts
skovati: These have been removed, it was a misunderstanding of mine adding them in the first place, since the `k6` scheduler will automatically run iterations as fast as is defined in the `scenarios`.

load-tests/src/requests.ts
skovati: Good call, I am checking the `response.errors` field now.

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,273 @@
...
+      // When the executor is `ramping-arrival-rate`, the stages define the number of requests
+      // per second we want to target. The k6 scheduler will then automatically scale the number of VUs
+      // in order to hit that desired rate
+
+      // https://k6.io/docs/using-k6/scenarios/executors/
+      executor: 'ramping-arrival-rate',
+      stages: [
+        // Start 10 iterations per second for the first 30 seconds,
+        { target: 10, duration: '30s' },
+        // Then, linearly scale up to 100 iterations/second for a minute,
```
Mythicaeda: Minor, typo:
```suggestion
        // Then, linearly scale up to 50 iterations/second for a minute,
```

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,273 @@
...
+export function setup(): VUSharedData {
+
+  const user = req.login(username);
+  const token = user.token;
+
+  check(user, {
+    'user has correct name': (user) => user.id === username
+  });
+
+  sleep(1);
```
Mythicaeda: What are the sleeps in this function for?

load-tests/src/load-test.ts
Mythicaeda: I'll let you resolve this one when you feel you're finished with the docs, but thank you for the detailed inline docs!

load-tests/src/load-test.ts
Mythicaeda: I don't feel like my issue has been addressed, as my concern was how selecting the 0th element of an empty array meaning that we get an `undefined` wasn't clear, especially to primarily Java developers, where the expectation is an exception. In fact, I feel it may be worse, as the refactor now selects a property of an `undefined` object, which actually _will_ throw an exception. Additionally, `number` doesn't have properties, so there are some typing errors going on in these lines:

```ts
const data = resp.json("data");
const plan_ids: number[] = data.plan;
const random_id = plan_ids[Math.floor(Math.random() * plan_ids.length)].id;
```

An example of the errors I'm referring to can be found by attempting to run this snippet in the [Typescript Playground](https://www.typescriptlang.org/play), which is a simplification of the above three lines. 
```ts
const emptyList: number[] = []
const vary = emptyList[0].id;
console.log(vary)
```

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,273 @@
...
+        // Then scale down again over 30 seconds.
+        { target: 10, duration: '30s' },
+      ],
+    },
+
+    insert_random_activities: {
+      exec: 'insert_random_activities',
+      executor: 'ramping-arrival-rate',
+      // This scenario will only start after 10 seconds, allowing plans to populate first
+      startTime: "10s",
```
Mythicaeda: Minor, nonblocking: Should this be `5s` to match with the delay on `insert_targeted_activities`?

load-tests/.dashboard.js
Mythicaeda, line 1: Are there any docs on how to make/edit this file?

load-tests/src/load-test.ts
```diff
@@ -0,0 +1,273 @@
...
+        { target: 10, duration: '1m' },
+        { target: 5, duration: '30s' },
+      ],
+    },
+  },
+
+  // If the statistics of the load test fall below the following threshold, it with exit with a nonzero exit code
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(50)<200', 'p(95)<1000'], // 90% below 400ms, 95% of requests should be below 1000ms
```
Mythicaeda: Minor, non-blocking: Clarity:
```suggestion
    http_req_duration: ['p(50)<200', 'p(95)<1000'], // 50% of requests should take less than 200ms, 95% less than 1000ms
```

Although you force-pushed changed this threshold so that half our queries need to be below 200ms. Can you clarify why this was necessary?

load-tests/src/load-test.ts
Mythicaeda: This also needs to be done on `run_targeted_simulations`

load-tests/src/load-test.ts
skovati: These are inspired from some of the e2e-tests, where we sleep to ensure merlin has enough time to generate the mission model.

load-tests/.dashboard.js
skovati: Working on it in the readme!

load-tests/src/load-test.ts
skovati: Just trial and error finding a decent load for our github actions. These numbers can (and probably should) be changed when running heavier tests locally, but are intentionally lax for these automated tests.

Mythicaeda: > Just trial and error finding a decent load for our github actions. 

Do we get slower responses in GH actions?

> These numbers can (and probably should) be changed when running heavier tests locally, but are intentionally lax for these automated tests.

Is this statement in the upcoming README? If not, then either there or above where these thresholds are in the code would be a good place to document this expectation.

load-tests/src/load-test.ts
skovati: Done!

load-tests/src/load-test.ts
skovati: Good catch!

load-tests/src/load-test.ts
skovati: Change has been made

load-tests/src/load-test.ts
skovati: I've added checks to catch several different errors now, which will show up in our statistics, rather than placing several `if === undefined` in the test functions themselves. I.e. we _want_ to fail loud in these scenarios, not gracefully handle errors.

How is the `number[]` type cast an error?

Mythicaeda: > I've added checks to catch several different errors now, which will show up in our statistics, rather than placing several if === undefined in the test functions themselves

I believe when I wrote this comment there were `!== undefined` checks in the test functions. The current error catching is good 👍 

> How is the `number[]` type cast an error?

This bit:

```ts
const random_plan = plan_ids[...];
return random_plan.id;
```

`plan_ids` is defined as `number[]` (which isn't the type of `data.plan`), which means that `random_plan` is of type `number`. The type `number` has no properties. Therefore, the fact that the code accesses the `.id` property of`random_plan` means that it isn't actually a `number`.

skovati: Ah duh... been starting at this code too long.

Mythicaeda (approved):
LGTM!

PR #1128, opened 2023-09-05
Implement dedicated rollingwindow constraint node

### LINKED ISSUE #1015: Implement dedicated RollingWindow constraint node

JoelCourtney:
* **Tickets addressed:** closes #1015 
* **Review:** By commit  

## Description

Adds a new node `Constraint.RollingThreshold(spans, width, threshold, algorithm)` that detects when the cumulative duration of `spans` exceeds or falls short of `threshold` in any `width` interval. `algorithm` decides both whether violations are reported by highlighting the individual spans that caused the violation or by highlighting the whole interval over which it happened, and whether it looks for excess or deficit duration.

## Verification
- [x] ast and edsl compilation tests

## Documentation
- [ ] tutorial docs

I know that the constraints documentation is seriously lacking, especially in tutorials for common constraint patterns. I'll get that section started with this.

## Future work
- [x] gimmie just a minute I'm gonna add an algorithm option to require minimum cumulative duration

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
@@ -88,12 +88,10 @@ public record RollingThreshold(Expression<Spans> spans, Expression<Duration> wid
         }
       }
       if (this.algorithm == RollingThresholdAlgorithm.ExcessHull || this.algorithm == RollingThresholdAlgorithm.DeficitHull) {
-        final var hull = Interval.between(
-            violationIntervals.get(0).start,
-            violationIntervals.get(0).startInclusivity,
-            violationIntervals.get(violationIntervals.size() - 1).end,
-            violationIntervals.get(violationIntervals.size() - 1).endInclusivity
-        );
+        var hull = violationIntervals.get(0);
+        for (final var interval: violationIntervals.subList(1, violationIntervals.size())) {
+          hull = Interval.unify(hull, interval);
+        }
         violationIntervals.clear();
         violationIntervals.add(hull);
       }
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
@@ -28,13 +28,15 @@ public record RollingThreshold(Expression<Spans> spans, Expression<Duration> wid
   @Override
   public ConstraintResult evaluate(SimulationResults results, final Interval bounds, EvaluationEnvironment environment) {
     final var width = this.width.evaluate(results, bounds, environment);
-    var spans = this.spans.evaluate(results, bounds, environment);
+    final var spans = this.spans.evaluate(results, bounds, environment);
 
     final Spans reportedSpans;
     if (algorithm == RollingThresholdAlgorithm.ExcessHull || algorithm == RollingThresholdAlgorithm.ExcessSpans) {
       reportedSpans = spans;
-    } else {
+    } else if (algorithm == RollingThresholdAlgorithm.DeficitHull || algorithm == RollingThresholdAlgorithm.DeficitSpans) {
       reportedSpans = spans.intoWindows().not().intoSpans(bounds);
+    } else {
+      throw new IllegalArgumentException("Algorithm not supported: " + algorithm);
     }
 
     final var threshold = this.threshold.evaluate(results, bounds, environment);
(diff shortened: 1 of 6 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+    ExcessSpans,
+    ExcessHull,
+    DeficitSpans,
+    DeficitHull
+  }
+
+  @Override
+  public ConstraintResult evaluate(SimulationResults results, final Interval bounds, EvaluationEnvironment environment) {
+    final var width = this.width.evaluate(results, bounds, environment);
+    var spans = this.spans.evaluate(results, bounds, environment);
```
Mythicaeda: Minor: `spans` can be `final`

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -69,6 +69,48 @@ export class Constraint {
...
+   * Detect when a spans object's cumulative duration either exceeds or falls short of a threshold within any interval of a given width.
+   *
+   * Violations can be reported in various ways by setting the `algorithm` argument:
+   * - `ExcessSpans` detects times when the duration exceeds the threshold and highlights the individual spans that
+   *    contributed to the threshold violation.
+   * - `ExcessHull` detects times when the duration exceeds the threshold and highlights the whole group of spans that
+   *    contributed to the threshold violation in one interval.
+   * - `DeficitSpans` detects times when the duration falls short of the threshold and highlights the individual gaps between spans
+   *    that contributed to the threshold violation.
+   * - `ExcessHull` detects times when the duration falls short of the threshold and highlights the whole group of gaps between
```
Mythicaeda: Typo here and on L1111:
```suggestion
   * - `DeficitHull` detects times when the duration falls short of the threshold and highlights the whole group of gaps between
```

Mythicaeda: Additionally can you remind me why this code needs to be written twice, once here and once in the `declare global` block?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+
+    final Spans reportedSpans;
+    if (algorithm == RollingThresholdAlgorithm.ExcessHull || algorithm == RollingThresholdAlgorithm.ExcessSpans) {
+      reportedSpans = spans;
+    } else {
+      reportedSpans = spans.intoWindows().not().intoSpans(bounds);
+    }
```
Mythicaeda: Are there more possible RollingWindows Algorithms? If so, I think a switch where the `default` case throws like what is done elsewhere in the codebase would make sense here and on L59

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+        Interval.FOREVER,
+        new LinearEquation(
+            Duration.ZERO,
+            1,
+            0
+        )
+    ));
+
+    if (algorithm == RollingThresholdAlgorithm.ExcessHull || algorithm == RollingThresholdAlgorithm.ExcessSpans) {
+      leftViolatingBounds = localAccDuration.greaterThan(thresholdEq);
```
Mythicaeda, line 62: Why isn't the `select` part needed in the `greaterThan` case?

JoelCourtney: This is because the end of every plan will drop below the minimum of a `deficit` thresholds when the interval starts to move off the end of the plan (because of course there are no more activities / spans). This isn't an issue for `excess` thresholds because this causes an automatic success instead of automatic violation.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/RollingThreshold.java
JoelCourtney: Yes, there probably are. I've added an else block that catches this.

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
JoelCourtney: The first declaration (and implementation) is needed to get the class objects added to `globalThis` in javascript. This allows only the class values to be accessible without importing. `declare global` is needed to get the class types accessible in the constraint code without importing.

The doc comments are duplicated because monaco uses docs from the global declaration for mouse-over hints, and typedoc uses the normal implementation for generating docs.

Mythicaeda (approved):

PR #1130, opened 2023-09-05
Replace Column List with '*' in Hasura Permissions

### LINKED ISSUE #1119: Replace Column List with '*'
While writing docs, [I discovered that Hasura now supports '*' for column lists when specified in the .YAMLs](https://hasura.io/docs/latest/api-reference/syntax-defs/#insertpermission)! I've messed around with this some and it does appear to work as expected, with giving the role access to all columns at the present moment and that are added/renamed/deleted. Our select permissions should be updated to use '*' instead of a column list where we allow for selection of all columns

This will not help in insert or update permissions, where we generally don't want to blindly give roles access to all columns.

Mythicaeda:
* **Tickets addressed:** Closes #1119.
* **Review:** By commit  

## Description

Replaces all select permissions where a role may select all columns with `'*'` rather than explicitly listing the column list. AKA all select permissions except for `viewer` and `user` on the `users` table.

Additionally, `merge_request` was missing a delete permission for `aerie_admin`. The role may now delete requests that are not in the middle of an active merge.

## Verification

Brought up Hasura and verified that the permissions summaries for all DBs and their schemas had not change

### DIFF
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/hasura_functions/begin_merge_return_value.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/hasura_functions/begin_merge_return_value.yaml
@@ -4,9 +4,9 @@ table:
 select_permissions:
   - role: aerie_admin
     permission:
-      columns: [merge_request_id, non_conflicting_activities, conflicting_activities]
+      columns: '*'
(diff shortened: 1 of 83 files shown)


camargo (approved):
Nice refactor. LGTM!

PR #1134, opened 2023-09-06
Remove tags trigger from GH pages workflow

camargo:
* GH does not allow deploying from tags for environments with branch protection rules that match the tag
* https://github.com/orgs/community/discussions/39054

For example the following `v*` rule used to work for tags, but does not anymore (for the `github-pages` environment rules): 

Screenshot 2023-09-06 at 4 05 44 PM

### DIFF
--- a/.github/workflows/deploy-to-gh-pages.yml
+++ b/.github/workflows/deploy-to-gh-pages.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - develop
-    tags:
-      - v*
   workflow_dispatch:
 
 permissions:


skovati (approved):
Classic GH...

skovati:
This reminds me, it would be nice to update our user docs and javadocs to be versioned with releases.

camargo:
@skovati Yeah agree. The user docs are pretty straightforward via [Docusaurus](https://docusaurus.io/docs/versioning). 
Not sure if Javadoc has a similar scheme to tag versions. Might need a custom workflow?

Mythicaeda:
Didn't we decide not to version with Docasaurus because it would require us to commit every version to the docs repo? Additionally, I don't think the API/UI changes frequently enough to warrant versioning per release. IE: the docs in `Planning` were written a while back, but they're still accurate, plus or minus mentioning new features (Activity Changelog, Creating and Restoring Snapshots). Additionally, there are docs improvements that can be made that would be useful to users on older Aerie versions (ie, providing more Constraints examples).

Alternative Proposal: We add admonitions at the top of new pages declaring when a feature was introduced. If the API changes, we update the docs and add an admonition that's "For versions prior to X: (original query)". We could then version the docs per major release (either Quarterly or v1.x, v2.x like how Hasura does it) and drop these admonitions on the "current" version.

> Not sure if Javadoc has a similar scheme to tag versions. Might need a custom workflow?

There are `@version` and `@since` tags, but if you mean versioning our entire docs, the Javadoc would need a custom workflow, from what I remember from the Scylla days. I _believe_ we used to use an extension of our TS Docs builder to version the EDSL docs.

camargo:
So we should version the user docs, but probably not for every version (think maybe the quarterly or even yearly boundary - like for major MGSS releases). But I 100% agree, we should try and do it without committing to the docs repo. My hunch is this could be done in the GH pages workflow in the docs repo, but it will require more investigation.

camargo:
@skovati is going to make a ticket. Let's take the discussion there.

PR #1136, opened 2023-09-07
Fix signature of `create_snapshot`

Mythicaeda:
* **Review:** By commit  

## Description

The initial signature of `create_snapshot` (_plan_id, _user, _snapshot_name), had `_user` and `_snapshot_name` in the opposite order to how the pair is written elsewhere in the database, including in the function itself. This caused a bug in the Hasura Function due to it supplying the parameters in the order `(_plan_id, _snapshot_name, _user)`. Since the underlying function's signature was off, I chose to fix that rather than fix its callsites.

## Verification

Three new DB tests were created in `PlanSnapshotTests`.


camargo (approved):
LGTM!

PR #1137, opened 2023-09-12
Array/List parameters are working again in scheduling

### LINKED ISSUE #981: Incomplete Activity Instances Scheduled when Optional List or Map Parameters Exist in Activity Type
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

1.6.0

### Describe the bug

**Description** : This bug shows up when scheduling activity types that have optional list or map parameters. After running a scheduling goal that places an activity instance of this type, only the first object in the parameter list is included in the activity instance when it is scheduled (even though all of the objects/their parameters in the goal are specified in the scheduling goal). 

An example of this is a Clipper specific goal EIS_NAC_Geometric_Cal with parameter "targetList "; a list of objects with each object composed of 11 parameters specifying the data pertaining to a target. I tried manually setting up the activity with multiple (3) target objects in the list and simulating and the activity works as expected, confirming this is an aerie scheduling bug.

### Reproduction

This bug shows up for activity types that have parameter lists or maps with optional elements. The problem is recursive too, so if an element in the list is a map, only the first element of this object appears as well. Testing scheduling goals that place activity types with optional list and/or map parameters (or an optional map nested in a list element) would be a good way to replicate this issue. 

Below, several Clipper-specific examples are provided for more context. The scheduling goals used for these examples were simple cardinality goals placing one occurrence of each activity type in the plan region >2AU away from the sun (utilizing the applyWhen method). 

**Example 1:** For the EIS_NAC_Geometric_Cal goal, the _expected_ post-scheduling parameter list for the EIS_NAC_Geometric_Cal activity instance scheduled in the plan is: 
eis-nac-geom-cal-expected
With the first target object's parameters expanded:
eis-nac-geom-cal-expected-expanded

The below is what was _actually_ populated in the plan after running the scheduling goal: 
eis-nac-geom-cal-actual
You can see that there is (1) only the first target element (out of 3) in the "targetList" scheduled and (2) only the first element of the "binRatios" map composite object type scheduled (out of 5).

**Example 2:** For the scheduling goal SUDA_Solar_Noise_Cal, the _expected_ post-scheduling parameter list for the SUDA_Solar_Noise_Cal activity instance scheduled in the plan is: 
suda-solar-noise-cal-expected

_Actual_ activity instance parameters after running the scheduling goal:
suda-solar-noise-cal-actual
You can observe that the first element of the list is the only one populated (out of 3).

### Logs

_No response_

### System Info

```shell
Google Chrome browser, macOS Ventura 13.3.1
```

### Severity

Moderate

adrienmaillard:
* **Tickets addressed:** Closes #981 
* **Review:** By commit  

## Description

An error in the algorithm transforming every user-provided typescript value into a profile was causing the transformation to stop after the first element of the array. 

## Verification

I added a test that was not passing before the fix was made.

## Documentation

None, it is a bug.

## Future work

None.

### DIFF
--- a/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
+++ b/scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/TypescriptCodeGenerationService.java
@@ -80,7 +80,7 @@ export function makeAllDiscreteProfile (argument: any) : any{
      }
   else if(Array.isArray(argument)){
    const arr: any[] = [];
-   argument.every((element) => { arr.push(makeAllDiscreteProfile(element))});
+   argument.forEach((element) => { arr.push(makeAllDiscreteProfile(element))});
    return Discrete.List(arr).__astNode;
  } else if (typeof argument === "object" ){
    const obj: { [k: string]: any } = {};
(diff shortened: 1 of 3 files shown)


Mythicaeda (approved):
🎉

PR #1146, opened 2023-09-14
Add line of code count to CI

### LINKED ISSUE #518: Add source lines of code count to run in GH Actions
We should run the count lines of code ([CLOC](https://github.com/AlDanial/cloc)) program in one of our GitHub Actions workflows to get a good idea of relative project code size. For reference the UI [does it here](https://github.com/NASA-AMMOS/aerie-ui/blob/develop/.github/workflows/lint.yml#L28). We should configure CLOC to only count developer written code and not dependencies.

skovati:
* **Tickets addressed:** closes #518 
* **Review:** By commit  

## Description

Uses [cloc](https://github.com/AlDanial/cloc) to count lines of code per module in Aerie. An `by-percent` option was added to `cloc` to output comment and blank as a percentage of total lines.

## Future work

`cloc` can output to several data formats (csv, json, sql), so we can add this additional output if we require further analysis.

### DIFF
--- a/.github/workflows/cloc.yml
+++ b/.github/workflows/cloc.yml
@@ -1,0 +1,33 @@
+name: Count Lines of Code
+
+on:
+  pull_request:
+    branches:
+      - develop
+  push:
+    branches:
+      - develop
+    tags:
+      - v*
+
+jobs:
+  cloc:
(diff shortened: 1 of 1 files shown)


camargo (approved):
LGTM

PR #1140, opened 2023-09-13
Validate User Role Permissions JSONs

Mythicaeda:
* **Tickets addressed:** Closes #990 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

A trigger was added to the `user_role_permission` table that checks if all the keys and permissions are actual Keys and Permissions. For Permissions, there is an additional check that Plan Merge Permissions are only applied to Plan Merge Functions.

Additionally, we were creating all the preloaded users/user-roles across several files. They've been moved to a single file: `default_user_roles.sql`

## Verification

New Permissions DB tests were added

## Documentation

No documentation changes are necessary

PR #1147, opened 2023-09-14
Add Snapshot Tags, Description

### LINKED ISSUE #1143: Add a description and tags to Snapshots
Plan Snapshots needs a `description` text field and a tags association table
Also, Plan needs a `description` text field

Mythicaeda:
* **Tickets addressed:** Closes #1143 
* **Review:** By commit  

## Description

Adds new metadata to Plan Snapshots: Tags, Description
Adds new metadata to Plan: Description

Adds Hasura relationship from Snapshots to their Activities and Tags

## Verification

Spun up Aerie and checked that the metadata was correct in Hasura.

## Future work

[The UI has a ticket for allowing Plan Metadata to be updated](https://github.com/NASA-AMMOS/aerie-ui/issues/883). The description column needs to be added to that list.

### DIFF
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
@@ -1,0 +1,48 @@
+table:
+  name: plan_snapshot_tags
+  schema: metadata
+configuration:
+  custom_name: "plan_snapshot_tags"
+object_relationships:
+  - name: plan_snapshot
+    using:
+      foreign_key_constraint_on: snapshot_id
+  - name: tag
+    using:
+      foreign_key_constraint_on: tag_id
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: user
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [snapshot_id, tag_id]
+      check: {}
+  - role: user
+    permission:
+      columns: [snapshot_id, tag_id]
+      check: {"plan_snapshot":{"plan":{"_or":[
+        {"owner":{"_eq":"X-Hasura-User-Id"}},
+        {"collaborators":{"collaborator":{"_eq":"X-Hasura-User-Id"}}}]}}}
+delete_permissions:
+  - role: aerie_admin
+    permission:
+      filter: {}
+  - role: user
+    permission:
+      filter: {"plan_snapshot":{"plan":{"_or":[
+        {"owner":{"_eq":"X-Hasura-User-Id"}},
+        {"collaborators":{"collaborator":{"_eq":"X-Hasura-User-Id"}}}]}}}
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_snapshot.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_snapshot.yaml
@@ -11,6 +11,21 @@ object_relationships:
       remote_table:
         name: plan
         schema: public
+array_relationships:
+  - name: activities
+    using:
+      foreign_key_constraint_on:
+        column: snapshot_id
+        table:
+          name: plan_snapshot_activities
+          schema: public
+  - name: tags
+    using:
+      foreign_key_constraint_on:
+        column: snapshot_id
+        table:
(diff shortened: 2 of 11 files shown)


merlin-server/sql/merlin/applied_migrations.sql
```diff
@@ -29,3 +29,4 @@ call migrations.mark_migration_applied('23');
 call migrations.mark_migration_applied('24');
 call migrations.mark_migration_applied('25');
 call migrations.mark_migration_applied('26');
+call migrations.mark_migration_applied('28');
```
skovati: Keep an eye on this migration ID, if this PR gets merged before #1140

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
```diff
@@ -0,0 +1,48 @@
...
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [snapshot_id, tag_id]
+      check: {}
+  - role: user
+    permission:
+      columns: [snapshot_id, tag_id]
```
skovati, line 36: Are we explicitly listing all columns instead of using `"*"` here in anticipation of other columns being added?

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
```diff
@@ -0,0 +1,48 @@
...
+      check: {"plan_snapshot":{"plan":{"_or":[
+        {"owner":{"_eq":"X-Hasura-User-Id"}},
+        {"collaborators":{"collaborator":{"_eq":"X-Hasura-User-Id"}}}]}}}
```
skovati, line 39: Update permissions aren't applicable, since we have to delete and recreate rows on tag change correct?

deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_snapshot.yaml
```diff
@@ -27,6 +42,17 @@ select_permissions:
       columns: '*'
       filter: {}
       allow_aggregations: true
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [description, taken_by]
```
skovati, line 48: What would be the use case for updating the `taken_by` column for a snapshot?

skovati (approved):
Looks good to me! Just some non-blocking questions / comments. Nice work 🚀

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
Mythicaeda: Technically, however on this table specifically it was mainly done in order to maintain a consistent style across all of the insert/update permissions in our metadata (that style being that insert/update permissions are **explicitly** granted rather than **implicitly** given via the `"*"`). This consistency is especially important as we expect users to actually view these files and use them as reference when making their own roles.

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
Mythicaeda: Can you clarify what you mean by "on tag change"?

Update Permissions aren't applicable b/c this is an association table. You can't "update" an association, it either exists or it doesn't. (And if by some miracle the ids are ever updated, there's a `cascade` on the underlying table)

deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan_snapshot.yaml
Mythicaeda: Weird username scenario, likely. In general, however, the `aerie_admin` is permitted to update `taken_by` for consistency, as they are allowed to update `owner`/`author`/`requested_by` columns everywhere else. Which, if you're wondering, it can update because it's the admin role and there is not a reason to prevent them from updating that column.

merlin-server/sql/merlin/applied_migrations.sql
Mythicaeda: Yeah, I was expecting the merge order to be 1140 -> this one, but it looks like I'll need to reverse the migration numbering 😔.

deployment/hasura/metadata/databases/AerieMerlin/tables/metadata/plan_snapshot_tags.yaml
skovati: Makes sense. I just wanted to double check that updating these rows wasn't a valid use case, i.e. atomically switching from one tag id to another _could_ be an update, but instead we have the client delete the existing association, and then create a new association.

PR #1152, opened 2023-09-25
Select windows by index 

adrienmaillard:
* **Review:** By commit  

## Description

This allows to select windows by their index in the sequence.
Example: `Real.Resource("state of charge").lessThan(0.3).starts().keepTrueSegment(2)`  will select the second window of `Real.Resource("state of charge").lessThan(0.3).starts()`. If it does not exist, it will return nothing.

If the user provides a negative index, the index starts from the last element (-1). 

## Verification

Some new tests.

## Documentation

Has to be added to constraints documentation.

## Future work

I want to add `before(Windows)` and `after(Windows)` operators to support more sequence operations.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -280,6 +280,15 @@ public final class ConstraintParsers {
               untuple((kind, alias) -> new StartOf(alias)),
               $ -> tuple(Unit.UNIT, $.activityAlias));
 
+  static JsonParser<KeepTrueSegment> keepTrueSegmentP(JsonParser<Expression<Windows>> windowsExpressionP) {
+    return productP
+        .field("kind", literalP("WindowsExpressionKeepTrueSegment"))
+        .field("expression", windowsExpressionP)
+        .field("index", intP)
+        .map(
+            untuple((kind, expression, index) -> new KeepTrueSegment(expression, index)),
+            $ -> tuple(Unit.UNIT, $.expression, $.i));
+  }
   static final JsonParser<Duration> durationP =
       longP
           .map(
@@ -580,7 +589,8 @@ public final class ConstraintParsers {
         windowsFromSpansF(spansP),
         activityWindowP,
         assignGapsF(selfP),
-        shiftByF(selfP)
+        shiftByF(selfP),
+        keepTrueSegmentP(selfP)
     ));
   }
 
--- a/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
+++ b/merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
@@ -164,6 +164,18 @@ export class Windows {
   }
 
   /**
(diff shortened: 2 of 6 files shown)


JoelCourtney:
Cool! The java Windows class already has a `keepTrueSegment` function which should have the same behavior. The AST node should only have to delegate to that function.

I'd prefer if it was named `keepTrueSegment` or something similarly descriptive, because to me `index` suggests that it selects the n'th segment, not the n'th `true` segment. I know most users don't think that way because much of the API prioritizes `true` segments, but we are going to have to be moving the mental model in that direction for procedural constraints anyway.

adrienmaillard:
> Cool! The java Windows class already has a `keepTrueSegment` function which should have the same behavior. The AST node should only have to delegate to that function.
> 
> I'd prefer if it was named `keepTrueSegment` or something similarly descriptive, because to me `index` suggests that it selects the n'th segment, not the n'th `true` segment. I know most users don't think that way because much of the API prioritizes `true` segments, but we are going to have to be moving the mental model in that direction for procedural constraints anyway.

Oh you're right, I was not aware. I'll replace the ad-hoc code with a call to `keepTrueSegment`. For the name, the closest I can agree to is `getTrueSegment`. `keepTrueSegment` works for getting a set of intervals but its meaning is odd if it is supposed to return only one window. Other ideas ?

adrienmaillard:
Actually, I like `keepTrueSegment` as it reflects better what's done. I have made the changes. If you want to review, it's ready ;)

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
```diff
@@ -163,6 +163,18 @@ export class Windows {
     });
   }
 
+  /**
+   * Selects the ith true window
```
JoelCourtney: Please add a bit that says it falsifies all the other true segments. Otherwise its unclear, could be interpreted as deleting the other segments.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
```diff
@@ -280,6 +280,15 @@ static JsonParser<SpansSelectWhenTrue> spansSelectWhenTrueF(JsonParser<Expressio
               untuple((kind, alias) -> new StartOf(alias)),
               $ -> tuple(Unit.UNIT, $.activityAlias));
 
+  static JsonParser<KeepTrueSegment> indexOfP(JsonParser<Expression<Windows>> windowsExpressionP) {
```
JoelCourtney: please rename this to `keepTrueSegmentP`

merlin-server/constraints-dsl-compiler/src/libs/constraints-edsl-fluent-api.ts
adrienmaillard: done

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
adrienmaillard: done

JoelCourtney (approved):

Issue #1057, opened 2023-08-02
Investigate Using Playwright's Java Interface for E2E Tests 


Mythicaeda:
[Playwright has a Java interface](https://playwright.dev/java/docs/intro). Given that the rest of the repo's tests and code is in Java (excluding the EDSL/Sequencing Server), it is worth investigating if it would be easier to write our Playwright tests in Java as well.

PR #1156, opened 2023-09-27
Standardize the cached Constraint results to match action results

### LINKED ISSUE #1150: Schema Mismatch Between Cached Constraint Run and Action Response
The cached results in the `constraint_run` table stores the names of the used resources under "resourceNames". The response from the Constraints action stores them under "resourceIds"

Additionally in the `violations`part of results, the violating windows are stored as `windows` in the action response and as `violationWindows` in the cached results

Also, `type` in the response vs `constraintType` in the cache

Mythicaeda:
* **Tickets addressed:** Closes #1150
* **Review:** By commit  

## Description

Updated the parser for constraint violations so that the cached run has the same key names as the action response.

## Verification

Wrote a JSON parser while working on #1057 that successfully parses both the `checkConstraints` action response and the cached values from `constraint_run`.

## Documentation

No docs were invalidated

## Future work

None

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/json/ConstraintParsers.java
@@ -299,7 +299,7 @@ public final class ConstraintParsers {
 
   public static final JsonParser<Violation> violationP =
       productP
-          .field("violationWindows", listP(intervalP))
+          .field("windows", listP(intervalP))
           .field("activityInstanceIds", listP(longP))
           .map(
               untuple(Violation::new),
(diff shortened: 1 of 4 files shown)


JoelCourtney:
@jeffpamer FYI this will impact your UI code

JoelCourtney (approved):

JoelCourtney:
Don't merge until the UI has a PR to keep up to date

jeffpamer:
Looks like this should be a pretty small change on our end if I'm interpreting this correctly. I'll try to get a UI PR up this afternoon or first thing tomorrow.

jeffpamer:
Upon further investigation this does not appear to be a breaking change for the UI as we are currently not reading anything out of `constraint_run`. If we do start using the cached run results in the future, it looks like after this change the schema will match the shape we are already expecting for constraint check results.

PR #1157, opened 2023-09-27
Fix Scheduling DSL Typescript Action

Mythicaeda:
* **Review:** By commit  

## Description

Two bugs were fixed:
1. The optional `planId` parameter for `schedulingDSLTypscript` was added to the Action definition. Because this parameter is optional, this change is not breaking (verified manually).
2. The parser has been fixed to properly handle `planId` not being specified. With fix (1), it was unable to parse a `null` `planId`.

## Verification

Re: 1: The GQL query has been updated with the action.

Re: 2: We already have two tests for that endpoint, one which provides the `planId` and one which doesn't. The reason this slipped by is the one that provides the `planId` was a Bindings Test, meaning it was hitting the scheduler directly instead of using a GQL query. I have corrected this in my work for #1057 by splitting the tests into two tests, a BindingsTest and a ScheudlingDSLTest. 

## Documentation

No docs need to be updated.

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -89,7 +89,7 @@ type Query {
 }
 
 type Query {
-  schedulingDslTypescript(missionModelId: Int!): DslTypescriptResponse
+  schedulingDslTypescript(missionModelId: Int!, planId: Int): DslTypescriptResponse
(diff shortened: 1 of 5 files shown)


adrienmaillard (approved):
Thank you for the fix

Issue #1087, opened 2023-08-10
Fix scheduler/simulation discrepancies about activity representation as intervals


adrienmaillard:
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

develop

### Describe the bug

In Aerie simulation, there is no order between activities being simulated at the same time except with anchoring. In simulation, an activity can act at both its time boundaries. They can all be potentially considered closed-closed intervals. Let's take A and B such that B starts at the exact same time as A is ending.

<img src="https://github.com/NASA-AMMOS/aerie/assets/94930916/002e4054-d281-4926-83fa-1c9e9efdcc99" width=40% height=40%>

There are two issues : 
- The scheduler assumes that activities can be represented by closed-open intervals. They should be represented with closed-closed intervals. 
- The second part has to do with `CoexistenceGoal`:
Currently, when we use `startsAt` or `endsAt` temporal constraints for placing coexisting activities, we make them start/end at one of the anchor time bounds which result in the above undefined behavior in simulation. In the above AB example the last instructions of A (after its last `delay()`) could execute after the first instructions of B (before its first `delay()`). Let's say that A depletes a battery and that B checks the level of the battery to act accordingly, this could lead to some unwanted/inaccurate behavior.

To fix that in `CoexistenceGoal` there are several possible paths that are currently being discussed:
- Either we add `startsJustAfter(TimeAnchor)` `endsJustBefore(TimeAnchor)` operators that makes start/end activities at bounds+1 microseconds, bounds-1 microseconds etc.
- We modify `startsAt`, `endsAt` behavior so that coexistent activities start/end at bounds+1 microseconds, bounds-1 microseconds etc. 
- We use anchors. If A is present and B has to be placed at the end of A, anchor B to the end of A. But if B is present and A has to finish at the start of A, the only way of using anchor would be to change the anchoring of B. The end of A cannot be anchored in the future. 
- Enforce order depending on which activity started first.
- Inject `delay(0)` at either start or end of activity effect model. Some discussion points below:
  - similar to what APGEN does
  - What would be the effects on decomposition ? Could the delay only apply to top-level activities ? 
  - That would change the semantics of spawning: the first child would see concurrent effect which was intended not to be the case 
  - Do we really need concurrency ? APGEN seems to work with total order 

### Reproduction

No reproduction.

### System Info

```shell
Not relevant
```

### Severity

Moderate

jmdelfa:
[...]

PR #1159, opened 2023-09-28
Add justAfter and justBefore operators to scheduling eDSL

adrienmaillard:
* **Tickets addressed:** #1087 partially
* **Review:** By commit  

## Description

- First commit: add `justAfter` and `justBefore` timing constraints in the scheduling edsl. The TimingConstraint.defaultPadding defines the default padding applied, 1 microsecond per default. The user can change it in the goal code.
- Second commit: Activities are now represented by closed-closed intervals. 
- The third commit is refactoring (collapsing) the `ActivityExpression` hierarchy 

## Verification

A few tests have been updated because of the first second commit. 

## Documentation

- [ ] Add documentation to aerie-docs

## Future work

None.

### DIFF
--- a/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
+++ b/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
@@ -901,7 +901,8 @@ public class TestApplyWhen {
 
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(1, Duration.SECONDS), actTypeA));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(14, Duration.SECONDS), actTypeA));
-    assertEquals(3, problem.getSimulationFacade().countSimulationRestarts());
+    assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(12, Duration.SECONDS), actTypeA));
+    assertEquals(4, problem.getSimulationFacade().countSimulationRestarts());
   }
 
   @Test
@@ -971,12 +972,13 @@ public class TestApplyWhen {
     for(SchedulingActivityDirective a : plan.get().getActivitiesByTime()){
       logger.debug(a.startOffset().toString() + ", " + a.duration().toString());
     }
-
+    assertEquals(10, plan.get().getActivitiesById().size());
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(2, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(10, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(16, Duration.SECONDS), actTypeB));
+    assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(23, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(25, Duration.SECONDS), actTypeB));
-    assertEquals(5, problem.getSimulationFacade().countSimulationRestarts());
+    assertEquals(6, problem.getSimulationFacade().countSimulationRestarts());
   }
 
   @Test
--- a/scheduler-worker/build.gradle
+++ b/scheduler-worker/build.gradle
@@ -126,7 +126,6 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
   testImplementation 'org.junit.jupiter:junit-jupiter-params:5.10.0'
   testImplementation 'org.assertj:assertj-core:3.24.2'
-  testImplementation 'junit:junit:4.13.2'
   testImplementation 'javax.json.bind:javax.json.bind-api:1.0'
   testImplementation 'org.glassfish:javax.json:1.1.4'
 }
--- a/scheduler-worker/build.gradle
+++ b/scheduler-worker/build.gradle
@@ -124,6 +124,7 @@ dependencies {
   testImplementation project(':examples:foo-missionmodel')
   testImplementation testFixtures(project(':scheduler-server'))
   testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
+  testImplementation 'org.junit.jupiter:junit-jupiter-params:5.10.0'
   testImplementation 'org.assertj:assertj-core:3.24.2'
   testImplementation 'junit:junit:4.13.2'
   testImplementation 'javax.json.bind:javax.json.bind-api:1.0'
--- a/scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
+++ b/scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
@@ -16,14 +16,11 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.HOUR;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.HOURS;
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECOND;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MILLISECONDS;
-import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTE;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTES;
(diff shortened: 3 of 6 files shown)


scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -2744,7 +2746,128 @@ export default () => Goal.CoexistenceGoal({
...
+        Map.of(
+            new ActivityDirectiveId(2L),
+            new ActivityDirective(
+                tenMinutes,
+                "GrowBanana",
+                Map.of(
+                    "quantity", SerializedValue.of(1),
+                    "growingDuration", SerializedValue.of(activityDuration.in(Duration.MICROSECONDS))),
+                new ActivityDirectiveId(1L),
+                true)),
```
Mythicaeda: Where is the `ActivityDirective` with an ID of `1L` defined? And if it isn't, then this test should be failing, since simulation can't simulate this activity. This applies to `testJustBefore` as well.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -19,6 +19,8 @@
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.HOUR;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.HOURS;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECOND;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MILLISECONDS;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTE;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MINUTES;
```
Mythicaeda: Minor, can ignore: can these imports be cleaned up so that they're only using either the plural or singular variant?

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
```diff
@@ -2744,7 +2746,128 @@ export default () => Goal.CoexistenceGoal({
     assertEquals(SerializedValue.of("fromStem"), peelBanana.serializedActivity().getArguments().get("peelDirection"));
   }
 
-    /**
+  @Test
+  void testJustAfter() {
```
Mythicaeda: Minor: Can there be two additional tests added that uses `WindowProperty.END` instead of `WindowProperty.START`? I believe you may even be able to do this by converting these tests into [parametrized tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests), rather than creating two new tests.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
```diff
@@ -971,12 +971,13 @@ public void testCoexistenceWindowsCutoffMidActivity() {
...
     }
-
+    assertEquals(10, plan.get().getActivitiesById().size());
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(2, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(10, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(16, Duration.SECONDS), actTypeB));
+    assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(23, Duration.SECONDS), actTypeB));
     assertTrue(TestUtility.activityStartingAtTime(plan.get(), Duration.of(25, Duration.SECONDS), actTypeB));
-    assertEquals(5, problem.getSimulationFacade().countSimulationRestarts());
+    assertEquals(6, problem.getSimulationFacade().countSimulationRestarts());
```
Mythicaeda, line 981: Why did using a closed interval for activities cause the number of sim restarts to increase? This question also applies to L904

Mythicaeda (commented):
Can Commit 3 be moved into its own PR? The work is most of this PR and is an unrelated refactor.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Nowhere. It simulates because the activity is not found in the greatly named `planActDirectiveIdToSimulationActivityDirectiveId` and thus its anchor id is null. I'll remove this anchor, it makes no sense here anyway.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Sure.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Sounds good.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/TestApplyWhen.java
adrienmaillard: Because as you can see, there is also a new activity created in both of these tests. That is because activity ending at 19+3 ends now into the [23,28] interval and a coexisting activity can be created. 
Same explanation for the 8+4 anchor activity falling in the [12...] interval in the other test.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Done

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
Mythicaeda: Does this meant that the scheduler alters the directives it is given? Or is removing the anchor on this directive a test-only oddity?

scheduler-worker/build.gradle
```diff
@@ -124,6 +124,7 @@ dependencies {
   testImplementation project(':examples:foo-missionmodel')
   testImplementation testFixtures(project(':scheduler-server'))
   testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
+  testImplementation 'org.junit.jupiter:junit-jupiter-params:5.10.0'
   testImplementation 'org.assertj:assertj-core:3.24.2'
   testImplementation 'junit:junit:4.13.2'
```
Mythicaeda: Minor, can ignore (might not even belong in this PR), but we actually don't need to be importing both JUnit 4 and JUnit 5.
```suggestion
  testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
  testImplementation 'org.junit.jupiter:junit-jupiter-params:5.10.0'
  testImplementation 'org.assertj:assertj-core:3.24.2'
```

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: no it does not. When looking for the corresponding activity anchor in simulation (in the map), it does not find it (as it does not exist).

Mythicaeda: Alright, I've traced the bug down to L335 of the facade. If the anchor activity does not exist in the map, we ought to throw, as the plan data is scuffed.

Mythicaeda (approved):
LGTM!

scheduler-worker/build.gradle
adrienmaillard: I have removed this dependency as the last commit.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingIntegrationTests.java
adrienmaillard: Added this bit 
```java
if(activity.anchorId()!= null && !planActDirectiveIdToSimulationActivityDirectiveId.containsKey(activity.anchorId())){
      throw new RuntimeException("Activity with id "+ activity.anchorId() + " referenced as an anchor by activity " + activity.toString() + " is not present in the plan");
}
```

PR #1169, opened 2023-09-29
Make Check Constraints Load Correct Sim Dataset

### LINKED ISSUE #1168: Check Constraints Action Does not used Specified Simulation Dataset
Currently, the check constraints action always uses the latest simulation results, even if an earlier sim dataset is requested. This occurs because it uses the resultsHandler object, which has fetched the most recent results in case `simDatasetId` wasn't specified, as the simulation results to check against.

Mythicaeda:
* **Tickets addressed:** Closes #1168.
* **Review:** By commit  

## Description

Resolves the bug by allowing the Simulation Service to look up either a specific SimulationDataset, or the latest one if one isn't specified.

## Verification

An e2e Test has been added to validate that constraints loads the specified Simulation Results, even if those results are outdated. It was placed in its own file to avoid any potential caching issues with the existing constraints tests.

A bindings tests has been added to validate the error response if an invalid sim dataset id is passed.

## Documentation

No docs need to be updated

## Future work

The `UncachedSimulationService` is not only unused, it's not representative of our system, which now generally expects results to have been cached. As a result, I'd like to flatten this interface.

### DIFF
--- a/e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
+++ b/e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
@@ -1,0 +1,108 @@
+import { expect, test } from '@playwright/test';
+import req, { awaitSimulation } from '../utilities/requests.js';
+
+test.describe.serial('Check Constraints Against Specific Sim Datasets', () => {
+    const constraintName = 'fruit_equal_peel';
+    const activity_offset_hours = 1;
+
+    const plan_start_timestamp = "2023-01-01T00:00:00+00:00";
+
+    let mission_model_id: number;
+    let plan_id: number;
+    let constraint_id: number;
+    let violation: Violation;
+    let activity_id: number;
+
+    test.beforeAll(async ({request}) => {
+        let rd = Math.random() * 100000;
+        let jar_id = await req.uploadJarFile(request);
+        // Add Mission Model
+        const model: MissionModelInsertInput = {
+            jar_id,
+            mission: 'aerie_e2e_tests',
+            name: 'Banananation (e2e tests)',
+            version: rd + "",
+        };
+        mission_model_id = await req.createMissionModel(request, model);
+
+        // Add Plan
+        const plan_input: CreatePlanInput = {
+            model_id: mission_model_id,
+            name: 'test_plan' + rd,
+            start_time: plan_start_timestamp,
+            duration: "24:00:00"
+        };
+        plan_id = await req.createPlan(request, plan_input);
+
+        // Add Activity
+        const activityToInsert: ActivityInsertInput = {
+            arguments: {
+                biteSize: 2,
+            },
+            plan_id: plan_id,
+            type: 'BiteBanana',
+            start_offset: activity_offset_hours + 'h',
+        };
+        activity_id = await req.insertActivity(request, activityToInsert);
+
+        // Add Constraint
+        const constraint: ConstraintInsertInput = {
+            name: constraintName,
+            definition: 'export default (): Constraint => Real.Resource("/fruit").equal(Real.Resource("/peel"))',
+            description: '',
+            model_id: null,
+            plan_id,
+        };
+        constraint_id = await req.insertConstraint(request, constraint);
+    });
+
+    test.afterAll(async ({request}) => {
+        // Delete Constraint
+        await req.deleteConstraint(request, constraint_id);
+
+        // Deleting Plan and Model cascades the rest of the cleanup
+        await req.deleteMissionModel(request, mission_model_id);
+        await req.deletePlan(request, plan_id);
+    });
+
+    test('Constraints Loads Specified Outdated Sim Dataset', async ({request}) => {
+        // Simulate
+        const oldSimDatasetId = (await awaitSimulation(request, plan_id)).simulationDatasetId;
+
+        // Invalidate Results and Resim
+        await req.deleteActivity(request, plan_id, activity_id);
+        const newSimulationDatasetId = (await awaitSimulation(request, plan_id)).simulationDatasetId;
+
+        // Assert No Violations on Newest Results
+        const newConstraintResults = await req.checkConstraints(request, plan_id, newSimulationDatasetId);
+        expect(newConstraintResults).toHaveLength(0);
+
+        // Assert One Violation on Old Results
+        const oldConstraintResults = await req.checkConstraints(request, plan_id, oldSimDatasetId);
+        expect(oldConstraintResults).toHaveLength(1);
+
+        // Assert the Results to be as expected
+        const result = oldConstraintResults.pop();
+        expect(result.constraintName).toEqual(constraintName);
+        expect(result.constraintId).toEqual(constraint_id);
+
+        // Resources
+        expect(result.resourceIds).toHaveLength(2);
+        expect(result.resourceIds).toContain('/fruit');
+        expect(result.resourceIds).toContain('/peel');
+
+        // Violation
+        expect(result.violations).toHaveLength(1);
+
+        // Violation Window
+        const plan_duration_micro = 24 * 60 * 60 * 1000 * 1000;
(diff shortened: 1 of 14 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -117,17 +117,22 @@ public Optional<ResultsProtocol.OwnerRole> claim(final PlanId planId, final Long
   }
 
   @Override
-  public Optional<ResultsProtocol.ReaderRole> lookup(final PlanId planId) {
+  public Optional<ResultsProtocol.ReaderRole> lookup(final PlanId planId, final Optional<SimulationDatasetId> simulationDatasetId) {
     try (final var connection = this.dataSource.getConnection()) {
       final var simulation = getSimulation(connection, planId);
-      final var datasetRecord = lookupSimulationDatasetRecord(
+      final Optional<SimulationDatasetRecord> datasetRecord;
+      if(simulationDatasetId.isPresent()) {
```
mattdailis: Formatting nit: missing space
```suggestion
      if (simulationDatasetId.isPresent()) {
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -117,17 +117,22 @@ public Optional<ResultsProtocol.OwnerRole> claim(final PlanId planId, final Long
...
+        datasetRecord = lookupSimulationDatasetRecord(
           connection,
           simulation.id());
-      final var datasetId$ = datasetRecord.map(SimulationDatasetRecord::datasetId);
-
-      if (datasetId$.isEmpty()) return Optional.empty();
+      }
+      if (datasetRecord.isEmpty()) return Optional.empty();
+      // If this assertion fails, then the specified sim dataset is not a simulation for the specified plan
+      assert(datasetRecord.get().simulationId() == simulation.id());
```
mattdailis: Good check - I think we'll want to go a step further and make this an if statement and throw our own exception. Would it be accurate to say that a mistake on an API user's part could result in this assertion failing? If it's something that the user has the power to fix, it would be best if we can return it as an official failure type. Tagging Luke just to keep this on his radar @skovati 

As an aside, I've grown distrustful of assertions outside of test cases because they are disabled by default - we'd need to start the JVM with a special flag to enable them. https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#enable-disable

e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
```diff
@@ -0,0 +1,111 @@
+import { expect, test } from '@playwright/test';
+import req, { awaitSimulation } from '../utilities/requests.js';
+
+test.describe.serial('Temporal Subset Simulation', () => {
```
mattdailis: Looks like we need a new title for this one. Perhaps "Check Constraints Against Given Simulation Dataset"?

e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        await req.deleteMissionModel(request, mission_model_id);
+        await req.deletePlan(request, plan_id);
+    });
+
+    test('Constraints Loads Specified Outdated Sim Dataset', async ({request}) => {
+        // Simulate
+        const oldSimDatasetId = (await awaitSimulation(request, plan_id)).simulationDatasetId;
+        //const x = await req.getSimulationDataset(request, oldSimDatasetId);
+        //console.log(x);
+        //console.log(x.simulated_activities)
```
mattdailis:
```suggestion
```

mattdailis (commented):
Looks great to me - a few minor comments to address, and then I say we ship it :shipit:

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
Mythicaeda: Ah, yeah, this could be an API mistake (as in the user literally puts in the wrong ID). I'll update this to an if->throw 👍

Mythicaeda: Done. Ended up splitting the method off into an override rather than an optional to reflect the divergent behavior (since the check will only potentially fail if the optional is present)

e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
Mythicaeda: Thanks for the catch. Test has been renamed

e2e-tests/src/tests/constraints-load-old-sim-results.test.ts
Mythicaeda: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
Mythicaeda: Done

skovati:
Record keeping: this was overlooked when I closed https://github.com/NASA-AMMOS/aerie/issues/1006, as I thought @cohansen's work with tracking constraint runs handled this case as well.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -120,14 +121,32 @@ public Optional<ResultsProtocol.OwnerRole> claim(final PlanId planId, final Long
...
+
+  @Override
+  public Optional<ResultsProtocol.ReaderRole> lookup(final PlanId planId, final SimulationDatasetId simulationDatasetId) throws SimulationDatasetMismatchException {
+    try (final var connection = this.dataSource.getConnection()) {
+      final var simulation = getSimulation(connection, planId);
+      final var datasetRecord = getSimulationDatasetRecordById(connection, simulationDatasetId.id());
+
+      if (datasetRecord.isEmpty()) return Optional.empty();
+      // If this check fails, then the specified sim dataset is not a simulation for the specified plan
+      if(datasetRecord.get().simulationId() != simulation.id()){
```
mattdailis: nit: 
```suggestion
      if (datasetRecord.get().simulationId() != simulation.id()) {
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -48,10 +48,17 @@ public ConstraintAction(
   }
 
   public List<ConstraintResult> getViolations(final PlanId planId, final Optional<SimulationDatasetId> simulationDatasetId)
-  throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException
-  {
+      throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException, SimulationDatasetMismatchException {
     final var plan = this.planService.getPlanForValidation(planId);
-    final var revisionData = this.planService.getPlanRevisionData(planId);
+    final Optional<SimulationResultsHandle> resultsHandle$;
+    if(simulationDatasetId.isPresent()){
```
mattdailis: nit: 
```suggestion
    if (simulationDatasetId.isPresent()) {
```

mattdailis (approved):
🚀

PR #1170, opened 2023-09-29
Fix scheduler issue

### LINKED ISSUE #1165: Scheduling a goal again immediately after the previous one finishes results in error
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

develop

### Describe the bug

The UI e2e test for scheduling the same goal twice has started to fail causing the PR checks to not pass. The PRs themselves don't have any scheduling related changes, so the suspicion is that scheduling changed in a way that makes the UI test obsolete.

We just need to know if there's a change that needs to be made to the tests.

### Reproduction

To reproduce:
1. Through the UI, create a Banananation plan
2. Then create the following scheduling goal and associate it to the plan created in step 1
```
export default (): Goal =>
    Goal.ActivityRecurrenceGoal({
        activityTemplate: ActivityTemplates.BakeBananaBread({
            temperature: 325.0,
            tbSugar: 2,
            glutenFree: false,
        }),
        interval: Temporal.Duration.from({ hours: 12 }),
    });
```
3. Navigate to the plan and open the scheduling panel
4. Click the `Schedule` button in the scheduling panel
5. Wait for the completion status
6. Click the `Schedule` button again
7. The error console should now have the following error:
```
java.lang.NullPointerException: Cannot read field "durationInMicroseconds" because "right" is null
	at gov.nasa.jpl.aerie.merlin.protocol.types.Duration.add(Duration.java:236)
	at gov.nasa.jpl.aerie.merlin.protocol.types.Duration.plus(Duration.java:361)
	at gov.nasa.jpl.aerie.scheduler.model.SchedulingActivityDirective.getEndTime(SchedulingActivityDirective.java:167)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.satisfyGoalGeneral(PrioritySolver.java:577)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.satisfyGoal(PrioritySolver.java:368)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.solve(PrioritySolver.java:325)
	at gov.nasa.jpl.aerie.scheduler.solver.PrioritySolver.getNextSolution(PrioritySolver.java:127)
	at gov.nasa.jpl.aerie.scheduler.worker.services.SynchronousSchedulerAgent.schedule(SynchronousSchedulerAgent.java:215)
	at gov.nasa.jpl.aerie.scheduler.worker.SchedulerWorkerAppDriver.main(SchedulerWorkerAppDriver.java:90)
```

### Logs

_No response_

### System Info

```shell
Chrome 117.0.5938.132, Mac OS 13.5.2 (22G91)
```

### Severity

Moderate

adrienmaillard:
* **Tickets addressed:** Fixes #1165 
* **Review:** By commit  

## Description

In the case that
- a goal is totally satisfied
- initial simulation results were used
- only association is checked 

The duration pulling from simulation is not performed (as there is no simulation performed) and the durations of activities were not pulled from the initial simulation results.  

The last commit of this branch makes it so durations are pulled from initial sim results if available.

On the way there, I found some things to improve.

## Verification

All tests are passing. 

- [x] reproduce bug as e2e test

## Documentation

None

## Future work

None

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationData.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationData.java
@@ -7,5 +7,5 @@ import java.util.Collection;
 
 public record SimulationData(
     SimulationResults driverResults,
-    gov.nasa.jpl.aerie.constraints.model.SimulationResults constraintsResults,
-    Collection<SchedulingActivityDirective> activitiesInPlan){}
+    gov.nasa.jpl.aerie.constraints.model.SimulationResults constraintsResults
+){}
--- a/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
+++ b/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
@@ -291,6 +291,7 @@ public record SynchronousSchedulerAgent(
                                                      final Map<SchedulingActivityDirective, ActivityDirectiveId> schedDirectiveToMerlinId)
   throws MerlinServiceException, IOException
   {
+    if(!simulationFacade.areInitialSimulationResultsStale()) return Optional.empty();
     //finish simulation until end of horizon before posting results
     try {
       simulationFacade.computeSimulationResultsUntil(planningHorizon.getEndAerie());
--- a/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
+++ b/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
@@ -132,7 +132,7 @@ public record SynchronousSchedulerAgent(
         final var externalProfiles = loadExternalProfiles(planMetadata.planId());
         final var initialSimulationResults = loadSimulationResults(planMetadata);
         //seed the problem with the initial plan contents
-        final var loadedPlanComponents = loadInitialPlan(planMetadata, problem);
+        final var loadedPlanComponents = loadInitialPlan(planMetadata, problem, initialSimulationResults);
         problem.setInitialPlan(loadedPlanComponents.schedulerPlan(), initialSimulationResults);
         problem.setExternalProfile(externalProfiles.realProfiles(), externalProfiles.discreteProfiles());
         //apply constraints/goals to the problem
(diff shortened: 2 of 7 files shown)


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationData.java
```diff
@@ -7,5 +7,5 @@
 
 public record SimulationData(
     SimulationResults driverResults,
-    gov.nasa.jpl.aerie.constraints.model.SimulationResults constraintsResults,
-    Collection<SchedulingActivityDirective> activitiesInPlan){}
```
Mythicaeda, line 11: Do you recall why `activitiesInPlan` used to be a field on this record?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
```diff
@@ -439,6 +444,16 @@ private PlanComponents loadInitialPlan(final PlanMetadata planMetadata, final Pr
...
+        //look for durations in initial sim results
+        if(initialSimulationResults.isPresent()){
+          for(final var simAct: initialSimulationResults.get().simulatedActivities.entrySet()){
+            if(simAct.getValue().directiveId().isPresent() &&
+               simAct.getValue().directiveId().get().equals(elem.getKey())){
+              actDuration = simAct.getValue().duration();
+            }
+          }
+        }
+
```
Mythicaeda: Should this block only be in the `// Do nothing` cases of Uncontrollable, Fixed, and Parametric? If the activity is Controllabe, its duration would've already been set on L432.

Additionally, is it possible to set the Duration of Fixed/Parametric activities during this method?

Mythicaeda (commented):
2 Things:
1) For clarification, is the bug occurring because the Scheduler is posting simulation results despite not actually simulating, causing issues when it goes to load those results again?
2) Can you add an e2e test that replicates this bug (make a plan that satisfies its goal, simulate, schedule, schedule again)?

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
adrienmaillard: > Should this block only be in the // Do nothing cases of Uncontrollable, Fixed, and Parametric? If the activity is Controllabe, its duration would've already been set on L432.

It is not really related to duration type and I did not want to make it look like it was. But why not.

> Additionally, is it possible to set the Duration of Fixed/Parametric activities during this method?

Absolutely, let's make it smarter

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationData.java
adrienmaillard: Because an early mental model I had included caching (and invalidating) `SimulationData` with respect to the activities in the plan.

adrienmaillard:
I have added an e2e test.

Mythicaeda (approved):

Issue #1173, opened 2023-10-02
Delete TimeRangeExpression hierarchy


adrienmaillard:
These are from a time where the scheduler was a stand-alone app. Although I am not really happy to have to go through simulation for every constraint we have to compute, it seems this the way to go for now so these classes have to be deleted so we do not spend time maintaining them.

PR #1175, opened 2023-10-03
Refactor of ActivityExpression and derived classes

adrienmaillard:
* **Tickets addressed:** None. Separated this commit from another PR on @Mythicaeda 's request.
* **Review:** By commit  

## Description

The hierarchy was an unnecessary mess. The approach taken is: get rid of all but `ActivityExpression`, the parent class of all. 

## Verification

It's a refactor. All existing tests pass. 

## Documentation

None. 

## Future work

Same kind of approach for #1173


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CoexistenceGoal.java
```diff
@@ -267,12 +260,8 @@ else if (this.initiallyEvaluatedTemporalContext == null) {
         }
       }
 
-      ActivityCreationTemplate temp;
-      if (disj) {
-        temp = (ActivityCreationTemplateDisjunction) activityCreationTemplate.build();
-      } else {
-        temp = (ActivityCreationTemplate) activityCreationTemplate.build();
-      }
+      final var temp = activityCreationTemplate.build();
```
Mythicaeda: Minor, can ignore: this variable can be inlined into its usage on L268. If this is done, I'd recommend reformatting the line for readability.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -89,11 +88,6 @@ public RootFindingResult<Duration, History> findRoot(
         public Duration valueAt(final Duration x, final History history) {
           return f.valueAt(x, history).minus(y);
         }
-
-        @Override
-        public boolean isApproximation() {
-          return f.isApproximation();
-        }
```
Mythicaeda, line 96: What was the original purpose of this field/method?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -43,24 +44,25 @@
  *
  * templates may be fluently constructed via builders that parse like first
  * order logic predicate clauses, used in building up scheduling rules
+ * @param startOrEndRange
```
Mythicaeda: Minor: This parameter is missing a description

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -43,24 +44,25 @@
...
+ * @param startRange range of allowed values for matching activity scheduled start times activities with null start time do not match any non-null range
+ * The range itself determines if endpoints are inclusive or exclusive
```
Mythicaeda: Minor: Readability:
```suggestion
 * @param startRange Range of allowed values for matching activity scheduled start times. Activities with null start time do not match any non-null range.
 * The range itself determines if endpoints are inclusive or exclusive
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -77,29 +79,32 @@ protected ActivityExpression() { }
...
     Map<String, ProfileExpression<?>> arguments = new HashMap<>();
 
+    protected @Nullable
+    ActivityType type;
+    protected @Nullable Interval startsIn;
+    protected @Nullable Windows startsOrEndsIn;
+    protected @Nullable Interval endsIn;
+    protected Windows startsInR;
+    protected @Nullable Pair<Expression<? extends Profile<?>>, Expression<? extends Profile<?>>> durationIn;
+    protected java.util.regex.Pattern nameRe;
```
Mythicaeda: Why do some of these have `In` appended to the name and others don't? Additionally, is the `R` at the end of `startsInR` a typo or does it have significance?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -553,28 +422,25 @@ public boolean matches(@NotNull gov.nasa.jpl.aerie.constraints.model.ActivityIns
...
+      final Optional<Duration> durRequirementLower = this.durationRange.getLeft()
+          .evaluate(simulationResults, evaluationEnvironment)
+          .valueAt(Duration.ZERO)
+          .flatMap($ -> $.asInt().map(i -> Duration.of(i, Duration.MICROSECOND)));
+      final Optional<Duration> durRequirementUpper = this.durationRange.getRight()
           .evaluate(simulationResults, evaluationEnvironment)
           .valueAt(Duration.ZERO)
           .flatMap($ -> $.asInt().map(i -> Duration.of(i, Duration.MICROSECOND)));
-      match = durRequirement.isEmpty() || (dur != null && durRequirement.get() == dur);
+      match = durRequirementUpper.isEmpty() || (dur != null && dur.noShorterThan(durRequirementLower.get()) && dur.noLongerThan(durRequirementUpper.get()));
```
Mythicaeda: I don't think this if statement is sufficient, as there are four possible checks that need to be run, and only one case four is properly checked:

1. Both are empty (no duration check)
2. Upper is empty, Lower isn't (match's duration must be >= Lower)
3. Lower is empty, Upper isn't (match's duration must be <= Upper)
4. Both are defined (match's duration must be between Lower and Upper)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -553,28 +422,25 @@ public boolean matches(@NotNull gov.nasa.jpl.aerie.constraints.model.ActivityIns
       final var startT = act.interval.start;
```
Mythicaeda: This comment actually is a suggestion for refining L413-L414: 
```java
match = type == null || type.getName().equals(act.type);
```
`match` is always true on this line, and since these are both Strings, there's no need to use `Objects.equals` instead of the String's `.equals` method, since it is null safe.
With this change the comment can go away as well

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -498,28 +370,25 @@ public boolean matches(@NotNull SchedulingActivityDirective act, SimulationResul
       final var startT = act.startOffset();
```
Mythicaeda: This comment is actually a suggestion for L361-362.
```java
match = type == null || type.equals(act.getType());
```
`match` is always true on this line and we've overriden `ActivityType.equals` to be a better check than `type == act.getType`.
With this change the comment can go away as well

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: I was envisioning that we would have access to approximations given by the model and that we could pass these functions around and act according to whether it was an approximation of the duration or not (maybe more/less checks depending on the context) but this was yet another hopeful hallucination.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: I have now removed it and replaced its only usage in CardinalityGoal by startsIn

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: the "in" is referring to the interval/windows, "starts in [0,11]". 
the R refers to Range that was used before for this. I have removed this member, it was used for search (only in one place) but was impossible to leverage in STN/rootfinding.

adrienmaillard:
@Mythicaeda I have made the changes. 
One notable thing: I had omitted to make one change in `PrioritySolver` prior to using its `getLatestSimResultsUpTo` in rootfinding (we had a very similar method in `ActivityCreationTemplate`...without synchronization). We don't want to synchronize the plan with sim when doing rootfinding so I had to pull that procedure. It's in the first commit.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -541,40 +343,37 @@ public boolean matches(@NotNull SchedulingActivityDirective act, SimulationResul
...
+      final Optional<Duration> durRequirementUpper = this.durationRange.getRight()
           .evaluate(simulationResults, evaluationEnvironment)
           .valueAt(Duration.ZERO)
           .flatMap($ -> $.asInt().map(i -> Duration.of(i, Duration.MICROSECOND)));
-      match = durRequirement.isEmpty() || (dur != null && durRequirement.get() == dur);
+      if(durRequirementLower.isEmpty() && durRequirementUpper.isEmpty()){
+        throw new RuntimeException("ActivityExpression is malformed, duration bounds are absent but the range is not null");
+      }
+      if(durRequirementLower.isPresent()){
+        match = match && dur.noShorterThan(durRequirementLower.get());
```
Mythicaeda: Minor, can ignore: The `match &&` on this line is unnecessary, as `match` hasn't been updated since entering the `if(match &&` block, meaning `match` must equal `true` (I agree that it is needed on L375, since this line might've set it to false).

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/TimeRangeExpression.java
```diff
@@ -48,7 +48,8 @@ public Windows computeRange(final SimulationResults simulationResults, final Pla
       if(minTimepoint.isPresent() && maxTimepoint.isPresent()) {
         final var anchorActSearch = new ActivityExpression.Builder()
             .basedOn(actTemplate)
-            .startsIn(inter).build(); //check if it exists IN the windows, not just the upper and lower bounds of the interval
+            //.startsIn(inter)
```
Mythicaeda, line 51: Minor: commented out code

adrienmaillard: Yeah, it's not worth it to update it because this class is going to be deleted soon.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
adrienmaillard: OK

adrienmaillard:
Made the changes + I have collapsed the 2 matches method into one with a supplementary parameter commanding whether the match on arguments should be on the subset of total equality.  Cleaner.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -483,106 +289,72 @@ ActivityExpression ofType(@NotNull ActivityType type) {
...
+  public boolean matches(
+      final @NotNull gov.nasa.jpl.aerie.constraints.model.ActivityInstance act,
+      final SimulationResults simulationResults,
+      final EvaluationEnvironment evaluationEnvironment,
+      final boolean matchArgumentsExactly) {
     boolean match = true;
 
-    //REVIEW: literal object equality is probably correct for type
-    match = match && (type == null || Objects.equals(type.getName(), act.type));
+    match = (type == null || type.getName().equals(act.type));
```
Mythicaeda: Minor, can ignore: declaration and assignment of `match` can be merged.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -483,106 +289,72 @@ ActivityExpression ofType(@NotNull ActivityType type) {
...
-
-    if (match && startOrEndRangeW != null) {
-      final var startT = act.interval.start;
-      final var endT = act.interval.end;
-      match = ((startT != null) && startOrEndRangeW.includes(Interval.at(startT))
-               || (endT != null) && startOrEndRangeW.includes(Interval.at(endT)));
-    }
-
     if (match && endRange != null) {
       final var endT = act.interval.end;;
```
Mythicaeda: Minor, can ignore: extra `;`

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -18,6 +19,7 @@
 import gov.nasa.jpl.aerie.scheduler.model.ActivityType;
```
Mythicaeda, line 19: Minor, can ignore: L10 (Windows import) is unused

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -609,7 +381,6 @@ public String prettyPrint(final String prefix) {
   @Override
   public void extractResources(final Set<String> names) { }
 
-
   /**
    * Evaluates whether a SerializedValue can be qualified as the subset of another SerializedValue or not
     * @param superset the proposed superset
```
Mythicaeda, line 383: Minor, can ignore: A little below here on L402, 408, 417: 
The returns can be cleaned up slightly (using Numeric as an example):
```java
return argumentsAsNumeric.map(bigDecimal -> bigDecimal.equals(value))
                         .orElse(false);
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/constraints/activities/ActivityExpression.java
```diff
@@ -333,124 +224,44 @@ Builder getThis() {
...
+    public @NotNull ActivityExpression build() {
+      final var template = new ActivityExpression(
+          startsIn,
+          endsIn,
+          durationIn,
+          type,
+          nameRe,
+          arguments
+      );
       return template;
```
Mythicaeda: Minor, can ignore: simplification
```suggestion
    public @NotNull ActivityExpression build() {
      return new ActivityExpression(
          startsIn,
          endsIn,
          durationIn,
          type,
          nameRe,
          arguments
      );
```

Mythicaeda (approved):
LGTM! Last couple comments are minor code cleanup and can be ignored

PR #1177, opened 2023-10-03
Improve ConstraintAction Error Messages

Mythicaeda:
* **Review:** By commit  

## Description

Follow up to #1169. 

The action will now correctly distinguish between a plan never being simulated and a specified sim dataset not existing when erroring on `resultsHandle$` being empty. 

Additionally, the serializer for SimDatasetMismatch now refers to the correct extension.

## Verification

A bindings test has been added. The bindings test fo the `SimDatasetMismatch` exception has been updated.

## Documentation

No docs changes needed.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
@@ -51,11 +51,18 @@ public class ConstraintAction {
       throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException, SimulationDatasetMismatchException {
     final var plan = this.planService.getPlanForValidation(planId);
     final Optional<SimulationResultsHandle> resultsHandle$;
+    final SimulationDatasetId simDatasetId;
     if (simulationDatasetId.isPresent()) {
       resultsHandle$ = this.simulationService.get(planId, simulationDatasetId.get());
+      simDatasetId = resultsHandle$
+        .map(SimulationResultsHandle::getSimulationDatasetId)
+        .orElseThrow(() -> new InputMismatchException("simulation dataset with id `" + simulationDatasetId.get().id() + "` does not exist"));
     } else {
       final var revisionData = this.planService.getPlanRevisionData(planId);
       resultsHandle$ = this.simulationService.get(planId, revisionData);
+      simDatasetId = resultsHandle$
+        .map(SimulationResultsHandle::getSimulationDatasetId)
+        .orElseThrow(() -> new InputMismatchException("plan with id " + planId.id() +" has not yet been simulated at its current revision"));
     }
 
 
@@ -68,9 +75,6 @@ public class ConstraintAction {
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
-    final var simDatasetId = resultsHandle$
-        .map(SimulationResultsHandle::getSimulationDatasetId)
-        .orElseThrow(() -> new InputMismatchException("no simulation datasets found for plan id " + planId.id()));
     final var violations = new HashMap<Long, ConstraintResult>();
 
     final var validConstraintRuns = this.constraintService.getValidConstraintRuns(constraintCode.values().stream().toList(), simDatasetId);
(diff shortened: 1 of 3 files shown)


mattdailis:
I'm highly supportive of improving these error messages 🚀 I was preparing for a demo this morning, and ran into "input mismatch exception". I think longer term we'll want the UI to specially handle these messages.

As far as this PR is concerned, I want to workshop the messages to make sure they are accurate enough to be actionable. The two cases seem to be:
- The provided simulation dataset id either doesn't exist, or it does exist, but isn't associated with this plan. Can we put this succinctly?
- No simulation dataset id was provided, and the latest revision of the plan has never been simulated

Mythicaeda:
> The two cases seem to be:

There are three cases, actually:
1. There was a simulation dataset id specified, but that dataset doesn't exist
2. There was a simulation dataset id specified, and that dataset exists, but it's a dataset for a different plan
3. There was no simulation dataset id specified, and the plan has never been simulated (note that plan revision is not actually used, it only still exists as a param due to the unused `UncachedSimulationService`. All checking a plan revision does is prove that the plan exists)

Cases 1 and 3 share an error class but have unique error messages in the `cause` field. Case 2 has its own error class.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -51,11 +51,18 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
     final Optional<SimulationResultsHandle> resultsHandle$;
+    final SimulationDatasetId simDatasetId;
     if (simulationDatasetId.isPresent()) {
       resultsHandle$ = this.simulationService.get(planId, simulationDatasetId.get());
+      simDatasetId = resultsHandle$
+        .map(SimulationResultsHandle::getSimulationDatasetId)
+        .orElseThrow(() -> new InputMismatchException("simulation dataset with id `" + simulationDatasetId.get().id() + "` does not exist"));
     } else {
       final var revisionData = this.planService.getPlanRevisionData(planId);
       resultsHandle$ = this.simulationService.get(planId, revisionData);
```
mattdailis, line 62: My expectation is that `.get` returns `Optional.empty()` in the event that this plan has been simulated before, but it has been edited since the last simulation, so there are no up-to-date simulation datasets. Does that track?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -51,11 +51,18 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
       throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException, SimulationDatasetMismatchException {
     final var plan = this.planService.getPlanForValidation(planId);
     final Optional<SimulationResultsHandle> resultsHandle$;
+    final SimulationDatasetId simDatasetId;
     if (simulationDatasetId.isPresent()) {
       resultsHandle$ = this.simulationService.get(planId, simulationDatasetId.get());
```
mattdailis, line 56: I'm under the impression that resultsHandle$ will be `Optional.empty()` if the given simulation dataset id does exist, but is associated with a different plan. Is that accurate?

mattdailis (commented):
Couple clarifying questions - I have a feeling the messages might still be a bit misleading

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: That is not accurate. If the sim dataset id does exist but is for a different plan, a `SimulationDatasetMismatchException` will be thrown. You can see it thrown on [L144 of `PostgresResultsCellRepository`](https://github.com/NASA-AMMOS/aerie/blob/refactor/adjust-constraints-error-messages/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java#L144), where it bubbles all the way up to `MerlinBindings`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: I thought it didn't based on the error message, but some testing indicates that it does. I have adjusted the message 👍.

mattdailis (approved):
💯

PR #1178, opened 2023-10-04
Fix Scheduling Assigning Incorrect Span Ids

### LINKED ISSUE #965: Activity directives disappear when running simulation after scheduling
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

1.6.0

### Describe the bug

After running a complex scheduling goal consisting of multiple placements of activity types and simulating, some of the activity directives disappear from the plan UI view. After the scheduling run, all activity directives are placed in the plan as expected, however after simulating, some seemingly random selection disappear. The activity directives and simulated activities still show up in the tables below the plan even though they don't have matches in the timeline view.

This bug makes debugging scheduling goals very challenging as the only way to get start/end times of the activity instances is clicking on the types in the UI view or referencing the activity/simulated activities table (which has duration and start offset which is not intuitive to debug off of).

Attaching two screen recordings of this happening. In the first one, you can see all the activity directives placed pre-simulation, then after, all but two activity directives disappear. In the second video, you can see only one chunk of the placed activity instances in the plan UI. The blue bars up top show where the others should be, but are not. I scrolled through the simulated activities table too, finding all correct instances that however do not show up in the UI.

https://github.com/NASA-AMMOS/aerie/assets/40210944/b2081e44-9d67-4ad2-ae10-04dabfdb97ee

https://github.com/NASA-AMMOS/aerie/assets/40210944/ea50de1b-71ad-4e66-a86d-427118df44e5

Attached below are the console logs when running the goal in the second video shared:
Screenshot 2023-06-01 at 4 41 08 PM
Screenshot 2023-06-01 at 4 41 14 PM

### Reproduction

This can be attempted to be reproduced by testing a scheduling goal (or set of goals) that place sets of activity instances (in the cases I've seen it has been multiple coexistence goals hooked to a primary anchor activity). Run the scheduler first and ensure all activity directives show up and then simulate and check again. In my case, the activity types always showed up in the activity directives/simulated activities table, yet some random assortment would not appear in the UI plan view.

### Logs

_No response_

### System Info

```shell
Google Chrome browser, macOS Ventura 13.3.1
```

### Severity

Moderate

Mythicaeda:
* **Tickets addressed:** Closes #965
* **Review:** By commit  

## Description

**Bug Description:** After inserting all the spans from simulation, the scheduler generates a map from its internal span ids to the ids the DB assigned to the spans. The original loop was missing a `++i` statement, meaning all of the Scheduler's span ids ended up mapped to the id of the first span posted. This would cause the scheduler to produce incorrect data when it went to update the `parent_id` field of the posted child spans.

This bug has been fixed by changing the `for-each` loop to a `for loop`. 

## Verification

A new e2e test was added that schedules a decomposing activity and checks that the post sim results have the correct `parentId`s assigned to all the spans.

### DIFF
--- a/e2e-tests/src/tests/scheduling-and-decomposition.test.ts
+++ b/e2e-tests/src/tests/scheduling-and-decomposition.test.ts
@@ -1,0 +1,128 @@
+import { expect, test } from '@playwright/test';
+import req, {awaitScheduling} from '../utilities/requests.js';
+
+test.describe('Scheduling and Decomposing Activities', () => {
+  const plan_start_timestamp = "2021-001T00:00:00.000";
+  const plan_end_timestamp = "2021-090T00:00:00.000";
+  let mission_model_id: number;
+  let plan_id: number;
+  let plan_revision: number;
+  let specification_id: number;
+
+  test.beforeAll(async ({request}) => {
+    let rd = Math.random() * 100000;
+    let jar_id = await req.uploadJarFile(request);
+    // Add Mission Model
+    const model: MissionModelInsertInput = {
+      jar_id,
+      mission: 'aerie_e2e_tests',
+      name: 'Banananation (e2e tests)',
+      version: rd + "",
+    };
+    mission_model_id = await req.createMissionModel(request, model);
+    await delay(1000);
+
+    // Add Plan
+    const plan_input: CreatePlanInput = {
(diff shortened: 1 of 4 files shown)


adrienmaillard (approved):
LGTM

PR #1179, opened 2023-10-04
Added an extension table and a table to keep track of roles for exten…

### LINKED ISSUE #1153: Custom Menu Items / Helper Scripts API
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

I'm frustrated when [...] happens as documented in issue-XYZ

Users would like to register custom helpers within Aerie. The UI will provide a menu to invoke the helpers which represent individual APIs that may generate reports, add activities to a plan, trigger archival workflows to OCS, etc.

We will need the following endpoints

- register helper
- unregister helper
- invoke helper

Registering a helper will require the following information

- URL to call - optionally split into multiple components (e.g. hostname and path)
- Label
- Description
- Roles that are allowed to invoke it

When invoked, Aerie will call the API and provide a JSON object containing
- plan url
- selected activity
- selected simulation run
- user information

```[tasklist]
### Tasks
- [ ] Add table to UI database
- [ ] Add hasura metadata
```

cohansen:
…sions

* **Tickets addressed:** Closes #1153.
* **Review:** By file

## Description
This PR adds the backend for custom helpers that we're calling extensions. An extension is an outside tool that can be invoked from within Aerie. This outside application my be pure, or have side effects but from our perspective we don't care as long as it accepts and responds with some known shape.

- [UI PR](https://github.com/NASA-AMMOS/aerie-ui/pull/920)
- [Documentation PR](https://github.com/NASA-AMMOS/aerie-docs/pull/87)

## Verification
Tested manually that extensions are being shown in the UI.

## Documentation
I'll link the documentation PR once it's open here.

## Future work
Right now we're sending the user cookie, `planId`, `selectedActivityDirectiveId`, `simulationDatasetId`, the gateway URL, and the Hasura URL in the request but this may change if we want to include more data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,8 @@ node_modules
 deployment/postgres-init-db/sql/**/*.sql
 !deployment/postgres-init-db/sql/ui/init.sql
 !deployment/postgres-init-db/sql/ui/tables/view.sql
+!deployment/postgres-init-db/sql/ui/tables/extension_roles.sql
+!deployment/postgres-init-db/sql/ui/tables/extensions.sql
 !deployment/postgres-init-db/sql/ui/tables/schema_migrations.sql
 !deployment/postgres-init-db/sql/ui/applied_migrations.sql
 
--- a/.gitignore
+++ b/.gitignore
@@ -41,12 +41,7 @@ node_modules
 
 # Ignore any SQL files copied from build tasks
 deployment/postgres-init-db/sql/**/*.sql
-!deployment/postgres-init-db/sql/ui/init.sql
-!deployment/postgres-init-db/sql/ui/tables/view.sql
-!deployment/postgres-init-db/sql/ui/tables/extension_roles.sql
-!deployment/postgres-init-db/sql/ui/tables/extensions.sql
-!deployment/postgres-init-db/sql/ui/tables/schema_migrations.sql
-!deployment/postgres-init-db/sql/ui/applied_migrations.sql
+!deployment/postgres-init-db/sql/ui
 
 # Ignore Gradle project-specific cache directory
 .gradle
--- a/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
+++ b/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
@@ -2,7 +2,7 @@ table:
   name: extension_roles
   schema: public
 object_relationships:
-  - name: extensions
+  - name: extension
     using:
       foreign_key_constraint_on: extension_id
 select_permissions:
--- a/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
+++ b/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
@@ -1,0 +1,37 @@
+table:
+  name: extension_roles
+  schema: public
+object_relationships:
+  - name: extensions
+    using:
+      foreign_key_constraint_on: extension_id
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: user
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [extension_id, role]
+      check: {}
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [extension_id, role]
+      filter: {}
+delete_permissions:
+  - role: aerie_admin
+    permission:
+      filter: {}
--- a/deployment/hasura/metadata/databases/AerieUI/tables/public_extensions.yaml
+++ b/deployment/hasura/metadata/databases/AerieUI/tables/public_extensions.yaml
@@ -1,0 +1,43 @@
+table:
+  name: extensions
+  schema: public
+array_relationships:
+  - name: extension_roles
+    using:
+      foreign_key_constraint_on:
+        column: extension_id
+        table:
+          name: extension_roles
+          schema: public
+select_permissions:
+  - role: aerie_admin
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: user
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: '*'
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [description, label, url]
+      check: {}
+      set:
+        owner: "x-hasura-user-id"
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [description, label, url]
+      filter: {}
+delete_permissions:
+  - role: aerie_admin
+    permission:
+      filter: {}
--- a/deployment/hasura/metadata/databases/AerieUI/tables/public_extensions.yaml
+++ b/deployment/hasura/metadata/databases/AerieUI/tables/public_extensions.yaml
@@ -35,7 +35,7 @@ insert_permissions:
 update_permissions:
   - role: aerie_admin
     permission:
-      columns: [description, label, url]
+      columns: [description, label, owner, url]
       filter: {}
 delete_permissions:
   - role: aerie_admin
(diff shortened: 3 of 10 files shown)


deployment/hasura/migrations/AerieUI/2_extensions/down.sql
```diff
@@ -0,0 +1,19 @@
+comment on column extension_roles.extension_id is null;
```
skovati: Minor: I believe these `comment is null` statements can be removed; comments are automatically `null`'ed when a corresponding column / table is dropped.

deployment/hasura/metadata/databases/AerieUI/tables/tables.yaml
```diff
@@ -1 +1,3 @@
+- "!include public_extension_roles.yaml"
```
skovati, line 1: Does the order need to be switched here since the former relies on the latter already existing? Or does Hasura just automagically handle that?

deployment/postgres-init-db/sql/ui/tables/extensions.sql
```diff
@@ -0,0 +1,36 @@
...
+comment on column extensions.owner is e''
+  'The user who owns the extension.';
+comment on column extensions.url is e''
+  'The URL of the API to be called.';
+comment on column extensions.updated_at is e''
+  'The time the extension was last updated.';
+
+create or replace function extensions_set_updated_at()
+returns trigger
+security definer
```
skovati: Why does this need `security definer` instead of the default `invoker`?

deployment/postgres-init-db/sql/ui/tables/extensions.sql
```diff
@@ -0,0 +1,36 @@
+create table extensions (
```
skovati, line 1: Doesn't belong in this first iteration, but would it be useful to store some input / output type definitions in the DB? That way, users could define a sort of custom API between the UI and their extensions, instead of our current hardcoded `ExtensionPayload`.

skovati (approved):
Looks great! Tested locally with the UI PR and it works flawlessly. Just have some non-blocking comments.

Ideally we would add an end-to-end test covering this feature, but that would probably live in the UI repo?

deployment/hasura/migrations/AerieUI/2_extensions/down.sql
cohansen: That'd be awesome, I saw them in other migrations so I copied it from there but I'll try this.

deployment/hasura/metadata/databases/AerieUI/tables/tables.yaml
cohansen: You're correct, order is important here.

deployment/postgres-init-db/sql/ui/tables/extensions.sql
cohansen: This is a great idea, we'll see if Clipper has a use-case that we're not accounting for with our initial pass but maybe this could be an enhancement.

deployment/postgres-init-db/sql/ui/tables/extensions.sql
cohansen: Good point, I can remove this!

.gitignore
```diff
@@ -43,6 +43,8 @@ node_modules
 deployment/postgres-init-db/sql/**/*.sql
 !deployment/postgres-init-db/sql/ui/init.sql
 !deployment/postgres-init-db/sql/ui/tables/view.sql
+!deployment/postgres-init-db/sql/ui/tables/extension_roles.sql
+!deployment/postgres-init-db/sql/ui/tables/extensions.sql
 !deployment/postgres-init-db/sql/ui/tables/schema_migrations.sql
 !deployment/postgres-init-db/sql/ui/applied_migrations.sql
```
Mythicaeda: I think this can be simplified to `!deployment/postgres-init-db/sql/ui`

deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
```diff
@@ -0,0 +1,37 @@
+table:
+  name: extension_roles
+  schema: public
+object_relationships:
+  - name: extensions
```
Mythicaeda: Minor: because this is an object relationship, the relationship name should be singular
```suggestion
  - name: extension
```

deployment/hasura/metadata/databases/AerieUI/tables/public_extensions.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - role: aerie_admin
+    permission:
+      columns: [description, label, url]
+      check: {}
+      set:
+        owner: "x-hasura-user-id"
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [description, label, url]
```
Mythicaeda: Minor: Consistency: `aerie_admin` should be allowed to update the `owner` field

deployment/hasura/migrations/AerieUI/2_extensions/up.sql
```diff
@@ -0,0 +1,56 @@
+create table extensions (
+  description text,
+  id integer generated always as identity,
```
Mythicaeda: Very minor, can ignore: can the PK please be moved to the first line in this `create table` statement?

deployment/postgres-init-db/sql/ui/tables/extension_roles.sql
```diff
@@ -0,0 +1,18 @@
+create table extension_roles (
+  extension_id  integer not null,
+  id integer generated always as identity,
```
Mythicaeda, line 2: Minor, can ignore: consistency: PK column should be declared first

deployment/postgres-init-db/sql/ui/tables/extensions.sql
```diff
@@ -0,0 +1,35 @@
...
+  description text,
+  id integer generated always as identity,
```
Mythicaeda, line 2: Minor, can ignore: consistency: PK column should be declared first

deployment/postgres-init-db/sql/ui/tables/extensions.sql
Mythicaeda: I'm not against this proposal, but how would the UI know how to handle arbitrary custom inputs/outputs from extensions? Or is the thought that there is a set of things an extension can ask for/return, rather than the current "the UI will always hand the entire payload"?

deployment/postgres-init-db/sql/ui/tables/extensions.sql
```diff
@@ -0,0 +1,35 @@
...
+comment on column extensions.label is e''
+  'The name of the extension that is displayed in the UI.';
+comment on column extensions.owner is e''
+  'The user who owns the extension.';
+comment on column extensions.url is e''
+  'The URL of the API to be called.';
+comment on column extensions.updated_at is e''
+  'The time the extension was last updated.';
+
+create or replace function extensions_set_updated_at()
```
Mythicaeda: This should be a `create` statement, as we want to break if this function already exists. This same feedback applies to the `up.sql`

deployment/postgres-init-db/sql/ui/tables/extension_roles.sql
```diff
@@ -0,0 +1,18 @@
...
+  role text not null,
+
+  constraint extension_roles_primary_key primary key (id),
+  constraint extension_roles_to_extension
+    foreign key (extension_id)
+      references "extensions"
+      on delete cascade
```
Mythicaeda: Minor: consistency: even FKs against generated PKs, we add `on update cascade`
```suggestion
    foreign key (extension_id)
      references "extensions"
      on update cascade
      on delete cascade
```

deployment/postgres-init-db/sql/ui/tables/extensions.sql
```diff
@@ -0,0 +1,35 @@
+create table extensions (
+  description text,
+  id integer generated always as identity,
+  label text not null,
+  owner text,
+  url text not null,
+  updated_at timestamptz not null default now(),
+
+  constraint extension_primary_key primary key (id)
```
Mythicaeda: Minor: typo:
```suggestion
  constraint extensions_primary_key primary key (id)
```

deployment/postgres-init-db/sql/ui/tables/extensions.sql
skovati: Interesting conversation we can have later if there is a need, doesn't need to block this PR though.

Mythicaeda (approved):

PR #1181, opened 2023-10-04
Add `created_by` column to `activity_directive`

### LINKED ISSUE #1176: Add `created_by` column to `activity_directive`
As part of the activity directive changelog work, it has been requested that the activity directive form include an original author `created_by` in addition to the `last_modified_by`.

Since this data is immutable, it wouldn't necessarily need to be tracked in the `activity_directive_changelog`.

skovati:
* **Tickets addressed:** closes #1176
* **Review:** By commit  

## Description
Adds a new column, `created_by`, to `activity_directive`, which is automatically populated on `activity_directive` `insert` by hasura. This field is nullable and will be empty for all existing activity directives.

## Verification
Manual testing mostly (these can be codified if desired):
 - [x] Verified new field is copied during plan merge (add)
 - [x] Verified that changing `created_by` by admin doesn't count towards diff in plan merge
 - [x] Verified that `created_by` is copied and restored during plan snapshot

Additionally relying on database migration tests for verification.

## Documentation
N/A

## Future work
Not a particular piece of work, but I'll note that changing relatively simple things like this in the database can be quite tedious, as several different stored procedures directly reference tables like `activity_directive`, so a developer needs to keep a mental list of these and what changes need to be made. This is especially quirky with database migrations, since we need to copy paste procedures just to change one line, which is what ~1400 lines of this PR is.


Mythicaeda (changes requested):
DBTests need to be updated. I know for sure off the top of my head that `Activity` and `SnapshotActivity` in `PlanCollaborationTests` need to be updated, as well as the method `assertActivityEquals`, and any tests that check that do Equivalency checks but can't use that method (ie, the test `snapshotCapturesAllActivities`, which is comparing between a `SnapshotActivity` and an `Activity`)

skovati:
@Mythicaeda DB tests have been updated

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -583,6 +614,7 @@ void snapshotCapturesAllActivities() throws SQLException {
 
         assertEquals(planActivities.get(i).sourceSchedulingGoalId, snapshotActivities.get(i).sourceSchedulingGoalId);
         assertEquals(planActivities.get(i).createdAt, snapshotActivities.get(i).createdAt);
+        assertEquals(planActivities.get(i).createdBy, snapshotActivities.get(i).createdBy);
```
Mythicaeda, line 618: This check also needs to be added in `assertActivitiesEquals` (L524)

skovati: Change has been made!

Mythicaeda (commented):
:shipit:

Mythicaeda (approved):

PR #1180, opened 2023-10-04
Convert e2eTests to Java

Mythicaeda:
* **Tickets addressed:** Closes #1057 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

Advantages to moving the e2eTests to Java:
- We now have the option to combine the DBTests with the rest of the e2eTests.
- We are now able to parallelize the e2eTests (Currently per-file parallelization is definitely supported, according to the [docs](https://playwright.dev/java/docs/test-runners#running-tests-in-parallel). Parallel Nested classes aren’t currently supported, since most of the nested classes are sharing a Playwright instance, but this could be altered. Our e2eTests are the slowest tests, so this is especially something worth looking into
- Easier to run/debug tests
- Minor, but we don't have to explicitly typecheck responses (`expect().notNull;`, `expect().toBe(number);` etc.)
- Minor, but developers don’t have to be familiar with TS to write/update e2eTests
   - Related, I think we’re generally more familiar with JUnit vs the Playwright TS interface

Outside of the conversion, the restructure from serial tests with side effects to isolated tests with proper Before and After tasks made it very easy to add new tests.

Disadvantages:
- Everything was a JSON in TS, which meant we could just define types and then declare the returns from Hasura requests as those types. In Java, we need to explicitly convert to and from JSON.

## Verification

Test updates beyond the conversion are:
- Tests hitting the deprecated `GetEffectiveActivityArguments` endpoint have been updated to hit the bulk endpoint
- Scheduling e2eTests have been fleshed out
- Model Effective Arguments tests have been added

## Documentation

The Readme in the e2eTest directory has been updated.

## Future work

I'd like to move the DBTests to be with the rest of the E2ETests, or at least have the them share some types.

PR #1182, opened 2023-10-05
Make workers exit when unexpectedly shutdown

Mythicaeda:
* **Review:** By commit  

## Description

Currently, our workers do not properly shut down when something goes wrong (ie a DB exception) in either the main thread or the listener thread, causing the workers to appear alive but unresponsive. This fixes that in two steps:

Commit 1: When the main thread exits, close the Javalin server and the listenerThread. Currently, neither of these threads are exited when the main thread stops, meaning the Java process is kept alive. Because the process is still alive, the container doesn't exit, preventing Docker from restarting it.

Commit 2: If the listener thread dies, kill the main thread. The worker cannot pick up any new work if its listener goes down, so if the listener dies, we need to restart the container and try to reconnect. This could also be achieved by trying to make a new listener thread if the original goes down, but by shutting down the container, the worker will accurately appear as struggling/restarting when viewed in Docker/AWS. This change **does not** affect any running simulations. Those will still complete (or attempt to complete).

## Verification

Both changes were verified manually.
For commit 1 (main loop terminates):
 - Checkout and compose up #1180. Run the `./gradlew e2e-tests:e2eTest --rerun-tasks` twice back to back. You should notice tests begin failing because of timeout errors during `awaitSimulation` and/or `awaitScheduling`.
 - Checkout and compose up this PR. Checkout #1180. Run `./gradlew e2e-tests:e2eTest --rerun-tasks` twice back to back. There should be no timeout errors. Open the logs for one of the workers and search for the messages `Listener has been interrupted` and `Listener has received interrupted signal`. One or both of the messages should appear in the logs.

For commit 2 (listener thread terminates):
- Checkout and compose up develop. Create a plan and simulate, then kill the postgres container. Start the Postgres container. Attempt to simulate. The simulation should remain stuck in 'pending'.
- Checkout and compose up this PR. Create a plan and simulate to prove that the workers are working. Kill the Postgres container. Start the Postgres container. Attempt to simulate. The simulation should be processed.

## Documentation

No docs need to be updated.

### DIFF
--- a/merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
+++ b/merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/ListenSimulationCapability.java
@@ -28,8 +28,8 @@ public class ListenSimulationCapability {
     this.notificationQueue = notificationQueue;
   }
 
-  public void registerListener() {
-    new Thread(() -> {
+  public Thread registerListener() {
+    final var listenThread = new Thread(() -> {
       try (final var connection = this.dataSource.getConnection()) {
         try (final var listenSimulationStatusAction = new ListenSimulationStatusAction(connection)) {
           listenSimulationStatusAction.apply();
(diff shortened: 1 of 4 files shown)


skovati (approved):
Looks good to me! My only concern is that killing the `/health` thread might hinder its use as a debugging tool if we add more info to the response in the future (e.g. `/health/dump_notification_queue`). This is much lower priority than the workers not restarting though.

PR #1183, opened 2023-10-05
User metadata annotations

### LINKED ISSUE #1034: Pull Units from mission model annotations and store them in DB

mattdailis:
* **Tickets addressed:** #1034 
* **Review:** By commit  

## Description

This is our second attempt. Our first attempt was #1055.

Short term goal: Allow users to provide unit information for parameters, computed attributes, and resources in their mission model, such that this information can be displayed in the UI and extracted via the API.

Broader goal: Set up a pattern that missions, and Aerie developers, can use to add more kinds of metadata in the future, e.g. Range.

There are two (mostly) independent problems to solve here:
1. How do mission modelers provide unit information, and define other kinds of custom metadata?
2. How is this custom metadata serialized and represented to downstream tools?

The following sections delve into these problems:

### How to define custom metadata
We knew from the outset that we would eventually want to be able to annotate deeply nested aspects of a value schema with metadata. E.g. a rover's pose could be characterized by a compound piece of data:

```yaml
pose:
  position:
    latitude: degrees
    longitude: degrees
    elevation: kilometers
  orientation: quaternion
    q1: unit vector component
    q2: unit vector component
    q3: unit vector component
    q4: scalar
```

We had originally descoped this capability, but then pivoted (i.e. we increased scope) when the path to implementing it became clear.

The two options to select between were Javadoc tags, and annotations. I will describe both approaches, and explain why we have settled on annotations.

#### Surface Syntax

This Javadoc tags approach amounted to parsing javadoc comments as strings, and extracting portions of them based on a set of tags, such as @aerie.unit.

It lended itself quite well to class style parameters, since it could be part of the javadoc comment attached to the parameter itself:

<details>
<summary>Javadoc tags approach</summary>

Example of using javadoc to add unit information to a parameter:

```java
@ActivityType("RunHeater")
public class RunHeater {
  /**
   * @aerie.unit seconds
   */
  @Parameter
  public long duration;

  /**
   * @aerie.unit kWh
   */
  @Parameter
  public int energyConsumptionRate;
}
```

Example of using javadoc to add unit information to a computed attributes record:

```java
  /**
   *
   * @aerie.computedAttribute durationInSeconds
   * @aerie.unit seconds
   * @aerie.computedAttribute energyConsumptionRate
   * @aerie.unit kWh
   */
  @AutoValueMapper.Record
  record ComputedAttributes(
    long durationInSeconds,
    int energyConsumptionRate
  ) {}
```

Example of using `@aerie.resourceName` to assign a unit to all resources that match a regular expression:

```java
/*
 * @aerie.resourceName \/prefix/.*
 * @aerie.unit F resources
 */
```

</details>

A limitation that we discovered of parsing the javadocs is that our implementation would only parse javadocs in a known set of locations. This works fine for parameters and computed attributes, but it breaks down for resources, which in practice are defined in submodels.

The second approach we pursued was to introduce a `@Unit` annotation, which could be attached to the _type_ of a parameter or computed attribute:

<details>
<summary>Annotation approach</summary>

Example of using an annotation to add unit information to a parameter:

```java
@ActivityType("RunHeater")
public class RunHeater {
  @Parameter
  public @Unit("seconds") long duration;

  @Parameter
  public @Unit("kWh") int energyConsumptionRate;
}
```

Example of using an annotation to add unit information to a computed attributes record:

```java
@AutoValueMapper.Record
record ComputedAttributes(
  @Unit("seconds") long durationInSeconds,
  @Unit("kWh") int energyConsumptionRate
) {}
```

</details>

What about resources? Those are defined through a different process from parameters and computed attributes. Resources get registered at runtime, in the mission model constructor, rather than at compile time like the other entities. The approach with javadocs would record a regular expression and do name matching at runtime. The current approach instead expects a mission model to provide the units to the registrar as runtime values. Two convenience methods have been defined:

```java
discreteResource(registrar, "/peel", this.peel, new DoubleValueMapper(), "kg");
realResource(registrar, "/fruit", this.fruit, "bananas");
```

This approach allows us to avoid any custom handling of units in the annotation processor, which means that it will generalize to any future metadata we would like to include. If it turns out that the regular expressions would be a significant quality of life improvement, we can consider that a separate feature.

### How is custom metadata serialized
In Java, there is now a new kind of ValueSchema: a `MetaSchema`, which carries a `Map<String, SerializedValue>`, as well as a nested `ValueSchema`. The interpretation is that all of the metadata in the `MetaSchema` applies to the nested `ValueSchema`.

The `withMeta` method, used to construct a `MetaSchema`, checks to see if the inner value is itself a `MetaSchema`, and if so merges the two, overlaying the outer one on top of the inner one.

When serializing to json, this `Map<String, SerializedValue>` is attached to the nested object, like so:

`MetaSchema(Map.of("unit", "m"),  IntSchema())` becomes

```json
{
  "type": "int",
  "metadata": {
    "unit": "m"
  }
}
```

### Future-proofing
One of the goals of this PR is to lay the groundwork for adding additional metadata to value schemas.

Part of this PR brings annotations into the `Resolver` and `TypePattern` machinery. This allows us to generate value mappers for annotated types.

This PR allows users to define their own annotations, and as long as a value mapper is provided for that annotation, Aerie will be able to serialize it and include it in the value schema.

One generalization that this PR leaves undone is to allow for using value mapper methods to define value mappers for annotated types. The nuance here is that, in this PR, we have users use the `@WithMetadata(Unit.class)` annotation to declare an annotation of interest, and we use that to generate a type rule, rather than the users writing the type rule as the return type from a static method, the way that other value mappers are defined. This is due to a limitation of type use annotations that we may be able to work around in the future.

## Verification

The Banananation example model was updated to include some unit information for parameters, computed attributes, and resources.

Things to test manually:
- [x] Do the value schemas look correct in the database?
- [x] Do scheduling and constraint checking still work with banananation? (I.e. do they correctly parse the new value schemas during typescript code generation)

The second bullet may be handled by e2e tests

## Documentation

- [ ] Using https://github.com/NASA-AMMOS/aerie-docs/pull/71 as an example, write a new page about metadata.

## Future work

- Incorporate unit information in typescript code generation
- Provide a convenience `@AutoValueMapper.Annotation` to make adding your own metadata annotations less verbose
- Find a workaround that allows type use annotations to be used for value mapper methods
- // int[] repeated(); // TODO add support for arrays in annotations for which we generate value mappers

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ActivityType.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ActivityType.java
@@ -5,14 +5,21 @@ import java.util.HashMap;
 import java.util.Map;
 
 
-public record ActivityType(String name, Map<String, Parameter> parameters) {
+public record ActivityType(String name, Map<String, Parameter> parameters, ValueSchema computedAttributes) {
+  /**
+   * Create an ActivityType with an empty computed attributes value schema.
+   */
+  public ActivityType(final String name, final Map<String, Parameter> parameters) {
+    this(name, parameters, new ValueSchema.ValueSchemaStruct(Map.of()));
+  }
+
   public static ActivityType fromJSON(JsonObject json) {
     final var parameters = json.getJsonObject("parameters");
     final var parameterMap = new HashMap<String, Parameter>();
     for (final var parameterName : parameters.keySet()) {
       parameterMap.put(parameterName, Parameter.fromJSON(parameters.getJsonObject(parameterName)));
     }
-    return new ActivityType(json.getString("name"), parameterMap);
+    return new ActivityType(json.getString("name"), parameterMap, ValueSchema.fromJSON(json.getJsonObject("computed_attributes_value_schema")));
   }
 
   public record Parameter(int order, ValueSchema schema) {
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
@@ -1,8 +1,0 @@
-package gov.nasa.jpl.aerie.banananation;
-
-import gov.nasa.jpl.aerie.banananation.activities.Banannotation;
-import gov.nasa.jpl.aerie.merlin.framework.annotations.MissionModel;
-
-@MissionModel.WithMetadata(name="banannotation", annotation= Banannotation.class)
-public class CustomValueMappers {
-}
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
@@ -1,0 +1,8 @@
+package gov.nasa.jpl.aerie.banananation;
+
+import gov.nasa.jpl.aerie.banananation.activities.Banannotation;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.MissionModel;
+
+@MissionModel.WithMetadata(name="banannotation", annotation= Banannotation.class)
+public class CustomValueMappers {
+}
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java
@@ -41,7 +41,7 @@ public final class Mission {
     registrar.discrete("/flag/conflicted", this.flag::isConflicted, new BooleanValueMapper());
     discreteResource(registrar, "/peel", this.peel, new DoubleValueMapper(), "kg");
     realResource(registrar, "/fruit", this.fruit, "bananas");
-    registrar.discrete("/plant", this.plant, new IntegerValueMapper());
+    discreteResource(registrar, "/plant", this.plant, new IntegerValueMapper(), "count");
     registrar.discrete("/producer", this.producer, new StringValueMapper());
     registrar.discrete("/data/line_count", this.dataLineCount, new IntegerValueMapper());
     registrar.topic("/producer", this.producer.ref, new StringValueMapper());
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java
@@ -17,6 +17,9 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
+import static gov.nasa.jpl.aerie.contrib.metadata.UnitRegistrar.discreteResource;
+import static gov.nasa.jpl.aerie.contrib.metadata.UnitRegistrar.realResource;
+
 public final class Mission {
   public final Accumulator fruit;
   public final AdditiveRegister peel;
@@ -36,8 +39,8 @@ public final class Mission {
 
     registrar.discrete("/flag", this.flag, new EnumValueMapper<>(Flag.class));
     registrar.discrete("/flag/conflicted", this.flag::isConflicted, new BooleanValueMapper());
-    registrar.discrete("/peel", this.peel, new DoubleValueMapper());
-    registrar.real("/fruit", this.fruit);
+    discreteResource(registrar, "/peel", this.peel, new DoubleValueMapper(), "kg");
+    realResource(registrar, "/fruit", this.fruit, "bananas");
     registrar.discrete("/plant", this.plant, new IntegerValueMapper());
     registrar.discrete("/producer", this.producer, new StringValueMapper());
     registrar.discrete("/data/line_count", this.dataLineCount, new IntegerValueMapper());
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/Banannotation.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/Banannotation.java
@@ -1,0 +1,16 @@
+package gov.nasa.jpl.aerie.banananation.activities;
+
+import gov.nasa.jpl.aerie.merlin.framework.annotations.AutoValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE_USE)
+@AutoValueMapper.Annotation
+public @interface Banannotation {
+  String value();
+}
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/ValueSchemaJsonParser.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/ValueSchemaJsonParser.java
@@ -4,6 +4,7 @@ import gov.nasa.jpl.aerie.json.JsonParseResult;
 import gov.nasa.jpl.aerie.json.JsonParser;
 import gov.nasa.jpl.aerie.json.SchemaCache;
 import gov.nasa.jpl.aerie.json.Unit;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 
 import javax.json.Json;
@@ -16,10 +17,12 @@ import java.util.function.Function;
 
 import static gov.nasa.jpl.aerie.json.BasicParsers.listP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.literalP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.mapP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
 import static gov.nasa.jpl.aerie.json.ProductParsers.productP;
 import static gov.nasa.jpl.aerie.json.Uncurry.tuple;
 import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
+import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
 
 public final class ValueSchemaJsonParser implements JsonParser<ValueSchema> {
   public static final JsonParser<ValueSchema> valueSchemaP = new ValueSchemaJsonParser();
@@ -38,7 +41,7 @@ public final class ValueSchemaJsonParser implements JsonParser<ValueSchema> {
     final var type = obj.get("type");
     if (!type.getValueType().equals(JsonValue.ValueType.STRING)) return JsonParseResult.failure("\"type\" field must be a string");
 
-    return switch (obj.getString("type")) {
+    JsonParseResult<ValueSchema> result = switch (obj.getString("type")) {
       case "real" -> JsonParseResult.success(ValueSchema.REAL);
       case "int" -> JsonParseResult.success(ValueSchema.INT);
       case "boolean" -> JsonParseResult.success(ValueSchema.BOOLEAN);
@@ -50,6 +53,13 @@ public final class ValueSchemaJsonParser implements JsonParser<ValueSchema> {
       case "variant" -> parseVariant(obj);
       default -> JsonParseResult.failure("Unrecognized value schema type");
     };
+
+    if (obj.containsKey("metadata")) {
+      final var metadata = mapP(serializedValueP).parse(obj.getJsonObject("metadata"));
+      return result.mapSuccess($ -> new ValueSchema.MetaSchema(metadata.getSuccessOrThrow(), $));
+    }
+
+    return result;
   }
 
   private JsonParseResult<ValueSchema> parseSeries(final JsonObject obj) {
@@ -174,6 +184,14 @@ public final class ValueSchemaJsonParser implements JsonParser<ValueSchema> {
                     .add("label", v.label())
                     .build(),
                 variants))
+            .build();
+      }
+
+      @Override
+      public JsonValue onMeta(final Map<String, SerializedValue> metadata, final ValueSchema target) {
+        return Json
+            .createObjectBuilder(target.match(this).asJsonObject())
+            .add("metadata", mapP(new SerializedValueJsonParser()).unparse(metadata))
             .build();
       }
     });
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
@@ -27,6 +27,7 @@ import javax.lang.model.element.Parameterizable;
 import javax.lang.model.element.RecordComponentElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeMirror;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -307,6 +308,11 @@ public class AutoValueMappers {
    * Turn a given string into a valid Java identifier
    * @param str a string that may or may not be a valid Java identifier
    * @return a string that is a valid Java identifier
+   *
+   * Any illegal characters are replaced by their UTF_8 integer representation as a decimal number surrounded by underscores.
+   * This helps ensure that strings the differ only in illegal characters get unique identifiers. (e.g. "m/s" and "m*s")
+   *
+   * E.g. "<Hello (Worl%d!)" becomes "_60_Hello_32__40_Worl_37_d_33__41_"
    */
   public static String getIdentifier(String str) {
     if (str.length() == 0) throw new IllegalArgumentException("Cannot turn empty string into an identifier");
@@ -314,13 +320,17 @@ public class AutoValueMappers {
     if (Character.isJavaIdentifierStart(str.charAt(0))) {
       identifier.append(str.charAt(0));
     } else {
-      identifier.append("_");
+      identifier.append('_');
+      identifier.append((int) str.charAt(0));
+      identifier.append('_');
     }
     for (int i = 1; i < str.length(); i++) {
       if (Character.isJavaIdentifierPart(str.charAt(i))) {
         identifier.append(str.charAt(i));
       } else {
-        identifier.append("_");
+        identifier.append('_');
+        identifier.append((int) str.charAt(i));
+        identifier.append('_');
       }
     }
     return identifier.toString();
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
@@ -11,13 +11,17 @@ import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
 import com.squareup.javapoet.TypeVariableName;
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.RecordValueMapper;
+import gov.nasa.jpl.aerie.merlin.framework.Result;
 import gov.nasa.jpl.aerie.merlin.framework.ValueMapper;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.AutoValueMapper;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.MissionModelRecord;
 import gov.nasa.jpl.aerie.merlin.processor.metamodel.TypeRule;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.Parameterizable;
 import javax.lang.model.element.RecordComponentElement;
@@ -27,10 +31,11 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 public class AutoValueMappers {
-  static TypeRule typeRule(final Element autoValueMapperElement, final ClassName generatedClassName) throws InvalidMissionModelException {
+  static TypeRule recordTypeRule(final Element autoValueMapperElement, final ClassName generatedClassName) throws InvalidMissionModelException {
     if (!autoValueMapperElement.getKind().equals(ElementKind.RECORD)) {
       throw new InvalidMissionModelException(
           "@%s.%s is only allowed on records".formatted(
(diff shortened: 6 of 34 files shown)


mattdailis:
Fixed a broken test - `valueSchemaIsNumeric` was returning false for the `/fruit` resource, because it's wrapped in a meta schema. I added the case below:

https://github.com/NASA-AMMOS/aerie/blob/416b628e82c1e6af1c0884c294aed5579713490a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/TypescriptCodeGenerationService.java#L392-L397

mattdailis:
New e2e tests should be passing now. Updates since the walkthrough:
- There is now an `@AutoValueMapper.Annotation` that can be used to generate a value mapper for a given annotation. It will generate a struct with the same keys as the annotation
  - We still need our own definition in BasicValueMappers, because our annotation processor doesn't apply to `contrib`. Perhaps a future work question of how we leverage our annotation processor in library code
- I looked into separating `TypePattern` from `ConcreteType`, but ran into difficulties deciding which of the two should be returned from `substitute`. I think keeping them together is prudent
- ParameterTest activity has been updated to use different units for the key and value of `doubleMap`, to allow us to ascertain that the annotations are applied in the right places
- I added `:examples:banananation:assemble` as a dependency to the e2e-tests, so that it always uploads the most up-to-date version

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -27,13 +31,14 @@
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 
 public class AutoValueMappers {
-  private static final Pattern JAVA_IDENTIFIER_ILLEGAL_CHARACTERS = Pattern.compile("[.><,\\[\\]]");
+  private static final Pattern JAVA_IDENTIFIER_ILLEGAL_CHARACTERS = Pattern.compile("[^a-zA-Z0-9_]");
```
Mythicaeda: Why are we locking down the whitelist into just the Latin ASCII characters? Why not permit, for example, á, ñ, or ç?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/ValueSchemaJsonParser.java
```diff
@@ -4,22 +4,27 @@
 import gov.nasa.jpl.aerie.json.JsonParser;
 import gov.nasa.jpl.aerie.json.SchemaCache;
 import gov.nasa.jpl.aerie.json.Unit;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
+import org.apache.commons.lang3.tuple.Pair;
```
Mythicaeda: Minor: Unused import

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/ValueSchemaJsonParser.java
```diff
@@ -4,22 +4,27 @@
...
 
 import javax.json.Json;
 import javax.json.JsonObject;
 import javax.json.JsonValue;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
```
Mythicaeda: Minor: Unused import

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/Mission.java
```diff
@@ -36,8 +39,8 @@ public Mission(final Registrar registrar, final Configuration config) {
 
     registrar.discrete("/flag", this.flag, new EnumValueMapper<>(Flag.class));
     registrar.discrete("/flag/conflicted", this.flag::isConflicted, new BooleanValueMapper());
-    registrar.discrete("/peel", this.peel, new DoubleValueMapper());
-    registrar.real("/fruit", this.fruit);
+    discreteResource(registrar, "/peel", this.peel, new DoubleValueMapper(), "kg");
+    realResource(registrar, "/fruit", this.fruit, "bananas");
     registrar.discrete("/plant", this.plant, new IntegerValueMapper());
```
Mythicaeda: Minor, ignorable: the `plant` resource lost the units assigned to its incon, `Configuration.initialPlant`.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/Banannotation.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.CLASS)
+@Target(ElementType.TYPE_USE)
+@AutoValueMapper.Annotation
+public @interface Banannotation {
+  String value();
+  // double optional() default 0.0; // TODO test support for defaults
+  // int[] repeated(); // TODO add support for arrays
```
Mythicaeda: Minor, can ignore: these should be Issues linked under `Future Work` rather than `TODO`s

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
```diff
@@ -0,0 +1,8 @@
+package gov.nasa.jpl.aerie.banananation;
+
+import gov.nasa.jpl.aerie.banananation.activities.Banannotation;
+import gov.nasa.jpl.aerie.merlin.framework.annotations.MissionModel;
+
+@MissionModel.WithMetadata(name="label", annotation= Banannotation.class)
+public class CustomValueMappers {
+}
```
Mythicaeda: As discussed during the walkthrough: consider permitting the `@MissionModel.WithMetadata` annotation in the `package-info.json`.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/TypeRule.java
```diff
@@ -4,27 +4,34 @@
...
       final ClassName factory,
-      final String method
-  ) {
-    this.head = Objects.requireNonNull(head);
-    this.enumBoundedTypeParameters = Objects.requireNonNull(enumBoundedTypeParameters);
-    this.parameters = Objects.requireNonNull(parameters);
-    this.factory = Objects.requireNonNull(factory);
-    this.method = Objects.requireNonNull(method);
+      final String method,
+      final String label)
```
Mythicaeda: Minor: inconsistency between `label` param in the constructor and `name` field on the Record

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/Resolver.java
```diff
@@ -40,7 +52,12 @@ public Optional<CodeBlock> instantiateMapperFor(final TypeMirror mirror) {
   }
 
   public Optional<CodeBlock> instantiateNullableMapperFor(final TypeMirror mirror) {
-    final var mapperCode = applyRules(createInitialGoal(mirror));
+    final TypePattern initialGoal = createInitialGoal(mirror);
+    final var mapperCode = applyRules(initialGoal);
+
+//    if (mapperCode.isEmpty() && mirror.toString().contains("Label")) {
+//      throw new Error("---" + initialGoal + "---" + mirror + "---" + String.join("---", this.typeRules.stream().map(Record::toString).toList()));
+//    }
```
Mythicaeda: Minor: commented out code

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
```diff
@@ -155,6 +166,23 @@ static ValueSchema ofVariant(final List<Variant> variants) {
     return new VariantSchema(variants);
   }
 
+  /**
+   * Augments a {@link ValueSchema} with the given metadata.
+   * If the underlying ValueSchema is already a MetaSchema, the new MetaSchema will be merged with the underlying one.
+   * In the event of a key collision, the outer MetaSchema's key shadows the inner one.
```
Mythicaeda: For clarity: is the `metadata` parameter the "outer" one?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/TypePattern.java
```diff
@@ -15,71 +16,72 @@
...
-        throw new Error("Cannot construct a pattern for type " + mirror);
+      default -> throw new Error("Cannot construct a pattern for type " + mirror);
     }
+
+    final var annotations = mirror.getAnnotationMirrors();
+    for (int i = annotations.size() - 1; i >= 0; i--) {
+      final var annotation = annotations.get(i);
+      final var className = ClassName.get((TypeElement) annotation.getAnnotationType().asElement());
+      result = new AnnotationPattern(className, Optional.of(annotation), result);
+    }
```
Mythicaeda, line 60: Why are annotations processed in reverse order?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/TypePattern.java
```diff
@@ -15,71 +16,72 @@
...
-  public abstract TypePattern substitute(Map<String, TypePattern> substitution);
-
-  public abstract Map<String, TypePattern> match(TypePattern other) throws UnificationException;
+  TypePattern substitute(Map<String, TypePattern> substitution);
 
-  public abstract boolean isGround();
+  Map<String, TypePattern> match(TypePattern other) throws UnificationException;
 
-  public abstract boolean isSyntacticallyEqualTo(TypePattern other);
+  boolean isGround();
```
Mythicaeda, line 90: What do `isGround`, `render`, `erasure`, and `box` do?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -153,6 +188,123 @@ record ComponentMapperNamePair(String componentName, String mapperName) {}
...
+      final var properties = new ArrayList<Property>();
+      final var necessaryMappers = new HashMap<TypeMirror, String>();
+      for (final var element : annotation.getEnclosedElements()) {
+        if (!(element.getKind() == ElementKind.METHOD)) continue;
+        final var typeMirror = ((ExecutableElement) element).getReturnType();
+        final var elementName = element.toString().replace("(", "").replace(")", "");
+        final var valueMapperIdentifier = JAVA_IDENTIFIER_ILLEGAL_CHARACTERS
+                                              .matcher(typeMirror.toString())
+                                              .replaceAll("_")
+                                          + "ValueMapper";
```
Mythicaeda: Is it possible to get illegal characters in this `toString`? I don't mind doing the regex match and replace for safety, I'm just not clear if how it would occur here, since `typeMirror` is a Java return type.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -153,6 +188,123 @@ record ComponentMapperNamePair(String componentName, String mapperName) {}
...
+                           TypeSpec
+                               .anonymousClassBuilder(CodeBlock.of(""))
+                               .addSuperinterface(
+                                   ParameterizedTypeName.get(ClassName.get(ValueMapper.class), TypeName.get(annotation.asType())))
+                               .addMethod(
+                                   MethodSpec
+                                       .methodBuilder("getValueSchema")
+                                       .addModifiers(Modifier.PUBLIC)
+                                       .returns(ValueSchema.class)
+                                       .addStatement("return $T.STRING", ValueSchema.class)
```
Mythicaeda: Bug on this line

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: The hope was to avoid the cat and mouse game of adding new illegal characters every time we encounter them. You're absolutely right that this whitelist is much too restrictive.

Looks like there are two methods we can use: `Character.isJavaIdentifierStart` and `Character.isJavaIdentifierPart`. I think this method will be sufficiently future-proof:
```java
  /**
   * Turn a given string into a valid Java identifier
   * @param str a string that may or may not be a valid Java identifier
   * @return a string that is a valid Java identifier
   */
  public static String getIdentifier(String str) {
    if (str.length() == 0) throw new IllegalArgumentException("Cannot turn empty string into an identifier");
    final var identifier = new StringBuilder();
    if (Character.isJavaIdentifierStart(str.charAt(0))) {
      identifier.append(str.charAt(0));
    } else {
      identifier.append("_");
    }
    for (int i = 1; i < str.length(); i++) {
      if (Character.isJavaIdentifierPart(str.charAt(i))) {
        identifier.append(str.charAt(i));
      } else {
        identifier.append("_");
      }
    }
    return identifier.toString();
  }
```

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
mattdailis: Yes - in a call like `ValueSchema.withMeta("foo", SerializedValue.of("bar"), target)`, `target` is the inner one, and `foo: bar` is the "outer" one.

Let me know if you think the javadoc could be clearer about this

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/TypePattern.java
mattdailis: We're building up a set of nested annotation patterns - so if we were to add imaginary parentheses to something like `@Unit @Range double`, we would put them like so: `(@Unit (@Range double))`, i.e. `AnnotationPattern(Unit, AnnotationPattern(Range, PrimitivePattern(double)))`. The innermost object needs to be constructed first

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/TypePattern.java
mattdailis: As we discussed verbally, these methods could really use some javadoc. Let me know what you think:

```java
  /**
   * Given a TypePattern, produce a new TypePattern where all type variables whose names are
   * in the given map have been replaced with the corresponding type patterns.
   * @param substitution A map from type variable name to the type pattern with which to replace it
   * @return a TypePattern with the same structure, but with any matching type variables replaced
   * E.g. ClassPattern(List, TypeVariablePattern(T)).substitute(Map.of("T", ClassPattern(Double))) will return
   *   ClassPattern(List, Double)
   */
  TypePattern substitute(Map<String, TypePattern> substitution);

  /**
   * Attempt to match this TypePattern against the given other TypePattern. Throw a UnificationException if they cannot be matched
   * @param other the ground TypePattern to match against this TypePattern
   * @return an assignment of type variables to TypePatterns, if match is successful
   * @throws UnificationException if the two patterns do not match
   */
  Map<String, TypePattern> match(TypePattern other) throws UnificationException;

  /**
   * Whether this TypePattern contains any type variables. Ground is true if there are no type variables present.
   * @return false if there is at least one type variable in this TypePattern. true if there are none.
   */
  boolean isGround();

  /**
   * Whether this TypePattern has the same structure, same types, and same type variable names as the other pattern
   * @param other the other TypePattern to compare with
   * @return true if the two TypePatterns are identical, false otherwise.
   */
  boolean isSyntacticallyEqualTo(TypePattern other);

  /**
   * Represent this TypePattern as a javapoet TypeName
   * @return a javapoet TypeName corresponding to this TypePattern
   *
   * E.g. ClassPattern(List, Double).erasure --> ClassName(List, Double)
   */
  TypeName render();

  /**
   * Represent this TypePattern as a javapoet TypeName without generics
   * @return a javapoet TypeName corresponding to this TypePattern without generics
   *
   * E.g. ClassPattern(List, Double).erasure --> ClassName(List)
   */
  TypeName erasure();

  /**
   * If this TypePattern represents a primitive, return a TypePattern that represents the corresponding Object type.
   * @return a TypePattern that is not a PrimitiveTypePattern
   *
   * E.g. PrimitivePattern(boolean) --> ClassPattern(Boolean)
   *
   * There is no need to recurse: if the top level is not a PrimitiveTypePattern,
   * the entire type pattern is already considered "boxed"
   */
  TypePattern box();
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: What we're trying to do here is turn a string representation of a type into a valid identifier. Unfortunately, types often contain characters that are illegal as Java identifiers - e.g. in `List<Double>`, `<` and `>` are illegal characters for identifiers.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: Great catch - I'll push the fix in a moment

mattdailis:
For ease of review, I've pushed all changes as `fixup` commits (and there's one `revert` commit) - no changes have been made to earlier commits.

I looked into whether `.box()` on arrays should return `this`, or `new ArrayPattern(this.element.box())`, and determined that actually `List<int[]>` is a valid type 😲. I also took a look at `TypeName`'s `box()` definition, and it returns `this` in all cases except for primitives.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
Mythicaeda: I agree. Adjacent topic that's out of scope for this PR: Can we use this method where we interpret the activity type name in `@ActivityType`? IIRR we're currently just assuming that they're valid Java/TS identifiers, and parts of Aerie break if that assumption is false (ie Constraints).

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
Mythicaeda: I think it could be clearer if, rather than using `outer` and `inner`, the javadoc calls out explicitly that `metadata` (or something like `the new MetaSchema`) will overwrite `key` in `target`'s metadata if there's a key collision. This is a nonblocking recommendation, however.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/TypePattern.java
Mythicaeda: This is perfect, thanks! One note: the example in `render`'s JDoc says `.erasure` instead of `.render`

Mythicaeda (approved):

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
mattdailis: Made a ticket for this: https://github.com/NASA-AMMOS/aerie/issues/1223

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/ValueSchema.java
mattdailis: Reworded:

```java
  /**
   * Augments a {@link ValueSchema} with the given metadata.
   * If the underlying ValueSchema is already a MetaSchema, the new MetaSchema will be merged with the underlying one.
   * In the event of a key collision, the given `metadata` takes precedence over the target's `metadata`
   * @return A new {@link ValueSchema} wrapping the original schema.
   */
  static ValueSchema withMeta(final String key, final SerializedValue metadata, final ValueSchema target) {
```

mattdailis:
I added a `DON'T MERGE` label while I try to fix a bug that clipper reported about an interaction between `@AutoValueMapper.Record` and unit annotations

Basically, if you create a record where multiple fields have the same type, and annotate them all with the same annotation, the autogenerated code won't quite be legal Java 🤔 This has to do with the logic that the record autovaluemapper generation code uses to de-duplicate value mappers for the same type.

mattdailis:
Bugfix update: I've added three commits, the first of which should just be a refactor that moves the relevant code into a helper method, and the second and third of which actually change behavior

### Background
The AutoValueMapper.Record generator code needs to declare a static method that takes as parameters the value mappers corresponding to the given record's fields (called "components"). For example:

```java
@AutoValueMapper.Record
record MyRecord(int x, int y, int z) {}
```

Results in:
```java
public static ValueMapper<MyRecord> generated(ValueMapper<Integer> integerValueMapper) { ... }
```

The `integerValueMapper` is then used for each of the components, which in this example all happen to have the same type.

(Original implementation from this PR: [link](https://github.com/NASA-AMMOS/aerie/pull/188#discussion_r905270957))

The ability to re-use the same value mapper across several components was implemented this way: the `HashSet<TypeMirror>` ensures that only one mapper per unique type is added as a parameter:

https://github.com/NASA-AMMOS/aerie/blob/56ddc0f30baf4fa55684096ec886de1242b4b8df/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java#L45-L50

### Bug 1 description

With the addition of metadata annotations, we have a scenario where two components of the same _type_ need different _mappers_, because they may have different annotations. Two components that have the same type as well as the same annotations may share a value mapper.

This commit changes the uniqueness check to use the type pattern's string representation, rather than the `TypeMirror`: [Fix AutoValueMapper.Record's behavior in the presence of multiple annotated components](https://github.com/NASA-AMMOS/aerie/pull/1183/commits/bd26b15c012ebaa97ebfda5d2d473a7220554ae5)

### Bug 2 description

The `getIdentifier` implementation earlier in this PR replaced all illegal characters with `_`. I tested with two unit annotations, one of which was `kg/s`, and the other was `kg*s`, and they were both converted to `kg_s`, and clashed. This commit makes use of the illegal character's UTF-8 value to help generate unique identifiers in this scenario: [Use UTF-8 value to ensure identifier uniqueness](https://github.com/NASA-AMMOS/aerie/pull/1183/commits/823c6294d301b5ef203d8f95be40f2341201265f)

mattdailis:
After chatting with @Mythicaeda, we decided to walk back the value mapper de-duplication logic described above (originally implemented [here](https://github.com/NASA-AMMOS/aerie/pull/188#discussion_r905270957)). Now, record value mappers declare one value mapper dependency per component.

The rationale was primarily to simplify logic, and generate readable parameter names (the annotated generic classes, once run through the `getIdentifier` method, were quite unwieldy)

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/CustomValueMappers.java
mattdailis: We pair-programmed moving WithMetadata to package-info.java

The implication is that mission models will need to include `@WithMetadata` in their own `package-info.java` file, since we cannot provide that out-of-the-box. We should add it to the aerie-mission-model-template repo so that new users don't need to think about this.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ActivityType.java
```diff
@@ -5,14 +5,21 @@
 import java.util.Map;
 
 
-public record ActivityType(String name, Map<String, Parameter> parameters) {
+public record ActivityType(String name, Map<String, Parameter> parameters, ValueSchema computedAttributes) {
+  /**
+   * Create an activity pe with an empty computed attributes value schema.
```
Mythicaeda: Minor, nonblocking: typo
```suggestion
   * Create an ActivityType with an empty computed attributes value schema.
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -167,4 +303,79 @@ private static CodeBlock castIfGeneric(final Element record) {
...
+
+  /**
+   * Turn a given string into a valid Java identifier
+   * @param str a string that may or may not be a valid Java identifier
+   * @return a string that is a valid Java identifier
+   *
+   * Any illegal characters are replaced by `$` followed by their UTF_8 integer representation as a decimal number.
+   * This helps ensure that strings the differ only in illegal characters get unique identifiers. (e.g. "m/s" and "m*s")
+   *
+   * E.g. "<Hello (Worl%d!)" becomes "$60Hello$32$40Worl$37d$33$41"
```
Mythicaeda: Minor, ignorable: this example is no longer accurate (it would become `_$60_Hello_$32__$40_Worl_$37_d_$33__$41_`)

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
```diff
@@ -167,4 +303,79 @@ private static CodeBlock castIfGeneric(final Element record) {
...
+   */
+  public static String getIdentifier(String str) {
+    if (str.length() == 0) throw new IllegalArgumentException("Cannot turn empty string into an identifier");
+    final var strBytes = str.getBytes(StandardCharsets.UTF_8);
+    final var identifier = new StringBuilder();
+    if (Character.isJavaIdentifierStart(str.charAt(0))) {
+      identifier.append(str.charAt(0));
+    } else {
+      identifier.append('_');
```
Mythicaeda, line 323: Minor, ignorable, the leading `_` is unnecessary on the first char in a string

Mythicaeda (approved):

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/AutoValueMappers.java
Mythicaeda: Closing as I mistakenly thought that this was generating `$INT` and not just `INT`

PR #1193, opened 2023-10-12
Make rootfinding robust to functions with discontinuities 

### LINKED ISSUE #1139: Scheduler rootfinding fails with ZeroDerivativeException
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

develop

### Describe the bug

Rootfinding is failing when one of the 2 init values (x1 and x2) leads to simulation exception. It needs to be more robust to this case. Maybe iterate in the domain until you have 2 valid values ? 

### Reproduction

With the 1.12 eurc model. 
Planning horizon : 2030-001 to 2030-002.
First goal: 
```
export default (): Goal => {
    interval: Temporal.Duration.from({hours: 3}),
    activityTemplate:ActivityTemplates.ECM_Power_ON()
  })
}
```
Second goal: 
```
  forEach: ActivityExpression.ofType(ActivityTypes.ECM_Power_ON),
  activityTemplate: ActivityTemplates.ECM_Power_OFF(),
  endsWithin: TimingConstraint.range(WindowProperty.END, Operator.PLUS, Temporal.Duration.from({ minutes: 15})),
})
```

The first goal is satisfied. 
The second goal misses the first two `ECM_Power_OFF` activities because rootfinding is failing. 
It is failing for the first `ECM_Power_OFF` activity because one of the bound is 2030-001T00:00:00. When a power off is simulated at this time, simulation fails (probably because it is a the same time that the power on?). Thus, one of the 2 init value of rootfinding is +inf. Which leads in the end to a derivative equal to 0.  

### Logs

_No response_

### System Info

```shell
develop, clipper eurc model
```

### Severity

Moderate

adrienmaillard:
* **Tickets addressed:** Fixes #1139 
* **Review:** By commit  

## Description

- To specifically handle discontinuities (=simulation exceptions here): when a discontinuity is detected when computing f(x), another x is chosen randomly in the interval of possible x. 
- The optimization discussed in #1046 is applied, that is, assume that the duration function is secretly constant.
- A refactor to separate solving f(x) = y and f(x) = 0 (and the simple coordinate change between the two) to make rootfinding more apparent, clearer 

## Verification

Two tests on discontinuous functions and one test on f(x) = x^2 

## Documentation

None

## Future work

None

### DIFF
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
@@ -440,6 +440,16 @@ public final class Duration implements Comparable<Duration> {
   }
 
   /**
+   * Determine whether this duration lies in [lower bound, upper bound], bounds comprised
+   * @param lowerBound the lower bound, inclusive
+   * @param upperBound the upper bound, inclusive
+   * @return true if lies in the interval, false otherwise
+   */
+  public boolean between(final Duration lowerBound, final Duration upperBound){
+    return (this.noShorterThan(lowerBound) && this.noLongerThan(upperBound));
+  }
+
+  /**
    * Determine whether this duration is longer than another.
    *
    * @see Duration#compareTo(Duration)
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
@@ -1,14 +1,22 @@
 package gov.nasa.jpl.aerie.scheduler;
 
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.random.RandomGenerator;
+import java.util.random.RandomGeneratorFactory;
 
 public class EquationSolvingAlgorithms {
 
   private static final Logger logger = LoggerFactory.getLogger(EquationSolvingAlgorithms.class);
 
-  public record RootFindingResult<T, History>(T x, T fx, History history){}
+  public record FunctionCoordinate<T>(T x, T fx){}
+
+  public record RootFindingResult<T, Metadata>(FunctionCoordinate<T> functionCoordinate, History<T, Metadata> history){}
 
   /**
    * Solves f(x) = y for x in [xLow, xHigh] with confidence interval [yLow, yHigh] around y such that we stop when
@@ -16,26 +24,40 @@ public class EquationSolvingAlgorithms {
    * x0 and x1 are initial guesses for x, they must be close to the solution to avoid diverging
    * It is considered that algorithm is diverging when the iterated value of x goes out of [xLow, xHigh].
    */
-  public interface SecantAlgorithm<T, History>{
-    RootFindingResult<T, History> findRoot(Function<T, History> f,
-                                  History history,
+  public interface SecantAlgorithm<T, Metadata>{
+    RootFindingResult<T, Metadata> findRoot(Function<T, Metadata> f,
+                                  History<T, Metadata> history,
                                   T x0,
-                                  T x1,
                                   T y,
                                   T toleranceYLow,
                                   T toleranceYHigh,
                                   T xLow,
                                   T xHigh,
-                                  int maxNbIterations) throws ZeroDerivativeException, InfiniteDerivativeException, DivergenceException,
-                                                              ExceededMaxIterationException, NoSolutionException;
+                                  int maxNbIterations) throws
+                                                       ZeroDerivativeException,
+                                                       InfiniteDerivativeException,
+                                                       DivergenceException,
+                                                       ExceededMaxIterationException,
+                                                       NoSolutionException;
   }
 
-  public interface Function<T, History> {
-    T valueAt(T x, History historyType);
+  public interface Function<T, Metadata> {
+    T valueAt(T x, History<T, Metadata> history) throws DiscontinuityException;
+  }
+
+  public interface History<T, Metadata>{
+    void add(FunctionCoordinate<T> functionCoordinate, Metadata metadata);
+    List<Pair<FunctionCoordinate<T>, Optional<Metadata>>> getHistory();
+    Optional<Pair<FunctionCoordinate<T>, Optional<Metadata>>> getLastEvent();
+    boolean alreadyVisited(T x);
   }
 
   public static class ZeroDerivativeException extends Exception{
     public ZeroDerivativeException() {}
+  }
+
+  public static class DiscontinuityException extends Exception{
+    public DiscontinuityException(){}
   }
 
   public static class InfiniteDerivativeException extends Exception{
@@ -44,11 +66,6 @@ public class EquationSolvingAlgorithms {
 
   public static class DivergenceException extends Exception{
     public DivergenceException(String errorMessage) {
-      super(errorMessage);
-    }
-  }
-  public static class WrongBracketingException extends Exception{
-    public WrongBracketingException(String errorMessage) {
       super(errorMessage);
     }
   }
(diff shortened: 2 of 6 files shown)


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -439,6 +439,16 @@ public boolean noShorterThan(final Duration other) {
     return this.compareTo(other) >= 0;
   }
 
+  /**
+   * Determine whether this duration lies in [lower bound, upper bound], bounds comprised
+   * @param lowerBound the lower bound, inclusive
+   * @param upperBound the upper bound, inclusive
+   * @return true if lies in the interval, false otherwise
+   */
+  public boolean liesIn(final Duration lowerBound, final Duration upperBound){
```
Mythicaeda: Minor, can ignore:
```suggestion
  public boolean between(final Duration lowerBound, final Duration upperBound){
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -64,79 +82,204 @@ public static class NoSolutionException extends Exception{
...
+    }
+
+    /**
+     * Solves x s.t. f(x) = 0
+     */
+    public RootFindingResult<Duration, Metadata> findRoot(
+        final Function<Duration, Metadata> f,
+        final History<Duration, Metadata> history,
+        final Duration x0,
+        final Duration x1,
```
adrienmaillard: remove x1, we don't need it anymore

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -219,11 +220,10 @@ public void testScheduleExceptionThrowingTask(){
...
-    assertTrue(TestUtility.containsActivity(plan,
-                                            planningHorizon.fromStart("PT120S"),
-                                            planningHorizon.fromStart("PT120S"),
-                                            problem.getActivityType("LateRiser")));
-    assertEquals(3, problem.getSimulationFacade().countSimulationRestarts());
+    assertTrue(TestUtility.atLeastOneActivityOfTypeInRange(plan,
+                                                           Interval.between(Duration.of(1, Duration.MINUTES), Duration.of(2,Duration.MINUTES)),
+                                                           problem.getActivityType("LateRiser")
+                                                           ));
   }
```
adrienmaillard: Update comment and put exact interval boundaries

adrienmaillard: ANd do not lose sim restart counting

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -64,79 +82,204 @@ public static class NoSolutionException extends Exception{
...
+     * @param function the function we are trying to call
+     * @param init the initial x value
+     * @param min the lower bound of the domain of x
+     * @param max the upper bound of the domain of x
+     * @param history the querying history of f
+     * @param maxIteration the maximum number of iteration possible
+     * @return a coordinate (x, f(x)) s.t. f is continuous at x.
+     * @throws ExceededMaxIterationException
+     */
+    private IteratingResult tryValueAtUntilExhaustion(
```
Mythicaeda: Minor, can ignore: I'm still trying to think if there's a more concise name that could be used. Would `valueAround` (or maybe `getValueAround`/`findValueAround`) fit? The `Around` communicates that the method doesn't necessarily return the function's output at `init`.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
```diff
@@ -0,0 +1,167 @@
...
+    if(power == 0){
+      return Duration.of(1, Duration.MICROSECONDS);
+    }
+    final var lv = duration.in(Duration.MICROSECONDS);
+    var res = lv;
+    for(int i = 1; i < power; i++){
+      res = res * lv;
+    }
+    return Duration.of(res, Duration.MICROSECONDS);
+  }
```
Mythicaeda: Minor, can ignore: this method can be simplified into a one liner at its callsite:

```java
final var ret = Duration.of((long) Math.pow(x.in(Duration.MICROSECONDS), 2), Duration.MICROSECONDS);
```

Mythicaeda (commented):
Looks good so far! Awaiting the changes we discussed yesterday.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
```diff
@@ -0,0 +1,167 @@
...
+  final Duration zeroSecond = Duration.of(0, Duration.SECONDS);
+  final static Duration oneSecond = Duration.of(1, Duration.SECONDS);
+  final Duration twoSecond = Duration.of(2, Duration.SECONDS);
+  final Duration threeSecond = Duration.of(3, Duration.SECONDS);
+  final Duration thirtySecond = Duration.of(30, Duration.SECONDS);
+
+  //compared with the testSimpleDiscontinuous, the function is discontinuous for all odd x, rootfinding is hitting
+  //discontinuous values multiple times
+  @Test
+  void testHighlyDiscontinuous()
```
Mythicaeda, line 19: Question: The tests cover `f(x)=2x, x is not odd`, which is a removable discontinuity, but how does the algorithm handle the jump discontinuities or vertical asymptotes? Especially asymptotes, as I can imagine it wouldn't notice jump discontinuities unless the jump wasn't defined on either side (ie `f(x) = floor(x), x != 1` would have no value at 1).

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: Not sure about that. I wanted to call it `infaillibleValueAt` but it can still fail if too many iterations...

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
adrienmaillard: haha indeed.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: `nextValueAt` it is

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
adrienmaillard: added the test

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/UncontrollableDurationTest.java
```diff
@@ -219,10 +220,10 @@ public void testScheduleExceptionThrowingTask(){
     final var plan = solver.getNextSolution().get();
     //Activity can be started in [0, 2m] but this activity will throw an exception if ran in [0, 1m] so it is scheduled at 2m (as being the second bounds the rootfinding tries before search).
     assertTrue(TestUtility.containsActivity(plan,
-                                            planningHorizon.fromStart("PT120S"),
-                                            planningHorizon.fromStart("PT120S"),
+                                            planningHorizon.fromStart("PT1M38.886061S"),
+                                            planningHorizon.fromStart("PT1M38.886061S"),
                                             problem.getActivityType("LateRiser")));
-    assertEquals(3, problem.getSimulationFacade().countSimulationRestarts());
+    assertEquals(4, problem.getSimulationFacade().countSimulationRestarts());
```
Mythicaeda, line 226: Why is there an additional restart now?

adrienmaillard: because before, the second bound (x1) was tested immediately after x0 and was outside of the non-working interval. The randomness needs one more iteraration to find such a point.

Mythicaeda (approved):

PR #1202, opened 2023-10-16
Simplify extension_roles

Mythicaeda:
* **Review:** By commit  

## Description

Simplifies the `extension_roles` table by removing the `id` PK and turning the `extension_id` and `role` column into the PK. This not only makes it consistent with our other association tables, it will avoid potential bugs caused by duplicate rows.

It also restricts the update permissions on this table. As an association table, this table should have no update permissions, however because `role` is a text field that's unable to have an FK against the `user_roles` table, we should permit this column to be edited in order to resolve typos/role name changes.

[There is a sister PR to accommodate the breaking change in the UI](https://github.com/NASA-AMMOS/aerie-ui/pull/941).

## Verification

Validated in Hasura.

## Documentation

The Extensions docs need to be slightly tweaked (the ones in `Planning` provide the table's entire schema, and the `API` docs are returning the `id` column)

### DIFF
--- a/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
+++ b/deployment/hasura/metadata/databases/AerieUI/tables/public_extension_roles.yaml
@@ -29,7 +29,7 @@ insert_permissions:
 update_permissions:
   - role: aerie_admin
     permission:
-      columns: [extension_id, role]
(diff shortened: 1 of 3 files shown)


cohansen (approved):
LGTM!

PR #1206, opened 2023-10-16
Implement Spans.contains eDSL method

### LINKED ISSUE #1113: Add operation to check coexistence of span inside span
Add an binary operation between two Spans collections (A and B) that returns a windows object defined by the following:

- `true` outside of the union of all A spans
- `true` inside any A spans if they contain at least one B span
- `false` inside any A spans that do not contain a B span

Before implementing, investigate the possibility of extending this to require more than one B span.

JoelCourtney:
* **Tickets addressed:** closes #1113
* **Review:** By commit  

## Description
This adds the coexistence-goal equivalent to the constraints edsl. It is meant to serve similar purposes, but is not a meant to be an exact replica of the judgement phase of coexistence goal.

I'm just going to copy-paste the doc-comment I wrote for the method, as a description of what it does:
```ts
/**
 * Creates a windows object that is false when one of these Spans does not contain a child span, and true otherwise.
 * The parents are the callee and the children are the argument, i.e. `parents.contains(children)`.
 * The default requirement of one child per parent can be modified.
 *
 * More concretely, for the expression `A.contains(B)`, the result is:
 * - `true` inside any A spans if (by default) they contain at least one B span
 *   - for counting spans, "contain" means that the entire B span is inside the A span.
 * - `true` (vacuously) outside the union of all A spans
 * - `false` inside any A spans that do not contain a B span
 *
 * The requirement for one child span can be optionally changed by providing the second argument:
 * - `{count: n}` requires *exactly* `n` children per parent.
 * - `{count: {min: n}}` requires at least `n` children per parent.
 * - `{count: {max: n}}` requires at most `n` children per parent
 * - `{duration: {min: d}}` requires a total duration of children of at least `d`
 * - `{duration: {max: d}}` requires a total duration of children of at most `d`
 *
 * Both `count` and `duration` can be provided at the same time
 * (e.g. `{count: 2, duration: {min: Temporal.Duration.from({hours: 1})}}`).
 * Both `min` and `max` can be provided at the same time (e.g. `{count: {min: 1, max: 3}}`.
 *
 * There is no option to require an exact duration, because the implementation uses floating point comparison.
 * If you need an exact duration, you can approximate it by using a small range around the desired value.
 *
 * @param children child spans to check the existence of.
 * @param requirement what to check for in each parent span.
 */
public contains(children: Spans, requirement?: SpansContainsRequirement): Windows;
```

## Verification
I've added the usual AST node and eDSL compilation tests

## Documentation
The methods are documented. This enables a new pattern of constraint, which deserves an addition to the constraints docs; but I've been procrastinating on that because procedural constraints / scheduling is going to require a massive rework of the documentation and it would be a waste of effort.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
@@ -37,7 +37,7 @@ public record SpansContains(Expression<Spans> parents, Expression<Spans> childre
     final var children = this.children.evaluate(results, bounds, environment);
     var falseIntervals = new ArrayList<Interval>();
 
-
+    // Check for count requirements if they exist
     if (this.requirement.minCount.isPresent() || this.requirement.maxCount.isPresent()) {
       final var sortedParents = StreamSupport.stream(parents.spliterator(), true).sorted((l, r) -> l
           .interval()
@@ -70,6 +70,8 @@ public record SpansContains(Expression<Spans> parents, Expression<Spans> childre
         }
       }
     }
+
+    // Check for duration requirements if they exist.
     if (this.requirement.minDur().isPresent() || this.requirement.maxDur.isPresent()) {
       final var accumulatedDuration = children.accumulatedDuration(Duration.MICROSECOND);
       final Optional<Long> minDur;
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
@@ -66,8 +66,8 @@ public record SpansContains(Expression<Spans> parents, Expression<Spans> childre
         var lookaheadChildIndex = childIndex;
         while (lookaheadChildIndex < childrenLength) {
           final var childInterval = sortedChildren.get(lookaheadChildIndex).interval();
-          if (!parentInterval.contains(childInterval)) break;
-          instanceCount.getAndIncrement();
+          if (parentInterval.contains(childInterval)) instanceCount.getAndIncrement();
+          else if (parentInterval.compareEndToStart(childInterval) < 1) break;
           lookaheadChildIndex++;
         }
 
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
@@ -1,0 +1,119 @@
+package gov.nasa.jpl.aerie.constraints.tree;
+
+import gov.nasa.jpl.aerie.constraints.model.EvaluationEnvironment;
+import gov.nasa.jpl.aerie.constraints.model.SimulationResults;
+import gov.nasa.jpl.aerie.constraints.time.Interval;
+import gov.nasa.jpl.aerie.constraints.time.Spans;
+import gov.nasa.jpl.aerie.constraints.time.Windows;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
+
+import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
```diff
@@ -0,0 +1,121 @@
...
+          falseIntervals.add(parentInterval);
+        } else if (this.requirement.maxCount.map(c -> instanceCount.get() > c).orElse(false)) {
+          falseIntervals.add(parentInterval);
+        }
+      }
+    }
+
+    // Check for duration requirements if they exist.
+    if (this.requirement.minDur().isPresent() || this.requirement.maxDur.isPresent()) {
+      final var accumulatedDuration = children.accumulatedDuration(Duration.MICROSECOND);
```
mattdailis, line 84: I know this isn't new in this PR, but could you help me understand what the duration argument to `accumulatedDuration` means?

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
```diff
@@ -0,0 +1,121 @@
...
+          childIndex++;
+          childInterval = sortedChildren.get(childIndex).interval();
+        }
+
+        var instanceCount = new AtomicInteger(0);
+        var transientChildIndex = childIndex;
+        while (parentInterval.contains(childInterval)) {
+          instanceCount.getAndIncrement();
+          transientChildIndex++;
+          childInterval = sortedChildren.get(transientChildIndex).interval();
```
mattdailis: In my attempt at a manual test I ran into an index out of bounds error here

```typescript
export default () => {
  return Spans.ForEachActivity(ActivityType.GrowBanana).contains(Spans.ForEachActivity(ActivityType.BiteBanana));
}
```


Screenshot 2023-10-18 at 11 50 10 AM

```
aerie_merlin  | java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
aerie_merlin  |         at java.base/java.util.ImmutableCollections$ListN.get(Unknown Source)
aerie_merlin  |         at gov.nasa.jpl.aerie.constraints.tree.SpansContains.evaluate(SpansContains.java:63)
aerie_merlin  |         at gov.nasa.jpl.aerie.constraints.tree.SpansContains.evaluate(SpansContains.java:17)
aerie_merlin  |         at gov.nasa.jpl.aerie.constraints.tree.ViolationsOfWindows.evaluate(ViolationsOfWindows.java:25)
aerie_merlin  |         at gov.nasa.jpl.aerie.constraints.tree.ViolationsOfWindows.evaluate(ViolationsOfWindows.java:16)
aerie_merlin  |         at gov.nasa.jpl.aerie.constraints.tree.Expression.evaluate(Expression.java:18)
aerie_merlin  |         at gov.nasa.jpl.aerie.merlin.server.services.ConstraintAction.getViolations(ConstraintAction.java:209)
aerie_merlin  |         at gov.nasa.jpl.aerie.merlin.server.http.MerlinBindings.getConstraintViolations(MerlinBindings.java:239)
```

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
JoelCourtney: We don't have duration profiles, so accumulated duration has to return a "unitless" real profile. The argument is the unit of the result.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/tree/SpansContains.java
JoelCourtney: Thanks, fixed. (and added a test)

mattdailis (approved):
Great work, looks good to me

PR #1208, opened 2023-10-18
Fixed an issue that was causing constraint checking to not load the c…

### LINKED ISSUE #1166: Constraints Behavior Changes Based on if SimDatasetId is Present
**Current Behavior:**
- If a simulation dataset id is present, only the external datasets associated with that sim run are loaded, while plan-wide external datasets are skipped.
- If a simulation dataset id is not present, all external datasets are loaded, including those associated with other sim datasets are loaded.

This causes the behavior of constraints to differ based on whether the SimulationDatasetId is provided.

**Expected Behavior:** regardless of if a sim dataset id is specified, plan-wide external datasets are loaded as well as any external datasets associated with the sim dataset (either the specified one or the latest).

**Proposed Solution:** Merge the `getPlanDatasets` and `getSimAssociatedPlanDatasets` action into one action that gets all plan datasets for the plan where the `sim_dataset_id` field is either `null` or equal to a passed in simDatasetId

We currently do not notice this inconsistent behavior in our e2eTests because the test suite is written as a serial suite, causing the cached results from prior tests to pollute later tests. Additionally, we do not notice the incorrect behavior because our e2eTests incorrectly expect an outdated sim-associated external dataset to have been loaded and caused a violation when checking constraints.

cohansen:
…orrect external datasets

* **Tickets addressed:** Closes #1166
* **Review:** By file

## Description
In order to get the correct external datasets to load, I've removed one of the actions that took a `simulationDatasetId` as a required parameter. We now only have one action that optionally takes the `simulationDatasetId`, if one is not provided then it **only** loads the plan level external datasets. If one **is** provided then it'll load the plan level external datasets and the simulation related external datasets as well.

## Verification
An existing broken test was changed to validate that the correct exception should be thrown.

## Documentation
NA

## Future work
NA


e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -442,25 +442,26 @@ void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
       final var message = exception.getMessage().split("\"message\":\"")[1].split("\"}]")[0];
       // Hasura strips the cause message ("Assumption falsified -- mission model for existing plan does not exist")
       // from the error it returns
-      if (!message.equals("input mismatch exception")) {
+      if (!message.equals("not a valid json response from webhook")) {
```
Mythicaeda: This error message doesn't look correct and indicates that a different issue may have been occurring. The same is true for the next test (L462). Do you have the full error statement?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/utils/HasuraRequests.java
```diff
@@ -584,6 +584,7 @@ public List<ConstraintRecord> checkConstraints(int planID, int simulationDataset
                               .add("simulationDatasetId", simulationDatasetID)
                               .build();
     final var constraintResults = makeRequest(GQL.CHECK_CONSTRAINTS, variables).getJsonArray("constraintViolations");
+    System.out.println(constraintResults);
```
Mythicaeda: PrintLn needs to be removed

Mythicaeda (commented):
Changes overall look good! Only concern is the expected error message change in the updated e2e tests, as missing resources during constraints compilation returned a `input mismatch exception` prior to this PR, and I'm not sure what caused the different (and, bluntly, even less informative) error message. @JoelCourtney, any insight?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
cohansen: Yeah I was confused by that as well, but I double checked recreating the error in the UI and I see the same error as well. I'm not sure if we're getting deeper into the constraint code checking or if something else is going on. I can get the full stack in a bit!

Mythicaeda: I took a quick dive into it and it looks like what happened is that the action will throw an Error on L167/L169 which isn't getting caught and serialized into a proper response by the Merlin Bindings. This causes it to have an invalid format when it returns to Hasura. Prior to this PR, it would be instead thrown as an `InputMismatchException`, which is properly handled. I'm not sure what change caused this, however.

skovati (approved):
Pending the error resolution, this looks great to me!

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -442,7 +442,7 @@ void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
       final var message = exception.getMessage().split("\"message\":\"")[1].split("\"}]")[0];
       // Hasura strips the cause message ("Assumption falsified -- mission model for existing plan does not exist")
```
Mythicaeda: Minor: can this comment be updated to reflect the error message now actually being stripped?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
Mythicaeda: Summary for posterity: Prior to this  PR, codegen would include unrelated external dataset resources, causing compilation to pass but evaluation of the constraint to fail (see DiscreteResource L29 for the original error message), but now compilation is failing because codegen (correctly) is not include those excess resources

Mythicaeda (approved):
Aside from the one test comment update, this LGTM! :shipit:

PR #1212, opened 2023-10-23
Use skip-list multi-map to speed up job queue

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
Changes the data structure for job queue from a priority queue to a skip-list-based multimap. During some profiling for the Clipper model, we found that removing condition jobs from the queue in order to reprioritize them was taking significant time. To mitigate this, we use a skip-list based map to order batches, and store a set for each batch. In practice, this reduced runtime for a 1-year plan with the Clipper model from ~60 seconds to ~20 seconds.

## Verification
Both the clipper model and a toy model were run with this changes while recording the number of jobs run and the number of resource samples taken. These numbers were verified not to change because of this change.

## Documentation
N/A - purely an internal refactor

## Future work
N/A - standalone change

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/JobSchedule.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/JobSchedule.java
@@ -1,15 +1,13 @@
 package gov.nasa.jpl.aerie.merlin.driver.engine;
 
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
-import org.apache.commons.lang3.tuple.Pair;
 
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.PriorityQueue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 public final class JobSchedule<JobRef, TimeRef extends SchedulingInstant> {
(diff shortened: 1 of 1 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/JobSchedule.java
```diff
@@ -1,59 +1,59 @@
...
 
-    if (oldTime != null) this.queue.remove(Pair.of(oldTime, job));
-    this.queue.add(Pair.of(time, job));
+    if (oldTime != null) removeJobFromQueue(oldTime, job);
+
+    this.queue.compute(time, (t, jobsAtNewTime) -> {
+      if (jobsAtNewTime == null) jobsAtNewTime = new HashSet<>();
+      jobsAtNewTime.add(job);
+      return jobsAtNewTime;
+    });
```
mattdailis: David and I chatted a bit and came up with this equivalent, but slightly easier to read version:

```suggestion
    this.queue.computeIfAbsent(time, $ -> new HashSet<>()).add(job);
```

Mythicaeda (approved):
Tested against develop and saw no issues. Thank you for this change!

For future reference, [here is the cheat sheet](https://gist.github.com/iSergius/e06963c6eca0a639023666097227427c) @DavidLegg used to judge which data structure would help most with performance (notice the difference in `remove`'s performance). This could be useful if we have other data structures we want to replace.

Issue #1194, opened 2023-10-12
Compilation errors in one constraint should not prevent checking others


mattdailis:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

There are situations when some constraints are expected to fail to compile, and a user may want to check other constraints nonetheless. An example where this came up recently is that some constraints depend on external datasets which were not yet available - which manifested itself as a compilation error.

### Describe the feature request

This ticket would be satisfied if there was a way that a user could, in the presence of constraints that cannot compile, still run a constraint check for the ones that do compile.

Some options are:
- Always check all constraints, and collect compilation errors and return them in bulk
- Allow selecting which constraints the user wants to check - in which case they can de-select the ones that fail compilation

The former seems like a simpler change to me, but it's worth considering the alternatives.

PR #1216, opened 2023-10-26
[Aerie-1117]  Return compilation errors with constraints

### LINKED ISSUE #1117: Return Compilation Errors When Constraints Fail to Compile
When running Constraints, if a constraint has errors that prevent it from compiling, the Merlin Server does not include these errors in its response. This means that the only way to see what the error was is to open the container's Docker log and look for the exception. These errors should instead be parsed and returned as part of the response payload.

goetzrrGit:
* **Tickets addressed:** Closes #1117
* **Review:** By commit  

## Description
Currently, if a constraint compilation error occurs, the server swallows it, making it difficult for users to develop constraints in Aerie. This is because users cannot see why the constraint is not running or where the error is located.

This PR changes this behavior by capturing compilation errors and sending them back to the UI, which will display them in the UI console. This will allow users to quickly identify and fix compilation errors, improving their constraint development experience.

Additionally, this PR updates the Hasura action to support sending back errors or results per constraint. This will be used in a future PR to stop the Merlin server from short-circuiting and not running the full list of constraints.

We will also need to merge in the UI PR which allows the frontend to support this modified Hasura call
https://github.com/NASA-AMMOS/aerie-ui/pull/967

## Verification
Updated e2e test 

## Documentation
None

## Future work
https://github.com/NASA-AMMOS/aerie/issues/1194

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
@@ -1,6 +1,8 @@
 package gov.nasa.jpl.aerie.e2e;
 
 import com.microsoft.playwright.Playwright;
+import gov.nasa.jpl.aerie.e2e.types.ConstraintError;
+import gov.nasa.jpl.aerie.e2e.types.ConstraintResult;
 import gov.nasa.jpl.aerie.e2e.types.ExternalDataset.ProfileInput;
 import gov.nasa.jpl.aerie.e2e.types.ExternalDataset.ProfileInput.ProfileSegmentInput;
 import gov.nasa.jpl.aerie.e2e.types.ValueSchema;
@@ -100,11 +102,13 @@ public class ConstraintsTests {
   @Test
   void constraintsSucceedOneViolation() throws IOException {
     hasura.awaitSimulation(planId);
-    final var constraintsResults = hasura.checkConstraints(planId);
-    assertEquals(1, constraintsResults.size());
+    final var constraintsResponses = hasura.checkConstraints(planId);
+    assertEquals(1, constraintsResponses.size());
 
     // Check the Result
-    final var constraintResult = constraintsResults.get(0);
+    final var constraintResponse = constraintsResponses.get(0);
+    assertEquals(true,constraintResponse.success());
+    final ConstraintResult constraintResult = constraintResponse.result();
     assertEquals(constraintId, constraintResult.constraintId());
     assertEquals(constraintName, constraintResult.constraintName());
 
@@ -155,7 +159,9 @@ public class ConstraintsTests {
 
     // Check results
     assertTrue(cachedRun.results().isPresent());
-    assertEquals(constraintRuns.get(0), cachedRun.results().get());
+    final var constraintResponse = constraintRuns.get(0);
+    assertEquals(true,constraintResponse.success());
+    assertEquals(constraintResponse.result(), cachedRun.results().get());
   }
 
   @Test
@@ -217,11 +223,13 @@ public class ConstraintsTests {
         "0h",
         Json.createObjectBuilder().add("duration", thirtyFiveDays).build());
     hasura.awaitSimulation(planId);
-    final var constraintResults = hasura.checkConstraints(planId);
-    assertEquals(1, constraintResults.size());
+    final var constraintsResponses = hasura.checkConstraints(planId);
+    assertEquals(1, constraintsResponses.size());
 
     // Check the Result
-    final var constraintResult = constraintResults.get(0);
+    final var constraintResponse = constraintsResponses.get(0);
+    assertEquals(true,constraintResponse.success());
+    final var constraintResult = constraintResponse.result();
     assertEquals(constraintId, constraintResult.constraintId());
     assertEquals(constraintName, constraintResult.constraintName());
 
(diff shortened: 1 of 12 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
Mythicaeda, line 1: Constraints can also fail with an `InputMismatchException`, either during part of the setup before compilation or during evaluation of the constraints. Since this exception is part of constraints, it would make sense for it to also have it's serlializer updated like the success and compilation failure cases

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
Mythicaeda, line 1: In this file, can you add some assertions about the `success` and `errors` fields on the tests where you replaced a `.get` with a `.get.result`?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/CachedConstraintRun.java
```diff
@@ -18,7 +18,7 @@ public static CachedConstraintRun fromJSON(JsonObject json) {
         json.getBoolean("definition_outdated"),
         json.getJsonObject("results").isEmpty() ?
             Optional.empty() :
-            Optional.of(ConstraintRecord.fromJSON(json.getJsonObject("results")))
+            Optional.of(ConstraintResult.fromJSON(json.getJsonObject("results")))
```
Mythicaeda, line 21: Just to confirm: the reason why the Cached Results no longer equal what is returned from the endpoint is because we only want to cache successful constraint runs (as in, there were no errors during execution)? If so, will #1194 cause any complications?

goetzrrGit: Since we weren't storing failed runs in the first place, I don't see this causing any issues. The original behavior would impose constraints that only compiled correctly, preventing compilation errors from being stored in the cache or database. We only care about runs that compiled and executed successfully on the server. With #1194, we'll filter out any errors and only store successful runs to the database and cache..

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
goetzrrGit: Yep, that is correct. I am going to handle all the setup/parsing errors with https://github.com/NASA-AMMOS/aerie/issues/1194 right now I was only focused on sending back a `compilation error` as that seems to be the most helpful.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Sure, I just updated the test to include that check.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -165,9 +166,9 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
         if (constraintCompilationResult instanceof ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Success success) {
           expression = success.constraintExpression();
         } else if (constraintCompilationResult instanceof ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Error error) {
-          throw new ConstraintCompilationException("Constraint compilation failed: " + error);
+          throw new ConstraintCompilationException(entry.getValue().name(),error);
         } else {
-          throw new ConstraintCompilationException("Unhandled variant of ConstraintsDSLCompilationResult: " + constraintCompilationResult);
+          throw new Error("Unhandled variant of ConstraintsDSLCompilationResult: " + constraintCompilationResult);
```
Mythicaeda: I think this makes sense as a ConstraintCompilationException, as it's an error that is explicitly called when Constraints Compilation returned an abnormal state.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -438,13 +455,13 @@ void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
       // This test causes the endpoint to throw an exception when it fails to compile the constraint,
       // as it cannot find the external dataset resource in the set of known resource types.
       // "input mismatch exception" is the return msg for this error
```
Mythicaeda: Minor: The "input mismatch exception" part of the comment needs to be updated

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Made the required change.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: oops, That is what I get for not reading the comment fully. I removed that line.

Mythicaeda (approved):
LGTM! Marking as `Don't Merge` until [the UI PR](https://github.com/NASA-AMMOS/aerie-ui/pull/967) is also ready to merge.

PR #1218, opened 2023-10-26
Add unit-awareness library

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By file

## Description
Adds a QUDV (Quantity, Unit, Dimension, and Value) library, generalized over the type of values. This is part of https://github.com/NASA-AMMOS/aerie/pull/1198, which applies units to numbers, dynamics, and resources, and allows adding units to other types.

In this library, base dimensions are declared at runtime, and derived dimensions are constructed by multiplying and taking powers of base dimensions. A dimension is a product of rational powers of base dimensions, and are represented exactly.

Units are a dimension and a scale. To use them properly, we declare a base unit for each base dimension, which implicitly has a scale of "1", and derive all other units from these through multiplication and powers. Derived units are anonymous by default, but can be given names. Since units are a multiplier and a dimension, we can only represent units with "absolute" scales. (So Kelvin is representable, but degrees Fahrenheit isn't.)

Quantities are represented by the `UnitAware<T>` interface, and represent a value of type `T` and a unit, with the ability to do dimensionality checks and rescale the value to convert to a different unit.

## Verification
There's a unit test for the Dimension class, but other classes have only been spot-checked. I'm working on converting the Clipper model, and parts of it use UnitAware types, so that should provide some test cases.

## Documentation
At the moment, only the inline Javadocs document this code, but we want to make wiki page(s) describing this and other parts of https://github.com/NASA-AMMOS/aerie/pull/1198. I'll update this description when that happens.

## Future work
See https://github.com/NASA-AMMOS/aerie/pull/1198

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/unit_aware/Dimension.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/unit_aware/Dimension.java
@@ -1,0 +1,175 @@
+package gov.nasa.jpl.aerie.contrib.streamline.unit_aware;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import static java.util.Collections.reverseOrder;
+import static java.util.Map.Entry.comparingByValue;
+import static java.util.stream.Collectors.joining;
+
+/**
+ * A kind of quantity, which can be measured.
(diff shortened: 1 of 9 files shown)


DavidLegg:
Closing this as redundant, because #1253 was merged.

PR #1222, opened 2023-10-27
Fix serialization/deserialization issue for controllable durations

### LINKED ISSUE #1220: Activities with controllable duration are not usable if their value mapper is custom
Interesting issue that has not blown up yet because Clipper does not actively use `ControllableDuration` yet but that we see on the external scheduler side. When we have such a parameter that controls the duration, we know that on the model side, it is represented by a `Duration` object. We assumed that the value mapper would our own but actually we don’t know what value mapper the mission model uses. Clipper uses another value mapper. The problem is that we need to be able to serialize/deserialize this parameter. The mappers are well hidden in the mission model and depend on a naming convention/code generation.

One path forward is to provide the value mapper in `DurationType.Controllable`. Problem, `ValueMapper` is in 1) merlin-framework and `DurationType` in 2) merlin-sdk and there is a dependency 1 ->2 but we need the opposite.

Other solution would be to rely on generated code in a similar fashion as `DurationType.ParametricDuration` and generate a new class that would call the value mapper of the controllable duration.

adrienmaillard:
* **Tickets addressed:** Fixes #1220 
* **Review:** By commit  

## Description

First commit: give access to mission model value mapper serialization/deserialization via the `SchedulerModel`.

Second commit is about fixing all the times that durations were assumed to be in a specific schema (long number of microseconds). 

Third commit: I added a (very smart) custom value mapper for durations to the foo mission model (as a test) and had to update 2 or 3 tests to use model serialization to build activities. 

## Verification

Foo has now a custom duration value mapper so all the tests using foo are testing the new pipeline

## Documentation

None. 

## Future work

None.

### DIFF
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
@@ -2,8 +2,8 @@
 
 @WithConfiguration(Configuration.class)
 
-@WithMappers(BasicValueMappers.class)
 @WithMappers(FooValueMappers.class)
+@WithMappers(BasicValueMappers.class)
 
 @WithActivityType(BasicActivity.class)
 @WithActivityType(FooActivity.class)
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
@@ -20,6 +20,7 @@ import javax.json.Json;
 import javax.json.JsonObject;
 import javax.json.JsonValue;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 
@@ -671,21 +672,28 @@ public class SchedulingTests {
                                              .toList();
       assertEquals(1, findFile.size());
 
-      final var resources = """
-          export type Resource = {
-            "/peel": number,
-            "/my_boolean": boolean,
-            "/fruit": {initial: number, rate: number, },
-            "/data/line_count": number,
-            "/flag/conflicted": boolean,
-            "/plant": number,
-            "/flag": ( | "A" | "B"),
-            "/producer": string,
-          };
-          """;
-      for(final var resourcesLine: resources.split("\n")){
-        assertTrue(findFile.get(0).content().contains(resourcesLine));
-      }
+      // Create a list of the exported resource types, one entry per new line
+      final List<String> resourceTypes = Arrays.stream(findFile.get(0)
+                                       .content()
+                                       .split("export type Resource = \\{\\n")[1]
+                                       .split("\\n};")[0] // isolate to export type Resource block
+                                       .split("\\n"))
+                                       .map(String::strip) // remove whitespace
+                                       .toList();
+
+      final var expectedResources = List.of(
+          "\"/data/line_count\": number,",
+          "\"/flag\": ( | \"A\" | \"B\"),",
+          "\"/flag/conflicted\": boolean,",
+          "\"/fruit\": {initial: number, rate: number, },",
+          "\"/my_boolean\": boolean,",
+          "\"/peel\": number,",
+          "\"/plant\": number,",
+          "\"/producer\": string,"
+      );
+
+      assertEquals(expectedResources.size(), resourceTypes.size());
+      assertTrue(resourceTypes.containsAll(expectedResources));
     }
 
     /**
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
@@ -683,7 +683,9 @@ public class SchedulingTests {
             "/producer": string,
           };
           """;
-      assertTrue(findFile.get(0).content().contains(resources));
+      for(final var resourcesLine: resources.split("\n")){
+        assertTrue(findFile.get(0).content().contains(resourcesLine));
+      }
     }
 
     /**
--- a/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/mappers/FooValueMappers.java
+++ b/examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/mappers/FooValueMappers.java
@@ -2,10 +2,15 @@ package gov.nasa.jpl.aerie.foomissionmodel.mappers;
 
 import gov.nasa.jpl.aerie.contrib.serialization.mappers.Vector3DValueMapper;
 import gov.nasa.jpl.aerie.merlin.framework.ValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import org.apache.commons.math3.geometry.euclidean.threed.Vector3D;
 
 public class FooValueMappers {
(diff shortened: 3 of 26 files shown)


mattdailis:
Nice work! Need to mull over some of the code generation, but it seems like you've figured out a way to expose the specific duration value mapper for each activity type. I'm debating whether it makes more sense to do this at the whole mission model level (single value mapper for Duration type), or do this per-activity, and conceivably allow activities with different duration serializations (?)

I'll try this out locally now

mattdailis:
Here's a piece of code that will need to change in this PR: https://github.com/NASA-AMMOS/aerie/blob/43a9ccf1c876d1ea64e73bbdf5f00c9665ff65b0/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java#L431-L437


Perhaps:
```java
if (serializedDuration != null) {
  actDuration = s.valueMapperInterface().deserializeControllableDuration(serializedDuration);
}
```

mattdailis:
A somewhat more insidious issue is that we need to deal with moving duration parameters from typescript into Java.

What I did was I changed the default duration serialization to be a string instead of a number (I wrapped it in `Long.toString()`), and ran tests to see what happened. Many failures were just test code that needed to create activities with duration parameters - those were easy fixes. I also flushed out the issue in the comment above, that's also an easy fix.

The test that is stumping me now is `testSingleActivityPlanSimpleCoexistenceGoal_AllenContains`, which has the following goal:

```typescript
  activityTemplate: (span) => ActivityTemplates.DurationParameterActivity({duration: Temporal.Duration.from({ minutes : 50})}),
  startsWithin: TimingConstraint.bounds(TimingConstraint.singleton(WindowProperty.START).plus(Temporal.Duration.from({ minutes : 5})), TimingConstraint.singleton(WindowProperty.START).plus(Temporal.Duration.from({minutes : 10}))),
  endsWithin: TimingConstraint.bounds(TimingConstraint.singleton(WindowProperty.END).minus(Temporal.Duration.from({ minutes : 10})), TimingConstraint.singleton(WindowProperty.END).minus(Temporal.Duration.from({ minutes : 5})))
})
```

The key part is `ActivityTemplates.DurationParameterActivity({duration: Temporal.Duration.from({minutes : 50})}),`. We need to intercept this serialized activity, and use the duration value mapper to convert it.

Things will be even more fun if the duration is nested somewhere inside a struct or a list parameter. Even if it isn't a ControllableDuration activity, we need to use the mission model's value mapper to be able to instantiate this activity.

With that I think I'll call it a weekend and we can brainstorm what to do about this on Monday 😅 


EDIT: Here is my branch with the experiment described above: `1220-fix-controllable-durations-serialization--dailis`

adrienmaillard:
I looked up usages of `SerializedValue.astInt()` and found even more usages of this assumption in the scheduler. One of those is in the specification of the duration range in an `ActivityExpression`. We expect a profile and we take its value at `Duration.ZERO`. If we keep this, we'd need to use the mission model value mapper. As this has nothing to do with controllable or uncontrollable durations (this is a constraint on the placement of the activity), I think we need a  duration value mapper that is not only available in `DurationType.Controllable` but either at the activity level (in all duration types then), to allow for different value mappers to be used, or at the mission model level (in `SchedulerModel` I believe) as you suggested.

adrienmaillard:
If users want to use several duration value mappers, could we maybe try them all in a mission model level meta vale mapper ? in the same manner as `chooseP` for json parsers.

adrienmaillard:
Continuing brain dump. It seems that we don't want a mission model level serializer because we don't want to pick the first one available. Serialization should be per activity. But deserialization could be at the mission model level (trying all possible duration deserializers).

mattdailis:
> we don't want a mission model level serializer because we don't want to pick the first one available

Hmm, I'm unaware of any use cases for having multiple different value mappers for the same concrete type in a single mission model 🤔 This might actually be difficult to achieve with our annotation processor, since it does use the first matching value mapper that it finds. Do you have a use case in mind?

adrienmaillard:
> Do you have a use case in mind?

Some instrument teams would be used to a certain format for inputting durations? If we can avoid it, it sounds better.

mattdailis:
> Some instrument teams would be used to a certain format for inputting durations? If we can avoid it, it sounds better.

I see... I still think this would not be straightforward to achieve in our current annotation processor - you'd need to define a custom wrapper type around each different subsystem's duration - which would be fine, except that controllable duration is requiring that the parameter's type be a raw `Duration`. Maybe it's this requirement that's the problem, and we need to explore a general controllable parameterized duration (i.e. a method from a desired duration to a set of parameters...?)

> the specification of the duration range in an ActivityExpression

This is one place that's completely in our control - the details of the AST are internal to Aerie, so the mission model has no direct visibility into the serialization format we choose. I think we should be okay keeping this as integer microseconds. Let me know if I'm underthinking this

Mythicaeda:
> Some instrument teams would be used to a certain format for inputting durations? 

I don't think this would be sufficient enough to make a model have multiple ways of serializing durations? As Matt said, this would require each way of inputting to be wrapped, since they would literally be different Duration-esque Classes.

adrienmaillard:
> This is one place that's completely in our control - the details of the AST are internal to Aerie, so the mission model has no direct visibility into the serialization format we choose. I think we should be okay keeping this as integer microseconds. Let me know if I'm underthinking this

It should be okay for now. I foresee that at some point, we'll build something more general to handle temporal constraints (for groups, n-ary) and that we'll need a way to combine "our" kind of durations with the user's kind (maybe the user has built a duration resource...).

adrienmaillard:
> > Some instrument teams would be used to a certain format for inputting durations?
> 
> I don't think this would be sufficient enough to make a model have multiple ways of serializing durations? As Matt said, this would require each way of inputting to be wrapped, since they would literally be different Duration-esque Classes.

Right, let's not allow it then. The debate is still open between per-activity or per-model mappers. We could make it accessible in `SchedulerModel` through a `getDurationValueMapper()` method instead of only in `DurationType.Controllable`.

Mythicaeda:
> The debate is still open between per-activity or per-model mappers

Could you give an example of how multiple per-activity mappers could be created for `Duration`? Or is it that the parameter used for the `Controllable` annotation does not have to be a `Duration`/`long` (which I didn't realize was an option until I checked the docs)?

adrienmaillard:
> Could you give an example of how multiple per-activity mappers could be created for Duration? 

What I did in this PR allows it because it provides the mapper associated to the controllable duration parameter of the activity. In other words, it does not assume that there is only one value mapper for `Duration` for the whole mission model. But we know that in reality, there is only one. 

> Or is it that the parameter used for the Controllable annotation does not have to be a Duration/long (which I didn't realize was an option until I checked the docs)?

I think it has to be a `Duration` on the java side. But Clipper has a different value mapper so in the end, it appears as something else than a long in the UI.

examples/foo-missionmodel/src/main/java/gov/nasa/jpl/aerie/foomissionmodel/package-info.java
```diff
@@ -2,8 +2,8 @@
 
 @WithConfiguration(Configuration.class)
 
-@WithMappers(BasicValueMappers.class)
 @WithMappers(FooValueMappers.class)
+@WithMappers(BasicValueMappers.class)
```
Mythicaeda, line 6: What does this reorder do? My best guess is that it ensures that the `SmartestDurationValueMapper` is used for `Duration`s instead of the one in `BasicValueMappers`. Is that correct?

adrienmaillard: Correct, it seems to pick and keep the first one encountered

Mythicaeda (approved):
LGTM!

PR #1234, opened 2023-11-07
Automatically calculate and cache directive validations

### LINKED ISSUE #502: Catch more activity directive validation errors
This issue is to catch more exceptions during activity directive validation. Generally activity validations should return a `200` unless the base exception is not caught.

Per @mattdailis On the backend - we can consider catching and collecting a list of exceptions that were thrown - although I’m not sure what level of exception we would catch (I don’t think we should catch `Throwable` since that could mask other fatal issues - maybe we could catch `Exception` or `RuntimeException`?)

We should also document in this issue the JSON format of `200`s (e.g. with and without `@Subject`). And the JSON format of `500`s.

This issue is part of this [internal request ticket](https://jira.jpl.nasa.gov/browse/AERIEQS-420). Also please see this [internal Slack thread](https://jpl.slack.com/archives/GJ07FGRPH/p1669664759005979) for more context.

### LINKED ISSUE #1195: Cache activity validation status
Backend implementation ticket for https://github.com/NASA-AMMOS/aerie-ui/issues/281

To implement the desired design, the backend will need to populate a table with the pre-computed results of an activity validation for a given activity directive revision.

There are two main chunks of work to implement this:
- [ ] Create a `jsonb` column in either `activity_directive_validation` or `activity_directive_changelog`, and populate it with the JSON that is returned from `ResponseSerializers.serializeValidationNotices`. This cached validation should be tied to an `activity_directive`'s revision, which is mostly tracked in `activity_directive_changelog`.
- [ ] Figure out some automation (database triggers) to run validation and cache results for every updated / inserted activity.

skovati:
* **Tickets addressed:** closes https://github.com/NASA-AMMOS/aerie/issues/1195 closes #502 
* **Review:** By commit  

## Description

Implements automatic activity directive validation (plus caching of results in DB). This PR adds a new worker thread to `merlin-server` which queries `activity_directive_validations` for any directive that doesn't yet have its `validations` field populated.

The `activity_directive_validations` table is populated by database triggers on `activity_directive` insert / update, that will clear out existing validations for a given activity directive, causing the validation thread to pick up that work, and write fresh validation results.

For performance, the validation thread will poll for new validation requests every 500ms (configurable by env var), batch by mission model, thereby only loading each mission model once per poll. The database action that writes validation results back is also capable of batching writes.

The UI is now able to subscribe as follows to get validations for all activities in the currently selected plan, which it can use to populate the [validation error console](https://github.com/NASA-AMMOS/aerie-ui/issues/281)

```graphql
subscription directiveValidations {
  activity_directive_validations(
    where: {status: {_eq: "complete"}, 
    _and: {plan_id: {_eq: 1}}}
  ) {
    validations
    directive_id
  }
}

```

The `validations` JSON will have variable fields depending on what errors were encountered during validation. The JSON could be parsed roughly as follows:
```javascript
if (!validations.success) {
  // there are errors, need to dig deeper
  const { type, errors } = validations;
 
  // these top-level cases are mutually exclusive
  switch (type) {
    case: "NO_SUCH_ACTIVITY_TYPE":
      const { noSuchActivityType } = errors;
      // add to UI error list
      ...
      break;
    case: "VALIDATION_NOTICES":
      const { validationNotices } = errors;
      // add to UI error list
      ...
      break;
    case: "INSTANTIATION_ERRORS":
      // we have instantiation errors, which have sub error
      // types that are all simultaneously possible
      // (need to check size of all three remaining lists)
      const { missingArguments,
              extraneousArguments,
              unconstructableArguments } = errors;
      // add each list to UI error list
      ...
      break;
  }
}
```

## Verification

So far most of the testing has been performance based, utilizing load tests to ensure performance is sufficient. Manual testing so far has shown promising results, with a single request taking ~20ms:
```
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processing batch for mission model: 1
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processed model batch of size 1 in 17.452291 ms
```

... and batches of ~1000 requests taking ~60ms (once this validation code has moved up a few compilation tiers in the JVM JIT)
```
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processing batch for mission model: 1
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processed model batch of size 445 in 22.684292 ms
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processing batch for mission model: 2
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processed model batch of size 416 in 36.900708 ms
```

torture test of 10000 inserts on a single mission model is still quite quick
```
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processing batch for mission model: 2
[Thread-0] INFO gov.nasa.jpl.aerie.merlin.server.services.ValidationWorker - processed model batch of size 10000 in 134.382375 ms
```

Correctness tests can be added if desired.

## Documentation

API docs need to be created, as querying this validation table might be useful to some API consumers.

## Future work

Unblocks https://github.com/NASA-AMMOS/aerie-ui/issues/936.
Long term, it might make sense to spin this validation work off into it's own container, e.g. `merlin-validator` that can be independently scaled from `merlin-server`.

### DIFF
--- a/merlin-server/sql/merlin/tables/activity_directive.sql
+++ b/merlin-server/sql/merlin/tables/activity_directive.sql
@@ -186,11 +186,19 @@ comment on trigger set_timestamp on activity_directive is e''
 create function activity_directive_set_arguments_updated_at()
   returns trigger
   security definer
-  language plpgsql as $$begin
-    call plan_locked_exception(new.plan_id);
-    new.last_modified_arguments_at = now();
+  language plpgsql as
+$$ begin
+  call plan_locked_exception(new.plan_id);
+  new.last_modified_arguments_at = now();
+
+  -- clear old validations
+  update activity_directive_validations
+    set last_modified_arguments_at = new.last_modified_arguments_at,
+        validations = '{}'
+    where (directive_id, plan_id) = (new.id, new.plan_id);
+
   return new;
-end$$;
+end $$;
 
 comment on function activity_directive_set_arguments_updated_at() is e''
   'Sets the last_modified_arguments_at field of an activity_directive to the current time.';
@@ -202,6 +210,22 @@ execute function activity_directive_set_arguments_updated_at();
 
 comment on trigger set_arguments_timestamp on activity_directive is e''
   'Sets the last_modified_arguments_at field of an activity_directive to the current time.';
+
+create function activity_directive_validation_entry()
+  returns trigger
+  security definer
+  language plpgsql as
+$$ begin
+  insert into activity_directive_validations
+    (directive_id, plan_id, last_modified_arguments_at)
+    values (new.id, new.plan_id, new.last_modified_arguments_at);
+  return new;
+end $$;
+
+create trigger validation_entry_on_insert
+  after insert on activity_directive
+  for each row
+execute function activity_directive_validation_entry();
 
 create function check_activity_directive_metadata()
 returns trigger
--- a/merlin-server/sql/merlin/tables/activity_directive.sql
+++ b/merlin-server/sql/merlin/tables/activity_directive.sql
@@ -191,10 +191,10 @@ $$ begin
   call plan_locked_exception(new.plan_id);
   new.last_modified_arguments_at = now();
 
-  -- clear old validations
+  -- request new validation
   update activity_directive_validations
     set last_modified_arguments_at = new.last_modified_arguments_at,
-        validations = '{}'
+        status = 'pending'
     where (directive_id, plan_id) = (new.id, new.plan_id);
 
   return new;
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
@@ -24,6 +24,7 @@ import gov.nasa.jpl.aerie.merlin.server.services.GetSimulationResultsAction;
 import gov.nasa.jpl.aerie.merlin.server.services.LocalMissionModelService;
 import gov.nasa.jpl.aerie.merlin.server.services.MissionModelService;
 import gov.nasa.jpl.aerie.merlin.server.services.MissionModelService.BulkEffectiveArgumentResponse;
+import gov.nasa.jpl.aerie.merlin.server.services.MissionModelService.BulkArgumentValidationResponse;
 import gov.nasa.jpl.aerie.merlin.server.services.UnexpectedSubtypeError;
 import org.apache.commons.lang3.tuple.Pair;
 
@@ -156,6 +157,41 @@ public final class ResponseSerializers {
         .add("success", JsonValue.FALSE)
         .add("errors", String.format("Internal error: %s", response))
         .build();
+  }
+
+  public static JsonValue serializeBulkArgumentValidationResponseList(final List<BulkArgumentValidationResponse> responses) {
+    return serializeIterable(ResponseSerializers::serializeBulkArgumentValidationResponse, responses);
+  }
+
+  public static JsonValue serializeBulkArgumentValidationResponse(BulkArgumentValidationResponse response) {
+    // TODO use pattern matching in switch statement with JDK 21
+    if (response instanceof BulkArgumentValidationResponse.Success) {
+      return Json.createObjectBuilder()
+                 .add("success", JsonValue.TRUE)
+                 .build();
+    } else if (response instanceof BulkArgumentValidationResponse.Validation v) {
+      return Json.createObjectBuilder()
+                 .add("success", JsonValue.FALSE)
+                 .add("errors", Json.createObjectBuilder()
+                     .add("validationNotices", serializeIterable(ResponseSerializers::serializeValidationNotice, v.notices()))
+                     .build())
+                 .build();
+    } else if (response instanceof BulkArgumentValidationResponse.NoSuchActivityError e) {
+      return Json.createObjectBuilder()
+                 .add("success", JsonValue.FALSE)
+                 .add("errors", Json.createObjectBuilder()
+                     .add("noSuchActivityError", serializeNoSuchActivityTypeException(e.ex()))
+                     .build())
+                 .build();
+    } else if (response instanceof BulkArgumentValidationResponse.InstantiationError f) {
+      return serializeInstantiationException(f.ex());
+    }
+
+    // This should never happen, but we don't have exhaustive pattern matching
+    return Json.createObjectBuilder()
+               .add("success", JsonValue.FALSE)
+               .add("errors", String.format("Internal error: %s", response))
+               .build();
   }
 
   public static JsonValue serializeCreatedDatasetId(final long datasetId) {
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
@@ -172,6 +172,7 @@ public final class ResponseSerializers {
     } else if (response instanceof BulkArgumentValidationResponse.Validation v) {
       return Json.createObjectBuilder()
                  .add("success", JsonValue.FALSE)
+                 .add("type", "VALIDATION_NOTICES")
                  .add("errors", Json.createObjectBuilder()
                      .add("validationNotices", serializeIterable(ResponseSerializers::serializeValidationNotice, v.notices()))
                      .build())
@@ -179,12 +180,15 @@ public final class ResponseSerializers {
     } else if (response instanceof BulkArgumentValidationResponse.NoSuchActivityError e) {
       return Json.createObjectBuilder()
                  .add("success", JsonValue.FALSE)
+                 .add("type", "NO_SUCH_ACTIVITY_TYPE")
                  .add("errors", Json.createObjectBuilder()
                      .add("noSuchActivityError", serializeNoSuchActivityTypeException(e.ex()))
                      .build())
                  .build();
     } else if (response instanceof BulkArgumentValidationResponse.InstantiationError f) {
-      return serializeInstantiationException(f.ex());
+      return Json.createObjectBuilder(serializeInstantiationException(f.ex()).asJsonObject())
+          .add("type", "INSTANTIATION_ERRORS")
+          .build();
     }
 
     // This should never happen, but we don't have exhaustive pattern matching
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetUnvalidatedDirectivesAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetUnvalidatedDirectivesAction.java
@@ -5,6 +5,7 @@ import gov.nasa.jpl.aerie.merlin.server.models.ActivityDirectiveForValidation;
 import gov.nasa.jpl.aerie.merlin.server.models.ActivityDirectiveId;
 import gov.nasa.jpl.aerie.merlin.server.models.MissionModelId;
 import gov.nasa.jpl.aerie.merlin.server.models.PlanId;
+import gov.nasa.jpl.aerie.merlin.server.models.Timestamp;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -19,7 +20,7 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.
 
 public class GetUnvalidatedDirectivesAction implements AutoCloseable {
   private static final String sql = """
-      select ad.id, ad.plan_id, ad.type, ad.arguments, p.model_id
+      select ad.id, ad.plan_id, ad.type, ad.arguments, p.model_id, adv.last_modified_arguments_at
         from activity_directive ad
         join activity_directive_validations adv
           on ad.id = adv.directive_id and ad.plan_id = adv.plan_id
@@ -45,11 +46,13 @@ public class GetUnvalidatedDirectivesAction implements AutoCloseable {
       final var type = results.getString("type");
       final var arguments = getJsonColumn(results, "arguments", activityArgumentsP)
           .getSuccessOrThrow($ -> new Error("Corrupt activity arguments cannot be parsed: " + $.reason()));
+      final var argumentTimestamp = results.getTimestamp("last_modified_arguments_at");
 
       map.computeIfAbsent(modelId, $ -> new ArrayList<>())
          .add(new ActivityDirectiveForValidation(
              new ActivityDirectiveId(directiveId),
              new PlanId(planId),
+             argumentTimestamp,
              new SerializedActivity(type, arguments)
          ));
     }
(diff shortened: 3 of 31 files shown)


merlin-server/sql/merlin/tables/activity_directive.sql
```diff
@@ -183,14 +183,22 @@ execute function activity_directive_set_updated_at();
 comment on trigger set_timestamp on activity_directive is e''
   'Sets the last_modified_at field of an activity_directive to the current time.';
 
-create function activity_directive_set_arguments_updated_at()
+create or replace function activity_directive_set_arguments_updated_at()
```
skovati: Remove this `or replace`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
```diff
@@ -107,4 +103,14 @@ record Success(SerializedActivity activity) implements  BulkEffectiveArgumentRes
...
+
+  sealed interface BulkArgumentValidationResponse {
+    record Success() implements BulkArgumentValidationResponse { }
+    record Validation(List<ValidationNotice> notices) implements
+        BulkArgumentValidationResponse { }
+    record NoSuchActivityError(NoSuchActivityTypeException ex) implements
+        BulkArgumentValidationResponse { }
+    record InstantiationError(InstantiationException ex) implements
+        BulkArgumentValidationResponse { }
+  }
```
mattdailis: 🔵 Extremely minor - I like to make these one-liners when I can
```suggestion
  sealed interface BulkArgumentValidationResponse {
    record Success() implements BulkArgumentValidationResponse { }
    record Validation(List<ValidationNotice> notices) implements BulkArgumentValidationResponse { }
    record NoSuchActivityError(NoSuchActivityTypeException ex) implements BulkArgumentValidationResponse { }
    record InstantiationError(InstantiationException ex) implements BulkArgumentValidationResponse { }
  }
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -158,6 +159,35 @@ public static JsonValue serializeBulkEffectiveArgumentResponse(BulkEffectiveArgu
...
+                 .add("success", JsonValue.FALSE)
+                 .add("errors", Json.createObjectBuilder()
+                     .add("validationNotices", serializeIterable(ResponseSerializers::serializeValidationNotice, v.notices()))
+                     .build())
+                 .build();
+    } else if (response instanceof BulkArgumentValidationResponse.NoSuchActivityError e) {
+      return serializeNoSuchActivityTypeException(e.ex());
+    } else if (response instanceof BulkArgumentValidationResponse.InstantiationError f) {
+      return serializeInstantiationException(f.ex());
+    }
```
mattdailis: 🔶 From a UI parsing perspective, it would be great if the structure of the object was consistent across all cases - I'm not 100% sure what that would look like, but I'll point out some inconsistencies here. Here's a quick survey of the current format:

```js
// Success
{
  "success": true
}
// Instantiation failure
{
  "success": false,
  "errors": {
    "missingArguments": [],
    "extraneousArguments": ["quanitid"],
    "unconstructableArguments": []
   },
  "arguments": {"quantity": 1, "growingDuration": 3600000000}
}
// Validation failure
{
  "success": false,
  "errors": {
    "validationNotices": [{"message": "Quantity must be positive", "subjects": ["quantity"]}]
  }
}
// No such activity type
{
  "message": "no such activity type",
  "activity_type": "GrowlBanana"
}
```

Some inconsistencies:
- The "no such activity type" response lacks a `success` property, and in the `"success": false` case it lacks an `errors` property
- While both instantiation failures and validation failures share an `errors` property, the structure of the `errors` object differs, and there's no way to figure out which is which except by checking for the presence or absence of fields. Perhaps we could add `"type": "INSTANTIATION_FAILURE"` and `"type": "VALIDATION_FAILURE"` respectively
- `"arguments"` shows up in just one of the four types

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -158,6 +159,35 @@ public static JsonValue serializeBulkEffectiveArgumentResponse(BulkEffectiveArgu
...
+
+    return Json.createObjectBuilder()
+               .add("success", JsonValue.FALSE)
+               .add("errors", String.format("Internal error: %s", response))
```
mattdailis, line 197: 🔵 I'd be okay with treating this code as "this should never happen" and leaving it out of the consistent error format discussion. Feel free to spruce it up if you'd like, though

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetUnvalidatedDirectivesAction.java
```diff
@@ -0,0 +1,68 @@
...
+      if (!map.containsKey(modelId)) {
+        map.put(modelId, new ArrayList<>());
+      }
+
+      map.get(modelId)
+         .add(new ActivityDirectiveForValidation(
+             new ActivityDirectiveId(directiveId),
+             new PlanId(planId),
+             new SerializedActivity(type, arguments)
+         ));
```
mattdailis: 🔵 As minor as it gets
```suggestion
      map.computeIfAbsent(modelId, $ -> new ArrayList<>())
         .add(new ActivityDirectiveForValidation(
             new ActivityDirectiveId(directiveId),
             new PlanId(planId),
             new SerializedActivity(type, arguments)
         ));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresMissionModelRepository.java
```diff
@@ -146,20 +146,29 @@ public void updateResourceTypes(final String missionModelId, final Map<String, R
...
+  public void updateDirectiveValidations(List<Pair<ActivityDirectiveForValidation, MissionModelService.BulkArgumentValidationResponse>> updates) {
+    try (final var connection = this.dataSource.getConnection()) {
+      try (final var updateAction = new UpdateActivityDirectiveValidationsAction(connection)) {
+        for (final var update : updates) {
+          final var directive = update.getLeft();
+          final var validation = update.getRight();
+          updateAction.addBatch(directive.id(), directive.planId(), validation);
+        }
+        updateAction.executeBatch();
+      }
```
mattdailis: 🔵 These two "try" statements can be combined into one
```suggestion
    try (final var connection = this.dataSource.getConnection();
         final var updateAction = new UpdateActivityDirectiveValidationsAction(connection)) {
      for (final var update : updates) {
        final var directive = update.getLeft();
        final var validation = update.getRight();
        updateAction.addBatch(directive.id(), directive.planId(), validation);
      }
      updateAction.executeBatch();
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/UpdateActivityDirectiveValidationsAction.java
```diff
@@ -27,17 +22,30 @@ public UpdateActivityDirectiveValidationsAction(final Connection connection) thr
     this.statement = connection.prepareStatement(sql);
   }
 
-  public void apply(final long directiveId, final long planId, final Timestamp argumentsModifiedTime, final List<ValidationNotice> notices)
+  public void apply(final ActivityDirectiveId directiveId, final PlanId planId, BulkArgumentValidationResponse results)
```
mattdailis: 🔶 If I understand correctly, this method is now unused? If so, we should consider deleting it.

🔵 Alternatively, `apply` could take in the whole list of updates, and have a for loop for updating the batch - this would be a bit more in line with some other actions (e.g. see `InsertActivityTypes`)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
```diff
@@ -0,0 +1,58 @@
...
+
+public record ValidationWorker(LocalMissionModelService missionModelService) {
+
+  private static final Logger logger = LoggerFactory.getLogger(ValidationWorker.class);
+
+  public void workerLoop() {
+    logger.info("validation worker starting...");
+    while (!Thread.interrupted()) {
+      try {
+        Thread.sleep(500);
```
mattdailis: 🔶 Since it doesn't hurt, let's make this a configurable environment variable

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
```diff
@@ -0,0 +1,58 @@
...
+
+        final var validationRequests = missionModelService.getUnvalidatedDirectives();
+        logger.info(
```
mattdailis: 🔶 The log feels a little noisy - maybe consider logging only when the number of validations is greater than zero? Or moving the log level to `debug`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
```diff
@@ -0,0 +1,58 @@
...
+
+          // write validations out to DB
+          missionModelService.updateDirectiveValidations(zippedList);
+          final var endTime = System.nanoTime();
+          final var duration = (endTime - beginTime) / 1_000_000.0;
+          logger.info("processed model batch of size {} in {} ms", unvalidatedDirectives.size(), duration);
+        }
+
+      } catch (NoSuchMissionModelException ex) {
+        throw new RuntimeException("Validation request failed due to missing mission model", ex);
```
mattdailis: 🔶 I'm curious about what happens when this exception is thrown (or if any other uncaught exception occurs while checking validations). Does the thread die? I'd like to dig into our fault behavior here a little bit to make sure we understand it

mattdailis (commented):
> Key:
> 🔻 Issues to address before merge
>  🔶 Requests that should not block merge, but should at least be discussed 
> 🔵 Recommendations that can be ignored if desired

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
skovati: intellij formatting 🙄, change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
skovati: Agreed. I originally wanted to reuse the existing `ResponseSerializers` functions, but unifying the response schema is worth adding some new serializers.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetUnvalidatedDirectivesAction.java
skovati: Clean. Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresMissionModelRepository.java
skovati: Change has been made to both functions added in this file.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/UpdateActivityDirectiveValidationsAction.java
skovati: I debated passing a list in, but figured it might be useful to keep around a version that acted on a single validation. Thinking about it more, I suppose the bulk version can just be passed a list of size 1 if we did want this functionality. Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
skovati: Change has been made.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
skovati: Good call, I've implemented a conditional check to only log when requests are `> 0`, and moved these log statements to `debug`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
skovati: Good question. On second thought, I've changed this case to an `error` log instead, although I can't see a case where this exception actually gets thrown, due to DB foreign key constraints.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
skovati: I've made a small change just to the "no such activity type" case that puts this error under an `errors` field like the other responses. I figured that the UI will either have to switch on fields being null or `type === "SOME_STRING"`, but this way we'll be storing (marginally) less data in the validation results for each directive in the DB.

skovati: After chatting with the UI folks, they agreed that switching on a string v.s. nullity of fields is easier, so I've added a enum style string to the schema for each of these error results.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
Mythicaeda: > I can't see a case where this exception actually gets thrown, due to DB foreign key constraints

`null` is permitted for a plan's `model_id`. In fact, if you delete a mission model, all plans using that model will automatically have their `model_id` set to `null`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
mattdailis: There's a file called `Environment.md` - let's make sure the two new environment variables are documented there

skovati: Updated!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java
skovati: Furthermore, I've added a `catch (Throwable)` to this error handling, which should keep this validation thread running regardless of exceptions thrown.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
skovati: I'll leave this as is for now, since Java doesn't know this case can't happen and needs it to compile.

mattdailis (approved):

PR #1235, opened 2023-11-07
Revert "[Aerie-1117]  Return compilation errors with constraints"

goetzrrGit:
Reverts NASA-AMMOS/aerie#1216


Mythicaeda (approved):

goetzrrGit:
We merged in the UI today so no need to revert.

PR #1242, opened 2023-11-09
Return all Constraint during Compilation

### LINKED ISSUE #1194: Compilation errors in one constraint should not prevent checking others

goetzrrGit:
* **Tickets addressed:** Closes #1194 Closes https://github.com/NASA-AMMOS/aerie-ui/issues/942
* **Review:** By commit  

## Description
Revise the ConstraintAction to avoid immediate termination and exception throwing in the event of a compilation error with a constraint. Instead, store these errors in a Failable object, enabling us to retrieve the information later for all constraints. Exceptions are retained for situations where it's illogical to proceed with checking a list of constraints without a plan, mission model, or simulation dataset.

I've also addressed an issue where a constraint previously compiled and ran successfully, but encountered compilation issues after a user modification. Now, when the user rectifies the constraint and it successfully compiles and runs, the SQL throws an exception saying the constraint already exists is `primary_id`. Now we do an update on collision.

## Verification
A lot of manual testing and e2e test.

## Future work
Add badges to constraints that have compilation errors.


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -49,26 +37,31 @@ public ConstraintAction(
     this.simulationService = simulationService;
   }
 
-  public List<ConstraintResult> getViolations(final PlanId planId, final Optional<SimulationDatasetId> simulationDatasetId)
-      throws NoSuchPlanException, MissionModelService.NoSuchMissionModelException, SimulationDatasetMismatchException , ConstraintCompilationException{
+  public List<Failable<?>> getViolations(final PlanId planId, final Optional<SimulationDatasetId> simulationDatasetId)
```
Mythicaeda, line 40: Why is this `Failable<?>` rather than `Failable<ConstraintResult>`?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -152,30 +147,58 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
         final Expression<ConstraintResult> expression;
 
         // TODO: cache these results, @JoelCourtney is this in reference to caching the output of the DSL compilation?
-        final var constraintCompilationResult = constraintsDSLCompilationService.compileConstraintsDSL(
-            plan.missionModelId,
-            Optional.of(planId),
-            Optional.of(simDatasetId),
-            constraint.definition()
-        );
+        ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult constraintCompilationResult = null;
```
Mythicaeda: Minor: `constraintCompilationResult` can be final
```suggestion
        final ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult constraintCompilationResult;
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -152,30 +147,58 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
+        } catch (MissionModelService.NoSuchMissionModelException ex) {
+          constraintResultMap.put(
+              constraint.id(),
+              Failable.failure(new Error("Constraint " + constraint.name() + ": " + ex.getMessage())));
+          continue;
+        } catch (NoSuchPlanException ex) {
+          constraintResultMap.put(
+              constraint.id(),
+              Failable.failure(new Error("Constraint " + constraint.name() + ": " + ex.getMessage())));
+          continue;
```
Mythicaeda: Minor: These catches are identical and should be collapsed

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -152,30 +147,58 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
         } else {
-          throw new ConstraintCompilationException(constraint.name(),new ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Error(new ArrayList<>(){{
-            add(new ConstraintsCompilationError.UserCodeError(
-                "Unhandled variant of ConstraintsDSLCompilationResult: " + constraintCompilationResult,
-                "",
-                new ConstraintsCompilationError.CodeLocation(0, 0),
-                ""));
-          }}));
+          final var finalConstraintCompilationResult = constraintCompilationResult;
```
Mythicaeda: Minor: This variable can be removed and its use inlined onto L194

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -152,30 +147,58 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
+                  new ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Error(
+                      new ArrayList<>() {{
+                        add(new ConstraintsCompilationError.UserCodeError(
+                            "Unhandled variant of ConstraintsDSLCompilationResult: "
+                            + finalConstraintCompilationResult,
+                            "",
+                            new ConstraintsCompilationError.CodeLocation(
+                                0,
+                                0),
+                            ""));
```
Mythicaeda: Minor, ignorable: formatting
```suggestion
                      new ArrayList<>() {{
                        add(new ConstraintsCompilationError.UserCodeError(
                            "Unhandled variant of ConstraintsDSLCompilationResult: " + finalConstraintCompilationResult,
                            "",
                            new ConstraintsCompilationError.CodeLocation(0,0),
                            ""));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -77,16 +70,18 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
       throw new RuntimeException("Assumption falsified -- mission model for existing plan does not exist");
     }
 
-    final var violations = new HashMap<Long, ConstraintResult>();
 
-    final var validConstraintRuns = this.constraintService.getValidConstraintRuns(constraintCode.values().stream().toList(), simDatasetId);
+    final var validConstraintRuns = this.constraintService.getValidConstraintRuns(constraintCode
+                                                                                      .values()
+                                                                                      .stream()
+                                                                                      .toList(), simDatasetId);
```
Mythicaeda, line 77: Minor, ignorable: this and L49-52 and L57-60 are reformattings that don't appear to improve readability. If these were intentional, feel free to ignore this comment

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -152,30 +147,58 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
+              Failable.failure(new Error("Constraint " + constraint.name() + ": " + ex.getMessage())));
+          continue;
+        } catch (NoSuchPlanException ex) {
+          constraintResultMap.put(
+              constraint.id(),
+              Failable.failure(new Error("Constraint " + constraint.name() + ": " + ex.getMessage())));
+          continue;
+        }
 
+        // Try to compile the constraint and caputure failure
```
Mythicaeda: Minor: typo
```suggestion
        // Try to compile the constraint and capture failures
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -217,22 +247,49 @@ public List<ConstraintResult> getViolations(final PlanId planId, final Optional<
...
+          constraintResult.constraintId = entry.getKey();
+          constraintResult.constraintType = entry.getValue().type();
+          constraintResult.resourceIds = List.copyOf(names);
+        }
 
-        constraintResult.constraintName = entry.getValue().name();
-        constraintResult.constraintId = entry.getKey();
-        constraintResult.constraintType = entry.getValue().type();
-        constraintResult.resourceIds = List.copyOf(names);
+        constraintResultMap.put(entry.getKey(), Failable.of(constraintResult));
```
Mythicaeda: To confirm my understanding, this change means we are now explicitly storing constraints that threw no violations?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
```diff
@@ -15,6 +15,8 @@
   private static final @Language("SQL") String sql = """
     insert into constraint_run (constraint_id, constraint_definition, simulation_dataset_id, results)
     values (?, ?, ?, ?::json)
+    on conflict (constraint_id, constraint_definition, simulation_dataset_id) do update
+    set results = excluded.results;
```
Mythicaeda: Major: This change does not fix the underlying bug, which is that `definition_outdated` is not being set to `false` in the DB when the constraint definition is back up to date.

The correct fix is to update the function `constraint_check_constraint_run` to

```sql
create function constraint_check_constraint_run()
  returns trigger
  security definer
  language plpgsql as $$
begin
  update constraint_run
  set definition_outdated = true
  where constraint_id = new.id
    and constraint_definition != new.definition
    and definition_outdated = false;

  update constraint_run
  set definition_outdated = false
  where constraint_id = new.id
    and constraint_definition = new.definition
    and definition_outdated = true;
  return new;
end$$;
```
This will now correctly update all rows in the constraint run table, rather than only ever running the first statement (which occurred because these statements were incorrectly wrapped in an `if-else` block)

Additionally, could you add an e2eTest for this fix?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
```diff
@@ -32,7 +34,7 @@ public void apply(
       statement.setString(2, constraint.definition());
       statement.setLong(3, simulationDatasetId);
 
-      if (results.get(constraint.id()) != null) {
+      if (results.get(constraint.id()) != null && !results.get(constraint.id()).isEmpty()) {
```
Mythicaeda: Could you explain what this change is doing? It's part of the bug-fix commit, but it does not seem related to the fix.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -402,12 +445,12 @@ public static JsonValue serializeInvalidJsonException(final InvalidJsonException
                .build();
   }
 
-  public static JsonValue serializeConstraintCompileException(final ConstraintCompilationException ex){
+  public static JsonValue serializeConstraintCompileErrors(final Failable<ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Error> ex){
 
-    final var userCodeError = ex.getErrors().errors().stream()
+    final var userCodeError = ex.getOrNull().errors().stream()
```
Mythicaeda: `ex.getOrNull.errors` could cause a NullPointerException. Not sure what to rank this as I don't know the likelihood of `ex.value` being `null`, but I believe using the `getOptional` method and changing the serialization based on if it's present will avoid this

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -137,7 +137,7 @@ void constraintsSucceedNoViolations() throws IOException {
     hasura.deleteActivity(planId, activityId);
     hasura.awaitSimulation(planId);
     final var constraintsResults = hasura.checkConstraints(planId);
-    assertEquals(0, constraintsResults.size());
+    assertEquals(1, constraintsResults.size());
```
Mythicaeda: What is this new result? Can you add asserts to describe it's properties?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -258,7 +258,8 @@ void runConstraintsOnOldSimulation() throws IOException {
 
     // Expect no violations on the new simulation
     final var newConstraintResults = hasura.checkConstraints(planId, newSimDatasetId);
-    assertEquals(0, newConstraintResults.size());
+    assertEquals(1, newConstraintResults.size());
+    assertEquals(null, newConstraintResults.get(0).result());
```
Mythicaeda: Why does `checkConstaints` now return `NULL` on success?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: This occurs because we store a subset of Java Errors as Failable objects, which are parsed when constructing the response to be sent back.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: I can't make this final because attempt to modify it inside a try-catch block which isn't allowed.

Mythicaeda: If you don't initialize it to `null`, it won't be considered a modification when you initialize it in the try-catch. We do the same thing with `simDatasetId` on L56 of this file -- declare it as `final` and initialize it later in the `if-else`.

goetzrrGit: ahhhh, I didn't see your suggestion removed the null.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: The way it used to work was that the loop would `continue,` and later when we hit `constraintService.createConstraintRuns` at line 281, it would end up storing an empty {} in the database. Personally, I wanted to hang onto these entries in the `constraintResultMap` for debugging `compiledConstraintMap` but still have things work the same way.

In`InsertConstraintRunsAction.java` we are throwing out the constraint that has no violations or is empty and insert `{}`

```
if (results.get(constraint.id()) != null && !results.get(constraint.id()).isEmpty()) {
        statement.setString(4, constraintResultP.unparse(results.get(constraint.id())).toString());
      } else {
        statement.setString(4, "{}");
      }
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: Got it, I'll revert my changes to the SQL. Just to clarify, given my limited knowledge of databases, is it considered better practice to embed logic like this directly in the database table rather than attempting to achieve the same action through SQL calls?

goetzrrGit: I will add an e2e test as well

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: I retain empty results to indicate a successful compilation of the constraint without any violations. At this point, I store constraints with violations in the database; otherwise, I store {}.

I should have broken this up into a separate commit.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Sure, I can modify the test. In this particular scenario, the constraint was successfully compiled and executed without any violations. The object we're returning appears as follows:

```js
{
status : success,
results: {},
errors: []
}
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: The data we get back from the server is below:

```js
{
status : success,
results: {},
errors: []
}
```

Converting this to java

```java
ConstraintRecord[success=true, result=null, errors=[]]
```

I made the conversion between an empty object from JS to java as `null`. 

I can change this to optional unless you have another suggestion?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: > Just to clarify, given my limited knowledge of databases, is it considered better practice to embed logic like this directly in the database table rather than attempting to achieve the same action through SQL calls?

If by "logic like this", you mean the choice to use an upsert vs fixing the function, it is better practice to fix the bugged function, as by not doing so you're actually redoing work that has already been done (that's why you were seeing the conflict here -- Constraints will only upload the results it didn't get from the cache)

If by "logic like this", you mean having the `definition_outdated` column, for constraints caching we decided it was clearer to have a boolean on `constraint_run` indicating if the definition it had run with was up-to-date rather than having external tools join on the `constraints` table in order to figure that out. For keeping this boolean up to date, it is much more efficient to use a trigger on the `constraints` table than trying to make SQL calls from external Java code.

Did that explain it or did I misunderstand the question?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: Is this change why the E2E tests now contain a `results` object in the successful cases?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
Mythicaeda: It should be a proper `ConstraintResult` object where the `violations` field is an empty List (as, for example, there still could've been gaps)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: Actually, thinking about it, a constraint run with no violations should have the exact same data structure as a constraint run with violations, however, the `violations` key should be an empty array.

goetzrrGit: Yes that is correct. There are 3 object that can be returned in the response array. This change affected the e2e test:

compiled but no violations/gaps
```js
{
success: true
result : {}
errors : []
}
```

compiled with violations
```js
{
success: true
result : {
   violations: [...],
   constraintID: 1,
   constriantName: "",
   gaps: [...]
}
errors : []
}
```

failed compiled
```js
{
success: false,
result : {},
errors : [
  {
        message: "",
        location: {}
        stack: ""
  }
]
}
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: You did, thank you, for answering the question.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: My intelij auto-formated things, so it was sort of intentional as I turned this feature on  a long time ago.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Done.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Ignore.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: ignore

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
goetzrrGit: Done

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: Done, Updated the `up.sql` and removed the modified sql expression

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
goetzrrGit: Done, Used optional

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Done, Updated the test

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: Added an e2e test for this bug.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Done, Return a ConstraintResult object

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: The `up.sql` can't be updated, as we've already released versions of Aerie with that file. Instead, it needs to be done in a new migration.

Mythicaeda: Additionally, the actual function in the DB needs to be changed (it is in `constraint.sql`).

Mythicaeda: For an explanation for future reference: when DB code changes, it needs to be reflected both in the actual `.sql` files (so that new DBs are initialized with the most up-to-date-state), and in a new migration (so that existing DBs can be caught up). Generally, we don't update old migrations, especially those that have already been included in a release, unless it's to fix a bug that would prevent the migration from running.

goetzrrGit: Got it, thank you for letting me know, I wasn't even thinking about the constraint.sql :P

Mythicaeda: The `up.sql` changes for migration 20 needs to be reverted, and the changes made to `constraint.sql` need to be applied in a new migration file (It will be either migration 33 or 34 depending on the relative ordering of this PR and the [Cancel Running Simulations PR](https://github.com/NASA-AMMOS/aerie/pull/1243)). Once that is done, you need to add mark the new migration file as applied in the `applied_migrations` file for Merlin.

Doing this should also resolve the function definition mismatch error causing the Migration Checks to fail. (Note that this comparison is whitespace-sensitive)

goetzrrGit: Took a couple of tires but I got it working. Sorry for the misunderstanding and delay.

Mythicaeda: The file `merlin-server/sql/merlin/applied_migrations.sql` needs the `call migrations.mark_migration_applied('34');` statement added there as well.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
Mythicaeda: Forgot to respond to this, but compiled with no violations or gaps should look like:

```json
{
"success": true
"result" : {
  "violations:" [],
  "constraintID": 1,
  "constriantName": "",
  "gaps": []
}
"errors": []
}
```

This keep's the response's structure consistent in the `success: true` case and does not cause information not related to violations or gaps to be lost.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: The catches were collapsed incorrectly, as we likely don't want to catch every possible exception there. What I meant by "collapsing" was the following syntax:

```java
catch (MissionModelService.NoSuchMissionModelException | NoSuchPlanException ex)
```

goetzrrGit: Ohh, got it. I changed it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: Now it is fully done :)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertConstraintRunsAction.java
goetzrrGit: Got it, I changed the code so it won't drop this information and send it back to the ui. 

https://github.com/NASA-AMMOS/aerie/pull/1242/files#diff-e4402938fdcefa189f062abfca8a610f16d1dd9272ccdaeb0c4b6ca59b383d4bL220

merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -83,23 +83,20 @@ before update on "constraint"
...
-      and constraint_definition != new.definition
-      and definition_outdated = false;
-  else
-    update constraint_run
-    set definition_outdated = false
-    where constraint_id = new.id
-      and constraint_definition == new.definition
-      and definition_outdated = true;
-  end if;
+create or replace function constraint_check_constraint_run()
```
Mythicaeda: Minor: This should be just `create` here.
```suggestion
create function constraint_check_constraint_run()
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,7 +487,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+      var constraintsResponses = hasura.checkConstraints(planId);
+
+      assertEquals(1, constraintsResponses.size());
+      assertEquals(true, constraintsResponses.get(0).success());
+
+      hasura.updateConstraint(
+          constraintId,
+          constraintDefinition+"compilation error");
+
+      var constraintsErrorResponses = hasura.checkConstraints(planId);
```
Mythicaeda: Minor, ignorable: this variable can be `final`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,7 +487,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+
+      assertEquals(1, constraintsErrorResponses.size());
+      assertEquals(false, constraintsErrorResponses.get(0).success());
+      assertEquals(3 , constraintsErrorResponses.get(0).errors().size());
+
+      hasura.updateConstraint(
+          constraintId,
+          constraintDefinition);
+      constraintsResponses = hasura.checkConstraints(planId);
```
Mythicaeda: Since this and the initial assignment should be being compared for equality, they should use separate variables rather than reusing `constraintsResponse`. 

Additionally, if separate variables are used, then this equality assertion can become very strong while remaining flexible to future changes by altering the  `assertEquals` on L521 to `assertEquals(initialResponse.get(0), newResponse.get(0))`.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,7 +487,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
       assertEquals(1,constraintRepsonse.errors().size());
       final ConstraintError error = constraintRepsonse.errors().get(0);
-      assertEquals("Constraint 'fruit_equal_peel': TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+      assertEquals("Constraint 'fruit_equal_peel' compilation failed:\n"
+                   + " TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+
+    }
+
+    @Test
+    @DisplayName("A modified and recompiled constraint that originally compiled successfully.")
```
Mythicaeda: This `DisplayName` doesn't communicate what the test is testing. 

This test is actually testing that, if a constraint is updated and then reverted, Constraints will load the cached value from before the update.

merlin-server/sql/merlin/tables/constraint.sql
goetzrrGit: Got it, I only updated `constraint.sql` and keep `up/down.sql` the same as It made sense to me to `create or replace` the migration file.

Mythicaeda: Yup, that was exactly what I meant 🚀

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -183,7 +187,7 @@ void constraintCachedNoViolations() throws IOException {
     assertEquals(constraintDefinition, cachedRun.constraintDefinition());
 
     // Check results
-    assertTrue(cachedRun.results().isEmpty());
+    assertTrue(!cachedRun.results().isEmpty());
```
Mythicaeda: This check needs to be fleshed out to validate the cached results are correct

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
```diff
@@ -14,7 +15,7 @@ public record ConstraintRecord(
   public static ConstraintRecord fromJSON(JsonObject json){
     return new ConstraintRecord(
         json.getBoolean("success"),
-        json.getJsonObject("results").isEmpty() ? null : ConstraintResult.fromJSON(json.getJsonObject("results")),
+        json.getJsonObject("results").isEmpty() ? new ConstraintResult(-1,"","",new ArrayList<>(), new ArrayList<>(),new ArrayList<>()) : ConstraintResult.fromJSON(json.getJsonObject("results")),
```
Mythicaeda: Minor: This change should be reverted, as none of this information is in the JSON object

merlin-server/sql/merlin/tables/constraint.sql
```diff
@@ -84,22 +84,19 @@ for each row
...
-    update constraint_run
-    set definition_outdated = false
-    where constraint_id = new.id
-      and constraint_definition == new.definition
-      and definition_outdated = true;
-  end if;
+  returns trigger
+  security definer
+  language plpgsql as $$begin
+  update constraint_run
```
Mythicaeda: Minor: readability:

```suggestion
  returns trigger
  security definer
  language plpgsql as $$
begin
  update constraint_run
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: I improved the test and now it is checking both cached and original version

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Got it, made the change.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Updated the display name

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: I am now checking the results object

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
goetzrrGit: got it, reverted

merlin-server/sql/merlin/tables/constraint.sql
goetzrrGit: made the change, my fingers are crossed hoping I didn't break the test

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
```diff
@@ -3,6 +3,7 @@
 import javax.json.JsonNumber;
 import javax.json.JsonObject;
 import javax.json.JsonString;
+import java.util.ArrayList;
 import java.util.List;
```
Mythicaeda: Minor: unused imports (only `JsonObject` and `List` are used)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -13,6 +13,7 @@
 import javax.json.Json;
 import javax.json.JsonValue;
 import java.io.IOException;
+import java.util.ArrayList;
```
Mythicaeda: Minor: unused import

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -137,7 +138,10 @@ void constraintsSucceedNoViolations() throws IOException {
     hasura.deleteActivity(planId, activityId);
     hasura.awaitSimulation(planId);
     final var constraintsResults = hasura.checkConstraints(planId);
-    assertEquals(0, constraintsResults.size());
+
+    assertEquals(1, constraintsResults.size());
+    assertEquals(true, constraintsResults.get(0).success());
```
Mythicaeda: `assertEquals(true, X)` should be replaced with `assertTrue(X)`. Similarly, `assertEquals(false, X)` should be replaced with `assertFalse(X)`. This comment applies throughout the file, I am just placing it on the first occurrence.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -183,7 +187,15 @@ void constraintCachedNoViolations() throws IOException {
     assertEquals(constraintDefinition, cachedRun.constraintDefinition());
 
     // Check results
-    assertTrue(cachedRun.results().isEmpty());
+    assertTrue(!cachedRun.results().isEmpty());
```
Mythicaeda: Minor, ignorable: `assertTrue(cachedRun.results().isPresent())` is clearer.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -258,7 +270,7 @@ void runConstraintsOnOldSimulation() throws IOException {
 
     // Expect no violations on the new simulation
     final var newConstraintResults = hasura.checkConstraints(planId, newSimDatasetId);
-    assertEquals(0, newConstraintResults.size());
+    assertEquals(1, newConstraintResults.size());
```
Mythicaeda: Check needs to be updated to satisfy the comment.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -460,7 +472,8 @@ void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
       assertEquals(false,constraintRepsonse.success());
       assertEquals(1,constraintRepsonse.errors().size());
       final ConstraintError error = constraintRepsonse.errors().get(0);
-      assertEquals("Constraint 'fruit_equal_peel': TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+      assertEquals("Constraint 'fruit_equal_peel' compilation failed:\n"
+                   + " TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
     }
```
Mythicaeda: Noticed a couple of things in this test that slipped by in the prior PR. Here's some updated code:

```java
    @Test
    @DisplayName("Simulation-Associated External Datasets aren't loaded when other SimIds are passed")
    void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
      // Delete Activity to make the simulation outdated, then resim
      hasura.deleteActivity(planId, activityId);
      final int newSimDatasetId = hasura.awaitSimulation(planId).simDatasetId();
      // This test causes the endpoint to throw an exception when it fails to compile the constraint,
      // as it cannot find the external dataset resource in the set of known resource types.
      final var constraintResponses = hasura.checkConstraints(planId, newSimDatasetId);
      assertEquals(1,constraintResponses.size());
      final var constraintResponse = constraintResponses.get(0);
      assertFalse(constraintResponse.success());
      assertEquals(1,constraintResponse.errors().size());
      final var error = constraintResponse.errors().get(0);
      assertEquals("Constraint 'fruit_equal_peel' compilation failed:\n"
                   + " TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
    }
```

The changes are:
1) Major: the test is for `External Datasets aren't loaded when other SimIds are passed`. In the last PR, `newSimDatasetId` stopped being passed into `checkConstraints` when the method was pulled out of the `assertThrows` state.
2) Minor: using `assertFalse` instead of `assertEquals(false)`
3) Minor: resolves typo in `constraintRepsonse`
4) Minor, for style consistency with the codebase: using `final var error` instead of `final ConstraintError error`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
Mythicaeda: Minor: please capitalize the first letter of the display name.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
Mythicaeda: Sorry to thrash on this advice, but I took a look at `CachedConstraintRun` and realized that the most accurate representation would be to make the `results` field on `ConstraintRecord` an `Optional` and have this line mirror that code:
```java
json.getJsonObject("results").isEmpty() ?
            Optional.empty() :
            Optional.of(ConstraintResult.fromJSON(json.getJsonObject("results"))
```
Doing this change would mean that in `ConstraintsTests`, it would need to be asserted that the results are present before they are fetched (`assertTrue(results.isPresent())`), and `assertTrue(results.isEmpty)` could be used to validate the response object in compilation failure cases.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -183,7 +187,15 @@ void constraintCachedNoViolations() throws IOException {
     assertEquals(constraintDefinition, cachedRun.constraintDefinition());
 
     // Check results
-    assertTrue(cachedRun.results().isEmpty());
+    assertTrue(!cachedRun.results().isEmpty());
+    final var results = cachedRun.results().get();
+    assertEquals("fruit_equal_peel",results.constraintName());
```
Mythicaeda, line 194: Minor: `assertEquals(constraintId, results.constraintId)` can be added here

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+    void constraintInvalidModification() throws IOException {
+
+      hasura.awaitSimulation(planId);
+      final var constraintsResponses = hasura.checkConstraints(planId);
+
+      assertEquals(1, constraintsResponses.size());
+      assertEquals(true, constraintsResponses.get(0).success());
+
+      hasura.updateConstraint(
+          constraintsResponses.get(0).result().constraintId(),
```
Mythicaeda: This value is already stored in `constraintId` and does not need to be fetched out of the response. This also applies to L519

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+          "export default (): Constraint => Discrete.Resource(\"/my_boolean\").equal(true); error");
+
+      final var constraintsErrorResponses = hasura.checkConstraints(planId);
+      assertEquals(1, constraintsErrorResponses.size());
+      assertEquals(false, constraintsErrorResponses.get(0).success());
+      assertEquals(1 , constraintsErrorResponses.get(0).errors().size());
```
Mythicaeda: Minor, whitespace:
```suggestion
      assertEquals(1, constraintsErrorResponses.get(0).errors().size());
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+      assertEquals("Constraint 'fruit_equal_peel' compilation failed:\n"
+                   + " TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+
+    }
+
+    @Test
+    @DisplayName("if a constraint is updated and then reverted, Constraints will load the cached value from before the update.")
+    void constraintInvalidModification() throws IOException {
+
+      hasura.awaitSimulation(planId);
```
Mythicaeda: Minor: whitespace:
```suggestion
    void constraintInvalidModification() throws IOException {
      hasura.awaitSimulation(planId);
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+          "export default (): Constraint => Discrete.Resource(\"/my_boolean\").equal(true); error");
+
+      final var constraintsErrorResponses = hasura.checkConstraints(planId);
+      assertEquals(1, constraintsErrorResponses.size());
+      assertEquals(false, constraintsErrorResponses.get(0).success());
+      assertEquals(1 , constraintsErrorResponses.get(0).errors().size());
+
+      hasura.updateConstraint(
+          constraintsResponses.get(0).result().constraintId(),
+          "export default (): Constraint => Discrete.Resource(\"/my_boolean\").equal(true)");
```
Mythicaeda: This definition should be pulled into a `cross-test constant` for the `WithExternalDatasets` tests (See L326). That way, we can update this test and the `beforeEach` at the same time in the future.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
Mythicaeda: Missing `assertNull(constraintsErrorResponses.results)`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
+      final var constraintsCacheResponses = hasura.checkConstraints(planId);
+
+      assertEquals(1, constraintsCacheResponses.size());
+      assertEquals(true, constraintsCacheResponses.get(0).success());
 
+      assertEquals(constraintsResponses.get(0),constraintsCacheResponses.get(0));
```
Mythicaeda: This can be simplified to `assertEquals(constraintsResponses, constraintsCacheResponses)`.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -478,8 +491,39 @@ void compilationFailsOutdatedSimulationNoSimDataset() throws IOException {
...
       final ConstraintError error = constraintRepsonse.errors().get(0);
-      assertEquals("Constraint 'fruit_equal_peel': TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+      assertEquals("Constraint 'fruit_equal_peel' compilation failed:\n"
+                   + " TypeError: TS2345 Argument of type '\"/my_boolean\"' is not assignable to parameter of type 'ResourceName'.",error.message());
+
+    }
+
+    @Test
+    @DisplayName("if a constraint is updated and then reverted, Constraints will load the cached value from before the update.")
+    void constraintInvalidModification() throws IOException {
```
Mythicaeda, line 521: Minor: Could you please add a couple brief comments in the same vein as how the rest of the tests are commented?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
goetzrrGit: Removed the unused imports

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Removed the unused imports

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: I went through and changed all `assertEquals` to there respected `assertTrue` and `assertFalse`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Using `isPresent` now

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Added a check to make sure that the violations are really empty.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Thanks for catching this. I replaced the test with the updated one.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Added the `assertEquals(constraintId, results.constraintId)`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: using the cross-test `constraintID` now

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: fixed

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: fixed

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Now using `WithExternalDatasets.constraintDefinition`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Updated with `assertTrue(constraintsErrorResponses.get(0).result().isEmpty());` as we are using `optionals` now

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Simplified as stated above

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: Added comments to this test case and a brief description

goetzrrGit:
@Mythicaeda I believe I fixed everything you suggested in the ConstraintTest. I have updated all the tests with optional checking and did some parameter renaming to make it more clear.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/types/ConstraintRecord.java
goetzrrGit: No worries, I have updated all the test and added this check:

```java
    final var constraintResponse = constraintsResponses.get(0);
    assertTrue(constraintResponse.success());
    assertTrue(constraintResponse.result().isPresent());
    final var constraintResult = constraintResponse.result().get();
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: It is now reworded and capitalized

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -136,19 +136,23 @@ void constraintsSucceedNoViolations() throws IOException {
...
+    assertTrue( constraintResponses.get(0).success());
+    assertTrue( constraintResponses.get(0).result().isPresent());
+    assertEquals(0, constraintResponses.get(0).result().get().violations().size());
   }
 
   @Test
   void constraintCachedViolation() throws IOException {
     final var simDatasetId = hasura.awaitSimulation(planId).simDatasetId();
-    final var constraintRuns = hasura.checkConstraints(planId);
+    final var constraintRepsonses = hasura.checkConstraints(planId);
```
Mythicaeda: Minor: Typo
```suggestion
    final var constraintResponses = hasura.checkConstraints(planId);
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
```diff
@@ -454,13 +478,15 @@ void compilationFailsOutdatedSimulationSimDatasetId() throws IOException {
       final int newSimDatasetId = hasura.awaitSimulation(planId).simDatasetId();
       // This test causes the endpoint to throw an exception when it fails to compile the constraint,
       // as it cannot find the external dataset resource in the set of known resource types.
-      final var constraintResponses = hasura.checkConstraints(planId);
+      final var constraintResponses = hasura.checkConstraints(planId, newSimDatasetId);
       assertEquals(1,constraintResponses.size());
       final var constraintRepsonse = constraintResponses.get(0);
```
Mythicaeda: Minor: typo
```suggestion
      final var constraintResponse = constraintResponses.get(0);
```

Mythicaeda (approved):
LGTM!

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/ConstraintsTests.java
goetzrrGit: noooooooo!!!!!! stupid fat fingers. I fixed it.

Issue #1241, opened 2023-11-09
Need a way to stop a scheduling run


mattdailis:
Follow up to #765, but specifically for a scheduling run.

PR #1243, opened 2023-11-14
Cancel Running Simulations

### LINKED ISSUE #765: Need a way to stop a running simulation
abercrom
https://jira.jpl.nasa.gov/browse/AERIEQS-451
Moderate
During thread test 2 we accidentally started a simulation with an incorrect configuration which caused the sim to to a very long time to run. We had to restart the worker container to get the sim to stop. Aerie needs a way for either the user or an administrator to stop a running simulation. (Restarting the container is a brute force solution, but application admins may not have permission to do that on production venues.)

Mythicaeda:
* **Tickets addressed:** Closes #765
* **Review:** By commit  

## Description

Adds a `SimulationCancelListener` that, upon receiving a signal to cancel a simulation from the DB, sets a flag to tell the Driver to stop before it either picks up or begins to perform more work. With this change, the only time a simulation will not be stoppable aside from shutting down the container is if the user code contains an infinite loop. It also alters how the `canceled` flag is set in the DB to accurately reflect that "if this flag is set, then the simulation was canceled".

Breakdown of commits:
- Commit 1 removes our auto-cancel triggers and sends a notification to the `simulation_cancel` channel when a simulation is canceled. The auto-cancel triggers were removed in order to give the users more control over when their simulations are canceled. This does mean that if anything is updated between a pending request and when it is picked up, it will now fail with an "Request is no longer relevant" message.
- Commit 2 adds Java code to listen for the new notification added in (1). Rather than using a separate DB connection for this, it was added to the list of subscribed channels in the Worker's Listeners. 
- Commit 3 threads this listener through to the SimulationDriver. **This is a breaking change**, because it alters the signature of `simulate`.
- Commit 4 alters the behavior of the Driver to check if the simulation has been canceled before it fetches or runs a batch, and, if it has, to return the results up until when it was canceled with its `reason` field set to how far into simulation it was (using the same JSON structure as when simulation fails with an exception).
- Commits 5-7 update our tests.

## Verification

- A new e2e test was added to check that canceling works and returns incomplete results. This test uses the `foo` model.
- DB Tests checking the removed triggers were also removed
- `SimulationDatasets` in our e2eTests now contain the information from the `status` and `reason` columns.
- Unit tests have been updated with the new `simulate` syntax. All of them were set to be uncancelable.

## Documentation

No docs need to be updated.

## Future work

- Is there a generic name for `SimulationFailure`? Prior to this PR, it was exclusively being used to fill in the `reason` column for failed simulations, but it's now also filling in the `reason` column for canceled sims.
- Scheduling ticket: #1241 
- UI needs to be updated to 1) allow canceling an inprogress simulation and 2) to properly display what a canceled simulation looks like (Useful note: canceled sims can only be `incomplete` or `pending`. A `failed` or `success` sim with `canceled` set to `true` implies that the simulation either finished or failed before it received the `canceled` signal)


merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
```diff
@@ -84,7 +86,9 @@ public static void main(String[] args) throws InterruptedException {
             notification.simulationTemplateRevision());
         final ResultsProtocol.WriterRole writer = owner.get();
         try {
-          simulationAgent.simulate(planId, revisionData, writer);
+          canceledListener.register(new DatasetId(datasetId));
```
mattdailis: 🔶 Some food for thought - I think moving `register` to before `claim` avoids a possible race condition where a cancel request comes in after `claim`, but before `register`

```diff
final var datasetId = notification.datasetId();
+
+canceledListener.register(new DatasetId(datasetId));
final Optional<ResultsProtocol.OwnerRole> owner = stores.results().claim(planId, datasetId);
if (owner.isEmpty()) continue;

final var revisionData = new PostgresPlanRevisionData(
    notification.modelRevision(),
    notification.planRevision(),
    notification.simulationRevision(),
    notification.simulationTemplateRevision());
final ResultsProtocol.WriterRole writer = owner.get();
try {
-  canceledListener.register(new DatasetId(datasetId));
  simulationAgent.simulate(planId, revisionData, writer, canceledListener);
```

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
```diff
@@ -84,7 +86,9 @@ public static void main(String[] args) throws InterruptedException {
...
+          simulationAgent.simulate(planId, revisionData, writer, canceledListener);
+          canceledListener.unregister();
         } catch (final Throwable ex) {
```
mattdailis: 🔵 I suspect that we might be able to do without `unregister` 🤔 

If we do keep `unregister`, we should consider what happens in the case that an exception is thrown during simulation. Perhaps moving it into a `finally` block, or making use of try-with-resources would ensure that it will definitely get called.

mattdailis (approved):
Awesome work! No major concerns

> Key:
> 🔻 Issues to address before merge
>  🔶 Requests that should not block merge, but should at least be discussed 
> 🔵 Recommendations that can be ignored if desired

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
Mythicaeda: Ah, thanks for the catch! Done

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/MerlinWorkerAppDriver.java
Mythicaeda: I've moved the `unregister` to a `finally` block 👍 

Re: using a `try-with-resources`: I did initially try to make the listener implement `Autoclosable`, however I didn't think it matched the interface's typical pattern (that being that `close` is called at the end of the object's lifecycle). I believe I also thought that `try-with-resources` had to create the object in the "with resources" part, but apparently that was just a misconception on my end.

Re: not having `unregister`: In my opinion, by having `unregister` and by calling it immediately after `simulate`, we're guaranteeing that once the worker is done with a request, it is done paying attention to that any information about that request. I felt that this was logically cleaner and less likely to cause potential future bugs then having the worker continue to acknowledge canceled requests in the time between it completing a request and it claiming a new one.

I'm happy to discuss either of these further, as I can see the merit of both methods (though obviously not at the same time).

mattdailis (approved):
Did a quick manual test, re-approving :+1:

PR #1251, opened 2023-12-04
Cancel a Scheduling Run

### LINKED ISSUE #1241: Need a way to stop a scheduling run

Mythicaeda:
* **Tickets addressed:** Closes #1241 
* **Review:** By commit  

## Description

Allows for a scheduling run to be canceled.

Scheduling will check if it has been canceled:
  - before computing simulation results
  - before processing a batch in simulation
  - before initializing the plan
  - at the top of `satisfyGoal`
  - at the top of `removeAndInsertActivitiesFromSimulation`
If it has been canceled, it will throw a `SchedulingInterruptedException`, which will bubble up until it hits the `SynchronousSchedulerAgent`, which will catch the exception and report the scheduling run as canceled.

Unlike with simulation, partial results are not returned.

## Verification

- Unit tests have been updated with the new `schedule` syntax. All of them have been set to uncancelable.
- A new e2e test was added to check that canceling works and updates the `reason` column.

## Documentation

No docs need to be updated.

## Future work

UI needs to be updated to allow for canceling an in-progress Scheduling Run

### DIFF
--- a/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulingCanceledListener.java
+++ b/scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulingCanceledListener.java
@@ -1,0 +1,55 @@
+package gov.nasa.jpl.aerie.scheduler.worker;
+
+import gov.nasa.jpl.aerie.scheduler.server.models.SpecificationId;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public class SchedulingCanceledListener implements Supplier<Boolean> {
+  private Optional<SpecificationId> registeredSchedulingRun;
+  private boolean canceled;
+
+  public SchedulingCanceledListener() {
+    registeredSchedulingRun = Optional.empty();
+    canceled = false;
+  }
+
+  /**
+   * Receive a canceled signal.
+   * All signals that are not for this object's registered scheduling run will be ignored.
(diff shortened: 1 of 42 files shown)


scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulingCanceledListener.java
```diff
@@ -0,0 +1,55 @@
...
+    canceled = false;
+  }
+
+  /**
+   * Receive a canceled signal.
+   * All signals that are not for this object's registered simulation will be ignored.
+   * @param payload The payload of the signal
+   */
+  public void receiveSignal(SpecificationId payload){
+    if (registeredSimulation.isEmpty() || !registeredSimulation.get().equals(payload)) return;
```
adrienmaillard: you left the word simulation in this class

adrienmaillard (approved):
As usual, nice and spotless work

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/SchedulingCanceledListener.java
Mythicaeda: Thanks for catching this! I've scrubbed the `simulation`s

PR #1258, opened 2023-12-14
Added a new ValidationResult type that can be used to return more spe…

### LINKED ISSUE #941: Allow validations to include failure reason
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

This issue is potentially related - since it would be possible to leverage exceptions to provide these strings: https://github.com/NASA-AMMOS/aerie/issues/502 Open question as to whether that is the best design direction here.

Currently, activity validations are methods that return booleans - in the `false` case, a static string is displayed in the UI that describes which validation was failed.

This is sufficient for simple validations - but for complex ones, it is desirable to communicate additional information about how exactly this validation was failed - for example, when validating a parameter of type "list", it would be helpful to communicate which element of the list caused the violation.

Source: [slack message](https://jpl.slack.com/archives/C0163E42UBF/p1683927568050659)

cohansen:
…cific validation notices

* **Tickets addressed:** #941.
* **Review:** By commit

## Description
I added a new `ValidationResult` class that can return more specific validation failures.

## Verification
Re-ran tests and manually tested against existing Bananation activities that use the new `ValidationResult` return type.

## Documentation
https://github.com/NASA-AMMOS/aerie-docs/pull/112

## Future work
Not sure, I'm hoping these changes capture what we wanted as part of #941.

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -191,7 +191,7 @@ type MerlinSimulationResponse {
 }
 
 type ValidationResponse {
-  errors: [ValidationNotice!]
+  errors: ValidationNotice!
   success: Boolean!
 }
 
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -191,7 +191,7 @@ type MerlinSimulationResponse {
 }
 
 type ValidationResponse {
-  errors: [ValidationNotice]!
+  errors: [ValidationNotice!]
   success: Boolean!
 }
 
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -191,7 +191,7 @@ type MerlinSimulationResponse {
 }
 
 type ValidationResponse {
-  errors: ValidationNotice!
+  errors: [ValidationNotice]!
   success: Boolean!
 }
 
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
@@ -15,12 +15,12 @@ public record BakeBananaBreadActivity(double temperature, int tbSugar, boolean g
   @Validation
   public ValidationResult validateTemperatures() {
     if (this.temperature < 0) {
-      return new ValidationResult(false, "temperature", Optional.of("Temperature must be positive"));
+      return new ValidationResult(false, "temperature", "Temperature must be positive");
     }
 
     return new ValidationResult(!glutenFree || temperature >= 100,
       "glutenFree",
-      Optional.of("Gluten-free bread must be baked at a temperature >= 100"));
+      "Gluten-free bread must be baked at a temperature >= 100");
   }
 
   @EffectModel
--- a/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
+++ b/examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
@@ -1,24 +1,26 @@
 package gov.nasa.jpl.aerie.banananation.activities;
 
 import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.contrib.models.ValidationResult;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Validation;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.WithDefaults;
 
+import java.util.Optional;
+
 @ActivityType("BakeBananaBread")
 public record BakeBananaBreadActivity(double temperature, int tbSugar, boolean glutenFree) {
 
-  @Validation("Temperature must be positive")
-  @Validation.Subject("temperature")
-  public boolean validateTemperature() {
-    return this.temperature() > 0;
-  }
+  @Validation
+  public ValidationResult validateTemperatures() {
+    if (this.temperature < 0) {
+      return new ValidationResult(false, "temperature", Optional.of("Temperature must be positive"));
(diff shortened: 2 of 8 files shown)


deployment/hasura/metadata/actions.graphql
```diff
@@ -191,7 +191,7 @@ type MerlinSimulationResponse {
 }
 
 type ValidationResponse {
-  errors: [ValidationNotice!]
+  errors: ValidationNotice!
   success: Boolean!
```
Mythicaeda: This change seems incorrect. The endpoint in the Java code returns a list of notices, and an activity can return a list of validation failures. I.e. GrowBanana can return a list of length two, if both the `quantity` and `growingDuration` parameters are invalid.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -502,6 +502,10 @@ private List<ParameterValidationRecord> getExportValidations(
     for (final var element : exportTypeElement.getEnclosedElements()) {
       if (element.getAnnotation(Export.Validation.class) == null) continue;
 
+      // Is there a way to tie this to the actual class rather than the string?
+      boolean isSimpleValidation = !(element.getKind() == ElementKind.METHOD &&
+          ((ExecutableElement) element).getReturnType().toString().equals("gov.nasa.jpl.aerie.contrib.models.ValidationResult"));
```
Mythicaeda: @mattdailis or @Twisol might know some fancier way, but if you want to tie it to the class, you could do this:
```suggestion
      // Is there a way to tie this to the actual class rather than the string?
      boolean isSimpleValidation = !(element.getKind() == ElementKind.METHOD &&
          ((ExecutableElement) element).getReturnType().toString().equals(ValidationResult.class.getTypeName()));
```

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MapperMethodMaker.java
```diff
@@ -151,26 +153,58 @@ public final MethodSpec makeGetValidationFailuresMethod() {
...
+                          .addStatement(
+                              "notices.add(new $T($T.of($L), ex.getMessage()))",
+                              ValidationNotice.class,
+                              List.class,
+                              subjects)
+                          .endControlFlow();
+                    }
+
+                    codeBlock.addStatement(
+                      "if (!$L.$L().$L()) notices.add(new $T($T.of($L.$L().$L()), $L.$L().$L().orElse(null)))",
```
Mythicaeda: If `Optional.empty()` is going to converted to `null` here, why use `Optional` in `ValidationResult`?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
cohansen: Awesome, this is exactly what I was looking for!

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MapperMethodMaker.java
cohansen: That's a good point, originally I wanted a `ValidationResult` to not have to include a `message`, but it probably makes more sense to just accept a `null` in that case.

`new ValidationResult(true, "biteSize", null)`

deployment/hasura/metadata/actions.graphql
Mythicaeda: Change was not properly reverted.

The original code was `errors: [ValidationNotice!]`, which means that `errors` is optional, but if it's there, there are no `null` entries. 

`errors: [ValidationNotice]!` means that `errors` _must_ be present, but that it may have `null` entries.

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BakeBananaBreadActivity.java
```diff
@@ -1,24 +1,26 @@
 package gov.nasa.jpl.aerie.banananation.activities;
 
 import gov.nasa.jpl.aerie.banananation.Mission;
+import gov.nasa.jpl.aerie.contrib.models.ValidationResult;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Validation;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.WithDefaults;
 
+import java.util.Optional;
```
Mythicaeda: Minor: Unused Import

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
```diff
@@ -3,12 +3,15 @@
...
 import gov.nasa.jpl.aerie.banananation.Mission;
 import gov.nasa.jpl.aerie.contrib.metadata.Unit;
+import gov.nasa.jpl.aerie.contrib.models.ValidationResult;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.ActivityType.EffectModel;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.AutoValueMapper;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Parameter;
 import gov.nasa.jpl.aerie.merlin.framework.annotations.Export.Validation;
 
+import java.util.Optional;
```
Mythicaeda: Minor: unused import

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MapperMethodMaker.java
Mythicaeda: Yeah, either an explicit `null` or adding an alternate constructor without a `message` field would work.

```java
public ValidationResult(boolean success, String subject) {
    this(success, subject, null);
  }
```

Mythicaeda (approved):
LGTM!

PR #1262, opened 2023-12-19
Handle task scheduling in ResumableSimulationDriver

### LINKED ISSUE #1256: State of resumable simulation driver is not properly maintained
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

develop

### Describe the bug

State of resumable simulation driver is not properly maintained. 
plannedDirectiveToTask is not updated with activities that have an anchor, therefore simulation is finishing early

### Reproduction

N/A

### Logs

```shell
N/A
```

### System Info

```shell
N/A
```

### Severity

Moderate

adrienmaillard:
* **Tickets addressed:** Fixes #1256 
* **Review:** By commit  

## Description

Here is an example of how the issue has manifested: 
- The scheduler has to simulate a plan with activities A and B with B anchored to A. 
- After simulation, the scheduler tries to query the duration of task B, it fails
- Indeed, the resumable simulation driver does not have a direct reference to task B because of how anchored activities were managed, that is, by creating an anonymous task that would include the A, a delay, then B. The driver would remember the id of this anonymous task but not the id of task B precisely. The id of a task is given when the task is scheduled and in this case, the simulation engine would schedule task A and B, not the driver. 

The solution consists in replacing this anonymous task creation with a mechanism that lets the driver schedule dependent tasks as soon as their dependent is finishing. Handling task creation allows to maintaining references to task ids and solve the above issue. 

## Verification

A simple test that was not passing before has been added.

## Documentation

N/A

## Future work

N/A

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
@@ -26,9 +26,11 @@ import org.slf4j.LoggerFactory;
 import java.time.Instant;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Supplier;
 
 public class ResumableSimulationDriver<Model> implements AutoCloseable {
@@ -52,6 +54,9 @@ public class ResumableSimulationDriver<Model> implements AutoCloseable {
   //mapping each activity name to its task id (in String form) in the simulation engine
   private final Map<ActivityDirectiveId, TaskId> plannedDirectiveToTask;
 
+  //subset of plannedDirectiveToTask to check for scheduling dependent tasks
+  private final Map<ActivityDirectiveId, TaskId> toCheckForDependencyScheduling;
+
   //simulation results so far
   private SimulationResults lastSimResults;
   //cached simulation results cover the period [Duration.ZERO, lastSimResultsEnd]
@@ -71,6 +76,7 @@ public class ResumableSimulationDriver<Model> implements AutoCloseable {
   ){
     this.missionModel = missionModel;
     plannedDirectiveToTask = new HashMap<>();
+    toCheckForDependencyScheduling = new HashMap<>();
     this.planDuration = planDuration;
     countSimulationRestarts = 0;
     this.canceledListener = canceledListener;
@@ -94,6 +100,7 @@ public class ResumableSimulationDriver<Model> implements AutoCloseable {
     printTimeSpent();
     durationSinceRestart = 0;
     plannedDirectiveToTask.clear();
+    toCheckForDependencyScheduling.clear();
     lastSimResults = null;
     lastSimResultsEnd = Duration.ZERO;
     long before = System.nanoTime();
@@ -273,13 +280,14 @@ public class ResumableSimulationDriver<Model> implements AutoCloseable {
         schedule).compute();
     // Filter out activities that are before the plan start
     resolved = StartOffsetReducer.filterOutNegativeStartOffset(resolved);
-
+    final var toSchedule = new HashSet<ActivityDirectiveId>();
+    toSchedule.add(null);
     scheduleActivities(
+        toSchedule,
         schedule,
         resolved,
         missionModel,
-        engine,
-        activityTopic
+        engine
     );
 
     var allTaskFinished = false;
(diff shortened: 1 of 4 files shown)


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
```diff
@@ -328,93 +338,67 @@ private void simulateSchedule(final Map<ActivityDirectiveId, ActivityDirective>
...
-                  resolved,
-                  missionModel,
-                  activityTopic
-              ));
-              return TaskStatus.completed(Unit.UNIT);
-            }));
-      }
-      return TaskStatus.completed(Unit.UNIT);
-    });
+    }), scheduler2 -> TaskStatus.completed(Unit.UNIT));
```
Mythicaeda: I don't think bit with the second scheduler is needed, judging by the code for `emitAndThen` (the predecessor of `makeTaskFactory`):
```java
private static <E, T>
  TaskFactory<T> emitAndThen(final E event, final Topic<E> topic, final TaskFactory<T> continuation) {
    return executor -> scheduler -> {
      scheduler.emit(event, topic);
      return continuation.create(executor).step(scheduler);
    };
  }
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
```diff
@@ -328,93 +338,67 @@ private void simulateSchedule(final Map<ActivityDirectiveId, ActivityDirective>
    * @return its duration if the activity has been simulated and has finished simulating, an IllegalArgumentException otherwise
    */
   public Optional<Duration> getActivityDuration(ActivityDirectiveId activityDirectiveId){
+    //potential cause of non presence: activity is outside plan bounds
```
Mythicaeda: Another potential cause of absence is if it hasn't yet been scheduled. Is this potential cause more expected than if the activity is outside plan bounds?

Mythicaeda (commented):
Could you either update or add a new Anchor Integration test in `SchedulingIntegrationTests` that uses a non-controllable duration activity? The fact that all of those tests use `GrowBanana` was how this bug slipped by for so long.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: not really, you're right, I was trying to write what we learned in this case

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
```diff
@@ -328,93 +338,67 @@ private void simulateSchedule(final Map<ActivityDirectiveId, ActivityDirective>
...
+    final var iterator = toCheckForDependencyScheduling.entrySet().iterator();
+    while(iterator.hasNext()){
+      final var taskToCheck = iterator.next();
+      if(engine.isTaskComplete(taskToCheck.getValue())){
+        toSchedule.add(taskToCheck.getKey());
+        iterator.remove();
+      }
+    }
+    return toSchedule;
+  }
```
Mythicaeda, line 357: Minor, nonblocking: proposal: this method could call `scheduleActivities` directly instead of doing `toSchedule.add`. Then we don't have to do create a set. Additionally, it would get rid of an unnecessary loop (we already loop over the set `toScheduleNow` in the process of generating it)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard:
```java
  private <Output> TaskFactory<Output> makeTaskFactory(
      final ActivityDirectiveId directiveId,
      final TaskFactory<Output> task,
      final Topic<ActivityDirectiveId> activityTopic
  )
  {
    return executor -> scheduler -> {
      scheduler.emit(directiveId, activityTopic);
      return task.create(executor).step(scheduler);
    };
  }
```
This is better

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/ResumableSimulationDriver.java
adrienmaillard: But we would be looping through `toCheckForDependencyScheduling` while we would be adding elements to it in `scheduleActivities`

Mythicaeda: ...Ah, yeah, forgot about that.

adrienmaillard:
Good catch, I forgot this, I have updated the `GrowBanana` of one test to a `PickBanana`.

adrienmaillard:
I'd like to point out that this PR is strictly removing code, which should be celebrated somehow :p

Mythicaeda (approved):
🎉

PR #1269, opened 2023-12-21
simulation performance improvements

martyvona:
includes cherry picks from @DavidLegg `DavidLegg/feature--job-schedule-speedup`

goes with corresponding PR in clipper-aerie

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
@@ -15,8 +15,18 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.jar.JarFile;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Arrays;
+import java.lang.ref.SoftReference;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class MissionModelLoader {
+
+    private static final Logger logger = LoggerFactory.getLogger(MissionModelLoader.class);
+
     public static ModelType<?, ?> loadModelType(final Path path, final String name, final String version)
     throws MissionModelLoadException
     {
@@ -58,6 +68,8 @@ public final class MissionModelLoader {
         }
     }
 
+    private static SoftReference<URLClassLoader> lastClassLoader;
+
     public static MerlinPlugin loadMissionModelProvider(final Path path, final String name, final String version)
     throws MissionModelLoadException
     {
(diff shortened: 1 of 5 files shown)


parkerabercrombie:
@mattdailis any estimate on when this might be reviewed? We don't have a strict deadline, but we do have some changed in our adaptation that depend on this, and it provides a significant performance benefit.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/MissionModelLoader.java
```diff
@@ -66,7 +78,21 @@ public static MerlinPlugin loadMissionModelProvider(final Path path, final Strin
         final var className = getImplementingClassName(path, name, version);
 
         // Construct a ClassLoader with access to classes in the mission model location.
-        final var classLoader = new URLClassLoader(new URL[] {missionModelPathToUrl(path)});
+        final var url = missionModelPathToUrl(path);
+
+        // Important optimization: if the last time we were called was for the same mission model then reuse the same
+        // classloader.  In particular this ensures that any static caches (e.g. for memoization optimizations) in the
+        // mission model are not reset across simulation runs.
```
Mythicaeda: While it is true that this change will permit an optimization done in this way, the state of the model when it loads _shouldn't_ be nondeterministic. Preserving the `static`s from the last model load (not necessarily simulation runs) is more often going to lead to unexpected bugs coming from state bleed, especially when you consider that that state could've been bled from an unrelated plan using the same model.

martyvona: Thanks for your review @Mythicaeda, your concerns are in line with some other conversations I've had.

Since there seem to be several concerns about this particular change I will remove it.  It's mostly aimed at helping the scheduler usecase which repeatedly runs simulations, and which is unusably slow currently for practical plans, which was the genesis of this effort.

It appears that even without persisting the memoization caches across simulation runs, there is still a lot of performance improvement.  For `eurc/sample-plans/cruise-rap-dev-9-26-old-updated.json` the sim time (on a particular machine) goes from 400s baseline to 126s on first run (thus not benefiting from persisted memoization caches) to 94s on subsequent runs with persisted caches.

So I will remove this because of the concern over it, and because it only seems to give an additional ~5-10% perf improvement anyway (though that could vary widely depending on the usecase).

That said, I'm not sure that it's really accurate to call what is happening here nondeterministic, nor is it necessarily the case that this type of caching would be a bug even if the same mission model is used to simulate multiple different plans.

The idea of memoized state like this is that each saved memo (function return value) is supposed to be indexed by a unique key that precisely and completely captures the conditions under which that function returns exactly that value.  Whether those conditions are just the values of the function arguments, or the arguments plus any additional state that the function depends on, is supposed to be captured by the implementer of the memoization.  If they did that correctly, then the saved memos should be valid across different plans or whatever.  The only other invariant that should need to be maintained is that the code of the function that was memoized should be the same.

In a related conversation, @DavidLegg suggested considering an alternate approach of having the mission model persist such caches, if it wants to, to disk at the end of simulation, and load them from disk at the start of any subsequent simulation.  While that is certainly an option in the space of possible alternatives to reusing the same ClassLoader instance, I don't think I'd recommend going down that path at this point.  Again, this is all only gaining a relatively small additional perf improvement.  Java serialization is relatively slow, not only in the actual disk accesses (which might not be too bad on modern hardware) but also in the typical reflection-based implementations.  And that slowness would possibly eat into the perf gain that is the point of this anyway.

The architecture of these memoization caches is that they are supposed to be fast and somewhat ephemeral in-memory things.  I use SoftReference to implement them in the eurc mission model which means that they could go away at any time if memory is needed.  Following that architecture, if people wanted to keep the option open to preserve them across simulation runs in Aerie, yet another way it could possibly be done is to amend the Aerie API to offer the mission model a general purpose SoftRef storage slot that is persisted (if memory allows) across simulations, e.g. perhaps as an additional parameter to gov.nasa.jpl.aerie.merlin.protocol.model.ModelType.instantiate() or something in that codepath.

Mythicaeda: Thank you for removing this, although two quick clarifying notes:

1) This change would not have impacted scheduling, as the mission model code is not reloaded between restarts. Model code is loaded at the start of a simulation or scheduling run.
2) The concern here isn't the memoized use case, it's the `static` states that _aren't_ part of the memoized use case. For example, singleton classes, especially singleton classes that take in any aspect of the simulation environment it's being run in (ie the start of the planning horizon)

Mythicaeda:
Closing this PR in favor of #1212, which has been merged.

parkerabercrombie:
This PR also included some logging which looks useful: 48f6039ab4343f35a590149f70f78e5309a42bad
Could we open another PR for that change?

PR #1271, opened 2024-01-02
Fix DB Lockup when Duplicating Large Plans

### LINKED ISSUE #1252: Database unresponsive if many users create branches of the same plan at the same time
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

2.1.0

### Describe the bug

In a multi-user test we were able to get the database to stop responding by having many people (~8) create branches of the same plan at the same time. The plan we used was from Clipper Thread Test 6 - a six week plan with about 200 activities. The system eventually recovered, but there was at least 5 min during which Aerie was completely unresponsive.

We attempted this twice with everyone creating branches from the same parent, and the system became unresponsive both times. We also tried with people branching from different plans, and did not reproduce the issue (branches were created quickly and successfully).

It is likely that this issue is what occurred during TT-6: https://github.com/NASA-AMMOS/aerie/issues/1044

### Reproduction

Have a bunch of people create branches at the same time. The size and complexity of the plan is probably important. Contact me for the plan that we used.

### Logs

_No response_

### System Info

```shell
Chrome
```

### Severity

Critical

Mythicaeda:
* **Tickets addressed:** Fixes #1252 
* **Review:** By commit  

## Description

The database was locking up due to running out of table-level locks available to it across all transactions. This occurs due to the `validate_anchors` function, which runs once per activity added, creating a temp table (which requires a table lock). This meant that branching a plan would require at least as many table locks as activities in the source plan, meaning that multiple users updating plans with a couple hundred activities could use up all of the table locks.

This has been fixed by extracting the temp table into a `STABLE` function `get_dependent_activities`. Because the function is `STABLE`, its value should be cached after running the first time in the transaction, so we shouldn't see much if any performance hit.

For information on how this bug was discovered and further discussion regarding adjusting the number of table locks we have in the DB (and why we have currently chosen not to), see [this Slack Thread](https://jpl.slack.com/archives/CJ07FGRPH/p1703184229901879).

## Verification

- A new load test was added, `db-lockup-test`, that replicates the scenario described in the ticket. This test fails on develop but passes when run on this branch.
- The hard coded "targeted plan id" in the existing load tests was replaced by a plan inserted during the setup phase. This avoids issues where there is no plan with id 1
- The existing load tests are no longer leaky. A `teardown` function has been added, tests that formerly selected random model ids always use the inserted model id, and tests that select random plan ids now use the inserted model id as a filter.

## Documentation

No docs need to be updated

## Future work

None currently needed. We do use temp tables in two other places: `begin_merge` and `metadata.validate_permissions_json`, but in both cases, it would **NOT** make sense to replace this with some other structure. In `begin_merge`, the function is continually updating its temp tables in order to properly fill out the state of the merge request at the end of the function. In `validate_permissions_json`, the two tables are generated by parsing JSON fields on trigger rows, making it possible but difficult to transform into a function. Additionally, these functions will always create a constant, low number of temp tables (3 and 2 respectively), instead of the variable amount seen when branching a plan. They are also run less frequently than `validate_anchors`.


load-tests/src/db-lockup-test.ts
```diff
@@ -0,0 +1,82 @@
...
+export const options = {
+  // A number specifying the number of VUs to run concurrently.
+  vus: 50,
+  // A string specifying the total duration of the test run.
+  duration: '10s',
+
+  // If the statistics of the load test fall below the following threshold, it with exit with a nonzero exit code
+  thresholds: {
+    http_req_failed: ['rate<0.01'], // http errors should be less than 1%
+    http_req_duration: ['p(50)<300', 'p(95)<1000'], // 50% below 300ms, 95% of requests should be below 1000ms
```
skovati, line 16: not a huge deal since this is going to need tuning based on compute resources (e.g. running these in GH actions), but this `http_req_duration` threshold actually fails for me locally.
Screenshot 2024-01-09 at 4 00 54 PM

skovati (approved):
Great work! I tested that the provided load test fails on develop and only passes once migration # 35 is applied (modulo `http_req_duration`). Thanks for refactoring the load tests and adding types!

load-tests/src/db-lockup-test.ts
Mythicaeda: Yeah, I've noticed that that threshold is flaky. I think setting it higher in this file might make sense (since duping a large plan is a large transaction), but I'm not sure what a good value is.

Issue #1277, opened 2024-01-11
Store Shared Versions of Scheduling Goals and Conditions


Mythicaeda:
Scheduling Goals and Conditions need to be versioned and those versions need to be reusable between plans.

PR #1278, opened 2024-01-11
Versioning Constraints - DON'T MERGE without UI

### LINKED ISSUE #882: Make sure constraint names are unique
Currently constraint names do not need to be unique. This can cause multiple violations to have the same name (violations do not point to a constraint via ID). We also throw an error when checking constraints if two constraints have the same name. This task is to just add a uniqueness constraint to the `name` column of the `constraint` table.

Additionally we should discuss and remove model constraints. See https://github.com/NASA-AMMOS/aerie/issues/892

### LINKED ISSUE #892: Remove ability to associate constraint with a model
Constraints only need to be associated with a plan. This is so constraints do not inadvertently change out from under a plan.

@JoelCourtney Notes we should have a nice way to copy constraints first otherwise this could be tedious. Imagine needing a set of 100 constraints for a new plan and needing to copy them over.

Related to https://github.com/NASA-AMMOS/aerie/issues/882

### LINKED ISSUE #1238: Store versions of constraints
Store and share versions of constraints

### LINKED ISSUE #1239: Evaluate only a specific version of plan level constraints
When not specified we should use the current behavior which is to evaluate using the latest value

Mythicaeda:
* **Tickets addressed:** 
    * Closes #1238,
    * Closes #892, 
    * Closes #1239, 
    * Closes #882 (for public constraints) 
* **Review:** By commit  

## Description

This PR is essentially an overhaul of how constraints were represented in the DB.

Below is a breakdown of the changes. Keep in mind when discussing permissions that unless a permission explicitly addresses the `aerie_admin` role, assume that it doesn't apply.
1) The `constraints` table has been replaced by the `constraint_metadata` and `constraint_definition` tables. The `constraint_definition` table contains the various versioned definitions of a constraint, while `constraint_metadata` contains the metadata consistent across all definitions (ie the name of the constraint). Once a version is created, it can only be updated by an admin user. The owner of a constraint can freely delete unused versions (see 3). Any user can create versions of public constraints (see 2). Both the overall constraint and each specific version can be tagged.
2) There is now the concept of public vs private constraints. All constraints start private unless otherwise specified and can be freely toggled between public and private. All public constraints must have unique names, but private constraints can have names that overlap. Only the owner of a private constraint can make new versions, and any user can make new versions of a public constraint. Only the owner of a private constraint can add it to constraint specifications (see 3), but once a private constraint has been added, any user able to edit the specification can freely manage the entry on the specification. Any user can add public constraints to constraint specifications they have permissions to manage. A constraint cannot be `SELECT`ed unless it is public, or the user doing the `SELECT` either owns the constraint or can manage a constraint specification it is on.
3) There is now the concept of a model and plan constraint specification, which replaces the former concept of "model constraints" and "plan constraints". A model specification is a recommended set of constraint and constraint versions for all plans using the model. A plan specification is the set of constraints that are checked during the action. If no version is specified (referred to as "version locking" in the tests), the latest version will be used. If a constraint is disabled on the specification, it will not be checked. When a plan is created, its constraint spec is prepopulated with the constraint spec of its model.

**QUICK LIST BREAKING CHANGES**:
- The structure of the return type for the `CheckConstraints` action has been altered:
    - the `type` field has been removed
    - the `constraintRevision` field has been added
- The `constraints` table has been split into two tables: `constraint_metadata` and `constraint_definition`. Additionally, plans and models now have constraint specifications. Any GQL queries referencing the `constraints` table will now need to reference at least one of these four tables, with which specific one varying based on what the query was doing.

## Verification

DB Tests:
- The DB Tags Tests were updated to test the new tags table
- Inserting a constraint in the DB has been updated to not depend on plans

e2eTests:
- Tests that checked/depended on `definitionOutdated` were removed
- ConstraintsTests for versioning and enabling/disabling constraints were added

## Documentation

[Open Docs ticket](https://github.com/NASA-AMMOS/aerie-docs/issues/120)

## Future work

[Scheduling Goals need to be changed in a similar way](https://github.com/NASA-AMMOS/aerie/issues/1277)


merlin-server/sql/merlin/tables/constraint_model_specification.sql
```diff
@@ -0,0 +1,27 @@
+create table constraint_model_specification(
+  model_id integer not null
+    references mission_model
+      on update cascade
+      on delete cascade,
+  constraint_id integer not null,
+  constraint_revision integer not null,
```
Mythicaeda: permit null here

deployment/hasura/migrations/AerieMerlin/36_constraint_versioning/up.sql
```diff
@@ -0,0 +1,281 @@
...
+'Whether to run a given constraint. Defaults to TRUE.';
+
+/*-- CONSTRAINT MODEL SPECIFICATION --*/
+create table constraint_model_specification(
+  model_id integer not null
+    references mission_model
+      on update cascade
+      on delete cascade,
+  constraint_id integer not null,
+  constraint_revision integer not null,
```
Mythicaeda: permit not null

deployment/hasura/migrations/AerieMerlin/37_constraint_versioning/up.sql
```diff
@@ -0,0 +1,279 @@
...
+  id integer generated by default as identity,
+
+  name text not null,
+  description text not null default '',
+  public boolean not null default false,
+
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+
+  owner text,
```
JoelCourtney: to make sure I'm understanding this right, the owner is not necessarily the person who created the constraint, it could be someone else and is configurable later, right?

Also, what case should the owner be null?

merlin-server/sql/merlin/tables/constraint_specification.sql
```diff
@@ -0,0 +1,28 @@
...
+  constraint_revision integer, -- latest is NULL
+  enabled boolean not null default true,
+
+  constraint constraint_specification_pkey
+    primary key (plan_id, constraint_id),
+  constraint plan_spec_references_existing_constraint
+    foreign key (constraint_id, constraint_revision)
+      references constraint_definition(constraint_id, revision)
+      on update cascade
+      on delete restrict
```
JoelCourtney, line 21: Seems like `restrict` could be a bit of a pain if they want to delete a definition that is used in a model spec, and therefore a bunch of plan specs. I'm not sure why it wouldn't be `cascade`. Are you intending that they won't ever need to delete a definition?

JoelCourtney (commented):
seems pretty clean and straightforward 👍 ! Just a couple questions

deployment/hasura/migrations/AerieMerlin/37_constraint_versioning/up.sql
Mythicaeda: Yeah, it will start as the person who created the constraint, but that person can pass their ownership to another user.

The only case the owner would be null is if the user who is the owner gets deleted (the `on delete set null` part of the `constraint_definition_author_exists` fkey).

merlin-server/sql/merlin/tables/constraint_specification.sql
Mythicaeda: I am not. Part of the user concerns that lead to versioning was the fear of constraints changing underfoot. Therefore, the decision was made that if a constraint definition is explicitly in use, it cannot be deleted. 

Ie, let's say constraint 10 has 2 definitions, `v1` and `v2`. If the model spec explicitly says to use `v2` (as in, the row on the column is literally `(10, 2)`), then that definition is in use and can't be deleted. If the model spec instead says to use whatever version is the latest (represented in the db as, `(10, NULL)`), then that definition isn't _explicitly_ in use and is fine to delete (from the model's POV, "latest" will become `v1`). 

Though this made me realize that we may need a restriction on the constraints definition table to prevent a constraint from having zero definitions after one's been created... 🤔

JoelCourtney (approved):
:shipit:

Mythicaeda:
Change just pushed was to tweak the Hasura Metadata to allow users to add and view all constraints on the plan's model spec, even if they are private. Users would've had issues when syncing their plan to the model spec without this change.

PR #1304, opened 2024-01-25
The timeline library

### LINKED ISSUE #1299: Timeline library
Provide a user-friendly library for querying and manipulating timeline objects (profiles, activity instances, directives). It should be well documented, tested, and have at least feature parity with the constraints package (but in a user-friendly and maintainable API).

JoelCourtney:
* **Tickets addressed:** closes #1299 

## Description
Creates a user-friendly and maintainable library for fetching and manipulating timeline objects, such as profiles, activity instances, and activity directives. This is more or less a giant rework of the backend constraints package, but since it required so much modification to make it user-friendly, I started over (while re-using bits and pieces of the old library for low-level functions, like `map2`).

I'll make a readme with a more formal description of the design before I do any walkthroughs, but the general idea is that all timeline classes (profiles, spans, activity instances, etc) contain a `Timeline` object, apply mixin interfaces to implement common operations, and then may have specialized operations if they would never be valid for any other timeline type.

You may notice that this is not Java 🤨. I wrote all this in Kotlin because
1. it general is cleaner and more concise
2. it has null safety in the type system
3. it doesn't actually matter to the user. It will look and behave and behave just like Java. Dokka, the kotlin documentation generator, can even output documentation that pretends the source code was actually the equivalent Java.

As for the code itself, all the operations are located in the mixin interfaces (the interfaces which end in `...Ops`). This idea turned out to be pretty complex, but it is the core of why this library is extensible and maintainable. It is so extensible that I could, for example, implement a Parallel (i.e. multi-valued) Windows class right now without re-implementing or even explicitly delegating the `.not` method; I'd just have to include the `BooleanOps` mixin. I haven't done this yet because there are some important mathematical considerations about how binary operations would work on parallel profiles, but that's another story.

I hope that, in the future, this design could be used to easily provide additional profile/spans types:
- time profiles (i.e. real profiles but for durations)
- event timelines
- parallel (multi-valued) profiles for reasoning about activity arguments without worrying about overlapping
- easy timeline type creation for users that have custom operations for special types in the mission model.

## Gradle commands

Build: `./gradlew :timeline:build`
Test: `./gradlew :timeline:test`
Package for testing in external projects: `./gradlew :timeline:shadowJar`
Generate documentation: `./gradlew :timeline:dokkaHtml` documentation is available in `timeline/build/dokka/html/index.html`

## Verification
- [x] unit tests are not yet complete

## Documentation
- [x] in-code documentation is complete

Tutorial documentation will come at the end of the procedural scheduling milestone.

## Future work
Aside from all the procedural scheduling work, this will make the vast majority of the constraints package obsolete, and once Clipper stops using it in the external scheduler, we could remove it entirely.


JoelCourtney:
I've run into two issues with dokka (Kotlin's documentation generator) that make the docs not quite as friendly as I'd like. They are still usable though.
- https://github.com/Kotlin/dokka/issues/3056 this one is specific to the kotlin-as-java plugin which makes kotlin docs look like they came from java code
- https://github.com/Kotlin/dokka/issues/3484 this one is a real issue regardless of the kotlin-as-java plugin. The workaround is to start each doc comment for a mixin method with a link to the method, which is ugly and time consuming.

JoelCourtney:
I ran into another problem with Dokka when trying to link the separate constraints library to the timeline library:

- https://github.com/Kotlin/dokka/issues/463

This means that the links on Violations (a constraints-only timeline type) to the default methods inherited from the ops interfaces are now *extra* broken. Even the manual links I put in the short descriptions do not work.

A possible workaround is to include a copy of the timeline library docs in the constraints docs, and not suppress them from the output. Not sure if this is worth it.

JoelCourtney:
I've marked this ready for review, but there are still a few things I need to do before merging
- [x] rebase to clean up history
- [x] unit tests for numeric profiles
- [x] e2e tests for AeriePostgresPlan

JoelCourtney:
@goetzrrGit @skovati I've rebased the entire thing to make the commit history cleaner. The general idea is that they are now sorted from most general/low-level to most specific/high-level. You'll need to force pull or hard reset.

JoelCourtney:
@goetzrrGit @skovati @bradNASA When looking at the generated docs, some of the features won't work due to CORS if you just open index.html directly. You can make IntelliJ run a little web server for you by finding the file, `Right click -> Open In -> Browser`. That way the search function and sidebar will work.

JoelCourtney:
Here is the project I demoed at the retro last week:
https://github.com/JoelCourtney/aerie-proc-constraint-demo

timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/Plan.kt
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Query a resource profile from the database
+   *
+   * @param ctor constructor of the profile, converting [SerializedValue]
+   * @param name string name of the resource
+   */
+  fun <V: Any, TL: CoalesceSegmentsOp<V, TL>> resource(name: String, ctor: (List<Segment<SerializedValue>>) -> TL): TL
+
+  /** Query all activity instances. */
+  fun allActivityInstances(): Instances<AnyInstance>
```
skovati: Is it possible to add some sort of `getActivityInstances(ActivityType)` or some other sort of filtering.

Same goes for directives.

timeline/build.gradle
skovati, line 1: Just curious if making this a `build.gradle.kts` would make for a cleaner integration with kotlin?

timeline/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+dependencies {
+  implementation project(':merlin-driver')
+  implementation project(':merlin-sdk')
+  implementation project(':parsing-utilities')
+
+  testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
+  testImplementation 'org.jetbrains.kotlin:kotlin-test-junit5'
+  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+
+//  dokkaHtmlPlugin 'org.jetbrains.dokka:kotlin-as-java-plugin:1.9.10'
```
skovati, line 23: Are the `dokka` bugs you mentioned related to this `as-java-plugin` format at all?

timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/AeriePostgresPlan.kt
```diff
@@ -0,0 +1,224 @@
...
+
+/** A connection to Aerie's database for a particular simulation result. */
+data class AeriePostgresPlan(
+    /** A connection to Aerie's database. */
+    val c: Connection,
+    /** The particular simulation dataset to query. */
+    val simDatasetId: Int
+): Plan {
+
+  private val datasetId by lazy {
```
skovati, line 35: dang... `by lazy` is _sweet_

skovati (commented):
Early reporting some questions / comments I had to allow for some concurrency in review

timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/Plan.kt
JoelCourtney: Yes, we can do that now, thanks for reminding me. I'm hoping to link to the mission model after MVP to allow you to deserialize as the original activity type, but I think I can do this now in a way that will be backward compatible.

Will resolve when done

timeline/build.gradle
JoelCourtney: I don't think so. Groovy or Kotlin for gradle is an independent choice from everything else going on. I don't really know what advantages kotlin gives over groovy, so I just stuck with what I'm familiar with. Not opposed to it though

timeline/build.gradle
JoelCourtney: Yes, one of them is. Static functions in interfaces are difficult to find when using the kotlin-as-java plugin. I'm still undecided if we should use it

timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/payloads/activities/Activity.kt
```diff
@@ -0,0 +1,13 @@
+package gov.nasa.jpl.aerie.timeline.payloads.activities
+
+import gov.nasa.jpl.aerie.timeline.Duration
+import gov.nasa.jpl.aerie.timeline.payloads.IntervalLike
+
+/** Unifying interface for activity instances and directives. */
+interface Activity<A: Any>: IntervalLike<A> {
```
mattdailis: Still building a mental model for this terminology - just looking at this line, I'm curious how a directive is represented as an interval. Is it instantaneous, or is there some notion of "unknown" end time...?

JoelCourtney: Yeah this is a weird one. In this case, `Directive` contains only its start time, and implements the `interval` property of `IntervalLike` by returning an instantaneous interval at the startTime. The `Directives` timeline is a bit unique in that all of its objects are guaranteed to have zero duration. So it is the only timeline type so far that does not apply (directly or indirectly) the `NonZeroDurationOps` mixin.

timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/Plan.kt
JoelCourtney: done

skovati (approved):
This looks really great overall, excellent job!

Any further questions I have are non-blocking and shouldn't hold up this PR, especially considering this is the first step of a larger work-in-progress, and I expect further iteration and refinement, so I'm GO for merge 🚀

PR #1313, opened 2024-02-01
Use dynamics instead of resources in LBCS

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
The LinearBoundaryConsistencySolver, a linear solver for polynomial resources driven by a boundary-consistency algorithm, currently computes its results using resources. This involves some more indirection than is necessary.

By changing LBCS to use dynamics instead of resources, we avoid that extra overhead, and improve performance somewhat.

## Verification
Existing unit test `LinearBoundaryConsistencySolverTest` continued to pass without modification, and Brad's branch of the SRL model has been using these changes successfully.

## Documentation
N/A - Refactor only

## Future work
N/A


bradNASA (approved):
I tested this with the SRL model for the battery and for the data model, and feel good about it.

mattdailis (approved):

PR #1314, opened 2024-02-01
Add more utilities for profiling resources

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Adds more support for profiling resource performance, especially around mutable resources.
Most importantly, it adds the static method `Resource.profileAllResources()`.
When called before constructing the model, this method turns on profiling for virtually every resource in the model.
This casts a broad net for early stages in a performance investigation, to highlight resources that are called unexpectedly frequently.

## Verification

These changes have been spot-checked by hand in the streamline example model and with the SRL model. As developer-oriented debugging tools with loose requirements, I don't think automated tests are appropriate for most of the behavior here.

## Documentation

Only in-line javadoc comments.

## Future work

N/A


bradNASA (approved):
I used this in the SRL model to help identify bottlenecks.  It was very useful.  I didn't see anything questionable in the edits here, but I did not get zen understanding of it, either.

mattdailis (approved):

PR #1322, opened 2024-02-06
Bugfix for polynomial root-finding

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit

## Description
Fixes a few bugs related to rootfinding and expiry-propagation for Polynomial resources. The most important of these fixes are:
* Taking expiry into account for dynamicsChange condition. This was an unusual case exposed by the solver: A solution could expire early due to inputs that expire, but the updated inputs return the same result. In this case, the solution data remains the same, but the solution expiry is extended. This needs to count as the dynamics "changing", since downstream consumers may be interested in the expiry, not just the data.
* Stability and performance improvements for Polynomial root-finding.
    * For stability, addresses an edge case where binary search for the exact root could overflow, and conditions the problem for the Laguerre solver by normalizing the polynomial's constant term to 1. This lets the solver converge in a few extra edge cases without affecting "regular" cases much.
    * For performance, adds a special case for linear polynomials, where we directly solve for the root instead of using the Laguerre solver. This tends to be faster and possibly more stable, and a large proportion of real-world cases tend to be linear.
* Passing expiry through the solver and clampedIntegrate functions. These components used to erase expiry information, as a patch over the bug in the dynamicsChange condition described above (though I didn't know at the time that was why erasing expiry information helped.) Now, by more carefully and selectively erasing expiry information outside the solver, we can preserve expiry information from inputs to the solver and clampedIntegrate through to the output. This in turn lets downstream components like approximations make better decisions. In particular, we have a use case for the SRL model with an approximation of a quadratic clamped integral. If we don't pass expiry information through, then the approximation "overshoots" the expiry, is interrupted by the next quadratic segment, and produces a discontinuity in the output. By passing the expiry through to the approximation, the approximation can align the approximation segment's endpoint with the expiry, and hence avoid this discontinuity.

Before:
image

After
image

## Verification
The changes in polynomial root-finding come with additional unit tests to detect some of these edge cases. Additionally, Brad has tested these changes with the SRL model, with positive results.

## Documentation
N/A

## Future work
N/A

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/core/Expiry.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/core/Expiry.java
@@ -8,7 +8,7 @@ import java.util.stream.Stream;
 /**
  * The time at which a value expires.
  */
-public record Expiry(Optional<Duration> value) {
+public record Expiry(Optional<Duration> value) implements Comparable<Expiry> {
   public static Expiry NEVER = expiry(Optional.empty());
 
   public static Expiry at(Duration t) {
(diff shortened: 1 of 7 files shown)


bradNASA (approved):
I've tested this.  SRL is having to use contrib jars from this branch until merged; the model is broken without it.  The performance improvement for SRL seems dramatic, maybe twice as fast.

mattdailis (approved):

Issue #1025, opened 2023-07-11
Sequencing Server crashes creating a large expansion set


cartermak:
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

1.6.1

### Describe the bug

When creating a moderate-sized expansion set (73 activity types) on a local instance of Aerie, the sequencing server crashes and fails to create the expansion set (presumably, during the process of validating the expansion rules against the activity schema). Increasing the memory allocated to Docker to 32 GB seems to have resolved the issue, but I haven't tested this extensively.

### Reproduction

On a local instance of Aerie with Docker allocated 16 GB of memory, create an expansion set with >70 activity types (arbitrary number). The container crashes and restarts; Hasura returns a "http exception when calling webhook" for the `createExpansionSet` action. 

A possible compounding factor is the size of the command dictionary.

Logs included below from the sequencing server container, although they don't seem too insightful.

### Logs

```shell
0 verbose cli /usr/local/bin/node /usr/local/bin/npm
1 info using npm@8.19.3
2 info using node@v18.13.0
3 timing npm:load:whichnode Completed in 0ms
4 timing config:load:defaults Completed in 0ms
5 timing config:load:file:/usr/local/lib/node_modules/npm/npmrc Completed in 0ms
6 timing config:load:builtin Completed in 1ms
7 timing config:load:cli Completed in 1ms
8 timing config:load:env Completed in 1ms
9 timing config:load:file:/app/.npmrc Completed in 2ms
10 timing config:load:project Completed in 7ms
11 timing config:load:file:/root/.npmrc Completed in 0ms
12 timing config:load:user Completed in 0ms
13 timing config:load:file:/usr/local/etc/npmrc Completed in 0ms
14 timing config:load:global Completed in 0ms
15 timing config:load:validate Completed in 1ms
16 timing config:load:credentials Completed in 1ms
17 timing config:load:setEnvs Completed in 0ms
18 timing config:load Completed in 13ms
19 timing npm:load:configload Completed in 14ms
20 timing npm:load:mkdirpcache Completed in 1ms
21 timing npm:load:mkdirplogs Completed in 1ms
22 verbose title npm start
23 verbose argv "start"
24 timing npm:load:setTitle Completed in 1ms
25 timing config:load:flatten Completed in 2ms
26 timing npm:load:display Completed in 2ms
27 verbose logfile logs-max:10 dir:/root/.npm/_logs
[...]

PR #1324, opened 2024-02-07
Improve large expansion sets

goetzrrGit:
* **Tickets addressed:** #1025 
* **Review:** By commit  

## Description
Part 1 of 2 for #1025 

This PR addresses a stability issue where creating large expansion sets caused server crashes due to excessive resource consumption by worker processes. Here's a summary of the improvements:

### Problem:

- Compiling individual authored logic large expansion sets led to high memory usage by worker processes.
- Queuing multiple compilation jobs per worker exacerbated the issue, contributing to server crashes.
- Workers had not cap on resources 

### Solution:

- Chunking: Implemented chunking to distribute authored logic compilation tasks among a subset of workers, preventing resource overload on individual processes.
- Caching: Introduced a TS transpilation cache to eliminate repetitive compilations for the same logic units, significantly reducing memory usage and avoiding unnecessary worker load.
- Exposing worker properties to fine-tune workers per project

By default, I am spinning up 8 workers and giving them 1GB of heap. 

## Verification
I was able to create an expansion set with 73 authored logic without a crash. Before the fix 20+ would crash the server.

## Future work
Implement a single background worker to transpile the authoring logic while the server is idle and cache the results. Help with the huge upfront cost .ex 70+ expansion logic takes about 13 minutes. This is cpu bounded so a beefier computer would help.

### DIFF
--- a/sequencing-server/src/routes/command-expansion.ts
+++ b/sequencing-server/src/routes/command-expansion.ts
@@ -1,6 +1,6 @@
-import type { CacheItem, UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
+import type { UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
 import pgFormat from 'pg-format';
-import { Context, db, piscina } from './../app.js';
+import { Context, db, piscina, promiseThrottler, typeCheckingCache } from './../app.js';
 import { Result } from '@nasa-jpl/aerie-ts-user-code-runner/build/utils/monads.js';
 import express from 'express';
 import { serializeWithTemporal } from './../utils/temporalSerializers.js';
@@ -13,6 +13,7 @@ import { unwrapPromiseSettledResults } from '../lib/batchLoaders/index.js';
 import { defaultSeqBuilder } from '../defaultSeqBuilder.js';
 import { ActivateStep, CommandStem, LoadStep } from './../lib/codegen/CommandEDSLPreface.js';
 import { getUsername } from '../utils/hasura.js';
+import * as crypto from 'crypto';
 
 const logger = getLogger('app');
 
@@ -54,17 +55,18 @@ commandExpansionRouter.post('/put-expansion', async (req, res, next) => {
     activityTypeName,
   });
   const activityTypescript = generateTypescriptForGraphQLActivitySchema(activitySchema);
-
-  const result = Result.fromJSON(
-    await (piscina.run(
-      {
-        expansionLogic,
-        commandTypes: commandTypes,
-        activityTypes: activityTypescript,
-      },
-      { name: 'typecheckExpansion' },
-    ) as ReturnType<typeof typecheckExpansion>),
-  );
+  const result = await promiseThrottler.run(() => {
+    return (
+      piscina.run(
+        {
+          commandTypes: commandTypes,
+          activityTypes: activityTypescript,
+          activityTypeName: activityTypeName,
+        },
+        { name: 'typecheckExpansion' },
+      ) as ReturnType<typeof typecheckExpansion>
+    ).then(Result.fromJSON);
+  });
 
   res.status(200).json({ id, errors: result.isErr() ? result.unwrapErr() : [] });
   return next();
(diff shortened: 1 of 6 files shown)


dyst5422:
I like the solution here with the caching.

I had a similar need recently for throttling the number of inflight promises and came up with this solution which might be inspiration for a less intrusive solution to the batching. Benefit here is that it allows us to limit the number of in process workers, BUT doesn't block on all of them completing. So we can ensure the workers are all always filled.

```ts
/** Throttle number of in flight promises. */
export class InFlightPromiseThrottler {
  private inFlightPromises: Promise<unknown>[] = [];
  private promiseLimit: number;

  public constructor(promiseLimit: number) {
    this.promiseLimit = promiseLimit;
  }

  public async run<T>(promiseFactory: () => Promise<T>): Promise<T> {
    while (this.inFlightPromises.length >= this.promiseLimit) {
      await Promise.race(this.inFlightPromises);
    }
    const promise = promiseFactory();
    this.inFlightPromises.push(promise);
    return promise.finally(() => {
      const index = this.inFlightPromises.indexOf(promise);
      if (index !== -1) {
        this.inFlightPromises.splice(index, 1);
      }
    });
  }
}
```

Used like
```ts
// Somewhere in a higher scope
const workerPromiseThrottler = new InFlightPromiseThrottler(8);

// When we actually do the worker run

const result = await workerPromiseThrottler.run(() => {\\ async code calling piscina.run});
```

goetzrrGit:
> I like the solution here with the caching.
> 
> I had a similar need recently for throttling the number of inflight promises and came up with this solution which might be inspiration for a less intrusive solution to the batching. Benefit here is that it allows us to limit the number of in process workers, BUT doesn't block on all of them completing. So we can ensure the workers are all always filled.
> 
> ```ts
> /** Throttle number of in flight promises. */
> export class InFlightPromiseThrottler {
>   private inFlightPromises: Promise<unknown>[] = [];
>   private promiseLimit: number;
> 
>   public constructor(promiseLimit: number) {
>     this.promiseLimit = promiseLimit;
>   }
> 
>   public async run<T>(promiseFactory: () => Promise<T>): Promise<T> {
>     while (this.inFlightPromises.length >= this.promiseLimit) {
>       await Promise.race(this.inFlightPromises);
>     }
>     const promise = promiseFactory();
>     this.inFlightPromises.push(promise);
>     return promise.finally(() => {
>       const index = this.inFlightPromises.indexOf(promise);
>       if (index !== -1) {
>         this.inFlightPromises.splice(index, 1);
>       }
>     });
>   }
> }
> ```
> 
> Used like
> 
> ```ts
> // Somewhere in a higher scope
> const workerPromiseThrottler = new InFlightPromiseThrottler(8);
> 
> // When we actually do the worker run
> 
> const result = await workerPromiseThrottler.run(() => {\\ async code calling piscina.run});
> ```

Hi Dylan, Thank you for the suggestion!!! Very cool approach and I will give it a try.  I hope all is well with you.

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -380,18 +425,19 @@ commandExpansionRouter.post('/expand-all-activity-instances', async (req, res, n
...
+                case 'command':
+                  return CommandStem.fromSeqJson(c);
+                case 'load':
+                  return LoadStep.fromSeqJson(c);
+                case 'activate':
+                  return ActivateStep.fromSeqJson(c);
+                default:
+                  throw new Error(`Unknown command type: ${c.type}`);
+              }
+            }) ?? null,
```
goetzrrGit: Intelij reformatted this. Ignore...

sequencing-server/src/utils/PromiseThrottler.ts
```diff
@@ -0,0 +1,22 @@
...
+  private runningPromises: Promise<unknown>[] = [];
+  private promiseLimit: number;
+
+  public constructor(promiseLimit: number) {
+    this.promiseLimit = promiseLimit;
+  }
+
+  public async run<T>(promiseFactory: () => Promise<T>): Promise<T> {
+    while (this.runningPromises.length >= this.promiseLimit) {
+      await Promise.race(this.runningPromises);
```
cohansen, line 11: Had to lookup this method, neat!

cohansen (approved):
Looks good to me, I saw your demo as well and it seemed like it was working well. My only question is do we need to enforce that `PromiseThrottler` is a singleton?

goetzrrGit:
> Looks good to me, I saw your demo as well and it seemed like it was working well. My only question is do we need to enforce that `PromiseThrottler` is a singleton?

I am using it as a singleton export but there is a ticket I created to use dependency injection so it wont be a singleton in the future.

https://github.com/NASA-AMMOS/aerie/issues/1350

PR #1328, opened 2024-02-13
Add Missing Filter in Commit Merge

Mythicaeda:
* **Review:** By commit  

## Description

Upserting the Presets for Activity Directives was missing filters on the snapshot id and the merge request id.

## Verification

DBTests were added.
- The test `presetOnlyPullsFromSourceSnapshot` asserts that the potential bug caused by a lack of snapshot id filter does not occur. 
- The test `presetUnaffectedByUnrelatedSnapshot` mirrors a strange scenario seen on a Clipper venue that was causing merge issues, where for some reason a preset recorded on an unrelated plan's snapshot was attempting to be set when committing the merge.

## Documentation

No docs need to be updated


skovati (approved):
LGTM! 🚀

PR #1332, opened 2024-02-15
Sequence background transpiler

goetzrrGit:
* **Tickets addressed:** #1324 
* **Review:** By commit  

## Description
Part 2 of 2 for #1324 

You need PR https://github.com/NASA-AMMOS/aerie/pull/1324 merged first

This PR introduces a background transpiler process to help improve the upfront cost of creating an expansion set or sequence expansion

### Changes:

* Fetches the latest command dictionary, mission model, and expansion logic every 5 minutes
* Transpiles and caches the results

## Verification
Manual testing with Clipper mission model and command dictionary

### DIFF
--- a/sequencing-server/src/app.ts
+++ b/sequencing-server/src/app.ts
@@ -27,6 +27,7 @@ import { getHasuraSession, canUserPerformAction, ENDPOINTS_WHITELIST } from './u
 import type { Result } from '@nasa-jpl/aerie-ts-user-code-runner/build/utils/monads';
 import type { CacheItem, UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
 import { PromiseThrottler } from './utils/PromiseThrottler.js';
+import { backgroundTranspiler } from './backgroundTranspiler.js';
 
 const logger = getLogger('app');
 
@@ -39,7 +40,9 @@ app.use(bodyParser.json({ limit: '100mb' }));
 
 DbExpansion.init();
 export const db = DbExpansion.getDb();
-
+export let graphqlClient = new GraphQLClient(getEnv().MERLIN_GRAPHQL_URL, {
+  headers: { 'x-hasura-admin-secret': getEnv().HASURA_GRAPHQL_ADMIN_SECRET },
+});
 export const piscina = new Piscina({
   filename: new URL('worker.js', import.meta.url).pathname,
   minThreads: parseInt(getEnv().SEQUENCING_WORKER_NUM),
@@ -62,10 +65,6 @@ export type Context = {
 };
 
 app.use(async (req: Request, res: Response, next: NextFunction) => {
-  const graphqlClient = new GraphQLClient(getEnv().MERLIN_GRAPHQL_URL, {
-    headers: { 'x-hasura-admin-secret': getEnv().HASURA_GRAPHQL_ADMIN_SECRET },
-  });
-
   // Check and make sure the user making the request has the required permissions.
   if (
     !ENDPOINTS_WHITELIST.has(req.url) &&
@@ -247,4 +246,30 @@ app.listen(PORT, () => {
   logger.info(`Worker pool initialized:
               Total workers started: ${piscina.threads.length},
               Heap Size per Worker: ${getEnv().SEQUENCING_MAX_WORKER_HEAP_MB} MB`);
+
+  if (getEnv().TRANSPILER_ENABLED === 'true') {
+    //log that the tranpiler is on
+    logger.info(`Background Transpiler is 'on'`);
+
+    let transpilerPromise: Promise<void> | undefined; // Holds the transpilation promise
+    async function invokeTranspiler() {
+      try {
+        await backgroundTranspiler();
+      } catch (error) {
+        console.error('Error during transpilation:', error);
+      } finally {
+        transpilerPromise = undefined; // Reset promise after completion
+      }
+    }
+
+    // Immediately call the background transpiler
+    transpilerPromise = invokeTranspiler();
+
+    // Schedule next execution after 2 minutes, handling ongoing transpilation
+    setInterval(async () => {
+      if (!transpilerPromise) {
+        transpilerPromise = invokeTranspiler(); // Start a new transpilation
+      }
+    }, 60 * 2 * 1000);
+  }
 });
--- a/sequencing-server/src/backgroundTranspiler.ts
+++ b/sequencing-server/src/backgroundTranspiler.ts
@@ -1,0 +1,143 @@
+import { piscina, graphqlClient, typeCheckingCache, promiseThrottler } from './app.js';
+import { objectCacheKeyFunction } from './lib/batchLoaders/index.js';
+import crypto from 'crypto';
+import { generateTypescriptForGraphQLActivitySchema } from './lib/codegen/ActivityTypescriptCodegen.js';
+import DataLoader from 'dataloader';
+import { activitySchemaBatchLoader } from './lib/batchLoaders/activitySchemaBatchLoader.js';
+import { commandDictionaryTypescriptBatchLoader } from './lib/batchLoaders/commandDictionaryTypescriptBatchLoader.js';
+import type { typecheckExpansion } from './worker';
+import { Result } from '@nasa-jpl/aerie-ts-user-code-runner/build/utils/monads.js';
+import { getLatestCommandDictionary, getLatestMissionModel, getExpansionRule } from './utils/hasura.js';
+
+export async function backgroundTranspiler(numberOfThreads: number = 2) {
+  if (graphqlClient === null) {
+    return;
+  }
+
+  // Fetch latest mission model
+  const { mission_model_aggregate } = await getLatestMissionModel(graphqlClient);
+  if (!mission_model_aggregate) {
+    console.log(
+      '[ Background Transpiler ] Unable to fetch the latest mission model. Aborting background transpiling...',
+    );
+    return;
+  }
+
+  // Fetch latest command dictionary
+  const { command_dictionary_aggregate } = await getLatestCommandDictionary(graphqlClient);
+  if (!command_dictionary_aggregate) {
+    console.log(
+      '[ Background Transpiler ] Unable to fetch the latest command dictionary. Aborting background transpiling...',
+    );
+    return;
+  }
+
+  const commandDictionaryId = command_dictionary_aggregate.aggregate.max.id;
+  const missionModelId = mission_model_aggregate.aggregate.max.id;
+  const { expansion_rule } = await getExpansionRule(graphqlClient, missionModelId, commandDictionaryId);
+
+  if (expansion_rule === null || expansion_rule.length === 0) {
+    console.log(`[ Background Transpiler ] No expansion rules to transpile.`);
+    return;
+  }
+
+  const commandTypescriptDataLoader = new DataLoader(commandDictionaryTypescriptBatchLoader({ graphqlClient }), {
+    cacheKeyFn: objectCacheKeyFunction,
+    name: null,
+  });
+  const activitySchemaDataLoader = new DataLoader(activitySchemaBatchLoader({ graphqlClient }), {
+    cacheKeyFn: objectCacheKeyFunction,
+    name: null,
+  });
+
+  const commandTypes = await commandTypescriptDataLoader.load({
+    dictionaryId: missionModelId,
+  });
+
+  if (commandTypes === null) {
+    console.log(`[ Background Transpiler ] Unable to fetch command ts lib.
+    Aborting transpiling...`);
+    return;
+  }
+
+  // only process 'numberOfThreads' worth at a time ex. transpile 2 logics at a time
+  // This allows for expansion set and sequence expansion to utilize the remaining workers
+  for (let i = 0; i < expansion_rule.length; i += numberOfThreads) {
+    await Promise.all(
+      expansion_rule.slice(i, i + numberOfThreads).map(async expansion => {
+        await promiseThrottler.run(async () => {
+          // Assuming expansion_rule elements have the same type
+          if (expansion instanceof Error) {
+            console.log(`[ Background Transpiler ] Expansion: ${expansion.name} could not be loaded`, expansion);
+            return Promise.reject(`Expansion: ${expansion.name} could not be loaded`);
+          }
+
+          const hash = crypto
+            .createHash('sha256')
+            .update(
+              JSON.stringify({
+                commandDictionaryId,
+                missionModelId,
+                id: expansion.id,
+                expansionLogic: expansion.expansion_logic,
+                activityType: expansion.activity_type,
+              }),
+            )
+            .digest('hex');
+
+          // ignore already transpiled hash info
+          if (typeCheckingCache.has(hash)) {
+            return Promise.resolve();
+          }
+
+          const activitySchema = await activitySchemaDataLoader.load({
+            missionModelId,
+            activityTypeName: expansion.activity_type,
+          });
+
+          // log error
+          if (!activitySchema) {
+            console.log(
+              `[ Background Transpiler ] Activity schema for ${expansion.activity_type} could not be loaded`,
+              activitySchema,
+            );
+            return Promise.reject('Activity schema for ${expansion.activity_type} could not be loaded');
+          }
+
+          const activityTypescript = generateTypescriptForGraphQLActivitySchema(activitySchema);
+
+          // log error
+          if (!activityTypescript) {
+            console.log(
+              `[ Background Transpiler ] Unable to generate typescript for activity ${expansion.activity_type}`,
+              activityTypescript,
+            );
+            return Promise.reject(`Unable to generate typescript for activity ${expansion.activity_type}`);
+          }
+
+          const typecheckingResult = (
+            piscina.run(
+              {
+                expansionLogic: expansion.expansion_logic,
+                commandTypes: commandTypes,
+                activityTypes: activityTypescript,
+                activityTypeName: expansion.activity_type,
+              },
+              { name: 'typecheckExpansion' },
+            ) as ReturnType<typeof typecheckExpansion>
+          ).then(Result.fromJSON);
+
+          //Display any errors
+          typecheckingResult.then(result => {
+            if (result.isErr()) {
+              console.log(`Error transpiling ${expansion.activity_type}:\n ${result.unwrapErr().map(e => e.message)}`);
+            }
+          });
+
+          typeCheckingCache.set(hash, typecheckingResult);
+          return typecheckingResult;
+        });
+      }),
+    );
+  }
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -95,6 +95,7 @@ services:
       SEQUENCING_LOCAL_STORE: /usr/src/app/sequencing_file_store
       SEQUENCING_WORKER_NUM: 8
       SEQUENCING_MAX_WORKER_HEAP_MB: 1000
+      TRANSPILER_ENABLED : "true"
     image: aerie_sequencing
     ports: ["27184:27184"]
     restart: always
--- a/sequencing-server/src/env.ts
+++ b/sequencing-server/src/env.ts
@@ -12,6 +12,7 @@ export type Env = {
   STORAGE: string;
   SEQUENCING_WORKER_NUM: string;
   SEQUENCING_MAX_WORKER_HEAP_MB: string;
+  TRANSPILER_ENABLED: string;
 };
 
 export const defaultEnv: Env = {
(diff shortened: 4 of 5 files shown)


sequencing-server/src/app.ts
```diff
@@ -234,4 +243,14 @@ app.use((err: any, _: Request, res: Response, next: NextFunction) => {
...
+              Total workers started: ${piscina.threads.length},
+              Heap Size per Worker: ${getEnv().SEQUENCING_MAX_WORKER_HEAP_MB} MB`);
 });
+
+// Immediately call the background transpiler
+await backgroundTranspiler();
+// Then continue calling it every 5 minutes
+setInterval(async () => {
+  await backgroundTranspiler();
+}, 60 * 5 * 1000);
```
cohansen: Is there ever a case where we don't want the background transpiler at all? Maybe add in a way to disable it?

sequencing-server/src/app.ts
```diff
@@ -36,8 +40,17 @@ app.use(bodyParser.json({ limit: '100mb' }));
 
 DbExpansion.init();
 export const db = DbExpansion.getDb();
+export let graphqlClient = new GraphQLClient(getEnv().MERLIN_GRAPHQL_URL, {
+  headers: { 'x-hasura-admin-secret': getEnv().HASURA_GRAPHQL_ADMIN_SECRET },
+});
+export const piscina = new Piscina({
+  filename: new URL('worker.js', import.meta.url).pathname,
+  minThreads: parseInt(getEnv().SEQUENCING_WORKER_NUM),
+  resourceLimits: { maxOldGenerationSizeMb: parseInt(getEnv().SEQUENCING_MAX_WORKER_HEAP_MB) },
```
cohansen: Default these env variables to a reasonable value in case they're not set?

sequencing-server/src/backgroundTranspiler.ts
```diff
@@ -0,0 +1,230 @@
...
+
+          typeCheckingCache.set(hash, typecheckingResult);
+          return typecheckingResult;
+        });
+      }),
+    );
+  }
+}
+
+async function fetchLatestMissionModel(): Promise<{
```
cohansen: Can we move all of the GQL function calls out to another file? We have `hasura.ts` with our existing GQL calls right now.

cohansen (changes requested):
I see your comment about merging this into the other PR, this was just an easier first review since it was more digestible.

sequencing-server/src/app.ts
goetzrrGit: Checkout `getEnv()` in `env.ts` If there isn't one provided in the environment from the docker-compose it will use the defaults.

sequencing-server/src/backgroundTranspiler.ts
goetzrrGit: Sure, I can do that

sequencing-server/src/app.ts
goetzrrGit: Added an option to turn it off via the environmental variables.

dyst5422:
Just want to ask a perhaps dumb question. This essentially polls every 5 minutes. Is there any way we can change from a pull to a push model? IE be notified of changes and trigger transpilation, but throttle it to every 5 minutes? That way, no cycles are wasted pulling and transpiling when there are no changes?

goetzrrGit:
@dyst5422 I like the idea of combining these two concepts. We can trigger transpilation whenever the user uploads a new command dictionary or mission model. The current polling mechanism can help address edge cases where users operate outside the sequencing server and Hasura GraphQL Actions. This typically occurs when users bypass the `/put-expansion `endpoint, often in cases like Clipper, which heavily relies on the API and not Aerie UI. 

@dandelany Do you want me to explore this option, or create a ticket to address this later? I want to get these new changes to Clipper to try it out and see if it helps them in the thread test.

sequencing-server/src/app.ts
```diff
@@ -234,4 +243,18 @@ app.use((err: any, _: Request, res: Response, next: NextFunction) => {
...
+              Heap Size per Worker: ${getEnv().SEQUENCING_MAX_WORKER_HEAP_MB} MB`);
 });
+
+if (getEnv().TRANSPILER_ENABLED === 'true') {
+  //log that the tranpiler is on
+  logger.info(`Background Transpiler is 'on'`);
+  // Immediately call the background transpiler
+  await backgroundTranspiler();
+  // Then continue calling it every 5 minutes
+  setInterval(async () => {
```
dandelany: discussed with @goetzrrGit - this will not wait for previous transpile to be done before setting timer - will update

goetzrrGit: Fixed

dyst5422:
> @dyst5422 I like the idea of combining these two concepts. We can trigger transpilation whenever the user uploads a new command dictionary or mission model. The current polling mechanism can help address edge cases where users operate outside the sequencing server and Hasura GraphQL Actions. This typically occurs when users bypass the `/put-expansion `endpoint, often in cases like Clipper, which heavily relies on the API and not Aerie UI.

Sorry for the ignorance, how does use of the api by clipper bypass `/put-expansion`?

goetzrrGit:
@dyst5422 Not at all, I noticed Aerie UI was doing this. When the user creates a new authoring logic it does not use the `/put-expansion` endpoint. It is using a different GQL call which bypasses the Hasura action and server. I am sure Clipper or other users can do the same. The problem is the user will always do something different than intended so the pull is a good way to catch those edge-cases. I  think the hybrid model will be good which will have the responsiveness of the push model but keep the pull model as a backup for the time the user does something they are not supposed to. 

I talked to @dandelany and  @joswig about this yesterday and we were going to make a ticket to add a push model to the code at a later time. We had some important EDSL tickets come from clipper that me and Matt need to prototype.

dyst5422:
> @dyst5422 Not at all, I noticed Aerie UI was doing this. When the user creates a new authoring logic it does not use the `/put-expansion` endpoint. It is using a different GQL call which bypasses the Hasura action and server. I am sure Clipper or other users can do the same. The problem is the user will always do something different than intended so the pull is a good way to catch those edge-cases. I think the hybrid model will be good which will have the responsiveness of the push model but keep the pull model as a backup for the time the user does something they are not supposed to.
> 
> I talked to @dandelany and @joswig about this yesterday and we were going to make a ticket to add a push model to the code at a later time. We had some important EDSL tickets come from clipper that me and Matt need to prototype.

It's concerning to me that there are multiple paths to get an expansion in the system as we clearly have processes we want to execute as part of the paths through and don't have control over that flow. Is there interest in closing those alternate paths to getting an expansion into the system?

100% support the appropriate prioritization.

sequencing-server/src/app.ts
dyst5422: I think this might be better expressed as a combination of a `throttle` with a function that calls itself on completion. I think the current implementation could end up with a 10 minute gap if the background transpiler is still running when the next invocation is intended to occur rather than an immediate invocation.

in psuedocode
```typescript
// The throttle wrap is lightly stateful, so we need to hold onto the same instance.
const trottledBackgroundTranspiler = throttle(backgroundTranspiler, 60 * 5 * 1000);
// This will trigger another run on completion. We need this here so that we attach the finally call to the promise returned each time its called
const runBackgroundTranspiler = () => trottledBackgroundTranspiler().finally(runBackgroundTranspiler);

// initially kick it off
runBackgrounTranspiler();
```

The throttle will delay execution for a minimum of 5 minutes after the previous execution, and the recursive call in finally will trigger the next invocation on completion (note that this is not a properly recursive call in the sense of adding another frame onto the stack, so won't produce a stack overflow).

A library like lodash will have a solid throttle function in it.

goetzrrGit: I did some testing and I think we are generating too much data on the heap with the throttle. I had the server up for about an hour in a meeting. 

```
[node] <--- Last few GCs --->
[node] 
[node] [1355:0x7f460ceb82a0]    82200 ms: Scavenge 2036.8 (2078.8) -> 2035.2 (2079.6) MB, 8.7 / 0.0 ms  (average mu = 0.152, current mu = 0.122) allocation failure; 
[node] [1355:0x7f460ceb82a0]    82219 ms: Scavenge 2037.4 (2079.6) -> 2036.2 (2081.1) MB, 16.8 / 0.0 ms  (average mu = 0.152, current mu = 0.122) allocation failure; 
[node] [1355:0x7f460ceb82a0]    82238 ms: Scavenge 2039.0 (2081.1) -> 2037.5 (2090.1) MB, 16.6 / 0.0 ms  (average mu = 0.152, current mu = 0.122) allocation failure; 
[node] 
[node] 
[node] <--- JS stacktrace --->
[node] 
[node] FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
[node] Program node --inspect=0.0.0.0 --experimental-vm-modules ./build/app.js exited with code null
```

goetzrrGit: I did try playing around with your suggestion but I keep getting a heap out of bounds. 

```
<--- Last few GCs --->
[node] 
[node] [1655:0x7fad990f52a0]    97704 ms: Scavenge 2027.3 (2069.3) -> 2026.1 (2071.1) MB, 11.1 / 0.0 ms  (average mu = 0.183, current mu = 0.146) allocation failure; 
[node] [1655:0x7fad990f52a0]    97718 ms: Scavenge 2028.8 (2071.1) -> 2027.5 (2080.1) MB, 12.2 / 0.0 ms  (average mu = 0.183, current mu = 0.146) allocation failure; 
[node] [1655:0x7fad990f52a0]    97737 ms: Scavenge 2033.8 (2080.1) -> 2030.7 (2081.3) MB, 13.9 / 0.0 ms  (average mu = 0.183, current mu = 0.146) allocation failure; 
[node] 
[node] 
[node] <--- JS stacktrace --->
[node] 
[node] FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
[node] Program node --inspect=0.0.0.0 --experimental-vm-modules ./build/app.js exited with code null
```

goetzrrGit: I tweaked the code a little but noticed that the server will not activate the transpiler after idle for more than 5 minutes. 

```
  // The throttle wrap is lightly stateful, so we need to hold onto the same instance.
  const trottledBackgroundTranspiler = throttle(backgroundTranspiler, 60 * 5 * 1000);
  // This will trigger another run on completion. We need this here so that we attach the finally call to the promise returned each time its called
  // @ts-ignore
  trottledBackgroundTranspiler().finally(trottledBackgroundTranspiler);

  // initially kick it off
  trottledBackgroundTranspiler();
```

goetzrrGit:
We will address the UI hole, which should resolve most of the issues. However, since missions/users have visibility into the Hasura CLI, they can still see all available GQL and potentially bypass our recommended queries. This is a very unlikely situation, but users are going to do weird things especially if they are outside of Aerie.

goetzrrGit:
Create a new ticket to create a push model.

https://github.com/NASA-AMMOS/aerie/issues/1349

sequencing-server/src/app.ts
dyst5422: You have to have the recursive call inside the function that gets kicked off, so that your tweaked code won't work after the second invocation of the background transpolar. I'm surprised by the memory usage as I tested this with the following test code and had memory stabilize for me with <30MB heap and garbage collection seemed to be working.

Thoughts on why you might be seeing garbage collection not happening? What does the size of your cache look like over time?


```
{
  rss: '111.5 MB -> Resident Set Size - total memory allocated for the process execution',
  heapTotal: '54.97 MB -> total size of the allocated heap',
  heapUsed: '27.52 MB -> actual memory used during the execution',
  external: '2.27 MB -> V8 external memory'
}
```

```typescript
import throttle from 'lodash.throttle';

async function someAsync() {
	await new Promise(resolve => setTimeout(resolve, 1));
	console.log(formatMemoryUsage(process.memoryUsage()));
	return;
}

const throttledSomeAsync = throttle(someAsync, 10);

function recursiveCall() {
	throttledSomeAsync().finally(recursiveCall)
}

recursiveCall()

function formatMemoryNumber(mem: number){
	return `${Math.round(mem / 1024 / 1024 * 100) / 100} MB`;
}

function formatMemoryUsage(memoryUsage: NodeJS.MemoryUsage) {
	return {
	  rss: `${formatMemoryNumber(memoryUsage.rss)} -> Resident Set Size - total memory allocated for the process execution`,
	  heapTotal: `${formatMemoryNumber(memoryUsage.heapTotal)} -> total size of the allocated heap`,
	  heapUsed: `${formatMemoryNumber(memoryUsage.heapUsed)} -> actual memory used during the execution`,
	  external: `${formatMemoryNumber(memoryUsage.external)} -> V8 external memory`,
	};
}
```

dyst5422: AHA! I found it! Because arrow functions `const functionName = () => { //body }` capture their context for execution, it causes memory to explode. Change it to the declared function like I have above and give it a try. With my small example above, I was able to get to gigabytes of memory in seconds just by changing it to an arrow function.

So
```typescript
// The throttle wrap is lightly stateful, so we need to hold onto the same instance.
const trottledBackgroundTranspiler = throttle(backgroundTranspiler, 60 * 5 * 1000);
// This will trigger another run on completion. We need this here so that we attach the finally call to the promise returned each time its called
function runBackgroundTranspiler() {
  return trottledBackgroundTranspiler().finally(runBackgroundTranspiler);
}

// initially kick it off
runBackgrounTranspiler();
```

goetzrrGit: Hi @dyst5422, thank you for looking into this! The server is now stable after removing the arrow function and throttling at 10 milliseconds. I ran it for 10 minutes with periodic updates to the expansion logic, and the memory holds during my testing.

```
[node].  rss: '1251.54 MB -> Resident Set Size - total memory allocated for the process execution',
[node]   heapTotal: '80.97 MB -> total size of the allocated heap',
[node]   heapUsed: '44.73 MB -> actual memory used during the execution',
[node]   external: '3.68 MB -> V8 external memory'
```

sequencing-server/src/backgroundTranspiler.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  if (commandTypes === null) {
+    console.log(`[ Background Transpiler ] Unable to fetch command ts lib.
+    Aborting transpiling...`);
+    return;
+  }
+
+  // only process 'numberOfThreads' worth at a time ex. transpile 2 logics at a time
+  // This allows for expansion set and sequence expansion to utilize the remaining workers
+  for (let i = 0; i < expansion_rule.length; i += numberOfThreads) {
```
dyst5422, line 65: Batching like this shouldn't be necessary with the promise throttler as it will simply queue up runs to execute max numberOfTreads at a time. So you should be able to just queue them all up without waiting on batches, maximizing our use of threads.

sequencing-server/src/backgroundTranspiler.ts
```diff
@@ -0,0 +1,144 @@
+import { piscina, graphqlClient, typeCheckingCache, promiseThrottler } from './app.js';
```
dyst5422, line 1: General warning that singleton exports like this make testing hard (ie have to mock inject into the runtime etc). It's probably fine here, but an alternative would be to do dependency injection for these elements.

ie
```
function createBackgroundTranspiler(graphqlClient: <whatever>, promiseThrottler: PromiseThrottler<whatever>, ...etc) {
  return function backgroundTranspiler(numberOfThreads: number = 2) {
    // Body of function
  }
}
```

and then initialize it before using it in the context where you have all the dependencies for injection.

```
const backgroundTranspiler = createBackgroundTranspiler(graphqlClient, promiseThrottler, ...etc);
```

sequencing-server/src/backgroundTranspiler.ts
goetzrrGit: Previously, the background transpiler didn't have batching functionality and all tasks were queued up in the PromiseThrottler. I encountered a scenario where I was transpiling 120 expansion logic files, with 50-58 currently processed in sequential order. If a user wanted to create an expansion set with logic files 118-120, they would face an estimated 18-minute delay due to the queue. This delay occurred between the user clicking "save" and the page refreshing with the new expansion set, as the cache held a pending promise waiting to finish. To address this, I implemented batching and limited the background transpiler to 2 workers, allowing the rest of the server to utilize the remaining workers.

A thing to note, my laptop is old 7+ year mac so the 18-minute delay while trying to typechecked could be less on Clipper's servers.

sequencing-server/src/backgroundTranspiler.ts
goetzrrGit: Thank you for the suggestion, I noticed we are doing this singleton export throughout the server which we probably need to refactor. Let me file a ticket to address this. 

https://github.com/NASA-AMMOS/aerie/issues/1350

sequencing-server/src/backgroundTranspiler.ts
dyst5422: Are you saying that the promise was getting inserted into the cache prior to resolving, and then being read out prior to resolving elsewhere?

I worry that this is a legitimate race condition that needs to be addressed and it's unclear to me how batching would address that.

goetzrrGit: There are two parts where transpilation can occur: the background transpiler and expansion set creation. During expansion set creation, the promise generated by the PromiseThrottler is either cached or retrieved from the cache. If the promise isn't yet resolved, Promise.allSettled() guarantees that the PromiseThrottler processes and completes all cached promises.

expansion-set code
```ts
 const typecheckErrorPromises = await Promise.allSettled(
    expansions.map(async (expansion, index) => {
    
      const hash = ...
     
      if (typeCheckingCache.has(hash)) {
        return typeCheckingCache.get(hash);
      }

      const typeCheckResult = promiseThrottler.run(() => {
        return (
          piscina.run(
           ...
          ) as ReturnType<typeof typecheckExpansion>
        ).then(Result.fromJSON);
      });

      typeCheckingCache.set(hash, typeCheckResult);
      return typeCheckResult;
    }),
  );
```

For the above example when implementing batching. We have 120 expansion logic files to transpile, only two are fetched at a time. These logic are either cached or retrieved from the cache. The process then waits for both promises to finish before moving on to the next two. This approach avoids placing all 120 promises in the cache simultaneously and sequentially adds them 2 at a time. 


When creating an expansion set with logic files 118-120, the background transpiler has finished/cached  1-58 promises, and the next two 59-60 are pending. We will add 118-120 to the cache and promiseThrottler has 6 threads free to start processing 118-120. The only blocking `Promise.allSettled() ` is doing is for the 118-120 promises.

goetzrrGit: On the other hand, creating an expansion set with all 120 logic files would queue all promises at once and wait for their completion. Although, the background transpiler can still assist by picking up pending promises from the shared cache and processing them, improving the overall transpilation time.

dyst5422: I feel like there's a solution where we don't have to wait for all of the in process transpilations, but just the ones we need to use. Unfortunately, the promise throttler doesn't consider priority in it's current form, so we can't really "jump" an expansion that has been requested to the front of the queue.

An alternative solution only has the background transpiler pushing into the cache on completion/promise resolution, rather than pushing the unresolved promise (that could wait a long time for 100 other expansions to transpile). Thus the cache has either those expansions immediately requested as unresolved, or those that are background transpiled pushed in already resolved.

It's not a fully fleshed out solution, but perhaps it sparks some ideas for you @goetzrrGit.

dyst5422: These are two-way doors where we can do better in a future change, so I don't want our pursuit of an optimal solution to delay a sufficient solution for Clipper.

goetzrrGit: It does, I was thinking about this more last night. I considered changing the Throttler into a priority queue, where all promises generated by the backgroundTranspiler would be treated as lower priority than requests from the `expansion-set` or `expansion-to-sequences`. Additionally, I agree with your suggestion of only adding the completed promise to the cache instead of pending promises. I believe there is room for improvement in this initial implementation. 

Yeah, we are in a situation now where Clipper needs solution now so we have to punt more design and ideals down the line atm.

sequencing-server/src/app.ts
goetzrrGit: Doing further testing, I discovered that the HTTP GET requests to the server are not responding or returning data. I am going to have to revert back to the original code for the time being. Unfortunately, due to time constraints, I cannot investigate the cause of this issue as we're releasing a new version of Aerie tomorrow.

cohansen (approved):

goetzrrGit:
Going to revise this code when I work on https://github.com/NASA-AMMOS/aerie/issues/1349

PR #1323, opened 2024-02-06
In-memory checkpoint simulation

adrienmaillard:
* **Tickets addressed:** Closes #1189, Closes #1190 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

The first part of this branch has been done by @mattdailis, it 
- enables to duplicate simulation engines
- then designs a specific simulate procedure that is able to use the best cached engine to simulate a given plan

Then, I took this work and integrated it into the scheduling agent. Here are some elements of this work:
- Separation between simulation and the computation of simulation results. E.g. rootfinding does not need simulation results (used only in scheduling) 
- Partial computation of simulation results for only a subset of the resources (used only in scheduling)
- `SimulationFacade` has been completely overhauled to account for checkpoint simulation. It now takes care of updating the plan to simulate with child activities and activity durations.
- The scheduler worker now maintain a cache of engines with a configurable size.  

Other stuff:
- A little bit of cleaning in the scheduler (`DurationExpression` mainly)
- In the priority solver, goals and activity expressions now explicitely compute resource dependencies and we use the partial sim computation above + we mock the simulation results when no resource is need, that is, we compute fake simulation results from grounding the schedule into a `SimulationResults` object.
- The interval of the last profile segment of constraint simulation results become closed-closed to be consistent with what the final state of the simulation engine means. When the engine is finishing at time t, everything at this time has been performed, meaning that the resource value at this time has been computed. 

This is not visible in the tests because the metric for simulation performance in scheduling has changed but all of this is greatly decreasing the required simulation time (for the scheduling tests at least...).

## Verification

Few tests were added.

## Documentation

Needs some documentation, at least for the configuration of cache size.

## Future work

- Smarter checkpoint policy
- Incremental simulation

Issue #1096, opened 2023-08-17
Stream simulation results into the database


camargo:

Mythicaeda:
[Excalidraw link](https://excalidraw.com/#room=4d39daf1c3ae97077256,sTk9Nliof5KKeU5ffdES6w) from [design thread](https://jpl.slack.com/archives/CJ07FGRPH/p1712951947360269).

Mythicaeda:
Our goal with this was to improve peak memory usage for large mission models. Simulation run time is not expected to change much, but a [follow-up ticket to improve it has been opened](https://github.com/NASA-AMMOS/aerie/issues/1516). Scheduling runtime is not anticipated to change at all.

You can [check the PR](#1507) for an in-depth comparison of memory usage made using the Europa Clipper Model.

For a better sense of impact, however, I performed a memory comparison between `develop` and `streaming` using a variant of Endurance's mission model that did not include power modelling. Of note:

- **Heap Size**: Streaming peaked at a heap size of 5GB while writing Simulation Results but generally was at 3-4GB, while Develop peaked at 6.5 GB but was generally at about 5 GB during that time.
- **Non-Heap Memory**: Streaming had a very level non-heap memory usage while Develop had an increase in non-heap memory usage around when Simulation Results were being processed. Streaming peaked at 42MB but generally used 39MB, while Develop peaked at 45MB, but generally used 35MB (pre-minute 7)  to 42MB (post-minute 7).
- The peak number of objects allocated in Streaming was 12 million, while the peak number of objects allocated on develop was 14 million.
- Streaming took 14 minutes to complete while Develop took 12 minutes.

Below are three screenshot’s showing the comparison for 1) heap 2) non heap and 3) recorded objects. Develop is shown on the left and Streaming is shown on the right. Please be aware in (2) that the vertical scale differs for the two graphs.

<img width="1493" alt="heap comparison" src="https://github.com/user-attachments/assets/77a8152d-9c63-4ecd-a6f8-f0cfb1bf160e">
<img width="1509" alt="non-heap comparison" src="https://github.com/user-attachments/assets/f82f9e58-3b35-4c89-9371-6a5289921b82">
<img width="1509" alt="recorded object comparison" src="https://github.com/user-attachments/assets/a9370aa3-624d-4872-9f99-1c5f1ccea522">

PR #1334, opened 2024-02-15
add some perf logging in SimulationDriver

parkerabercrombie:
Cherry pick performance logging change by @martyvona, from https://github.com/NASA-AMMOS/aerie/pull/1269. This seems useful as we continue to work on optimizing simulation performance.

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -27,6 +27,8 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.TaskStatus;
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.lang3.tuple.Triple;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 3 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -50,6 +52,11 @@
  * A representation of the work remaining to do during a simulation, and its accumulated results.
  */
 public final class SimulationEngine implements AutoCloseable {
+
+  private static final Logger logger = LoggerFactory.getLogger(SimulationEngine.class);
```
JoelCourtney, line 56: This logger looks unused

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -50,6 +52,11 @@
...
+
+  private static final boolean PROFILE_RESOURCES = false;
```
JoelCourtney, line 58: It would be nice to be able to configure this with an env var like `MERLIN_WORKER_PROFILE_RESOURCES`

JoelCourtney:
Seems to work as expected 👍

martyvona:
Thanks @parkerabercrombie and @JoelCourtney, I'd be happy to resolve the PR review comments above, however, I think perhaps we should just close this PR without merging for the following reasons:
1. it looks like @DavidLegg has a [different branch](https://github.com/DavidLegg/aerie/tree/feature--streamline-resource-profiling) now with a more sophisticated implementation of resource profiling
2. it's just a small PR, not MVP, and we all have a lot going on right now

mattdailis:
Closing this PR based on @martyvona 's comment above. I'll frame the rationale a couple additional ways:
1. The Clipper mission's requirements for performance are (in theory) more easily met with recent simplifications to their approach to using Aerie
2. There are many efforts in flight that affect the simulation driver (#1096 , #1323 , #669), so the kinds of things that would be useful to log are in flux

PR #1340, opened 2024-02-17
acknowledge code inheritance from Blackbird

bradNASA:
A reference to Blackbird in the README acknowledges that there is some code inheritance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ The scripts and documentation in this project are released under the [MIT Licens
 
 ## Attribution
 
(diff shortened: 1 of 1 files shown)


Twisol:
(my comment on [NASA-AMMOS/aerie-docs#126](https://github.com/NASA-AMMOS/aerie-docs/pull/126#issuecomment-1950269670) applies here too; just commenting here to link the two issues together.)

bradNASA:
I got the history wrong--some miscommunication.   The update to aerie-docs accomplishes the intent here, and this can go away.

PR #1341, opened 2024-02-20
Update Java version to 21 in CI

skovati:


.github/workflows/create_jnispice.yml
Mythicaeda, line 1: We should recompile the JAR portion of this workflow, so that all of our artifacts are compiled on the same Java version.

Mythicaeda (approved):

.github/workflows/create_jnispice.yml
skovati: I've created a ticket to track this. https://github.com/NASA-AMMOS/aerie/issues/1344

PR #711, opened 2023-03-03
Constrain scheduling goals and specifications

mattdailis:
* **Tickets addressed:** Closes #471 
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This PR constrains scheduling goals to at most one scheduling specification, and constrains scheduling specifications to at most one plan.

This will need to be rebased onto #708 (or vice versa), to keep our migrations in linear order. 

## Verification

Using the API, I checked that it was no longer possible to associate a goal with multiple specifications. I created Goal 1 with the UI, and it was automatically associated with Specification 1. Here, I attempted to associate it with Specification 2:
```graphql
mutation AssociateSchedulingGoalWithNewSpec{
  insert_scheduling_specification_goals_one(object:{
    goal_id: 1,
    specification_id: 2
  }) {
    enabled
  }
}
```

Error message:
```json
{
  "errors": [
    {
      "extensions": {
        "code": "constraint-violation",
        "path": "$.selectionSet.insert_scheduling_specification_goals_one.args.object"
      },
      "message": "Uniqueness violation. duplicate key value violates unique constraint \"scheduling_specification_unique_goal_id\""
    }
  ]
}
```

Similarly, I tried to add a new specification for plan 1:

```graphql
mutation CreateSchedulingSpecification{
  insert_scheduling_specification_one(object:{
    plan_id:1,
    plan_revision: 1,
    horizon_start: "2023-03-03T00:00:00+00:00",
    horizon_end: "2023-03-17T00:00:00+00:00",
    simulation_arguments: {},
    analysis_only: false
  }) {
    id
  }
}
```

```json
{
  "errors": [
    {
      "extensions": {
        "code": "constraint-violation",
        "path": "$.selectionSet.insert_scheduling_specification_one.args.object"
      },
      "message": "Uniqueness violation. duplicate key value violates unique constraint \"scheduling_specification_unique_plan_id\""
    }
  ]
}
```

## Documentation

https://github.com/NASA-AMMOS/aerie-docs/pull/18

## Future work

If it turns out that this is a limitation we want to keep for the long term, we can consider normalizing our schema

PR #1355, opened 2024-03-04
Update Plan Metadata

Mythicaeda:
* **Tickets addressed:** none
* **Review:** By commit  

## Description

#711 changed the relationship of plan to scheduling spec from an array relationship to an object relationship but did not update the Hasura metadata to reflect this. This PR updates that metadata and renames the relationship from the plural `scheduling_specifications` to the singular `scheduling_specification`. This metadata change is minorly breaking for anyone using that relationship.

## Verification

The `GET_PLAN` query in the E2E tests was updated, but no other changes were necessary.

## Documentation

No docs need to be updated, as we never referred to this relationship in them.

### DIFF
--- a/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan.yaml
+++ b/deployment/hasura/metadata/databases/AerieMerlin/tables/public_plan.yaml
@@ -67,10 +67,10 @@ array_relationships:
         name: plan
         schema: public
 remote_relationships:
-- name: scheduling_specifications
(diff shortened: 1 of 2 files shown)


JoelCourtney (approved):

JoelCourtney:
Does the UI need to be updated because of the name change?

Mythicaeda:
Yeah, a mixture of the name change and the return type going from an array to an object, based on a discussion with @duranb.

PR #1359, opened 2024-03-11
Add Ability to Forcibly Resimulate

### LINKED ISSUE #1317: Add Optional "Force Resim" Parameter to Simulation Endpoint
If this optional parameter is set to true, the simulation endpoint will update the _Simulation Specification_'s revision between checking permissions and running the standard simulation logic. This update is to avoid having two requests, one completed and one that needs to be picked up, at the same revision set. The simulation specification is being chosen as we would want both runs to display on the same Plan Snapshot, and updating the plan revision would prevent that.

If this optional parameter is set to false or missing, the simulation endpoint will follow its current logic.

Additional work that may come up during this ticket: the "update on revision" trigger on the sim spec table ought to be fixed to update the row as it's updated, instead of sending a second update request afterwards. (This is also true of the plan, mission model, and simulation template) This change will guarantee that these revisions are actually monotonically increasing, as they currently can be set to any arbitrary value + 1.

Mythicaeda:
* **Tickets addressed:** Closes #1317
* **Review:** By commit  

## Description

Adds a new optional parameter to the `Simulate` action, `force`. If this parameter is present and set to `true`, the backend will ignore any cached sim results that may satisfy request and create a new request, updating the revision of the sim config in order to avoid there existing two requests for the exact same revision sets.

## Verification

New tests were added to `Bindings` and `SimulationTests` exercising this new parameter.

## Documentation

API Docs will need to be updated. PR to do so is open [here](https://github.com/NASA-AMMOS/aerie-docs/pull/134).

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/HasuraParsers.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/HasuraParsers.java
@@ -7,6 +7,7 @@ import gov.nasa.jpl.aerie.merlin.server.models.HasuraMissionModelEvent;
 
 import java.util.Optional;
 
+import static gov.nasa.jpl.aerie.json.BasicParsers.boolP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.listP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.longP;
 import static gov.nasa.jpl.aerie.json.BasicParsers.mapP;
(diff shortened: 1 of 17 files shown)


AaronPlave:
Working for me - I'm able to re-run sims from my UI branch https://github.com/NASA-AMMOS/aerie-ui/pull/1153

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/HasuraParsers.java
```diff
@@ -54,6 +55,17 @@ private static <I extends HasuraAction.Input> JsonParser<HasuraAction<I>> hasura
...
                           .map(HasuraAction.PlanInput::new, HasuraAction.PlanInput::planId));
 
+  public static final JsonParser<HasuraAction<HasuraAction.SimulateInput>> hasuraSimulateActionP
+      = hasuraActionF(
+          productP
+              .field("planId", planIdP)
+              .optionalField("force", nullableP(boolP))
+              .map(
+                  untuple((planId, force) -> new HasuraAction.SimulateInput(planId, force.flatMap($ -> $))),
+                  $ -> tuple($.planId(), Optional.of($.force()))
```
skovati, line 65: Non-blocker, but it might be clearer to just have this line be the `$.orElse(false)` so we can propagate a `Boolean` down to `getSimulationResults` instead of an `Optional<Boolean>`

skovati (approved):
LGTM!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/HasuraParsers.java
Mythicaeda: I've addressed this slightly differently. Rather than update the parser, I've instead updated when `force` is fetched from `body` in `MerlinBindings` to flatten it to a boolean as you suggested.

skovati: Looks great!

skovati (approved):
🚀

PR #1363, opened 2024-03-14
Procedural constraints library

### LINKED ISSUE #1302: Create a small constraints-only library
Create a small library that only constraints will use, which includes

```[tasklist]
### Tasks
- [ ] constraint interface
- [ ] violation type
```

JoelCourtney:
* **Tickets addressed:** closes #1302 
* **Review:** By commit  

## Description
This builds off the timeline library to provide a minimal interface for procedural constraints. It really only provides two things:
- a new timeline type for Violations, with a bunch of static methods for constructing Violations timelines from types in the timeline library
- an interface for the users' constraints to implement

This PR also moves the timeline library to a gradle subproject called `procedural` so that the libraries we provide to the users for procedural scheduling/constraints can live together. For reviewers / testers, this means the gradle commands and artifact locations are now nested in the procedural subproject.

## Verification
I actually don't see much opportunity for helpful unit tests, because the Violations timeline delegates and reuses things from the timeline library. There's very little interesting new code in here. 🤷 

## Documentation
See `procedural/README.md`.


procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/ActivityId.kt
```diff
@@ -0,0 +1,7 @@
+package gov.nasa.jpl.aerie.constraints
+
+/** An activity ID, referencing either a directive or instance. */
+sealed interface ActivityId {
+  /***/ data class InstanceId(/***/ val id: Long): ActivityId
```
skovati: Could this be a `typealias` instead? Would remove the need to `instanceId.id`

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/Violation.kt
```diff
@@ -0,0 +1,22 @@
...
+    override val interval: Interval,
+
+    /** List of associated activities (directives or instances) that are related to the violation. */
+    val ids: List<ActivityId> = listOf()
+) : IntervalLike<Violation> {
+
+  override fun withNewInterval(i: Interval) = Violation(i, ids)
+
+  /** Constructs a violation on the same interval with a different list of ids. */
+  fun withNewIds(vararg id: ActivityId) = Violation(interval, id.asList())
```
skovati: What is the use case for reconstructing a violation with new IDs?

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/Violations.kt
```diff
@@ -0,0 +1,85 @@
...
+    NonZeroDurationOps<Violation, Violations>,
+    CoalesceNoOp<Violation, Violations>
+{
+  constructor(vararg violation: Violation): this(violation.asList())
+  constructor(violations: List<Violation>): this(BaseTimeline(::Violations, preprocessList(violations, null)))
+
+  /**
+   * Maps the list of associated activity ids on each violation.
+   *
+   * @param f a function which takes a [Violation] and returns a new list of ids.
```
skovati: Continuation of my previous question, what would an example `f: Violation -> List<ActivityId>` be?

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/ActivityId.kt
JoelCourtney: Unfortunately we wouldn't be able to differentiate between instance and directive in Java. Typealiases don't exist in Java.

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/Violation.kt
JoelCourtney: Its possible to lose the relevant ids along the way, and end up with violations that are logically associated with activities, but don't actually contain the ids. Or maybe you want to violate on a child activity but associate it with the parent activity.

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/ActivityId.kt
```diff
@@ -0,0 +1,7 @@
+package gov.nasa.jpl.aerie.constraints
```
skovati: Do we want this to be `gov.nasa.jpl.aerie.procedural.constraints` (likewise for the timeline library)? Having the timeline library just be `jpl.aerie.timeline` is perfectly clear, but I can see it being a bit confusing having the same namespace as the original constraints library (`:constraints`).

Either way is fine with me, you can resolve this comment if you decide to stick with just `gov.nasa.jpl.aerie.constraints` and `gov.nasa.jpl.aerie.timeline`

skovati (approved):
Looks great to me! Really impressive how terse the constraints + violations definitions can be with the timeline lib mixins

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/Violations.kt
skovati: Keep it!

procedural/constraints/src/main/kotlin/gov/nasa/jpl/aerie/constraints/ActivityId.kt
JoelCourtney: made this change for the constraints package but not timeline package, like we talked about with Dan (i think)

Issue #1364, opened 2024-03-14
Add a mission model behavioral framework


adrienmaillard:
This is the second version of this ticket. This writeup is as much of a plan as it is a basis for my own reflection, sorry for the endless rambling. This is in coordination with the grouping work #1073 #1451 #870 as it also involves declarative modelling in the mission model.

We believe Aerie needs a **declarative** task framework to allow for better automated reasoning which will lead to a better performance and will also allow us to move from low-level goal specification (current goals based on batch insertion of activities) to high-level goal specification (e.g. state based).  

What’s not declarative right now ? 
the behavior of a task is pretty much entirely contained in its `run()` method. There’s no way of knowing what an activity is doing without simulating it. And simulating it in one set of initial conditions is not sufficient to know all of its potential behaviors, as conditional statements are allowed and may change the behavior of the task. Let's qualify the current task specification framework as **procedural**. The behavior of a task is not explicitly modeled, it’s the result of the execution of a piece of code. 

As I see it, there are two sides to this work, the aspects related to changes in the mission model and the aspects related to the uses of this new knowledge about tasks in the scheduler (and planner in the future hopefully).

### What do we need to express in this framework ?
The behavior of a task, what it does. 
- What resources are read, why are they read (in a conditional statement ?)
- What resources are written, how much there are affected (is it increased ? decreased ? by how much ?) 
- Tasks can be durative so we can also describe when things happen. 

And we need to record the relationships/dependencies between _resources_ and _tasks_.

In a traditional/academic planning and scheduling language, this is straightforward because there is a small number of well defined ways of expressing task behaviors and resource types. And these languages have been slowly improved/enriched year after year, their creators having in mind the consequences of adding each new block to the language. In our case, we start from the full expressivity of java and the endless imagination of mission modelers. Let's say right away that we won't be able to represent the current expressivity of mission models, as we would not be able to perform tractable reasoning over all the java constructs and so on and so on. 

Let's talk first about resources. I looked at our toy models and David Legg also helped me look at what Clipper does with resources. What we found is that:
- some resources are registered because the user is interested in seeing them in the UI
- but other resources are usually complex in that they form dependency chains with _anonymous_ resources in the middle that are often mappings/transformations of the input into another output (one resource X is a sample version of Y which is the Z multiplied by 2 which is A divided by 3...) which implies many resources are not registered, _they may not have a name that can be referenced from the activity model_

[...]

PR #1365, opened 2024-03-15
Add @MaximumDuration annotation and use it in scheduling

### LINKED ISSUE #1172: Reduce temporal domain of CoexistenceGoal by adding a @MaximumDuration annotation to models
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

The coexistence goal produces conflicts with a temporal context in the form of a `Windows` object. In absence of `applyWhen`, the temporal context is set to the planning horizon. 

This context is used to determine the domain of global scheduling condition checking (that is performed for EACH activity insertion). 

Simulation ends up being ran for the whole horizon because we do not have a way of computing an upper bound that is not the planning horizon. 

Here below is what we could do to remedy this:
if we have `end` timing constraint on the goal, we are good, we can take the upper bound of the interval. 
if we have `start` and `duration` timing constraint on the goal, we can add upper bounds of both.
if we have `start` constraints and `FixedDuration` or `ControllableDuration` duration types, we are also good. 
if we have `start` and `UncontrollableDuration` duration type, we **cannot** compute an upper bound. 

Thus, the need for a @MaximumDuration annotation. This duration will also be used to reduce domains of rootfinding/static temporal constraint checking.  This duration could be a very pessimistic upper bounds (10hours for a 1 minute activity) as it would already be much much better than the entire planning horizon. 

This ticket is about : 
- exploiting the existing duration types to provide such bounds 
- adding a @MaximumDuration annotation and exploit it

adrienmaillard:
* **Tickets addressed:** Resolves #1172 
* **Review:** By commit  

## Description

An optional `@MaximumDuration` mission model annotation has been added. This annotation can be used in conjunction of @ParametricDuration and @UncontrollableDuration to signify the maximum duration that an activity 

## Verification

I have verified that the domain of coexistence goal was actually cut. I have added a test with a coexistence goal that inserts a lot of coexisting activities and noticed a difference between the use of the maximum duration bounding and no use. 
The gain with this test is not significant yet (15-20%) because the simulation domain is dominated by simulation to get child activities. With changes proposed in #1364 related to dependency analysis, we would be able to bypass this step if new activities do no produce child activities. Then this change will have a significant effect on the runtime of`CoexistenceGoal` and rootfinding. Indeed, with future user option (in #1364) to bypass rootfinding when placing activities, knowing the maximum duration of an activity allows to place it without violating temporal constraints. 

## Documentation

Code documentation has been added for the annotation. User documentation has to be done. 

## Future work

Changes in #1364.

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
@@ -550,6 +550,7 @@ import java.util.stream.Collectors;
   {
     Optional<String> fixedDuration = Optional.empty();
     Optional<String> parameterizedDuration = Optional.empty();
+    Optional<String> maximumDuration = Optional.empty();
     for (final var element: activityTypeElement.getEnclosedElements()) {
       if (element.getAnnotation(ActivityType.FixedDuration.class) != null) {
         if (fixedDuration.isPresent()) throw new InvalidMissionModelException(
@@ -581,6 +582,22 @@ import java.util.stream.Collectors;
         );
 
         parameterizedDuration = Optional.of(executableElement.getSimpleName().toString());
+      } else if (element.getAnnotation(ActivityType.MaximumDuration.class) != null){
+        if (maximumDuration.isPresent()) throw new InvalidMissionModelException(
+            "MaximumDuration annotation cannot be applied multiple times in one activity type."
+        );
+
+        if (element.getKind() == ElementKind.METHOD) {
+          if (!(element instanceof ExecutableElement executableElement)) throw new InvalidMissionModelException(
+              "MaximumDuration method annotation must be an executable element.");
+
+          if (!executableElement.getParameters().isEmpty()) throw new InvalidMissionModelException(
+              "MaximumDuration annotation must be applied to a method with no arguments."
+          );
+          maximumDuration = Optional.of(executableElement.getSimpleName().toString() + "()");
+        } else if (element.getKind() == ElementKind.FIELD) {
+          maximumDuration = Optional.of(element.getSimpleName().toString());
+        }
       }
     }
 
@@ -609,7 +626,7 @@ import java.util.stream.Collectors;
           ? Optional.<TypeMirror>empty()
           : Optional.of(returnType);
 
-      return Optional.of(new EffectModelRecord(element.getSimpleName().toString(), executorAnnotation.value(), nonVoidReturnType, durationParameter, fixedDuration, parameterizedDuration));
+      return Optional.of(new EffectModelRecord(element.getSimpleName().toString(), executorAnnotation.value(), nonVoidReturnType, durationParameter, fixedDuration, parameterizedDuration, maximumDuration));
     }
 
     return Optional.empty();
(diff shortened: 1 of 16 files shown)


merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -581,6 +582,22 @@ private Optional<EffectModelRecord> getActivityEffectModel(final TypeElement act
         );
 
         parameterizedDuration = Optional.of(executableElement.getSimpleName().toString());
+      } else  if (element.getAnnotation(ActivityType.MaximumDuration.class) != null){
```
Mythicaeda: Minor, nonblocking: extra space
```suggestion
      } else if (element.getAnnotation(ActivityType.MaximumDuration.class) != null){
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -947,52 +946,16 @@ private Optional<SchedulingActivityDirective> instantiateActivity(
...
-      logger.debug(
-          "Duration range [" +
-          (durRequirementLower.isPresent() ? durRequirementLower.get() : "-inf") +
-          ", " +
-          (durRequirementUpper.isPresent() ? durRequirementUpper.get() : "+inf"));
-      logger.debug("Interval range " + interval);
-      return Optional.empty();
-    }
-    final var solved = taskNetwork.getAllData(name);
+    final var enveloppes = new ArrayList<Interval>();
```
Mythicaeda: Minor: typo:
```suggestion
    final var envelopes = new ArrayList<Interval>();
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CoexistenceGoal.java
```diff
@@ -266,11 +277,19 @@ else if (this.initiallyEvaluatedTemporalContext == null) {
       if (!alreadyOneActivityAssociated) {
         //create conflict if no matching target activity found
         if (existingActs.isEmpty()) {
+          var temporalContext = this.temporalContext.evaluate(simulationResults, evaluationEnvironment);
+          final var activityCreationTemplateBuilt = activityCreationTemplate.build();
+          final var evaluationEnvironement = createEvaluationEnvironmentFromAnchor(evaluationEnvironment, window);
```
Mythicaeda: Minor: typo: 
```suggestion
          final var evaluationEnvironment = createEvaluationEnvironmentFromAnchor(evaluationEnvironment, window);
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/CoexistenceGoal.java
```diff
@@ -266,11 +277,19 @@ else if (this.initiallyEvaluatedTemporalContext == null) {
...
+          final var newTemporalContext = boundTemporalContextWithSchedulerModel(
```
Mythicaeda, line 283: Why is CoexistenceGoal the only goal type that received the additional bounding when making a `MissingActivityTemplateConflict`?

adrienmaillard: Because only in coexistence goal there is such a precise notion of where the activity must be located. Cardinality and Recurrence are vague over each of their intervals.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: you can find this typo in many of my PRs :D

Mythicaeda (approved):

PR #1366, opened 2024-03-16
Fix Background Transpiler `Latest` Check

Mythicaeda:
* **Review:** By commit  

## Description

Fixed how the background transpiler is checking if there are any mission models or command dictionaries uploaded. Before, it would fail to notice that no models/cdicts had been uploaded, as `mission_model_aggregate` / `command_dictionary_aggregate` is never `null`.

## Verification

Spun up a fresh aerie instance and saw the correct logs printed. On develop, an exception would instead be printed.

## Documentation

No docs updates needed.

### DIFF
--- a/sequencing-server/src/backgroundTranspiler.ts
+++ b/sequencing-server/src/backgroundTranspiler.ts
@@ -15,8 +15,8 @@ export async function backgroundTranspiler(numberOfThreads: number = 2) {
   }
 
   // Fetch latest mission model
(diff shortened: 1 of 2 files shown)


goetzrrGit (approved):
I noticed a tiny bug when working on the query refactor branch. I pushed up a commit and It should now pull the right stuff. :P

Your changes look good and worked on my local computer.

PR #1367, opened 2024-03-18
Add @AllChildren annotation

adrienmaillard:
* **Tickets addressed:** Part of #1364 but does not resolve it
* **Review:** By commit  

## Description

Add a `@AllChildren` annotation to mission model so mission modelers can report, for a given activity, what other activity is called/spawned. This information will be used by the scheduler to evaluate whether it is useful to resimulate or not before goal evaluation: let's have G goal being satisfied by an activity of type A, G is the next goal to be evaluated. If an activity of type B was inserted at the last iteration, has B any chance of generating A and thus of satisfying the goal G ? If yes, simulation is required, otherwise it's not (for that only purpose, there might be other things that warrant a simulation, e.g. the goal needs to also evaluate an expression...).

Some edge cases:
- If the user does not use the annotation for an activity, we make the worst case assumption that the activity might generate all activities (itself included). 
- Activities without effect model cannot generate children 

This PR does not contain changed to the scheduler to use this information. These changes would overlap too much  with changes under review in #1323 

## Verification

A test with the banananation model has been added to verify that the generated code is accurate.

## Documentation

To be done.

## Future work

Other things in #1364 + use the annotation in the scheduler.

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
@@ -46,6 +46,7 @@ import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic;
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -265,7 +266,7 @@ public record MissionModelGenerator(Elements elementUtils, Types typeUtils, Mess
             .classBuilder(typeName)
             .addAnnotation(
                 AnnotationSpec
-                    .builder(javax.annotation.processing.Generated.class)
+                    .builder(Generated.class)
                     .addMember("value", "$S", MissionModelProcessor.class.getCanonicalName())
                     .build())
             .addModifiers(Modifier.PUBLIC, Modifier.FINAL)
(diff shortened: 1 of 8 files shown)


merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -336,6 +354,26 @@ public JavaFile generateSchedulerModel(final MissionModelRecord missionModel) {
...
+  private CodeBlock.Builder getChildrenStatement(final MissionModelRecord missionModel, final ActivityTypeRecord activityTypeRecord){
+    String[] children = null;
+    //if the effect model is empty, an activity cannot have children
+    if(activityTypeRecord.effectModel().isEmpty()) children = new String[]{};
+    else if(activityTypeRecord.effectModel().get().children().isPresent()){
+      //if children have been declared with an annotation, record those
+      children = activityTypeRecord.effectModel().get().children().get();
+    } else {
+      //if children have not been declared with an annotation, assume the activity might generate all activity types
+      children = missionModel.activityTypes().stream().map(ActivityTypeRecord::name).toArray(String[]::new);
```
Mythicaeda, line 366: Memory-wise, rather than returning a list of every activity in the plan, what about wrapping the return in `Optional` and returning `Optional.empty`? It would communicate the same idea-- that `getChildren` is returning the _known_ children and that we don't know what the children are-- without needing a map of 300+ elements with each containing a list of 300+ elements (ie, in the Clipper case).

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/SchedulerModel.java
```diff
@@ -4,10 +4,12 @@
...
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 
+import java.util.List;
 import java.util.Map;
 
 public interface SchedulerModel {
   Map<String, DurationType> getDurationTypes();
   SerializedValue serializeDuration(final Duration duration);
   Duration deserializeDuration(final SerializedValue serializedValue);
+  Map<String, List<String>> getChildren();
```
Mythicaeda, line 14: Minor: If the data is provided in an array, what is gained by converting it into a List? Are there some List-only methods you expect the scheduler to use? Was it simpler to return a List in the code generator?

Mythicaeda, line 14: I believe this change makes this PR `breaking` for existing models.

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/generator/MissionModelGenerator.java
```diff
@@ -327,6 +328,23 @@ public JavaFile generateSchedulerModel(final MissionModelRecord missionModel) {
...
+                           .addCode(
+                               missionModel
+                                   .activityTypes()
+                                   .stream()
+                                   .map(
+                                       activityTypeRecord ->
+                                           getChildrenStatement(missionModel, activityTypeRecord))
+                                   .reduce((x, y) -> x.add("$L", y.build()))
+                                   .orElse(CodeBlock.builder()).build())
+                           .addStatement("return result")
```
Mythicaeda, line 346: Minor, nonblocking: this method could return an immutable map if it uses the `Map.ofEntries()` method. 
Then instead of each `getChildrenStatement` returning `result.put("TYPE", LIST)`, it would return `Map.entry("TYPE", LIST)`.

adrienmaillard:
This is blocked because part of an ongoing prototyping for dependency modeling in #1364.

adrienmaillard:
@dandelany I think we can close this PR.

Mythicaeda:
Closing as per Adrien's note

PR #1369, opened 2024-03-19
Enable continuous validation thread by default

### LINKED ISSUE #1333: Enable automatic simulation validation thread by default
`aerie-merlin`'s `ENABLE_CONTINOUS_VALIDATION_THREAD` env var originally defaulted to false as a performance consideration, but now that the corresponding UI has been merged (and heavily depends on this flag being set), the feature should be enabled by default to avoid confusion.

skovati:
* **Tickets addressed:** closes #1333 
* **Review:** By commit  

## Description

s/false/true

Additionally, I removed this env var from the `docker-compose.yml` files, to avoid cluttering these files with an env var that will very likely never change from the default.

# Future work
Could be nice to clean up / consolidate all non-default env vars we set into a single `.env` that [`docker compose` could source](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute)

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -35,7 +35,6 @@ services:
       MERLIN_DB_USER: "${AERIE_USERNAME}"
       MERLIN_LOCAL_STORE: /usr/src/app/merlin_file_store
       MERLIN_PORT: 27183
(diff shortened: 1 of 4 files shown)


mattdailis (approved):
Good idea. We've had a few new users be confused by this

PR #1371, opened 2024-03-21
Performance/intervalmap

bradNASA:
* **Tickets addressed:** none, afaik
* **Review:** By file ?

## Description
IntervalMap changed to use TreeSet instead of ArrayList for Segments to speed up set(), which is used to construct an IntervalMap from a list of Segments.  Some SPHEREx scheduling code for merging SimulationResults is very slow because set() makes IntervalMap construction O(n^2).  The use of TreeSet makes that O(n log n).  I am not sure what other code may benefit from this.

Joel Courtney pointed out that the use of a list is more efficient in some cases, for example, for appending non-overlapping Segments.  He thought it might make sense to keep the list and use a tree in some cases.  There may be some better data structure that takes advantage of both, like an ordered list with binary search operations; maybe a skip list would, too.

There's a shortcut checking whether the input segments meet invariant conditions in order to ingest Segments without using set(), which handles overlapping.

There are [changes to IntervalAlgebra](https://github.com/NASA-AMMOS/aerie/pull/1371/files#diff-2265dc5cd805d49c34a2cb478c43f8ddae1127272d5f78043cc34032f7933b12R161) for faster low-level interval queries.

## Verification
Unit tests pass seemingly faster (maybe 3.25 instead of 3.75 minutess), but variance is high.  I still need to verify it solves the problem in the SPHEREx model.

## Documentation
None, afaik

## Future work
Interval code elsewhere in Aerie may also need updating like this.

### DIFF
--- a/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalAlgebra.java
+++ b/constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalAlgebra.java
@@ -158,11 +158,8 @@ public class IntervalAlgebra {
    * @return whether the operands overlap
    */
   static boolean overlaps(Interval x, Interval y) {
-    // First try for a fast shortcut that doesn't require allocating a new interval.
-    if (!x.isEmpty() && !y.isEmpty()) {
-        return !endBeforeStart(x, y) && !endBeforeStart(y, x);
-    }
-    return !isEmpty(intersect(x, y));
+    if (x.isEmpty() || y.isEmpty()) return false;
+    return !endBeforeStart(x, y) && !endBeforeStart(y, x);
   }
 
   /**
@@ -173,14 +170,8 @@ public class IntervalAlgebra {
    * @return whether `outer` contains every point in `inner`
    */
   static boolean contains(Interval outer, Interval inner) {
-    // First try for a fast shortcut that doesn't require allocating a new interval.
-    if (!outer.isEmpty() && !inner.isEmpty()) {
-      return !startBeforeStart(inner, outer) && !endBeforeEnd(outer, inner);
-    }
-
-    // If `inner` doesn't overlap with the complement of `outer`,
-    // then `inner` must exist entirely within `outer`.
-    return !(overlaps(inner, strictUpperBoundsOf(outer)) || overlaps(inner, strictLowerBoundsOf(outer)));
+    if (outer.isEmpty() || inner.isEmpty()) return false;
+    return !startBeforeStart(inner, outer) && !endBeforeEnd(outer, inner);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/SerializedValue.java
```diff
@@ -73,6 +116,11 @@ public <T> T match(final Visitor<T> visitor) {
       return visitor.onNumeric(value);
     }
 
+    @Override
```
bradNASA, line 96: These getValue() implementations could be private, but it seems like a convenient method to have.  I suspect there's a good reason why SerializedValue is not SerializedValue\<T\>.  I guess it might be a breaking change at this point.

JoelCourtney: Yeah, SerializedValue is not generic because it is intended to be basically a JSON object, which is type checked at runtime. Its always used in a context that is polymorphic over all types, so if we parameterized it, it would just be used as `SerializedValue<Object>` all the time.

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/time/IntervalAlgebra.java
```diff
@@ -224,6 +294,11 @@ static boolean endsStrictlyBefore(Interval x, Interval y) {
    * @return whether x ends when y begins, with no overlap and no gap
    */
   static boolean meets(Interval x, Interval y) {
+    // First try for a fast shortcut that doesn't require allocating a new interval.
+    if (!x.isEmpty() && !y.isEmpty()) {
+      return x.end.isEqualTo(y.start) && (x.endInclusivity != y.startInclusivity);
+    }
+
```
JoelCourtney: I think rather than being a "shortcut", all of these are just faster ways of representing the same logic. Assuming these changes are all sound, anything that is not caught by the "shortcut" should return `false`, because one of the intervals is empty.

bradNASA: Okay, for some reason I thought the existing implementation was being smart about empty and negative intervals.

bradNASA: I pushed a commit that uses the simple versions and removes the previous implementation.  They pass the tests.  That doesn't necessarily mean that the logic is the same as before.

JoelCourtney: Yeah, not really. I think we were just focusing more on readability and delegation than performance.

Also there are no negative intervals; negative intervals are considered empty.

bradNASA:
@mattdailis Any reason we shouldn't merge this?

JoelCourtney (approved):
I haven't looked at every change to IntervalMap in detail, but I do know that Pranav and I tested the bejeezus out of that class back in the day. It sounds theoretically sound and the tests pass. 👍

bradNASA:
> I haven't looked at every change to IntervalMap in detail, but I do know that Pranav and I tested the bejeezus out of that class back in the day. It sounds theoretically sound and the tests pass. 👍

Yeah, I was impressed with the detail of the test suite.

dandelany:
Thanks @bradNASA , merging!

PR #1385, opened 2024-04-03
Return the Executing Directive Id in Simulation Error Response

### LINKED ISSUE #659: Include more information with simulation error messages

Mythicaeda:
* **Tickets addressed:** Closes #659 
* **Review:** By commit  

## Description

If an activity directive throws an exception while executing, the offending directive id will now be included in the `data` part of the `reason` under the `executingDirectiveId` key.

This was accomplished by wrapping exceptions raised while stepping tasks in a `SpanException`. When this exception is caught, the driver will attempt to get the `directiveId` from that span before rewrapping the exception into a `SimulationException`, providing this id if it was found. Additionally, when a SimulationException is returned as the `reason` for a SimFailure, the `message` field is now populated with the exceptions message;

## Verification

The `SimulationExceptionResponse` tests were added.

In order to support these tests, the foo mission model gained a simulation configuration parameter and a new activity type. 
- The `raiseException` parameter will trigger a daemon task that throws a runtime exception 1hr in a plan if enabled
- The `DaemonCheckerSpawner` will call a `DaemonCheckerActivity` after `spawnDelay` minutes have passed, allowing us to throw runtime exceptions on demand/see the interaction between child activities and daemon tasks.

## Documentation

No docs need to be updated.

## Future work

- Posting the failed sim results to the DB
- Returning more information about the executing directive (ie name, the lineage from the source directive if it was a child that threw)
- If a daemon task throws an exception, is there anything we can return?

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -80,7 +80,10 @@ public final class SimulationDriver {
         {
           final var batch = engine.extractNextJobs(Duration.MAX_VALUE);
           final var commit = engine.performJobs(batch.jobs(), cells, elapsedTime, Duration.MAX_VALUE);
-          timeline.add(commit);
+          timeline.add(commit.getLeft());
+          if(commit.getRight().isPresent()) {
+            throw commit.getRight().get();
+          }
         }
 
         // Get all activities as close as possible to absolute time
@@ -88,7 +91,7 @@ public final class SimulationDriver {
         // Using HashMap explicitly because it allows `null` as a key.
         // `null` key means that an activity is not waiting on another activity to finish to know its start time
         HashMap<ActivityDirectiveId, List<Pair<ActivityDirectiveId, Duration>>> resolved = new StartOffsetReducer(planDuration, schedule).compute();
-        if(!resolved.isEmpty()) {
+        if (!resolved.isEmpty()) {
           resolved.put(
               null,
               StartOffsetReducer.adjustStartOffset(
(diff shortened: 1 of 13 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -161,8 +177,11 @@ void simulateTask(final MissionModel<Model> missionModel, final TaskFactory<Retu
       engine.scheduleTask(Duration.ZERO, missionModel.getDaemon());
       {
         final var batch = engine.extractNextJobs(Duration.MAX_VALUE);
-        final var commit = engine.performJobs(batch.jobs(), cells, elapsedTime, Duration.MAX_VALUE);
-        timeline.add(commit);
+          final var commit = engine.performJobs(batch.jobs(), cells, elapsedTime, Duration.MAX_VALUE);
+          timeline.add(commit.getLeft());
+          if(commit.getRight().isPresent()) {
+             throw new Error("Exception thrown while starting daemon tasks", commit.getRight().get());
```
mattdailis: I think `Error` is a little strong here - perhaps `RuntimeException`? Generally, we use `Error` to indicate a bug in our own code, kind of a "this shouldn't happen". If it can be triggered by user actions, let's use a subclass of Exception instead of Error, which is a Throwable but not an Exception.

Mythicaeda: I've changed them to `RuntimeException`s 👍

mattdailis (approved):
🎉

PR #1396, opened 2024-04-09
Allow Users to write to Model Scheduling Spec's `priority` Column

Mythicaeda:
* **Review:** By commit  

## Description

Users need to be able to set and update the priority of scheduling goals on the mission model spec, but this column was initially excluded from the insert/update set.

## Verification

Verified via the UI's upsert query for model specs.

## Documentation

No docs update are needed.


skovati (approved):
🚀

PR #1401, opened 2024-04-11
Fix Id Sequences on Constraint, Goals, and Conditions

Mythicaeda:
* **Review:** By commit  

## Description

During the migrations that split constraints, goals, and conditions into metadata and definitions, we preserved the IDs. This meant, however, that the ID sequence used to generate new IDs were not advanced, causing issues with inserting new rows in these tables initially. 

This PR creates a migration that advances the ID sequence on these tables to where it should be (the current max id + 1, or 1 if the table is empty). The `down` migrations are empty aside from marking the rollback as there is nothing to revert.

## Verification

I first tested the migration against an empty database. 

I then:
1) Spun up a 2.4.0 Aerie instance (the last one before versioning constraints)
2) Added 5 Constraints, Goals, and Conditions
3) Migrated to this branch
4) Created a new Constraint, Goal, and Condition

## Documentation

No docs need to be updated.

### DIFF
--- a/deployment/hasura/migrations/AerieMerlin/39_update_constraints_sequence/down.sql
+++ b/deployment/hasura/migrations/AerieMerlin/39_update_constraints_sequence/down.sql
@@ -1,0 +1,1 @@
+call migrations.mark_migration_rolled_back('39');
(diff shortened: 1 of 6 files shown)


mattdailis (approved):
Looks good to me. Do we feel pretty good that we don't need to do this for other tables?

Mythicaeda:
Yeah. I went and checked the rest of the repo and the only other time we ever used the phrase "generated by default" (including within migrations) is on `activity_directive` and `merge_staging_area`. I'm not thrilled about the usage in `merge_staging_area` and imagine that that was just a result of copying the `activity_directive` table definition and tweaking it, but felt fixing it was out of scope for this PR.

PR #1405, opened 2024-04-16
flags for turning naming on/off and specifically for emits

bradNASA:
* **Tickets addressed:** none
* **Review:** By commit

## Description
Brad Clement's SPHEREx model was unusually slow (~100X slower) because of string operations in creating names for emits on MutableResources.  These changes add a flag that by default turns off this naming.  It also adds a flag for turning
all of Naming on/off (default on).  Naming is part of the streamline resource library that helps with debugging, profiling, and
maybe error messages, too.

## Verification
The SPHEREx model's performance is greatly improved.  Unit tests pass.

## Documentation
descriptions of things in comments -- help me find where it may be missing, pretty please.

## Future work
Clipper may want to include the call to turn emit naming on in their mission model.  I recommend testing the performance of this on Clipper to see if this makes a difference.


bradNASA:
#1486 makes this PR unnecessary.

PR #1407, opened 2024-04-17
Add Missing `merlin` Prefix

Mythicaeda:
* **Review:** By commit  

## Description

Adds a missing `merlin` prefix in the force-resim workflow.

Additionally, updates the default provided service usernames in the `deployment/.env` to not have them match names of services.

This slipped by our tests initially because we had the usernames for services set to the name of the service schemas. When a user connects to the DB, [their default search path is `$username, public`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH). Since this bug was in the merlin-server, the service had a default search path of `merlin, public`, which allowed it to successfully find and update the `merlin.simulation` table.

## Verification

Manual verification, as the E2E tests were okay with this missing prefix.

CI Tests will now be effective thanks to @dandelany's tweaking of the `e2eTest` environment.

The DB Test Helper now grabs the Aerie and Postgres credentials from the user's environment/`.env` instead of assuming them. Note that currently the helper will prefer to use the `.env` in the `e2eTest` directory over the one at the root of the repo.


dandelany (approved):
👍 thanks!

PR #1410, opened 2024-04-17
Add "order by" when populating a plan spec

Mythicaeda:
* **Review:** By commit  

## Description

@duranb was running into an issue where plans were failing to have their scheduling specs created and populated due to priority gaps. While I was unable to repro, I found this was caused by a missing `order by` in the function to create and populate the plan spec that enforces that the spec is populated in priority order.

This PR adds that `order by` statement in the base code and in the migration.

## Verification

Manually verified on @duranb's deployment

This PR will fail the DB Migration check as 2.8.0 has yet to be released.


dandelany (approved):
👍 thanks for catching this! I'll go ahead and merge.

PR #1412, opened 2024-04-18
Update to Postgres 16.2

### LINKED ISSUE #1357: Update Postgres Version
We have been stuck on version 14 of Postgres due to creation restrictions in the `public` schema. Once #1311 is completed, however, that restriction will no longer apply and we should upgrade Postgres to the latest version (as of writing this ticket 16).

This ticket is being written as a follow-up to #1311 rather than as part of it to avoid changing both the DB structure and version at the same time.

Mythicaeda:
* **Tickets addressed:** Closes #1357
* **Review:** By commit  

## Description

Updates our base Postgres image from 14.8 to 16.2.

## Verification

Started a fresh DB to ensure there were no issues.

Spun up a DB on 14.8 and following the `pg_dumpall` -> `psql` upgrade procedure from the [Postgres upgrade docs](https://www.postgresql.org/docs/current/upgrading.html) to verify there were no issues upgrading. 

## Documentation

I don't believe any docs were invalidated.

### DIFF
--- a/.github/workflows/pgcmp.yml
+++ b/.github/workflows/pgcmp.yml
@@ -273,7 +273,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install --yes postgresql-client
       - name: Start Postgres
-        run: docker run -d -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust --name=postgres postgres:14.8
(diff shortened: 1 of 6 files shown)


skovati (approved):
`aerie-dev` dump and restore on pg 16.2 worked flawlessly!

PR #1418, opened 2024-04-24
Add Hasura Events Views

### LINKED ISSUE #1273: Show error if mission model upload fails
### Checked for duplicates

No - I haven't checked

### Alternatives considered

### Related problems

_No response_

Aerie should show an error to the user if an error occurs during mission model upload. We ran into a situation where an error occured when Aerie processed the mission model jar, but no error appeared to the user. The mission model appeared in the list of models, you could create plans with it, but there were no activity types in the model.

We found this error in the hasura trigger logs:
```
"message": "{\"message\":\"gov.nasa.jpl.aerie.merlin.driver.MissionModelLoader$MissionModelLoadException: No implementation found for `MerlinPlugin` at path `/usr/src/app/merlin_file_store/eurc-1704400986577-neM51CdWwMPdBN.jar` wih name \\\"eurc\\\" and version \\\"main-bfad1f70f3\\\"\"}"
```

Mythicaeda:
* **Tickets addressed:** Closes #1273 (backend portion)
* **Review:** By commit  

## Description

Creates a SQL function to get the event logs for a specified Hasura event, and three views that select from this function, one for each of our Hasura events.

While the views could not be necessary had I actually created and tracked a custom table for the function return type in order to [satisfy the Hasura Custom Function API ](https://hasura.io/docs/latest/schema/postgres/custom-functions/#pg-supported-sql-functions)(as is done with our other Hasura functions), by using views we can establish relationships between the results and the `mission_model` table, allowing for queries like so:

```gql
query {
  mission_model{
    id
    refresh_activity_type_logs { success }
    refresh_resource_type_logs { success }
    refresh_model_parameter_logs { success }
  }
}
```

Note that using a function for the view definitions is required, as the Hasura tables the views reference _**do not exist**_ during database initialization (Hasura creates and manages them). View definitions must be executable at the time of creation, but functions do not have this restriction.

This PR also fixes some minor bugs in our Migrations pipeline. Specifically:
- Fixed the PGCMP Workflow. Checking out `develop` requires re-cloning the `PGCMP` repo. The workflow was doing this pre-2.8.0, but I accidentally removed it when updating it for 2.8.0.
- Fixed a bug in the migration script that was causing hasura to reapply migration 1 when bulk migrating down from version 1. This bug was introduced in 2.8.0. 
- Fixed a typo in the migration script preventing the "mark correct version" command from applying. The `hasura` command is case-sensitive. This bug was introduced in 2.8.0.

## Verification

Manual verification was done to figure out the schema for errors and to check the Hasura metadata.

An E2E test was added to the `MissionModelTests` suite for testing the ability to get these logs. Additionally, the model upload process now awaits these logs, replacing the `Thread.sleep` in `HasuraFunctions.uploadMissionModel`

While checking if any tests were delaying for these events to complete, I noticed that `SchedulingTests.BeforeEach` was inconsistent in how it uploaded the model compared to other tests classes (it caught and rethrew the potential `InterruptedException` as a `RuntimeException`). This has been resolved.

## Documentation

Docs ticket here: https://github.com/NASA-AMMOS/aerie-docs/issues/147

## Future work

A front-end ticket is needed to track displaying success/failures for model uploads.

### DIFF
--- a/deployment/postgres-init-db/sql/views/hasura/hasura_event_logs.sql
+++ b/deployment/postgres-init-db/sql/views/hasura/hasura_event_logs.sql
@@ -1,0 +1,60 @@
+create function hasura.get_event_logs(_trigger_name text)
+returns table (
+  model_id int,
+  model_name text,
+  model_version text,
+  triggering_user text,
+  delivered boolean,
+  success boolean,
+  tries int,
+  created_at timestamp,
+  next_retry_at timestamp,
+  status int,
+  error jsonb,
+  error_message text,
+  error_type text
+)
+stable
+security invoker
+language plpgsql as $$
+begin
+  return query (
+    select
+      (el.payload->'data'->'new'->>'id')::int as model_id,
+      el.payload->'data'->'new'->>'name' as model_name,
+      el.payload->'data'->'new'->>'version' as model_version,
+      el.payload->'session_variables'->>'x-hasura-user-id' as triggering_user,
+      el.delivered,
+      eil.status is not distinct from 200 as success, -- is not distinct from to catch `null`
+      el.tries,
+      el.created_at,
+      el.next_retry_at,
+      eil.status,
+      (eil.response -> 'data'->> 'message')::jsonb as error,
+      (eil.response -> 'data'->> 'message')::jsonb->>'message' as error_message,
+      (eil.response -> 'data'->> 'message')::jsonb->>'type' as error_type
+      from hdb_catalog.event_log el
+      join hdb_catalog.event_invocation_logs eil on el.id = eil.event_id
+      where trigger_name = _trigger_name);
+end;
+$$;
+comment on function hasura.get_event_logs(_trigger_name text) is e''
+ 'Get the logs for every run of a Hasura event with the specified trigger name.';
+
+create view hasura.refresh_activity_type_logs as
+  select * from hasura.get_event_logs('refreshActivityTypes');
+comment on view hasura.refresh_activity_type_logs is e''
+ 'View containing logs for every run of the Hasura event `refreshActivityTypes`.';
+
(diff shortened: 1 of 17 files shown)


deployment/postgres-init-db/sql/views/hasura/hasura_event_logs.sql
```diff
@@ -0,0 +1,60 @@
...
+      el.payload->'data'->'new'->>'name' as model_name,
+      el.payload->'data'->'new'->>'version' as model_version,
+      el.payload->'session_variables'->>'x-hasura-user-id' as triggering_user,
+      el.delivered,
+      eil.status is not distinct from 200 as success, -- is not distinct from to catch `null`
+      el.tries,
+      el.created_at,
+      el.next_retry_at,
+      eil.status,
+      (eil.response -> 'data'->> 'message')::jsonb as error,
```
skovati, line 33: Non-blocking: Why have both the error and its fields in this view?

skovati (approved):
Looks good to me! Excited to have our first migration in the new DB!

skovati:
Forgot to ask in my review, where is the migration script "typo" that you fixed? Just the "aerie" vs "Aerie"?

deployment/postgres-init-db/sql/views/hasura/hasura_event_logs.sql
Mythicaeda: The fields picked out are the ones that are most likely to contain the most relevant info for debugging the failure, but the overall `error` field is there in case there's additional information that wasn't in those fields that are important. I included `error` since these views are the only insight most users will be able to have into these event tables.

Mythicaeda:
> Just the "aerie" vs "Aerie"?

Yeah, apparently Hasura's case sensitive 🥲

PR #1420, opened 2024-04-25
Feature/channel dictionary hookup

cohansen:
More sequence editor work, ignoring the PR template for now and we can add that as part of the final merge into develop.

Front-end PR: https://github.com/NASA-AMMOS/aerie-ui/pull/1244


goetzrrGit (approved):
Honestly not much to change here as it works with the frontend.

PR #1423, opened 2024-04-26
Procedural scheduling library

### LINKED ISSUE #1362: Design & implement plan editing and simulation control
Design the interface for procedural goal plan edits and simulation. To be presented on friday

JoelCourtney:
* **Tickets addressed:** closes #1362 
* **Review:** By commit  

## Description
- The main part of this PR is the new interfaces for scheduling procedures, plan editing, and simulation. These are just interfaces, not implementations. The implementations are in another branch, `feature/procedural-scheduling` because they are part of scheduler-driver.
- This PR undoes the creation of the timeline Duration class, instead merging the functionality with the old merlin-protocol version. Some extension functions to give Duration some operator overloading in Kotlin are also included. As part of this, I've reduced `Duration.MIN_VALUE` and `MAX_VALUE` to half of long min and max. See `Duration.MIN_VALUE` for an explanation of why.
- It also splits the `Plan` interface into both `Plan` and `SimulationResults`, to work better with the new `EditablePlan` interface.

## Verification
Since this PR doesn't include the implementation of `EditablePlan`, there isn't really anything to test. The changes to the Duration class are already extensively tested everywhere throughout Aerie.

## Documentation
The doc comments on the public facing API are complete. You can view the documentation by running `./gradlew :procedural:dokkaHtmlMultiModule`. You should see the scheduling lib docs next to and integrated with the docs for the timeline and constraints libraries.

## Future work
Currently all the simulation options are commented out, because I haven't supported them in the EditablePlan implementation. I'll uncomment them as they become supported later.


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -118,25 +110,34 @@ private Duration(final long durationInMicroseconds) {
...
-   * The value of this quantity should not be assumed.
-   * Currently, this is precisely -9,223,372,036,854,775,808 microseconds, or approximately -293,274 years.
+   * The value of this quantity should not be assumed. Currently, it equals half of long-min microseconds.
+   * This is for two reasons:
+   * 1. It may avoid some overflow errors by giving a 2x margin.
+   * 2. Java durations are serialized to ISO8601 duration format using only hours, minutes, and fractional seconds.
+   *    This is to avoid timekeeping issues related to leap seconds and leap years. Postgres intervals can represent
+   *    durations up to hundreds of millions of years, but only by using days and months. Thus, attempting to
+   *    interact with Postgres using a duration of long-min or long-max microseconds, represented as hours, causes an overflow.
+   *    However, long-min / 2 and long-max / 2 microseconds are small enough to not cause overflow.
```
mattdailis, line 120: Discussion of serialization concerns, and specific database systems, are implementation details of the clients of the Duration class, not inherent properties. I do think it's valuable to discuss rationale, but it should be in general terms that are relevant to all possible clients of this class.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -148,6 +149,7 @@ private Duration(final long durationInMicroseconds) {
   public static final Duration MINUTE = SECOND.times(60);
   /** One hour (h), equal to 60m. */
   public static final Duration HOUR = MINUTE.times(60);
+  public static final Duration DAY = HOUR.times(24);
```
mattdailis, line 152: Take a look at the javadoc comment at the top of the file, which provides rationale for not including DAY as a primitive time unit. What are your thoughts on that?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -233,19 +265,19 @@ public static Duration negate(final Duration duration) throws ArithmeticExceptio
...
 
   public static Duration saturatingAdd(final Duration left, final Duration right) {
-    return new Duration(saturatingAddInternal(left.durationInMicroseconds, right.durationInMicroseconds));
+    return new Duration(saturatingAddInternal(left.micros, right.micros));
   }
 
   private static long saturatingAddInternal(final long left, final long right) {
     long result = left + right;
     if (((result ^ left) & (result ^ right)) < 0) {
       return Long.MIN_VALUE - (result >>> (Long.SIZE - 1));
```
mattdailis: I haven't looked deeply into this bit shifting logic - does it continue to apply with the proposed reduced range of representable durations?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -118,25 +110,34 @@ private Duration(final long durationInMicroseconds) {
...
   /**
-   * The largest observable negative span of time. Attempting to go "more negative" will cause an exception.
+   * The largest valid negative span of time. Attempting to go "more negative" may cause an exception.
    *
    * <p>
-   * The value of this quantity should not be assumed.
-   * Currently, this is precisely -9,223,372,036,854,775,808 microseconds, or approximately -293,274 years.
+   * The value of this quantity should not be assumed. Currently, it equals half of long-min microseconds.
+   * This is for two reasons:
+   * 1. It may avoid some overflow errors by giving a 2x margin.
```
mattdailis, line 115: Can we quantify what the new range of durations is - is it around -140 to 140 years?

procedural/scheduling/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+var timelineSource = "${project(":procedural:timeline").projectDir}/src/main/kotlin"
+
+dokkaHtmlPartial.configure {
+  dokkaSourceSets {
+    configureEach {
+      // used as project name in the header
+      moduleName.set("Scheduling")
+
+      reportUndocumented.set(true)
+//      failOnWarning.set(true)
```
mattdailis: Commented code - uncomment, delete, or elaborate?

procedural/scheduling/src/main/kotlin/gov/nasa/jpl/aerie/procedural/scheduling/Procedure.kt
```diff
@@ -0,0 +1,14 @@
...
+import gov.nasa.jpl.aerie.timeline.CollectOptions
+
+/** The interface that all scheduling procedures must satisfy. */
+interface Procedure {
+  /**
+   * Run the procedure.
+   *
+   * @param plan A plan representation that can be edited and simulated.
+   */
+  fun run(plan: EditablePlan, options: CollectOptions)
```
mattdailis: Did we make a call on the user-facing nature of `CollectOptions` here?

procedural/scheduling/src/main/kotlin/gov/nasa/jpl/aerie/procedural/scheduling/simulation/SimulateOptions.kt
```diff
@@ -0,0 +1,8 @@
+package gov.nasa.jpl.aerie.procedural.scheduling.simulation
+
+/** Configuration for simulation. */
+/* data */ class SimulateOptions(
+//    val checkPointGeneration: CheckpointGeneration = CheckpointGeneration.None,
```
mattdailis, line 6: Commented code - uncomment, delete, or elaborate?

If we decide to leave it - let's add a TODO explaining what metric a future developer can use to decide if this code is ready to be uncommented

procedural/timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/AeriePostgresSimulationResults.kt
```diff
@@ -0,0 +1,167 @@
...
+import gov.nasa.jpl.aerie.timeline.ops.coalesce.CoalesceSegmentsOp
+import gov.nasa.jpl.aerie.timeline.payloads.Segment
+import gov.nasa.jpl.aerie.timeline.payloads.activities.Instance
+import java.io.StringReader
+import java.sql.Connection
+import javax.json.Json
+import kotlin.jvm.optionals.getOrNull
+
+/** A connection to Aerie's database for a particular simulation result. */
+data class AeriePostgresSimulationResults(
```
mattdailis: My understanding is that we're moving the postgres capabilities out of the library itself - is that accurate?

mattdailis (commented):
Dropping comments on the first things that jumped out at me - I'd still like to give NewDirective and friends a closer read to make sure I understand how they're intended to be used

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
JoelCourtney: Oops, I missed the early return. I'm just adding a clamp. Its less efficient but I didn't write the bitwise logic to begin with (that was someone on stackoverflow that I did not have the foresight to document 🤦 ).

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
JoelCourtney: I agree that there are multiple definitions of a day that could be useful, but the usual definition of day is so ubiquitous that omitting it has always confused me. I think if you ask anyone, even a rover planner, how long a day is, they will say "24 hours", not "are we asking about Mars?". Even when martian sols or sidereal days are relevant, I don't think anyone will blindly assume that `Duration.DAY` refers to anything other than a synodic day, and if we don't define it, most of our users will. The same applies to a week but I forgot to add it.

I agree that we shouldn't provide months or years, but not for the stated reasons. Days have multiple definitions, but months arguably have no definition as a pure duration. It would be problematic to define a month as 30 days, not because there are alternate definitions, but because it is wrong.

edit: changed "all of our users will" to "most of..."

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
JoelCourtney: Moved into a non-doc comment

procedural/scheduling/src/main/kotlin/gov/nasa/jpl/aerie/procedural/scheduling/Procedure.kt
JoelCourtney: In my demos and experimenting I did not find it useful, so I'll remove it. I think it should stay in the constraint interface though.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
JoelCourtney: Added. Its around 145,000 years

procedural/scheduling/src/main/kotlin/gov/nasa/jpl/aerie/procedural/scheduling/simulation/SimulateOptions.kt
JoelCourtney: Added a comment saying when to uncomment those lines. The pause behavior is MVP and should already be working on the other branch, but I haven't tested it yet.

procedural/timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/AeriePostgresSimulationResults.kt
JoelCourtney: Yes. I've already moved some of the implementations out of the scheduling and timeline libraries into scheduler-driver, but not this one because I don't have a place to put it yet. I can make a procedural-utils library quickly.

procedural/scheduling/build.gradle
JoelCourtney: deleted. Dokka does something weird for this package and reports some undocumented items in java.lang.annotation that we can't fix.

procedural/timeline/src/main/kotlin/gov/nasa/jpl/aerie/timeline/plan/AeriePostgresSimulationResults.kt
JoelCourtney: Added `...aerie.procedural.remote`.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
mattdailis: Given that no duration longer than a second has a reliable definition (leap seconds result in minutes with 59, 60, or 61 seconds, which propagate up to affect the containing hour and day. Days may have 23, 24, or 25 hours around daylight savings boundaries), let's not let the edge cases get in the way of providing useful default durations. 

I think having DAY as shorthand for 24 hours follows the principle of least surprise. If someone needs a different definition, they can define one. 

Longer term, I'd like to look into moving some of the convenience functionality out of the protocol, whose primary (and, ideally, sole) responsibility is to define the methods and data structures that the merlin driver uses to interrogate a mission model

dandelany:
Closing this in favor of combined Procedural Scheduling PR: https://github.com/NASA-AMMOS/aerie/pull/1496

PR #1427, opened 2024-04-26
Quiet down successful builds

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  

## Description
This PR fixes or silences all warnings in the current build process. For silenced warnings that need justification, I've put my reasons in the commit descriptions.

The only one that I'm not sure about is the first commit. @goetzrrGit I specifically need your review for that one line. Basically, the TS compiler in sequencing-server gave a warning that said something like "import assertions aren't allowed unless module is set to esnext or nodenext". So I set `"module": "esnext"` in tsconfig. I have no idea what this change means or if it will cause problems.

## Verification
If tests pass and successful builds are silent, I'm happy. You should be able run `./gradlew build --warning-mode all` and see no output other than the Build Successful message.

## Future work
The war for quiet builds is eternal.

### DIFF
--- a/constraints/build.gradle
+++ b/constraints/build.gradle
@@ -34,6 +34,8 @@ dependencies {
   implementation project(':parsing-utilities')
 
   testImplementation 'org.junit.jupiter:junit-jupiter-engine:5.10.0'
+
+  testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
 
 publishing {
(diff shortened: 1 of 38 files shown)


goetzrrGit (approved):
I tested the changes with the sequencing server and did not have an issues with manual and e2e sequence expansion. Sorry about the confusion at standup.

dandelany (approved):
Built successfully 👍 LGTM

PR #1435, opened 2024-05-02
Fix Delete trigger for v2.6.3

Mythicaeda:
* **Review:** By commit  

## Description

This PR and its sister PR against develop fix two bugs with our scheduling specifications:
1) The migration for Versioning Scheduling Goals did not update the `when` condition of the `on update` trigger for scheduling specifications, leading to the stack depth unexpectedly exploding when trying to update a spec 38+ goals (the threshold may be lower, but I did not look into this).
2) The function used for `on delete` did not properly handle bulk deletes (as in a delete statement that deletes multiple rows). This function and its trigger have been updated to operate "per-statement" and go through all the rows deleted in reverse priority order. This ensures that 

Independently of these bugs, I also put a depth check on the `increment spec update when the goals subtable is updated` trigger, as there's no reason for us to update the revision when a trigger changes a goal's priority.

A simple example of the second bug follows:
Let's say we have the following specification. Specification are being expressed as an array ordered by priority and "tailing" priorities will be expressed as `null` in order to avoid confusion caused by the array changing size mid-walkthrough.
`[ Goal 1, Goal 2, Goal 3, Goal 4, Goal 5, Goal 6 ]`

Now let us delete Goal 1, Goal 2, and Goal 5. The expected output for this action is`[Goal 3, Goal 4, Goal 6, null, null, null]`

Under the old logic, the following would happen:

1. Goal 1, Goal 2, and Goal 5 are deleted: `[null, null, Goal 3, Goal 4, null, Goal 6]`
2. The `after delete` trigger for Goal 1 is processed. As Goal 3, Goal 4, and Goal 6 have a strictly higher priority than Goal 1, their priorities are reduced by 1: `[null, Goal 3, Goal 4, null, Goal 6, null]`
3. The `after delete` trigger for Goal 2 is processed. As Goal 4 and Goal 6 have a strictly higher priority than Goal 2, their priorities are reduced by 1: `[null, Goal 3 and Goal 4, null, Goal 6, null, null]`
4. The `after delete` trigger for Goal 5 is processed. As no Goals have a strictly higher priority than Goal 5, no rows are updated: `[null, Goal 3 and Goal 4, null, Goal 6, null, null]`
5. An exception is raised, as Goal 3 and Goal 4 cannot share a priority.

Note that preserving the old per-row function but instead reducing goals with a priority greater than or equal to the current goal's would _not_ have been sufficient, as it would lead to gaps in priorities (as well as conflicting priorities in different setups). For example, in the above example, `Goal 6` has been placed at priority 4 before we process `Goal 5`'s deletion, meaning it will not be moved to priority 3, even though that's where we would expect it to be at the end of the transaction, leaving a gap at priority 3.

Under the new logic, the following occurs:

1. Goal 1, Goal 2, and Goal 5 are deleted: `[null, null, Goal 3, Goal 4, null, Goal 6]`
2. The `after delete` for step 1 begins. `Goal 5`, as the deleted Goal with the highest priority, is processed first: `[null, null, Goal 3, Goal 4, Goal 6, null]`
3. Goal 2, as the deleted Goal with the second-highest priority, is processed next: `[null, Goal 3, Goal 4, Goal 6, null, null]`
4. Goal 1, as the deleted Goal with the lowest priority, is processed last: `[Goal 3, Goal 4, Goal 6, null, null, null]`
5. The `after delete` for step 1 completes.

The output in this case is exactly as expected, as by processing in reverse priority order, we guarantee that all goals with a strictly higher priority than that of the deleted goal we're processing is still at a strictly higher priority than that deleted goal, even if that priority has changed.

## Verification

New DB tests will be included in the variant of this branch to be merge into main.

## Documentation

We need to document that PGCMP does not catch `when` conditions on triggers so that viewers are aware to keep their eye out for it.

## Future work

Looking into extending PGCMP to implement the `when` condition trigger match.

### DIFF
--- a/.github/scripts/explanations
+++ b/.github/scripts/explanations
@@ -1,8 +1,8 @@
 schema	migrations	migrations	missing in 1st DB	From Migration 0	2
 function definition	migrations	migrations.migrations.mark_migration_applied(character varying)	missing in 1st DB	From Migration 0	3
 function definition	migrations	migrations.migrations.mark_migration_rolled_back(character varying)	missing in 1st DB	From Migration 0	3
-function permissions	migrations	migrations.mark_migration_applied(character varying)-role:aerie	missing in 1st DB	From Migration 0	3
-function permissions	migrations	migrations.mark_migration_rolled_back(character varying)-role:aerie	missing in 1st DB	From Migration 0	3
+function permissions	migrations	migrations.mark_migration_applied(character varying)-role:aerie_test_username	missing in 1st DB	From Migration 0	3
+function permissions	migrations	migrations.mark_migration_rolled_back(character varying)-role:aerie_test_username	missing in 1st DB	From Migration 0	3
 function permissions	migrations	migrations.mark_migration_applied(character varying)-role:PUBLIC	missing in 1st DB	From Migration 0	3
 function permissions	migrations	migrations.mark_migration_rolled_back(character varying)-role:PUBLIC	missing in 1st DB	From Migration 0	3
 function owner	migrations	migrations.mark_migration_applied(character varying)	missing in 1st DB	From Migration 0	3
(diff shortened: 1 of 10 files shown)


skovati (approved):
Crazy bug... nice find. Approving now, but depending on how soon we need `2.6.3` released, maybe we can wait to merge and cut a release until DB tests are passing on the sister `develop` PR as a sanity check.

Mythicaeda:
Change just pushed was a fix for the model version of delete discovered while writing DB Tests for the develop version of this PR. In summary, forgot to replace `old` with `r` when updating the model function, causing rows to be improperly updated.

PR #1443, opened 2024-05-10
Write parents unfinished spans

Mythicaeda:
* **Review:** By commit  

## Description

@AaronPlave noticed a bug where unfinished spans would always have their `parentId` set to `null`. This was fixed by providing the list `allActivityRecords`, which contains both complete and unfinished spans, to the `updateSimulatedActivityParentsAction` instead of `simulatedActivityRecords`, which only contains the completed spans.

## Verification

A new E2E test was added. This test simulates a plan containing two decomposing activities, one which finishes, and the other which is unfinished. It then verifies that **all** child spans have their `parentId` fields populated.

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SimulationTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SimulationTests.java
@@ -16,7 +16,9 @@ import org.junit.jupiter.api.TestInstance;
 import javax.json.Json;
 import javax.json.JsonValue;
 import java.io.IOException;
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


mattdailis (approved):
This rings a bell... I believe #1400 made the same change. You've added a test, though, so let's merge this PR and close the older one

Mythicaeda:
> This rings a bell... I believe #1400 made the same change. You've added a test, though, so let's merge this PR and close the older one

Looking at it, it did. Ah, and I see why it hadn't been approved yet--I had left a comment requesting the addition of a test.
Do we have a ticket for the `Future Work` part of that PR body yet, by the way? If not, I can make one.

PR #1446, opened 2024-05-13
Test Private Repo Cloning

Mythicaeda:
Test PR to determine if we can clone a private repo within the same organization.


.github/workflows/private-clone.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+      - name: Clone Private Repo
+        uses: actions/checkout@v4
+        with:
+          repository: NASA-AMMOS/aerie-orbiter-model
+          path: aerie-orbiter-model
+          ref: main
```
skovati, line 26: this should be possible if you create a fine-grained token (with read access to `aerie-orbiter-model`) secret and reference it like:

```yaml
    token: ${{ secrets.READ_ONLY_TOKEN }}
```

PR #1456, opened 2024-05-23
Update Docker-Compose for the Sequencing Server

goetzrrGit:
* **Tickets addressed:** NONE
* **Review:** By commit  

## Description
@dandelany noticed high CPU usage on the sequencing server, even during idle periods. This seems to be caused by a development-related watcher process used when developing the sequencing server.

To address this, I set the default values in the Docker Compose to match `deployment` which makes the CPU utilization normal. If you specifically need to develop using the sequencing server and hook up a debugger within the ide, you can uncomment the necessary sections in the Docker Compose configuration file.

Additionally, I exposed the knobs to the sequencing server in the deployment Docker Compose as they were missing. 

## Verification
There is now very low CPU usage in a development environment.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -74,7 +74,8 @@ services:
   aerie_sequencing:
     build:
       context: ./sequencing-server
-      dockerfile: Dockerfile.dev
+      dockerfile: Dockerfile
+      #dockerfile: Dockerfile.dev #-- Uncomment for development
     container_name: aerie_sequencing
     depends_on: ["postgres"]
     environment:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -74,7 +74,8 @@ services:
   aerie_sequencing:
     build:
       context: ./sequencing-server
-      dockerfile: Dockerfile.dev
+      dockerfile: Dockerfile
+      #dockerfile: Dockerfile.dev -- Uncomment for development
```
Mythicaeda: Minor, nonblocking: If you use another `#` instead of a `--` here and L97, `Uncomment for development` will automatically become commented out when you uncomment the line.

Mythicaeda (approved):

docker-compose.yml
goetzrrGit: Good idea. I just made the change.

dandelany (commented):
👍 thanks @goetzrrGit !

PR #1461, opened 2024-05-30
Added a table to track the default view for a given mission model

### LINKED ISSUE #1403: Specify default view for a model
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

_No response_

Users should be able to specify a default view for a model. This would enable projects to load a particular view that is tuned to their needs when creating new plan or opening an existing plan.

cohansen:
* **Tickets addressed:** Closes #1403.
* **Review:** By file

## Description
This PR adds a mapping table that users can use to make a default view for a given mission model. When the user navigates to the plan the UI will load the default view if one is set.

## Verification
Manually tested in the UI to make sure the default view is being loaded if it's set.

## Documentation
None

## Future work
Add a UI to enable creating / deleting these default views.

Associated frontend PR: https://github.com/NASA-AMMOS/aerie-ui/pull/1310


deployment/hasura/migrations/Aerie/4_mission_model_default_view/down.sql
```diff
@@ -0,0 +1,5 @@
+drop trigger set_timestamp on view_to_mission_model;
+
+drop table view_to_mission_model;
```
Mythicaeda: `view_to_mission_model` needs the schema qualifier: `ui.view_to_mission_model`

deployment/hasura/migrations/Aerie/4_mission_model_default_view/up.sql
```diff
@@ -0,0 +1,41 @@
...
+    unique (mission_model_id, view_id),
+
+  foreign key (mission_model_id)
+    references merlin.mission_model (id)
+    on delete cascade,
+
+  foreign key (view_id)
+    references ui.view (id)
+    on delete cascade
+);
```
Mythicaeda: While refining this write-up to properly break down my proposed changes, I realized that this table could be simplified into a new column on the `mission_model` table: `default_view_id int default null references ui.view`. This is because a model will always have up to one association with a view, making it a bad candidate for an association table like this, which is used when we have a one-to-many or many-to-many association (ie, **one** activity directive can have **many** tags). Additionally, this is metadata relevant to a specific mission model, and will be commonly looked up by the `model_id`, which further supports it being on the `mission_model` table.


Original write-up breakdown of this table is below:

1) Update the name to better describe its contents. This table specifically contains the default view for a mission model. A better name would be something like `model_default_view`.
2) Use `mission_model_id` as the pk instead of an auto-generated `id`. Each mission model can only have exactly 1 default view (so `mission_model_id` must be unique), and this is the column we will be using to look up entries in this table by. Therefore, it should be the PK.
3) Remove the `updated_at` and `created_at` columns. We don't include these on association tables, and `updates` are forbidden on this table anyway.

```sql
create table ui.model_default_view (
  mission_model_id integer not null,
  view_id integer not null,

  primary key (mission_model_id)

  foreign key (mission_model_id)
    references merlin.mission_model (id)
    on delete cascade,

  foreign key (view_id)
    references ui.view (id)
    on delete cascade
);
```

deployment/postgres-init-db/sql/tables/merlin/mission_model.sql
```diff
@@ -48,6 +52,8 @@ comment on column merlin.mission_model.created_at is e''
   'The time this mission model was uploaded into Aerie.';
 comment on column merlin.mission_model.description is e''
   'A human-meaningful description of the mission model.';
+comment on column merlin.mission_model.default_view_id is e''
+  'The ID of an option default view for the mission model.';
```
Mythicaeda: Nitpick, typo: 
```suggestion
  'The ID of an optional default view for the mission model.';
```

deployment/hasura/metadata/databases/tables/merlin/mission_model.yaml
```diff
@@ -103,7 +106,7 @@ insert_permissions:
       columns: [mission, name, version, description, jar_id]
```
Mythicaeda: The `default_view` column needs to be added to the `columns` list here and on L113. Otherwise nobody can set a default view :-P

cohansen: We don't care about this column triggering any of the `event_triggers` right?

Mythicaeda: Yup!

Mythicaeda (approved):
LGTM!

PR #1471, opened 2024-06-11
Added a default value to sequence adaptation names so migrations dont…

cohansen:
… break

* **Tickets addressed:** NA
* **Review:** By commit

## Description
The previous PR broke against a real migration because we didn't provide a default value for the `name` of a `sequence_adaptation`. This adds a UUID.

I tried using the current ID but that seemed more complicated for a problem that should be very low impact.

## Verification
NA

## Documentation
NA

## Future work
NA

### DIFF
--- a/deployment/hasura/migrations/Aerie/5_sequence_adaptation_name/down.sql
+++ b/deployment/hasura/migrations/Aerie/5_sequence_adaptation_name/down.sql
@@ -1,5 +1,5 @@
 alter table sequencing.sequence_adaptation
-  drop constraint sequence_adaptation_natural_key,
+  drop constraint sequence_adaptation_unique_key,
   drop column name;
 
 call migrations.mark_migration_rolled_back('5');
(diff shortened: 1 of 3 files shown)


deployment/hasura/migrations/Aerie/5_sequence_adaptation_name/down.sql
```diff
@@ -1,5 +1,5 @@
 alter table sequencing.sequence_adaptation
-  drop constraint sequence_adaptation_natural_key,
+  drop constraint sequence_adaptation_unique_key,
```
Mythicaeda: Oh I was saying in the thread that the usual naming pattern would be like `sequence_adaptation_name_unique`

deployment/postgres-init-db/sql/tables/sequencing/sequence_adaptation.sql
```diff
@@ -1,15 +1,15 @@
 create table sequencing.sequence_adaptation (
   id integer generated always as identity,
   adaptation text not null,
-  name text not null,
+  name text not null default gen_random_uuid(),
```
Mythicaeda, line 4: This lets users leave `name` blank while creating new `sequence_adaptation`s--is this desired?

cohansen: The UI blocks this behavior so I don't think this is a huge deal, if they upload directly through the API without providing a name I think a generated name is alright.

Mythicaeda (approved):

PR #1477, opened 2024-06-13
Remove unnecessary cast from `hasura.get_event_logs`

Mythicaeda:
* **Review:** By commit  

## Description

Grabbing the `message` field from `errors` as a `json` rather than as `text` that is then converted back to `jsonb` is more efficient, more straightforward, and avoids unnecessary errors when `message` is not a valid json.

Additionally, the type of `errors` in the function has been updated from `jsonb` to `json` to reflect the type that Hasura stores the underlying field as.

## Verification

Query was tested against the `aerie_dev` database, which contains mission model logs where `message` is not a valid json.

### DIFF
--- a/deployment/hasura/migrations/Aerie/6_hasura_events_views_500_fix/down.sql
+++ b/deployment/hasura/migrations/Aerie/6_hasura_events_views_500_fix/down.sql
@@ -1,4 +1,10 @@
-create or replace function hasura.get_event_logs(_trigger_name text)
+drop view hasura.refresh_resource_type_logs;
+drop view hasura.refresh_model_parameter_logs;
+drop view hasura.refresh_activity_type_logs;
+drop function hasura.get_event_logs(_trigger_name text);
+
+-- Recreate with new function signature
+create function hasura.get_event_logs(_trigger_name text)
 returns table (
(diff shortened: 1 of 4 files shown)


mattdailis:
> and avoids unnecessary errors when message is not a valid json

Should I take this to mean that `json` is allowed to be any text, while `jsonb` must be valid json?

Mythicaeda:
> Should I take this to mean that json is allowed to be any text, while jsonb must be valid json?

No, it's the active cast that's the issue, and this issue would occur regardless of if we were casting to `jsonb` or `json`.

When you do `->` it comes out as "JSON object field" [according to the docs](https://www.postgresql.org/docs/current/functions-json.html). I'm not sure _exactly_ what Postgres is doing under the hood, but a "JSON object field" is flexible in definition (ie it's allowed to be a string or an int or a nested JSON Object, etc). However, when you actively cast to `json` via `::json`, Postgres is looking for either `null` or something like `'{}'` or `'5'` or `'true'` or `'"a"'`. Since `''` is none of these, the cast throws an exception rather than converting to the JSON Object version of the string `''`.

skovati (approved):

PR #1486, opened 2024-06-20
Naming efficiency

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Several proposed performance optimizations for the Aerie streamline framework, all indicated by profiling done on the CADRE simplified data model. In particular:
1. Changes many more string operations done when computing names to be lazy; those string operations are only done if a name is requested. Since names are primarily a debugging concept, in the vast majority of cases, names are assigned but never requested, making this a significant performance improvement.
2. Refrains from adding duplicate constraints to the active queue in LinearBoundaryConsistencySolver, limiting re-work when solving.
3. Optimizes Expiry.or to use Optional methods instead of converting to Streams. This is less elegant but much more performant.
4. Using effectively immutable lists for Context, so that daemon tasks can share references with their parent task's context rather than copying and rebuilding the list. Since daemon tasks contribute most of the load of context management, it makes sense to optimize for their primary use case.
5. Re-writing the clampedIntegrate function to directly calculate its results, rather than leaning on the LinearBoundaryConsistencySolver. While this requires some tinkering with lower-level concepts, it roughly doubles the speed of clampedIntegrate. For models that use this function, it can be a performance bottleneck.

## Verification

All changes were tested by hand on the CADRE simplified data model.

That model, running a 1 day plan, went from taking a few minutes to taking ~15s, with all the changes in this PR. Also, it looks like ~5s of that time is spent in simulation, and the remaining ~10s are spent saving the results.

## Documentation

N/A - pure refactor

## Future work

N/A

Note: This PR makes #1483 and #1405 obsolete.

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Naming.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Naming.java
@@ -28,15 +28,19 @@ public final class Naming {
   // Use a WeakHashMap so that naming a thing doesn't prevent it from being garbage-collected.
   private static final WeakHashMap<Object, Function<NamingContext, Optional<String>>> NAMES = new WeakHashMap<>();
 
-  private record NamingContext(Set<Object> visited, Optional<String> anonymousName) {
+  private record NamingContext(Set<Object> visited, Function<Object, Optional<String>> anonymousName) {
     NamingContext visit(Object thing) {
       var newVisited = new HashSet<>(visited);
       newVisited.add(thing);
       return new NamingContext(newVisited, anonymousName);
     }
 
+    public NamingContext() {
+      this(Set.of(), $ -> Optional.empty());
+    }
+
     public NamingContext(String anonymousName) {
-      this(Set.of(), Optional.ofNullable(anonymousName));
+      this(Set.of(), anonymousName == null ? $ -> Optional.of(Objects.toString($)) : $ -> Optional.of(anonymousName));
     }
   }
 
@@ -52,7 +56,9 @@ public final class Naming {
       for (int i = 0; i < args$.length; ++i) {
         var argName$ = Optional.ofNullable(args$[i].get())
                 .flatMap(argRef -> getName(argRef, context));
-        if (argName$.isEmpty()) return context.anonymousName();
+        if (argName$.isEmpty()) {
+          return context.anonymousName.apply(args$[i].get());
+        }
         argNames[i] = argName$.get();
       }
       return Optional.of(nameFormat.formatted(argNames));
@@ -66,8 +72,13 @@ public final class Naming {
    * returns empty.
    */
   public static Optional<String> getName(Object thing) {
-    return getName(thing, new NamingContext(null));
+    return getName(thing, new NamingContext());
   }
+
+  // TODO - build a version of getName, or make this the default behavior, (maybe if anonymousName is null?)
+  //   which calls Object.toString on things that don't otherwise have a name.
+  //   This supports the use of literals and dynamics objects in effect names,
+  //   without needing to eagerly compute their names, which involves costly string ops.
 
   /**
    * Get the name for thing.
@@ -80,8 +91,8 @@ public final class Naming {
 
   private static Optional<String> getName(Object thing, NamingContext context) {
     return context.visited.contains(thing)
-            ? context.anonymousName
-            : NAMES.getOrDefault(thing, NamingContext::anonymousName).apply(context.visit(thing));
+            ? context.anonymousName.apply(thing)
+            : NAMES.getOrDefault(thing, ctx -> ctx.anonymousName.apply(thing)).apply(context.visit(thing));
   }
 
   public static String argsFormat(Collection<?> collection) {
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
@@ -9,14 +9,7 @@ import gov.nasa.jpl.aerie.contrib.streamline.core.Resources;
 import gov.nasa.jpl.aerie.contrib.streamline.core.monads.ExpiringMonad;
 import gov.nasa.jpl.aerie.merlin.framework.Condition;
 
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Queue;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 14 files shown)


contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/PolynomialResources.java
```diff
@@ -359,52 +361,125 @@ public static Resource<Polynomial> integrate(Resource<Polynomial> integrand, dou
    *   0   /             \-----
    * time  0        10        20
    * </pre>
+   * <p>
+   *     NOTE: This method assumes that lowerBound <= upperBound at all times.
+   *     Failure to meet this precondition may incorrect outputs, crashing the simulation, stalling in an infinite loop,
```
mattdailis:
```suggestion
   *     Failure to meet this precondition may lead to incorrect outputs, crashing the simulation, stalling in an infinite loop,
```

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Naming.java
```diff
@@ -66,9 +72,14 @@ public static <T> T name(T thing, String nameFormat, Object... args) {
    * returns empty.
    */
   public static Optional<String> getName(Object thing) {
-    return getName(thing, new NamingContext(null));
+    return getName(thing, new NamingContext());
   }
 
+  // TODO - build a version of getName, or make this the default behavior, (maybe if anonymousName is null?)
```
mattdailis, line 78: @DavidLegg correct me if I'm wrong - it appears to me that this TODO describes the current behavior of NamingContext when anonymousName is null. Should we delete the comment as done?

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
```diff
@@ -132,8 +125,11 @@ private void solve() {
                   "LinearBoundaryConsistencySolver %s failed. Domain for %s is empty: [%s, %s]".formatted(
                       getName(this).orElseThrow(), D.variable, D.lowerBound, D.upperBound));
             }
-            // TODO: Make this more efficient by not adding constraints that are already in the queue.
-            remainingConstraints.addAll(neighboringConstraints.get(D.variable));
+            for (DirectionalConstraint constraintToAdd : neighboringConstraints.get(D.variable)) {
+              if (!remainingConstraints.contains(constraintToAdd)) {
```
mattdailis, line 129: Makes me wonder if a LinkedHashSet would be appropriate here

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
```diff
@@ -327,9 +323,50 @@ private Stream<DirectionalConstraint> directionalConstraints(Variable constraine
...
+    public Function<Map<Variable, ? extends Domain>, Expiring<Polynomial>> bound() {
+      return bound;
+    }
+
+    public Set<Variable> drivingVariables() {
+      return drivingVariables;
+    }
+
+    // *Don't* override hashCode and equals - we want to use object identity here instead of value equality.
+    // This makes it more performant during solving to check whether a constraint is already in the queue to be propagated.
```
mattdailis, line 359: I appreciate the comment. It does make me wonder whether we're missing out on the ability to prune distinct, but equivalent constraints. Perhaps that's rare enough not to warrant special consideration.

mattdailis (approved):

dandelany:
Thanks @DavidLegg for the massive performance improvement and @mattdailis for reviewing! Merging so that we can include this in the 2.14 release. FYI I have to "resolve" the conversations above in order to merge, but feel free to unresolve & continue the convo if there's more to be said.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Naming.java
DavidLegg: Yeah, good catch. We can just delete that comment.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
DavidLegg: That seems like it would work. With the change to clampedIntegrate away from using this solver, optimizing the solver is less important. Using a set instead of a list would simplify the code, though.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/polynomial/LinearBoundaryConsistencySolver.java
DavidLegg: Just me thinking-in-text. TL;DR - I think there's something here, but I'm not sure it's worth chasing until we see a real use case that's bottlenecked by the solver.

My first guess would be that that's rare... The constraint expansion basically re-arranges the constraints given by the user to phrase them in terms of an expression constraining each involved variable. If you give a set of independent constraints, I think the expansions should all be independent too, at least to start. It's possible that one constraint will dominate another over some restriction of the domains, though. Consider "X <= Y + 10" and "X <= 2Y". For Y <= 5, the latter controls X, for Y >= 5 the former does. If at some point you restricted the domain for Y to just one of those conditions, say you find that Y is in [7, 20], then at least in theory you could throw away the non-dominant constraint for the rest of this solving iteration. The fact that the constraints are linear ought to help here. There's just one or two levels too many of linear algebra for me to fully complete that thought at the moment, though...

Kind of expanding on that idea though... It may also be possible to determine something about when a constraint is "tight" (i.e., re-evaluating it might actually restrict a domain further) vs. "slack" (it's not the limiting constraint, no need to re-evaluate), even if neither one is strictly dominant like in the prior example. I don't know if the bookkeeping to maintain that would be more or less expensive than just running the restriction process, though. I kind of suspect that level of granularity is equivalent to just propagating the constraint.

The other common way of optimizing a solver like this would be constraint ordering. If we could sort the constraints well, we might be able to reduce the number of constraint propagations we do. Something like "the constraint with the most variable domains that have changed is propagated next", or "the constraint with the most restricted domains is propagated next" might be good enough. Note that we don't have to do backtracking, so the usual search heuristics kind of don't apply.

DavidLegg: As I was thinking through this a bit more, I think there's one easier optimization we could make without too much effort. Since all constraints are linear inequalities, they do the restriction by looking at one endpoint of each variables' domain, either high or low. Currently, we add constraints to re-propagate whenever an involved domain changes, but we don't take into account which endpoint of the domain changed. I.e., "X <= 2Y" only cares about the high end of Y's domain, but currently, if Y were restricted from [0, 10] to [5, 10], we'd still queue this constraint to be re-propagated.

If the keys in `neighboringConstraints` were not just variables, but variable + which side of the domain changed, we could probably throw out at least some irrelevant constraints with minimal extra effort.

Issue #1468, opened 2024-06-07
Scheduler does not respect mutex condition


parkerabercrombie:
### Checked for duplicates

No - I haven't checked

### Is this a regression?

Yes - This worked in a previous version

### Version

2.11.0

### Describe the bug

I ran scheduling with two goals (in priority order):

B/6hr
```
export default (): Goal => {
  return Goal.ActivityRecurrenceGoal({
    interval: Temporal.Duration.from({hours: 6}),
     activityTemplate: ActivityTemplates.TEL_Set_GroupID({ groupID: 34 }) }) }
```

A/1d
```
export default (): Goal => {
  return Goal.ActivityRecurrenceGoal({
    interval: Temporal.Duration.from({days: 1}),
    activityTemplate:ActivityTemplates.GENERIC_Placeholder({
      duration:"000T12:00:00",
      team: "TEST",
      activityName: "TEST" }) }) }
```

And one scheduling condition:

```
 export default (): GlobalSchedulingCondition => {
  return GlobalSchedulingCondition.mutex(
    [ActivityType.GENERIC_Placeholder], [ActivityType.TEL_Set_GroupID]
); }
```

When I run the scheduler I get a plan with overlapping GENERIC_Placeholder and TEL_Set_GroupID activites, which is in violation of the scheduling condition.
![image](https://github.com/NASA-AMMOS/aerie/assets/1181836/1b24826a-ab35-444d-8099-e60b46a9ef8c)

Plan start: 2030-001T00:00:00.000
Plan end: 2030-005T00:00:00.000

### Reproduction

See bug description above.

### Logs

```shell
N/A
```

### System Info

```shell
Chrome
```

### Severity

Moderate

parkerabercrombie:
Clipper is not relying on scheduling conditions for current conop, so not a high priority bug for us.

dandelany:
Thanks @parkerabercrombie - related to or duplicate of #1336

adrienmaillard:
@parkerabercrombie @srschaffJPL 
So far, can't reproduce. I have a mission model jar from march 18, 2024. When I try running this example, I get an error on the compilation of this global scheduling condition. It seems there is an issue with `GENERIC_Placeholder` activity type being absent from generated types. I have `GENERIC_Power_Data` and `GENERIC_Epoch` but not `GENERIC_Placeholder`. 
(1) do you have any other goal with that activity type ? do you use it otherwise ? is it draggable from the UI ? 
[...]

PR #1488, opened 2024-06-21
Fix rootfinding by enforcing bounds of initial x-values

### LINKED ISSUE #1336: scheduling mutex condition does not prevent overlap
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

2.2.1

### Describe the bug

Using a mutex scheduling condition does not prevent scheduling of conflicting activities that have duration spans that overlap.

It does appear to prevent them from scheduling the same _start time_ though. It also does work in some other priority orderings.

### Reproduction

Create a plan:
- 2030-001T00:00:00 to 2030-005T00:00:00
- eurc model (g24 release)

This condition:
- A+B: export default (): GlobalSchedulingCondition => { 
  return GlobalSchedulingCondition.mutex(
    [ActivityType.ECM_Power_ON, ActivityType.GENERIC_Power_Data],
    [ActivityType.ECM_Power_ON, ActivityType.GENERIC_Power_Data]
    ); }

These goals:
- B/6h: export default (): Goal => { return Goal.ActivityRecurrenceGoal({
  interval: Temporal.Duration.from({hours: 6}),
  activityTemplate: ActivityTemplates.ECM_Power_ON()
  }) }
- A/1d: export default (): Goal => { return Goal.ActivityRecurrenceGoal({
  interval: Temporal.Duration.from({days: 1}),
  activityTemplate: ActivityTemplates.GENERIC_Power_Data({
    duration:"000T12:00:00", type: "A",
    loadWatts: 0, data: []
    }) }) }

Press schedule. It places 16 activities for B (ECM_Power_On) as expected.
Considering the mutex, this should leave no room at all to place a 12hr long A (GENERIC_Power_Data) anywhere in the plan.
However, it also places 4 instances for A.

(Notably, the mutex works as expected in the reverse priority ordering, ie A then B)

### Logs

_No response_

### System Info

```shell
2023 M2 Macbook Pro Ventura 13.6.4 (22G513)
Chrome 121.0.6167.184
G24 eurc model release https://github.jpl.nasa.gov/Europa-PESS/clipper-aerie/releases
```

### Severity

Moderate

### Workaround

None currently outside of manually scheduling

adrienmaillard:
* **Tickets addressed:** Closes #1336 #1468 
* **Review:** By commit  

## Description

There was a bug in the rootfinding algorithm. A heuristic value was provided without checking its bounds. Then the algorithm was using it and going off bound, causing it to find a solution that did not exist, violating the mutex conditions. 

## Verification

Validated against the clipper case, the bug disappeared. 

## Documentation

None. 

## Future work

None.

### DIFF
--- a/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
+++ b/scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
@@ -127,7 +127,7 @@ public class RootfindingTest {
         Duration.of(-2, Duration.SECONDS),
         twoSecond,
         100);
-    assertEquals(29, solution.history().getHistory().size());
+    assertEquals(30, solution.history().getHistory().size());
     assertEquals(new EquationSolvingAlgorithms.FunctionCoordinate<>(Duration.of(0, Duration.MICROSECONDS), Duration.of(0, Duration.MICROSECONDS)), solution.functionCoordinate());
   }
 
@@ -164,7 +164,7 @@ public class RootfindingTest {
         Duration.of(-2, Duration.SECONDS),
         twoSecond,
         100);
-    assertEquals(2, solution.history().getHistory().size());
-    assertEquals(new EquationSolvingAlgorithms.FunctionCoordinate<>(Duration.of(0, Duration.MICROSECONDS), Duration.of(0, Duration.MICROSECONDS)), solution.functionCoordinate());
+    assertEquals(1, solution.history().getHistory().size());
+    assertEquals(new EquationSolvingAlgorithms.FunctionCoordinate<>(Duration.of(-860925, Duration.MICROSECONDS), Duration.of(0, Duration.MICROSECONDS)), solution.functionCoordinate());
   }
 }
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
@@ -131,7 +131,7 @@ public class EquationSolvingAlgorithms {
       int i = 0;
       do {
         //we should not come back to previously visited values
-        if (!history.alreadyVisited(cur)) {
+        if (!history.alreadyVisited(cur) && cur.between(min, max)) {
           i++;
           try {
             final var value = function.valueAt(cur, history);


adrienmaillard:
Clarification about why this happens as I have seen this is not clear. Let's say you want to find the root of f. Your x domain is [1,3], your y domain is [5,6]. You start at x0, provided as an argument of the rootfinding procedure. In our case, it's gonna be 1 (we take the lower bound of the x domain). The algorithm actually needs 2 starting values to really begin. we compute f(x0) = y0, let's say that f(1) = 8. Then we use a heuristic that assumes that the function is constant so we say x1 (the next x to be tested) is x1 = x0 - 8 (as f(x) is assumed to be x+8 and we want f(x) = 0). x1 = -7 which is outside of the x domain [1,3]. This x1 is passed to the `nextValueAt` as an init value. The domain of values found by `nextValueAt` was checked... except for the init value. So in the faulty case, it computes f(x1) and returns. f(x1) satisfies constraints on the y domain and rootfinding is done even though x1 is outside of the x domain. The x domain is determined by the application of constraints beforehand, of which mutex is one.

Mythicaeda:
> Clarification about why this happens as I have seen this is not clear. Let's say you want to find the root of f. Your x domain is [1,3], your y domain is [5,6]. You start at x0, provided as an argument of the rootfinding procedure. In our case, it's gonna be 1 (we take the lower bound of the x domain). The algorithm actually needs 2 starting values to really begin. we compute f(x0) = y0, let's say that f(1) = 8. Then we use a heuristic that assumes that the function is constant so we say x1 (the next x to be tested) is x1 = x0 - 8 (as f(x) is assumed to be x+8 and we want f(x) = 0). x1 = -7 which is outside of the x domain [1,3]. This x1 is passed to the `nextValueAt` as an init value. The domain of values found by `nextValueAt` was checked... except for the init value. So in the faulty case, it computes f(x1) and returns. f(x1) satisfies constraints on the y domain and rootfinding is done even though x1 is outside of the x domain. The x domain is determined by the application of constraints beforehand, of which mutex is one.

Why does the algorithm want the second value to be `f(x)=0` instead of `f(x_max)=y`? If it took `f(x_min)` and `f(x_max)` as its first two values, then it shouldn't ever escape the fn's x-domain (and would remove the need for `x_init`, assuming of course that the caller isn't optimizing this). Even if `x_init` was kept, having the second point be the x-domain bound it's further from would still give the alg two points while ensuring it's in the bounds (while avoiding issues should `x_init` be `x_min` or `x_max`.

dandelany:
@Mythicaeda and I talked this morning & confirmed that the change in this PR is a) safe and b) mitigates the issue reported by users, so I'd like to get this merged for our `v2.14.0` release today. I've opened #1490 to continue the conversation from above, answer any remaining questions & determine if there's a deeper improvement to the algo to be made.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
```diff
@@ -164,7 +164,7 @@ public Duration valueAt(
         Duration.of(-2, Duration.SECONDS),
         twoSecond,
         100);
-    assertEquals(2, solution.history().getHistory().size());
-    assertEquals(new EquationSolvingAlgorithms.FunctionCoordinate<>(Duration.of(0, Duration.MICROSECONDS), Duration.of(0, Duration.MICROSECONDS)), solution.functionCoordinate());
+    assertEquals(1, solution.history().getHistory().size());
+    assertEquals(new EquationSolvingAlgorithms.FunctionCoordinate<>(Duration.of(-860925, Duration.MICROSECONDS), Duration.of(0, Duration.MICROSECONDS)), solution.functionCoordinate());
```
Mythicaeda, line 168: Why did this value change? Additionally, why did the history _decrease_ in this test and _increase_ in the other test?

adrienmaillard:
> Why does the algorithm want the second value to be f(x)=0 instead of f(x_max)=y? If it took f(x_min) and f(x_max) as its first two values, then it shouldn't ever escape the fn's x-domain (and would remove the need for x_init, assuming of course that the caller isn't optimizing this). Even if x_init was kept, having the second point be the x-domain bound it's further from would still give the alg two points while ensuring it's in the bounds (while avoiding issues should x_init be x_min or x_max.

Our overall use of rootfinding is to find f(x) = y but we transpose this problem to a traditional rootfinding problem (there is one method with the y argument that transposes to another method without the y... because it's gonna be 0).
The goal of traditional rootfinding is to find x such that f(x) = 0. Here we add bounds for x and f(x). The bounds for f(x) are just the wiggle room around 0 that we are willing to go with. 

And yes it can escape because of the heuristic I talked about. This heuristic is good in most cases but it can sometimes get us outside of the bounds... not anymore though.

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/RootfindingTest.java
adrienmaillard: it increased because the x1 value was out of the domain but still came back to the domain afterwards. But it still needed one more step to solve, it happens. You can go out of the x domain and then come back. 

For this one, the init value I am giving in this test is actually out of the x domain in the beginning. So instead of using the heuristic value, it uses another value randomly and finds the solution straight away.

Mythicaeda (approved):

dandelany:
Thanks for the fix @adrienmaillard and for the detailed review @Mythicaeda - discussion continued in #1490. I'm gonna go ahead and merge this for our `2.14` release today.

PR #1492, opened 2024-06-27
Added a method to check for static parameters inside of an ActivityType

### LINKED ISSUE #1309: Disallow static Activity Parameters
The `@Export.Parameter` annotation should not be permitted on `static` fields. Permitting Activity Parameters to be static can lead to activities running with parameters that they were not assigned to. The attached images show an example of this.

Screenshot 2024-01-30 at 3 03 42 PM
Screenshot 2024-01-30 at 3 03 49 PM

cohansen:
* **Tickets addressed:** Closes #1309.
* **Review:** By commit

## Description
After loading the activity types during mission model building I added a check for any parameters we've loaded that are marked as being static.

## Verification
Tested manually, to test mark a parameter inside of Bananation and run the assemble task, it should fail.

## Documentation
NA

## Future work
NA

### DIFF
--- a/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
+++ b/merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
@@ -386,6 +386,10 @@ import java.util.stream.Collectors;
     final var validations = this.getExportValidations(activityTypeElement, parameters);
     final var effectModel = this.getActivityEffectModel(activityTypeElement);
 
+    if (!parameters.isEmpty()) {
+      checkForStaticParameters(name, parameters);
+    }
+
     final var durationParameterName = effectModel.flatMap(EffectModelRecord::durationParameter);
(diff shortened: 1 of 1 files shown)


merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/MissionModelParser.java
```diff
@@ -409,6 +413,17 @@ class (old-style) or as a record (new-style) by determining
         effectModel);
   }
 
+  private void checkForStaticParameters(String activityName, List<ParameterRecord> parameters) throws InvalidMissionModelException {
+    for (final var parameter: parameters) {
+      if (parameter.element.getModifiers().contains(Modifier.STATIC)) {
+        throw new InvalidMissionModelException(
```
mattdailis, line 419: I think there's a way to have the error point directly to a code location... but I don't see any examples of that in this class, so no changes requested.

mattdailis (approved):
Looks great!

PR #1495, opened 2024-07-08
Span to Event Traceability

### LINKED ISSUE #1426: Associate simulation events with spans for better traceability
### Background

Simulations can emit **simulation events** when they are run. This happens automatically when activities are created, and the user can also add their own events to act as useful log messages or warnings. Thanks to the changes in https://github.com/NASA-AMMOS/aerie-ui/pull/1236 these events will be visible in the UI to provide extra visibility into what's going on "under the hood" when a sim is run.

Per @mattdailis we should be able to associate these events with the spans/activities that create them, which would help users trace them to their source.

### Requirements

* Add information to simulation events to reference their associated `Span`s
* Display the span info in the sim events table (https://github.com/NASA-AMMOS/aerie-ui/pull/1236)
* Ideally have UI which lets you link/scroll to the Span in question from the sim events table

goetzrrGit:
* **Tickets addressed:** Closes #1426 
* **Review:** By commit  

## Description
Currently, users can't easily trace log messages (created as SimulationEvents) back to the specific activity (Span) in the timeline UI. This traceability gap makes it difficult to troubleshoot errors captured by a log message within an activities models.

To address this, this PR proposes adding a new span_index column to both the `span` and `event` tables. This column will link each `event` to its corresponding `span`, allowing users to jump the timeline from an event within the `Simulation Event Table` to the relevant activity in the timeline with a single click. 

There is a UI PR as well you can find it here: https://github.com/NASA-AMMOS/aerie-ui/pull/1363

## Verification
Successfully implemented event-to-span linking in the UI! Users can now click a simulation event and jump directly to the corresponding timeline span.


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
```diff
@@ -29,7 +31,8 @@
           e.transaction_index,
           e.causal_time,
           e.topic_index,
-          e.value
+          e.value,
+          e.span_index
```
mattdailis: One lingering span_index that should be span_id

```suggestion
          e.span_id
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -560,10 +561,44 @@ public static SimulationActivityExtract computeActivitySimulationResults(
         engine,
         startTime,
         elapsedTime,
-        computeTaskInfo(timeline, activityTopic, serializableTopics)
+        computeSpanInfo(timeline, activityTopic, serializableTopics)
     );
   }
 
+  public static HashMap<SpanId, ActivityDirectiveId> spanToActivityDirectiveId(
```
mattdailis: This method should be private - in part because it's a helper method for computeResults, and in part because it requires the caller to provide a SpanInfo object, which is a private inner class of SimulationEngine.

```suggestion
  private static HashMap<SpanId, ActivityDirectiveId> spanToActivityDirectiveId(
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -560,10 +561,44 @@ public static SimulationActivityExtract computeActivitySimulationResults(
...
+  ){
+    final var activityDirectiveIds = new HashMap<SpanId, ActivityDirectiveId>();
+    engine.spans.forEach((span, state) -> {
+      if (!spanInfo.isActivity(span)) return;
+      if (spanInfo.isDirective(span)) activityDirectiveIds.put(span, spanInfo.getDirective(span));
+    });
+    return activityDirectiveIds;
+  }
+
+  public static HashMap<SpanId, SimulatedActivityId> spanToSimulatedActivities(
```
mattdailis: This method should also be private
```suggestion
  private static HashMap<SpanId, SimulatedActivityId> spanToSimulatedActivities(
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -728,19 +748,50 @@ public static SimulationResults computeResults(
...
                 Optional<SerializedValue> serializedEvent = trySerializeEvent(event, serializableTopic);
                 if (serializedEvent.isPresent()) {
-                  output = EventGraph.concurrently(output, EventGraph.atom(Pair.of(serializableTopicToId.get(serializableTopic), serializedEvent.get())));
+
+
+                  // Make sure we have a true mapping between span and events
+                  // if not loop back up the parents untill we find one
+                  // we can potentially not find a parent
+                  var activitySpanID = Optional.ofNullable(spanToActivities.get(event.provenance()).id());
+                  if (activitySpanID.isEmpty()) {
```
mattdailis: I'm going to need to pore over this section for a little longer - leaving a comment for now so I can submit the rest of this review

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
```diff
@@ -63,24 +66,32 @@ public SortedMap<Duration, List<EventGraph<Pair<Integer, SerializedValue>>>> get
...
       final var causalTime = resultSet.getString(3);
       final var topicIndex = resultSet.getInt(4);
       final var serializedValue = parseSerializedValue(resultSet.getString(5));
+      Optional<Long> spanIndex = Optional.empty();
+      try {
+        final var index = resultSet.getLong(6);
+        spanIndex = Optional.ofNullable( index == 0 ? null : index );
+      } catch (final SQLException e) {
+        // ignore
+      }
```
mattdailis: Can you help me understand this section a bit better? What is the special meaning of index === 0?

Let's rename the `spanIndex` variable to `spanId` here.

Here's another part of the code where we're loading a nullable long - perhaps this pattern is applicable here as well? https://github.com/NASA-AMMOS/aerie/blob/2cfc9cd42bd1210c5dba85256c82ab3d5c766723/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSpanRecords.java#L50

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
```diff
@@ -20,8 +20,8 @@
 
 /*package-local*/ final class PostSpansAction implements AutoCloseable {
   private static final @Language("SQL") String sql = """
-      insert into merlin.span (dataset_id, start_offset, duration, type, attributes)
-      values (?, ?::timestamptz - ?::timestamptz, ?::timestamptz - ?::timestamptz, ?, ?::jsonb)
+      insert into merlin.span (span_id,dataset_id, start_offset, duration, type, attributes)
```
mattdailis: At this point, I believe we already have the `parent_id` available. Let's include it in the `insert` statement, which will allow us to skip the "UpdateSimulatedActivityParent" action

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -420,15 +421,14 @@ private static void postActivities(
...
+      postActivitiesAction.apply(
           datasetId,
           allActivityRecords,
           simulationStart);
 
       updateSimulatedActivityParentsAction.apply(
           datasetId,
-          allActivityRecords,
-          simIdToPgId);
+          allActivityRecords);
```
mattdailis: Once we update the `PostSpansAction` to set the parent id, we can delete the update action altogether

mattdailis (changes requested):
Overall direction looks great to me - a few small but important details left to work out

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Experimenting here... not sure this quite has all the properties we want yet, but don't want to leave you hanging too long

```java
// If the event's `provenance` has no simulated activity id, search its ancestors to find the nearest
// simulated activity id, if one exists
if (!spanToActivities.containsKey(event.provenance())) {
  var spanId = event.provenance();
  while (true) {
    if (spanToActivities.containsKey(spanId)) {
      spanToActivities.put(event.provenance(), spanToActivities.get(spanId));
      break;
    }
    final var parentSpanId = engine.getSpan(spanId).parent();
    if (parentSpanId.isEmpty()) {
      break;
    }
  }
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
goetzrrGit: oops, one got by my eyes. Thanks for catching this.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
goetzrrGit: From the getLong() javadocs

```
Returns:
the column value; if the value is SQL NULL, the value returned is 0
```

if the column is null then I will get a 0 value. I don't want to store `Optional.of(0) `I want `Optional.empty()` because I don't have a spanID associated for the event. 

I will inline things to make it better:

```java
final var serializedValue = parseSerializedValue(resultSet.getString(5));
final Optional<Long> spanId  = resultSet.getLong(6) == 0 ? Optional.empty() : Optional.of(
            Long.valueOf(resultSet.getLong(6)));
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
goetzrrGit: Got it made it private

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
goetzrrGit: Got it made it private

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
goetzrrGit: Quick question: Are we going to get into a situation where the `parent_id` is referencing a `span_id` that is not inserted into the table yet?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
mattdailis: Thanks for the clarification - unfortunately I think 0 should be considered a valid span id (even if we don't currently generate any zeros). `resultSet.getObject(6) == null` may be a more robust null check here.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
mattdailis: Ah... that's a good point. I'd like this to be a bulk insert, if possible (one insert statement, many values). If that proves difficult (or if we're concerned that can get too large), we can do a quick topological sort and insert in parents-first order

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetSimulationEventsAction.java
goetzrrGit: Got it let me change this up a little

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
goetzrrGit: Modified the above code a little and added it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
goetzrrGit: We are going s batch insert but that still caused a problem. Sorting by adding the span that have null parents helped.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
goetzrrGit: I removed the function

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Looks good to me! Good catch that it was never updating `spanId` 🤦

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostSpansAction.java
mattdailis: Nice 👍 That sort was a good first step - it certainly helps for parents and children - but I think if there are grandchildren we're back to the same problem. I pushed a commit adding a full topological sort, which essentially does what you did over and over, first all the top level parents, then all their children, then all their grandchildren, etc...

https://github.com/NASA-AMMOS/aerie/pull/1495/commits/8e11693a906fa258328ced0f4127fd7b1f5c8c05

mattdailis (approved):

PR #1500, opened 2024-07-10
Add streamline Logger

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Adds a Logger class to the streamline framework, which uses topics to emit log messages directly as events. Also builds such a Logger when the Registrar is initialized, and reconfigures errors to go to that logger when the error behavior is set to LOG. Before this, errors were collected awkwardly into a discrete string resource.

## Verification

Since the logger is connected to the Registrar's error behavior, this was tested using the streamline-demo example model. By adding "CauseError" activities to the plan, we can trip the resource error handling machinery, and verify that appropriate error messages appear in the Simulation Events tab.

image

## Documentation

None yet - #1494 should probably be updated to tell users how to use the logger in their model/activities.

This does remove the `errors` resource, but it leaves the `numberOfErrors` resource intact, which we should note for any users who may have depended on `errors`.

## Future work

None planned

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Logger.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Logger.java
@@ -1,0 +1,60 @@
+package gov.nasa.jpl.aerie.contrib.streamline.debugging;
+
+import gov.nasa.jpl.aerie.contrib.streamline.core.Resources;
+import gov.nasa.jpl.aerie.merlin.framework.Registrar;
+
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Map;
+
+import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECONDS;
+import static java.util.stream.Collectors.toMap;
+
+public class Logger {
+    private static final int LEVEL_INDICATOR_SIZE = Arrays.stream(LogLevel.values())
+            .map(v -> v.toString().length())
+            .max(Integer::compareTo)
+            .orElseThrow();
+    private static final String LOG_MESSAGE_FORMAT = "%s [%-" + LEVEL_INDICATOR_SIZE + "s] %s";
+
+    private final Map<LogLevel, SimpleLogger> subLoggers;
+    private final Instant planStart;
+
+    public Logger(Registrar registrar, Instant planStart) {
+        // TODO - refactor the plan start out, by instead building a global sim clock to do the same job.
+        //  To do that "right", the global clock should be a resource with custom dynamics for continuous Instant clocks.
+        subLoggers = Arrays.stream(LogLevel.values()).collect(toMap(
+                level -> level,
+                level -> new SimpleLogger(level.name(), registrar)));
+        this.planStart = planStart;
+    }
+
+    public void log(LogLevel level, String messageFormat, Object... args) {
+        Instant time = planStart.plusNanos(Resources.currentTime().in(MICROSECONDS) * 1_000);
+        String message = messageFormat.formatted(args);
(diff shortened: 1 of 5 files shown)


contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/debugging/Logger.java
```diff
@@ -0,0 +1,60 @@
...
+            .map(v -> v.toString().length())
+            .max(Integer::compareTo)
+            .orElseThrow();
+    private static final String LOG_MESSAGE_FORMAT = "%s [%-" + LEVEL_INDICATOR_SIZE + "s] %s";
+
+    private final Map<LogLevel, SimpleLogger> subLoggers;
+    private final Instant planStart;
+
+    public Logger(Registrar registrar, Instant planStart) {
+        // TODO - refactor the plan start out, by instead building a global sim clock to do the same job.
```
mattdailis: @DavidLegg and I discussed that this TODO comment should be addressed before merging this PR. In particular - there's no need to explicitly include the log timestamp in the message, because every serializable event is time-tagged

mattdailis (changes requested):
See https://github.com/NASA-AMMOS/aerie/pull/1500/files#r1705853363

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/Registrar.java
```diff
@@ -64,17 +59,13 @@ public enum ErrorBehavior {
     Throw
   }
 
-  public Registrar(final gov.nasa.jpl.aerie.merlin.framework.Registrar baseRegistrar, final ErrorBehavior errorBehavior) {
+  public Registrar(final gov.nasa.jpl.aerie.merlin.framework.Registrar baseRegistrar, final Instant planStart, final ErrorBehavior errorBehavior) {
```
mattdailis: If we can avoid changing the signature of the Registrar constructor, I think this PR can be backwards compatible with code that already uses the streamline framework

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/modeling/Registrar.java
```diff
@@ -66,27 +56,11 @@ public enum ErrorBehavior {
...
-    return "Error affecting %s:%n%s".formatted(
-        String.join(", ", affectedResources),
-        formatException(e));
-  }
 
-  private static String formatException(Throwable e) {
-    return ExceptionUtils.stream(e)
-        .map(ExceptionUtils::getMessage)
-        .collect(joining("\nCaused by: "));
+    discrete("numberOfErrors", numberOfErrors, new IntegerValueMapper());
```
mattdailis, line 63: In the event that users have written constraints that depend on the previous resource definition, it may be worth calling out this change in the release notes. Not a reason not to make this change, but worth communicating explicitly.

It also highlights that constraints (and scheduling?) would benefit from being able to refer to simulation events.

mattdailis (approved):
Thanks for making this backwards compatible! Looks good to me

EDIT: Resolving conversation below so it doesn't block merge, but copying up here for visibility:
> In the event that users have written constraints that depend on the previous resource definition, it may be worth calling out in the release notes [that the errors resources has been replaced with a numberOfErrors resource]. Not a reason not to make this change, but worth communicating explicitly.

DavidLegg:
One slight correction to that note on the `errors` and `numberOfErrors` resources - We actually had `numberOfErrors` already, just derived in a different way. This PR does remove `errors` itself, though, so I agree we should put that in the release notes just in case.

PR #1507, opened 2024-07-18
Stream Simulation Resources

### LINKED ISSUE #1096: Stream simulation results into the database

Mythicaeda:
* **Tickets addressed:** Closes #1096 
* **Review:** By commit  

## Description

Should not be breaking. 

Extracts resource management out of the engine and into a new class called the `SimulationResourceManager` the `InMemorySimulationResourceManager` operates nearly the same way as on `develop`, except that the serialized form of the profile segment is kept around, rather than the pre-serialized form. The `Simulate` action uses the `StreamingSimulationResourceManager`, which streams profile segments to the DB in batches.

Additionally, the engine has been slightly refactored: 
- the `step` and `init` logic which was repeated verbatim in all `simulate` methods has been pulled into methods in the engine
- the engine now tracks its own `timeline`, `cells`, and `elapsedTime`
    - related, the logic of tracking mutliple timelines for checkpoint simulation has been simplified into the engine containing a `referenceTimeline`, which is defined as the `referenceTimeline` combined with the `timeline` of the engine it was copied from
- `computeResults` and related methods have been made non-static, as they require an instance of a `simulationEngine` to execute and now rely on internal fields of the engine.

Additionally, since all but the `SynchronousSimulationAgent` are unused, the interface `SimulationAgent` has been flattened.

## Verification

Manually verified that streaming is working.

Did some profiling on `streaming` vs `develop`. I used Clipper’s `gnc-all-activities` plan and ran profiling twice, once with high-overhead memory instrumentation and once with async sampling. Findings:
- Streaming has a positive impact on overall Heap memory usage. From what I can tell, the heap usage is always <= develop, with the  difference being ~0.1GB, and given that the max heap usage is ~1GB, that’s up to a 10% save (tho its closer to a 0% save while streaming). I’d definitely prefer to do this comparison again against a plan that consumes more memory, though.
    - I took a look at the breakdowns and, from what I can tell, in Streaming most Heap memory was in the young “Eden” scape, while on develop most Heap memory was in the older “Old Gen/Survivor” scape (preference seemed to depend on whether I was profiling memory)
- For non-heap memory, streaming appeared to have a neglible to slight negative impact--although we’re talking within a couple MBs. Of note is that Develop has an increase of non-heap memory at the end, while Streaming is consistent all the way through--at least when doing memory profiling. When not profiling, while Streaming starts with a higher non-heap memory, at about halfway through the usage is just under halved, while develop’s usage remains approximately constant. I don’t know what the subcategories for non-heap memory mean, so I’ll skip an analysis.
- Streaming took slightly longer than develop - 27min vs 25min with profiling on, 2min v 1min with profiling off. This may be avoidable by having streaming work in its own thread.

I also ran a JDBC profiler alongside the memory profiler and of note:
- the threshold for streaming may need to be tweaked, as Streaming spent 50 seconds total awaiting Network I/O calls when inserting profile segments (reminder that this simulation took 27 minutes total), although the average individual wait was 150 microseconds. Similarly, it spent 18 seconds total on the segments being actively written, with the average execution of 50 microseconds
   - Taking a quick look at the code, I accidentally enabled the “returned generated columns” flag on the “insert segments” statement, which may be the reason the Network I/O was so high for that statement, since that flags asks the JDBC to request excess data back from the DB
- We may want to take a quick peek at simulation_extent, as an abnormally long amount of time was spent updating this table--8 seconds overall on develop, 3 seconds on streaming, with the average execution time for the statement being 11ms on develop, 3ms on streaming.

Raw Snapshots for those curious: [Streaming.zip](https://github.com/user-attachments/files/16367229/Streaming.zip)

## Documentation

No docs need to be updated.

## Future work

[Investigate `simulation_extent`](https://github.com/NASA-AMMOS/aerie/issues/1508)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/SimulationResourceManager.java
```diff
@@ -0,0 +1,39 @@
...
+import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.Map;
+import java.util.Set;
+
+public interface SimulationResourceManager {
+  int DEFAULT_THRESHOLD = 1024;
```
Mythicaeda: make `final` and drag down to `Streaming`, add a comment explaining what its for

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/InMemorySimulationResourceManager.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.Set;
+
+/**
+ * A variant of the SimulationResourceManager that keeps all segments in memory
+ */
+public class InMemorySimulationResourceManager implements SimulationResourceManager {
+  private final HashMap<String, ResourceSegments<RealDynamics>> realResourceSegments;
+  private final HashMap<String, ResourceSegments<SerializedValue>> discreteResourceSegments;
+
+  private final int threshold;
```
Mythicaeda: Remove threshold

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/StreamingSimulationResourceManager.java
```diff
@@ -0,0 +1,198 @@
...
+    return profiles;
+  }
+
+  /**
+   * Compute a subset of the ProfileSegments stored in this resource manager
+   * @param elapsedDuration the amount of time elapsed since the start of simulation.
+   * @param resources the set of names of the resources to be computed
+   * @throws UnsupportedOperationException this class streams all resources it has as it accepts updates,
+   *    so it cannot only compute a subset of ProfileSegments.
+   */
```
Mythicaeda: Just say we don't support it

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/SimulationResourceManager.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Compute a subset of the ProfileSegments stored in this resource manager
+   * @param elapsedDuration the amount of time elapsed since the start of simulation.
+   * @param resources the set of names of the resources to be computed
+   */
+  ResourceProfiles computeProfiles(final Duration elapsedDuration, Set<String> resources);
+
+  /**
+   * Process resource updates for a given time.
```
Mythicaeda, line 28: Times given must be monotonically increasing

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -93,8 +95,17 @@ public static int getNumActiveSimulationEngines() {
   /** A thread pool that modeled tasks can use to keep track of their state between steps. */
   private final ExecutorService executor;
 
-  public SimulationEngine() {
+  /* The top-level simulation timeline. */
+  private final TemporalEventSource timeline;
+  private final LiveCells cells;
+  private Duration elapsedTime;
+
+  public SimulationEngine(MissionModel<?> model) {
```
Mythicaeda: just pass in the cells

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -138,6 +153,100 @@ private SimulationEngine(SimulationEngine other) {
...
+      if(results.error.isPresent()) {
+        throw results.error.get();
+      }
+    }
+  }
+
+  public sealed interface Status {
+    record NoJobs() implements Status {}
+    record AtDuration() implements Status{}
+    record Nominal(
```
Mythicaeda, line 180: Consider a different word @mattdailis

mattdailis: Are NoJobs and AtDuration different? My understanding was that one implies the other

Mythicaeda: If an activity is unfinished at the end of simulation, my understanding is that you will have jobs but be AtDuration

Mythicaeda:
Just pushed changes addressing the set of comments left during the walkthrough.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -816,6 +832,14 @@ public static SimulationResults computeResults(
...
   }
 
+  private static <V1, V2> void append(
+      final Duration currentTime,
+      ProfilingState<V2> profilingState,
+      V1 value
+  ) {
+    profilingState.profile().append(currentTime, (V2) value);
+  }
+
```
adrienmaillard: This method has been removed later, remove here directly ?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/InMemorySimulationResourceManager.java
```diff
@@ -0,0 +1,155 @@
...
+      final var finalSegment = segments.getLast();
+      profile.add(new ProfileSegment<>(elapsedDuration.minus(finalSegment.startOffset()), finalSegment.dynamics()));
+    }
+
+    return profiles;
+  }
+
+  /**
+   * Add new segments to this manager's internal store of segments.
+   * @param elapsedTime the amount of time elapsed since the start of simulation.  Must be monotonically increasing on subsequent calls.
```
adrienmaillard, line 125: Any way to enforce this constraint ?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
```diff
@@ -62,8 +65,12 @@ Map<String, SerializedValue> getModelEffectiveArguments(String missionModelId, M
          LocalMissionModelService.MissionModelLoadException,
          InstantiationException;
 
-  SimulationResults runSimulation(CreateSimulationMessage message, Consumer<Duration> writer, Supplier<Boolean> canceledListener)
-          throws NoSuchMissionModelException, MissionModelService.NoSuchActivityTypeException;
+  SimulationResults runSimulation(
+      CreateSimulationMessage message,
+      Consumer<Duration> writer,
+      Supplier<Boolean> canceledListener,
+      final SimulationResourceManager resourceManager
```
adrienmaillard: why did not you add the final modifier to these other arguments ?

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
```diff
@@ -198,8 +196,9 @@ public Map<String, SerializedValue> getModelEffectiveArguments(
   public SimulationResults runSimulation(
       final CreateSimulationMessage message,
       Consumer<Duration> simulationExtentConsumer,
-      Supplier<Boolean> canceledListener)
-  throws NoSuchMissionModelException {
+      Supplier<Boolean> canceledListener,
+      final SimulationResourceManager resourceManager
```
adrienmaillard: Same question as above

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -734,35 +733,94 @@ public static SimulationResults computeResults(
...
+                  }
+                  var activitySpanID = Optional.ofNullable(spanToActivities.get(event.provenance()).id());
+                  output = EventGraph.concurrently(
+                      output,
+                      EventGraph.atom(
+                          new EventRecord(serializableTopicToId.get(serializableTopic),
+                                          activitySpanID,
+                                          serializedEvent.get())));
+                }
+              }
```
adrienmaillard: Looks like code duplication from the computeResults below, can you explain ?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
```diff
@@ -139,6 +153,96 @@ private SimulationEngine(SimulationEngine other) {
...
+            Pair.of(
+                resource.getOutputType().getSchema(),
+                serializeProfile(elapsedTime, state, SimulationEngine::extractDiscreteDynamics)));
+
+        default ->
+            throw new IllegalArgumentException(
+                "Resource `%s` has unknown type `%s`".formatted(name, resource.getType()));
+      }
+    }
+     */
```
adrienmaillard: Remove commented code

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/CheckpointSimulationDriver.java
```diff
@@ -148,22 +150,25 @@ public static Function<SimulationState, Boolean> checkpointAtEnd(Function<Simula
...
     // this will give you its exact start time, but if there are any end-time anchors, this will give you the minimum time the activity could start at.
     final var minimumStartTimes = new HashMap<ActivityDirectiveId, Duration>();
-    for(final var activity : schedule.entrySet()){
+    for (final var activity : schedule.entrySet()) {
       var curInChain = activity;
       var curSum = ZERO;
-      while(true){
-        if(curInChain.getValue().anchorId() == null){
+      while (true) {
+        if (curInChain.getValue().anchorId() == null) {
```
adrienmaillard, line 123: This practice about spacing between reserved keywords and their arguments seems new to me in Aerie, right ? Why and is it part of a consistent style guide that we can find somewhere ?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/CheckpointSimulationFacade.java
```diff
@@ -89,15 +88,6 @@ public CheckpointSimulationFacade(
...
 
-  /**
-   * Returns the total simulated time
-   * @return
-   */
-  @Override
-  public Duration totalSimulationTime(){
-    return totalSimulationTime;
-  }
-
```
adrienmaillard: I don't think this should be removed, I want to use that as a performance metric that we can add to tests so we can see if changes to sim/scheduling decrease perf

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
```diff
@@ -0,0 +1,180 @@
+package gov.nasa.jpl.aerie.merlin.worker.postgres;
+
+import gov.nasa.jpl.aerie.json.JsonParser;
```
adrienmaillard, line 3: should not this commit and the previous one switched so first you create the streamer and then you use it ?

adrienmaillard (commented):
No big concern, just a few comments.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
mattdailis: Having trouble coming up with a different word too - let's not block this PR on it

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Mythicaeda: This code is already removed in a later commit (`Update Resource Tracking in Engine`).

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/CheckpointSimulationDriver.java
Mythicaeda: This spacing with `(` and `)` and similar symbols is something I've had @mattdailis suggest tweaks for on my PRs in the past and is how IntelliJ automatically respaces the file. I'm not sure if there is a style guide it's being sourced from (though he may know). 

If you are strongly against the reformatting, I can drop the commit/parts of the commit (I am strongly for breaking up long lines into multiple lines for readability)

adrienmaillard: I am not against it. IMHO, if something is not helping in terms of readability in these expressions is the absence of space between parentheses and the logical expression inside them. Between the `if` and the opening `parenthese`, I don't really see the point... Anyway, not blocking, just commenting that it'd be good to have one style shared by everyone that we all agree on and that we can input to intellij once and for all.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/CheckpointSimulationFacade.java
Mythicaeda: Can you give an example of what such a test would look like? While I see the hypothetical value in having a metric that can assess that we're selecting checkpoints that minimize the amount of time to be simulated, I cannot picture what a test using it would actually look like.

adrienmaillard: We could add one line to every scheduler test that would assert the current "simulation time".

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/CheckpointSimulationDriver.java
mattdailis: That's a good point - I think we have two different styles, one that's more prevalent in the scheduler, and one that occurs more in merlin. Personally, I prefer the extra spaces - I think they help make it clear what is a function call, and what is a keyword, but of course consistency is more important than going one way or the other.

The closest we have to a style guide is this .editorconfig file (which, in theory, is supposed to guide the IntelliJ autoformatter, but I haven't investigated whether that works or not). That file says there should be a space before the parens in an if statement:

https://github.com/NASA-AMMOS/aerie/blob/7b13d58d66c6229c8fe7405a251f87cb8d8fd781/.editorconfig#L185

Here's a convention that recommends the extra spaces - but again, I do believe this is a bit subjective.
https://www.oracle.com/java/technologies/javase/codeconventions-whitespace.html

Let's treat the extra spaces as our default "Aerie style" going forward.

@adrienmaillard that's an interesting point about the spacing inside parens - here's where that setting is set to false:
https://github.com/NASA-AMMOS/aerie/blob/7b13d58d66c6229c8fe7405a251f87cb8d8fd781/.editorconfig#L225

We can chat as a team to see how we feel about those interior spaces.

adrienmaillard: I have the `EditorConfig` enabled but intellij would not correct anything if I am not adding these spaces when I am writing new code. However, if I select some code and click on `Code` -> `Reformat code` I get the spaces.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/MissionModelService.java
Mythicaeda: Done

merlin-server/src/test/java/gov/nasa/jpl/aerie/merlin/server/mocks/StubMissionModelService.java
Mythicaeda: Done

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Mythicaeda: I've extracted this code into it's own method rather than repeating it in both `computeResults`. I've left it as it's own commit near the end to avoid rebase headaches.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
Mythicaeda: Commits have been reversed.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/CheckpointSimulationFacade.java
Mythicaeda: Method has been restored.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
Mythicaeda: Done

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/resources/InMemorySimulationResourceManager.java
Mythicaeda: Yes, I've updated the ResourceManagers to track what the last timestep they received an update for was, and to throw an exception if they get an earlier update.

adrienmaillard (approved):
Great work \o/

PR #1509, opened 2024-07-18
Refactor directive IDs in the scheduler

### LINKED ISSUE #1505: Refactor directive ids in scheduler
My understanding of the current use of directive IDs in the scheduler is this:
- The database acts as the source of truth for what the directive IDs "actually" are.
- In the scheduler, existing directive IDs are negated, while newly created activities use positive IDs.
- The scheduler also reindexes IDs entirely when calling simulation, using a counter in `SimulationFacadeUtils`.
- The scheduler performs one last map when uploading new activities to the database, allowing the database to decide the final ID in case new activities were added independently during scheduling.

This doesn't include any mapping between directive and instance IDs. All of these steps are mapping directive IDs around, and I believe the middle two are unnecessary and can be removed. Essentially we should always be consistent with the database when possible; it is only impossible for activities that aren't uploaded yet, and we should regain consistency as soon as they are inserted.

I will give the priority solver an autoincrementing ID cell that it can use to assign new IDs, starting 1 above the highest ID already in the plan. Since the cell will be unique per scheduling run, we won't get any collisions.

- I will probably delete the `SchedulingActivityDirectiveId` class in favor of just using `ActivityDirectiveId`.
- I will delete all of the directive-to-directive ID maps except for any in the upload step. This will impact `SchedulingActivityDirective`, `SimulationData`, `PrioritySolver`, `SimulationFacade.PlanSimCorrespondence`, `SimulationFacadeUtils` and `SynchronousSchedulerAgent`. The plan-sim-correspondence part of `CheckpointSimulationFacade` should be a bit simpler but still needs to exist to help searching for cached sim engines.

JoelCourtney:
* **Tickets addressed:** closes #1505 
* **Review:** By commit  

## Description
This contains a couple behavioral changes:
- All activity directive id mapping or transforming has been removed, except for the upload step where directive ids might not match what they ultimately become in the database.
- Generated activities are no longer assigned fake directive IDs. They are not directives, so this was unsound.

And several renames / trivial refactors:
- Renamed `SimulatedActivity[Id]` to `ActivityInstance[Id]` to better contrast with `ActivityDirective`.
- Deleted `SchedulingActivityDirectiveId` and replaced all usages with `ActivityDirectiveId`.
- Renamed `SchedulingActivityDirective` to `SchedulingActivity` because it also holds instance information.
- Renamed `MerlinService` and its implementors to `MerlinDatabaseService` because merlin is not the database.

## Verification
All tests still pass. Please suggest any new tests you think would be helpful.

## Documentation
I don't think we had any internal documentation of the ID maps, which is part of what made them so annoying. So none needs to be updated!


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/SchedulingActivity.java
```diff
@@ -0,0 +1,283 @@
...
+ * @param topParent the parent activity if any
+ * @param isNew whether this activity was created in this scheduling run, or already existed in the plan
+ */
+public record SchedulingActivity(
+    ActivityDirectiveId id,
+    ActivityType type,
+    Duration startOffset,
+    Duration duration,
+    Map<String, SerializedValue> arguments,
+    ActivityDirectiveId topParent,
```
mattdailis, line 41: Consider Optional here

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/SchedulingActivity.java
```diff
@@ -0,0 +1,283 @@
...
+   *
+   * @return the activity
+   */
+  public static SchedulingActivity getActWithEarliestStartTime(List<SchedulingActivity> acts) {
+    if (!acts.isEmpty()) {
+      acts.sort(Comparator.comparing(SchedulingActivity::startOffset));
+
+      return acts.getFirst();
+    }
+    return null;
```
mattdailis, line 213: Consider Optional here

mattdailis (approved):
A lot of very sensible changes here. Looks good to me based on visual inspection - if you can get the tests pass I'd say ship it

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/SchedulingActivity.java
JoelCourtney: Yeah that's a valid concern. This class is generally a mess because of its dual meaning (it represents all known data about an activity, combining directive and/or instance). All of the fields except `type`, `startOffset` and `arguments` may or may not be valid depending on whether the activity has been simulated, is an instance-only spawned activity, is anchored, etc. The id fields may be `null`. The boolean fields will always be defined but their values may not mean anything depending on context.

I don't like it at all. But it is so deeply coupled to the rest of the scheduler that changing it, even to use `Optional`s, would be a pretty big ordeal I think.

Issue #1073, opened 2023-08-08
Ability to "Group" Activities Linked via Scheduling


meganmjones:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

No - I haven't considered

### Related problems

This is somewhat related to the issue found with the activityFinder (https://github.com/NASA-AMMOS/aerie/issues/1061). If the ability to group scheduled activity instances and space them out from other distinct groupings was available this activityFinder bug would be a more minor problem.

### Describe the feature request

When scheduling groupings of activity instances together it would be helpful to have the option to prevent that "group" from overlapping with other "groups" (via group mutexes or a new global scheduling condition). There are many instances where one scheduling goal sets up GNC turns and array articulations surrounding a primary instrument activity instance. When setting these up in a larger plan, in many cases we want these groupings to not overlap with other goal's groupings that involved GNC/array articulations. 

Currently, mutexes can be written to prevent overlap of the GNC and array activity types, but it would be beneficial to have the capability to write mutexes preventing overlap of scheduled groupings of activity instances relative to one another. Similarly, being able to set spacing preferences between scheduled groupings would aide significantly in planning (i.e. allow 30 minutes on either end of X grouping from other groupings). This could potentially be handled with a global scheduling condition and would help break up spacing of linked activities in the plan.

A workaround that was put in place to build Clipper's Thread Test 6 plan was to shift the applyWhen window for select anchor activity placements to X hrs after a notable event (in TT6 the >2AU transition) instead of leaving the goal with simply the true constraint (>2AU). This was to space out the scheduled activity groupings to fall in a more logical ordering along with buffer time in between groups.

jmdelfa:
Slides with design approach for groups [Grouping.pptx](https://github.com/NASA-AMMOS/aerie/files/13384548/Grouping.pptx)

Need to discuss:
- How to store compound activities. Involvement from DB recommended
- How to enable users to re-use compound activities. Involvement from user interface team recommended

dandelany:
@jmdelfa and others met yesterday to discuss the details of this + HTNs, detailed notes in Slack here:

https://jpl.slack.com/archives/CJ07FGRPH/p1701910742191329

dandelany:
This is related to #1364 and we'll discuss more next week to scope both of them together.

dandelany:
Also related to older ticket: #870

jmdelfa:
Updated slides
[...]

Issue #870, opened 2023-04-13
Allow groupings of sequential activities in eDSL


anselrd:
One common use case for clipper is multiple activities that are performed in serial. It is difficult to accomplish this with the current set of scheduling goals, but is fundamentally a simple operation. Please add a capability to treat a specified sequence of activities like an activity, so things can be scheduled relative to it and it can be scheduled with recurrence and cardinality rules. Additionally, there should be a mechanism for providing relative timing within the sequence to encode wait times between activities.

The project should have the flexibility to encode these types of rules in scheduling rather than as decompositions because having too many decompositions clutters the timeline and obfuscates the lineage between activities because there are too many relationships.

PR #1510, opened 2024-07-19
Refactor RecurrenceGoal conflict production and solving

### LINKED ISSUE #1135: Activity recurrence goals allow too large an interval
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

1.13 - commit 67800ea

### Describe the bug

Activity recurrence goals allow for consecutive activities that are up to twice the interval size apart when checking for conflicts.

### Reproduction

Load the banananation model in a local simulation, and create a one-day plan with it, starting exactly at midnight for ease of seeing the timing window boundaries.

Add the following scheduling goal:
```
export default (): Goal => {
    activityTemplate: ActivityTemplates.BiteBanana({
      biteSize: 0.2
    }),
    interval: Temporal.Duration.from({ hours: 1 })
  })
}
```

Run scheduling. You should see a BiteBanana activity scheduled every hour, on the hour.

Delete one of the activities. Now re-run scheduling or analyze goals. Both report that the scheduling goal is fully satisfied.
image
This is wrong according to the documentation, as there's a two-hour gap between activities, twice the required interval. It's consistent with the interpretation that "every 1-hour slice has at least 1 activity", if slices include their endpoints.

Now move the activity after the one you deleted forward by any amount. Analyze scheduling goals, and it reports a failure:
image

Re-run scheduling goals, and it inserts two new activities:
image
This is consistent with the documentation, but it's not consistent with the alternate interpretation proposed above.

### Logs

_No response_

### System Info

```shell
Running Aerie locally
```

### Severity

Moderate

### Workaround

Manually add missing activity

adrienmaillard:
* **Tickets addressed:** Resolves #1135 
* **Review:** By commit  

## Description

See the ticket for the original issue. The cause for this issue is that 
- `RecurrenceGoal` produces all the conflicts at once
- each conflict has no relationship with the other. 

Because of this independence between conflicts, we had to find a way to specify temporal constraints for each conflict. What is done is that if the recurrence interval is 1 hour, each conflict is being given a 1-hour period to insert an activity. But you can see how this can result in 2 contiguous activities being separated by more than 1 hour (the first is scheduled at the start of its interval and the other at the end of its interval (because of a mutex possibly)). 

I tried 2 approaches: 
1. going back to an iterative conflict production that was ditched less than a year ago (I'd say) in which for each goal we call `getConflicts` not only before but also after solving to see if new conflicts are discovered. I combined this with a change to `RecurrenceGoal` so that it produces only one conflict at a time. Because conflicts would be produced one at a time, it would take into account the last activity insertion. Unfortunately, it required adding a specific block for `RecurrenceGoal` in the solving loop and I had to update the definition of `RecurrenceGoal` at each iteration so it would not stay stuck in case one of the activity insertion fails. All of this just looked like a workaround for a bigger issue.
2. creating a new type of conflict that takes relationships between subsequent activities into account. Instead of multiple 1-activity template conflicts, we have one big `MissingNetwork` conflict that contains a list of activities and their temporal constraints encoded as an STN. To solve this conflict, we go through each activity in the list, build a `MissingActivityTemplateConflict` with the most up-to-date STN temporal constraints and give that to the regular solver for that type of conflict. Then, we update the STN bounds with the start/end times of the inserted activity, we propagate and we find the bounds for the next activity. And so on and so on.  

I ended up with 2 as this is more elegant and also will be useful to instantiate task networks in #1073 #870.

I am also updating how the API of this goal to include an explicit temporal flexibility definition. Instead of just the `every(Duration)`, I am adding `separatedByNoMoreThan(Duration)` and `separatedByNoLessThan(Duration)`. The first is the one leading to conflict production. Which means that there is no penalty by having smaller separations in existing plans but... when we create new activities we use the second operator to determine the minimum space between activities.

I have refactored the `satisfyGoalGeneral` method of `PrioritySolver` to put the business logic of solving each type of conflict in its own method. It's just easier to read. And I needed to use one for this new work.

I have also refactored the satisfaction concept in the scheduler so it's not just a ill-defined number. 

## Verification

- [x] fix the failing tests (in most of them, one too many activity is added to the plan)
- [x] add at least one test reproducing the original issue 

## Documentation

- [x] update the documentation for the new behavior/operators

## Future work

None.


adrienmaillard:
I faced some bugs that made me reconsider the semantics of the goal. @srschaffJPL helped me reflect on this subject, here is a summary. 

## Alternative (easier) way of solving this issue 
When the user writes `.every(1hour)`, we don't only allow activities to start at times 0, 1, 2, 3 but we allow a default/implicit flexibility (to account for mutex/state constraints) of 1 hour. In other words we allow the first activity to start in [0, 1[, the second in [1, 2[ and so on. Which can cause activities to be separated by more than 1 hour.

We could be explicit about this flexibility with additional operators which would have the user write something like 
```
.every(1 hour)
.plus(30 minutes)
.minus(10 minutes)
```
We would not have to depart from the current approach of independent conflict and activities could still be separated by more than 1 hour but it would be the user's problem, not ours. Note that the scheduler now schedules activities at their earliest start time.   
 
Important point: in Clipper recurrence goals are mostly concerned by enforcing that 2 activities are not separated by more than some duration. So the strictness is important which would go against this alternative way of solving the issue. 

## Making `MissingNetwork` conflicts
Now let's come back to the `MissingNetwork` solution. One issue I was struggling with were the difference between activity and absolute timepoint boundaries.

There is one basic use case that is recurrence(activity1, activity2) that arises when there is too much time passing between the starts of activity1 and activity2. In this case, we can put constraints between missing template activities in the STN and existing activities. In other words we can bound the space between all activities. This is the easiest case.

The other basic use case that I’ll note recurrence(timepoint1, timepoint2) is the one when the plan is empty and the timepoints correspond to the start and end of the planning horizon. In this case, 
- what is the constraint between timepoint1 and the first template activity ? Should we consider that there exists a satisfying activity before timepoint1 (before the planning horizon start sometimes) ? 
- what is the constraint between the last template activity and timepoint2 ?  and so on.
 
What got out of the conversation is that we should let the users tell us where the activity (it’s fake though, there is no activity) before timepoint1 is, which would say when to actually start the serie. It would look like that
```
.separatedByNoLessThan(1)
.separatedByNoMoreThan(3)
.lastActOccurred(1)
```
The default value for this parameter could be the negated value of `separatedByNoMoreThan`, here -3 then. This would allow to conserve the usual behavior that have activities start at timepoint1 (0 if planning horizon start). 

As for the timepoint2, we would not put a specific distance constraint but more of a separation constraint between the last template activity and this absolute boundary. The reason for that is that we care about what happened before the plan but not about what happens after. 

## What happens when one of the template in a `MissingNetwork` cannot be scheduled ? 

Sometimes, a constraint or the mission model will make it impossible to schedule one of the activities in the “chain”, let’s say it’s activity i. That is a problem because the start bounds of activities i+1…k depend on the actual start times of activity i (they are being refined with propagation). So it seems we have to fix the STN to continue solving if we do not want to have a degraded solution. Note that this does not happen currently because each activity template is independent from others (which contributes to create the issue we are trying to solve…). 

Some ways of fixing it : 
1. transitive closures: we get the windows in which the activity has been scheduled.  We act like activity has been scheduled at its earliest start time, propagate and log that it has failed for later report
2. relax constraint for activity i and try again 
3. insert absolute timepoints and reconnect following activities in the graph

From the point of view of the user, what is the best behavior is something fails in the middle ? 
- fit an activity somehow even with looser constraints ?  
- drop the activity ?

adrienmaillard:
Screenshot 2024-07-25 at 2 37 39 PM

dandelany:
Discussed this with @adrienmaillard and @mattdailis earlier this week - looking good except it needs a rebase & potential conflict resolution.

Also @adrienmaillard when you get a chance - please write a couple sentences for the Upgrade Guide when this is released, to advise users on the impact of the change. Thanks!

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityNetworkConflict.java
```diff
@@ -0,0 +1,41 @@
...
+    return temporalContext;
+  }
+
+  public record ActivityDef(ActivityExpression activityExpression){}
+  public TaskNetworkAdapter temporalConstraints;
+  public Map<String, ActivityDef> arguments;
+  public List<String> schedulingOrder;
+  public Windows temporalContext;
+  public List<SimpleMissingActivityConflict> staticActivityConflicts;
+  public MissingActivityNetworkConflict(
```
JoelCourtney, line 24: formatting looks a bit weird, please add a newline

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/ConflictSolverResult.java
```diff
@@ -0,0 +1,36 @@
+package gov.nasa.jpl.aerie.scheduler.solver;
+import gov.nasa.jpl.aerie.scheduler.model.SchedulingActivityDirective;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class ConflictSolverResult {
+  private ConflictSatisfaction satisfaction;
+  private Set<SchedulingActivityDirective> activitiesCreated;
```
JoelCourtney: Since I don't see an `addActivity` method similar to `setSatisfaction`, I'm just checking that the activities set is never modified after creation. As in, you don't do something like:

```java
final var activities = new HashSet<...>();
final var result = new ConflictSolverResult(sat, activities);
activities.add(new ...);
```

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
```diff
@@ -81,21 +117,27 @@ protected RecurrenceGoal fill(RecurrenceGoal goal) {
       //first fill in any general specifiers from parents
       super.fill(goal);
 
-      if (every == null) {
-        throw new IllegalArgumentException("creating recurrence goal requires non-null \"every\" duration interval");
+      if (separatedByAtMost == null) {
```
JoelCourtney, line 113: I think if `separatedByAtLeast` is null it should be replaced with `Duration.ZERO`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -671,7 +664,8 @@ private ConflictSolverResult solveActivityNetworkConflict(
                                                                                       1,
                                                                                       Optional.empty(),
                                                                                       Optional.of(true),
-                                                                                      Optional.empty());
+                                                                                      Optional.empty(),
+                                                                                      ScheduleAt.LATEST);
```
JoelCourtney: I think that `ScheduleAt.LATEST` should be chosen by Recurrence Goal, and not be mandatory for all network conflicts.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityTemplateConflict.java
```diff
@@ -75,6 +80,11 @@ public Optional<Duration> getTotalDuration(){
     return totalDuration;
   }
 
+  @Override
+  public ScheduleAt scheduleAtEarliest() {
```
JoelCourtney: I think these methods should just be called `scheduleAt()` because it is returning an enum not a boolean.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -1135,25 +1130,36 @@ private Windows narrowGlobalConstraints(
       final MissingActivityTemplateConflict missingConflict,
       final String name,
       final Windows windows,
-      final EvaluationEnvironment evaluationEnvironment) throws SchedulingInterruptedException
+      final EvaluationEnvironment evaluationEnvironment,
+      final ScheduleAt chooseEarliestStart) throws SchedulingInterruptedException
```
JoelCourtney: Please also rename this, because its an enum not a boolean.

JoelCourtney (commented):
~Did you make changes so that recurrence goal would add more activities if the minimum wasn't enough? I can't find it.~

Nevermind, I found it.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -513,6 +521,176 @@ private void satisfyGoalGeneral(Goal goal) throws SchedulingInterruptedException
...
+        } else {
+          plan.remove(conflictResult.activitiesCreated());
+          if(hasDuplicates(conflictResult.activitiesCreated().stream().toList())){
+            //duplicates means it is unsolvable and it is no use increasing the number of activities
+            break;
+          }
+          if(conflictResult.satisfaction().equals(ConflictSatisfaction.PARTIALLY_SAT)){
+            lastPartiallySat = conflictResult.activitiesCreated();
+          }
+        }
```
JoelCourtney: If I'm interpreting this right, it looks like if solving the conflict with `n` activities fails, it starts over and tries again with `n+1`, instead of trying to continue with the existing activities. So it will potentially trigger another `n` duplicate simulations, with a worst case of `N^2` total simulations, where `N` is the final number of activities that actually succeeds.

JoelCourtney: Or actually it would start over and try again with `n+step` activities, but `step` is based on `maxIterations`, which is never set, so `step` will always be `1`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityNetworkConflict.java
adrienmaillard: where exactly ? I can't see the weirdness

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/ConflictSolverResult.java
adrienmaillard: It's not happening.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
adrienmaillard: You're right. Refactored so it uses previously inserted activities.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/RecurrenceGoal.java
adrienmaillard: OK.

adrienmaillard:
I have made the changes you mentioned.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/conflicts/MissingActivityNetworkConflict.java
JoelCourtney: you already fixed it :)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -513,6 +521,199 @@ private void satisfyGoalGeneral(Goal goal) throws SchedulingInterruptedException
...
+          default:
+        }
+        if(hasDuplicates(conflictResult.activitiesCreated().stream().toList())){
+          //duplicates means it is unsolvable and it is no use increasing the number of activities
+          break;
+        }
+      } else {
+        conflictProductionHasNotProgressedSince++;
+      }
+      if(nbActivities < maxNbActivities && conflictProductionHasNotProgressedSince < maxNoProgress){
```
JoelCourtney, line 559: Does `makeActivityNetworkConflict` use a simulation, or do some other expensive operation? If not, I don't see the purpose of using `conflictProductionHasNotProgressedSince` to limit the number of times it can try to create the network.

adrienmaillard: no expensive operation but it could loop a lot (when the user is setting the min separation to 0, the number of iterations is going to be `separatedByAtMost.in(MICROSECONDS)`) if `makeActivityNetworkConflict` is failing repeateadly. At some point, we know it won't progress anymore.

JoelCourtney:
Is there any check to make sure `separatedByAtLeast > separatedByAtMost`?

adrienmaillard:
I have just added a check in `RecurrenceGoal`. And I think it's separatedByAtLeast < separatedByAtMost :)

JoelCourtney (approved):

adrienmaillard:
Upgrade note for @dandelany 

RecurrenceGoal changes

The behavior of the goal has been changed when setting its `interval` parameter
When setting the `interval` parameter to 1 hour, the old behavior would create 1-hour bins for the duration of the horizon (or the more restrictive applyWhen period) and place 1 activity in each of these bins, whenever it is possible considering global scheduling conditions.  
With the new behavior, the `interval` parameter means a strict 1-hour distance between consecutive activities.  

New parameters have been introduced to give more control over the behavior of the goal and how it is solved:
- A user can now give lower (`separatedByAtLeast`) and upper bounds (`separatedByAtMost`)  on the separation between activities. Only the upper bound will lead to conflicts being created. In other words, having activities separated by less than `separatedByAtLeast`will not create a conflict. But, when a conflict exists, `separatedByAtLeast` will be used for solving.  
- As said above, the old `interval` parameter has been left but setting it to X is equivalent to setting both `separatedByAtMost` and `separatedByAtLeast` to X. 
- The user can either use the `interval` parameter or the 2 new bounds parameters.
- Another new parameter `previousActivityStartedAt` allows the user to specify when the “last” activity happened. It can potentially be a negative duration if it is before the start of the planning horizon. If not set, this parameter is set to `-separatedByAtMost`. 

The solver will now try several combinations to solve this goal, starting with the least number of activities possible (separated by a greater distance), and increase the number of activities until either all the temporal constraints are satisfied or it reaches a dead-end.

PR #1533, opened 2024-08-15
Use MissionModelId Class instead of String

### LINKED ISSUE #1063: Unify how MissionModelIds are parsed in Scheduler and Merlin Bindings
In the Scheduler, `MissionModelIds` are parsed using `missionModelP`, which is a wrapper around a `longP`. In Merlin, `MissionModelIds` are parsed as Strings. Additionally, the Hasura Actions use the ID type for MissionModelIds rather than Int. Merlin needs to be updated to match how the Scheduler handles MissionModelIds.

Mythicaeda:
* **Tickets addressed:** Closes #1063
* **Review:** By commit  

## Description

[Related Frontend PR](https://github.com/NASA-AMMOS/aerie-ui/pull/1427).

Updates the Merlin part of the codebase to use a numerical MissionModelId record instead of a raw String to represent model ids. This makes it consistent with the rest of the codebase and the database.

Breaking change as this replaces the `ID!` in `actions.graphql` with `Int!`. This change may impact users of the following actions:
- `getModelEffectiveArguments`
- `getActivityEffectiveArgumentsBulk`
- `constraintsDslTypescript`
- `validateActivityArguments`
- `validateModelArguments`

Users will be impacted only if they were using a query variable for the `missionModelId` argument of these actions. If they were, this can be fixed by changing the type of that query variable from `ID!` to `Int!`

Two deprecated actions have also been impacted. `getActivityEffectiveArguments` and `resourceTypes`. Use `getActivityEffectiveArgumentsBulk` in place of `getActivityEffectiveArguments`. Query the `resource_type` table in place of the `resourceTypes` action.

## Verification

Unit tests were updated to reflect the change in type.

## Documentation

[Docs PR](https://github.com/NASA-AMMOS/aerie-docs/pull/181)


goetzrrGit (approved):
I am not sure if I can approve this PR since I helped work on this, but it all looks good from my point of view.

dandelany:
👍 looked at this with @Mythicaeda on a call also, thanks!

Issue #1506, opened 2024-07-18
Stateless Aerie - Scheduling


dandelany:
### Description

Per discussions with @ewferg, some users would like an officially-supported way to use Aerie's simulation and scheduling engines as a simpler "stateless" and "headless" program.
- **Stateless** means the user runs Aerie simulation/scheduling with some inputs, Aerie exits and returns some output, but nothing is persisted in the Aerie app/DB beyond that point - as opposed to our current standard Aerie deployment, which is a persistent web service
- **Headless** means without a UI, ie. likely a CLI which reads from `stdin` and writes results to `stdout` when complete.

### Requirements 
We don't yet have all of the requirements or implementation fully planned out yet, but I wanted to open this to track our progress. I will meet with @ewferg soon to discuss - please comment here or in Slack if you're interested in participating. As a starting point, here's a **rough draft of some requirements**:

1. Users should be able to run Aerie's **simulation engine alone** in a stateless way, by running a command and providing two inputs: a mission model JAR file, and an Aerie plan file containing activity directives. Aerie should run the simulation, output any useful results (grounded activity instances, profiles, other TBD?) and exit.
2. Users should also be able to run **scheduling alone** statelessly, by running a command and providing a mission model, an Aerie plan, and **scheduling goals** as inputs. Aerie should execute the scheduling goals (possibly running more simulations in the process), output useful results, and exit.
3. For a reasonably small model/plan/set of goals (to be defined), stateless simulation should be able to run in < 10 seconds, and stateless scheduling should run in < 30 seconds, to be competitive with existing tools that users are using today instead of Aerie.

### Notes & Questions
* Our first goal is to determine how feasible this is given our current architecture, & the level of effort it will take. Once the requirements are further fleshed out, we'll meet with devs to come up with a proposed implementation.
* Can stateless sim/scheduling be done *without* running our (hasura/postgres) database service, since nothing is being persisted, or is the DB too tightly coupled to Aerie sim/scheduling for this to work? The performance requirements will be difficult to meet if we have to wait for Hasura's significant "spin-up" time.
* Stateless scheduling will be required to support procedural scheduling JARs. Does it also need to support EDSL goals/interleaving like the UI does?
* Is there any overlap here with `aerie-cli` functionality, and should we consider taking ownership of `aerie-cli` and implementing this feature there, or is this a part of Aerie core, or a new tool altogether that interfaces with Aerie core?

ewferg:
[...]

PR #1536, opened 2024-08-20
Stateless Aerie - Simulation

Mythicaeda:
* **Tickets addressed:** Addresses simulation portion of #1506 
* **Review:** By commit  

## Description

Adds three packages:
- `orchestration`: package containing classes useful for orchestrating aerie (ie, calling simulation, loading a mission model)
- `stateless-aerie`: package containing a Java CLI that runs aerie in a stateless mode (compared to the Python CLI, which runs against a venue)
- `type-utils`: package containing types commonly used by various parts of the system (ie Plan, MissionModelId). Created as a first step towards not requiring so many parts of the system to import other packages in order to use common types: ie `merlin_driver`. @mattdailis I'd like your thoughts on this package's existence/contents.

## Verification

Unit tests were added that the orchestration works as expected. Reference JSONs of simulation results for tests were manually validated.

## Documentation

Docs need to be created for the new JSON schemas: Sim Configs, Sim Results.

## Future work

Integration with Procedural Scheduling.
Further work to extract more common types into `type-utils` and common parsers into a `parsing-utils`-type package.


merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
```diff
@@ -496,16 +527,15 @@ public boolean isZero() {
...
   }
 
   /** Determine whether two durations are the same. */
   @Override
   @Deprecated
   public boolean equals(final Object o) {
-    if (!(o instanceof Duration)) return false;
-    final var other = (Duration)o;
-
+    if (!(o instanceof final Duration other)) return false;
```
goetzrrGit, line 548: non-blocker do we have another method we can use if this one is `deprecated`?

stateless-aerie/src/main/java/gov/nasa/jpl/aerie/stateless/simulation/SimulationExtentConsumer.java
```diff
@@ -0,0 +1,52 @@
...
+      @Override
+      public void run() {
+        // Only print if simulation time has progressed.
+        if(!lastAcceptedDuration.isEqualTo(lastReportedDuration)) {
+          System.out.println("Current simulation time: " + lastAcceptedDuration);
+          lastReportedDuration = lastAcceptedDuration;
+        }
+      }
+    };
+    printTimer.scheduleAtFixedRate(timerTask, periodMillis, periodMillis);
```
goetzrrGit: Non-blocker: You can inline this if you want

```java
printTimer.scheduleAtFixedRate(new TimerTask() {
      @Override
      public void run() {
        // Only print if simulation time has progressed.
        if(!lastAcceptedDuration.isEqualTo(lastReportedDuration)) {
          System.out.println("Current simulation time: " + lastAcceptedDuration);
          lastReportedDuration = lastAcceptedDuration;
        }
      }
    }, periodMillis, periodMillis)
```

stateless-aerie/src/main/java/gov/nasa/jpl/aerie/stateless/simulation/SimulationResultsWriter.java
```diff
@@ -0,0 +1,431 @@
...
+  realProfiles: [
+   {
+     name: string
+     schema: ValueSchema
+     segments: [
+       extent: Duration
+       dynamics: {}//arbitrary value based on schema
+     ]
+   }
+  ],
```
goetzrrGit: non-blocker: We can probably get rid of this comment unless you want to keep it around as a reference.

goetzrrGit (approved):
Some non-blocker comments. You mentioned in the standup 'type-utils' needs to be discussed on further. I am not sure if you were going to wait on that meeting or remove the commit from this PR and merged?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
Mythicaeda: `Duration.isEqualTo`, which this method calls. This is an override of a default Object method, though, so we can't remove it or the `deprecated` tag.

stateless-aerie/src/main/java/gov/nasa/jpl/aerie/stateless/simulation/SimulationResultsWriter.java
Mythicaeda: I'd like to keep it around until we decide somewhere better to document this, though I agree that I'd prefer to do that before merging this PR.

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/types/Duration.java
JoelCourtney: FYI in procedural scheduling branch I have converted `Duration` to a record. So I deleted this function, because it is automatically implemented (and not deprecated).

mattdailis (approved):
Nice work!

stateless-aerie/src/main/java/gov/nasa/jpl/aerie/stateless/simulation/SimulationExtentConsumer.java
Mythicaeda: Done!

PR #1539, opened 2024-08-26
fix for infinite loop when time range is less than 20 micros

bradNASA:
* **Tickets addressed:** none
* **Review:** By commit 
* **Merge strategy:** Merge 

## Description
Infinite loop when searching in a time range of micros less than the max iterations where no solutions exist

## Verification
ran unit tests, and it fixed this bug occurring on another branch

## Documentation
none

## Future work
add a test case that searches within a <20 microsecond window

### DIFF
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
@@ -152,7 +152,7 @@ public class EquationSolvingAlgorithms {
     /**
      * Addition of longs with safety from overflow
      */
-    private static long plus( long rd1, long rd2 ) {
+    private static long plus(final long rd1, final long rd2 ) {
       long result;
       // check for overflow
       if ( rd1 >= 0 && Long.MAX_VALUE - rd1 <= rd2 ) {
@@ -168,7 +168,7 @@ public class EquationSolvingAlgorithms {
     /**
      * Subtraction of longs with safety from overflow
      */
-    private static long minus( long rd1, long rd2 ) {
+    private static long minus(final long rd1, final long rd2 ) {
       long result;
       // use plus, but don't risk trying to negate +/-inf
       if ( rd2 == Long.MAX_VALUE ) result = plus( rd1, Long.MIN_VALUE );
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
@@ -127,6 +127,9 @@ public class EquationSolvingAlgorithms {
         final int maxIteration)
     throws ExceededMaxIterationException, SchedulingInterruptedException
     {
+      // the number of possible values may be less than the number of iterations, so stop after all have been visited.
+      long numTimepoints = max.in(Duration.MICROSECONDS) - min.in(Duration.MICROSECONDS) - 1;
+      long maxIters = Long.min(maxIteration, numTimepoints);
       var cur = init;
       int i = 0;
       do {
@@ -141,8 +144,8 @@ public class EquationSolvingAlgorithms {
           }
         }
         cur = chooseRandomX(min, max);
-        //if min == max, another call to random will have no effect and thus we should exit
-      } while(i < maxIteration && !min.isEqualTo(max));
+        //if all timepoints have been visited or min == max, another call to random will have no effect and thus we should exit
+      } while(i < maxIters);
       throw new ExceededMaxIterationException();
     }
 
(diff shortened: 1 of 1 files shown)


scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -141,8 +144,8 @@ private IteratingResult nextValueAt(
           }
         }
         cur = chooseRandomX(min, max);
-        //if min == max, another call to random will have no effect and thus we should exit
-      } while(i < maxIteration && !min.isEqualTo(max));
+        //if all timepoints have been visited or min == max, another call to random will have no effect and thus we should exit
+      } while(i < maxIters);
```
adrienmaillard, line 148: why removing the second `!min.isEqualTo(max)` ?

adrienmaillard, line 148: I don't see how this leads to an infinite loop in the case you describe ? Can you elaborate either on the PR description or in an answer to that comment ?

adrienmaillard (commented):
Some questions first

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
bradNASA: if min == max, `numTimepoints` will be -1, so `maxIters` will be -1

bradNASA: I came across a run that called this method with min = 4 micros and max = 6 micros, so only 4 or 5 would be randomly encountered.  At some point, the `if` statement always fails, and `i` is never incremented, so the loop never ends.

adrienmaillard: can you then remove the "or min == max" part of the comment ?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -127,6 +127,9 @@ private IteratingResult nextValueAt(
         final int maxIteration)
     throws ExceededMaxIterationException, SchedulingInterruptedException
     {
+      // the number of possible values may be less than the number of iterations, so stop after all have been visited.
+      long numTimepoints = max.in(Duration.MICROSECONDS) - min.in(Duration.MICROSECONDS) - 1;
```
adrienmaillard: this is a numeric overflow if `max` is `Long.MAX_VALUE` and `min` is `Long.MIN_VALUE`.

adrienmaillard: maybe use `BigInteger` ?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
adrienmaillard: OK! it sounds like the increment of `i` is at fault here but that's a reasonable workaround.

bradNASA: removed

bradNASA: `i` was only incremented for values that weren't already visited.  So, I don't think always incrementing `i` is a good fix.  It think this fix is appropriate.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
bradNASA: I pasted in some code for handling overflow.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/EquationSolvingAlgorithms.java
```diff
@@ -127,25 +127,56 @@ private IteratingResult nextValueAt(
...
-      } while(i < maxIteration && !min.isEqualTo(max));
+        //if all timepoints have been visited, another call to random will have no effect and thus we should exit
+      } while(i < maxIters);
       throw new ExceededMaxIterationException();
     }
 
+    /**
+     * Addition of longs with safety from overflow
+     */
+    private static long plus( long rd1, long rd2 ) {
```
adrienmaillard: can you final rd1 and rd2 ? same in the minus.

bradNASA: "final" modifiers added to plus() and minus()

dandelany (approved):
Thanks Brad - all issues above resolved & tests passing 👍 I'll approve and merge this on @adrienmaillard 's behalf since he's out this week.

Issue #1538, opened 2024-08-22
Process + workflows for testing UI PRs against specified backend branch


dandelany:
## Background

We don't have a good process today for dealing with changes/features that require *both* an `aerie-ui` frontend PR and an `aerie` backend PR (or `aerie-gateway` PR). When the UI PR is submitted, the e2e test suite will fail because it is run against the `develop` version of the backend (ie. it uses Docker images tagged `develop`) which does not have the changes from the backend feature branch.

The only ways to work around this today are:
* Manually checkout the backend + UI branches locally - this works but isn't reflected in the PR's automated CI checks so you can't tell when it's safe to merge.
* Change the `docker-compose-test.yaml` on the UI branch to point to a different Docker tag temporarily - works, but then the PR is not safe to merge until you switch it back - otherwise would change the test branch for everyone on merge.
* Merge the backend PR first, then re-run checks on the frontend PR - works but leaves the system in an unsafe "halfway" state in the meantime until UI PR is merged. The combination of `develop` backend + `develop` UI should always work together.

### Proposed Process

After discussing with a few devs, here's what I'm proposing as the process for this scenario:

* When the backend PR is submitted (or later), you tag it with a GH label called "**publish**". 
  - This causes a GH workflow to publish docker images for the PR branch, tagged with the branch name, and re-publish them anytime the PR is updated with new commits.
  - This could be done automatically for **all** PRs, but we worry this could create a very large number of published images, so @Mythicaeda suggested we do it selectively.
* When the frontend PR is submitted (or later), you add a specific string to the body of the PR that looks like `___AERIE_IMAGE_TAG___="name-of-my-backend-branch"`
  - When the UI e2e Test workflow is run, it looks for a string like this in the PR body and sets the branch name in an environment variable if so
  - The env var is passed to the `docker-compose-test.yaml` & used as the tag name for the `aerie` images during e2e testing
  - Similarly, to specify an `aerie-gateway` branch you would put `___GATEWAY_IMAGE_TAG___="name-of-my-gateway-branch"`

If this is done correctly, all test suites should be able to pass on the frontend PR. You will still have to merge the backend PR before the frontend PR or else tests will fail in the merged `develop` branch but at least this avoids the unsafe workarounds mentioned above.

dandelany:
Document this process in Developer/contributing docs...

Mythicaeda:
The Gateway needs to additionally be able to pull in a Backend image, as was added to the UI.

[...]

PR #1540, opened 2024-08-27
GH workflow to publish images for PRs labeled 'publish'

dandelany:
Supporting this issue: https://github.com/NASA-AMMOS/aerie/issues/1538

This change will cause the GH "Publish" workflow to run whenever an `aerie` PR is given the "publish" label, which will cause Docker images to be published from the PR branch, tagged with the PR number eg. `pr-1540`

The published aerie PR can then be referenced in an `aerie-ui` PR with a flag in the PR body that looks like eg. `___REQUIRES_AERIE_PR___="1540"`, which will cause the `aerie-ui` e2e tests to use the published `pr-1540` aerie images during testing rather than the default `develop` branch.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -7,18 +7,28 @@ on:
       - dev-[0-9]+.[0-9]+.[0-9]+
     tags:
       - v*
+  pull_request:
+    types:
+      # publish images for PRs labeled "publish" whenever changed/labeled
+      [opened, reopened, synchronize, labeled]
   workflow_dispatch:
 
 env:
   REGISTRY: ghcr.io
   OWNER: nasa-ammos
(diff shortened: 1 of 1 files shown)


Mythicaeda:
In order to test the `synchronization` event, could you flatten the working commits and force push?

skovati (commented):
Will removing the `publish` tag on this PR remove the failed check and allow the PR to be merged?

Mythicaeda:
The PR isn't actually blocked on the failure, it's blocked on a lack of an approval. The reason it isn't blocked is because the `Publish` workflow isn't a required workflow, unlike our `test` workflow.

dandelany:
> In order to test the `synchronization` event, could you flatten the working commits and force push?

Done! will do the same for the other 2 PRs.

Mythicaeda (approved):

PR #1541, opened 2024-08-30
Support events emitted from daemon tasks

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Tweaks the handling of event provenance to account for events emitted from daemon tasks to avoid NullPointerException.

All credit to @mattdailis for this fix.

## Verification
I did some manual testing with a local Aerie instance, on a model that used the streamline Logger to emit events from daemon tasks, and verified that the simulation completes without error and the events appear in the simulation events table in the UI.

## Documentation
N/A - bug fix

## Future work
N/A


mattdailis (approved):
Nice catch!

PR #1559, opened 2024-09-12
Generator Constraints and Violation Messages

JoelCourtney:
* **Tickets addressed:**
* **Review:** By commit  

## Description
This adds the `GeneratorConstraint` abstract class, which is a helper class that enables a more imperative programing style in constraints. When writing goal I've found it pretty natural to switch from a functional style when getting and transforming inputs, to an imperative style when creating outputs. So I made a similar system for constraints where you can repeatedly call `violate(Violation)` or one of its derivatives instead of aggregating everything in a single `Violations` timeline.

This also adds the first step for support of violation messages, as in customizable strings that can explain the violation to the user in the UI, and can be different for each violation generated. This doesn't do anything with the messages, it just stores them in the `Violation` class.

## Verification
I made a unit test to make sure the generator aggregates the results correctly.

## Documentation
I wrote the docs after I made these changes; they are actually already live.


mattdailis (approved):
I like the new imperative interface - and also that it augments rather than replaces the functional one 💯

JoelCourtney:
Docs update at https://github.com/NASA-AMMOS/aerie-docs/pull/187

PR #1560, opened 2024-09-13
External profiles in procedural scheduling

JoelCourtney:
* **Tickets addressed:** probably should have made a ticket
* **Review:** By commit  

## Description
This PR adds a `resource(...)` method to the `Plan` object, with the exact same semantics as `simResults.resource(...)`. I put it on `Plan`, because it is statically known data that doesn't change with simulation.

## Verification

This PR also refactors some e2e tests to make a dedicated package for procedural scheduling, including building goals inside the e2e package. And it adds a couple tests for both the AeriePostgresPlan object and the main scheduler backend.

## Documentation
[docs pr](https://github.com/NASA-AMMOS/aerie-docs/pull/188)


e2e-tests/build.gradle
```diff
@@ -54,12 +57,17 @@ task e2eTest(type: Test) {
 }
 
 dependencies {
+  testAnnotationProcessor project(':procedural:processor')
```
skovati: We might want to keep these procedure definitions outside of the e2e-tests submodule, similar to how example mission models that are used in our tests live in the `examples` directory (since they are useful for more than just e2e tests). Having them in this submodule feels a bit "overloaded" in my opinion

.github/workflows/test.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: Assemble
         run: ./gradlew assemble --parallel
       - name: Build scheduling procedure jars for testing
-        run: ./gradlew procedural:examples:foo-procedures:buildAllSchedulingProcedureJars
+        run: ./gradlew e2e-test:build && ./gradlew e2e-tests:buildAllSchedulingProcedureJars
```
skovati: I believe we're already running the annotation processing on these example procedures in the `./gradlew assemble --parallel` step above

```suggestion
        run: ./gradlew e2e-tests:buildAllSchedulingProcedureJars
```

e2e-tests/build.gradle
JoelCourtney: I actually don't like calling these examples. They are a little too simplistic and unrealistic, and I wrote them for testing anyway. These are basically just assets for the e2e tests, and its normal to put test assets in the same package.

.github/workflows/test.yml
JoelCourtney: done

e2e-tests/build.gradle
skovati: That's fair, although I do think it would be fruitful to surface some useful example procedures in the top level `examples` folder eventually, but maybe that can result from our eventual monorepo w/ example mission models / schedulers in the future. Certainly not a blocker for this PR, happy to table it.

skovati (approved):
🚀

PR #1568, opened 2024-09-25
Save activity names from procedural goals

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  

## Description
This PR saves activity names set by procedural goals to the database.

## Verification

I added an e2e test to make sure the names successfully made it to the db.

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
@@ -1156,5 +1156,22 @@ public class SchedulingTests {
 
       assertEquals(52, activities.size());
     }
+
+    /**
+     * Run a spec with one procedure and make sure the activity names are saved to the database
+     */
+    @Test
+    void saveActivityName() throws IOException {
+      final var args = Json.createObjectBuilder().add("quantity", 2).build();
+      hasura.updateSchedulingSpecGoalArguments(procedureId.invocationId(), args);
+
+      final var resp = hasura.awaitScheduling(specId);
+
+      final var plan = hasura.getPlan(planId);
+      final var activities = plan.activityDirectives();
+
+      assertEquals(2, activities.size());
+      assertEquals("It's a bite banana activity", activities.getFirst().name());
+    }
   }
 }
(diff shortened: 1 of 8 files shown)


e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/SchedulingTests.java
skovati, line 1: Will unnamed activies get the "Unnamed Activity" `name` stored to the DB? We might want to add a test to ensure the behavior is as expected

procedural/examples/foo-procedures/src/main/java/gov/nasa/ammos/aerie/procedural/examples/fooprocedures/procedures/SampleProcedure.java
```diff
@@ -19,9 +21,12 @@ public void run(@NotNull final EditablePlan plan) {
     var currentTime = firstTime;
     for (var i = 0; i < quantity; i++) {
       plan.create(
-          "BiteBanana",
-          new DirectiveStart.Absolute(currentTime),
-          Map.of()
+          new NewDirective(
```
skovati, line 24: Is naming an activity common enough that we should provide a `plan.create` overload?

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/SchedulingActivity.java
```diff
@@ -86,12 +88,13 @@ public static SchedulingActivity of(
...
         anchoredToStart,
-        isNew
+        isNew,
+        null
     );
   }
 
   public SchedulingActivity withNewDuration(Duration duration){
-    return SchedulingActivity.of(
+    return new SchedulingActivity(
```
skovati, line 97: Previous question applies here too, if naming an activity is common enough, maybe we want to overload `SchedulingActivity.of` instead of forcing the use of `new SchedulingActivity`, if that is the main pattern in the scheduler code.

procedural/examples/foo-procedures/src/main/java/gov/nasa/ammos/aerie/procedural/examples/fooprocedures/procedures/SampleProcedure.java
```diff
@@ -19,9 +21,12 @@ public void run(@NotNull final EditablePlan plan) {
...
+              new AnyDirective(Map.of()),
+              "It's a bite banana activity",
```
skovati, line 26: Make sure to add this name field insertion to your new procedures in your e2e test improvement branch 👍

skovati (approved):
LGTM, just a few non-blocking comments

procedural/examples/foo-procedures/src/main/java/gov/nasa/ammos/aerie/procedural/examples/fooprocedures/procedures/SampleProcedure.java
JoelCourtney: I don't think so

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/SchedulingActivity.java
JoelCourtney: I found the many overloads of `SchedulingActivity.of` to be impossible to understand and keep track of, which is why I've deleted most of them. Since java doesn't have keyword args, and many of the arguments are nullable, its hard to tell what each argument is. I definitely don't want to add another overload, even if it was common.

dandelany:
Merging for release (with permission from @JoelCourtney) - thanks!

PR #1569, opened 2024-09-25
Generate JVM default methods for procedural interfaces

JoelCourtney:
* **Tickets addressed:** None
* **Review:** By commit  

## Description
For some reason Kotlin doesn't automatically generate JVM interface default methods by default, and you have to enable it with a compiler flag. In a few interfaces, I've provided a couple default overload specializations of a method (i.e. `plan.directives()` and `directives(type)` delegate to `directives(type, deserializer)`). If you implement the interface in Kotlin, you don't have to override the overloads. But in Java you do, unless we have this flag.

## Verification
I tested this by converting the `NotImplementedPlan` test stub from Kotlin to Java.


skovati (approved):
LGTM, looks like they are tracking this (very long term) issue

https://youtrack.jetbrains.com/issue/KT-4779/Generate-default-methods-for-implementations-in-interfaces

PR #1574, opened 2024-10-02
NPE fix

bradNASA:
NPE fix, untested

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java
@@ -843,7 +843,7 @@ public final class SimulationEngine implements AutoCloseable {
                 .stream()
                 .map(spanToActivityInstanceId::get)
                 .toList(),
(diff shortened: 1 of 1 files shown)


mattdailis (approved):
Thank for the fix, I've run into this a number of times myself. This can occur if a daemon task spawns an activity, and that activity doesn't finish by the end of simulation

mattdailis:
Discussed with Theresa, and we decided to refactor these two lines of code:

https://github.com/NASA-AMMOS/aerie/blob/ac8720e5a10a391373724ead9033d01d226a5bf7/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SimulationEngine.java#L831

To this:
```diff
+(activityParents.containsKey(span)) ? Optional.empty() : Optional.ofNullable(directiveId)
-Optional.ofNullable(directiveId)
```

The ternary dates back to a time when we used the absence of a parent span to determine whether a span is "top level", and assumed that all top level spans came from the plan. Now, we have a more direct way to compute directive ids, which is robust to the situation where a parent-less span comes from a daemon task rather than a directive in the plan.

dandelany:
Thanks all! Merging this fix.

PR #1513, opened 2024-07-23
Add External Events

pranav-super:
* **Tickets addressed:** MPS-118
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description
A new construct in AERIE is that of External Events. In this PR, we will introduce External Events, their containing entities (External Sources), their grouping mechanisms (Derivation Groups), as well as how they manifest in the UI and in the backend. Detail on what External Events are is provided in the [aerie-ui](https://github.com/NASA-AMMOS/aerie-ui/pull/1396) PR, but as a quick summary:

> An external event is a construct that represents the occurrence of something outside of what is modeled in the mission model. They could be events that are close to the spacecraft and directly affect behavior on it, like a DSN visibility event that might affect when communications happen, or they could be much more remote, like a constellation event. Functionally, these are at the interstice between external datasets and activities, in that they manifest on the timeline as activities (discrete blocks of time), but they themselves do not affect the mission model or invoke any actions much like external datasets. 

One component of that discussion that _is_ deferred to this description, as it is here that it is primarily handled, is that of derivation. We discuss that in detail here.

### Derivation

Derivation was an operation conceived for the reconciliation of overlapping External Sources. The problem at hand was that it is entirely possible for sources to be generated at later points in time, that cover slightly different windows on plan time than their predecessors. 

image

If this is the case, some reconciliation scheme is needed to determine which External Events from these External Sources are valid and should be visible on the timeline.

#### Reconiciliation Mechanism
To begin this discussion, we should first talk about when External Sources apply.

External Sources, as discussed earlier, have within them a `valid_at` field, which was described as a more particular form of a version number. With this, we know we can impose an ordering of External Sources. Given that ordering, figuring out when each External Source actually applies is as simple as determining for a given time (or slot of time), what the most recently valid External Source is. For example, the following set of 4 External Sources would be applicable in the following order. Note that one of the Exteranl Sources, `B`, is valid twice, as it is the most recently valid External Source for a given slot of time:

image

Given this, we can now discuss how to compare the External Events within each External Source, based on how they fall across these windows/against future External Sources. There are 4 rules (though the first two are effectively cases of each other):
1. _An External Event superceded by nothing will be present in the final, derived result._ The logic behind this is that this event has no future data that might challenge its validity, so it reasonably would trickle down as an event in a final result.
[...]

PR #1585, opened 2024-10-23
Schedule Activities Based On External Events

pranav-super:
* **Tickets addressed:** MPS-124
* **Review:** By commit

## Description
Following the introduction of External Events in [#1513](https://github.com/NASA-AMMOS/aerie/pull/1513), this PR introduces the ability to schedule Activities based off of External Event occurrences, using the procedural scheduler.

This PR introduces 4 new types which users can use in their definition of scheduling procedures to narrow down the external events they would like to schedule off of and which properties they would like to access:
- an `ExternalEvent` type, which contains event-specific properties,
- an `ExternalEvents` type, which represents a filterable collection of said events,
- an `EventQuery` type, which can be used to filter the set of all `ExternalEvent`s associated with a given plan by their `ExternalSource`, derivation group, or event type.
- and an `ExternalSource` type, which can be used by `EventQuery`s to narrow down the set of events being examined by a procedure to those contained in specific sources.

To make this more explicitly clear, we will illustrate an example,  based on the tests in `ExternalEventsTests.java` under the `e2e-tests` directory.

Consider this set of events:
image

Consider scheduling the Banananation activity `BiteBanana` off of these events. 
This can mean a variety of different things, depending on the exact behavior we desire:
- schedule the activity coincident with all events
- schedule the activity coincident with events belonging to derivation group `"TestGroup"`
- schedule the activity coincident with events belonging to the second source
- schedule the activity based on events of type `"Test_Type"`
- schedule the activity based on events with the substring `"01"` in their key.

For the sake of brevity, we will explore just one case - events belonging to the second source with the substring `"01"` in their key. This case makes use of all of the types described above.

```
@SchedulingProcedure
public record ExternalEventsSourceQueryGoal() implements Goal {
  @Override

    // extract events belonging to the second source
    EventQuery eventQuery = new EventQuery(
        null,
        null,
        List.of(new ExternalSource("NewTest.json", "TestGroup_2"))
    );

    for (final var e: plan.events(eventQuery)) {
      // filter events that we schedule off of by key
      if (e.key.contains("01")) {
        plan.create(
            "BiteBanana",
            // place the directive such that it is coincident with the event's start
            new DirectiveStart.Absolute(e.getInterval().start),
            Map.of("biteSize", SerializedValue.of(1)));
      }
    }
    plan.commit();
  }
}
```

This goal does the following:
1. create an `EventQuery` to select events belonging to the second source. To specify that selector, we create an `ExternalSource` object, whose properties are the two properties necessary to uniquely identify two `ExternalSource`s in AERIE: a `key` and a `derivation_group_name`. 
2. Call `plan.events(eventQuery)`. This returns all external events associated with a plan (based on the associated derivation groups), and filters them according to the passed-in query.
3. Filter by key. We directly access the properties of each event (namely, `key`) now to further match our requirements.
4. Create the activity, placing it coincident with the event's start, accessed via `getInterval().start`.

After running it, we get the following result in AERIE:
image

As expected, there is a single activity, scheduled off of an `Event_01` from the second source.

## Verification
The changes were verified both by testing with our own models, and subsequent testing of the 4 new types listed above in a few end-to-end tests (see `ExternalEventsTests.java` under the `e2e-tests` directory).

## Documentation
Documentation will need to be added to [aerie-docs](https://vcs.ssmo.gsfc.nasa.gov/SSMO-MPS/aerie-docs/tree/master/docs) to describe how scheduling using external events work. As this feature is solely an addition, no existing documentation is invalidated; the documentation is simply expanded.

## Future work
The next steps for External Event work are:
- additional cleanup and optimization of External Events in the front and backend
- adding properties to events and sources, and exposing those to the scheduler
- integration with the procedural constraint system
- introduction of a formal JSON schema for events and sources and support of said schema in the database

### DIFF
--- a/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventsSourceQueryGoal.java
+++ b/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventsSourceQueryGoal.java
@@ -1,0 +1,39 @@
+package gov.nasa.jpl.aerie.e2e.procedural.scheduling.procedures;
+
+import gov.nasa.ammos.aerie.procedural.scheduling.Goal;
+import gov.nasa.ammos.aerie.procedural.scheduling.annotations.SchedulingProcedure;
+import gov.nasa.ammos.aerie.procedural.scheduling.plan.EditablePlan;
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.ExternalSource;
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.DirectiveStart;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.EventQuery;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.Map;
+
+@SchedulingProcedure
+public record ExternalEventsSourceQueryGoal() implements Goal {
+  @Override
+  public void run(@NotNull final EditablePlan plan) {
+
+    // extract events belonging to the second source
+    EventQuery eventQuery = new EventQuery(
+        null,
+        null,
+        List.of(new ExternalSource("NewTest.json", "TestGroup_2"))
+    );
+
+    for (final var e: plan.events(eventQuery)) {
+      // filter events that we schedule off of by key
+      if (e.key.contains("01")) {
+        plan.create(
+            "BiteBanana",
+            // place the directive such that it is coincident with the event's start
+            new DirectiveStart.Absolute(e.getInterval().start),
+            Map.of("biteSize", SerializedValue.of(1)));
+      }
+    }
+    plan.commit();
+  }
+}
--- a/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/ExternalEvents.kt
+++ b/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/ExternalEvents.kt
@@ -1,0 +1,31 @@
+package gov.nasa.ammos.aerie.procedural.timeline.collections
+
+import gov.nasa.ammos.aerie.procedural.timeline.Timeline
+import gov.nasa.ammos.aerie.procedural.timeline.BaseTimeline
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.Instance
+import gov.nasa.ammos.aerie.procedural.timeline.ops.*
+import gov.nasa.ammos.aerie.procedural.timeline.ops.coalesce.CoalesceNoOp
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.ExternalEvent
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.ExternalSource
+import gov.nasa.ammos.aerie.procedural.timeline.util.preprocessList
+
+/**
+ * A timeline of external events.
+ */
+data class ExternalEvents(private val timeline: Timeline<ExternalEvent, ExternalEvents>):
+    Timeline<ExternalEvent, ExternalEvents> by timeline,
+    NonZeroDurationOps<ExternalEvent, ExternalEvents>,
+    ParallelOps<ExternalEvent, ExternalEvents>
+{
+  constructor(vararg events: ExternalEvent): this(events.asList())
+  constructor(events: List<ExternalEvent>): this(BaseTimeline(::ExternalEvents, preprocessList(events, null)))
+
+  /** Filter by one or more types. */
+  fun filterByType(vararg types: String) = filter { it.type in types }
+
+  /** Filter by one or more event sources. */
+  fun filterBySource(vararg sources: ExternalSource) = filter { it.source in sources }
+
+  /** Filter by one or more derivation groups. */
+  fun filterByDerivationGroup(vararg groups: String) = filter { it.source.derivationGroup in groups }
+}
--- a/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
+++ b/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
@@ -7,6 +7,7 @@ import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.AnyDirective
 import gov.nasa.ammos.aerie.procedural.timeline.collections.Directives
 import gov.nasa.ammos.aerie.procedural.timeline.ops.SerialSegmentOps
 import gov.nasa.ammos.aerie.procedural.timeline.payloads.Segment
+import gov.nasa.ammos.aerie.procedural.timeline.collections.ExternalEvents
 import java.time.Instant
 
 /** An interface for querying plan information and simulation results. */
@@ -41,4 +42,9 @@ interface Plan {
    * @param name string name of the resource
    */
   fun <V: Any, TL: SerialSegmentOps<V, TL>> resource(name: String, deserializer: (List<Segment<SerializedValue>>) -> TL): TL
+
+  /** Get external events associated with this plan. */
+  fun events(query: EventQuery): ExternalEvents
+  /** Get all external events across all derivation groups associated with this plan. */
+  fun events() = events(EventQuery())
 }
--- a/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
+++ b/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
@@ -46,7 +46,6 @@ import java.util.stream.Collectors;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.MICROSECOND;
 import static gov.nasa.jpl.aerie.merlin.protocol.types.Duration.ZERO;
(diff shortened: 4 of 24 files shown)


JoelCourtney:
@pranav-super you forgot to push the change to GraphQLMerlinDatabaseService

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/solver/PrioritySolver.java
```diff
@@ -609,7 +617,7 @@ private Optional<MissingActivityNetworkConflict> makeActivityNetworkConflict(
         curPreviouslyInstantiated = null;
       }
       taskNetwork.startsAfterStart(
-          allActivitiesInNetwork.getLast(),
```
JoelCourtney: why did you change this?

pranav-super: IntelliJ was getting very upset with me using `getLast()` and `getFirst()`, saying it wasn't supported on `List`. Checking the `List` interface supported this assertion, as did [this](https://stackoverflow.com/questions/1049518/why-does-the-java-list-interface-not-support-getlast). I'm pretty skeptical but this was the only way I could compile and get things to run!

JoelCourtney: It is supported in Java 21, so I think the problem must be in your setup, not the code. Please revert this.

On Mac in IntelliJ, you can go to `File > Project Structure > Project > SDK` to change your jdk version.

JoelCourtney:
Rebased to clean up the git history

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
```diff
@@ -41,4 +42,13 @@ interface Plan {
    * @param name string name of the resource
    */
   fun <V: Any, TL: SerialSegmentOps<V, TL>> resource(name: String, deserializer: (List<Segment<SerializedValue>>) -> TL): TL
+
+  /** Get external events associated with this plan. */
+  fun events(query: EventQuery): ExternalEvents
```
skovati, line 47: I'm not quite seeing the purpose of `EventQuery`, since we already provide filtering capabilities in these other event function overloads. Why not just have a function in the `Plan` interface here with nullable arguments (`derivationGroup: String?, eventType: String?, source: String?`) that conditionally handles the filtering?

Not a big deal to me, but it does seem a little confusing that we're exposing multiple filtering APIs to users, especially when they could do this filtering in "userspace" anyways.

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/ExternalEvents.kt
```diff
@@ -0,0 +1,31 @@
...
+data class ExternalEvents(private val timeline: Timeline<ExternalEvent, ExternalEvents>):
+    Timeline<ExternalEvent, ExternalEvents> by timeline,
+    NonZeroDurationOps<ExternalEvent, ExternalEvents>,
+    ParallelOps<ExternalEvent, ExternalEvents>
+{
+  constructor(vararg events: ExternalEvent): this(events.asList())
+  constructor(events: List<ExternalEvent>): this(BaseTimeline(::ExternalEvents, preprocessList(events, null)))
+
+  /** Filter by one or more types. */
+  fun filterByType(vararg types: String) = filter { it.type in types }
```
skovati, line 24: This looks like yet another filtering API we're exposing to users. When would we expect users to use these functions v.s. querying while getting the `ExternalEvents` object?

skovati (changes requested):
Looks great so far! Nice work. Just have a few questions about streamlining the filtering API we're exposing to users to simplify things.

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
pranav-super: The only thing this provides in addition is the ability to specify multiple filters in one single query, without needing multiple filtering function calls. They are be functionally equivalent, but, at least in testing of the feature, this query did seem to be pretty helpful. 

@JoelCourtney Any thoughts on this? Should we remove either the `EventQuery` or the other function options?

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/ExternalEvents.kt
pranav-super: If there are conditional filters we want to apply on an `ExternalEvents` object, this is how we could do it without having to pull repeatedly from the `Plan` object.

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
skovati: Gotcha, I just worry about exposing too many filtering objects to end users. It looks like we can filter when requesting events in two ways:
1. `val events = plan.events("SomeGroupName")`
2. `val events = plan.events(new EventQuery("SomeGroupName"))`

And once we have an `ExternalEvents` object, we can filter in two ways:
1. `val justFooType = events.filterByType("Foo")`
2. `val justFooType = events.filter { it.type == "Foo }`

Like you said, all are functionally equivalent, but I can imagine a user being a bit confused when they see multiple ways to filter events without realizing they're functionally equivalent.

Now (while introducing this feature) is a good time to think about the most clear / concise API we can expose, since future changes would (likely?) be breaking.

pranav-super: Sure! I think allowing filtering from the `Plan` and filtering from the `ExternalEvents` objects should both be available, but I agree we should limit each to only 1 method.

pranav-super: Actually wait - I think I change my mind. I think we should actually allow all of these. The point of the methods on `Plan`, despite the fact that they can be replicated in `EventQuery`, is to provide idiomatic ways of filtering events from a plan without needing to specify `EventQuery`. We will add more to events (like metadata) in the future and as such having both an easy idiomatic way as well as a more complicated, query way, to fetch events I would argue is helpful. At the current moment there is complete overlap but going forward I don't think this will be the case!

skovati: Gotcha! That's why I had originally asked about the point of `EventQuery` since it (currently) feels like an unnecessary layer of indirection around simple filtering logic, but if you see the `EventQuery` object encapsulating logic around more complex filters (e.g. some sort of metadata many-to-one mapping or something), then I'm fine keeping it in.

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventsSourceQueryGoal.java
```diff
@@ -0,0 +1,39 @@
...
+@SchedulingProcedure
+public record ExternalEventsSourceQueryGoal() implements Goal {
+  @Override
+  public void run(@NotNull final EditablePlan plan) {
+
+    // extract events belonging to the second source
+    EventQuery eventQuery = new EventQuery(
+        null,
+        null,
+        List.of(new ExternalSource("NewTest.json", "TestGroup_2"))
```
skovati, line 24: Is it safe to assume that users will know external event source file names _a priori_? Or would it be helpful to include some way to query for which sources had been registered with a plan?

I'm thinking of cases where source file names might be autogenerated or have some unknown timestamp, and users might just want to reference events where `source.includes("DSN")` or something.

While typing this out, I'm realizing that particular use case could be accomplished by "manually" filtering on event sources after we've run the `EventQuery` against the `Plan` 🤔

pranav-super: Not necessarily a safe assumption; as such it is possible to allow users to filter for that kind of thing: 
```
var events = plan.events();
events.filter(false, event -> event.source.key.contains("DSN"));
```

skovati: My thoughts exactly!

skovati (approved):

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/Plan.kt
JoelCourtney: Actually EventQuery and the other `events` overloads could provide some extra utility, and the explicit filter functions are just sugar. I made EventQuery to give the Plan implementation the option to optimize the events database lookup so that it only loads the events you need (although the built in scheduler doesn't actually do this).

I think there are situations that are best suited to each of the three ways of filtering, but I agree that having three ways is kinda overkill. So I'd be happy removing the `events(derivGroup)` and `events(group, type)` overloads, since I think they're the least useful.

pranav-super: These are the ones defined as methods on `Plan`, right?
EDIT: they have been factored out.

Issue #1563, opened 2024-09-18
Fix sim results staleness check in Procedural Scheduling


dandelany:
## Description

The `SimulationResults` interface in the `aerie.procedural.timeline.plan` package has a function [`isStale()`](https://github.com/NASA-AMMOS/aerie/blob/9af4b3c2abe43dccebe1c4dbd64194be024fe224/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/SimulationResults.kt#L13), which is supposed to return `true` if the sim results are stale/out-of-date compared to what's in the plan - ie. because the plan has changed and the simulation needs to be re-run - or `false` if they're up to date.

The current implementation has a bug which causes `isStale()` to always return `false` (ie. not stale), which will cause problems for goals relying on this flag to know when the sim needs to be re-run. 

@JoelCourtney is investigating. The fix is a bit more complicated than expected - due to the fact that scheduling goals may drop & re-create plans multiple times, it's not quite as simple as just checking the revision of the plan against the revision stored on the `SimResults` object.

dandelany:
Closed with #1595

PR #1595, opened 2024-11-08
Simulation result staleness checking

JoelCourtney:
* **Tickets addressed:** #1563
* **Review:** By commit  

## Description
This PR implements a staleness check for the internal procedural scheduling implementations. The editable plan instance keeps track of sim results that it has produced using weak references, and can dynamically update their staleness if the plan is changed after it was simulated. The process is this:

1. InMemoryEditablePlan has a set of weak references to simulation results objects that are currently up-to-date. I used weak references because if the user can't access it anymore, staleness doesn't matter and we might as well let it get gc'ed.
2. When the user gets simulation results, either through simulation or by getting the latest, it always checks for plan equality between the returned results and the current plan, even if we just simulated. If it is up-to-date, a weak ref is added to the set.
3. When an edit is made, the sim results in the current set are marked stale; then the set is reset to new reference to an empty set.
4. When a commit is made, the commit object takes *shared ownership* of the set. If a new simulation is run (step 2) the plan can still add to the set while it is still jointly owned by the commit. Then when an edit is made (step 3) the commit will become the sole owner of the set.
5. When changes are rolled back, any sim results currently in the plan's set are marked stale, the previous commit's sim results are marked not stale, then the plan will resume joint ownership of the previous commit's set.

The joint ownership freaks me out a wee bit, but I think it's safe because the commits are only used to keep the previous sets from getting gc'ed in the event of a rollback. Only the plan object actually mutates the set.

## Verification

I added some unit tests to the scheduler-driver.

## Documentation
This doesn't need to be explained to the user, but I've copied the above description into the doc comment on `InMemoryEditablePlan`.

## Future work
- The stateless scheduler will probably need something like this too.
- I'm expecting that when I implement deletion, we will get false positives of staleness, because I've intentionally stayed away from constantly performing plan equality checks and storing a bunch of copies of the plan. Maybe that was premature optimization. The false positive would happen like this:
  1. simulate
  2. add activity; sim results marked stale
  3. delete activity (not rollback); sim results not marked un-stale.
  4. sim results now say they are stale when they actually aren't.


scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
```diff
@@ -11,14 +11,41 @@ import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.AnyDirective
...
+ *
+ * The editable plan instance keeps track of sim results that it has produced using weak references, and can dynamically
+ * update their staleness if the plan is changed after it was simulated. The process is this:
+ *
+ * 1. [InMemoryEditablePlan] has a set of weak references to simulation results objects that are currently up-to-date.
+ *    I used weak references because if the user can't access it anymore, staleness doesn't matter and we might as well
+ *    let it get gc'ed.
+ * 2. When the user gets simulation results, either through simulation or by getting the latest, it always checks for
+ *    plan equality between the returned results and the current plan, even if we just simulated. If it is up-to-date, a
+ *    weak ref is added to the set.
```
Mythicaeda, line 37: ~~Does this mean that the user could repeatedly add to this set by, ie, getting the latest results in a loop (without doing anything else in the loop)?~~

I forgot how Sets work for a second 🤦‍♀️. This question is irrelevant as long as the _same_ `WeakRef` is returned each time.

JoelCourtney: I actually don't know, it depends on if the simulation facade returns the exact same pointer, rather than equivalent data, when the user simulates the same plan multiple times. And I don't know which one it does.

But since they are weak references, they will not prevent the sim results from being garbage collected when the user code loses all strong references to it. So it doesn't actually matter; the set could contain a thousand weak references to the same Sim Results instance, or it could contain a thousand references to different instances with the same data. But since they are all weak references, the memory usage will be the same as before this PR.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
```diff
@@ -11,14 +11,41 @@ import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.AnyDirective
...
+ * 3. When an edit is made, the sim results in the current set are marked stale; then the set is reset to new reference
+ *    to an empty set.
+ * 4. When a commit is made, the commit object takes *shared ownership* of the set. If a new simulation is run (step 2)
+ *    the plan can still add to the set while it is still jointly owned by the commit. Then when an edit is made (step 3)
+ *    the commit will become the sole owner of the set.
+ * 5. When changes are rolled back, any sim results currently in the plan's set are marked stale, the previous commit's
+ *    sim results are marked not stale, then the plan will resume joint ownership of the previous commit's set.
+ *
+ * The joint ownership freaks me out a wee bit, but I think it's safe because the commits are only used to keep the
+ * previous sets from getting gc'ed in the event of a rollback. Only the plan object actually mutates the set.
```
Mythicaeda, line 47: Are users meant to see this comment (as in, are they expected to be making/managing `InMemoryEditablePlan`s? If so, the language across this comment should be formalized, this paragraph especially.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
```diff
@@ -34,9 +61,19 @@ data class InMemoryEditablePlan(
   val totalDiff: List<Edit>
     get() = commits.flatMap { it.diff }
 
-  override fun latestResults() =
-    simulationFacade.latestSimulationData.getOrNull()
-      ?.let { MerlinToProcedureSimulationResultsAdapter(it.driverResults, false, plan) }
+  // Jointly owned set of up-to-date simulation results. See class-level comment for algorithm explanation.
+  private var simResultsUpToDate: MutableSet<WeakReference<MerlinToProcedureSimulationResultsAdapter>> = mutableSetOf()
```
Mythicaeda: Please rename this variable, the current name makes me think it's a `boolean` and not a `set`. This also applies to the field in `Commit.kt`

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
```diff
@@ -34,9 +61,19 @@ data class InMemoryEditablePlan(
...
+
+  override fun latestResults(): SimulationResults? {
+    val merlinResults = simulationFacade.latestSimulationData.getOrNull() ?: return null
+
+    // kotlin checks structural equality by default, not referential equality.
+    val isStale = merlinResults.plan.activities != plan.activities
+
+    val results = MerlinToProcedureSimulationResultsAdapter(merlinResults.driverResults, isStale, plan)
+    if (!isStale) simResultsUpToDate.add(WeakReference(results))
+    return results
```
Mythicaeda, line 88: Why are the results still returned if they're stale? Doesn't that mean that the user of this function has to be aware of and check the staleness to know if they can use `latestResults`?

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
```diff
@@ -55,16 +92,33 @@ data class InMemoryEditablePlan(
...
+    // because new sets are only created when edits are made.
+    // Probably unnecessary, but shared ownership freaks me out enough already.
+    if (uncommittedChanges.isEmpty()) return
+
     val committedEdits = uncommittedChanges
     uncommittedChanges = mutableListOf()
-    commits.add(Commit(committedEdits))
+
+    // Create a commit that shares ownership of the simResults set.
+    commits.add(Commit(committedEdits, simResultsUpToDate))
```
Mythicaeda: Idle question: Why is the entire history of commits saved as opposed to just the most recent commit and a `totalDiff` `List<Edit>`?

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
Mythicaeda: > I actually don't know, it depends on if the simulation facade returns the exact same pointer, rather than equivalent data, when the user simulates the same plan multiple times. And I don't know which one it does.

While I see logic to avoid a resim if the plan matches the initial results, I see no such logic for anything past that. It looks like at best, the facade will recalculate a new Results object with the same data, and at worst it will resim the portion of the plan between the Checkpoint and the end time (Which is a minor regression. Near as I can tell, the reason we aren't noticing this regression is because our test for this behavior (`noNeedForResimulation`) doesn't actually check that any of the objects are the same, just that the results are _equivalent_. But the engine still goes all the way into the sim loop and just happens to stop immediately because the checkpoint was already after the end of the requested horizon).

> But since they are all weak references, the memory usage will be the same as before this PR.

Understood, although could you quickly go over why we need a _Set_ of up to date results for a commit? Is it because we don't know _which_ sim results the user still has a strong reference to?

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: Not really. It won't be included in the procedural scheduling docs because its in the scheduler driver. But I can make it a regular comment anyway.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: The point of `latestResults` is to get the most recent results regardless of whether they're up to date. The user does need to be informed if its stale or not though.

If they want to guarantee they get up to date results, they should call simulate. The sim facade is responsible for making sure that it doesn't resimulate if the latest results are already up-to-date.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: This PR doesn't change any of the behavior relating to when the sim facade does or doesn't simulate, or whether the sim facade returns a cached results object if simulated twice. That is up to the sim facade and its behavior should be exactly the same after this change.

> why we need a Set of up to date results for a commit?

The user will probably only simulate at most once for any given plan state, but we don't know that for sure. The user can simulate as much as they want, and the sim facade can return whatever it wants, and this PR treats both of them as black boxes. So there might be multiple equivalent sim results objects floating around. Its unlikely but I wanted to be safe.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: I suppose there's no good reason anymore. It's leftover from an earlier version of this class. I think I can rework it pretty easily.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: Renamed to `upToDateSimResultsSet`

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/InMemoryEditablePlan.kt
JoelCourtney: Ok I've reworked it to flatten the committed changes into one list of Edits instead of keeping a list of Commits. I'm still using the `Commit` object though just to keep the committed changes and up-to-date sim results for that commit associated together; but there's now only one Commit object at a time.

Mythicaeda (approved):
LGTM! Does moving `Commit` inside of the `InMemoryEditablePlan` help ease any anxieties about the shared ownership?

JoelCourtney:
> Does moving `Commit` inside of the `InMemoryEditablePlan` help ease any anxieties about the shared ownership?

Nope. I just did it because the Commit class isn't used anywhere else, so I didn't want to make it look like it was part of a public interface.

PR #1596, opened 2024-11-08
Procedural Constraints Action/Database Support

### LINKED ISSUE #1572: DB updates for Procedural Constraints
### Background

Now that [Procedural Scheduling](https://nasa-ammos.github.io/aerie-docs/scheduling-and-constraints/procedural/introduction/) has been released, we are working on implementing [Procedural Constraints](https://nasa-ammos.github.io/aerie-docs/scheduling-and-constraints/procedural/constraints/). The workflow for using them will be very similar to scheduling goals: write a constraint in Java, compile a JAR, upload it via the Aerie UI, and add it to a plan using the same list UI that is used for eDSL constraints. Whereas procedural scheduling **goals** create **activities** on an editable plan, procedural **constraints** will return **violations**, ie. windows of time when the constraint is violated, and will not be able to edit the plan directly.

Implementing procedural **goals** required some changes to the database schema, to support storing them in the database and associating them with a plan:

[deployment/hasura/migrations/Aerie/10_procedural_scheduling/up.sql](https://github.com/NASA-AMMOS/aerie/blob/develop/deployment/hasura/migrations/Aerie/10_procedural_scheduling/up.sql)

In particular, since multiple invocations of the *same goal* are now allowed in a plan spec, we had to switch from a primary key of `goal_id` to `goal_invocation_id`, representing a single invocation of a goal.

### Requirements

* Define a new database schema & write a migration similar (nearly identical) to the one for procedural scheduling, to support procedural constraints
* Just like goals:
  * Multiple invocations of the same constraint will be allowed on a plan, necessitating a switch to `constraint_invocation_id`
  * Procedural constraints may have parameters/arguments
  * User should be able to have a mix of eDSL constraints & procedural constraints on their plan spec

### Related Issues
https://github.com/NASA-AMMOS/aerie/issues/1571, https://github.com/NASA-AMMOS/aerie-ui/issues/1478

### LINKED ISSUE #1633: Reordering Multiple Invocations of Scheduling Goals Fails
The update functions for `scheduling_specification_goals` and `scheduling_model_specification_goals` does not work with multiple invocations of the same goal.

If you have two instances of a goal with sequential priorities, then try to reorder them, the update fails.

This occurs because the priority-cleanup functions filter on `goal_id` and not `invocation_id`

skovati:
* **Tickets addressed:** Closes #1572, Closes #1633
* **Review:** By commit  

## Description

This PR adds support for Procedural Constraints in the Constraints Action and the Database.

Please disregard all commits prior to `Add sequencing build files to .gitignore`, those are from #1611 and should be reviewed in that PR.

Database Changes:
 - Adds invocation/arguments support to scheduling model specification
 - Fixes a bug with procedural constraints
 - Updates `constraint_definition` to support procedural JARs
 - Adds invocation/arguments/priority to constraints specification and constraint model specification
 - Replaces the current cache `constraint_run` table with a trio of tables to track constraint results and requests:
     - `constraint_request`: log an executed request. As we are not currently tracking multiple statuses for these requests, the request is only created when the rest of the results are posted.
     - `constraint_result`: A table containing the output of all constraints ran. Entries are deduplicated as the action inserts them (see `InsertConstraintResultsAction`), but this is not enforced in the database, as a duplicated result is not invalid, just not preferred. If a constraint has an error while running, those errors are now stored in an `errors` column, though results with errors will not be considered when checking for a cached result (see `GetValidConstraintRunsAction`)
     - `constraint_run`: An association table between `constraint_request` and `constraint_result`. Entries are one row per enabled constraint in the specification during the request. As Aerie does not consider `priority` as part of the inputs to the `result` or as a factor for cache staleness, it is noted on this table.

Backend Changes:
- General work to make the Constraints Action read in and run procedural constraints.
- Added an optional `force` flag to the Constraint Action. If set to `true`, the Action will skip loading results from the cache
- Creating a `ConstraintResult` interface with an implementation per results source: `EDSLConstraintResult`, `ProceduralConstraintResult`, `DBConstraintResult`
- Refactored `Fallible` to use a `SuccessType` and `FailureType`, instead of a single `?` type.
- Returns the generated request id as part of the action response (BREAKING CHANGE)

## Verification

E2E Tests have been updated, none have been added yet due to time.

## Documentation

[Some of the Constraints API docs have been invalidated](https://nasa-ammos.github.io/aerie-docs/api/examples/constraints/)

## Future work

- Change how procedural constraints get their resources in order to load them lazily, instead of preloading them prior to run time. If this is not feasible, alter the Constraint Action such that if there are _only_ EDSL constraints, it will lazy-load resources as it needs to the same way it did before, and if there are only Procedural constraints, avoid instantiating a Constraint representation of a SimulationResult, as that uses a separate set of objects for resources/activities. See [this comment](https://github.com/NASA-AMMOS/aerie/pull/1596#issuecomment-2761741368) for some explorations done by @mattdailis.
- We may want to pull the Constraint Action code out of the `merlin_server`.


skovati:
Active works in progress:
- [ ] database migrations
- [ ] comments for new tables
- [ ] verification of migrations against real data

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_run.sql
```diff
@@ -10,7 +12,7 @@ create table merlin.constraint_run (
   requested_at timestamptz not null default now(),
 
   constraint constraint_run_key
-    primary key (constraint_id, constraint_revision, simulation_dataset_id),
+    primary key (constraint_invocation_id, simulation_dataset_id),
```
Mythicaeda: Blocking: This pkey is not sufficient. If the user decides to check constraints twice, updating the definition of a constraint between checks, they will get a pkey conflict when the action attempts to cache the output of the second definition.

(This comment is here to prevent me from not fixing this)

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_request.sql
```diff
@@ -0,0 +1,37 @@
...
+  id integer not null generated always as identity
+    primary key,
+  plan_id integer not null,
+  simulation_dataset_id integer not null, -- "not null", as we reject rqs if there aren't up-to-date sim results
+                                          -- (or if the provided sim dataset id is invalid)
+  force_rerun boolean,
+
+  -- Additional Metadata
+  requested_by text,
+  requested_at timestamptz not null default now(),
```
mattdailis, line 11: Not sure if we typically comment these kinds of columns, but I noticed these ones are missing comments

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_results.sql
```diff
@@ -0,0 +1,45 @@
...
+
+  -- inputs
+  constraint_id integer not null,
+  constraint_revision integer not null,
+  simulation_dataset_id integer not null,
+  arguments jsonb not null,
+
+  -- outputs
+  results jsonb not null default '{}',
+  errors jsonb not null default '{}',
```
mattdailis, line 12: Just for my understanding - is the expectation that only one of `results` or `errors` will be populated for a given row? Would documenting the answer to that question in a brief comment in this sql file be worthwhile?

deployment/postgres-init-db/sql/tables/scheduler/scheduling_specification/scheduling_specification_goals.sql
```diff
@@ -98,34 +98,31 @@ create function scheduler.update_scheduling_specification_goal_func()
...
-    (select priority
-     from scheduler.scheduling_specification_goals ssg
-     where ssg.specification_id = new.specification_id
-     order by priority desc
-     limit 1), -1) + 1
-  into next_priority;
+  select count(goal_invocation_id) - 1
+  from scheduler.scheduling_specification_goals ssg
+  where ssg.specification_id = new.specification_id
+  into max_priority;
```
mattdailis, line 106: Looking at the name of the variable (`max_priority`), would this perhaps be a more direct way to express this query?

```suggestion
  select max(priority)
  from scheduler.scheduling_specification_goals ssg
  where ssg.specification_id = new.specification_id
  into max_priority;
```

mattdailis (commented):
I'm making slow, but steady progress - unfortunately I need to pause review for a few hours now, but I wanted to give you the comments I have so far

deployment/postgres-init-db/sql/tables/scheduler/scheduling_specification/scheduling_specification_goals.sql
Mythicaeda: Yeah, it is. I went with the `count` approach initially, as I wanted to bound `priority` not by its own value but by a property of the specification: that is: the `max_permitted_priority` (I dropped `permitted` when naming the variable) is the number of constraints/goals on the spec (minus 1 b/c priority is 0-based). I was also testing this against a DB where I had already let `priority` escape this upper bound, which meant I couldn't use it as the bounding column.

mattdailis: That's a good point - I poked around on aerie-dev and found cases where the bounds had been escaped. The questions that come to mind are:
1. Do we feel like we understand how that happened?
2. Is it a problem?
3. How does this PR change that story? Will existing scheduling specifications be affected?

Mythicaeda: 1. Yes, [I've explained it how it happens in the UI PR](https://github.com/NASA-AMMOS/aerie-ui/pull/1537#discussion_r2001567862)
2. It is not a problem with execution AFAIK
3. Currently, this PR isn't cleaning up existing specs (they're allowed to exist but will have issues changing the priorities unless they explicit set them to be back in-bounds), but I can add a step to the migration to fix the existing specs (keep the relative order but bring them back in-bounds)

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_results.sql
Mythicaeda: That's how the table is written to, and I'm trying to think of a case where you'd want to write to both. I hadn't considered it a restriction of the table, though, which is why there's no constraint or comment.

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_request.sql
Mythicaeda: These are usually commented, thanks for catching this!

deployment/postgres-init-db/sql/tables/scheduler/scheduling_specification/scheduling_specification_goals.sql
mattdailis: > Currently, this PR isn't cleaning up existing specs (they're allowed to exist but will have issues changing the priorities unless they explicit set them to be back in-bounds), but I can add a step to the migration to fix the existing specs (keep the relative order but bring them back in-bounds)

Got it - is it easy for a user to fix this issue? It seems like it could be rather disruptive if a user is prevented from doing their work due to our "bugfix".

Any of the following would be okay:
- A migration automatically fixes the priorities on existing specifications
- Existing specifications autofix when a user changes the priority of a goal
- We continue to permit specifications that don't strictly go from 0 to n - 1 (or is it 1 to n? I'm not sure)

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/SimulationResults.java
```diff
@@ -27,6 +32,39 @@ public SimulationResults(
     this.discreteProfiles = discreteProfiles;
   }
 
+  public SimulationResults(
```
mattdailis, line 35: Here's another piece of code that does the same thing - perhaps the scheduler can call this new constructor instead?

https://github.com/NASA-AMMOS/aerie/blob/f32a88973c70ab29a57dd6c6c03117864a50b440/scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/simulation/SimulationResultsConverter.java#L25

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ConstraintParsers.java
```diff
@@ -0,0 +1,34 @@
...
+import static gov.nasa.jpl.aerie.constraints.json.ConstraintParsers.violationP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.listP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.longP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.productP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
+import static gov.nasa.jpl.aerie.json.Uncurry.tuple;
+import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
+
+public final class ConstraintParsers {
+  public static final JsonParser<ProceduralConstraintResult> proceduralConstraintResultP =
```
mattdailis, line 18: There is an existing class called ConstraintParsers.java - is there a need for two of them? Does it have something to do with the fact that they're in different subprojects?

More specifically, I'm a bit worried about the duplication between `proceduralConstraintResultP` and `edslConstraintResultP` - if I understand correctly, the two are expected to have the same JSON representation, is that right?

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_results.sql
```diff
@@ -0,0 +1,45 @@
...
+  'The constraint that was checked.';
+comment on column merlin.constraint_results.constraint_revision is e''
+  'The version of the constraint definition that was checked.';
+comment on column merlin.constraint_results.simulation_dataset_id is e''
+  'The simulation dataset the constraint was checked against.';
+comment on column merlin.constraint_results.arguments is e''
+  'The arguments provided to the constraint when checked.';
+comment on column merlin.constraint_results.results is e''
+  'Results that were computed during the constraint check.';
+comment on column merlin.constraint_results.results is e''
```
mattdailis:
```suggestion
comment on column merlin.constraint_results.errors is e''
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ConstraintParsers.java
Mythicaeda: > Does it have something to do with the fact that they're in different subprojects?

Yes. There are two files in different subprojects are because the different representation of `ConstraintResults` live in different subprojects -- that is `EdslConstraintResult` is in the `constraints` package, while `ProceduralConstraintResult` is in the `merlin-server` package. `EdslConstraintResult` wants to reference its own parser for its `toJson()`, so I can't move the parser to the `merlin-server`

> the two are expected to have the same JSON representation, is that right?

For now, yeah. We may want to consider diverging them in the future because, for example, procedural constraints have no concept of gaps at all. (This is in contrast to how they _could_ have a concept of resources referenced, but we currently don't have that programmed)

I think the correct solution is likely to remove the `ResultP` parsers from `ConstraintParsers` and put them inside the specific class, similar to how `ConstraintsCompilationError` contains its own parser. Thoughts?

mattdailis: If the long term plan is for the two features to diverge into separate features, I think that makes sense :+1: My understanding, however, was that these are intended to be two different ways to write constraints, and that we may want the same downstream processes to handle violations from both kinds of constraints without needing to care which kind of constraint they came from. To me that implies that these two results should converge over time, rather than diverge. While gaps may not be straightforward to implement for procedural constraints, I don't think they're inherently inapplicable.

mattdailis: If the long term plan is for the two features to diverge into separate features, I think that makes sense :+1: My understanding, however, was that these are intended to be two different ways to write constraints, and that we may want the same downstream processes to handle violations from both kinds of constraints without needing to care which kind of constraint they came from. To me that implies that these two results should converge over time, rather than diverge.

Regarding gaps - while they may not be straightforward to implement for procedural constraints, I don't think they're inherently inapplicable. I'd be curious to hear @JoelCourtney 's thoughts here

Mythicaeda: > While gaps may not be straightforward to implement for procedural constraints, I don't think they're inherently inapplicable.

Breakdown of the situation that @JoelCourtney gave me is that in procedural scheduling, Violations aren't profiles, they're a collection of unordered parallel timelines that can freely overlap (as opposed to in the EDSL, where IIRR overlapping violations just become a single violation), so the concept of "gaps" doesn't make sense.  @JoelCourtney

> My understanding, however, was that these are intended to be two different ways to write constraints

My understanding is that EDSL constraints are going into LTS towards eventual deprecation with the release of Proc Constraints.

mattdailis: My (less technical) understanding of gaps is it's a mechanism for a constraint to provide feedback to the user that it was unable to conclude one way or the other whether it is satisfied in a given time range. That's pretty important - we had users chime in during design reviews asking for this feature when we were discussing a flight rule checking feature. I think the contents of constraint results are driven by the needs of the consumers rather than the implementations of the producers

Regarding the long term plan for EDSL constraints - I think that's contingent on the successful adoption of procedural constraints, so I wouldn't place any bets until we get some user feedback.

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_results.sql
Mythicaeda: Thanks! Done!

constraints/src/main/java/gov/nasa/jpl/aerie/constraints/model/SimulationResults.java
Mythicaeda: Yep! Done!

deployment/postgres-init-db/sql/tables/scheduler/scheduling_specification/scheduling_specification_goals.sql
Mythicaeda: I've done the first option, and "compressed" the priorities back in bounds during the migration (plan version shared below):

```sql
with priorities as (
  select
    goal_invocation_id,
    row_number() over (partition by specification_id) as new_prio
  from scheduler.scheduling_specification_goals ssg
  order by priority
)
update scheduler.scheduling_specification_goals ssg
set priority = p.new_prio - 1 -- -1, as priority starts at 0
from priorities p
where p.goal_invocation_id = ssg.goal_invocation_id;
```

deployment/postgres-init-db/sql/tables/merlin/constraints/constraint_request.sql
Mythicaeda: Fixed!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ConstraintParsers.java
JoelCourtney: I didn't implement gaps for violations for a couple reasons
- The fact that violations is a parallel timeline makes gaps hard to define and kinda unsound. Gaps are for things that are supposed to have continuous, unambiguous coverage of the timeline (like a profile). `Violations` should be thought of as "a collection of violation objects" instead of "a profile that indicates when the system is in a violating state". Gaps are only well-defined in the latter. We did it in the eDSL anyway because it was a very controlled environment, and because I hadn't thought it all the way through.
- We don't really want first-class support for gaps in violations anyway. What we really want is first-class support for *warnings*. Then, gaps in resource profiles used in the constraint can be just one of many possible sources of warnings. This is what I originally wanted in the eDSL, but the API would have been pretty heinous, so we only supported gaps. I think warnings will be a lot easier in procedural constraints, but not part of MVP.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -120,145 +121,126 @@ public Map<Constraint, Fallible<?>> getViolations(final PlanId planId, final Opt
...
+    } catch (MissionModelService.NoSuchMissionModelException | NoSuchPlanException ex) {
+      return Fallible.failure(
+          new ConstraintsDSLCompilationService.ConstraintsDSLCompilationResult.Error(
+              new ArrayList<>() {{
+                add(new ConstraintsCompilationError(
+                    "Constraint '" +constraint.name()+ "' compilation failed:\n " + ex.getMessage(),
+                    ex.toString(),
+                    new ConstraintsCompilationError.CodeLocation(0,0),
+                    ex.toString()));
+              }}
```
mattdailis: This is interesting! I'm unfamiliar with this syntax. I noticed you changed it below to `List.of()`, does it make sense to do that here too? What _does_ `new Foo() {{ ...code here... }}` do?

Mythicaeda: It's known as a "Double Brackets Initializer". From my understanding of the syntax, the first set of brackets creates an anonymous inner class (same as when you do `new Thread() {  public void run() {..}}`, and the inner set opens up the constructor for said anonymous class. 

I think the syntax is neat, but sonarQube doesn't like it, hence why I changed the lower instance to `List.of` after much deliberation (since the List isn't meant to be updated, `List.of` is an acceptable substitute). I can change this one too (or change the lower one to go back to a DBI)

mattdailis: Fascinating - that's good to know. I think I agree with SonarQube on this one

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanConstraintsAction.java
```diff
@@ -49,17 +81,32 @@ public Optional<List<ConstraintRecord>> get(final long planId) throws SQLExcepti
       do {
         if (results.getObject(1) == null) continue;
 
+        final var typeString =  results.getString("type");
+        final ConstraintType type;
+        switch (typeString) {
+          case "EDSL" -> type = new ConstraintType.EDSL(results.getString("definition"));
+          case "JAR" -> type = new ConstraintType.JAR(Path.of("/usr/src/app/merlin_file_store", results.getString("path")));
```
mattdailis: I believe we can get this path from the `MERLIN_LOCAL_STORE` environment variable

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -34,74 +31,78 @@ public ConstraintAction(
...
+                                                + planId.id()
+                                                + " has not yet been simulated at its current revision"));
     }
 
-    final var constraintCode = new HashMap<>(this.planService.getConstraintsForPlan(planId));
-    final var constraintResultMap = new HashMap<Constraint, Fallible<?>>();
+    final SimulationDatasetId simDatasetId = resultsHandle.getSimulationDatasetId();
 
-    final var validConstraintRuns = this.constraintService.getValidConstraintRuns(constraintCode, simDatasetId);
+    final var constraintCode = new ArrayList<>(this.planService.getConstraintsForPlan(planId));
```
mattdailis: `constraintCode` seems a bit misleading now that it includes both EDSL and JAR constraints. Perhaps `constraintRecords` would be more descriptive?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
```diff
@@ -120,145 +121,126 @@ public Map<Constraint, Fallible<?>> getViolations(final PlanId planId, final Opt
...
+      Collections.sort(compiledConstraints);
 
-        constraintResult.constraintName = entry.getValue().name();
-        constraintResult.constraintRevision = entry.getValue().revision();
-        constraintResult.constraintId = entry.getKey();
-        constraintResult.resourceIds = List.copyOf(names);
+      // prepare simulation results -- all resources need to be fetched ahead of time as it is unknown what profiles
+      //    a procedural constraint will access
+      final var merlinResults = resultsHandle.getSimulationResults();
+      final var constraintsResults = new SimulationResults(merlinResults);
```
mattdailis: On my first read through, `constraintResults` evoked the result of running a constraint, so I was surprised to discover that it's referring to simulation results in the format expected by constraints. `constraintResultMap` is similarly named, but means something different

mattdailis:
While reviewing I did a little exploratory refactoring - I'll leave it here for your reference if you're interested:
- https://github.com/NASA-AMMOS/aerie/commit/bdb22ccd2b52715a9eb148bd00cee39d77590875 explores the idea you and I discussed regarding loading simulation results differently if the user provides a set of EDSL-only constraints.
  - The "Mixed" case is exactly what this PR has now
  - The EDSL-only case avoids loading any profiles that aren't referenced by any constraints
  - The Procedural-only case avoids duplicating simulation results (i.e it uses the merlin results without ever constructing the constraint results)
 
My expectation is that most users will use EDSL constraints or JAR constraints, but not a mixture of the two, so most users would get some performance benefit from the specialized code

- https://github.com/NASA-AMMOS/aerie/commit/c818195100a3df5cabb7f2b620d19796aca49cce contains some exploratory refactoring related to constraint results. You had mentioned to me that `ConstraintResult` could in theory be a sealed interface, so I took a swing at what it could look like if it were. I settled on two variants: `Cached` for the case where we're simply forwarding constraint results from the database to the HTTP response, and `Uncached` for the case where all of the Java objects (violations etc) are available
- https://github.com/NASA-AMMOS/aerie/commit/2d2b237207431ba905bb08f88d94c08f2edd2d07 while reviewing `InsertConstraintResultAction` I found myself shuffling around the if statements trying to understand all the cases. This commit is what ended up making most sense to me

mattdailis (approved):
No show-stoppers. I've left comments on everything that stood out to me, but none of those need to block merge

Review based on visual inspection - I haven't run any code

ewferg:
A quick note that I have run a number of constraints (10+) I authored against a couple different models, including our tutorial model, and the code that @mattdailis reviewed produced the results I as an end user expected.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanConstraintsAction.java
Mythicaeda: I didn't realize that was an envvar, as it's hardcoded in procedural scheduling. Let me thread that part of the App Configuration into here.

Mythicaeda: Done! The file store is now threaded through the Plan Repository to here.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: I've renamed it to `constraints`

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: I've renamed `merlinResults` to `merlinSimResults` and `constraintsResults` to `edslSimResults`.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java
Mythicaeda: I swapped this instance to `List.of`

Mythicaeda:
> While reviewing I did a little exploratory refactoring - I'll leave it here for your reference if you're interested:
> 
>     * [bdb22cc](https://github.com/NASA-AMMOS/aerie/commit/bdb22ccd2b52715a9eb148bd00cee39d77590875) explores the idea you and I discussed regarding loading simulation results differently if the user provides a set of EDSL-only constraints.
>       
>       * The "Mixed" case is exactly what this PR has now
>       * The EDSL-only case avoids loading any profiles that aren't referenced by any constraints
>       * The Procedural-only case avoids duplicating simulation results (i.e it uses the merlin results without ever constructing the constraint results)
> 
> 
> My expectation is that most users will use EDSL constraints or JAR constraints, but not a mixture of the two, so most users would get some performance benefit from the specialized code
> 
>     * [c818195](https://github.com/NASA-AMMOS/aerie/commit/c818195100a3df5cabb7f2b620d19796aca49cce) contains some exploratory refactoring related to constraint results. You had mentioned to me that `ConstraintResult` could in theory be a sealed interface, so I took a swing at what it could look like if it were. I settled on two variants: `Cached` for the case where we're simply forwarding constraint results from the database to the HTTP response, and `Uncached` for the case where all of the Java objects (violations etc) are available
> 
>     * [2d2b237](https://github.com/NASA-AMMOS/aerie/commit/2d2b237207431ba905bb08f88d94c08f2edd2d07) while reviewing `InsertConstraintResultAction` I found myself shuffling around the if statements trying to understand all the cases. This commit is what ended up making most sense to me

I've added this comment to the relevant section of "Future Work", and will reference it in the ticket. Thank you for doing this!

Mythicaeda:
Test was failing because I forgot to pass the `directiveId` from the merlin simulation results to the constraints one. Tests should pass now.

PR #1598, opened 2024-11-08
Ignore generated sequencing parsers

JoelCourtney:
This PR adds the generated fprime and ampcs parsers to a local gitignore.


Mythicaeda (approved):
Thanks!

PR #1610, opened 2024-11-22
Procedural Scheduling activity deletion

### LINKED ISSUE #1594: Delete activities via Procedural Scheduling API
## Background

Users can use the Procedural Scheduling API [to add activities to a plan](https://nasa-ammos.github.io/aerie-docs/scheduling-and-constraints/procedural/scheduling/examples/), but there is currently no way to delete them.

## Requirements

* Add "delete activity" function(s) to procedural scheduling API
* Like other changes, deletions should not modify the actual plan until user calls `plan.commit()` at the end

JoelCourtney:
* **Tickets addressed:** closes #1594 
* **Review:** By commit  

## Description
This PR does two main things:
- allows procedural goals to delete activities
  - If the deleted activity has others anchored to it, they can choose whether to cascade delete them, reanchor them, or throw an error. Default is error
  - Unlike the UI, the ReAnchor option doesn't let you specify "reanchor to plan" vs "reanchor to grandparent". This is because allowing forced reanchoring to the plan is doesn't produce a predictable start time (see below). This unpredictability is more acceptable in the UI because the user will immediately see the results of their action and correct any mistakes. But in a scheduling goal, many activities might be updated unsoundly making it hard to fix the damage.
- Separates responsibility for satisfying the behavior guarantees of the editable plan interface into a reusable driver called `EasyEditablePlanDriver`. This way all the adapter has to do is translate the plan representations.

### Unpredictability of re-anchoring

The timeline/scheduling procedural libraries try to keep track of an estimated start time for anchored activities, but it is not guaranteed to be right. In practice, I expect the estimates will usually be wrong due to the interop between eDSL and procedural; and even if an estimate is right, it can become wrong in the future. This can definitely be improved, but its a separate task and it will probably never be perfect.

This is usually fine because its an "unofficial" estimate. But if we have the activity chain `A <- B <- C`, and we delete `B` and try to reanchor `C` to the plan, we need to turn the start estimate into an official start time. This *way* too fallible and unpredictable to do automatically on a large batch of activities.

On the other hand, we could reanchor `A <- C` in a predictable and reliable way. Its still not necessarily accurate, because if the original state was `A <- B (end)<- C` then `C` will be shifted forward by the duration of `B`, because `B` is assumed to be instantaneous. So its not perfect, but its at least reproducible and predictable.

## Verification
I've added unit tests for the changes to `InMemoryEditablePlan`, and e2e tests to make sure the changes are reflected in the database.

## Documentation
I've added doc comments, and I'll open a PR for the docs website soon.

## Future work
It turns out that implementing activity modification is so easy that I went ahead and did it in this PR, hidden in the backend. You'll notice in `GraphQLMerlinDatabaseService` that it not has routines for uploading creations, deletions, *and* modifications. The modifications part is used to update anchored activities during a deletion. I think extending support for this to the user will be almost trivial, but does deserve a separate PR.


JoelCourtney:
Notes from walkthrough:

- [x] require that modified activities stay the same type
  - [x] for now, require that only startOffset/anchorId can change
- [x] rename `ReAnchor` (now `PreserveTree`)
- [x] rename `EasyEditablePlanDriver` and use composition
- [x] add test for rolling back a deleted anchor

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/plan/Edit.kt
```diff
@@ -2,13 +2,28 @@ package gov.nasa.ammos.aerie.procedural.scheduling.plan
...
+  /**
+   * Returns the reverse operation.
+   *
+   * If both `E` and `E.inverse()` are applied, the plan is unchanged.
+   */
+  fun inverse(): Edit
+
   /** Create a new activity from a given directive. */
-  data class Create(/***/ val directive: Directive<AnyDirective>): Edit
+  data class Create(/***/ val directive: Directive<AnyDirective>): Edit {
```
Mythicaeda, line 21: Why is there `/***/` before `val` on this line (and on L26)?

procedural/constraints/src/test/kotlin/gov/nasa/ammos/aerie/procedural/constraints/NotImplementedSimulationResults.kt
Mythicaeda, line 1: What is this class for?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/ProceduralSchedulingSetup.java
```diff
@@ -26,13 +26,7 @@ public abstract class ProceduralSchedulingSetup {
...
   // Cross-Test Constants
   protected final String planStartTimestamp = "2023-01-01T00:00:00+00:00";
-  protected final String recurrenceGoalDefinition =
-      """
-      export default function myGoal() {
-        return Goal.ActivityRecurrenceGoal({
-          activityTemplate: ActivityTemplates.PeelBanana({peelDirection: 'fromStem'}),
-          interval: Temporal.Duration.from({hours:1})
-      })}""";
+
```
Mythicaeda, line 28: import on L4 is unused

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/EasyEditablePlanDriver.kt
```diff
@@ -0,0 +1,312 @@
...
+ * 4. When a commit is made, the commit object takes *shared ownership* of the set. If a new simulation is run (step 2)
+ *    the plan can still add to the set while it is still jointly owned by the commit. Then when an edit is made (step 3)
+ *    the commit will become the sole owner of the set.
+ * 5. When changes are rolled back, any sim results currently in the plan's set are marked stale, the previous commit's
+ *    sim results are marked not stale, then the plan will resume joint ownership of the previous commit's set.
+ *
+ * The joint ownership freaks me out a wee bit, but I think it's safe because the commits are only used to keep the
+ * previous sets from getting gc'ed in the event of a rollback. Only the plan object actually mutates the set.
+ */
+abstract class EasyEditablePlanDriver(
```
Mythicaeda: Name ideas: `DefaultEditablePlanDriver`? `EditablePlanDriverImpl`?

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ActivityDeletionGoal.java
```diff
@@ -0,0 +1,47 @@
...
+    );
+    ids[2] = plan.create(
+        "BiteBanana",
+        new DirectiveStart.Anchor(ids[1], Duration.HOUR, DirectiveStart.Anchor.AnchorPoint.Start),
+        Map.of("biteSize", SerializedValue.of(2))
+    );
+
+    if (whichToDelete >= 0) {
+      plan.delete(ids[whichToDelete], anchorStrategy);
+    }
```
Mythicaeda, line 46: Nonblocking: Is this goal meant to just add three activities in the event that `whichToDelete` is out of bounds? Nonblocking as this is test code, but brought up since it may be referenced as an example.

Mythicaeda (commented):
Mostly minor comments, as we already discussed in the walkthrough the things that need to be tweaked with this PR (most blocking for me is restricting modifying activity directives)

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/plan/Edit.kt
JoelCourtney: Those are scattered all around the Kotlin code, because I have the docs generator set to warn me if anything is undocumented. This is an empty doc comment that I put in because I thought the meaning was obvious from the name or context. I just wanted to please the docs generator.

procedural/constraints/src/test/kotlin/gov/nasa/ammos/aerie/procedural/constraints/NotImplementedSimulationResults.kt
JoelCourtney: I use it in tests to stub things. Right now its used in the generator constraint test, because I wanted a hard-coded resource profile and nothing else. So all the other methods throw.

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/EasyEditablePlanDriver.kt
JoelCourtney: I like default 👍

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ActivityDeletionGoal.java
JoelCourtney: Yes, if it is negative it leaves all three activities. Its used in the deletion e2e tests to make sure all three activities are created correctly first, before deleting anything. I'll add to the comment to document this.

Mythicaeda: It also keeps them if it's >=3

JoelCourtney: I think it probably throws an index out of bounds error if its 3 or higher

Mythicaeda: Oh, yeah 🤦‍♀️

JoelCourtney:
I was accidentally pushing to the wrong branch for the last two hours 🤦  @Mythicaeda all changes should be done now.

scheduler-driver/src/main/kotlin/gov/nasa/jpl/aerie/scheduler/plan/SchedulerPlanEditAdapter.kt
```diff
@@ -0,0 +1,90 @@
...
+import gov.nasa.jpl.aerie.scheduler.model.*
+import gov.nasa.jpl.aerie.types.ActivityDirectiveId
+import java.time.Instant
+import kotlin.jvm.optionals.getOrNull
+import gov.nasa.ammos.aerie.procedural.timeline.plan.Plan;
+
+/*
+ * An implementation of [EditablePlan] that stores the plan in memory for use in the internal scheduler.
+ *
+
```
Mythicaeda: Nonblocking: Extra lines

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -636,6 +652,66 @@ mutation createAllPlanActivityDirectives($activities: [activity_directive_insert
...
+                         update_%d: update_activity_directive_by_pk(pk_columns: {id: %d, plan_id: %d}, _set: $activity_%d) {
+                           affected_rows
+                         }
+                         """.formatted(id, id, planId.id(), id));
+
+      final var activityObject = Json
+          .createObjectBuilder()
+          .add("start_offset", act.startOffset().toString())
+          .add("anchored_to_start", act.anchoredToStart())
+          .add("name", act.name());
```
Mythicaeda: This ought to be:
```java
      final var activityObject = Json
          .createObjectBuilder()
          .add("start_offset", act.startOffset().toString())
          .add("anchored_to_start", act.anchoredToStart())
          .add("anchor_id", act.anchorId())
          .add("name", act.name());
```
Otherwise you can't actually save what the activity got reanchored to.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -636,6 +652,66 @@ mutation createAllPlanActivityDirectives($activities: [activity_directive_insert
...
+
+      final var insertionObjectArguments = Json.createObjectBuilder();
+      for (final var arg : act.arguments().entrySet()) {
+        insertionObjectArguments.add(arg.getKey(), serializedValueP.unparse(arg.getValue()));
+      }
+      activityObject.add("arguments", insertionObjectArguments.build());
+      arguments.add("activity_%d".formatted(id), activityObject);
+    }
+    postRequest(request.toString(), arguments.build()).orElseThrow(() -> new NoSuchPlanException(planId));
```
Mythicaeda, line 717: Blocking: Missing a `request.append("}")` before `postRequest`. See L710 for reference.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -636,6 +652,66 @@ mutation createAllPlanActivityDirectives($activities: [activity_directive_insert
...
+  {
+    if (ids.isEmpty()) return;
+    ensurePlanExists(planId);
+    final var request = new StringBuilder();
+    request.append("mutation deletePlanActivityDirectives {");
+    for (final var id : ids) {
+      request.append("""
+                         delete_%d: delete_activity_directive_by_pk(id: %d, plan_id: %d) {affected_rows}
+                         """.formatted(id.id(), id.id(), planId.id()));
+    }
```
Mythicaeda: Suggestion: Alternate GQL Proposal:
```gql
mutation deletePlanActivityDirectives($planId: Int! = %d, $directiveIds: [Int!]!=%s) {
  delete_activity_directive(where: {_and: {plan_id: {_eq: $planId}, id: {_in: $directiveIds}}}) {
    affected_rows
  }
}
```
`planId` has the format `1`, `directiveIds` has the format `[1, 2, 3]`.
This lets you avoid needing to build up the statement per-directive. 

A test that will show that only the relevant activities will be deleted is to create a plan with some activites, branch it, then perform a scheudling run that deletes some of the activities. The branch will be unaffected, as will the remaining activities in the plan.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -440,8 +442,21 @@ public Map<ActivityDirectiveId, ActivityDirectiveId> updatePlanActivityDirective
             activity.anchoredToStart()
         );
         if (!activityDirectiveFromSchedulingDirective.equals(actFromInitialPlan.get())) {
-          throw new MerlinServiceException("The scheduler should not be updating activity instances");
-          //updateActivityDirective(planId, schedulerActIntoMerlinAct, activityDirectiveId, activityToGoalId.get(activity));
+          final var newState = activityDirectiveFromSchedulingDirective.serializedActivity();
+          final var oldState = actFromInitialPlan.get().serializedActivity();
```
Mythicaeda: Idle question seeing this and L444: why is `actFromInitialPlan` an `Optional` instead of calling `.get` during assignment? 

That is, why isn't L435 `final var actFromInitialPlan = initialPlan.getActivityById(activity.id()).get();`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -440,8 +442,21 @@ public Map<ActivityDirectiveId, ActivityDirectiveId> updatePlanActivityDirective
             activity.anchoredToStart()
         );
```
Mythicaeda, line 442: I can't select it in a comment, but this is a question re: L429-434: Why are we changing the activity's arguments on these lines? Asking as I'm concerned about the following case: 

An activity type has a controllable duration, and the parameter that controls it has a default value. There's an instance in the plan that is using this default duration  (so its arguments would not contain the parameter). This plan is then scheduled. Wouldn't this activity cause an exception to be raised on L455?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -452,13 +467,14 @@ public Map<ActivityDirectiveId, ActivityDirectiveId> updatePlanActivityDirective
...
-        throw new MerlinServiceException("The scheduler should not be deleting activity instances");
-        //deleteActivityDirective(idActInInitialPlan.getValue());
+        toDelete.add(idInInitialPlan);
       }
     }
 
     //Create
     ids.putAll(createActivityDirectives(planId, toAdd, activityToGoalId, schedulerModel));
+    modifyActivityDirectives(planId, toModify);
+    deleteActivityDirectives(planId, toDelete);
```
Mythicaeda, line 476: `updatePlanActivityDirectiveAnchors` ought to be called between `modify` and `delete`. Not only would this let the method become private (and have `storeFinalPlan` _actually_ store the final plan), but if it's not done in this order, then a preserved anchor tree will cause the scheduling run to fail during `deleteActivityDirectives`, since you can't delete an activity that is an active anchor.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/DeletionTests.java
Mythicaeda, line 1: Blocking: None of these tests appear to excersize deleting and reanchoring activities that were already in the plan, only ones that were created as part of scheduling. Please add tests where activities in the plan prior to scheduling are 1) deleted and 2) reachored so that the new GQL code can be tested. I would especially prefer if these tests involved multiple existing activities being reanchored/have their start times adjusted to unique offsets to fully test the `modifyActivityDirecitves` behavior, since it's chaining sub-mutations.

Mythicaeda (commented):
Files left unreviewed as of this review: `DefaultEditablePlanDriver.kt`, `Procedure.java`, `SchedulerPlanEditAdapter.kt`, and `EditablePlanTest.java`

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
Mythicaeda: Actually, I just checked `updatePlanActivityDirectiveAnchors` and it _only_ runs for new activities. This means that either 
 1) that method needs to be changed to run for modified activities too, including supporting reanchoring existing activities to the plan, (in which case, [this comment](https://github.com/NASA-AMMOS/aerie/pull/1610#discussion_r1927950763) can be disregarded), or 
 2) [the comment about including `anchorId` in the fields being updated in `modifyActivityDirectives`](https://github.com/NASA-AMMOS/aerie/pull/1610#discussion_r1927950763) needs to have its scope increased to support reanchoring to activities that are newly-created in scheduling.

If 2) is chosen, I still think `updatePlanActivityDirectiveAnchors` ought to be moved into this function

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Procedure.java
```diff
@@ -64,24 +65,24 @@ public void run(
...
 
-    if (!editablePlan.getUncommittedChanges().isEmpty()) {
+    if (editablePlan.isDirty()) {
       throw new IllegalStateException("procedural goal %s had changes that were not committed or rolled back".formatted(jarPath.getFileName()));
     }
     for (final var edit : editablePlan.getTotalDiff()) {
       if (edit instanceof Edit.Create c) {
         newActivities.add(toSchedulingActivity(c.getDirective(), lookupActivityType::apply, true));
-      } else {
-        throw new IllegalStateException("Unexpected value: " + edit);
```
Mythicaeda, line 84: Why was this removed instead of changing the `else` to an `else if (!(edit instanceof Edit.Delete))`? Additionally, is there any value in having a procedure track updated and deleted activities as well as added ones?

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/DefaultEditablePlanDriver.kt
```diff
@@ -0,0 +1,315 @@
...
+    for (simResults in upToDateSimResultsSet) {
+      simResults.get()?.setStale(true)
+    }
+    // create a new list instead of `.clear` because commit objects have the same reference
+    upToDateSimResultsSet = mutableSetOf()
+
+    return id
+  }
+
+  override fun delete(directive: Directive<AnyDirective>, strategy: DeletedAnchorStrategy) {
```
Mythicaeda, line 187: Nonblocking: could you add a couple of inline comments to this function? The outer `else` branch isn't doing much, but that's hard to see without dissecting the Kotlin-isms.

Also L188 is an extra newline.

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/DefaultEditablePlanDriver.kt
```diff
@@ -0,0 +1,315 @@
...
+    val directivesToCreate: Set<Directive<AnyDirective>>
+
+    if (strategy == DeletedAnchorStrategy.Cascade) {
+      directivesToDelete = deleteCascadeRecursive(directive, directives).toSet()
+      directivesToCreate = mutableSetOf()
+    } else {
+      directivesToDelete = mutableSetOf(directive)
+      directivesToCreate = mutableSetOf()
+      for (d in directives) {
+        when (val childStart = d.start) {
```
Mythicaeda: Nonblocking: Does Kotlin have the same kind of streamable support Java does? Ie `val childDirectives = directives.filter{ d -> d.start is DirectiveStart.Anchor && d.start.parentId == directive.id }`. (or even skipping `childDirectives` and chaining into a `forEach`)

I don't know if that would be more performant, but it may help collapse L199-L202 from a forloop -> switch -> if statement into what that works out into, which is, "get the activities directly anchored to this one, then loop over them and resolve according to anchor strat"

JoelCourtney: I can't think of a way to do that while keeping the smart cast of the first `when` block. Yes it would be possible, but the only aesthetically pleasing way I can think of has an unchecked cast in it.

(Kotlin doesn't like smart casting objects whose origins it can't statically analyze. That's why I have to assign `childStart` instead of just using a smart-casted `d.start`.)

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/goals/Procedure.java
JoelCourtney: I can put the error back, no reason not to.

I don't think the procedure could do anything with the record of deleted activities, yet. We only record the new activities so the `source_scheduling_goal` field can be set in the database, which obviously doesn't exist for deleted activities.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
JoelCourtney: Fascinating question... I have no idea why this is here. We do have an algorithmic guarantee that this won't ever actually change the duration though, thats why it doesn't throw an error.

This is probably just a mixup from when I added the `isNew` flag. There's a similar statement in `createActivityDirectives` that sets the controllable duration, that's the only one we should need.

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/DefaultEditablePlanDriver.kt
Mythicaeda: Understood, though can you explain it? In Java, I would expect there to be no issue with the cast since the "check" `d.start is DirectiveStart.Anchor` would be guarding the cast.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
JoelCourtney: No reason, I think. I'll fix it

JoelCourtney:
Looks like sequencing-server e2e tests are failing 😠

Mythicaeda:
> Looks like sequencing-server e2e tests are failing 😠

Yeah, they've been extremely flaky the past couple days 😞

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
JoelCourtney: I've deleted the update-anchor-step entirely, and now I just check and upload anchor changes in the general modify step.

Mythicaeda: The `update-anchor` step was for **new** activities and is still very much needed in the `new` case, as if any activity is added to Aerie during a scheduling run, the ids the scheduler server _thinks_ new activities will have will **not** be the ids that get assigned to them by the DB

Mythicaeda: Basically:
 - new activities need to come in without anchors (due to a combined issue of 1) the activity its anchored to may not yet exist and 2) the final ids need to be "locked in" by the db), then have them assigned as a follow-up step 
 - modified activities can have their anchors assigned at the same as the rest of the activity is modified, as long it runs after the ids for new activities are "locked in" by the db

JoelCourtney: you right

JoelCourtney: Would it be vulgar of me to just add all created activities to the "modified" list? 🫥

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/DeleteBiteBananasGoal.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Deletes all Bite Bananas with extreme prejudice. Used to test that updated
+ * anchors are saved in the database properly.
+ */
+@SchedulingProcedure
+public record DeleteBiteBananasGoal() implements Goal {
+  @Override
+  public void run(@NotNull final EditablePlan plan) {
+    plan.directives("BiteBanana").forEach($ -> plan.delete($, DeletedAnchorStrategy.PreserveTree));
```
Mythicaeda: I'd like to request tests that test the behavior of the other two `DeletedAnchorStrategy`s (`Error` and `Cascade`)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/DeletionTests.java
```diff
@@ -7,6 +7,7 @@
 import org.junit.jupiter.api.Test;
 
 import javax.json.Json;
+import javax.json.JsonValue;
```
Mythicaeda: This added import is unused.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/DatabaseDeletionTests.java
```diff
@@ -0,0 +1,177 @@
...
+        && Objects.equals(it.startOffset(), "02:00:00")
+    ));
+  }
+
+  @Test
+  void deletesDirectiveInDatabaseInMiddleOfChain() throws IOException {
+
+    // Creates 5 activities, deletes "Bite".
+    // grow1 <- bite
+    // bite <- grow (id lost)
```
Mythicaeda: What does `id lost` mean here?

Mythicaeda (commented):
This review skips over `GraphQLMerlinDatabaseServer.java`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/DatabaseDeletionTests.java
JoelCourtney: It just means I didn't record it in a variable, because I didn't anchor anything to it. I'll clarify.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
JoelCourtney: I've refactored the modify step to accept a configurable list of changes for each activity. The create step still ignores the anchor id, and then the anchors are added to the modification list in between the "create" and "modify" steps.

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/DeleteBiteBananasGoal.java
JoelCourtney: Done, see bottom of databasedeletiontests

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -452,40 +452,71 @@ public Map<ActivityDirectiveId, ActivityDirectiveId> updatePlanActivityDirective
...
+        toDelete.add(idInInitialPlan);
       }
     }
 
     //Create
     ids.putAll(createActivityDirectives(planId, toAdd, activityToGoalId, schedulerModel));
+
+    // Create does not upload the anchor ids, because directive IDs can change during upload
+    // and it would cause a foreign key violation. So we map the anchor ids using the creation results
+    // and add an anchor-modification entry to the `toModify` list after the fact.
```
Mythicaeda, line 465: Thank you for adding this as a comment.

Mythicaeda (approved):
LGTM!

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -636,6 +667,69 @@ mutation createAllPlanActivityDirectives($activities: [activity_directive_insert
...
+        ",",
+        modifications.stream().map($ -> "$activity_%d: activity_directive_set_input!".formatted($.id().id())).toList()
+    ));
+    request.append(") {");
+    final var arguments = Json.createObjectBuilder();
+    for (final var mod : modifications) {
+      final var id = mod.id().id();
+      request.append("""
+                         update_%d: update_activity_directive_by_pk(pk_columns: {id: %d, plan_id: %d}, _set: $activity_%d) {
+                           affected_rows
```
Mythicaeda: `affected_rows` doesn't exist when you do a `pk` operation. I suggest `id`

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/DeleteBiteBananasGoal.java
Mythicaeda, line 1: Is it possible to add assembling these files to the `assemble` task of `e2e-tests`? It's annoying to have to remember to assemble these afterwards.

JoelCourtney: I've tried, and I think Luke has tried too. Its probably possible, but I haven't figured it out. The main problem is that `assemble` generates the Goal mapper objects that Gradle then detects to create the `jar` tasks themselves. Each `jar` task is created dynamically *when gradle starts*, and creating a new task in the middle of gradle eval is tough.

JoelCourtney:
Updates since last review:
- in "fix e2e tests"
  - made Directive.name nullable. I thought it was not-null in the database, but apparently it is
  - in the updating-new-activity-anchors step, I had only mapped the anchor id to the new value, not the main activity id
- Removed `SchedulingActivity.isNew`. I originally made it so that I didn't have to do a full plan-diff during the upload step, but it turns out its pretty difficult to maintain the accuracy of `isNew` when modifying anchors. So I deleted the field and I'm instead relying only on activity IDs to find modified activities.

Mythicaeda (approved):
🎉

PR #1611, opened 2024-11-22
Procedural constraints annotation processor + build logic

### LINKED ISSUE #1571: Annotation Processor updates for Procedural Constraints
### Background

Now that [Procedural Scheduling](https://nasa-ammos.github.io/aerie-docs/scheduling-and-constraints/procedural/introduction/) has been released, we are working on implementing [Procedural Constraints](https://nasa-ammos.github.io/aerie-docs/scheduling-and-constraints/procedural/constraints/). The workflow for using them will be very similar to scheduling goals: write a constraint in Java, compile a JAR, upload it via the Aerie UI, and add it to a plan using the same list UI that is used for eDSL constraints. Whereas procedural scheduling **goals** create **activities** on an editable plan, procedural **constraints** will return **violations**, ie. windows of time when the constraint is violated, and will not be able to edit the plan directly.

Procedural Scheduling goals use an **annotation processor** to look for classes with the annotation `@SchedulingProcedure` and process them with `ValueMappers` to generate the final goal/procedure code. It's located here:

[procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/SchedulingProcedureProcessor.java](https://github.com/NASA-AMMOS/aerie/blob/develop/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/SchedulingProcedureProcessor.java)

### Requirements

- Define a new annotation for constraints - `@SchedulingConstraint` or `@ProceduralConstraint`?
  - or could re-use existing `@SchedulingProcedure` annotation, but might be confusing
- Update the annotation processor (or create a new one) to process constraints and generate code that implements the correct interface
- Populate constraint parameter schemas stored in the DB based on the parameters from the generated code

### Related Issues
https://github.com/NASA-AMMOS/aerie/issues/1572, https://github.com/NASA-AMMOS/aerie-ui/issues/1478

skovati:
* **Tickets addressed:** closes https://github.com/NASA-AMMOS/aerie/issues/1571
* **Review:** By commit  

## Description
This PR does a few things related to procedural constraints workflow:
- creates the `ConstraintProcedure` annotation and handles it together with scheduling procedures in the same processor.
- Modifies the procedure-building gradle script to do scheduling and constraints procedures together.
- Implements the `EditablePlan`, `Plan`, and `SImulationResults` interfaces for stateless-aerie plan objects.
- Creates a basic test constraint `ObeyConservationOfBanana` and tests it using plan stubs and stateless simulation. This is less a test of the constraint itself, and more a test-that-testing-is-actually-possible kind of thing.

## Documentation

I still need to update the procedural constraints docs.

## Future work

We now have nearly everything for a full-loop external scheduler run. Once we make a plan serializer, you could make a driver that loads a plan from file, runs a series of constraints and goals (with simulation), and outputs a result plan.


Mythicaeda:
This PR is currently orphaned, pair programming hosted by @mattdailis prior to handoff.

This PR has compiled a constraint w/o arguments, so it's potentially close.

@JoelCourtney to add example constraints to this branch

Decision to be made: support non-records or updating docs? @JoelCourtney proposing to just update docs and allow non-records later.

Let's start with only records and double-back for non-record support.

JoelCourtney:
State of PR:

The annotation processor worked without any extra work. It was actually so good at combining the scheduling / constraints workflows that the gradle task can't tell them apart, and the original `buildAllSchedulingProcedureJars` task would build the constraints as well. So I've just removed the `Scheduling`, making it `buildAllProcedureJars`. I don't see why the user would need separate workflows for them anyway. The annotation processor does still generate different value mapper types for goals & constraints. Since this gives us a bit of type safety during procedure loading, I figured I'd leave it and not combine them. So:
- two different annotations
- one annotation processor
- two value mappers
- one gradle command

I haven't made the equivalent of the scheduler's `Procedure.java` for constraints. I'm not sure that's included in this task, but I can work on it because it seems like the logical next step.

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/SchedulingProcedureMapper.kt
```diff
@@ -3,7 +3,7 @@ package gov.nasa.ammos.aerie.procedural.scheduling
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue
 import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema
 
-interface ProcedureMapper<T: Goal> {
+interface SchedulingProcedureMapper<T: Goal> {
```
Mythicaeda: Heads up: this rename will break existing uploaded procedural scheduling goals. I actually encountered the reverse of this, where I built a goal on this branch and tried to use it on the db support branch, only to get `implementation not found` exception.

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/StartOffsetReducer.java
```diff
@@ -104,9 +104,16 @@ private Pair<ActivityDirectiveId, Duration> getNetOffset(ActivityDirective ad){
     boolean anchoredToStart = ad.anchoredToStart();
     Duration netOffset = ad.startOffset();
 
+    final var firstId = currentAnchorId;
+
     while(currentAnchorId != null && anchoredToStart){
       currentActivityDirective = completeMapOfDirectives.get(currentAnchorId);
       currentAnchorId = currentActivityDirective.anchorId();
+
+      if (currentAnchorId.equals(firstId)) {
```
Mythicaeda: This line needs to be reversed to avoid a NPE: `if(firstId.equals(currentAnchorId)`

This is because we know for a fact that `firstId` is not null, as we otherwise wouldn't have entered the while loop, but we expect `currentAnchorId` to become null at some point, assuming there is no cycle.

Mythicaeda: Additionally, I'm not sure if tracking a single id (as opposed to the set of all seen ids so far) is sufficient to detect a cycle, as the cycle may start further up than the initial directive id. If it is, could you please break down _why_ it is?

For context, the database uses the "set of all seen ids" approach to detecting cycles.

`Set` proposal based on what I did to get tests working on my branch:
```java
final var seenIds = new HashSet<ActivityDirectiveId>();
seenIds.add(currentAnchorId);

while(currentAnchorId != null && anchoredToStart){
    currentActivityDirective = completeMapOfDirectives.get(currentAnchorId);
    currentAnchorId = currentActivityDirective.anchorId();

    if (seenIds.contains(currentAnchorId)) {
        throw new IllegalStateException("Anchor ID cycle detected on " + currentAnchorId);
    }
//...
```

procedural/examples/banana-procedures/build.gradle
```diff
@@ -90,5 +92,8 @@ private static String getMainClassFromGeneratedFile(File file) {
 }
 
 test {
+  testLogging {
+    showStandardStreams = true
+  }
```
Mythicaeda: What does this do?

procedural/utils/MODULE_DOCS.md
Mythicaeda, line 1: These docs are empty. Did you mean to move/rework the contents?

type-utils/src/main/java/gov/nasa/jpl/aerie/types/ActivityDirective.java
```diff
@@ -4,6 +4,7 @@
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 
 import java.util.Map;
+import java.util.Objects;
```
Mythicaeda: Unused added import.

procedural/utils/src/main/java/gov/nasa/ammos/aerie/procedural/utils/TypeUtilsPlanAdapter.java
```diff
@@ -0,0 +1,99 @@
...
+      @NotNull final Function1<? super List<Segment<SerializedValue>>, ? extends TL> deserializer)
+  {
+    throw new NotImplementedException();
+  }
+
+  @NotNull
+  @Override
+  public ExternalEvents events(@NotNull final EventQuery query) {
+    throw new NotImplementedException();
+  }
```
Mythicaeda, line 98: Are these supposed to be implemented?

procedural/utils/src/main/java/gov/nasa/ammos/aerie/procedural/utils/TypeUtilsEditablePlanAdapter.java
```diff
@@ -0,0 +1,185 @@
...
+  private final SimulationUtility simUtility;
+  private final MissionModel<?> model;
+  private long idCounter;
+  private SimulationResults latestResults;
+
+  private boolean changedSinceLastSim = true;
+
+  public TypeUtilsEditablePlanAdapter(
+      TypeUtilsPlanAdapter plan,
+      SimulationUtility simUtility,
```
Mythicaeda, line 73: Major: I get the following error from this class: `Cannot resolve symbol "SimulationUtility". Add dependency on module "orchestration-utils"`. Please update the `build.gradle` to resolve this.

procedural/examples/banana-procedures/src/test/java/gov/nasa/ammos/aerie/procedural/examples/bananaprocedures/constraints/TestBananaConservationStub.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class TestBananaConservationStub {
```
Mythicaeda, line 32: Minor: Since this and `TestBananaConservationSim` are meant to be used for user reference, could you add a class-level comment briefly going over the approach to generating simulation data. Ie, how `Stub` stubs out just the needed data, and how `Sim` runs a simulation against the built model.

procedural/examples/banana-procedures/build.gradle
JoelCourtney: Whoops thats just for test debugging. It makes gradle not capture stdout.

procedural/utils/MODULE_DOCS.md
JoelCourtney: Forgot to fill it in, that's all. I'll add a bit to it

procedural/utils/src/main/java/gov/nasa/ammos/aerie/procedural/utils/TypeUtilsPlanAdapter.java
JoelCourtney: Type-utils plan doesn't have external resource profiles or external events, so no. This is just an adapter, its not intended to do anything that type-utils Plan doesn't already do.

procedural/utils/src/main/java/gov/nasa/ammos/aerie/procedural/utils/TypeUtilsEditablePlanAdapter.java
JoelCourtney: The `build.gradle` already does have a dependency on `orchestration-utils`. Try a clean build?

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/StartOffsetReducer.java
JoelCourtney: Fixed both

procedural/utils/src/main/java/gov/nasa/ammos/aerie/procedural/utils/TypeUtilsEditablePlanAdapter.java
Mythicaeda: For some reason `clean`ing wasn't effective and I had to hard-refresh my dependencies, but yes, the issue has gone away, and I see the entry in `build.gradle`

Mythicaeda (approved):
LGTM!

PR #1617, opened 2024-12-09
Refine migration script

### LINKED ISSUE #852: Use dotenv to load in .env in Database Migration Script
As mentioned in [this comment](https://github.com/NASA-AMMOS/aerie/pull/845#discussion_r1161821025), rather than manually parsing a .env, we should use a library like [python-dotenv](https://github.com/theskumar/python-dotenv) to get the Postgres credentials. This will:
1. Make extracting the username/password from the .env easier and
2. Give the option to pass in the username/password as an environment variable in the shell rather than as part of a .env file.

### LINKED ISSUE #1353: aerie_db_migration.py output to display correct hasura endpoint
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

No - I haven't considered

### Related problems

When I ran below commands, `aerie_db_migraiont.py` didn't display correct hasura url endpoint.  
There were two config.yaml files in this [repo](https://github.jpl.nasa.gov/Europa-PESS/clipper-aerie/tree/G24.0/deployment).
It seems like that the command output displayed endpoint specified in ./config.yaml while it was actually connecting to the address specified in ./hasura/config.yaml as described in the code.
Would you please update the script so it'll display hasura endpoint actually connecting?

```
% python3 aerie_db_migration.py -a --all -n <rds_endpoint> -e aerie.env
```

Would you please update the script so it'll display hasura endpoint actually connecting?

Mythicaeda:
* **Tickets addressed:** Closes #1353 and Closes #852 
* **Review:** By commit  

## Description

Adds a tracked view to Hasura showing the integer contents of the `migrations.schema_migrations` table. This allows Aerie Admins on a venue to view which migrations have been applied (was formerly locked behind Hasura Admins/Postgres Admins).

Heavily refactors the Aerie DB Migration Script:
 - Adds the 'status' subcommand to view the current status of migrations without applying them
 - Moves the migration logic into the 'migrate' subcommand
 - Adds a Hasura class for communicating via the CLI or API. This ensures that all necessary flags are always passed to the CLI. Necessary flags include: the Hasura project root, the Hasura endpoint and admin secret, the location of the .env to load values from, the database to apply migrations to
 - Loads in endpoint and admin secret information the same way that the Hasura CLI does (flags, then envvars, then .env, then config.yaml)
 - Gets the current database migration status by running a [`run_sql` query against the Hasura venue](https://hasura.io/docs/2.0/api-reference/schema-api/run-sql/), removing the need to provide database connection information
 - **Fixes the bug where Hasura metadata would be reloaded but not applied, forcing the user to restart Hasura**. The new metadata is now _applied_ before it's reloaded

## Verification

Manually tested

## Documentation

Added doc comments and type hints throughout the script.

The website docs were edited to reflect the changes to the command line arguments and introduction of subcommands. 
- [Docs PR can be found here](https://github.com/NASA-AMMOS/aerie-docs/pull/201).

## Future work

1) Refining the output when a migration fails to run. Currently the blob JSON is printed and it's hard to parse even if you know what you're looking for.
2) Perhaps adding an "until" flag to the bulk migration (as in, "apply[revert] migrations until you reach this one"). Not sure if there's a pressing use case for this currently.
3) #1627

### DIFF
--- a/.github/workflows/pgcmp.yml
+++ b/.github/workflows/pgcmp.yml
@@ -111,7 +111,7 @@ jobs:
           AERIE_PASSWORD=${AERIE_PASSWORD}
           EOF
           python -m pip install -r requirements.txt
-          python aerie_db_migration.py --apply --all
+          python aerie_db_migration.py migrate --apply --all
           cd ..
       - name: Clone PGCMP
         uses: actions/checkout@v4
@@ -204,7 +204,7 @@ jobs:
           AERIE_PASSWORD=${AERIE_PASSWORD}
           EOF
           python -m pip install -r requirements.txt
-          python aerie_db_migration.py --revert --all
+          python aerie_db_migration.py migrate --revert --all
           cd ..
       - name: Dump Migrated Database
         run: |
--- a/deployment/aerie_db_migration.py
+++ b/deployment/aerie_db_migration.py
@@ -279,6 +279,74 @@ def bulk_migration(hasura: Hasura, apply: bool):
   exit(exit_with)
 
 
+def migrate(args: argparse.Namespace):
+  hasura = create_hasura(arguments)
+
+  clear_screen()
+  print(f'\n###############################'
+        f'\nAERIE DATABASE MIGRATION HELPER'
+        f'\n###############################'
+        f'\n\nMigrating database at {hasura.endpoint}')
+  # Enter step-by-step mode if not otherwise specified
+  if not args.all:
+    # Find all migration folders for the database
+    migration_path = os.path.abspath(args.hasura_path+"/migrations/Aerie")
+    migration = DB_Migration(migration_path, args.revert)
+
+    # Go step-by-step through the migrations available for the selected database
+    step_by_step_migration(hasura, migration, args.apply)
+  else:
+    bulk_migration(hasura, args.apply)
+
+
+def status(args: argparse.Namespace):
+  hasura = create_hasura(args)
+
+  clear_screen()
+  print(f'\n###############################'
+        f'\nAERIE DATABASE MIGRATION STATUS'
+        f'\n###############################'
+        f'\n\nDisplaying status of database at {hasura.endpoint}')
+
+  display_string = f"\n\033[4mMIGRATION STATUS:\033[0m\n"
+  output = hasura.get_migrate_output('status')
+  del output[0:3]
+  display_string += "\n".join(output)
+  print(display_string)
+
+
+def create_hasura(args: argparse.Namespace) -> Hasura:
+  """
+  Create a Hasura object from the CLI arguments
+
+  :param args: Namespace containing the CLI arguments passed to the script. Relevant fields in Namespace:
+    - hasura_path (mandatory): Directory containing the config.yaml and migrations folder for the venue
+    - env_path (optional): Envfile to load envvars from
+    - endpoint (optional): Http(s) endpoint for the venue's Hasura instance
+    - admin_secret (optional): Admin secret for the venue's Hasura instance
+  :return: A Hasura object connected to the specified instance
+  """
+  if args.env_path:
+    if not os.path.isfile(args.env_path):
+      exit_with_error(f'Specified envfile does not exist: {args.env_path}')
+    load_dotenv(args.env_path)
+
+  # Grab the credentials from the environment if needed
+  hasura_endpoint = args.endpoint if args.endpoint else os.environ.get('HASURA_GRAPHQL_ENDPOINT', "")
+  hasura_admin_secret = args.admin_secret if args.admin_secret else os.environ.get('HASURA_GRAPHQL_ADMIN_SECRET', "")
+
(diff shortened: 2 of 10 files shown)


dandelany (changes requested):
@mattdailis and I successfully tested and reviewed this yesterday - thanks @Mythicaeda ! There was only one issue that rose to the level of requested change:

As discussed, we accept that this is a non-backwards-compatible change to the migration script command (since subcommands `migrate` and `status` are now used). However, if users accidentally call it **without** a subcommand/arguments (ie. `python aerie_db_migration.py`) due to using an old CI script or referencing old docs, it now fails with an inscrutable error message:

```
Traceback (most recent call last):
  File "/Users/dailis/projects/aerie/migration-script-testing-2/aerie_db_migration.py", line 512, in <module>
    arguments.func(arguments)
    ^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'func'
```
Can we **improve the message in this case** so users know what the issue is?

Other notes/non-blockers from testing:
* [ ] I will need to include this change to the command in the upgrade guide when released
* nonblocker: If you make a mistake while running the script, you can't Ctrl+C to exit during the "safe" part while Hasura is checking applied migrations, it only recognizes Ctrl+C after it starts running them.
* nonblocker: I'd like to eventually improve docs/support for downgrading/running reverse migrations, since it's not super intuitive and requires knowing which down migrations need to be run in advance.
* Unrelated issue: we noticed that in our `deployment.zip` files for each release, the default docker tag is `latest`, we should consider (separately) making releases use the release tag.

Our testing procedure:
* We spun up an old version (v2.20?) of Aerie locally and created a plan, some activities, and simulated. Then we used the script on this branch to upgrade to the latest Aerie version, confirmed the necessary forward migrations ran, and tested with the new version with no issues.
* We created a new dummy migration, and migrated both forward and backward using this script with no issues, with verification steps in between to make sure the correct database changes were applied.

Mythicaeda:
> However, if users accidentally call it without a subcommand/arguments (ie. python aerie_db_migration.py) due to using an old CI script or referencing old docs, it now fails with an inscrutable error message:

Thanks for catching this! I've updated the script to print the `help` message in this case. 

```
$ aerie_db_migration.py 
usage: aerie_db_migration.py [-h] <command> ...

Migrate the database of an Aerie venue.

options:
  -h, --help  show this help message and exit

commands:
  <command>
    status    Get the current migration status of the database
    migrate   Migrate the database
```

Are there any similar issues with the behavior when a subcommand is omitted entirely?

```
$ aerie_db_migration.py -e ../.env
usage: aerie_db_migration.py [-h] <command> ...
aerie_db_migration.py: error: argument <command>: invalid choice: '../.env' (choose from 'status', 'migrate')
```

> nonblocker: If you make a mistake while running the script, you can't Ctrl+C to exit during the "safe" part while Hasura is checking applied migrations, it only recognizes Ctrl+C after it starts running them.

Can you clarify if you mean during `mark_current_version` at the start of execution or `reload_metadata` between steps (slash at the end in bulk-mode)?

> nonblocker: I'd like to eventually improve docs/support for downgrading/running reverse migrations, since it's not super intuitive and requires knowing which down migrations need to be run in advance.

This feels beyond the scope of this PR. Could you make a ticket for this so that it's not lost?  I'll link it in the "Future Work" section.

dandelany (approved):
> I've updated the script to print the help message in this case.

Thanks! Approved.

> Are there any similar issues with the behavior when a subcommand is omitted entirely?

Not that I've seen, but I'll keep an eye out.

> (re: Ctrl-C) Can you clarify if you mean during mark_current_version at the start of execution or reload_metadata between steps (slash at the end in bulk-mode)?

I'm not sure - when you run the `migrate` command there is a long (10-30 second) wait time while it does "something" before actually running migrations - it's during this time that Ctrl+C seemed to be ignored.

> This feels beyond the scope of this PR. Could you make a ticket for this so that it's not lost? I'll link it in the "Future Work" section.

👍 done - filed #1627 (not prioritized yet)

Mythicaeda:
> > Are there any similar issues with the behavior when a subcommand is omitted entirely?

> Not that I've seen, but I'll keep an eye out.

To be explicit, the code block below that comment (and I'll include it here again) was the output when there is no subcommand:

```
$ aerie_db_migration.py -e ../.env
usage: aerie_db_migration.py [-h] <command> ...
aerie_db_migration.py: error: argument <command>: invalid choice: '../.env' (choose from 'status', 'migrate')
```

--------

> I'm not sure - when you run the migrate command there is a long (10-30 second) wait time while it does "something" before actually running migrations - it's during this time that Ctrl+C seemed to be ignored.

That's `mark_current_version`. That means the slowdown's either on the request getting the migration information (where I don't think there's any improvements that can be made) or, more likely, it's in the loop afterwards where all the migrations that are in the DB are marked as applied in Hasura. I can refine this loop so that it only applies unmarked migrations, meaning that you have the same slowdown the very first time a DB is ever migrated but shouldn't experience it again afterwards.

PR #1618, opened 2024-12-13
Scheduling activity auto-deletion

### LINKED ISSUE #1609: Activity auto-deletion/"easy idempotence" in Procedural Scheduling
## Background
Proposed new feature for the Java Procedural Scheduling library.

Procedural scheduling goals are essentially Java programs that can make arbitrary changes to the Plan, such as adding, editing or removing activities. This was an intentional decision to maximize their expressiveness - however, since Aerie provides no guarantees or restrictions on how goals will behave or what kind of "side effects" they have, they may end up having behavior which is *tied to the number of times they have been run*. For example, if a goal simply adds an activity to the plan every time it runs, then you'll get a new activity every time you run the goal, with a total number of activities = number of runs.

This kind of situation can be hard to reason about, and cause unpredictable results, so generally we recommend that users try to make their activities **idempotent**, that is, to have their side effects be *independent of how many times they were run*. Running 10 times should produce the same result as once. However, we can't guarantee idempotence of goals **in general** because some important use cases and performance optimizations may rely on intentional non-idempotence.

## Description
The proposal from @JoelCourtney (and echoed in requests from users) is to provide a method in the scheduling library which allows users to guarantee that their goals will be idempotent by **automatically cleaning up** the activities they create. Specifically:

* We should add a function call or different class to inherit from to the Procedural Scheduling API
  - `guarantee_idempotence()`? `clean_up_activities()`? or maybe `IdempotentGoal`? open to suggestions 
* The user can use this function/class in their goal to intentionally opt-in if desired
* When the goal is run, we first look for any activity directives in the plan that were placed by a **previous invocation of the same goal**, and delete them.
  - If the plan goal specification contains *multiple* invocations of the same goal (ie. with different parameters), this deletion should *only* apply to activities created by the previous run of the same invocation, rather than all activities created by the goal in all of its invocations.
* Then run the goal as usual, adding any new activities to the plan

## Extra Credit
Something to think about when implementing this feature - this addresses "cleaning up" activities from past runs before the next run, but what if a user decides a goal invocation is no longer relevant, and removes it from the plan spec entirely - the activities it created will remain, with no easy way to delete them all (if desired). Maybe we should also think about an option to "remove activities created by this goal invocation" when removing a goal from a plan?

JoelCourtney:
* **Tickets addressed:** closes #1609 
* **Review:** By commit  

Turn on "Hide whitespace" in the github diff viewer, because I fixed a large formatting error in `SynchronousSchedulerAgent` and it hides the real change.

## Description
This PR allows scheduling goals to choose to automatically have their previous outputs deleted during scheduling.
- Added a column to `activity_directive` to track not just the goal id it was created by, but the invocation id too. I also added foreign key relationships to them, which required reorganizing the database init process.
- In a scheduling goal you can now optionally override a method `shouldDeletePastCreations` that returns an `ActivityAutoDelete` object. The choices are:
  - `No`: don't auto delete (default if not overridden)
  - `JustBefore(DeletedAnchorStrategy)`: deletes the activities in the middle of the scheduling run, just before this goal executes. This means previous goals will have seen the past outputs before they got deleted and recreated.
  - `AtBeginning(DeletedAnchorStrategy, boolean simulateAfter)`: deletes the activities at the beginning of the scheduling run, with the option to simulate on the cleaned plan before running the first goal. This means previous goals will not see the deleted directives, and if `simulateAfter` is true, they won't see those activities' effects either.
 - The `PrioritySolver` runs this function on all procedural goals at the beginning of the run and behaves accordingly.
 - Also I moved the `GoalId` record out of the scheduler-server into scheduler-driver, which made a pretty big commit that is mostly just changing imports.

See `ActivityAutoDeletionGoal` in the e2e tests for an example.

This doesn't allow the user to undo a previous deletion from that goal, because we don't track deletions. It can only undo a previous creation.

## Verification
I've made some e2e tests that checks the interaction between an auto-deleting goal and an eDSL CoexistenceGoal over multiple runs.

## Documentation

I'll add a section to the scheduling docs.


deployment/hasura/migrations/Aerie/18_activity_goal_invocation_id/down.sql
Mythicaeda, line 1: Style issue (I'll call this "technically nonblocking" since the PR Check is passing): On Aerie, we write our migrations such that the `up` and `down` flow in reverse order -- that is, if the first thing the `up.sql` does is add a table column, the very last thing the `down` migration should do is drop that table column. We do this for two reasons:

1) It helps developers avoid debugging dependency issues while writing migrations. This is especially helpful if a migration is touching more than a single table (ie, if you had had to alter the schema of both `activity_directive` _and_ `scheduling_specification`, rather than just `activity_directive`.)
2) It helps the reviewer review the code. If the migrations are written in this way, then the reviewer can pull the two migrations up side-by-side and check that everything that happens in the `up` is undone in the `down` and vice versa without having to search the other file for the inverting change.

Just to be explicit about the style, I'll mention that this is specifically about _schema_ changes (which are triggers, functions, table structure, and types). _Data_ changes (statements that use `insert`, `update`, `delete` or _execute_ a function) may not have an "inverting" action or may need to happen slightly "out of order" (but should try to be "in order" when applicable).

Mythicaeda (commented):
SQL Code looks good, just one comment about migration styles.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
```diff
@@ -122,6 +124,13 @@ public Map<String, List<ExternalEvent>> getExternalEvents(final PlanId planId, f
     public Collection<ResourceType> getResourceTypes(final PlanId planId) {
       return null;
     }
+
+    @Override
```
Mythicaeda, line 128: Nonblocking: Idle thought looking at the changes between this file and `MockMerlinDatabaseService`: why does this file have its own mock implementation rather than using an instance of that class? `MockMerlinDatabaseService` implements `OwnerRole`, which is a subclass of `ReaderRole`.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/GoalId.java
```diff
@@ -1,4 +1,4 @@
-package gov.nasa.jpl.aerie.scheduler.server.models;
+package gov.nasa.jpl.aerie.scheduler.model;
```
Mythicaeda, line 1: Nonblocking: What was the driver for moving this file?

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/DefaultEditablePlanDriver.kt
```diff
@@ -49,7 +49,7 @@ import java.lang.ref.WeakReference
  * previous sets from getting gc'ed in the event of a rollback. Only the plan object actually mutates the set.
  */
 class DefaultEditablePlanDriver(
-  private val adapter: PlanEditAdapter
+  val adapter: PlanEditAdapter
```
Mythicaeda, line 52: What was the driver for making this non-private? Does making it non-private cause Kotlin to auto-generate the `getAdapter` method used in `Procedure::deletePastCreations`?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
```diff
@@ -897,6 +899,36 @@ public Collection<ResourceType> getResourceTypes(final PlanId planId)
...
+          final var obj = $.asJsonObject();
+          final var id = new ActivityDirectiveId(obj.getInt("id"));
+          if (obj.isNull("source_scheduling_goal_id")) return null;
+          final var source_goal = obj.getInt("source_scheduling_goal_id");
+          if (obj.isNull("source_scheduling_goal_invocation_id")) return entry(id, new GoalId(source_goal, -1, Optional.empty()));
+          final Long source_invocation = (long) obj.getInt("source_scheduling_goal_invocation_id");
+          return entry(id, new GoalId(source_goal, -1, Optional.of(source_invocation)));
+        }
+    );
+    return Map.ofEntries(results.stream().filter(Objects::nonNull).toArray(Map.Entry[]::new));
```
Mythicaeda, line 929: Nonblocking: Is there a reason other than code style to build the map like this vs via declaring the map and using `data.getJsonArray("activity_directive").forEach()` to fill it out? Is it for the immutability of the Map returned by `Map.ofEntries`? No change is needed, I'm asking because the code stood out.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/GoalId.java
```diff
@@ -1,4 +1,4 @@
...
 
 import java.util.Optional;
```
Mythicaeda, line 4: Nonblocking because this could be split into its own work, but I'm bringing this up because your changes need to interact with this design choice: Why is `goalInvocationId` `Optional`? The only time it is not defined outside of tests is for `refreshSchedulingProcedureParameterTypes`, and more specifically for `SpecificationRepository::updateGoalParameterSchema`. My gut preference is to just continue threading `goalId`, `revision` down into this function rather than making a field that is **always** otherwise present `Optional`

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/AutoDeletionTests.java
```diff
@@ -0,0 +1,137 @@
...
+          """
+          export default function myGoal() {
+            return Goal.CoexistenceGoal({
+              forEach: ActivityExpression.ofType(ActivityTypes.BiteBanana),
+              activityTemplate: ActivityTemplates.GrowBanana({quantity: 1, growingDuration: Temporal.Duration.from({minutes:1})}),
+              startsAt:TimingConstraint.singleton(WindowProperty.START).plus(Temporal.Duration.from({ minutes : 5}))
+            })
+          }""";
+
+      edslId = hasura.createSchedulingSpecGoal(
```
Mythicaeda, line 46: Minor, nonblocking: Could you add a comment somewhere about how this goal is being used to test if the plan is empty? I want to make sure that we don't lose this understanding if we ever have to refactor these tests.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/AutoDeletionTests.java
```diff
@@ -0,0 +1,137 @@
...
+
+    assertEquals(1, activities.size());
+
+    assertTrue(activities.stream().anyMatch(
+        it -> Objects.equals(it.type(), "BiteBanana")
+    ));
+  }
+
+  @Test
+  void createsTwoActivitiesSteadyState_JustBefore() throws IOException {
```
Mythicaeda, line 101: Minor, nonblocking: Could you add a method-level comment on this and `createsOneActivitySteadyState_AtBeginning` that's going over what this is testing? Even after reading through and understanding the tests, I'm still finding myself squinting at the test name (mainly because of the `SteadyState`, though I know why you used it)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/AutoDeletionTests.java
```diff
@@ -0,0 +1,137 @@
...
+
+      assertEquals(2, activities.size());
+
+      assertTrue(activities.stream().anyMatch(
+          it -> Objects.equals(it.type(), "BiteBanana")
+      ));
+
+      assertTrue(activities.stream().anyMatch(
+          it -> Objects.equals(it.type(), "GrowBanana")
+      ));
```
Mythicaeda: Might be worth asserting three additional pieces of information:

1) that there is only one activity after the first run: `BiteBanana`
2) that the activityId of `BiteBanana` is never the same (that is, that the goal is placing a _new_ `BiteBanana` each loop)
3) that the activityId of `GrowBanana` is the same (that is, that the coex goal is not placing a new `GrowBanana` each loop)

If I had to pick only one of these, I'd pick (2), since that id changing proves the thesis that the prior activity is being deleted and replaced. Without that check, you would get the same results if the activity wasn't deleted and the goal was marked as satisfied because of the activity already in the plan (though I understand satisfaction works differently for Procedural Goals).

(1) and (2) also apply for `createsOneActivitySteadyState_AtBeginning`, though not as strongly, since the CoEx not placing anything is pretty strong evidence that the prior activity is being deleted.

Mythicaeda (commented):
Primarily nonblocking comments

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/GoalId.java
JoelCourtney: If the invocation is deleted in the database, an activity can end up where the goal id is non-null and the invocation id is null. So since the `GoalId` record is used in other places I figured we should represent that.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/AutoDeletionTests.java
JoelCourtney: Good point

procedural/scheduling/src/main/kotlin/gov/nasa/ammos/aerie/procedural/scheduling/utils/DefaultEditablePlanDriver.kt
JoelCourtney: Yeah that's exactly it. If its private then `Procedure` cant call `getAdapter`. It was definitely a hack, but I did this because I intended source goal/invocation ids on Directives to be a hidden thing that the user's goal can't access from the Plan interface. So `Procedure` had to reach deeper into the data structure to get the scheduler-internal plan representation that has the goal ids.

scheduler-driver/src/main/java/gov/nasa/jpl/aerie/scheduler/model/GoalId.java
JoelCourtney: Before this change the record was only used in the scheduler server, so scheduler-driver/worker didn't have access to it. It used to be that goal ids were non-behavioral bookkeeping that was only needed for db interaction. Now that we actually need it for scheduler behavior I had to expose it to the other packages.

JoelCourtney:
@Mythicaeda I caught a bug and pushed a fairly large commit to fix it. The sim facade checks if the plan has been simulated before so it can use cached results, but it (correctly) doesn't check for ID equality, only activity type/args/time equality. If the IDs are unequal (which is now common behavior with auto-deletion) the sim facade returns sim results with IDs that don't match the input plan. So I made the facade map the returned results to use the new ids from the input plan.

scheduler-worker/src/test/java/gov/nasa/jpl/aerie/scheduler/worker/services/SchedulingDSLCompilationServiceTests.java
JoelCourtney: No idea, since I didn't create either class. I didn't look very closely but just blindly substituting this with the official mock service makes a bunch of test failures, so 🤷

deployment/hasura/migrations/Aerie/18_activity_goal_invocation_id/down.sql
JoelCourtney: done 👍

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/services/GraphQLMerlinDatabaseService.java
JoelCourtney: Its just style. I tend to think in functional style when working with collections.

deployment/hasura/migrations/Aerie/18_activity_goal_invocation_id/down.sql
JoelCourtney: Actually, not done 👎 . Reversing the order of the triggers in the down migration causes pgcmp down to fail. No idea why.

JoelCourtney: ok done fr this time 👍

dandelany:
@Mythicaeda will review this for next release

scheduler-driver/src/test/java/gov/nasa/jpl/aerie/scheduler/SimulationFacadeTest.java
```diff
@@ -353,4 +356,21 @@ public void testActivityTypeWithResourceConstraint() throws SchedulingInterrupte
     assertTrue(TestUtility.containsExactlyActivity(plan, act2));
     assertTrue(TestUtility.doesNotContainActivity(plan, act1));
   }
+
+  @Test
+  public void testIdMapOnCachedPlan() throws SchedulingInterruptedException, SimulationFacade.SimulationException {
```
Mythicaeda, line 361: I'm not sure I follow how this test shows that the plan was not resimulated, but instead a cache was loaded with updated ids. Could you please explain it?

JoelCourtney: The test assumes that the plan wasn't resimulated, because I assumed that Adrien had already tested the caching appropriately. This just checks that the new sim results have the updated ID, and doesn't check whether that ID came from replacement or resimulation.

Mythicaeda: Ah, I see. I misread the bug and thought there was an issue in the caching, not the returned ids. You're right, just checking that the new ids are applied is correct for the test, since caching is tested elsewhere (ie `testOptionalSimulationAfterGoal_staleResources`)

Mythicaeda (approved):
LGTM!

PR #1624, opened 2025-01-03
Run backend e2e tests with given gateway docker tag when provided in PR body, add dev docs

### LINKED ISSUE #1538: Process + workflows for testing UI PRs against specified backend branch

dandelany:
```___REQUIRES_GATEWAY_PR___="116"```
### Description
Closes https://github.com/NASA-AMMOS/aerie/issues/1538

For reference see https://github.com/NASA-AMMOS/aerie-ui/pull/1439

This PR contains two things:
* An update to GH workflows to allow `aerie` backend PRs to specify an `aerie-gateway` PR/image to use for e2e tests - since, as noted in [this comment from @Mythicaeda](https://github.com/NASA-AMMOS/aerie/issues/1538#issuecomment-2322105332), the backend e2e tests may rely on changes in Gateway. This is an exact copy of the approach from https://github.com/NASA-AMMOS/aerie-ui/pull/1439 applied to the `aerie` backend e2e tests
* Updates to the CONTRIBUTING.md developer docs to explain how to use this process. Also added some notes on Required Checks in general since they weren't mentioned before.

### Verification

I tested this PR both with and without the `REQUIRES_GATEWAY_PR` tag at the top, and verified that when it is present, the e2e test action correctly pulls and runs the gateway image tagged as `pr-116` (shown in the workflow run logs)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - develop
       - dev-[0-9]+.[0-9]+.[0-9]+
+    types:
+      # defaults + 'edited', to handle changes to flags like ___REQUIRES_GATEWAY_PR___ in the PR body
+      [ opened, synchronize, reopened, edited ]
   push:
     branches:
       - develop
@@ -27,6 +30,7 @@ env:
   SCHEDULER_PASSWORD: "${{secrets.SCHEDULER_PASSWORD}}"
   SEQUENCING_USERNAME: "${{secrets.SEQUENCING_USERNAME}}"
   SEQUENCING_PASSWORD: "${{secrets.SEQUENCING_PASSWORD}}"
+  PR_BODY: "${{github.event.pull_request.body}}"
 
 jobs:
   unit-test:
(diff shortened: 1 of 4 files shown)


Mythicaeda:
~~I'd recommend using `___REQUIRES_GATEWAY_PR___="119"` instead, as that contains no behavior changes to the Gateway code.~~

Ah, I see why you're using 116, since it's already publishing an image.

Mythicaeda (approved):

PR #1628, opened 2025-01-28
Streamline system initialization refactor

DavidLegg:
DRAFT pending #1589 , on which this PR builds.

* **Tickets addressed:** N/A
* **Review:** By commit  

## Description
Refactors the initialization of streamline to an explicit call to StreamlineSystem.init. This new method is responsible for building all the singletons used by streamline, rather than putting that responsibility on the Registrar. This promotes a better separation of concerns, especially as we add more singleton-type functionality to initialize.

This also standardizes re-initialization behavior. Re-initializing an already-initialized system will simply rebuild all the singletons. While this would be wasteful if abused in production, it supports testing scenarios where multiple models may be built and re-built as tests are run.

Finally, after merging in #1589, this adds a globally-visible absolute clock, and relatedly the `StreamlineSystem.currentInstant()` method to get the current time as an Instant.

## Verification

Existing unit tests were updated to call StreamlineSystem.init, and they continue to function properly.
The streamline-demo model was also updated to call this method, and was tested manually in a local aerie deployment. It continues to function correctly as well.

## Documentation

IMPORTANT: This is a breaking change!
All projects that use streamline should change
```java
var registrar = new Registrar(aerieRegistrar, errorBehavior);
```
into
```java
StreamlineSystem.init(InitArgs.builder()
    .baseRegistrar(aerieRegistrar)
    .errorBehavior(errorBehavior)
    .planStart(planStart)
    .build());
var registrar = Registration.REGISTRAR;
```
when constructing their model. This will properly initialize the full streamline system.

Additionally, `currentTime()` has moved from `Resources` to `StreamlineSystem`.

Finally, while not a breaking change, this does expose `StreamlineSystem.currentInstant()`, which returns the current absolute time as an Instant. This is something I expect many might find useful.

## Future work

N/A

### DIFF
--- a/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/StreamlineSystem.java
+++ b/contrib/src/main/java/gov/nasa/jpl/aerie/contrib/streamline/StreamlineSystem.java
@@ -1,0 +1,52 @@
+package gov.nasa.jpl.aerie.contrib.streamline;
+
+import gov.nasa.jpl.aerie.contrib.streamline.core.Resource;
+import gov.nasa.jpl.aerie.contrib.streamline.debugging.Logging;
+import gov.nasa.jpl.aerie.contrib.streamline.modeling.Registrar;
+import gov.nasa.jpl.aerie.contrib.streamline.modeling.Registration;
+import gov.nasa.jpl.aerie.contrib.streamline.modeling.clocks.Clock;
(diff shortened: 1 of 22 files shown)


DavidLegg:
Closing this because I opened this against my fork of Aerie instead of AMMOS repo by mistake. See #1629 for correct PR.

PR #1630, opened 2025-01-28
Stream simulation results in their own thread

### LINKED ISSUE #1516: Stream Simulation Results in their own thread
Follow up to #1507.

Move streaming resources into its own thread in order to increase time performance of simulation.

joshhaug:
* **Tickets addressed:** Closes https://github.com/NASA-AMMOS/aerie/issues/1516.
* **Review:** By commit  

## Description

This should not be a breaking change. 

This PR introduces threaded uploads of simulation results to the database. This is a follow up to [this PR](https://github.com/NASA-AMMOS/aerie/pull/1507) in which streaming simulation results were introduced.  

### Implementation Details
* Creates a new `PostgresProfileQueryHandler` class which abstracts out database query logic that was previously in `PostgresProfileStreamer`. 
* Due to some database design decisions, sim result segments must be inserted in-order. 
  * Note that `PostgresProfileStreamer.accept()` calls will be in order.  
    * But, if we were to spin up each upload in a separate thread, the upload completion times for each of those calls may not be in-order.
  * To get around this, this PR creates a new `PostgresQueryQueue` which enqueues concurrent uploads using an `ExecutorService`.
* `PostgresProfileStreamer` creates a single instance of the query queue, which handles all uploads and is automatically closed. 

## Verification

Initially I had hoped to profile against the status prior to the [original streaming PR](https://github.com/NASA-AMMOS/aerie/pull/1507) but some version mismatches in clipper made this difficult. 

I instead used the `feature--aerie-3.2.0` branch of `clipper-aerie`. I then extended Clipper's `gnc-all-activities` plan by repeating its content every day for five days. I'm happy to make this plan available to anyone who would like to see it or use it for future reference, but I doubt I can upload it here in this public venue. I ran profiling against this extended `gnc-all-activities` twice, once with high-overhead memory instrumentation and once with async sampling. I encountered errors when extracting mission model parameters, but this apparently did not impact simulation results, and the error message in the console was not helpful. 

Here are the results from using async sampling:

| Branch Name                | Sim Dur | Heap Mem | Non-Heap Mem |
| -------------------------- | ------- | -------- | ------------ |
| `develop`                  | 2:48    | 1.68 GB  | 62.42 MB     |
| `feature/streaming-thread` | 2:20    | 1.62 GB  | 62.78 MB     |

Statistics for these results:
* 4% decrease in heap memory usage
* 1% increase in non-heap memory usage
* 16.6% speedup in sim duration

As expected, the primary gain here is in simulation duration, as we are no longer blocking simulation to upload results to the database. I'll attach some further sim results for anyone interested. 

[ProfilingResults.zip](https://github.com/user-attachments/files/18578611/ProfilingResults.zip)

## Documentation

No documentation needs updating.


merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
```diff
@@ -0,0 +1,190 @@
...
+    postSegmentsStatement = connection.prepareStatement(postSegmentsSql, PreparedStatement.NO_GENERATED_KEYS);
+    updateDurationStatement = connection.prepareStatement(updateDurationSql, PreparedStatement.NO_GENERATED_KEYS);
+  }
+
+  /**
+   * Upload profiles, profile segments, and corresponding profile durations to the database.
+   * */
+  public void uploadResourceProfiles(final ResourceProfiles resourceProfiles) {
+    try {
+      prepareStatements();
```
Mythicaeda: This method call should go back into the constructor, as prepared statements only need to be prepped once per `connection`. (As in, the `connection` field on the class, not each time that field is used)

joshhaug: Duh, good point. Updated in 4a820b8f72af92f1118d430c2ba0ab2a0f6fc54c.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
```diff
@@ -1,180 +1,37 @@
...
-          addDiscreteProfileToBatch(discreteEntry.getKey(), discreteEntry.getValue());
-        }
-      }
-      postProfiles();
-
-      // Post Segments
-      for(final var realEntry : resourceProfiles.realProfiles().entrySet()){
-        addProfileSegmentsToBatch(realEntry.getKey(), realEntry.getValue(), realDynamicsP);
+    queryQueue.addToQueue(() -> {
+      try (var transaction = new PostgresProfileQueryHandler(dataSource, datasetId)) {
```
Mythicaeda: Why make a new Query Handler per-transaction instead of having a single `handler` instance that is close in the `close` method? Asking as I'm concerned about running out of DB connections if the queue gets backed up, since the constructor for this class opens a new one.

joshhaug: Good question. Creating a single handler instance caused me to run into an `PSQLException: This statement has been closed` error. I believe this was due to the HikariCP lifecycle auto-closing the connection. The good news is that each PostgresProfileQueryHandler will be created as part of the queued task, so it won't be created until the previous one is completed. So we should only have one DB connection at a time. 

That said, I'm not sure of the overhead of creating "new" connections. I assumed HikariCP maintaining a connection pool meant that the overhead was low, but I could be wrong. There's likely a way of maintaining a single `handler` if we think the overhead is non-negligible.

Mythicaeda: Hikari does not auto-close the connection, otherwise we would have run into this issue with the `develop` implementation, alongside several other Hikari DB connections that have long lifespans with low activity.

[Let me amend this: if Hikari auto-closes a connection, it recreates one to refill the pool.](https://medium.com/@rajchandak1993/understanding-hikaricps-connection-pooling-behaviour-467c5a1a1506)

Mythicaeda: Additionally `this statement has been closed` isn't about the connection, it's about a `Statement` object, which has a lifespan independent of the `connection` object. Basically, when you use the `connection` to make a new `Statement`, the prior `statement` (and corresponding `ResultsSet`) is closed, and trying to access it gives the error you received.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
```diff
@@ -0,0 +1,190 @@
...
+public class PostgresProfileQueryHandler implements AutoCloseable {
+  private final Connection connection;
+  private final HashMap<String, Integer> profileIds;
+  private final HashMap<String, Duration> profileDurations;
+
+  private PreparedStatement postProfileStatement;
+  private PreparedStatement postSegmentsStatement;
+  private PreparedStatement updateDurationStatement;
+
+  private long datasetId;
```
Mythicaeda: This can be `final`

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
```diff
@@ -0,0 +1,190 @@
...
+ * own Connection object.
+ * */
+public class PostgresProfileQueryHandler implements AutoCloseable {
+  private final Connection connection;
+  private final HashMap<String, Integer> profileIds;
+  private final HashMap<String, Duration> profileDurations;
+
+  private PreparedStatement postProfileStatement;
+  private PreparedStatement postSegmentsStatement;
+  private PreparedStatement updateDurationStatement;
```
Mythicaeda: If `prepareStatements` was inlined into the constructor, these could be `final`.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.HashMap;
+
+import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
+import static gov.nasa.jpl.aerie.merlin.server.http.ProfileParsers.realDynamicsP;
+import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.discreteProfileTypeP;
+import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.realProfileTypeP;
+
+/**
+ * Utility class to handle upload of resource profiles to the database. This class allows for each upload to have its
+ * own Connection object.
```
Mythicaeda: Nonblocking Suggestion: remove the second part of this class comment, since "each upload having its own connection object" isn't an inherit property of the class (since the same instance can work for multiple uploads):
```suggestion
 * Utility class to handle upload of resource profiles to the database.
```

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
```diff
@@ -1,180 +1,37 @@
...
-  private final Connection connection;
-  private final HashMap<String, Integer> profileIds;
-  private final HashMap<String, Duration> profileDurations;
-
-  private final PreparedStatement postProfileStatement;
-  private final PreparedStatement postSegmentsStatement;
-  private final PreparedStatement updateDurationStatement;
+  private final DataSource dataSource;
+  private long datasetId;
+  private PostgresQueryQueue queryQueue;
```
Mythicaeda: These can be `final`

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresQueryQueue.java
```diff
@@ -0,0 +1,17 @@
+package gov.nasa.jpl.aerie.merlin.worker.postgres;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public class PostgresQueryQueue {
```
Mythicaeda: This class can be inlined into `PostgresProfileStreamer`, since it's just a wrapper around an `executor`, albeit one with most methods hidden. I'd recommend doing this unless you see a use elsewhere/feel it's important to hide these methods.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
joshhaug: After discussing this with @Mythicaeda offline, I'm convinced the auto-closing issue was due to a separate bug stemming from my usage of a diy query queue. This implementation never made it to commit level, as I decided to just move forward with an ExecutorService instead. 

Refactor to use a shared `queryHandler` has been implemented in 24221cb51036a46a1dc8b4a7774a9e033073279e.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
joshhaug: Done in 24221cb51036a46a1dc8b4a7774a9e033073279e,  good catch.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
joshhaug: Yep, done in 24221cb51036a46a1dc8b4a7774a9e033073279e.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileQueryHandler.java
joshhaug: Good idea. Implemented in 24221cb51036a46a1dc8b4a7774a9e033073279e.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresProfileStreamer.java
joshhaug: Made `queryQueue` a `final` member, and removed `datasetId` as it's no longer needed as a member, in 24221cb51036a46a1dc8b4a7774a9e033073279e.

merlin-worker/src/main/java/gov/nasa/jpl/aerie/merlin/worker/postgres/PostgresQueryQueue.java
joshhaug: Yep, I removed the class entirely in 24221cb51036a46a1dc8b4a7774a9e033073279e and just folded the logic straight into `PostgresProfileStreamer`.  Good point.

Mythicaeda (approved):

PR #1634, opened 2025-02-12
Aerie Actions Backend

goetzrrGit:
## Description
Aerie Action Server is a new Aerie Core service. This will be handling User Actions Logic currently be designed. This is a basic server atm with unit test hooked up. 

## Verification
Just make sure the service starts up when you start aerie via docker

## Future work
Hook up a way to execute `js` code written by the user via the endpoint


action-server/src/utils/codeRunner.ts
```diff
@@ -0,0 +1,61 @@
...
+      },
+      error: (...args: any[]) => {
+        consoleOutput.error.push(args.join(' '))
+      }
+  }
+
+  const context = vm.createContext(aerieGlobal)
+
+  try{
+    vm.runInContext(code, context);
```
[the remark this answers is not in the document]
dandelany: In case it's not obvious to readers - this is both a) intentional and b) absolutely a big security hole under normal "open to the public" web circumstances - this is the mechanism by which we allow our (trusted) users to do dangerous things by executing their arbitrary code.

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
```diff
@@ -0,0 +1,47 @@
...
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: "*"
+      check: {}
+      set:
+        created_by: "x-hasura-user-id"
+  - role: user
+    permission:
+      columns: "*"
```
Mythicaeda: I recommend replacing `columns` with a specific list of columns, as API users should _not_ be able to set, ie, the `status` or `created_at` columns on insert

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
```diff
@@ -0,0 +1,47 @@
...
+      check: {}
+      set:
+        created_by: "x-hasura-user-id"
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [logs, error, results, status]
+      filter: {}
```
Mythicaeda, line 43: Why are these columns editable? This comment is directed more towards the first three columns, as I can understand the admin updating `status` should, ie, the system go down while running an action.

deployment/postgres-init-db/sql/init.sql
```diff
@@ -28,7 +28,7 @@ begin;
   -- UI
   \ir init_ui.sql
 
-  -- Merlin
+-- Merlin
```
Mythicaeda: Minor: Spacing got accidentally tweaked

deployment/hasura/migrations/Aerie/14_actions/up.sql
```diff
@@ -0,0 +1,162 @@
...
+comment on column actions.action_run.parameters is e''
+  'The supplied parameters for the run of the action.';
+comment on column actions.action_run.logs is e''
+  'The logs produced by the action run.';
+comment on column actions.action_run.error is e''
+  'The error produced by the action run.';
+comment on column actions.action_run.results is e''
+  'The results produced by the action run.';
+comment on column actions.action_run.status is e''
+  'The status of the action run. pending -> in-progress -> failed or complete';
```
Mythicaeda: I'd recommend removing the part after the first sentence.
```suggestion
  'The status of the action run.
```

deployment/hasura/migrations/Aerie/14_actions/up.sql
```diff
@@ -0,0 +1,162 @@
...
+                 parameters,
+                 action_definition_id,
+                 workspace_id,
+                 action_file_path) as
+           (
+             select NEW.settings,
+                    NEW.parameters,
+                    NEW.action_definition_id,
+                    ad.workspace_id,
+                    uf.name
```
Mythicaeda: I'd recommend changing this code to match how we get the path for other uploaded files in the backend: `encode(uf.path, 'escape') as path,`.

This also applies to `notify_action_definition_inserted`

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  // todo: set schemas on the DB row?
+  const pool = ActionsDbManager.getDb();
+  const query = `
+    UPDATE actions.action_definition
+    SET
+      parameter_schema = parameter_schema || $1::jsonb,
+      settings_schema = settings_schema || $2::jsonb
+    WHERE id = $3
+      RETURNING *;
+  `;
```
goetzrrGit: @dandelany I believe we are setting the schema in the DB with the query below. Is this TODO needed?

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  console.log(`action run ${actionRunId} inserted (${actionFilePath})`);
+  console.info(payload);
+  // event payload contains a file path for the action file which should be run
+  const actionJS = await readFileFromStore(actionFilePath);
+
+  // TODO: how to handle auth tokens??
+  // const authToken = req.header("authorization");
+  // if (!authToken) console.warn("No valid `authorization` header in action-run request");
+
+  // todo: run the action file, put results in the same DB row and mark status as successful
```
goetzrrGit: This TODO is complete as well

action-server/src/threads/worker.ts
```diff
@@ -0,0 +1,38 @@
+import { threadId } from "worker_threads";
+import { jsExecute } from "../utils/codeRunner";
+import { ActionResponse, ActionTask } from "../type/types";
+import { ActionsDbManager } from "../db";
+
+//export async function add({n} : { n : number}) :Promise<number> { debugger;return n; }
+
```
goetzrrGit: We can remove this as this was to test the initial picscina hookup and make sure it was working with a simple function

action-server/tests/codeRunner.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    const result: ActionResponse = await jsExecute(code, {}, {}, undefined, {} as PoolClient, 1);
+
+    assert(result.console[0].includes("[INFO] Hello World"));
+
+    // Check console outputs
+    // assert.strictEqual(result.console.log.join(), "Hello World");
+    // assert.strictEqual(result.console.debug.join(), "Debug message");
+    // assert.strictEqual(result.console.info.join(), "Info message");
+    // assert.strictEqual(result.console.warn.join(), "Warning message");
+    // assert.strictEqual(result.console.error.join(), "Error message");
```
goetzrrGit: Should we check if the other messages are within the console array?

goetzrrGit (commented):
Very minor things I noticed.

action-server/src/listeners/dbListeners.ts
Mythicaeda: Better question: why is this query merging the schemas? When we update a schema, we _replace_ the old value with the new one. If you merge it, then you run into issues where, for example, a parameter is dropped, but that parameter is still in the schema. 

Please reference `CreateModelParametersAction.java` and `UpdateConstraintParametersAction.java` for examples.

.env.template
```diff
@@ -13,6 +13,9 @@ SCHEDULER_PASSWORD=
 SEQUENCING_USERNAME=sequencing_user
 SEQUENCING_PASSWORD=
 
+ACTION_USERNAME=action_user
+ACTION_PASSWORD=
```
Mythicaeda: These ought to be removed from the template, as currently they are unused.

deployment/docker-compose.yml
```diff
@@ -1,5 +1,32 @@
...
+    environment:
+      HASURA_GRAPHQL_ADMIN_SECRET: "${HASURA_GRAPHQL_ADMIN_SECRET}"
+      LOG_FILE: console
+      LOG_LEVEL: debug
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      ACTION_SERVER_PORT: 27186
+      AERIE_DB_HOST: postgres
+      AERIE_DB_PORT: 5432
+      ACTION_DB_USER: "${ACTION_USERNAME}"
+      ACTION_DB_PASSWORD: "${ACTION_PASSWORD}"
```
Mythicaeda: Bug: These need to point to the `SEQUENCING` version of these envvars, as the `ACTION` version is not a real PG user

.github/workflows/pgcmp.yml
```diff
@@ -35,6 +35,8 @@ env:
   SCHEDULER_PASSWORD: "${{secrets.SCHEDULER_PASSWORD}}"
   SEQUENCING_USERNAME: "${{secrets.SEQUENCING_USERNAME}}"
   SEQUENCING_PASSWORD: "${{secrets.SEQUENCING_PASSWORD}}"
+  ACTION_USERNAME: "${{secrets.ACTION_USERNAME}}"
+  ACTION_PASSWORD: "${{secrets.ACTION_PASSWORD}}"
```
Mythicaeda: These ought to be removed when `ACTION_USERNAME` and  `ACTION_PASSWORD` are removed from the `.env.template`

deployment/.env
```diff
@@ -18,6 +18,9 @@ SCHEDULER_PASSWORD=
 SEQUENCING_USERNAME=sequencing_service
 SEQUENCING_PASSWORD=
 
+ACTION_USERNAME=action_service
+ACTION_PASSWORD=
+
```
Mythicaeda: This should be removed. See comment on the `.env.template`

action-server/eslint.config.mjs
```diff
@@ -0,0 +1,20 @@
...
+  "jest": false,
+  "vitest": false
+};
+
+export default tseslint.config(sheriff(sheriffOptions),
+    {
+      rules: {
+        "@typescript-eslint/no-explicit-any": ["off"]
+      },
+    },
```
Mythicaeda, line 19: I'm not that familiar with `eslint`. Is this turning off all of the linting settings?

action-server/src/app.ts
```diff
@@ -0,0 +1,31 @@
+import express from "express";
+import { configuration } from "./config";
+import { corsMiddleware, jsonErrorMiddleware } from "./middleware";
+import { ActionWorkerPool } from "./threads/workerPool";
+import { cleanup, setupListeners } from "./listeners/dbListeners";
+
+const port = configuration().PORT;
+
+// init express app and middleware
+const app = express();
```
Mythicaeda, line 10: Blocking: Please add a `/health` endpoint. Clipper has encountered issues with deployed containers being restarted by their container manager when a container doesn't provide this endpoint. The endpoint can literally just return a `200` status. You can use L153 of `sequencing-server/src/app.ts` for reference.

Additionally, the E2E Health tests will look for this endpoint.

deployment/docker-compose.yml
```diff
@@ -1,5 +1,32 @@
...
+      ACTION_LOCAL_STORE: /usr/src/app/action_file_store
+      ACTION_WORKER_NUM: 8
+      ACTION_MAX_WORKER_NUM: 8
+    image: aerie_action
+    ports: [ "27186:27186" ]
```
Mythicaeda, line 22: Please add the following line to `BaseURL.java` in order to update the E2E Health Tests:

```
ACTION_SERVER("http://localhost:27186"),
```

action-server/src/config.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const configuration = (): Config => {
+  const { env } = process;
+
+  return {
+    AERIE_DB: env.AERIE_DB ?? "aerie",
+    AERIE_DB_HOST: env.AERIE_DB_HOST ?? "postgres",
+    AERIE_DB_PORT: env.AERIE_DB_PORT ?? "5432",
+    ACTION_DB_USER: env.ACTION_DB_USER ?? "postgres",
+    ACTION_DB_PASSWORD: env.ACTION_DB_PASSWORD ?? "password",
```
Mythicaeda: Credentials should always default to an empty string (see L30 of this file):
```suggestion
    ACTION_DB_USER: env.ACTION_DB_USER ?? "",
    ACTION_DB_PASSWORD: env.ACTION_DB_PASSWORD ?? "",
```

action-server/src/db.ts
```diff
@@ -0,0 +1,43 @@
+import type { Pool, PoolConfig } from "pg";
+import pg from "pg";
+import { configuration } from "./config";
+// import getLogger from './utils/logger.js';
+
+const { Pool: DbPool } = pg;
+
+const { AERIE_DB, AERIE_DB_HOST, AERIE_DB_PORT, ACTION_DB_USER, ACTION_DB_PASSWORD } = configuration();
+// const logger = getLogger('packages/db/db');
```
Mythicaeda: Minor: Please remove commented out code.

action-server/src/db.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  static getDb(): Pool {
+    return ActionsDbManager.pool;
+  }
+
+  static init() {
+    try {
+      const config: PoolConfig = {
+        host: AERIE_DB_HOST,
+        port: parseInt(AERIE_DB_PORT, 10),
```
Mythicaeda: Bug: the default value for `port` should be `5432`, not `10`.

Additionally, please set the pool size to the minimum amount of needed connections. Since three threads need to be connected to the DB at once (the listen thread, the definition update thread, and the execution thread), this should be at least 3.

action-server/src/db.ts
```diff
@@ -0,0 +1,43 @@
...
+      };
+
+      logger.info(`Creating PG pool`);
+      logger.debug(`
+      {
+         database: ${config.database},
+         host: ${config.host},
+         port: ${config.port},
+         user: ${config.user}
+      }`);
```
Mythicaeda: Major: Do **_not_** log credentials, even to the `debug` channel. We have in the past had to remove this exact log statement from the Gateway.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+async function readFileFromStore(fileName: string): Promise<string> {
+  // read file from aerie file store and return [resolve] it as a string
+  const fileStoreBasePath = configuration().ACTION_LOCAL_STORE;
+  const filePath = path.join(fileStoreBasePath, fileName);
+  console.log(`path is ${filePath}`);
+  return await readFile(filePath, "utf-8");
+}
+
+async function handleActionDefinition(payload: ActionDefinitionInsertedPayload) {
+  console.log("action definition inserted");
```
Mythicaeda: Minor: This log shouldn't go _before_ the success it's reporting. Additionally, this `console.log` and L29's `console.info` look like debugging statements and should probably be removed.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // read the action file and extract parameter/setting schemas
+  const actionJS = await readFileFromStore(payload.action_file_path);
+  // console.debug(actionJS);
```
Mythicaeda: Please remove commented out code.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  try {
+    const res = await pool.query(query, [
+      JSON.stringify(schemas.paramDefs),
+      JSON.stringify(schemas.settingDefs),
+      payload.action_definition_id,
+    ]);
+    console.log("Updated action_definition:", res.rows[0]);
+  } catch (error) {
+    console.error("Error updating row:", error);
```
Mythicaeda: Since these seem like proper logs and not debugging statements, why are these using `console` instead of the logger? This applies to L71, L83, L89, and L90 as well.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+
+async function readFileFromStore(fileName: string): Promise<string> {
+  // read file from aerie file store and return [resolve] it as a string
+  const fileStoreBasePath = configuration().ACTION_LOCAL_STORE;
+  const filePath = path.join(fileStoreBasePath, fileName);
+  console.log(`path is ${filePath}`);
+  return await readFile(filePath, "utf-8");
+}
+
+async function handleActionDefinition(payload: ActionDefinitionInsertedPayload) {
```
Mythicaeda: Nonblocking: I suggest renaming this to match the pattern of similar endpoints in Aerie: 

```suggestion
async function refreshActionDefinitionSchema(payload: ActionDefinitionInsertedPayload) {
```

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+      JSON.stringify(schemas.settingDefs),
+      payload.action_definition_id,
+    ]);
+    console.log("Updated action_definition:", res.rows[0]);
+  } catch (error) {
+    console.error("Error updating row:", error);
+  }
+}
+
+async function handleActionRun(payload: ActionRunInsertedPayload) {
```
Mythicaeda: Nonblocking: Suggestion:

```suggestion
async function runAction(payload: ActionRunInsertedPayload) {
```

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  const actionRunId = payload.action_run_id;
+  const actionFilePath = payload.action_file_path;
+  console.log(`action run ${actionRunId} inserted (${actionFilePath})`);
+  console.info(payload);
```
Mythicaeda: Debug statements should be removed.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  // event payload contains a file path for the action file which should be run
+  const actionJS = await readFileFromStore(actionFilePath);
+
+  // TODO: how to handle auth tokens??
+  // const authToken = req.header("authorization");
+  // if (!authToken) console.warn("No valid `authorization` header in action-run request");
```
Mythicaeda: You cannot handle Auth tokens while using Postgres Listen/Notify. If you need to access the auth token, you need a Hasura Action, Hasura Event, or Hasura Function.

I'd recommend replacing/updating this TODO with the above explanation.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // todo: run the action file, put results in the same DB row and mark status as successful
+  // todo: try/catch - need to handle errors manually since not in express handler?
```
Mythicaeda: This TODO looks complete? I'd keep the `Need to handle errors manually since not in express handler` as a comment above the `try` block

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  const { parameters, settings } = payload;
+  const workspaceId = payload.workspace_id;
+  const pool = ActionsDbManager.getDb(); // cant seralize pool as there is data that is unserializable DOMException: DataCloneError
```
Mythicaeda: What is the comment on this line referencing?

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+    console.error("Error running task:", error);
+    taskError = { message: error.message, stack: error.stack };
+  }
+
+  const duration = Math.round(performance.now() - start);
+  const status = taskError || run?.errors ? "failed" : "complete";
+  console.log(`Finished run ${actionRunId} in ${duration / 1000}s - ${status}`);
+  console.info(run);
+
+  const logStr = run ? run.console.join("\n") : "";
```
Mythicaeda: Why is the `run` info initially logged unformatted? Is L90 meant to be a debug statement?

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+      `
+      UPDATE actions.action_run
+      SET
+        status = $1,
+        error = $2::jsonb,
+        results = $3::jsonb,
+        logs = $4,
+        duration = $5
+      WHERE id = $6
+        RETURNING *;
```
Mythicaeda, line 98: Minor: Do you really need `returning *`? I see you have a debug `console.log` that prints the whole row after the update, but I'm not sure that's needed. This also applies to L39.

docker-compose.yml
```diff
@@ -1,5 +1,32 @@
...
+      LOG_LEVEL: debug
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      PORT: 27186
+      AERIE_DB_HOST: postgres
+      AERIE_DB_PORT: 5432
+      ACTION_DB_USER: "${SEQUENCING_USERNAME}"
+      ACTION_DB_PASSWORD: "${SEQUENCING_PASSWORD}"
+      ACTION_LOCAL_STORE: /usr/src/app/action_file_store
+      ACTION_WORKER_NUM: 8
+      ACTION_MAX_WORKER_NUM: 8
```
Mythicaeda: I thought the Action Server was supposed to be single-threaded initially? Why does it have a Worker configuration?

docker-compose.yml
```diff
@@ -17,6 +44,7 @@ services:
       AERIE_DB_PORT: 5432
       GATEWAY_DB_USER: "${GATEWAY_USERNAME}"
       GATEWAY_DB_PASSWORD: "${GATEWAY_PASSWORD}"
+      FILE_PATH: /app/files
```
Mythicaeda: Please remove this change due to changes done on the Gateway PR

docker-compose.yml
```diff
@@ -221,7 +249,7 @@ services:
     container_name: aerie_hasura
     depends_on: ["postgres"]
     environment:
-      AERIE_DATABASE_URL: "postgres://${AERIE_USERNAME}:${AERIE_PASSWORD}@postgres:5432/aerie?options=-c%20search_path%3Dutil_functions%2Chasura%2Cpermissions%2Ctags%2Cmerlin%2Cscheduler%2Csequencing%2Cpublic"
+      AERIE_DATABASE_URL: "postgres://${AERIE_USERNAME}:${AERIE_PASSWORD}@postgres:5432/aerie?options=-c%20search_path%3Dutil_functions%2Chasura%2Cpermissions%2Ctags%2Cmerlin%2Cscheduler%2Csequencing%2Cpublic%2Cactions"
```
Mythicaeda: `actions` and `public` should be reversed, otherwise you're telling Hasura to let objects in the `public` schema overshadow objects in the `actions` schema when the schema isn't explicitly specified (this shouldn't _currently_ cause a bug as we keep our `public` schema empty atm, but it is not advised).

```suggestion
      AERIE_DATABASE_URL: "postgres://${AERIE_USERNAME}:${AERIE_PASSWORD}@postgres:5432/aerie?options=-c%20search_path%3Dutil_functions%2Chasura%2Cpermissions%2Ctags%2Cmerlin%2Cscheduler%2Csequencing%2Cactions%2Cpublic"
```

docker-compose.yml
```diff
@@ -255,6 +283,8 @@ services:
       SCHEDULER_DB_PASSWORD: "${SCHEDULER_PASSWORD}"
       SEQUENCING_DB_USER: "${SEQUENCING_USERNAME}"
       SEQUENCING_DB_PASSWORD: "${SEQUENCING_PASSWORD}"
+      ACTION_DB_USER: "${ACTION_USERNAME}"
+      ACTION_DB_PASSWORD: "${ACTION_PASSWORD}"
```
Mythicaeda: Please remove unneeded envvars from the `postgres` container.

deployment/docker-compose.yml
```diff
@@ -1,5 +1,32 @@
 version: "3.7"
 services:
+  aerie_action:
+    build:
+      context: ./action-server
+      dockerfile: Dockerfile
+      #dockerfile: Dockerfile.dev #-- Uncomment for development
```
Mythicaeda: This chunk needs to be removed, as this is the `deployment` compose file and should use `image` instead (see the `aerie_merlin` container in this file)

docker-compose.yml
```diff
@@ -1,5 +1,32 @@
 version: "3.7"
 services:
+  aerie_action:
```
Mythicaeda, line 3: Major: This container needs to be added to the other `docker-composes`: `docker-compose-many-workers.yaml`, `docker-compose-proxy.yaml`, and `docker-compose-test.yaml`. Additionally, this container needs Kubernetes files (`/deployment/kubernetes`)

docker-compose.yml
```diff
@@ -1,5 +1,32 @@
...
+      dockerfile: Dockerfile
+      #dockerfile: Dockerfile.dev #-- Uncomment for development
+    container_name: aerie_action
+    depends_on: [ "postgres" ]
+    environment:
+      HASURA_GRAPHQL_ADMIN_SECRET: "${HASURA_GRAPHQL_ADMIN_SECRET}"
+      LOG_FILE: console
+      LOG_LEVEL: debug
+      MERLIN_GRAPHQL_URL: http://hasura:8080/v1/graphql
+      PORT: 27186
```
Mythicaeda: Why is `PORT` configurable? The action server will **always** have`27186` available to it within the container, and the `ports` tag on L24 is where the host machine picks a port to communicate with the container on.

For reference, see L216 of this file, the `ports` configuration for `scheduler_worker_2`: `ports: ["5010:5005", "27190:8080"]`. Within the container, port `5005` is open for virtual debugging. The `5010` part tells the host machine to open up port `5010` and to forward all requests it receives on that port to the containers `5005` port. [Here's some additional docs on the concept from Docker.](https://docs.docker.com/get-started/docker-concepts/running-containers/publishing-ports/)

action-server/src/threads/worker.ts
```diff
@@ -0,0 +1,38 @@
...
+export async function runAction(task: ActionTask): Promise<ActionResponse> {
+  console.log(`Worker [${threadId}] running task`);
+  console.info(`Parameters: ${JSON.stringify(task.parameters, null, 2)}`);
+  console.info(`Settings: ${JSON.stringify(task.parameters, null, 2)}`);
+
+  // console.debug(task);
```
Mythicaeda: Commented out code should be removed.

action-server/src/db.ts
```diff
@@ -0,0 +1,43 @@
...
+const logger = console;
+
+export class ActionsDbManager {
+  private static pool: Pool;
+
+  static getDb(): Pool {
+    return ActionsDbManager.pool;
+  }
+
+  static init() {
```
Mythicaeda: Major: This `init` code is being called by each thread that wants a connection (`dbListeners.ts` and `worker.ts`). This is an incorrect pattern: the purpose of having a connection pool is that there is _one_ pool and each thread that wants a connection requests a thread from said pool.

The Gateway code demonstrates the correct pattern: in the `main` method that starts up the container, the pool is initialized _once_. Endpoints that need a DB connection then call `getDb` to access the pool.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+    );
+    console.log("Updated action_run:", res.rows[0]);
+  } catch (error) {
+    console.error("Error updating row:", error);
+  }
+}
+
+export async function setupListeners() {
+  // initialize a database connection pool
+  ActionsDbManager.init();
```
Mythicaeda: These two lines should be removed.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+  console.log("Initialized PG event listeners");
+}
+
+export function cleanup(server: http.Server) {
+  console.log("shutting down...");
+  if (listenClient) {
+    listenClient.release();
+  }
+  server.close(() => {
+    console.log("server closed.");
```
Mythicaeda: Please either cleanup or properly log these debugging statements.

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+      return;
+    }
+    const payload = JSON.parse(msg.payload);
+
+    if (msg.channel === "action_definition_inserted") {
+      // todo should these be awaited?
+      await handleActionDefinition(payload);
+    } else if (msg.channel === "action_run_inserted") {
+      await handleActionRun(payload);
+    }
```
Mythicaeda: In answer to this `TODO`: `handleActionDefinition`, and `handleActionRun` should be putting jobs in two separate thread pools and not awaiting their completion. The `handleActionRun` queue should be a thread pool of size 1, as @dandelany specified in the demo that the Action Server will run only one job at a time. The `handleActionDefinition` queue _can_ included a thread pool of a larger size, as the server can handle multiple action definition updates simultaneously (if this is chosen, I'd recommend making the thread pool at most `DB connection pool size - 2`, to ensure that the `action run` thread pool always has a connection available to it (the listener is also always using a connection)). 

With the current code, I would expect that a long-running action will block the ability to update Action Definition schemas until it completes (easy test: create an action that includes some kind of "wait 2 mins" and then, while its running, attempt to upload a new Action and observe if the schema is updated while the action is running or only after it completes).

action-server/src/listeners/dbListeners.ts
```diff
@@ -0,0 +1,163 @@
...
+
+async function handleActionDefinition(payload: ActionDefinitionInsertedPayload) {
+  console.log("action definition inserted");
+
+  // read the action file and extract parameter/setting schemas
+  const actionJS = await readFileFromStore(payload.action_file_path);
+  // console.debug(actionJS);
+  const schemas = await extractSchemas(actionJS);
+
+  console.info(`schemas ${JSON.stringify(schemas, null, 2)}`);
```
Mythicaeda: Please remove debug print statement.

action-server/src/threads/worker.ts
```diff
@@ -0,0 +1,38 @@
...
+  ActionsDbManager.init();
+  const pool = ActionsDbManager.getDb();
```
Mythicaeda: Aside from L14 needing to be removed, as the DB pool should not be re-initializing the DBPool per-action run, can you please clarify why the action is being granted a DB connection?

action-server/src/type/types.ts
```diff
@@ -0,0 +1,59 @@
...
+      errors: null;
+    }
+  | {
+      results: null;
+      console: string[];
+      errors: {
+        stack: string | undefined;
+        message: string;
+        cause: unknown;
+      }; // TODO: should this be an error array
```
Mythicaeda: Unaddressed `TODO` that looks significant.

action-server/src/utils/codeRunner.ts
```diff
@@ -0,0 +1,121 @@
+import * as vm from "node:vm";
+import type { ActionResponse } from "../type/types";
+import { Actions } from "aerie-actions/dist/helpers";
+import { PoolClient } from "pg";
+import { createLogger, format, transports } from "winston";
+
+// todo put this inside a more limited closure scope or it will get reused...
+// const logBuffer: string[] = [];
```
Mythicaeda, line 8: There are a lot of leftover TODOs in this file. Please review and remove the ones that have been addressed.

action-server/tests/codeRunner.test.ts
Mythicaeda: Yes, and the test should assert that there are no more messages after the final one.

deployment/postgres-init-db/sql/tables/actions/action_definition.sql
```diff
@@ -0,0 +1,84 @@
+create table actions.action_definition (
+  id integer generated always as identity,
```
Mythicaeda, line 2: Column is missing comment.

deployment/postgres-init-db/sql/tables/actions/action_definition.sql
```diff
@@ -0,0 +1,84 @@
...
+  settings_schema jsonb not null default '{}'::jsonb,
+  settings jsonb not null,
+
+  action_file_id integer not null,
+  workspace_id integer not null,
+
+  created_at timestamptz not null default now(),
+  owner text,
+  updated_at timestamptz not null default now(),
+  updated_by text,
```
Mythicaeda, line 16: Columns are missing comments.

deployment/postgres-init-db/sql/tables/actions/action_definition.sql
```diff
@@ -0,0 +1,84 @@
...
+
+  constraint action_definition_synthetic_key
+    primary key (id),
+
+  foreign key (workspace_id)
+    references sequencing.workspace (id)
+    on delete set null,
```
Mythicaeda: Bug: this foreign key will cause issues, as `workspace_id` is a `not null` column. I'm assuming that actions can only exist within the concept of a specific workspace. Therefore, this should be: 

```suggestion
  foreign key (workspace_id)
    references sequencing.workspace (id)
    on delete cascade,
```

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
+create table actions.action_run (
+  id integer generated always as identity,
```
Mythicaeda, line 2: Column is missing comment.

deployment/postgres-init-db/sql/types/actions/action-types.sql
Mythicaeda, line 1: This type is 1-1 equivalent with  `util_functions.request_status` and all usages should be replaced with that type.

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
...
+
+  settings jsonb not null,
+  parameters jsonb not null,
+  logs text,
+  error jsonb,
+  results jsonb,
```
Mythicaeda: Why is `null` permitted as opposed to `not null default '{}'`, like we do with our other JSONB columns (see `action_definition.sql`).

deployment/postgres-init-db/sql/tables/actions/action_definition.sql
```diff
@@ -0,0 +1,84 @@
+create table actions.action_definition (
+  id integer generated always as identity,
+
+  name text not null,
+  description text null,
+  parameter_schema jsonb not null default '{}'::jsonb,
+  settings_schema jsonb not null default '{}'::jsonb,
+  settings jsonb not null,
```
Mythicaeda: This should include `default '{}'::jsonb`

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
...
+
+  created_by text,
+  created_at timestamptz not null default now(),
+  duration integer,
+
+  constraint action_run_synthetic_key
+    primary key (id),
+  foreign key (action_definition_id)
+    references actions.action_definition (id)
+    on delete set null,
```
Mythicaeda: Bug: `action_definition_id` is `not null`. That makes sense, as what is the purpose of an action run without a corresponding action? This should be `on delete cascade`
```suggestion
  foreign key (action_definition_id)
    references actions.action_definition (id)
    on delete cascade,
```

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
...
+  parameters jsonb not null,
+  logs text,
+  error jsonb,
+  results jsonb,
+  status actions.action_run_status not null default 'pending',
+
+  action_definition_id integer not null,
+
+  created_by text,
+  created_at timestamptz not null default now(),
```
Mythicaeda: Please rename to `requested_by` and `requested_at` to be consistent with similar tables in the DB. Additionally, these columns are missing comments.

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
...
+comment on column actions.action_run.error is e''
+  'The error produced by the action run.';
+comment on column actions.action_run.results is e''
+  'The results produced by the action run.';
+comment on column actions.action_run.status is e''
+  'The status of the action run.';
+comment on column actions.action_run.duration is e''
+  'The duration of the action run, if it has completed; null otherwise';
+comment on column actions.action_run.action_definition_id is e''
+  'The ID of the definition of the action.';
```
Mythicaeda: Nitpick: Please order your comments to be in the same order that the columns are declared on the table. This makes them easier to maintain.

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -0,0 +1,79 @@
...
+                    encode(uf.path, 'escape') as path
+             from actions.action_definition ad
+                    left join merlin.uploaded_file uf on uf.id = ad.action_file_id
+                    where ad.id = NEW.action_definition_id
+           )
+    select pg_notify('action_run_inserted', json_strip_nulls(row_to_json(payload))::text)
+    from payload
+  );
+  return null;
+end$$;
```
Mythicaeda, line 80: Nitpick, ignorable: while some of our older functions use `end$$;`, our styling for new functions is actually to have them on separate lines: 
```sql
end
$$;
```

deployment/hasura/metadata/databases/tables/actions/action_definition.yaml
```diff
@@ -0,0 +1,47 @@
+table:
+  name: action_definition
+  schema: actions
+configuration:
+  custom_name: "action_definition"
+object_relationships:
+  - name: workspace
+    using:
+      foreign_key_constraint_on: workspace_id
+  - name: uploaded_file
```
Mythicaeda: This relationship should be renamed to something meaningful, for example, `action_file`.

deployment/hasura/metadata/databases/tables/actions/action_definition.yaml
```diff
@@ -0,0 +1,47 @@
...
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: "*"
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: "*"
```
Mythicaeda: Major: Column list should not be `"*"`, as we do not, for example, want the user to specify `id`. I'd recommend the following column list:

```suggestion
      columns: [name, description, action_file_id, workspace_id]
```

`parameter_schema` and `settings_schema` will be set by the server after insertion. `settings` requires `settings_schema` to be set before it can have a value. `owner` and `updated_by` are set by the `set` block on L34, and `created_at` and `updated_at` will be prefilled with the current time.

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
```diff
@@ -0,0 +1,47 @@
+table:
+  name: action_run
+  schema: actions
+configuration:
+  custom_name: "action_run"
+object_relationships:
+  - name: action_definition
```
Mythicaeda, line 7: Nonblocking: rename suggestion 
```suggestion
  - name: action
```

action-server/src/listeners/dbListeners.ts
dandelany: Good catch @Mythicaeda - updated this to replace instead of merging, thanks

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
```diff
@@ -0,0 +1,47 @@
...
+      allow_aggregations: true
+  - role: viewer
+    permission:
+      columns: "*"
+      filter: {}
+      allow_aggregations: true
+insert_permissions:
+  - role: aerie_admin
+    permission:
+      columns: "*"
```
Mythicaeda: `aerie_admin` should also have insert column restrictions (see comment on `action_definition` about not letting them set, for example, `id`). I recommend the following:

```suggestion
      columns: [settings, parameters, action_definition_id]
```

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
```diff
@@ -0,0 +1,47 @@
...
+      check: {}
+      set:
+        created_by: "x-hasura-user-id"
+  - role: user
+    permission:
+      columns: [settings, parameters]
```
Mythicaeda: Bug: how can the `user` role make an action run request if it cannot specify what action it wants to run?

```suggestion
      columns: [settings, parameters, action_definition_id]
```

deployment/hasura/migrations/Aerie/14_actions/up.sql
```diff
@@ -0,0 +1,188 @@
...
+
+DO $$
+  DECLARE seq_user text;
+  BEGIN
+    SELECT into seq_user grantee
+    FROM information_schema.role_table_grants
+    WHERE table_schema = 'sequencing'
+      AND table_name = 'user_sequence'
+      and privilege_type = 'INSERT'
+      and grantee != (select current_user as usersss)
```
Mythicaeda: Minor: please remove the `as usersss` part. The alias isn't used, and also I just added all the s's to avoid getting distracted while working out the function, not as an acceptable alias in published code:

```suggestion
      and grantee != (select current_user)
```

deployment/hasura/migrations/Aerie/14_actions/up.sql
```diff
@@ -0,0 +1,188 @@
...
+    limit 1;
+
+    EXECUTE format('grant create, usage on schema actions to %I', seq_user);
+    EXECUTE format('grant select, insert, update, delete on all tables in schema actions to %I', seq_user);
+    EXECUTE format('grant execute on all routines in schema actions to %I', seq_user);
+    EXECUTE format('alter default privileges in schema actions grant select, insert, update, delete on tables to %I', seq_user);
+    EXECUTE format('alter default privileges in schema actions grant execute on routines to %I', seq_user);
+  END $$;
```
Mythicaeda: Please recapitalize to be consistent with the rest of our DB code. I also shifted the `into seq_user` to the bottom of the `select` block to be consistent with how we usually write `select into`s, but where it was is also fine.

```sql
do $$
  declare
    seq_user text;
  begin
    select grantee
    from information_schema.role_table_grants
    where table_schema = 'sequencing'
      and table_name = 'user_sequence'
      and privilege_type = 'INSERT'
      and grantee != (select current_user)
    limit 1
    into seq_user;

    execute format('grant create, usage on schema actions to %I', seq_user);
    execute format('grant select, insert, update, delete on all tables in schema actions to %I', seq_user);
    execute format('grant execute on all routines in schema actions to %I', seq_user);
    execute format('alter default privileges in schema actions grant select, insert, update, delete on tables to %I', seq_user);
    execute format('alter default privileges in schema actions grant execute on routines to %I', seq_user);
  end 
$$;
```

Mythicaeda:
I noticed there are 0 E2E Tests added. Please add even just a basic test of the workflow.

action-server/tests/codeRunner.test.ts
Mythicaeda: Unresolving because I added a comment to the thread while @goetzrrGit's feedback was being addressed and wanted to make sure they didn't get missed. If you feel the changes address my additions, then feel free to re-resolve

action-server/eslint.config.mjs
goetzrrGit: This one is turning off the warning that we are using `any` as a type. We are okay using `any` as a type

action-server/src/listeners/dbListeners.ts
goetzrrGit: This note explains why we cannot pass a pool object directly into a Piscina worker, as initially planned. I will move this to the Piscina area of the code.

docker-compose.yml
goetzrrGit: To prevent delays caused by long-running actions, we've implemented workers.  Imagine a single-threaded system: if User A's action takes 10 minutes, User B would be blocked for that entire duration.  Workers allow User B to perform actions without waiting for User A's action to complete

action-server/src/listeners/dbListeners.ts
Mythicaeda: Got it. Can the comment be reworded? Reading it with the context now, I understand, but I expect that when I'm reading it again without this context in mind I'll be confused by it again.

docker-compose.yml
Mythicaeda: Yes, but the feature was demo'd on Monday with @dandelany explicitly noting that it only handled one request at a time due to being uncertain if parallel runs were threadsafe. I of course understand the benefits of handling requests in parallel, but that's not how the feature has been advertised. Additionally, due to you `awaiting` on each request, I don't think the application is actually running multiple requests at once at the moment. ([See this comment](https://github.com/NASA-AMMOS/aerie/pull/1634#discussion_r2003971015), the summary is that the code is `await`ing a method that is `await`ing the finished run.)

action-server/src/app.ts
goetzrrGit: I added a health check and updated the e2e test with the action server url

deployment/docker-compose.yml
goetzrrGit: I added a health check and updated the e2e test with the action server url

action-server/src/config.ts
goetzrrGit: Removed the user and password from the configuration

action-server/src/db.ts
goetzrrGit: Updated the port to 5432 and added a minimum connection of 5.

action-server/src/db.ts
goetzrrGit: Removed the logging of credentials

action-server/src/listeners/dbListeners.ts
goetzrrGit: Removed the log message as it is reported at the end of this function

action-server/src/listeners/dbListeners.ts
goetzrrGit: Removed

action-server/src/listeners/dbListeners.ts
goetzrrGit: I added a logger to the server now and replaced most of the console.log/info/error with the winston logger.

action-server/src/listeners/dbListeners.ts
goetzrrGit: Thank you for the suggestion. I renamed it

action-server/src/listeners/dbListeners.ts
goetzrrGit: Thank you for the suggestion. I renamed it

action-server/src/listeners/dbListeners.ts
goetzrrGit: removed

action-server/src/listeners/dbListeners.ts
goetzrrGit: I reworded the message.

action-server/src/listeners/dbListeners.ts
goetzrrGit: It was for debugging. I removed it.

action-server/src/db.ts
goetzrrGit: Comment removed and logger added

action-server/src/listeners/dbListeners.ts
goetzrrGit: Reviewed this with @dandelany and we removed the TODO

action-server/src/listeners/dbListeners.ts
goetzrrGit: No problem. I moved and reworded it.

action-server/src/listeners/dbListeners.ts
goetzrrGit: I removed this message

docker-compose.yml
goetzrrGit: Got it, I updated the url

docker-compose.yml
goetzrrGit: I removed this line

deployment/docker-compose.yml
goetzrrGit: I removed this from the deployment docker compose file

docker-compose.yml
goetzrrGit: These env are now removed

docker-compose.yml
goetzrrGit: Action-server container's in the Docker Compose files now. Still figuring out the Kubernetes part, haven't done that before, so I'm learning it on the fly.

action-server/src/listeners/dbListeners.ts
goetzrrGit: I replaced console with logger

action-server/src/threads/worker.ts
goetzrrGit: remove commented out code and updated to use logger

docker-compose.yml
goetzrrGit: I added kubernete files for the action server

action-server/src/db.ts
goetzrrGit: We originally planned for the main thread to establish a pool and provide a connection to the Piscina worker.  However, this approach failed due to our inability to serialize the connection for transfer to the worker.  As a workaround, each worker now independently creates a pool, acquires a client connection, and passes it to the action.  The worker then closes its pool after executing the action code.

action-server/src/listeners/dbListeners.ts
goetzrrGit: It think we might need  `ActionsDbManager.init()` Removing ActionsDbManager.init() will prevent the creation of the listening client and associated PostgreSQL hooks.

action-server/src/utils/codeRunner.ts
goetzrrGit: @dandelany and I went through the TODO an left the ones we will address in v2 of the Action Server. I know we kinda have a lot atm. Not enough time for this cycle to address them.

action-server/src/threads/worker.ts
goetzrrGit: The actionAPI is used by each action JavaScript file. This API exposes three methods: `listSequences`, `readSequences`, and `writeSequences`.  Developers writing actions can leverage these methods to which will query the PostgreSQL database directly and require an active database connection.

action-server/src/db.ts
dandelany: Just to be clear - this was **_not_** logging the credential (password).

goetzrrGit:
> I noticed there are 0 E2E Tests added. Please add even just a basic test of the workflow.

We might need to create a ticket or added it to V2 as I don't think we have time to write one atm. @dandelany thoughts on this.

action-server/src/db.ts
Mythicaeda: But the pool's a `static` field. Did you try making the initing the DB connection pool in the startup method before _any_ threads are made (both worker and listener), and just using `ActionsDbManager.getDb();` where you're currently calling `init` and `getDb`? If you did and you got errors, please share the errors.

I don't mean serializing the pool as part of `actionTask`, since if you want `runAction` to have access to a `DB` connection, `runAction` _should_ call `getDb`

Mythicaeda: > each worker now independently creates a pool, acquires a client connection, and passes it to the action. The worker then closes its pool after executing the action code.

If you're going to code it this way, then a pool is a waste of resources and the workers should just manage a single connection, rather than an entire pool. 

However, I stand by that the correct approach is to have a static pool that the workers request a single connection from, which is nearly but not quite what is currently implemented.

action-server/src/listeners/dbListeners.ts
dandelany: > With the current code, I would expect that a long-running action will block the ability to update Action Definition schemas until it completes 

This is not the case (I have tested it) & not how `await` works with callbacks. The execution *of that function context* will be paused until the worker resolves with a result from the action, but the JS event loop keeps running and handling other tasks, including new incoming notifications. On the contrary, if you *don't* `await` here, you lose structured error propagation - ie. if `runAction` throws, the error will correctly bubble up the call stack with `await`, whereas it will be an unhandled rejection without it (these luckily still get logged in Node 15+, but different from a normal error). So I believe it is better to `await` here.

Re: the thread pool comments, this is *sort of* what it's doing already - at least, the pool used by `handleActionDefinition` is only used for that, and workers get their own DB connections when they are run. I don't think there is any need to have more than one DB connection for handling action definitions (+1 for the listener). 

The process by which workers get their own DB connection needs some refinement that we can discuss offline, bc it's too much to type out here - namely it should be initialized differently from the parent pool. But your concerns re: blocking are not an issue.

docker-compose.yml
dandelany: > Action Server was supposed to be single-threaded initially? Why does it have a Worker configuration?

V1 is meant to only run a single action at a time. However that **does not** mean it's single threaded. Even in a "one action at a time" world we need to run the action in a separate worker thread, or else it may block execution of the main thread and we may fail to handle & enqueue new requests while the previous action is running.

That said, this **should be set to `1`, not `8`,** this is the mechanism by which we enforce the "one action at a time" rule today and it is not being enforced correctly. It was originally hardcoded to `1` in code but must have changed when it was moved out to an env var. I've [updated this](https://github.com/NASA-AMMOS/aerie/pull/1634/commits/04d27c84bc0d1ec38b5350edc4120609e065036d) to `1` but left it alone otherwise.

> I don't think the application is actually running multiple requests at once

Explained in [other comment](https://github.com/NASA-AMMOS/aerie/pull/1634#discussion_r2006690697). It is.

Mythicaeda: > Explained in https://github.com/NASA-AMMOS/aerie/pull/1634#discussion_r2006690697. It is.

Understood. Thank you for the clear breakdown in that comment as to how the multithreading works.

> V1 is meant to only run a single action at a time. However that does not mean it's single threaded.

Ah, yes, calling it "single threaded" was a misspeak on my part throughout this thread. I meant "single _worker_ thread" (as opposed to an 8-thread pool). Changing it to `1` addresses that concern.

deployment/.env
Mythicaeda: Reminder: this has not been addressed yet. Given the 👍 from @cohansen, though, that might be just because the change hasn't been pushed yet.

action-server/src/listeners/dbListeners.ts
Mythicaeda: Bumping this comment, since it has yet to be addressed either way.

docker-compose.yml
Mythicaeda: Thanks!
Follow-up to this: I just realized that only the `develop` `docker-compose.yml` has its `AERIE_DATABASE_URL` updated. Please update this in the other `docker-compose`s.

docker-compose.yml
Mythicaeda: I've just noticed that only the `develop` docker-compose has been updated and that the others are setting the pool size to 8. Please update the other composes.

deployment/docker-compose.yml
Mythicaeda: The `image` path has not been updated. It needs to be the following:

```yaml
image: "${REPOSITORY_DOCKER_URL}/aerie-action:${DOCKER_TAG}"
```

(I'm assuming the image is supposed to be named `aerie-action` with a hyphen, as that is the naming convention of our other images. If it is actually getting published as `aerie_action` with an underscore, please fix this to avoid user confusion)

docker-compose.yml
Mythicaeda: Bumping as this comment has not yet been addressed.

action-server/src/listeners/dbListeners.ts
Mythicaeda: > So I believe it is better to await here.

Understood. Then my suggestion is to remove the `todo`, unless there's a larger discussion about it that I'm missing.

action-server/src/threads/worker.ts
Mythicaeda: Understood, that's a reasonable use case. Follow-up question: Is the pool at all accessible or can the user only see these three API methods? I'd prefer them not be able to call something like `API.connection` and start using it to run arbitrary SQL, even if the scope of their damage would likely be limited to only the `sequencing` and `actions` schemas.

action-server/src/type/types.ts
Mythicaeda: Bumping this comment.

action-server/src/utils/codeRunner.ts
Mythicaeda: That's fine, I just wanted to make sure that none of them were already accomplished, since there were several finished TODOs left behind in other files.

deployment/hasura/metadata/databases/tables/actions/action_run.yaml
cohansen: This broke some of the ui, so we reverted this change. Going to leave the relationship named `action_definition` for now.

Mythicaeda: Bumping as @cohansen left a 👍 but nothing was changed. Did you mean to commit a change or were you rejecting the suggestion?

Mythicaeda (commented):
This rereview skips over any DB-related comments that were not reacted to, but otherwise bumps and follows-up on any unresolved conversations.

deployment/postgres-init-db/sql/tables/actions/action_run.sql
cohansen: @AaronPlave This change will break the UI if it's referencing these columns.

deployment/hasura/metadata/databases/tables/actions/action_definition.yaml
cohansen: @AaronPlave Will this break a lot if we rename?

docker-compose.yml
goetzrrGit: Sorry I am a little confused about this question. I wanted to enable users to use a different port if needed. My thought was that they could change the PORT environment variable and then update the ports: [ "27186:27186" ] line to reflect that change. Are you saying I should change `ports: [ "27186:8080" ]` which gives even more flexibility?

docker-compose.yml
goetzrrGit: I looked over the docker compose and changed any pool size from 8 back to 1.

docker-compose.yml
goetzrrGit: I believe I got them all.

docker-compose.yml
goetzrrGit: Thanks for the chat @Mythicaeda I removed the ENV PORT from the docker-compose files and hardcoded the port in the server `config.ts` to **27186**
https://github.com/NASA-AMMOS/aerie/pull/1634/commits/d3f54a49c63736d1d267517ead3ecc8b1a744361

deployment/hasura/metadata/databases/tables/actions/action_definition.yaml
AaronPlave: Nope!

action-server/src/type/types.ts
dandelany: removed - should just be one error

action-server/src/threads/worker.ts
```diff
@@ -1,23 +1,39 @@
...
+    host: AERIE_DB_HOST,
+    port: parseInt(AERIE_DB_PORT, 5432),
+    database: AERIE_DB,
+    user: ACTION_DB_USER,
+    password: ACTION_DB_PASSWORD,
+    // should have exactly one client/connection
+    min: 1,
+    max: 1
+  });
+}
```
goetzrrGit, line 25: Non-blocking: I was thinking a little about whether we should keep the database pool alive in the worker threads, as a singleton, to maintain a constant connection. I'm leaning towards your approach of creating a new pool per action run. This releases connections when they're not needed, which is could happen if the user isn't using any actions for a while and the action server is just idle.

action-server/src/listeners/dbListeners.ts
```diff
@@ -78,7 +78,7 @@ async function runAction(payload: ActionRunInsertedPayload) {
   }
 
   const duration = Math.round(performance.now() - start);
-  const status = taskError || run?.errors ? "failed" : "complete";
+  const status = taskError || run?.errors ? "failed" : "success";
```
Mythicaeda, line 81: I don't see a similar update for setting the status to `incomplete`. I would expect to see such a line around `worker::runAction`. Is Github hiding it from me or does the Action Server not post when it starts _unning an action?

goetzrrGit: You are correct, This came up yesterday during Cody's work. Dan noted in the Action subteam thread that we plan to resolve it in V2 if that is okay with you?

Mythicaeda: If you're not logging it at all? Then yes, it can wait until v2.

Mythicaeda (approved):
Approving but please DO NOT MERGE until the Gateway is approved.

action-server/src/threads/worker.ts
dandelany: Yeah that's an interesting idea, the implications aren't completely obvious to me in terms of how it would behave so I think it's best left for later - we're coming back to the pooling question in V2 so I will do some testing with that approach.

PR #1638, opened 2025-03-03
Sequence Templates

cartermak:
 -->

### Manual Testing Procedure
- Make a fresh Aerie deployment
- Use the corresponding UI branch in local development mode
- Upload a mission model (e.g. banananation)
- Upload a command dictionary
- Make a parcel
- Make a new template with the contents `R00:00:00 GROW_BANANA 0 0`
- Make a plan
- Add an activity
- Simulate
- Navigate to sequencing panel
- Create a sequence filter
- ... etc

## Outstanding Work

What do we need to do before this is ready to go?

- [x] Testing
- [ ] Code cleanup (var naming, clean up commits, rebase or merge from develop)

## Future work

Open tickets for:

- [ ] Refactor SeqN parsing to a separate NPM package that can be shared by UI and sequencing server (I started this on the side: https://github.com/cartermak/aerie-seq-utils)
- [ ] Clean up the Parcels story (set "language" as a property of parcels?)
- [ ] Some user research/feedback?


deployment/postgres-init-db/sql/tables/sequencing/sequence_template.sql
```diff
@@ -0,0 +1,53 @@
...
+        on delete no action
+        NOT VALID,
+    constraint "model_id -> merlin.mission_model.id" foreign key (model_id)
+        references merlin.mission_model (id) match simple
+        on update no action
+        on delete no action,
+    constraint "parcel_id -> sequencing.parcel.id" foreign key (parcel_id)
+        references sequencing.parcel (id) match simple
+        on update no action
+        on delete no action,
```
mattdailis: I'm curious about the `on update no action` and `NOT VALID` here. Any objections to removing `NOT VALID` and changing the triggers to `on update cascade` and `on delete cascade`?

deployment/hasura/metadata/actions.graphql
```diff
@@ -38,6 +38,27 @@ type Mutation {
...
+    filterId: Int!,
+    simulationDatasetId: Int!,
+    seqId: String!,
+    timeRangeStart: String!,
+    timeRangeEnd: String!
+  ): AssignActivitiesByFilterResponse
+}
+
+type Mutation {
+  addTemplateExpansion(
```
DavidLegg: The "addTemplateExpansion" name suggests to me this is adding the result of expanding a template, but the arguments suggest this adds the template itself. If this is adding the template itself, I think "addTemplate" may be clearer.

sequencing-server/src/builders/seqJsonBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+              stem: '$$ERROR$$',
+              arguments: [{ message: e.message }],
+            }).METADATA({ simulatedActivityId: ai.id }),
+          ),
+        );
+      }
+
+      // Typeguard only
+      if (ai.expansionResult === null) {
+        break;
```
DavidLegg, line 36: I know we _shouldn't_ ever be able to reach this line, but just in case we do, how about `continue` instead of `break`?

sequencing-server/src/builders/seqJsonBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+       * TODO: we argue that this is actually ideal behavior, but it warrants discussion.
+       */
+      previousTime = ai.startTime;
+      for (const command of ai.expansionResult) {
+        const currentCommand =
+          command instanceof CommandStem
+            ? (command as CommandStem)
+            : command instanceof LoadStep
+            ? (command as LoadStep)
+            : (command as ActivateStep);
```
DavidLegg: I'm still fairly new to TS, but would this work?
```suggestion
        const currentCommand = command as (CommandStem | LoadStep | ActivateStep);
```

sequencing-server/src/builders/seqJsonBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  // If we never found a condition that prohibits sorting, then sort the converted commands and assign those as the output
+  if (activityInstanceCount > 0 && shouldSort) {
+    timeSorted = true;
+    allCommands = convertedCommands.sort((a, b) => {
+      const aStep =
+        a instanceof CommandStem ? (a as CommandStem) : a instanceof LoadStep ? (a as LoadStep) : (a as ActivateStep);
+      const bStep =
+        b instanceof CommandStem ? (b as CommandStem) : b instanceof LoadStep ? (b as LoadStep) : (b as ActivateStep);
```
DavidLegg: Do we need `aStep` and `bStep`, or can we just use `a` and `b` directly below?

sequencing-server/src/lib/batchLoaders/sequenceTemplateBatchLoader.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const sequenceTemplateBatchLoader: BatchLoader<
+    { modelId: number },
+    SequenceTemplate[],
+    { graphqlClient: GraphQLClient}
+> = opts => async keys => {
+  const { sequence_templates } = await opts.graphqlClient.request<{ sequence_templates: SequenceTemplate[] }>(
+      gql`
+      query GetSequenceTemplates($modelIds: [Int!]!) {
+        sequence_templates: sequence_template(where: {_and: [{ model_id: { _in: $modelIds }}]}) {
```
DavidLegg:
```suggestion
        sequence_templates: sequence_template(where: { model_id: { _in: $modelIds }}) {
```

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -3233,6 +3233,36 @@ export function doyToInstant(doy: DOY_STRING): Temporal.Instant {
...
+
+  while (minutes >= 60) {
+    hours += 1;
+    minutes -= 60;
+  }
+
+  while (minutes <= -60) {
+    hours -= 1;
+    minutes += 60;
+  }
```
DavidLegg:
```suggestion
function normalize(lg_time: number, sm_time: number, ratio: number): [number, number] {
    let sm = sm_time % ratio;
    return [lg_time + (sm_time - sm) / ratio, sm];
}

[seconds, milliseconds] = normalize(seconds, milliseconds, 1000)
[minutes, seconds] = normalize(minutes, seconds, 60)
[hours, minutes] = normalize(hours, minutes, 60)
```
etc. will do the same thing without a loop.

Just to double check the logic for negative times, though: This leaves negative time components negative, just clamped to within the appropriate unit. Is that what we want?

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,273 @@
...
+        },
+        {},
+    );
+
+    return simulatedActivities.filter(simAct => {
+        return applyFiltersToDirectiveOrSpan(simAct, filter, staticTypeMap, timeRangeStart, timeRangeEnd);
+    });
+}
+
+function applyFiltersToDirectiveOrSpan(
```
DavidLegg, line 34: Maybe I just haven't had enough coffee today, but I'm struggling to follow the logic of this routine. I think what's tripping me up is the way `included` gets re-incorporated along the way. Is there a short description of how these various kinds of filters interact?

I tried to write this out as boolean logic. I managed to show that `included` on line 71 is equivalent to `!!staticTypeMap[simulatedActivity.activityTypeName] || directiveOrSpanMatchesDynamicFilters(simulatedActivity, filter.dynamic_type_filters)`. After that, I drowned in boolean algebra. I strongly doubt we mean for this expression to be as complicated as it is, and there ought to be a simpler way to write this.

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -49,6 +49,7 @@ commandExpansionRouter.post('/put-expansion', async (req, res, next) => {
     return next();
   }
 
+  // WHY NOT DO THIS FIRST?
```
mattdailis, line 63: @pranav-super do you feel like this comment has been addressed?

docker-compose.yml
```diff
@@ -90,6 +90,7 @@ services:
       SEQUENCING_DB_USER: "${SEQUENCING_USERNAME}"
       SEQUENCING_DB_PASSWORD: "${SEQUENCING_PASSWORD}"
       SEQUENCING_LOCAL_STORE: /usr/src/app/sequencing_file_store
+      SEQUENCING_LANGUAGE: "SEQN"
```
mattdailis, line 93: We've changed one docker-compose file, which means:
- [x] We should change the other ones in the repo
- [x] We should document the new environment variables in Environment.md

sequencing-server/src/builders/seqJsonBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+
+      // Typeguard only
+      if (ai.expansionResult === null) {
+        break;
+      }
+
+      /**
+       * Treat the activity start time as the "previous time" for sorting expanded commands that begin with a command-relative time tag.
+       *
+       * TODO: we argue that this is actually ideal behavior, but it warrants discussion.
```
mattdailis, line 42: Should we have this discussion before merging?

sequencing-server/src/lib/batchLoaders/sequenceTemplateBatchLoader.ts
```diff
@@ -0,0 +1,49 @@
+import type { BatchLoader } from './index';
+import { gql, GraphQLClient } from 'graphql-request';
+import { ErrorWithStatusCode } from '../../utils/ErrorWithStatusCode.js';
+
+// TODO: grab all sequence templates for this model.
```
mattdailis: I believe this TODO comment has already been satisfied

sequencing-server/src/lib/batchLoaders/sequenceTemplateBatchLoader.ts
mattdailis: Possible simplification here
```suggestion
        sequence_templates: sequence_template(where: {model_id: { _in: $modelIds }}) {
```

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,272 @@
...
+    dynamicFilters: ActivityLayerDynamicFilter<typeof ActivityLayerFilterField>[],
+    // activityTypeDefMap: Record<string, ActivityType>,
+    // defaultArgumentsMap: DefaultEffectiveArgumentsMap,
+): boolean {
+    return dynamicFilters.reduce((acc, curr) => {
+        let matches = false;
+        if (curr.field === 'Type') {
+            matches = matchesDynamicFilter(simulatedActivity.activityTypeName, curr.operator, curr.value);
+        } else if (curr.field === 'Name') {
+            // TODO: is id correct?
```
mattdailis: Is it?

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,272 @@
...
+            matches = matchesDynamicFilter(simulatedActivity.id, curr.operator, curr.value);
+        // } else if (curr.field === 'Subsystem') {
```
mattdailis: Large blocks of commented-out code here. Are they present in the original, or is this a deviation?

sequencing-server/src/lib/mustache/util/index.ts
```diff
@@ -0,0 +1,97 @@
...
+function formatAsDate(date: string): string {
+    if (environment.language === SequencingLanguage.STOL) {
+        return ISO8061toSTOL(STOLToISO8061(date))
+    }
+    else {
+        return ISO8061toSeqN(SeqNToISO8061(date))
+    }
+}
+
+// TODO: FIGURE OUT A GOOD WAY TO ALLOW USER UPLOADS/DEFINITION IN AERIE MAIN? Ask about this.
```
mattdailis: We won't address this feature in this PR - I suggest we remove this comment and make a ticket instead

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -219,7 +346,250 @@ commandExpansionRouter.post('/put-expansion-set', async (req, res, next) => {
...
+    // filteredActivities is a list of the SimulatedActivities for the current seqId
+    const filteredActivities = filteredSimulatedActivitiesBySeqId[index]
+    if (filteredActivities && !(filteredActivities instanceof Error)) {
+      // Extract just the id and start offset from each simulated activity
+      seqIdToFilteredActivities[seqId] = filteredActivities.map(act => {
+        return { id: act.id, startOffset: act.startOffset }
+      });
+
+      // Add this simulated activity to allFilteredActivities if it's not already there
+      // TODO: figure out whether we need to handle the case where a simulated activity is included with multiple
```
mattdailis: This is a good point. The database does seem to permit this situation, but you're right that the frontend most likely doesn't. It's worth bringing up

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -430,18 +804,30 @@ commandExpansionRouter.post('/expand-all-activity-instances', async (req, res, n
...
 
+    logger.info(`POST /command-expansion/expand-all-activity-instances:\n` + JSON.stringify(seqIdToSimActivityId));
+
     // If the user has created a sequence, we can try to save the expanded sequences when an expansion runs.
+    logger.info('ORIGINAL SIMULATED ACTIVITIES: ' + JSON.stringify(simulatedActivities));
     for (const seqRow of seqRows.rows) {
       const seqId = seqRow.seq_id;
       const seqMetadata = seqRow.metadata;
 
+      // move this outside of the loop? why not use simulated activities from before? they use similar loaders (the gql query is the same, and we override an existing variable...why not delete?)
```
mattdailis, line 830: Has this question been addressed?

sequencing-server/src/routes/seqjson.ts
```diff
@@ -132,7 +132,7 @@ seqjsonRouter.post('/get-seqjson-for-seqid-and-simulation-dataset', async (req,
 
   // This is here to easily enable a future feature of allowing the mission to configure their own sequence
   // building. For now, we just use the 'defaultSeqBuilder' until such a feature request is made.
-  const seqBuilder = defaultSeqBuilder;
+  const seqBuilder = seqJsonBuilder; // TODO: make this compatible with strings too. also...why do we even have this? why not just pull from or update expanded sequences???
```
mattdailis: Has this todo comment been addressed?

sequencing-server/src/routes/seqjson.ts
```diff
@@ -226,7 +226,7 @@ seqjsonRouter.post('/bulk-get-seqjson-for-seqid-and-simulation-dataset', async (
...
```
mattdailis, line 229: Duplicate todo comment - has it been addressed?

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,272 @@
...
+        case 'is_less_than':
+            return itemValue < filterValue;
+        case 'is_within':
+            if (
+                Array.isArray(filterValue) &&
+                filterValue.length === 2 &&
+                typeof filterValue[0] === 'number' &&
+                typeof filterValue[1] === 'number'
+            ) {
+                // TODO should upper bound be inclusive or exclusive?
```
DavidLegg, line 168: I think inclusive is fine. I think that's the more intuitive meaning, if the number is an integer. If the number is a float, I don't think it matters. If they want exclusive, users can do (upper bound - epsilon) easily enough.

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,272 @@
...
+            }
+            return false;
+        case 'is_not_within':
+            if (
+                Array.isArray(filterValue) &&
+                filterValue.length === 2 &&
+                typeof filterValue[0] === 'number' &&
+                typeof filterValue[1] === 'number'
+            ) {
+                // TODO should upper bound be inclusive or exclusive?
```
DavidLegg, line 179: Ibid, inclusive is fine. As long as it exactly inverts the `is_within` case, as this code does.

sequencing-server/src/lib/filters/utilities.ts
```diff
@@ -0,0 +1,272 @@
...
+        // Check to see if type is included in static list
+        if (filter.static_types?.length) {
+            included = !!staticTypeMap[type.name];
+        }
+
+        // Check if necessary to see if type is included in dynamic list
+        if ((!filter.static_types?.length || !included) && filter.dynamic_type_filters?.length) {
+            included = typeMatchesDynamicFilters(type, filter.dynamic_type_filters);
+        }
+        return included;
```
DavidLegg, line 221: After a bit of boolean algebra, this is equivalent to the following:
```suggestion
    const includedInStaticTypes = !filter.static_types?.length || !!staticTypeMap[type.name];
    const includedInDynamicTypes = !filter.dynamic_type_filters?.length || typeMatchesDynamicFilters(type, filter.dynamic_type_filters);
    return includedInStaticTypes && includedInDynamicTypes;
```

sequencing-server/src/lib/mustache/util/index.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// helper to clean dates. must be manually invoked by the user, just in case. here, Text and SeqN are handled the same.
+function formatAsDate(date: string): string {
+    if (environment.language === SequencingLanguage.STOL) {
+        return ISO8061toSTOL(STOLToISO8061(date))
+    }
+    else {
+        return ISO8061toSeqN(SeqNToISO8061(date))
+    }
+}
```
DavidLegg: > Text and SeqN are handled the same

Small nit: Can I suggest using ISO8601 for text, since we have that format already? If the user hasn't given a language, defaulting to the international standard seems better than defaulting to JPL's idiosyncrasies.

sequencing-server/src/lib/mustache/util/index.ts
```diff
@@ -0,0 +1,97 @@
...
+                    let castedFunctions: { [key: string]: any } = module
+
+                    Object.keys(castedFunctions).forEach(funcName => {
+                        Handlebars.registerHelper(moduleName + "." + funcName, castedFunctions[funcName])
+                    })
+                })
+            }
+        });
+    }
+});
```
DavidLegg: Maybe I'm misremembering, but didn't we de-scope user-provided helpers in favor of just the four hard-coded ones for now? Is this code needed?

If we're keeping this code, we should make sure to mention how this works in our documentation.

sequencing-server/src/lib/mustache/util/index.ts
```diff
@@ -0,0 +1,97 @@
...
+export class Mustache {
+    private template: HandlebarsTemplateDelegate<any>
+
+    constructor(template: string, language?: SequencingLanguage) {
+        environment.language = language ?? environment.language
+        this.template = Handlebars.compile(template)
+    }
+
+    public execute(data: any) {
+        // TODO: AUTOMATICALLY FORMAT TIMES IN DATA? Ask about this
```
DavidLegg: I believe we decided not to automatically format times and dates - if so, we can remove this comment.

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+
+/////////////// SEQUENCING-SERVER-SPECIFIC HELPERS ///////////////
+export function addTime(startTime: string, duration: string, environment: { language: SequencingLanguage }): string {
+  let date: Temporal.Instant
+  if (environment.language === SequencingLanguage.STOL) {
+    date = STOLToISO8061(startTime)
+  }
+  else { // Text and SeqN are handled the same.
+    date = SeqNToISO8061(startTime)
+  }
```
DavidLegg: Nit: Maybe we should add a `parseDatetime` / `formatDatetime` method pair, which reads `environment.language` and dispatches to the language-appropriate parser/formatter, to keep things consistent across the time-related helpers.

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+
+function checkNumDurationComponents(split: string[]): split is [string, string, string] {
+  return split.length === 3;
+}
+
+export function AERIEDurationToISO8061(duration: string): string {
+  // HHHHHH...:MM:SS.mmmuuu -> PHHMMSS.mmmuuuS
+  let split = duration.split(":")
+
+  if (!checkNumDurationComponents(split)) {
```
DavidLegg: This method can be inlined for clarity.
```suggestion
  if (split.length !== 3) {
```

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+    throw new Error(`Invalid duration string: ${duration}`);
+  }
+
+  let hours = parseInt(split[0])
+  let minutes = parseInt(split[1])
+  let split2 = (split[2]).split(".")
+  let seconds = parseInt(split2[0] as string)
+  if (split2.length > 1) {
+    let microseconds = parseInt(split2[1] as string)
```
DavidLegg: What happens if I use the string `00:00:00.100`? Will that get misinterpreted as 100 micros here?

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+    return `PT${hours > 0 ? `${hours}H` : ""}${minutes > 0 ? `${minutes}M` : ""}${seconds > 0 && microseconds > 0 ? `${seconds}.${microseconds}S` : (seconds > 0) ? `${seconds}$` : (microseconds > 0) ? `0.${microseconds}S` : ""}`
```
DavidLegg: Aerie durations could be negative, e.g. `-00:05:00`. Is this a case we need to consider?
As written, I think this code would misinterpret this as "-0 hours + 5 minutes", i.e., positive 5 minutes.
For example, I could image an activity parameter that schedules a command relative to some mid-activity event, where both positive and negative offsets are sensible (so you couldn't rely on `subtract-time` to force the offset to be positive).

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+  else {
+    return `PT${hours > 0 ? `${hours}H` : ""}${minutes > 0 ? `${minutes}M` : ""}${seconds > 0 ? `${seconds}S` : ""}`
+  }
+}
+
+// CONVERSION BACK TO STRINGS
+export function ISO8061toSTOL(date: Temporal.Instant): string {
```
DavidLegg: Nit: ISO8601 is a string format, these functions convert to and from `Temporal.Instant` objects, not ISO8601. This name is mildly misleading.

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
+// TODO: Ensure absolute and relative times resolve correctly. Ask about this later.
```
DavidLegg: Can you elaborate on this question? Are your concerns resolved here?

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,313 @@
...
+    return doyDuration;
+  }
+
+  return null;
+}
+
+/**
+ * Parses a duration string (DOYTHH:mm:ss.ms) into its separate components
+ */
+function parseDOYDurationTime(doyTime: string): ParsedDurationString | null {
```
DavidLegg, line 293: Nit: The "days" component of a duration are just a number of days, not specifically day-of-year (DOY). I think this can be renamed `parseDuration`.

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -219,7 +346,250 @@ commandExpansionRouter.post('/put-expansion-set', async (req, res, next) => {
...
+  /**
+   * ARGUMENTS
+   * {
+   *    modelId: Int!,
+   *    simulationDatasetId: Int!,
+   *    seqIds: [Int!]!
+   * }
+   */
+
+  // const defaultTemplate = "CMD {{formatAsDate startTime}} {{name}} {{duration}}"; //req.body.input.template;
```
DavidLegg: Looks like some debugging code, I think we can remove this?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -219,7 +346,250 @@ commandExpansionRouter.post('/put-expansion-set', async (req, res, next) => {
...
+    context.sequenceTemplateDataLoader.load({ modelId }),
+    context.simulatedActivityInstanceBySeqIdBatchLoader.loadMany(seqIds.map(seqId => {
+      return { simulationDatasetId, seqId }
+    }))
+  ]);
+
+  //  2. Determine the language being used (SeqN vs. STOL)
+  //        Presently, we assume based on a database constraint, that all templates pulled for a given model/parcel combo have
+  //        the same language. While this constraint will remain true its exact enforcement and therefore implementation in SQL
+  //        and here may be subject to change.
```
DavidLegg, line 383: Is this comment still accurate, or did we drop the database constraint?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -219,7 +346,250 @@ commandExpansionRouter.post('/put-expansion-set', async (req, res, next) => {
...
+        //    i.e. "CMD {{ dsvsdfs }}" expands to "CMD ".
+        currentTemplate.setLanguage(languages[0])
+        const commandString = currentTemplate.execute(stringifyActivity(simulatedActivity))
+
+        // add to results
+        expandedActivities[simulatedActivityId] = {
+          value: {
+            ...simulatedActivity,
+            expansionResult: commandString,
+            errors: [] // TODO: pass the errors, once we have the errors, if we even can
```
DavidLegg, line 507: Does this need to be addressed? Should we have a try/catch around `execute` a few lines up to facilitate this?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -219,7 +346,250 @@ commandExpansionRouter.post('/put-expansion-set', async (req, res, next) => {
...
+    const sortedSimulatedActivitiesWithCommands: ExpandedActivity<string>[] = sortedActivityInstances.reduce((result: ExpandedActivity<string>[], current) => {
+      const expandedActivity = expandedActivities[current.id];
+      if (!expandedActivity) {
+        // Case: this activity wasn't expanded because we didn't have a template for it
+        return result;
+      } else {
+        result.push(expandedActivity.value);
+        return result
+      }
+    }, [])
```
DavidLegg, line 530:
```suggestion
    const sortedSimulatedActivitiesWithCommands: ExpandedActivity<string>[] = sortedActivityInstances.map(current => expandedActivities[current.id]).filter(e => e);
```

sequencing-server/test/templating/mustache.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { SequencingLanguage } from '../../src/lib/mustache/enums/language.js';
+import { Mustache } from '../../src/lib/mustache/util/index.js';
+
+describe('Basic template functionality', () => {
+
+    // nested in object to demonstrate functionality
+    let genericInput = { input: { name: "<name>" } }
+
+    it('can escape inputs in filling out templates', () => {
+        // handlebars does automatically escape the input
```
DavidLegg, line 10: This might be something to call out in our documentation - html entities likely aren't what users are expecting to see in a spacecraft sequencing system.

sequencing-server/test/templating/mustache.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    
+        it('should decrement correctly', () => {
+            let templateRaw = 'Subtracting dates: {{ subtract-time date duration }}'
+            let template = new Mustache(templateRaw, SequencingLanguage.STOL)
+            expect(template.execute(input))
+                .toEqual('Subtracting dates: 2025-001/00:57:01.877800Z')
+        })
+    });
+
+    it('should flatten arrays correctly', () => {
```
DavidLegg, line 66: Not something we explicitly need to handle, but what happens if I indicate an object from a template field? For example, let's say I wanted `{{ activity.params.foo }}` but I mess up and put `{{ activity.params }}`?
We might want to say what that does in some documentation, even if it just blows up in an opaque way.

sequencing-server/test/templating/time.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+        expect(ISO8061duration).toEqual('PT12345H67M89.101112S')
+    });
+});
+
+// test conversion back to strings
+describe('String conversion', () => {
+    describe('convert ISO8061 to SeqN', () => {
+        it('should convert correctly', () => {
+            let instant: Temporal.Instant = Temporal.Instant.fromEpochMicroseconds(1735732883000123n);
```
DavidLegg: Could we replace `fromEpochMicroseconds` with [Temporal.Instant.from(string)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant/from) so the test case is more legible?

sequencing-server/test/templating/time.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+            let seqnYMD = "2025-01-01T00:00:00.000123Z"
+            expect(convertDoyToYmd(seqnYMD)).toEqual('2025-01-01T00:00:00.000123Z')
+        });
+
+        it('should handle conversion from STOL YMD', () => {
+            let stolYMD = "2025-01-01/00:00:00.000123Z"
+            expect(convertDoyToYmd(stolYMD)).toEqual('2025-01-01T00:00:00.000123Z')
+        });
+
+        it('should operate agnostic of Zulu abbrevriation', () => {
```
DavidLegg:
```suggestion
        it('should operate agnostic of Zulu abbreviation', () => {
```

deployment/postgres-init-db/sql/tables/sequencing/sequence_template.sql
```diff
@@ -0,0 +1,28 @@
...
+    on update cascade
+    on delete set null,
+  constraint "model_id -> merlin.mission_model.id" foreign key (model_id)
+    references merlin.mission_model (id) match simple
+    on update cascade
+    on delete set null,
+  constraint "parcel_id -> sequencing.parcel.id" foreign key (parcel_id)
+    references sequencing.parcel (id) match simple
+    on update cascade
+    on delete cascade,
```
DavidLegg: Why do we have `on delete cascade` for parcel, but we use `on delete set null` for the model and activity type?

sequencing-server/src/lib/filters/utilities.ts
cartermak: De Morgan would be proud

docker-compose.yml
```diff
@@ -114,6 +115,7 @@ services:
       PUBLIC_HASURA_CLIENT_URL: http://localhost:8080/v1/graphql
       PUBLIC_HASURA_SERVER_URL: http://hasura:8080/v1/graphql
       PUBLIC_HASURA_WEB_SOCKET_URL: ws://localhost:8080/v1/graphql
+      PUBLIC_SEQUENCING_MODE: "templating"
```
mattdailis: We should figure out which templating mode should be the default in what environments (e.g. development, new users, existing users)

sequencing-server/src/routes/command-expansion.ts
pranav-super: I don't feel like it has been, however as it's not the focus of the PR this comment is fine safe to remove.

docker-compose.yml
pranav-super: Updated [here](https://github.com/NASA-AMMOS/aerie/commit/10086e6fc2a48a667e20f458a73c36fd2e7dc843).

sequencing-server/src/lib/mustache/util/index.ts
pranav-super: How is [this ticket](https://github.com/NASA-AMMOS/aerie/issues/1646)?

sequencing-server/src/lib/mustache/util/index.ts
```diff
@@ -0,0 +1,97 @@
...
+        });
+    }
+});
+
+
+/////////////// AERIE HELPER REGISTRATION ///////////////
+Handlebars.registerHelper("add-time", addTime)
+Handlebars.registerHelper("subtract-time", subtractTime)
+Handlebars.registerHelper("flatten", flatten)
+Handlebars.registerHelper("formatAsDate", formatAsDate)
```
cartermak: We should pick a consistent style for our helper names

mattdailis: Good catch. Let's change this to `format-as-date`

sequencing-server/src/builders/seqJsonBuilder.ts
```diff
@@ -0,0 +1,113 @@
...
+        if (currentCommand.GET_RELATIVE_TIME() && previousTime) {
+          const absoluteCommand: CommandStem | LoadStep | ActivateStep = currentCommand.absoluteTiming(
+            previousTime.add(currentCommand.GET_RELATIVE_TIME() as Temporal.Duration),
+          );
+          convertedCommands.push(absoluteCommand);
+          previousTime = absoluteCommand.GET_ABSOLUTE_TIME();
+        } else {
+          convertedCommands.push(currentCommand);
+          previousTime = currentCommand.GET_ABSOLUTE_TIME();
+        }
```
cartermak, line 69: Not something this PR introduced, but suggest that we report a warning or error if a single source activity provides out-of-order commands (i.e., check that the current command's absolute time is after `previousTime`).

sequencing-server/src/lib/mustache/util/index.ts
mattdailis: Looks great, thanks! I'm trying to decide if we should keep this code, or if we should delete it for now and add it back when we address that ticket. Thoughts?

sequencing-server/src/lib/filters/utilities.ts
pranav-super: It's been working as far as we have tested so I would say so. I'll go ahead and clean up the comments here.

sequencing-server/src/lib/filters/utilities.ts
pranav-super: They can be removed.

sequencing-server/src/routes/command-expansion.ts
pranav-super: It does not and seems out of scope for now, as there aren't clear use cases. Is it okay to leave this TODO up, just so this functionality is clearly pointed out?

sequencing-server/src/routes/command-expansion.ts
pranav-super: This hasn't! I noticed it when going through the method earlier so that I could understand how the process worked before writing the templating equivalent, so it's not a new problem in any way, but it's not clear why this code is organized the way it is. This implementation does seem redundant.

sequencing-server/src/routes/seqjson.ts
pranav-super: Not exactly. 

Since this route is explicitly named after seqjson, should such compatibility be allowed? Should we introduce another route? 

It also seems like the uses for this route could (should?) be limited altogether, as it has a single use in the UI that effectively regenerates a sequence when expanding it in the UI, as opposed to just grabbing the result from the appropriate table (`expanded_sequences`). The "bulk" equivalent of this route _has_ been removed from the UI.

sequencing-server/src/routes/seqjson.ts
pranav-super: See above.

sequencing-server/src/routes/command-expansion.ts
mattdailis: Yes, let's leave the comment, but reword it to be explanatory rather than a TODO comment. Maybe something like this?

```
// NOTE: The database schema permits a simulated activity to be associated with multiple seq IDs, even though
// there is no way to create that multi-association using the UI. This code will honor the multi-association
```

deployment/postgres-init-db/sql/tables/sequencing/sequence_template.sql
cartermak: I think `on delete set null` would make sense for parcels too, but this is consistent with the parcel and workspace linkage for user sequences (which are both `on delete cascade`. Take your pick for consistency with either 1) other FKs on this table or 2) other tables in Aerie?

sequencing-server/src/builders/seqJsonBuilder.ts
cartermak: This came from the existing `seqBuilder`; suggest we boot to #1647

sequencing-server/src/builders/seqJsonBuilder.ts
cartermak: I suppose we should...do you have thoughts? I came up with this change, although it looks like I might've also been involved in the original design to not sort single-activity sequences: #879 

I can't think of why we'd want that other than to preserve relative timing, which isn't a thing we ended up using in expanded commands for Clipper.

sequencing-server/src/routes/seqjson.ts
cartermak: Suggest we resolve separately since that point mostly relates to an existing bug. I just wrote #1647 and #1648 to capture the actions to cleanup these builders and fix the bug with inconsistency between the "bulk" route and the `expanded_sequences` table.

sequencing-server/src/lib/mustache/util/index.ts
pranav-super: I think it makes sense to keep, just so it's also there in the code at the exact point of insertion. However, if that's not standard practice then we should remove it.

sequencing-server/src/builders/seqJsonBuilder.ts
pranav-super: Fixed!

sequencing-server/src/builders/seqJsonBuilder.ts
pranav-super: We should be able to just use `a` and `b` directly.

(Fix has been committed)

sequencing-server/src/lib/filters/utilities.ts
pranav-super: That seems reasonable! This implementation was borrowed directly from the UI so that we could ensure that we accurately mimicked UI filtering behavior - should we open a small ticket about resolving this in both spots so that it's simple, readable, and consistent between the two?

sequencing-server/src/lib/filters/utilities.ts
pranav-super: This TODO is from the UI code as well.

sequencing-server/src/routes/command-expansion.ts
pranav-super: It has been removed!

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: It seems that the current notation was required so our editors wouldn't flag a type error. Inlining it leads the type checking system to believe that `split[0]` is `string|undefined`, but as is, it believes `split` is exactly `[string, string, string]`.

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: Just added it to a test! Seems like it interprets it correctly, as `PT0.100S`.

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: A fix has been committed and corresponding tests added!

sequencing-server/test/templating/time.spec.ts
pranav-super: I think the point of this _was_ to use a date format that was a little opaque to make it clear that conversion from Instant, in whatever format, to a clear STOL format, is doable. I could add a comment annotating this with the date that number represents, or replace with a string - what do you think?

sequencing-server/test/templating/mustache.spec.ts
pranav-super: Committed to docs!

sequencing-server/src/builders/seqJsonBuilder.ts
pranav-super: We decided to push this to #1647.

deployment/postgres-init-db/sql/tables/sequencing/sequence_template.sql
pranav-super: Commit pushed - this is subject to change in the future, however, as the model is updated further! We should make sure to add a note that this behavior likely _will_ change, or that we might want UI support for reassigning existing templates to new parcels/models/activity types.

We are very much not at the final stage for the template authoring process; there are definitely unanswered questions.

sequencing-server/src/builders/seqJsonBuilder.ts
pranav-super: This is work for #1647.

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
pranav-super: @mattdailis This might have already been merged!

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: Since there are only 2 instances, we will avoid this simplification for now.

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: This is also borrowed from the UI!

sequencing-server/src/routes/command-expansion.ts
pranav-super: Error handling with Handlebars should get its own ticket or should be integrated with a later templating PR.

We have silent failures if the input is missing, but _also_ loud, uncaught failures it the template is wrong somehow or a helper is called on an incorrect type. We want to make both transparent to the user somehow.

Related but separate is the lack of clear proper error propagation from `sequencing-server` to `aerie-ui`.

sequencing-server/src/routes/command-expansion.ts
pranav-super: Pushed to #1647.

sequencing-server/src/routes/command-expansion.ts
cartermak: #1649

sequencing-server/src/routes/seqjson.ts
pranav-super: We should remove the deprecated one OR make both point to a table, as opposed to doing this on-the-fly recompilation. We can handle this in #1648. That being said, the TODO itself can be removed as the method here IS SeqJSON-specific so generic string handling wouldn't really make sense.

We don't want to change legacy expansion API right now, in this PR.

sequencing-server/src/routes/command-expansion.ts
pranav-super: Ensure this is in #1647. Otherwise, it can be resolved.

sequencing-server/src/routes/command-expansion.ts
pranav-super: This opportunity is referenced [here](https://github.com/NASA-AMMOS/aerie-team/discussions/57#discussioncomment-12554051).

sequencing-server/src/builders/seqJsonBuilder.ts
pranav-super: Moving to be addressed in #1647.

sequencing-server/src/lib/filters/utilities.ts
pranav-super: Comment removed.

sequencing-server/src/lib/mustache/util/index.ts
pranav-super: All of this will be migrated into [this issue](https://github.com/NASA-AMMOS/aerie/issues/1646) and removed from the code here.

sequencing-server/src/lib/mustache/util/index.ts
pranav-super: This will be moved to [this issue](https://github.com/NASA-AMMOS/aerie/issues/1646) and removed from the current codebase here.

sequencing-server/src/lib/mustache/util/time.ts
pranav-super: This comment has been removed.

sequencing-server/src/lib/mustache/util/index.ts
pranav-super: This has been handled.

pranav-super: This has been included in a newer commit!

sequencing-server/test/templating/time.spec.ts
pranav-super: This test has been updated for readability.

sequencing-server/test/templating/mustache.spec.ts
pranav-super: This behavior has been documented - we decided not to do anything special for objects _for now_.

docker-compose.yml
pranav-super: `"legacy"` has now been updated as the default here and in the UI!

pranav-super (approved):
Looks ready to go!

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// TIME PARSING
+export function SeqNToInstant(date: string): Temporal.Instant {
+  return Temporal.Instant.from(convertDoyToYmd(date))
+}
+
+export function STOLToInstant(date: string): Temporal.Instant {
+  return Temporal.Instant.from(convertDoyToYmd(date))
+}
```
mattdailis, line 79: The definitions of `SeqNToInstant` and `STOLToInstant` are identical; what do we gain by having two separate functions here?

pranav-super: If we introduce more formats (for example, one called `Language`), the implementation of `LanguageToInstant` might differ. As such, it might be a little confusing to _not_ have a method for each, if `Language` requires its own `LanguageToInstant` method. In lieu of having:

```
export function STOLorSeqNToInstant(date: string): Temporal.Instant {
  return Temporal.Instant.from(convertDoyToYmd(date))
}

export function LanguageToInstant(date: string): Temporal.Instant {
  // implementation...
}
```

I think the following, while redundant, provides a more clear API:

```
export function SeqNToInstant(date: string): Temporal.Instant {
  return Temporal.Instant.from(convertDoyToYmd(date))
}

export function STOLToInstant(date: string): Temporal.Instant {
  return Temporal.Instant.from(convertDoyToYmd(date))
}

export function LanguageToInstant(date: string): Temporal.Instant {
  // implementation...
}
```

mattdailis: I see... what's giving me pause is my understanding of the use case for `add-time` was that it would take in things like the activity start time or end time. That time is coming from the simulated_activities table - I'm not sure what format it's in by the time it gets to handlebars, but I would be surprised if it was in STOL format. Let me know if I'm misreading the use case here

I'm looking at the callers of this function in `addTime` and `subtractTime`:
https://github.com/NASA-AMMOS/aerie/blob/5894b2d7f248af858b8d05e8ff5dd27caf4ecb7c/sequencing-server/src/lib/mustache/util/time.ts#L7-L17

pranav-super: It would still be in its native format! It's _in_ the implementation for `addTime` that the language is checked and the conversion is performed:

```
export function addTime(startTime: string, duration: string, environment: { language: SequencingLanguage }): string {
  let date: Temporal.Instant
  if (environment.language === SequencingLanguage.STOL) {
    date = STOLToInstant(startTime)
  }
  else if (environment.language === SequencingLanguage.TEXT) {
    date = Temporal.Instant.from(TextToISO8601(startTime))
  }
  else {
    date = SeqNToInstant(startTime)
  }
  // ...
}
```

deployment/docker-compose.yml
JosephVolosin, line 1: Believe we need to add `PUBLIC_COMMAND_EXPANSION_MODE: typescript` to the `environment` definition for `aerie-ui` to get up-to-date with the changes on the UI side

deployment/hasura/metadata/databases/tables/sequencing/sequence_template.yaml
```diff
@@ -0,0 +1,50 @@
...
+      set:
+        owner: "x-hasura-user-id"
+update_permissions:
+  - role: aerie_admin
+    permission:
+      columns: [name, model_id, parcel_id, template_definition, activity_type, language, owner]
+      filter: {}
+  - role: user
+    permission:
+      columns: [name, template_definition, language]
```
JosephVolosin: Is there a reason these permissions don't match up with a user's `insert` permissions?

deployment/docker-compose.yml
pranav-super: Is it not already in? [I know `.env` _has_ been updated](https://github.com/NASA-AMMOS/aerie-ui/blob/69c324a41607b9993ac1246e884bd68dc914c2ac/.env)...

deployment/hasura/metadata/databases/tables/sequencing/sequence_template.yaml
pranav-super: I don't believe so. My guess would be because lack of supporting functionality, but that historically hasn't been a good reason to restrict permissions. I think this could be updated to match that of a the `insert` permissions...any thoughts @mattdailis?

mattdailis: No reason I can think of. Good catch. 4815889bcc702c1feb49cdbbb21ceff79de56e02

deployment/docker-compose.yml
mattdailis: Good catch - I've updated the docker-compose file.

If the env variable is missing, or if it is set to anything except "templating", the UI will go into typescript command expansion mode. I think it's a bit less disruptive that way - people who want to try out sequence templates can opt in, but no one is required to add `PUBLIC_COMMAND_EXPANSION_MODE` to their environment just to opt out.

mattdailis (approved):

PR #1641, opened 2025-03-11
Add Ability to Lock User Sequences

duranb:
* **Tickets addressed:** [AERIE UI-1633](https://github.com/NASA-AMMOS/aerie-ui/issues/1633)
* **Review:** By commit

## Description
This adds an `is_locked` column to the `user_sequences` table that prevents the sequence entry from being edited when true.

## Future work

TODO:

- [x] add db tests

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
@@ -278,7 +278,7 @@ public class UserSequenceTests {
   }
 
   @Test
-  void canDeleteLockedSequence() throws SQLException {
+  void canDeleteUnlockedSequence() throws SQLException {
     final var otherSequenceId = createUserSequence("Test Sequence 3", parcelId, workspaceId, sequenceDefinition, sequenceUser.name());
 
     assertDoesNotThrow(()->deleteUserSequence(otherSequenceId));
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
@@ -1,0 +1,286 @@
+package gov.nasa.jpl.aerie.database;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.ArrayList;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@SuppressWarnings("SqlSourceToSinkFlow")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class UserSequenceTests {
+  private final String sequenceDefinition = """
+        C BAKE_BREAD
+        C PEEL_BANANA "fromStem"
+        C GROW_BANANA 0 0
+      """;
+  private DatabaseTestHelper helper;
+  private MerlinDatabaseTestHelper merlinHelper;
+
+  private Connection connection;
+  private int commandDictionaryId;
+  private int parcelId;
+  private int workspaceId;
+  private int sequenceId;
+  private MerlinDatabaseTestHelper.User sequenceUser;
+
+  void setConnection(DatabaseTestHelper helper) {
+    connection = helper.connection();
+  }
+
+
+  @BeforeEach
+  void beforeEach() throws SQLException {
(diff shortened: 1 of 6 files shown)


db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
```diff
@@ -0,0 +1,209 @@
...
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@SuppressWarnings("SqlSourceToSinkFlow")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class UserSequenceTests {
+  private final String sequenceDefinition = """
+        C BAKE_BREAD
+        C PEEL_BANANA \"fromStem\"
```
Mythicaeda: Very Minor, ignorable: Because you're in a block string, you don't need to escape the `"` here!

db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
```diff
@@ -0,0 +1,209 @@
...
+    assertDoesNotThrow(()->assignParcelToSequence(testParcelId, sequenceId));
+  }
+
+  @Test
+  void lockUserSequence() throws SQLException {
+    final var testParcelId = createParcel("Test Parcel 2 ");
+    updateUserSequenceIsLocked(sequenceId, true);
+
+    assertThrows(SQLException.class, ()->assignParcelToSequence(testParcelId, sequenceId));
+  }
```
Mythicaeda: I'd like to request 4 similar tests.

1) `lockDoesNotAffectOtherSequence` or sth like that. Basically, same set up, _but_ you create a second user sequence and add an `asserDoesNotThrow( assignParcelToSequence(testParcel, unrelatedSequence)`
2) `unlockUserSequence`. Starts the same as `lockUserSequence`, but the test then unlocks the sequence and tries to update the parcel again (with the expected result being a success this time)
3) `cannotDeleteLockedSequence`. Like `lockUserSequence`, but it attempts to delete a locked sequence instead of updating it. It would then assign the returned exception and check that the error message contains `Cannot delete locked user sequence.`
4) `canDeleteUnlockedSequence`. Like `updateUserSequenceParcel`, but with a delete instead of an update.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
```diff
@@ -0,0 +1,209 @@
...
+    final var testParcelId = createParcel("Test Parcel 2 ");
+    assertDoesNotThrow(()->assignParcelToSequence(testParcelId, sequenceId));
+  }
+
+  @Test
+  void lockUserSequence() throws SQLException {
+    final var testParcelId = createParcel("Test Parcel 2 ");
+    updateUserSequenceIsLocked(sequenceId, true);
+
+    assertThrows(SQLException.class, ()->assignParcelToSequence(testParcelId, sequenceId));
```
Mythicaeda: Minor, I'd prefer if you catch this and check that the error message is the one we're expecting. This will help catch if the DB code starts throwing an exception for the wrong reason in the future:
```suggestion
    final var ex = assertThrows(SQLException.class, ()->assignParcelToSequence(testParcelId, sequenceId));
    assertTrue(ex.getMessage().contains("Cannot update locked user sequence."));
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
```diff
@@ -0,0 +1,209 @@
...
+      """;
+  private DatabaseTestHelper helper;
+  private MerlinDatabaseTestHelper merlinHelper;
+
+  private Connection connection;
+  int commandDictionaryId;
+  int parcelId;
+  int workspaceId;
+  int sequenceId;
+  MerlinDatabaseTestHelper.User sequenceUser;
```
Mythicaeda: Minor: these should all be `private`, as should the Helper Functions and the record UserSequence

Mythicaeda (commented):
A couple of minor things. The one comment really blocking this from an approval is just the request for those tests.

duranb:
@Mythicaeda all comments shooould be addressed finally. Thanks for taking a look!

db-tests/src/test/java/gov/nasa/jpl/aerie/database/UserSequenceTests.java
```diff
@@ -0,0 +1,286 @@
...
+      deleteUserSequence(sequenceId);
+    } catch (SQLException ex) {
+      if (!ex.getMessage().contains("Cannot delete locked user sequence.")) {
+        throw ex;
+      }
+    }
+  }
+
+  @Test
+  void canDeleteLockedSequence() throws SQLException {
```
Mythicaeda: Minor: name is incorrect, should be:
```suggestion
  void canDeleteUnlockedSequence() throws SQLException {
```

Mythicaeda (approved):

PR #1642, opened 2025-03-12
Bugfix: Sequencing Server Duration Parsing and Rebalancing

### LINKED ISSUE #1625: Sequencing server fails to deserialize `Duration`-type activity arguments
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

3.0.1

### Describe the bug

The command expansion service incorrectly interprets all `Duration`-type activity arguments (that is, `Duration` as a value schema type) as having a value of zero. From a little debugging, it seems like [`parse`](https://www.npmjs.com/package/postgres-interval#parsepginterval---interval) is returning a duration of zero when parsing the serialized value, which comes as an integer number of microseconds.

Relevant sequencing server function: https://github.com/NASA-AMMOS/aerie/blob/e92f98a0ffc60a5f48d3b75b81129bde008950cb/sequencing-server/src/lib/batchLoaders/simulatedActivityBatchLoader.ts#L304-L349

### Reproduction

Write a command expansion rule that uses a `Duration`-type activity parameter and see that the value of the parameter is deserialized for use in the expansion rule as zero. 

For example, the below uses `GrowBanana` and prints out the string value of the duration (which comes back as `PT0S`).

```ts
export default function MyExpansion(props: {
  activityInstance: ActivityType,
  channelDictionary: ChannelDictionary | null
  parameterDictionaries : ParameterDictionary[]
}): ExpansionReturn {
  const { startTime, attributes } = props.activityInstance;
  return [A(startTime.add(attributes.arguments.growingDuration)).SEQ_ECHO(attributes.arguments.growingDuration.toString())];
}

```

### Logs

_No response_

### System Info

```shell
Aerie 3.0.1 local
```

### Severity

Moderate

mattdailis:
* **Tickets addressed:** Fixes #1625 
* **Review:** By commit  

## Description

Durations in the Aerie system take a few different forms. Sometimes they're postgres intervals, sometimes they're ISO8601, and yet other times they're integers representing microseconds.

The Duration Value Schema type uses the integer in microseconds representation.

Fixing this bug uncovered another: the relative-timed commands coming out of the EDSL expansion could end up with unbalanced durations. This PR also adds some code to rebalance durations in `durationToHms`. The approach I chose for rebalancing is simple, but not necessarily fast. My assumption is that the durations are likely to be _nearly_ balanced, and the while loops shouldn't need to do more than a handful of iterations.

Balancing stops at hours and does not continue on into days etc.

## Verification

An existing EDSL command expansion test was modified to hit this code path. It uses GrowBanana's "growingDuration" parameter as a relative time command offset.

## Documentation

N/A

## Future work

N/A

### DIFF
--- a/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
+++ b/sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
@@ -3233,13 +3233,36 @@ export function doyToInstant(doy: DOY_STRING): Temporal.Instant {
 export function durationToHms(time: Temporal.Duration): HMS_STRING {
   let { days, hours, minutes, seconds, milliseconds } = time;
 
+  const isNegative = days < 0 || hours < 0 || minutes < 0 || seconds < 0 || milliseconds < 0;
+
+  if (isNegative) {
+    days = Math.abs(days);
+    hours = Math.abs(hours);
+    minutes = Math.abs(minutes);
+    seconds = Math.abs(seconds);
+    milliseconds = Math.abs(milliseconds);
+  }
+
+  seconds += Math.floor(milliseconds / 1000);
+  milliseconds %= 1000;
+
+  minutes += Math.floor(seconds / 60);
+  seconds %= 60;
+
+  hours += Math.floor(minutes / 60);
+  minutes %= 60;
+
+  days += Math.floor(hours / 24);
+  hours %= 24;
+
+
   const DD = days !== 0 ? `${formatNumber(days, 3)}T` : '';
   const HH = days !== 0 ? formatNumber(hours, 2).replace('-', '') : formatNumber(hours, 2);
   const MM = formatNumber(minutes, 2).replace('-', '');
   const SS = formatNumber(seconds, 2).replace('-', '');
   const sss = formatNumber(milliseconds, 3).replace('-', '');
 
-  return `${DD}${HH}:${MM}:${SS}.${sss}` as HMS_STRING;
+  return `${isNegative ? '-' : ''}${DD}${HH}:${MM}:${SS}.${sss}` as HMS_STRING;
 }
 
 export function hmsToDuration(hms: HMS_STRING, epoch: boolean = false): Temporal.Duration {
--- a/sequencing-server/test/__snapshots__/command-types.spec.ts.snap
+++ b/sequencing-server/test/__snapshots__/command-types.spec.ts.snap
@@ -3236,13 +3236,36 @@ export function doyToInstant(doy: DOY_STRING): Temporal.Instant {
 export function durationToHms(time: Temporal.Duration): HMS_STRING {
   let { days, hours, minutes, seconds, milliseconds } = time;
 
+  const isNegative = days < 0 || hours < 0 || minutes < 0 || seconds < 0 || milliseconds < 0;
+
+  if (isNegative) {
+    days = Math.abs(days);
+    hours = Math.abs(hours);
+    minutes = Math.abs(minutes);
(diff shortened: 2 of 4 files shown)


sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
```diff
@@ -3233,6 +3233,36 @@ export function doyToInstant(doy: DOY_STRING): Temporal.Instant {
...
+  while (minutes >= 60) {
+    hours += 1;
+    minutes -= 60;
+  }
+
+  while (minutes <= -60) {
+    hours -= 1;
+    minutes += 60;
+  }
+
```
goetzrrGit: Instead of looping have you thought about doing something like this? I did something similar with the TS custom working before reworking it to use the time library written for the planning side.  https://github.com/NASA-AMMOS/aerie-ui/pull/820/files#diff-005afc3206d16625a1a62c09f8ba9d96cbe846da3a31c9b3c0b3822fad77cc0cR367-R389

`
  seconds += Math.floor(milliseconds / 1000);
  milliseconds %= 1000;

  minutes += Math.floor(seconds / 60);
  seconds %= 60;

  hours += Math.floor(minutes / 60);
  minutes %= 60;

  days += Math.floor(hours / 24);
  hours %= 24;
  `
You should probably do something if days are over 365 as years are not supported in a hms durations format

goetzrrGit (changes requested):
Thank you for fixing this bug. I had another approach if you wanted to take a look. Might be faster than looping if you have a big unbalanced time

sequencing-server/src/lib/codegen/CommandEDSLPreface.ts
mattdailis: Unfortunately this alternative algorithm doesn't seem to work quite right for negative values

```
  "metadata": Object {
            "simulatedActivityId": 2028,
          },
          "stem": "PREPARE_LOAF",
          "time": Object {
    -       "tag": "-001T12:16:54.000",
    +       "tag": "-003T14:18:54.000",
            "type": "EPOCH_RELATIVE",
          },
          "type": "command",
        },
```

mattdailis: The code in that time library seems to assume that the following is true:

```
x == y * Math.floor(x / y) + (x % y) 
```

Testing with a positive value:
```
var x = 11;
var y = 5;
console.log(y * Math.floor(x / y) + (x % y));

// 11
```

Testing with a negative value:
```
var x = -11;
var y = 5;
console.log(y * Math.floor(x / y) + (x % y));

// -16
```

goetzrrGit: awww yeah I forgot you can have a negative duration.

sequencing-server/test/__snapshots__/command-types.spec.ts.snap
```diff
@@ -3236,6 +3236,18 @@ export function doyToInstant(doy: DOY_STRING): Temporal.Instant {
...
+
+  minutes += Math.floor(seconds / 60);
+  seconds %= 60;
+
+  hours += Math.floor(minutes / 60);
+  minutes %= 60;
+
+  days += Math.floor(hours / 24);
+  hours %= 24;
+
```
goetzrrGit: You should probably update the snapshot again. I know it is a pain to do.

goetzrrGit (changes requested):

goetzrrGit (approved):
Looks great!!!

PR #1656, opened 2025-03-29
Bugfix: Use simulated_activity instead of span for sequence templates

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This is a follow-up to #1655 . That PR had fixed the `sequence_to_simulated_activity` -> `simulated_activity` relationship. This PR makes the sequence templates feature use that relationship instead of the `sequence_to_simulated_activity` -> `span` relationship. This PR also removes that relationship, because it turned out to be unsound.

## Verification

I did a quick manual test to make sure sequence template expansion worked. Existing end-to-end tests should also cover sequence template functionality.

## Documentation

N/A

## Future work

- There are two different simulated activity batch loaders - consider consolidating

### DIFF
--- a/deployment/hasura/metadata/databases/tables/sequencing/sequence_to_simulated_activity.yaml
+++ b/deployment/hasura/metadata/databases/tables/sequencing/sequence_to_simulated_activity.yaml
@@ -16,14 +16,6 @@ object_relationships:
       remote_table:
         name: simulated_activity
         schema: merlin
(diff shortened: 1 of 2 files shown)


pranav-super (approved):
Looks good to me! After our discussion about the difference between `dataset_id` and `simulated_dataset_id`, I think I agree with this change.

PR #1657, opened 2025-03-31
remove build context for aerie_action in deployment/docker-compose

dandelany:
## Description
Makes aerie_action consistent with other services, hopefully to fix aerie-dev deployment


goetzrrGit (approved):
Looked over this live with @dandelany

Issue #1663, opened 2025-04-01
Psyche adaptation not working/not producing valid SATFs


goetzrrGit:
### Describe the bug

Psyche currently uses a prototype adaptation layer for conversions between seqN and SATF formats. This prototype is fragile and requires frequent modifications. To address this, we aim to rewrite the adaptation layer, leveraging the satf/sasf npm library package. This will significantly reduce the amount of code the adaptation developer needs to maintain, simplifying their workflow. By utilizing the satf/sasf library package to generate 90% of the SATF/SASF structure, the adaptation layer will only need to handle the remaining 10% of mission-specific information.

Issue #1530, opened 2024-08-14
Better support for migrating plans to a new mission model version


parkerabercrombie:
### Checked for duplicates

No - I haven't checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

When a new version of a mission model is available, we will need to migrate existing plans to the new model. Right now we download the plan from the API, run an ad-hoc upgrade script, and then upload to the new model. Aerie should provide a standard workflow for this since this is a very common need.

dandelany:
Met with @lklyne and developers to start gathering requirements on this, Lyle took some notes in this Figma document:

https://www.figma.com/board/PykIfZW7SNdGoySgQjC81O/AERIE-Plan-Migration-Explorations?node-id=0-1&node-type=canvas&t=MAkhPE1CMmFN1Xxm-0

Lyle will be working on UI design for this next sprint, and we'll kick off dev/implementation discussions in earnest next week.

lklyne:
**Notes and working direction from UIWG 2024-11-20**

We have a working direction and a few open questions. General approach is to have a less structured mission model migration process that relies on our built in error handling to identify and resolve errors that come up. Notes here: https://www.figma.com/board/PykIfZW7SNdGoySgQjC81O/AERIE-Plan-Migration-Explorations?node-id=100-1459&t=akjfG5N3Eng6Ed22-4

**Workflow plans**
1. When a plan is selected on the plans page, show a button to change the mission model
2. Button opens a dialog where user can select a mission model. List of mission models is sorted reversed chronologically, and should display the date a model was uploaded. On this page, we should also show a list of potential errors and a note that a snapshot will be created.
3. After confirming, the mission model for the plan changes, and any errors are flagged in the row item
4. When a plan model migration errors is opened, we show those errors in the activity validation section of the error console.

**Updates to activity validation errors console:**
- Add "Extraneous activity types" section to the top of the list. This shows all the activity types that are not present in the new mission model. Option to "rename" these, which will associate the missing activity type with a new type. Clicking replace presents a new TBD dialog which lists activity types, and prompts the user to rename a single one, or rename all associated types.
- We should provide similar replace functionality for extraneous parameters. 
- Lyle to sync with @dandelany about details for the rename dialog. 

**Limitations with this approach**
- This pattern disallows the ability to do combined swaps. Eg no a -> b and b -> c. The workaround would be a two step migration process.

**View model migration questions**
[...]

PR #1665, opened 2025-04-02
Adding plan model migration support

joshhaug:
* **Tickets addressed:** [AERIE-1530](https://github.com/NASA-AMMOS/aerie/issues/1530)
* **Review:** By file  

## Description

This PR implements backend support for migrating plans from one model to another. It is designed to be used in conjunction with [aerie-ui PR #1663](https://github.com/NASA-AMMOS/aerie-ui/pull/1663), which provides frontend support. 

### Summary of Changes
This PR adds two Hasura functions:
* `check_model_compatability` checks whether a plan with `_plan_id` can be migrated to model `_new_model_id` by querying for:
  * any activity directives in the plan with a type that does not exist in the new model
  * any activity directives in the plan with parameter schema mismatches to the activity type in the new model
  * This function returns these stats to the UI to display. There may another way to format this info that will result in more useful messages to the user (e.g. which parameters are mismatched? which activity types are missing?). 
* `migrate_plan_to_model` checks for open merge requests, and if none exist, it:
    * creates a uniquely-named snapshot of the plan, then
    * updates the plan to a the new model
  
 Both of these functions have fine-grained permissions; only a plan owner or collaborator can perform these queries. 

## Verification

Ran tests in conjunction with the [aerie-ui changes](https://github.com/NASA-AMMOS/aerie-ui/pull/1663) to upload incompatible and compatible banananation models. 

## Documentation

I believe the docs will need to be uploaded to include these new Hasura functions. 

## Future work

* Better formatting for migration checks
* The checks I listed above + sim-level args are all I could think of that could pose an issue for a plan model migration, but there very well could be others.


deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,82 @@
...
+  from merlin.merge_request
+  where _plan_id = migrate_plan_to_model._plan_id
+    and status in ('pending', 'in-progress');
+
+  if _open_merge_count > 0 then
+      raise exception 'Cannot migrate plan %: it has open merge requests.', _plan_id;
+  end if;
+
+  -- Create snapshot before migration
+  perform merlin.create_snapshot(_plan_id, 'Migration to model '|| _new_model_id || ' on ' || NOW(), 'Automatic snapshot before attempting migration to model id ' || _new_model_id || ' on ' || NOW(), _requester_username);
```
cartermak: Note to future Carter, check whether snapshots store the model ID and, if they don't, should we change that? Then, migrated plans could still store snapshots for previous models. But maybe that's a pain...

dandelany (changes requested):
Did some testing on this with @mattdailis yesterday and it's getting close! The main issue that we saw is that the background **activity validator** doesn't get run on the updated plan after model migration, and we think it should. This validator (see `merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ValidationWorker.java`) runs in the background looking for unvalidated activities, ie. rows in the `activity_directive_validations` table with status `pending`, and validates them.

Since swapping out the mission model can change the definitions/types of activities & their parameters (or remove them entirely), we should invalidate any existing `activity_directive_validations` at the time of model migrations, causing the validator to re-run automatically. This can be done in SQL by setting their status to `"pending"`, eg. by adding this to the `hasura.migrate_plan_to_model` function: https://github.com/mattdailis/aerie/commit/2758a6537 (thanks @mattdailis )

This was the main issue, a few more notes:
* We made a new version of banananation for testing, i'll add this in another comment
* We currently don't have any validation error for "extraneous activities" & ideally should, though debatable if this is a v1 or v2 thing.
* `pgcmp` tests are failing, I think this also needs db migrations for the new functions

dandelany:
@mattdailis & I created an updated version of the Banananation model in order to test edge cases of mission model migration - attached here 
[banananation-v2.jar.zip](https://github.com/user-attachments/files/19694690/banananation-v2.jar.zip) as it will be useful for future testing. It has the following changes vs. the Banananation model in the repo:

* added `newOptionalParameter` to `BiteBanana` (as optional parameter with default)
* added`newRequiredParameter` (string) to `ChangeProducerActivity`
* removed `growingDuration` from GrowBanana
* deleted an activity type - removed BananaNapActivity
* in `PeelBanana` - changed parameter `PeelDirection` from enum type to string
* PickBanana - changed `quantity` from number to string
* in `bakeBananaBread` - changed temperature to struct (w an enum) from double
* `DurationParameterActivity` - renamed to `NewDurationParameterActivity`
* in `LineCount` - renamed "path" input to "file" (Path object)

dandelany:
I pushed the change to the validator mentioned above - now when you migrate the model, it automatically shows the errors in the new plan when you open it.

Handing this off to @AaronPlave for further testing & identifying any more issues... Others we noted:
* the table on Change Mission Model Modal should have a "version" column
* when showing expected conflicts, should either show more info, or not be selectable
   - track down what are the expected conflicts - some may not actually be

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
```
AaronPlave: Would it be possible to mention the model that is being migrated from here? Something like "Automatic snapshot before attempting migration from model X to model Y on ..."?

mattdailis:
If anyone wants to fiddle around with the modified banananation activity types, I've pushed them to a branch here: [feature/plan-model-migration--testing](https://github.com/NASA-AMMOS/aerie/tree/feature/plan-model-migration--testing)

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  update merlin.activity_directive_validations
+  set status = 'pending'
+  where plan_id = _plan_id;
+
+  return row('success')::hasura.migrate_plan_to_model_return_value;
+end
+$$;
+
+
+create table hasura.check_model_compatability_return_value(result json);
```
AaronPlave: Would be nice if this could return the list of missing activity types as well as a list of activity type + invalid parameter pairings. The UI could sort out the rollup itself but having some hint as to what the actual changes are before hitting migrate would be super useful.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  returns hasura.migrate_plan_to_model_return_value
+  volatile
+  language plpgsql as $$
+declare
+  _requester_username text;
+  _function_permission permissions.permission;
+  _open_merge_count integer;
+begin
+  _requester_username := (hasura_session ->> 'x-hasura-user-id');
+  _function_permission := permissions.get_function_permissions('migrate_plan_to_model', hasura_session);
```
Mythicaeda, line 15: Calling this out explicitly since the Docs section of the PR body seemed uncertain, [this](https://nasa-ammos.github.io/aerie-docs/deployment/advanced-permissions/#action-and-function-permissions) is the section of our website that needs to be updated with new function keys.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  perform permissions.raise_if_plan_merge_permission('migrate_plan_to_model', _function_permission);
```
Mythicaeda, line 16: Major: This currently does not check the permissions, just checks that the permission isn't a plan merging-specific key. The following code checks the permission:
```suggestion
  perform permissions.raise_if_plan_merge_permission('migrate_plan_to_model', _function_permission);
    if not _function_permission = 'NO_CHECK' then
      call permissions.check_general_permissions('migrate_plan_to_model', _function_permission, _plan_id, _requester_username);
    end if;
```

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+
+  -- Check for open merge requests
+  select count(*) into _open_merge_count
+  from merlin.merge_request
+  where _plan_id = migrate_plan_to_model._plan_id
+    and status in ('pending', 'in-progress');
+
+  if _open_merge_count > 0 then
```
Mythicaeda: There's a nonblocking and blocking component to this comment:

Blocking: `merlin.merge_request` does not have a `_plan_id` column, it has a `plan_id_receiving_changes` column (and a `plan_id_supplying_changes` column, but this is a check for if the plan has _incoming_ open merge requests). The current code is checking if `_plan_id` is equal to itself.

Strong preference: Please add DB tests for these functions. Specifically, in addition to any tests for these functions directly, I'd like a subsuite in the `Plan Collaboration` tests that checks that this function actually raises the expected exception if the plan has `pending` requests,  an `in-progress` request, both `pending` and `in-progress` requests, no requests, `pending` and `accepted` requests (or `pending` and `rejected`), and only `accepted` (or `rejected`) requests, as well as that another plan with open merge requests does not prevent a different plan from being migrated. That last test would have caught this bug in the code.

Nonblocking: This section of the code can be simplified, as you don't actually care about the count, just the existence:

```suggestion
  -- Check for open merge requests
  if exists(select from merlin.merge_request mr 
                             where mr.plan_id_receiving_changes = _plan_id 
                             and status in ('pending', 'in-progress')) then
```

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
Mythicaeda: Major: they do not and they _should_. Otherwise, we have lost the needed information to revert the plan to the old model by loading the snapshot (at least without string parsing, which is a little silly to do when we could just add a column to the `plan_snapshot` table with the info and update `create_snapshot` to fill this column). 

**Blocking**: Additionally, the function `merlin.restore_from_snapshot` needs to be updated to update the `model_id` of the plan, or else the MVP functionality of undoing a model migration by restoring from snapshot is not complete.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  _requester_username text;
+  _function_permission permissions.permission;
+  _old_model_id integer;
+  _param_mismatch_count integer := 0;
+  _missing_count integer := 0;
+
+begin
+  _requester_username := (hasura_session ->> 'x-hasura-user-id');
+  _function_permission := permissions.get_function_permissions('migrate_plan_to_model', hasura_session);
+  perform permissions.raise_if_plan_merge_permission('migrate_plan_to_model', _function_permission);
```
Mythicaeda: Why is this function using a **fine grained** permission check (that's incomplete, see comment on the other Hasura function)? This function solely **queries** information from the DB, it does not perform any mutations and therefore I would not expect it to contain a fine-grained check.

For context, in general, we only put fine-grained permissions on functions that mutate the data in the DB in some way, and more specifically on functions where users would want fine-grained control over who specifically with a given role may perform that mutation. 

If you take a look at our `.yaml` files for, let's say, `activity_directive`, you'll see that for users using the `user` role, we specify `filter`s and `check`s for `inserting`, `updating`, and `deleting`, but that we let them `select` freely. Fine-grained permissions are our way of providing these `filter`s and `check`s to our functions, since Hasura does not natively support any such logic.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+
+  -- Get the old model ID associated with the plan
+  select model_id into _old_model_id from merlin.plan where id = _plan_id;
+
+  -- Count missing activity types (activities that exist in the old model but not in the new model)
```
Mythicaeda: This code doesn't count missing activity types, it counts activities on the plan that do not have a type in the new model. This can cause the count to be inaccurate. Unless that is the intended functionality? If so, then the comments within the function does not align with what the code is doing and should be altered.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  _param_mismatch_count integer := 0;
+  _missing_count integer := 0;
+
+begin
+  _requester_username := (hasura_session ->> 'x-hasura-user-id');
+  _function_permission := permissions.get_function_permissions('migrate_plan_to_model', hasura_session);
+  perform permissions.raise_if_plan_merge_permission('migrate_plan_to_model', _function_permission);
+
+  -- Get the old model ID associated with the plan
+  select model_id into _old_model_id from merlin.plan where id = _plan_id;
```
Mythicaeda: I'd actually recommend making the function take `_old_model_id` and `_new_model_id`, but that's inspired by my proposal to essentially make this function check if two models are compatible and not how compatible the plan is with the new model. If this is checking intends to check the latter, then the current params make sense.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+  set status = 'pending'
+  where plan_id = _plan_id;
+
+  return row('success')::hasura.migrate_plan_to_model_return_value;
+end
+$$;
+
+
+create table hasura.check_model_compatability_return_value(result json);
+create function hasura.check_model_compatability(_plan_id integer, _new_model_id integer, hasura_session json)
```
Mythicaeda: Minor: please add a comment to both the new functions.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,87 @@
...
+         join merlin.activity_type old_at on old_at.model_id = _old_model_id and old_at.name = ad.name
+         join merlin.activity_type new_at on new_at.model_id = _new_model_id and new_at.name = ad.name
+  where ad.plan_id = _plan_id
+    and old_at.parameters <> new_at.parameters;
+
+  -- Return JSON object with counts
+  return row(json_build_object(
+      'missing_count', _missing_count,
+      'param_mismatch_count', _param_mismatch_count
+         ))::hasura.check_model_compatability_return_value;
```
Mythicaeda: Changing the `json_build_object` to the following will return which activity **types** have been removed and which have had their parameter schemas changed, which should address @AaronPlave's request. If you still want to provide a list of the impacted **directives** in the plan, they can be added as additional entries into the return JSON:

```sql
_removed_activity_types := coalesce((
      select json_agg(name)
      from merlin.activity_type old_at
      where old_at.model_id = 1
        and not exists(select from merlin.activity_type new_at
                              where new_at.name = old_at.name
                                and new_at.model_id = 2)
      ), '{}'::json);

  _altered_activity_types := coalesce((
      select json_object_agg(types.n, types.t) from (
        select type.name as n,
               json_build_object('old_parameter_schema', old_params, 'new_parameter_schema', new_params) as t
        from (
          select new_type.name,
                 old_type.parameters as old_params,
                 new_type.parameters as new_params
          from merlin.activity_type new_type
          left join (select name, parameters from merlin.activity_type where model_id = 1) old_type
          using (name)
          where new_type.model_id = 2
            and old_type.parameters <> new_type.parameters
          ) type
        ) types
    ), '{}'::json);

return row(json_build_object(
    'removed_activity_types', _removed_activity_types
    'altered_activity_types', _altered_activity_types
))::hasura.check_model_compatability_return_value;

```

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
Mythicaeda: Additionally, these permissions are not using the key added in `permissions.sql`

deployment/postgres-init-db/sql/types/permissions/permissions.sql
```diff
@@ -35,6 +35,7 @@ create type permissions.function_permission_key
     'begin_merge',
     'branch_plan',
     'cancel_merge',
+    'check_model_compatability',
```
Mythicaeda: This key is unused (see comment on the `check_model_compatability` function)

Mythicaeda (changes requested):
Marking this as "Request Changes" due to some bugs in the functions.

Major: `simulation_dataset` needs to be updated to track the `model_id`, otherwise the UI will be unable to properly display old simulation results following a model migration.

Major: `create_merge_request` needs to be updated to _block_ merge requests between plans using different mission models (If this aspect of the feature has been changed, please let me know and I will update this part of the comment).

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Thanks, updated in ea5fdc94d74554be63cd2417e642012851e439cc.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Good point, added in 4c782dd6e54846f56f486795371f48d0cf62dde1.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Thanks, fixed in dac23b04b5d4bd907cb88fc5bd5f31e66178d384.

Mythicaeda: You also need to remove it from `default_user_roles.sql`

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: PR for the docs change is available [here](https://github.com/NASA-AMMOS/aerie-docs/pull/227).

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Check is updated in 61f4210dd9ad5e623d5531d0bcba2af12db482da, working on DB tests now.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Duh, updated in 67753a4b7fc53d2e4e06392ae388a43b3a5b36f5

deployment/postgres-init-db/sql/types/permissions/permissions.sql
joshhaug: Removed in https://github.com/NASA-AMMOS/aerie/commit/dac23b04b5d4bd907cb88fc5bd5f31e66178d384.

AaronPlave:
> If anyone wants to fiddle around with the modified banananation activity types, I've pushed them to a branch here: [feature/plan-model-migration--testing](https://github.com/NASA-AMMOS/aerie/tree/feature/plan-model-migration--testing)

@mattdailis and others – Any thoughts on how we might store this sort of altered bananation modal for end to end testing in the UI (not sure if the backend needs such a model for testing). Easy enough to store this altered model as it is now in the UI repo but if the model needs to be updated in the future things could get weird.

AaronPlave:
This may be a different chunk of work but is there a way to determine which model was used for a simulation? Now that a plan can change models at any time there is no stable reference to a model ID used for a simulation (if I'm not mistaken).

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Good point. See changes in 20a0851d99dc995e7ff34eddb6783b674f04fe79 which invalidate this comment

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Done, good point: 20a0851d99dc995e7ff34eddb6783b674f04fe79

joshhaug:
Added plan snapshot support in d4ecbe169dea727786603f3ca938cebe7e995324. See https://github.com/NASA-AMMOS/aerie/pull/1665#discussion_r2046988746 for the bug this fixes.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Done in d4ecbe169dea727786603f3ca938cebe7e995324. We now return a json object like so:
```json
{
    "removed_activity_types": {}
    "altered_activity_types": { "BiteBanana" : {"old_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real"}}}, "new_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}, "secondParameter": {"order": 1, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}}} } 
}
```

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Good point, updated in 53b9451ce11cd2e6458c3974ef676ad53f3b1cfe.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Updated in 53b9451ce11cd2e6458c3974ef676ad53f3b1cfe.

joshhaug:
@AaronPlave some breaking changes to the backend, both concerning the `check_model_compatability` function:
* The function now takes parameters `(_old_model_id integer, _new_model_id integer)` and does not require the `_plan_id`. 
* The output of this function is now a json object like:
```json
{
    "removed_activity_types": {}
    "altered_activity_types": { "BiteBanana" : {"old_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real"}}}, "new_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}, "secondParameter": {"order": 1, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}}} } 
}
```

If you'd like me to update the UI for these changes just lmk, I'm happy to take a stab at it.

AaronPlave:
> @AaronPlave some breaking changes to the backend, both concerning the `check_model_compatability` function:
> 
> * The function now takes parameters `(_old_model_id integer, _new_model_id integer)` and does not require the `_plan_id`.
> * The output of this function is now a json object like:
> 
> ```json
> {
>     "removed_activity_types": {}
>     "altered_activity_types": { "BiteBanana" : {"old_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real"}}}, "new_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}, "secondParameter": {"order": 1, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}}} } 
> }
> ```
> 
> If you'd like me to update the UI for these changes just lmk, I'm happy to take a stab at it.

Thanks! I'll take a crack at it and will let you know if I run into any issues.

AaronPlave:
> > @AaronPlave some breaking changes to the backend, both concerning the `check_model_compatability` function:
> > 
> > * The function now takes parameters `(_old_model_id integer, _new_model_id integer)` and does not require the `_plan_id`.
> > * The output of this function is now a json object like:
> > 
> > ```json
> > {
> >     "removed_activity_types": {}
> >     "altered_activity_types": { "BiteBanana" : {"old_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real"}}}, "new_parameter_schema" : {"biteSize": {"order": 0, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}, "secondParameter": {"order": 1, "schema": {"type": "real", "metadata": {"unit": {"value": "m"}}}}}} } 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you'd like me to update the UI for these changes just lmk, I'm happy to take a stab at it.
> 
> Thanks! I'll take a crack at it and will let you know if I run into any issues.

@joshhaug Seeing an issue where `altered_activity_types` and `removed_activity_types` seem to be strings instead of actual json objects.

Also would it be possible to add in a list of directives that are impacted for the plan? My gut here is that users would probably be after getting a sense of how many problems they will have when migrating and for each category (altered and removed), which types are involved plus a count of those types. Could let users dive a little deeper into which parameters have actually changed but that feels out of scope. The general diff between two models is definitely useful but I don't think it is quite enough in the context of a specific plan when trying to figure out if migration is roughly feasible.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Added some db tests in a770b5b8704d8f82ae13535af64b6d9c6052f5d0

joshhaug:
@AaronPlave There's another function available as of 11466fa8ca665624d806ae5f56187ae1b90b8bbb, function name is `check_model_compatability_for_plan` and takes a plan id and new model id. It returns something like the following:
```json
{
  "data": {
    "check_model_compatability_for_plan": {
      "result": {
        "removed_activity_types": [],
        "altered_activity_types": {
          "BiteBanana": {
            "old_parameter_schema": {
              "biteSize": {
                "order": 0,
                "schema": {
                  "type": "real"
                }
              }
            },
            "new_parameter_schema": {
              "biteSize": {
                "order": 0,
                "schema": {
                  "type": "real",
                  "metadata": {
                    "unit": {
                      "value": "m"
                    }
                  }
                }
              },
              "secondParameter": {
                "order": 1,
                "schema": {
                  "type": "real",
                  "metadata": {
                    "unit": {
                      "value": "m"
                    }
                  }
                }
              }
            }
          }
        },
        "problematic_directives": [
          {
            "activity_directive": {
              "id": 3,
              "plan_id": 1,
              "name": "BiteBanana",
              "source_scheduling_goal_id": null,
              "created_at": "2025-05-01T22:54:59.010601+00:00",
              "created_by": "josh",
              "last_modified_at": "2025-05-01T22:54:59.010601+00:00",
              "last_modified_by": "josh",
              "start_offset": "09:18:35.044",
              "type": "BiteBanana",
              "arguments": {},
              "last_modified_arguments_at": "2025-05-01T22:54:59.010601+00:00",
              "metadata": {},
              "anchor_id": null,
              "anchored_to_start": true
            },
            "issue": "altered"
          }
        ]
      }
    }
  }
}
```

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
```diff
@@ -315,4 +329,77 @@ begin
...
-call migrations.mark_migration_applied('18');
+
+create table hasura.check_model_compatability_for_plan_return_value(result json);
+
+/*
+* This function checks whether a plan is compatible with a given model. It returns a json object containing:
+*     * removed_activity_types, containing the activity types that are in the old model and not in the new model
+*     * altered_activity_types, containing the activity types with dissimilar parameter schemas, and the old and
+*            new parameter schemas for this activity type
+*     * impacted_directives, containing a list of the directives in the plan that fall into one of the two above categories
```
AaronPlave: Should "impacted_directives" be "problematic_directives"?

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
```diff
@@ -0,0 +1,405 @@
+alter table merlin.plan_snapshot
+add column model_id integer references merlin.mission_model on delete set null default null;
```
Mythicaeda: Column comment needs to be added.

deployment/postgres-init-db/sql/functions/merlin/snapshots/restore_from_snapshot.sql
```diff
@@ -22,10 +24,20 @@ create procedure merlin.restore_from_snapshot(_plan_id integer, _snapshot_id int
           _snapshot_id, _plan_name, _plan_id;
       end if;
     end if;
+    select model_id from merlin.plan_snapshot where snapshot_id = _snapshot_id into _model_id;
+    select name from merlin.mission_model where id = _model_id into _model_name;
+		if _model_name is null then
+      raise exception 'Cannot Restore: Model with ID % does not exist.', _model_id;
```
Mythicaeda: Minor: you don't need a whole variable for the model name when this is just `if exists(select from merlin.mission_model m where m.id = _model_id)`

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,199 @@
...
+  language plpgsql as $$
+declare
+  _removed_activity_types json;
+  _altered_activity_types json;
+
+begin
+
+  if not exists (select 1 from merlin.mission_model where id = _old_model_id)
+    or not exists (select 1 from merlin.mission_model where id = _new_model_id) then
+    raise exception 'One or both models (% and %) do not exist, not proceeding with plan compatability check.', _old_model_id, _new_model_id;
```
Mythicaeda: Minor, typo: 
```suggestion
    raise exception 'One or both models (% and %) do not exist, not proceeding with model compatibility check.', _old_model_id, _new_model_id;
```

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
```diff
@@ -0,0 +1,405 @@
...
+
+-- Backfill the model_id column for existing snapshots with the current model_id for the snapshot's plan
+update merlin.plan_snapshot snap
+set model_id = plan.model_id
+from merlin.plan
+where plan.id = snap.plan_id;
+
+
+-- Alter snapshot functions
+drop function merlin.create_snapshot();
```
Mythicaeda: Major: wrong function signature:
```suggestion
drop function merlin.create_snapshot(_plan_id integer);
```

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
Mythicaeda: I think `impacted_directives` makes more sense, as the directives aren't inherently `problematic`, they're just ones that will be impacted by the change.

AaronPlave: Ah fair, I was mostly after having this word match the actual output below.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,199 @@
...
+
+  -- Get the old model ID associated with the plan
+  select model_id into _old_model_id from merlin.plan where id = _plan_id;
+
+  -- Create snapshot before migration
+  perform merlin.create_snapshot(_plan_id,
+                                 'Migration from model ' || _old_model_id ||
+                                 ' to model ' || _new_model_id ||
+                                 ' on ' || NOW(),
+                                 'Automatic snapshot before attempting migration from model id ' || _old_model_id ||
```
AaronPlave: Would you be open to also including model names in this message? Makes it more human readable inside of the plan when looking at snapshots.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,199 @@
...
+                                 ' to model id ' || _new_model_id ||
+                                 ' on ' || NOW(),
+                                 _requester_username);
```
Mythicaeda: Nonblocking, ignorable: Condensing snapshot title and removing incorrect phrase "attempting":
```suggestion
  perform merlin.create_snapshot(_plan_id,
                                 'Migration from model ' || _old_model_id ||
                                 ' to model ' || _new_model_id,
                                 'Automatic snapshot before migrating from model id ' || _old_model_id ||
                                 ' to model id ' || _new_model_id ||
                                 ' on ' || NOW(),
                                 _requester_username);
```

AaronPlave:
While I was testing with the UI branch I discovered that I was able to make a merge request from a branch with a different model than the parent and successfully merge that branch. Not sure if this has been talked about in a comment somewhere or if this behavior is desired (probably not?) so wanted to mention it here to see if the backend could prevent those merge requests from being submitted/merged.

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
joshhaug: Good point, addressed in 45d1fae0e61406dbdcf40f42bf25a6c0691a9bed.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Functions, files, etc. were all named `compatability`. Fixed all these in fc5d295031ec5019a22a1879d6d943cf8496c339. @AaronPlave this will likely need to be adjusted in the UI too. Sorry for the typo.

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
joshhaug: Added in ac9c10f7cc1b2300d1a6d8cb39fc85e9b8ca3705.

deployment/postgres-init-db/sql/functions/merlin/snapshots/restore_from_snapshot.sql
joshhaug: Thanks, added in ac9c10f7cc1b2300d1a6d8cb39fc85e9b8ca3705.

deployment/hasura/migrations/Aerie/18_plan_model_migration/up.sql
joshhaug: Good catch, fixed in ac9c10f7cc1b2300d1a6d8cb39fc85e9b8ca3705

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Updated in ac9c10f7cc1b2300d1a6d8cb39fc85e9b8ca3705

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: Good point, added in ac9c10f7cc1b2300d1a6d8cb39fc85e9b8ca3705

joshhaug:
> While I was testing with the UI branch I discovered that I was able to make a merge request from a branch with a different model than the parent and successfully merge that branch. Not sure if this has been talked about in a comment somewhere or if this behavior is desired (probably not?) so wanted to mention it here to see if the backend could prevent those merge requests from being submitted/merged.

We do check to see whether there are any open merge requests when performing migration, but we don't check for this specific use case. Is it reasonable to say that merges should only be allowed between plans with the same `model_id`?

joshhaug:
Added `model_id` column to `simulation_dataset` in dfbe4a12b1dad072565f60a53d97432d9f75af50. This may need some extra review, I modified the trigger function to auto-populate this field but I was unable to find other locations in the codebase where this is being explicitly set.

mattdailis:
> Is it reasonable to say that merges should only be allowed between plans with the same `model_id`?

I think that's a reasonable restriction for the time being. The remedy is to migrate one or both plans to the same model prior to initiating a merge - that seems like a straightforward story to tell users

AaronPlave:
> > While I was testing with the UI branch I discovered that I was able to make a merge request from a branch with a different model than the parent and successfully merge that branch. Not sure if this has been talked about in a comment somewhere or if this behavior is desired (probably not?) so wanted to mention it here to see if the backend could prevent those merge requests from being submitted/merged.
> 
> We do check to see whether there are any open merge requests when performing migration, but we don't check for this specific use case. Is it reasonable to say that merges should only be allowed between plans with the same `model_id`?

I think that is a reasonable restriction though probably good to check with @Mythicaeda

Mythicaeda:
> > Is it reasonable to say that merges should only be allowed between plans with the same `model_id`?
> 
> I think that's a reasonable restriction for the time being. The remedy is to migrate one or both plans to the same model prior to initiating a merge - that seems like a straightforward story to tell users

I agree, and it's one of the blocking issues I left in my initial review

mattdailis:
A question from demos: what happens if a model is deleted?

Mythicaeda:
> A question from demos: what happens if a model is deleted?

This is ill-defined in Aerie in general, but the general current practice is to `set null` rather than cascade-delete

dandelany:
@mattdailis & I did some work on this yesterday and:

* Added a backend check to disallow beginning a merge request if the branches have different `model_id`s
* Fixed most of the migration checks, though a few "down" checks are still broken

Additionally, @joshhaug has added (since last review):

* tracking `model_id` in simulation dataset
* updating compatibility check to check *plan* compatibility, so it only shows relevant errors

@Mythicaeda I think all of your "majors" have been addressed, we just need to wrap up fixing any remaining broken tests

dandelany:
@Mythicaeda will review final DB changes

@dandelany will work on fixing remaining pgcmp-down test failures

@joshhaug will look into E2E test failure - per @Mythicaeda we suspect a missing cast to jsonb on a string entry.

deployment/postgres-init-db/sql/functions/merlin/snapshots/restore_from_snapshot.sql
```diff
@@ -22,10 +24,19 @@ create procedure merlin.restore_from_snapshot(_plan_id integer, _snapshot_id int
           _snapshot_id, _plan_name, _plan_id;
       end if;
     end if;
+    select model_id from merlin.plan_snapshot where snapshot_id = _snapshot_id into _model_id;
+    if not exists(select from merlin.mission_model m where m.id = _model_id) then
```
Mythicaeda, line 27: Nonblocking Question: Why do we want to prevent restoring a deleted model?

To be clear, I'm okay with this check, I'm just curious about _your_ reasoning in particular, as I could also see an argument to allow this but to set the `model_id` to `null` instead of to a non-existent value

joshhaug: Good question. If I were using this, restoring to a "broken" state where the model no longer exists seems like it may be confusing. My mental model here is: snapshots represent working past iterations of the plan/model state. If the model is deleted, there was _probably_ a reason and therefore that snapshot is made retroactively invalid. I don't have a strong rationale beyond that, and I'm happy to change it if we think this is a use case we should support.

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
joshhaug: For posterity, we are now tracking/setting `model_id` in `create_snapshot` and `restore_from_snapshot`

joshhaug:
@Mythicaeda if all your comments are resolved, would you mind adding an approval? 🙏

Mythicaeda:
@joshhaug Yes, I would mind adding an approval, because the DB migrations are not passing, meaning the DB code is not ready. Additionally, I know that you will need to update your DB code following the merge of https://github.com/NASA-AMMOS/aerie/pull/1689, as that will update the plan snapshot functions, which your code touches.

dandelany:
As of *now* all checks are passing on this branch, blockers for merging this are:
* #1689 should be merged when ready
* After that merges, the migration number on this branch will need to be incremented one more time & any new pgcmp failures fixed - & then I will approve
* Per standup @Mythicaeda will double check that her blocker issues have been addressed & then she will approve

joshhaug:
@Mythicaeda @dandelany PR has been rebased/updated to include the latest changes

deployment/postgres-init-db/sql/functions/merlin/merging/get_merge_base.sql
```diff
@@ -1,3 +1,6 @@
+/*
+  - Find the latest snapshot that exists in the history of both the receiving and supplying plans
```
Mythicaeda: I have mixed feelings on this comment's accuracy, as it doesn't get the latest snapshot that exists in the history of both plans, it gets the latest snapshot in the history up to `snapshot_id_supplying_changes`, as indicated by the second parameter being not a `plan_id` but a `snapshot_id`. 

In practice, those should usually be the same thing, but if you were to:
1.  Open a merge request.
2.  Make an edit on the supplying plan an open a new merge request
3. Merge that new merge request
then when you get the merge base of the request open in step (1) it should not be the snapshot taken when the merge is committed (as would be the case with any new merge requests), but whatever the original merge base for the request was.

I can whip up a db test for this, since I'm not seeing one in the `MergeBaseTests`

dandelany (approved):
Rebased & updated migration again now that the cancellation PR merged - this is ready to go from my perspective. Requesting final approval from @Mythicaeda

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,205 @@
...
+$$;
+
+
+create table hasura.check_model_compatibility_return_value(result json);
+/*
+* This function checks whether two models are compatible. It returns a json object containing:
+*     * removed_activity_types, containing the activity types that are in the old model and not in the new model
+*     * modified_activity_types, containing the activity types with dissimilar parameter schemas, and the old and
+*            new parameter schemas for this activity type
+*/
```
Mythicaeda: Nonblocking: the proper way to add a comment to a function in SQL is via the `comment on` command. While commenting it this way lets the develop read it whenever they look at this file, by adding a comment via the `comment on` command, that comment is attached to the function, and will appear when asking for its DDL representation. 
This allows 
1) for this function definition to be accessible to an API user
2) IntelliJ to find the function comment in the same way it is able to find JavaDoc comments on Java methods.
3) Developers to see the comment should they examine the function in the DB, as opposed to just in the SQL code

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
```diff
@@ -0,0 +1,205 @@
...
+create function hasura.check_model_compatibility(_old_model_id integer, _new_model_id integer, hasura_session json)
```
Mythicaeda: `hasura_session` on this function is unused. Removing it would also make it unused on `check_model_compatibility_for_plan`. Please remove it from both functions, alongside the `session_argument` parameter in `functions.yaml`. 

This change will **not** impact any downstream clients, as Hasura does not expose the `session_argument` parameter of a function.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -246,6 +248,21 @@ int getParentPlanId(final int planId) throws SQLException{
     }
   }
 
+  private List<Integer> getPlanMetadata(final int planId) throws SQLException{
```
Mythicaeda: Why is this called `getPlanMetadata`? That's not what it does, it gets the plan's ancestry list.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3546,4 +3614,129 @@ void tagsAreNotShuffledDuringMerge() throws SQLException {
       assertEquals(case8Tags, tagsHelper.getTagsOnActivity(case8Id, planId));
     }
   }
+
+  @Nested
+  class PlanMigrationTests{
```
Mythicaeda: These tests belong in their own file. `PlanCollaborationTests` are for Plan Branching and Merging-related tests. That means that a test suite about how a feature interacts with merging could go in this file (ie, `checkFailsForOpenMergeRequests`), but not general tests for the feature.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3546,4 +3614,129 @@ void tagsAreNotShuffledDuringMerge() throws SQLException {
...
+      final int existingPlanId = merlinHelper.insertPlan(missionModelId);
+
+      try{
+        migratePlanToModel(-1, missionModelId);
+        fail();
+      }
+      catch (SQLException sqEx){
+        if(!sqEx.getMessage().contains("Plan -1 does not exist"))
+          throw sqEx;
+      }
```
Mythicaeda: Please use `assertThrows` to model this behavior. This applies to every location of this pattern. The `try-catch-fail` pattern is vulnerable to forgetting to add the `fail` statement (which has occurred in the past).

```suggestion
final var sqEx = assertThrows(SQLException.class, () -> migratePlanToModel(-1, missionModelId));
if(!sqEx.getMessage().contains("Plan -1 does not exist")) {
    throw sqEx;
}
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3546,4 +3614,129 @@ void tagsAreNotShuffledDuringMerge() throws SQLException {
...
+
+
+    /**
+     * Check that migration fails if there exist open merge requests on _plan_id.
+     */
+    @Test
+    void checkFailsForOpenMergeRequests() throws SQLException {
+      final int planId = merlinHelper.insertPlan(missionModelId);
+      final int modelFileId = merlinHelper.insertFileUpload();
+      final int modelId = merlinHelper.insertMissionModel(modelFileId);
```
Mythicaeda: The initial plan and the second model upload should be handled in a `before` for this test suite, as every test but the `input validation` test relies on them.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3546,4 +3614,129 @@ void tagsAreNotShuffledDuringMerge() throws SQLException {
...
+        throw sqEx;
+      }
+    }
+
+
+    /**
+     * Verify that a compatability check will not show any differences for identical models.
+     */
+    @Test
+    void verifyCompatibilityCheckWorks() throws SQLException {
```
Mythicaeda: This does not adequately check that the compatibility check works, just that it returns nothing when there are no changes. It doesn't prove that it returns changes when they exist.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3546,4 +3614,129 @@ void tagsAreNotShuffledDuringMerge() throws SQLException {
...
+    @Test
+    void verifySnapshotCreated() throws SQLException {
+      final int planId = merlinHelper.insertPlan(missionModelId);
+      final int newModelFileId = merlinHelper.insertFileUpload();
+      final int newModelId = merlinHelper.insertMissionModel(newModelFileId);
+
+      try {
+        migratePlanToModel(planId, newModelId);
+        final List<Integer> snapshotIds = getLatestSnapshots(planId);
+        assertEquals(snapshotIds.size(), 1);
```
Mythicaeda: Given that the method makes a snapshot with a deliberate name and description, this test should check for it using a `startsWith` check. (Keep in mind that the current merge tests only don't check for such information because they're not providing it into the snapshot)

deployment/postgres-init-db/sql/functions/merlin/snapshots/restore_from_snapshot.sql
```diff
@@ -3,6 +3,8 @@ create procedure merlin.restore_from_snapshot(_plan_id integer, _snapshot_id int
 	declare
 		_snapshot_name text;
 		_plan_name text;
+		_model_name text;
```
Mythicaeda: Minor: `_model_name` is unused, please remove it.

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
```diff
@@ -0,0 +1,514 @@
...
+  returning * into dataset_ref;
+  new.dataset_id = dataset_ref.id;
+  new.dataset_revision = dataset_ref.revision;
+return new;
+end$$;
+
+create trigger set_revisions_and_initialize_dataset_on_insert_trigger
+  before insert on merlin.simulation_dataset
+  for each row
+execute function merlin.set_revisions_and_initialize_dataset_on_insert();
```
Mythicaeda: Continual issue in both up and down: When _updating_ a function, the preference is to use `create or replace` instead of `drop` and `create`, unless there is an issue that would prevent that (such as adding/removing a parameter, altering a parameter's type, etc). 

The "drop and create" approach can cause the migration to otherwise needlessly touch other parts of the DB (ie, the trigger `set_revisions_and_initialize_dataset_on_insert_trigger`). It also makes it harder to notice comment changes, as in "drop and create" the comment must always be recreate, while with "create or replace", the comment will only be set if it changed.

For this instance, and similar ones where "drop and create" require what should be unimpacted parts of the schema to be dropped and recreated, I'd like it fixed. For instances where this is not the case (ie L16-17), it _can_ be considered non-blocking, but I'd strongly _prefer_ if it was addressed, especially since its a very quick fix (just remove the `drop` statement and add `or replace` to the `create` statement).

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
```diff
@@ -0,0 +1,514 @@
...
+
+-- Modify simulation_dataset to include model_id
+alter table merlin.simulation_dataset
```
Mythicaeda: Nonblocking, but style critique: you should alter a table _before_ you change the functions on that table.

deployment/postgres-init-db/sql/types/permissions/permissions.sql
```diff
@@ -48,6 +48,7 @@ create type permissions.function_permission_key
     'get_conflicting_activities',
     'get_non_conflicting_activities',
     'get_plan_history',
+    'migrate_plan_to_model',
```
Mythicaeda, line 53: MAJOR, BLOCKING: This change was not included in the migration (no, I don't know why the migration is passing with the enum definitions differing)

deployment/postgres-init-db/sql/default_user_roles.sql
```diff
@@ -38,6 +38,7 @@ set action_permissions = '{
       "get_conflicting_activities": "NO_CHECK",
       "get_non_conflicting_activities": "NO_CHECK",
       "get_plan_history": "NO_CHECK",
+      "migrate_plan_to_model": "PLAN_OWNER_COLLABORATOR",
```
Mythicaeda, line 43: MAJOR, BLOCKING: This change was not in the migration. You can address it by adding the following line:

In the `up.sql`
```sql
update permissions.user_role_permission
  set function_permissions = function_permissions
                               || jsonb_build_object('migrate_plan_to_model', 'PLAN_OWNER_COLLABORATOR')
where role = 'user';
```

In the `down.sql`
```sql
update metadata.user_role_permission
  set function_permissions = function_permissions  - 'migrate_plan_to_model';
```

Note that in the `down`, we don't provide a `where` clause, as we want to remove the key from any role that has it assigned, while in the `up`, we only want to update our `user` role while not touching the client's custom roles/`aerie_admin`/`viewer`

Mythicaeda (commented):
There are two `MAJOR, BLOCKING` issues with the migration preventing approval. I will investigate why `pgcmp` missed the `enum` issue.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
Mythicaeda: It's also unused, please remove it.

deployment/postgres-init-db/sql/functions/merlin/merging/get_merge_base.sql
dandelany: improved the wording of the comment, we can talk about whether or not a test makes sense

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
dandelany: fixed (here & in a few other places at least)

deployment/postgres-init-db/sql/functions/hasura/plan_migration_functions.sql
dandelany: fixed

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
dandelany: removed

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
dandelany: fixed

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
dandelany: fixed in up and down migrations 👍

deployment/postgres-init-db/sql/types/permissions/permissions.sql
dandelany: fixed, maybe we can talk offline about better strategies for `enum`s since migrating them is painful

deployment/postgres-init-db/sql/default_user_roles.sql
dandelany: fixed

deployment/postgres-init-db/sql/functions/merlin/snapshots/restore_from_snapshot.sql
dandelany: removed

dandelany:
@Mythicaeda I've addressed all of your blockers above and most of the smaller ones - there are a few remaining, mostly testing-related, that would be good to discuss with @joshhaug next week. Tests are passing and this is ready for another look from you

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
```diff
@@ -3628,33 +3614,23 @@ class PlanMigrationTests{
...
-        fail();
-      }
-      catch (SQLException sqEx){
-        if(!sqEx.getMessage().contains("Model -1 does not exist"))
-          throw sqEx;
-      }
+      assertThrows(
+          SQLException.class,
+          () -> migratePlanToModel(-1, missionModelId),
+          "Plan -1 does not exist"
```
Mythicaeda: This is the message the test fails with if the code doesn't throw a SQLException, and is not equivalent to the `if` check from before. The following code is:

```java
final var sqEx =  assertThrows(SQLException.class, () -> migratePlanToModel(-1, missionModelId));
if(!sqEx.getMessage().contains("Model -1 does not exist")) {
   throw sqEx;
}
```

dandelany: ~fixed in [4eb45a1](https://github.com/NASA-AMMOS/aerie/pull/1665/commits/4eb45a14c2bb86afa51cb26e3eb8284cfb7ae624)~ not actually fixed yet, will be soon

dandelany: Okay, this is back to passing and ready for another look @Mythicaeda, TL;DR good catch and sorry for the confusion.

After trying a few things with @mattdailis I think my preferred pattern here is something like:
```
assertTrue(
  sqlEx.getMessage().contains("Plan -1 does not exist"), 
  "bad error message, got " + sqlEx.getMessage()
);
```
since it is still an assertion failure but preserves the message so you can see what was unexpected? This might be a good thing to have a util for (`assertThrowsWithMessage`), but I think we've gone deep enough on this one for now :)

Mythicaeda: That's fine, although some context as to why the reason that the general pattern has been `if(!ex.contains(message)) { throw ex }` (plus or minus checking the SQL exception code, if a custom one has been set):

If there is an exception of any kind during a DB request, it will come back as a SQLException. This includes exceptions due to DB connection issues. The purpose of the "if-not-throw" pattern is to have the test correctly end in an exception if one was thrown, similar to how currently in the E2E tests, if there's an IOException while connecting to the Gateway, the test is ended via an IOException.

dandelany: Fair, will follow that pattern in future

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
```diff
@@ -0,0 +1,526 @@
...
+  insert into merlin.merge_request(plan_id_receiving_changes, snapshot_id_supplying_changes, merge_base_snapshot_id, requester_username)
+    values(plan_id_receiving, supplying_snapshot_id, merge_base_snapshot_id, request_username)
+    returning id into merge_request_id;
+  return merge_request_id;
+end
+$$;
+
+-- Alter set_revisions_and_initialize_dataset_on_insert
+-- drop trigger first & recreate after
+drop trigger set_revisions_and_initialize_dataset_on_insert_trigger on merlin.simulation_dataset;
```
Mythicaeda: This `drop trigger` and the `create trigger` on L276 are not needed. Similarly, in the `down.sql`, the statements on L314 and L346 aren't needed.

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
Mythicaeda: This comment was unaddressed (neither accepted nor rejected)

deployment/hasura/migrations/Aerie/22_plan_model_migration/down.sql
```diff
@@ -0,0 +1,568 @@
...
+drop function permissions.get_function_permissions(_function permissions.function_permission_key, hasura_session json);
+
+------ 2. drop permissions.function_permission_key and recreate without 'migrate_plan_to_model'
+drop type permissions.function_permission_key;
+create type permissions.function_permission_key
+as enum ('apply_preset', 'begin_merge', 'branch_plan', 'cancel_merge', 'commit_merge', 'create_merge_rq',
+  'create_snapshot', 'delete_activity_reanchor', 'delete_activity_reanchor_bulk', 'delete_activity_reanchor_plan',
+  'delete_activity_reanchor_plan_bulk', 'delete_activity_subtree', 'delete_activity_subtree_bulk', 'deny_merge',
+  'get_conflicting_activities', 'get_non_conflicting_activities', 'get_plan_history',  'restore_activity_changelog',
+  'restore_snapshot', 'set_resolution', 'set_resolution_bulk', 'withdraw_merge_rq');
```
Mythicaeda: Nonblocking, Readability: please put the enum entries on separate lines. It makes reviewing the statement easier, as it lets me compare it side by side with how the enum is declared in `permissions.sql`. (Only flagging it as "nonblocking" because I already reviewed this statement).

```suggestion
create type permissions.function_permission_key
as enum (
  'apply_preset',
  'begin_merge',
  'branch_plan', 
  'cancel_merge', 
  'commit_merge', 
  'create_merge_rq',
  'create_snapshot', 
  'delete_activity_reanchor', 
  'delete_activity_reanchor_bulk', 
  'delete_activity_reanchor_plan',
  'delete_activity_reanchor_plan_bulk', 
  'delete_activity_subtree', 
  'delete_activity_subtree_bulk', 
  'deny_merge',
  'get_conflicting_activities', 
  'get_non_conflicting_activities', 
  'get_plan_history',  
  'restore_activity_changelog',
  'restore_snapshot', 
  'set_resolution', 
  'set_resolution_bulk', 
  'withdraw_merge_rq');
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
joshhaug: I've decided to just move all plan migration related tests into their own file in 72602862b2bb156f32d33b4aa69bb1bc6ad1bb21

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
joshhaug: Updated in fb73e8ebcf97cb8c392b6a29d540dedc42c4bc4e

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
joshhaug: Addressed in 2c7728be52a8626e817e28e8c631095f70c7f2b5

deployment/hasura/migrations/Aerie/22_plan_model_migration/up.sql
joshhaug: Good hygiene noted, punting for now to expedite delivery

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
joshhaug: Fixed in 133f84c46f28ec092668214a552224d7a78aeeca

joshhaug: While fixing this, discovered a bug in the way that `impacted_directives` is being populated. In `plan_migration_functions.sql` we were doing the following:
```sql
    problematic as (
      select to_json(ad) as activity_directive, 'removed' as issue
      from merlin.activity_directive ad
             join removed_names r on r.name = ad.name
      where ad.plan_id = _plan_id

      union all

      select to_json(ad) as activity_directive, 'altered' as issue
      from merlin.activity_directive ad
             join altered_names a on a.name = ad.name
      where ad.plan_id = _plan_id
    )
```
However, we should've been comparing to the activity directive `type`. Fixed as well in this commit.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanCollaborationTests.java
joshhaug: Good idea, fixed in dabfe71efc862580c31516342c6efadb03e4829d

deployment/postgres-init-db/sql/functions/merlin/merging/get_merge_base.sql
joshhaug: I think there may be value in future discussion on this, but this is probably not in scope or a blocker for this PR

joshhaug:
Ready for a final look from @Mythicaeda!

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
```diff
@@ -0,0 +1,356 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+@SuppressWarnings("SqlSourceToSinkFlow")
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class PlanMigrationTests {
+  private DatabaseTestHelper helper;
+  private MerlinDatabaseTestHelper merlinHelper;
+
+  private Connection connection;
+  int fileId;
```
Mythicaeda: Unused variable

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
```diff
@@ -0,0 +1,356 @@
...
+        resultMap.put("modified_activity_types", res.getString("modified_activity_types"));
+        resultMap.put("impacted_directives", res.getString("impacted_directives"));
+        return resultMap;
+      } else {
+        throw new SQLException("No result returned from check_model_compatibility");
+      }
+    }
+  }
+
+  private Boolean migratePlanToModel(final int planId, final int newModelId) throws SQLException{
```
Mythicaeda: I think you meant `boolean` and not `Boolean`

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
```diff
@@ -0,0 +1,356 @@
...
+          //language=sql
+          """
+          select hasura.migrate_plan_to_model(%d, %d, '%s'::json);
+          """.formatted(planId, newModelId, merlinHelper.admin.session())
+      );
+      return res.next();
+    }
+  }
+
+  private Integer getPlanModel(final int planId) throws SQLException {
```
Mythicaeda:
```suggestion
  private int getPlanModel(final int planId) throws SQLException {
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+
+  /**
+   * Check that migration fails if there exist open merge requests on _plan_id.
+   */
+  @Test
+  void checkFailsForOpenMergeRequests() throws SQLException {
+    final int branchId = duplicatePlan(planId, "MultiTags Test");
+    final int mergeRQ = createMergeRequest(planId, branchId);
```
Mythicaeda: The value of `mergeRQ` is unused and the assignment can be removed.

```suggestion
   createMergeRequest(planId, branchId);
```

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
```diff
@@ -0,0 +1,356 @@
...
+    );
+  }
+
+  /**
+   * Check that migration creates a snapshot.
+   */
+  @Test
+  void verifySnapshotCreated() throws SQLException {
+
+    try {
```
Mythicaeda: `try-catch` block is unneeded in this test, `verifyMigrationWorks`, `verifyCompatibilityCheckWorksForIdenticalModels`, and `verifyCompatibilityCheckWorksForDissimilarModels` -- the code just rethrows the only caught exception. Please remove these.

Mythicaeda (approved):
Approving as the only new comments are minor code hygiene ones.

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
joshhaug: Fixed in 83096223a92a5f4f051e3a2804195e018703f48b

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
joshhaug: Done in 83096223a92a5f4f051e3a2804195e018703f48b

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
joshhaug: Done in 83096223a92a5f4f051e3a2804195e018703f48b

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
joshhaug: Yep, done in 83096223a92a5f4f051e3a2804195e018703f48b

db-tests/src/test/java/gov/nasa/jpl/aerie/database/PlanMigrationTests.java
joshhaug: Removed in 83096223a92a5f4f051e3a2804195e018703f48b

joshhaug:
Addressed all remaining comments 👍

PR #1666, opened 2025-04-02
Disable on-update triggers during Procedural Constraint's Migration

Mythicaeda:
* **Review:** By commit  

## Description

The update statement in the migration will update every row in the table, so the triggers do not need to fire. In fact, these triggers firing may cause issues when the migration is run against a database with priority gaps.

## Verification

Verified by testing against aerie-dev's database, which did encounter the bug.


dandelany (approved):
👍 reviewed with @Mythicaeda on a Huddle

PR #1667, opened 2025-04-03
Disable on-update triggers much more aggressively during Proc Constraints Migration

Mythicaeda:
## Description

Within a single transaction, just disabling the trigger will still cause it to gather up events to fire on, causing issues when it is reenabled in the same transaction. Rather than include a `commit` in a migration (which would break the fact that our migrations are atomic), I decided to just drop and recreate the trigger rather than disable and reenable it. The only difference in effect is that a dropped trigger cannot gather events like a disabled trigger.

## Verification

I wrapped the up migration in a transaction (with a `rollback` at the end) when I tested this time, to mirror how the migration is wrapped.

### DIFF
--- a/deployment/hasura/migrations/Aerie/16_procedural_constraints/up.sql
+++ b/deployment/hasura/migrations/Aerie/16_procedural_constraints/up.sql
@@ -129,9 +129,8 @@ drop trigger update_scheduling_model_specification_goal on scheduler.scheduling_
 with priorities as (
   select
     goal_invocation_id,
-    row_number() over (partition by model_id) as new_prio
+    row_number() over (partition by model_id order by priority) as new_prio
   from scheduler.scheduling_model_specification_goals smg
-  order by priority
 )
 update scheduler.scheduling_model_specification_goals smg
 set priority = p.new_prio - 1 -- -1, as priority starts at 0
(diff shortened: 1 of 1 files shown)


mattdailis:
Seems reasonable to me. While you're at it, there's one more bug I found while doing some manual testing. For window functions, the order by clause goes inside the parentheses alongside `partition by`. As written, I think this migration shuffles the user's goals into an unspecified order (in the screenshot it happens to be reverse order, but I don't think that will always be the case)

Screenshot 2025-04-03 at 8 05 49 AM


```diff
with priorities as (
  select
    goal_invocation_id,
-   row_number() over (partition by model_id) as new_prio
+   row_number() over (partition by model_id order by priority) as new_prio
  from scheduler.scheduling_model_specification_goals smg
- order by priority
)
```

Mythicaeda:
Aw geez, that didn't come up while I was testing the select. I'll push that fix.

mattdailis:
Thanks for the quick fix. Here are a couple more places where order by is outside the window function - they don't matter as much because constraints never had a defined order to begin with, but since they do have order by clauses, I want to make sure you've given them a glance and think they're fine.

https://github.com/NASA-AMMOS/aerie/blob/a0eceba70304d35cf9ea3d71423a6802b60126c9/deployment/hasura/migrations/Aerie/16_procedural_constraints/up.sql#L235-L241

https://github.com/NASA-AMMOS/aerie/blob/a0eceba70304d35cf9ea3d71423a6802b60126c9/deployment/hasura/migrations/Aerie/16_procedural_constraints/up.sql#L390-L396

Mythicaeda:
> I want to make sure you've given them a glance and think they're fine.

Yea, I'm not worried about it, as, like you said, constraints were unordered prior to this, so users wouldn't have that expectation like they would for scheduling, where we should avoid reordering their goals.

dandelany (approved):
👍 thanks @Mythicaeda - tested with @mattdailis after the latest commit was added and all seems to work now - will merge & test dev deployment

PR #1672, opened 2025-04-16
Added support for saving command dictionaries to the file store

cohansen:
* **Tickets addressed:** NA
* **Review:** By commit

## Description
This adds support for saving command dictionary files when they're uploaded through the UI. Right now we're only saving the parsed json and the typescript representation of a dictionary but sometimes we need the actual dictionary to pass to an action.

The PR looks like more changes then it really is due to prettier going crazy when I touched a few files.

## Verification
Manual testing

## Documentation
NA

## Future work
Right now we don't include the filetype of the command dictionary in the hasura action so this will need to change to support non xml command dictionaries in the future.

### DIFF
--- a/sequencing-server/src/app.ts
+++ b/sequencing-server/src/app.ts
@@ -209,9 +209,19 @@ app.post('/put-dictionary', async (req, res, next) => {
     let parsedDictionary: CommandDictionary | ParameterDictionary | ChannelDictionary | undefined;
     let db_table_name = '';
     let db_value: any[] = [];
+    let columns = 'dictionary_path, mission, version, parsed_json';
+    let conflict = 'dictionary_path = $1, parsed_json = $4';
+    let values = '$1, $2, $3, $4';
+    let returning = 'id, dictionary_path, mission, version, parsed_json, created_at';
+
     if (dictionaryType == DictionaryType.COMMAND && parsedDictionaries.commandDictionary) {
       db_table_name = 'command_dictionary';
       parsedDictionary = parsedDictionaries.commandDictionary as CommandDictionary;
+      // Only command dictionaries have a command_dictionary_file_path so add the necessary column, value, set, and return statement.
+      columns = `${columns}, command_dictionary_file_path`;
+      conflict = `${conflict}, command_dictionary_file_path = $5`;
+      values = `${values}, $5`;
+      returning = `${returning}, command_dictionary_file_path`;
     } else if (dictionaryType == DictionaryType.CHANNEL && parsedDictionaries.channelDictionary) {
       db_table_name = 'channel_dictionary';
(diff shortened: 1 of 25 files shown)


sequencing-server/src/app.ts
```diff
@@ -184,58 +183,76 @@ app.get('/health', (_: Request, res: Response) => {
...
       continue;
     }
 
     const dictionaryPath = await processDictionary(parsedDictionary, dictionaryType as DictionaryType);
+    let commandDictionaryFilePath = undefined;
+
+    if (persistDictionaryToFilesystem) {
+      // TODO: We'll eventually need to be able to handle dynamic filetypes, right now those aren't passed through in the Hasura action.
+      commandDictionaryFilePath = await writeFile(
+        `command.${parsedDictionary.header.version}.xml`,
```
dandelany: Two questions here:
* this is inside a loop which loops over all three dictionary types, should this `if` have a `&& dictionaryType == DictionaryType.COMMAND` if it's meant to be the command dictionary?
* It makes me a bit uncomfortable to be relying directly on the values from the dictionary headers to generate the path, does it make sense to have fallbacks/checks here or does the dictionary parser give us strong enough guarantees?

dandelany: Per discussion with Cody - I'm OK with assuming these will have values (as guaranteed by another part of the system) but we should make sure they're sanitized to be used as filenames (ie. don't have slashes, etc)

sequencing-server/src/utils/file.ts
```diff
@@ -0,0 +1,14 @@
+import fs from 'fs';
+import { getEnv } from '../env.js';
+
+export async function writeFile(fileName: string, folderName: string, data: string): Promise<string> {
+  const folder = `${getEnv().STORAGE}/${folderName}/`;
```
dandelany: minor: node-ish way to do this is `path.join(a, b)` instead of `${a}/${b}`

dandelany: This could be where we do the sanitizing of filename (make sure user doesn't have slashes in filename)

dandelany (changes requested):
reviewed with @cohansen on slack - comments below, 👍 otherwise

cohansen:
Tests are failing until https://github.com/NASA-AMMOS/aerie-actions/pull/2 is pulled and published to NPM.

sequencing-server/src/app.ts
dandelany: Update: discussed with Cody and we decided to make the backend allow uploading *any* kinds of dictionaries, not just command dictionaries, so this code needs to be updated again to allow any type

dandelany (approved):
👍 @cohansen updated the code to allow uploading any type of dictionary file, looks good to me!

PR #1675, opened 2025-04-22
Action Cancellation

joshhaug:
* **Ticket:** No ticket created for this item.
* **Review:** By file  

## Description
This adds the ability for a user to cancel an action run that is currently in progress. This is designed to be used in conjunction with the [aerie-ui PR for action cancellation support](https://github.com/NASA-AMMOS/aerie-ui/pull/1673).

### Summary of Changes
* Added a `canceled` key to the `action_run` schema, this is set to `true` _once the user requests the action to be cancelled_
* Added a trigger function `notify_action_run_cancel_requested` that will alert listeners when an `action_run`'s `canceled` key is set to `true`
* Added a listener for this trigger in the `actions-server`
* Updated `workerPool` to:
  * Keep track of `AbortController`s and `MessagePort`s for action runs
  * Send/receive messages to/from the worker thread prior to aborting (see diagram for more info)
* Updated `worker` to:
  * Update the `action_run` status to incomplete when starting
  * Send/receive messages to/from the main thread to coordinate db cleanup
  * Did some general cleanup

Here's a rough diagram showing how the cancel signal flows across the different elements of the stack:
Screenshot 2025-04-22 at 11 56 50 AM

## Verification
Verified by testing with several instances of the action from [the `aerie-action-template` repo](https://github.com/NASA-AMMOS/aerie-action-template). Tested in conjunction with [UI changes](https://github.com/NASA-AMMOS/aerie-ui/pull/1673). 

## Future work
* When the action server starts, we should attempt to rebuild the in-memory queue of tasks. This could be done by looking for `action_run`s in the DB that have `canceled==false && status== pending`.  Ppresumably, any with `status==incomplete` were being worked on at the time of crash, and could be marked as `failed` at this time. 
* There is an argument to be made for a reserved `cleanup()` or `onCancel()` function that a user can implement in their `action.js` implementation, which could be called when the worker receives a cancellation request to allow the action author to perform any cleanup they deem necessary.

### DIFF
--- a/action-server/src/listeners/dbListeners.ts
+++ b/action-server/src/listeners/dbListeners.ts
@@ -9,6 +9,7 @@ import type { ActionDefinitionInsertedPayload, ActionResponse, ActionRunInserted
 import { extractSchemas } from "../utils/codeRunner";
 import { createLogger, format, transports } from "winston";
 import logger from "../utils/logger";
+import {ActionRunCancellationRequestPayload} from "../type/types";
 
 let listenClient: PoolClient | undefined;
 
@@ -47,6 +48,10 @@ async function refreshActionDefinitionSchema(payload: ActionDefinitionInsertedPa
   }
 }
 
+async function cancelAction(payload: ActionRunCancellationRequestPayload) {
+  ActionWorkerPool.cancelTask(payload.action_run_id);
+}
+
 async function runAction(payload: ActionRunInsertedPayload) {
   const actionRunId = payload.action_run_id;
   const actionFilePath = payload.action_file_path;
@@ -63,28 +68,40 @@ async function runAction(payload: ActionRunInsertedPayload) {
   const pool = ActionsDbManager.getDb();
   logger.info(`Submitting task to worker pool for action run ${actionRunId}`);
   const start = performance.now();
-
   let run, taskError;
   try {
     run = (await ActionWorkerPool.submitTask({
-      actionJS,
-      parameters,
-      settings,
-      workspaceId,
-    })) satisfies ActionResponse;
+          actionJS: actionJS,
+          action_run_id: actionRunId,
+          parameters: parameters,
+          settings: settings,
+          workspaceId: workspaceId,
+          abort_controller: new AbortController()
+        }
+    )) satisfies ActionResponse;
   } catch (error: any) {
-    logger.error("Error running task:", error);
-    taskError = { message: error.message, stack: error.stack };
+    if (error?.name === "AbortError") {
+      // todo: verify that this will only happen on deliberate cancel
+      logger.info(`Action run ${actionRunId} has been canceled`);
+    } else {
+      logger.error("Error running task:", error);
+      taskError = {message: error.message, stack: error.stack};
+      logger.error(JSON.stringify(taskError));
+      logger.error(`Run is ${JSON.stringify(run || "")}`);
+    }
   }
 
   const duration = Math.round(performance.now() - start);
   const status = taskError || run?.errors ? "failed" : "success";
   logger.info(`Finished run ${actionRunId} in ${duration / 1000}s - ${status}`);
+  const errorValue = JSON.stringify(taskError || run?.errors || {});
+
 
   const logStr = run ? run.console.join("\n") : "";
 
   // update action_run row in DB with status/results/errors/logs
   try {
+
     const res = await pool.query(
       `
       UPDATE actions.action_run
@@ -99,8 +116,8 @@ async function runAction(payload: ActionRunInsertedPayload) {
     `,
       [
         status,
-        JSON.stringify(taskError || run?.errors),
-        run ? JSON.stringify(run.results) : undefined,
+        errorValue,
+        run && run.results ? JSON.stringify(run.results) : JSON.stringify("{}"),
         logStr,
         duration,
         payload.action_run_id,
@@ -124,6 +141,8 @@ export async function setupListeners() {
   listenClient.query("LISTEN action_definition_inserted");
   // these occur when a user inserts a row in the `action_run` table, signifying a run request
   listenClient.query("LISTEN action_run_inserted");
+  // these occur when a user sets the `canceled` of an `action_run` to true, signifying a cancellation request
+  listenClient.query("LISTEN action_run_cancel_requested");
 
   listenClient.on("notification", async (msg) => {
     console.info(`PG notify event: ${JSON.stringify(msg, null, 2)}`);
@@ -137,6 +156,8 @@ export async function setupListeners() {
       await refreshActionDefinitionSchema(payload);
     } else if (msg.channel === "action_run_inserted") {
       await runAction(payload);
+    } else if (msg.channel === "action_run_cancel_requested") {
+      await cancelAction(payload);
     }
   });
   logger.info("Initialized PG event listeners");
--- a/action-server/src/listeners/dbListeners.ts
+++ b/action-server/src/listeners/dbListeners.ts
@@ -9,7 +9,7 @@ import type { ActionDefinitionInsertedPayload, ActionResponse, ActionRunInserted
 import { extractSchemas } from "../utils/codeRunner";
 import { createLogger, format, transports } from "winston";
 import logger from "../utils/logger";
-import {ActionRunCancellationRequestPayload} from "../type/types";
+import { ActionRunCancellationRequestPayload } from "../type/types";
 
 let listenClient: PoolClient | undefined;
 
@@ -71,22 +71,21 @@ async function runAction(payload: ActionRunInsertedPayload) {
   let run, taskError;
   try {
     run = (await ActionWorkerPool.submitTask({
-          actionJS: actionJS,
-          action_run_id: actionRunId,
-          parameters: parameters,
-          settings: settings,
-          workspaceId: workspaceId,
-          abort_controller: new AbortController(),
-          message_port: null,
-        }
-    )) satisfies ActionResponse;
+      actionJS: actionJS,
+      action_run_id: actionRunId,
+      parameters: parameters,
+      settings: settings,
+      workspaceId: workspaceId,
+      abort_controller: new AbortController(),
+      message_port: null,
+    })) satisfies ActionResponse;
   } catch (error: any) {
     if (error?.name === "AbortError") {
       // todo: verify that this will only happen on deliberate cancel
       logger.info(`Action run ${actionRunId} has been canceled`);
     } else {
       logger.error("Error running task:", error);
-      taskError = {message: error.message, stack: error.stack};
+      taskError = { message: error.message, stack: error.stack };
       logger.error(JSON.stringify(taskError));
       logger.error(`Run is ${JSON.stringify(run || "")}`);
     }
@@ -97,12 +96,10 @@ async function runAction(payload: ActionRunInsertedPayload) {
   logger.info(`Finished run ${actionRunId} in ${duration / 1000}s - ${status}`);
   const errorValue = JSON.stringify(taskError || run?.errors || {});
 
-
   const logStr = run ? run.console.join("\n") : "";
 
   // update action_run row in DB with status/results/errors/logs
   try {
-
     const res = await pool.query(
       `
       UPDATE actions.action_run
--- a/action-server/src/listeners/dbListeners.ts
+++ b/action-server/src/listeners/dbListeners.ts
@@ -73,18 +73,16 @@ async function runAction(payload: ActionRunInsertedPayload) {
     run = (await ActionWorkerPool.submitTask({
       actionJS: actionJS,
       action_run_id: actionRunId,
+      message_port: null,
       parameters: parameters,
       settings: settings,
       workspaceId: workspaceId,
-      abort_controller: new AbortController(),
-      message_port: null,
     })) satisfies ActionResponse;
   } catch (error: any) {
     if (error?.name === "AbortError") {
       logger.info(`Action run ${actionRunId} has been canceled`);
     } else {
       logger.error("Error running task:", error);
-      logger.error(`Run is ${JSON.stringify(run || "")}`);
     }
     taskError = { message: error.message, stack: error.stack };
     logger.error(JSON.stringify(taskError));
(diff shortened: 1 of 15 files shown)


joshhaug:
@Mythicaeda if you have some free time, I'd appreciate your feedback on the DB changes here. Nothing major

action-server/src/listeners/dbListeners.ts
```diff
@@ -63,23 +68,33 @@ async function runAction(payload: ActionRunInsertedPayload) {
...
+      workspaceId: workspaceId,
+      abort_controller: new AbortController(),
+      message_port: null,
     })) satisfies ActionResponse;
   } catch (error: any) {
-    logger.error("Error running task:", error);
-    taskError = { message: error.message, stack: error.stack };
+    if (error?.name === "AbortError") {
+      // todo: verify that this will only happen on deliberate cancel
+      logger.info(`Action run ${actionRunId} has been canceled`);
```
dandelany, line 83: Per discussion:
* let's treat canceled as `failed` and set status to `failed` after cancellation
* and set taskError to the `AbortError`

action-server/src/threads/workerPool.ts
```diff
@@ -21,11 +24,58 @@ export class ActionWorkerPool {
...
+    this.messagePortsForActionRun.set(task.action_run_id, port1);
+
+    console.log("Submitted new task with ID " + task.action_run_id);
+
     try {
-      return await ActionWorkerPool.piscina.run(task, { name: "runAction" });
+      return await ActionWorkerPool.piscina.run(task, {
+        name: "runAction",
+        signal: task.abort_controller.signal,
+        transferList: [port2],
```
dandelany, line 47: Dig into this a little more & confirm we understand correctly why this is needed, & leave a one-line comment

action-server/src/threads/workerPool.ts
```diff
@@ -21,11 +24,58 @@ export class ActionWorkerPool {
...
     }
   }
+
+  static cancelTask(action_run_id: string) {
+    // kill the task and delete from the abortControllers data structure
+    logger.info(`Attempting to cancel task ${action_run_id}`);
+    if (this.abortControllerForActionRun.has(action_run_id)) {
+      const port = this.messagePortsForActionRun.get(action_run_id);
+      if (port) {
+        port.on("message", async (msg) => {
```
dandelany: minor: look into this & see if there is a common pattern for unbinding this listener

action-server/src/threads/worker.ts
```diff
@@ -2,54 +2,116 @@ import { threadId } from "worker_threads";
...
-  });
+  // Set up the message listener
+  if (task.message_port) {
+    task.message_port.on("message", async (msg) => {
+      if (msg.type === "abort") {
+        logger.info(`[${threadId}] Received abort message`);
+        try {
+          await releaseDbPoolAndClient();
+          logger.info(`[${threadId}] Async cleanup complete`);
+          task.message_port?.postMessage({ type: "cleanup_complete" });
```
dandelany: Minor: look into whether the worker can kill itself rather than waiting for `abort()` from the workerPool, to avoid any overlap time where user code could still be running, after DB cleanup but before `abort`

action-server/src/threads/worker.ts
```diff
@@ -2,54 +2,116 @@ import { threadId } from "worker_threads";
...
-    logger.info(`[${threadId}] released DB connection`);
     return jsRun;
   } catch (e) {
     logger.info(`[${threadId}] error while executing`);
-    if (client) await client.release();
-    await pool.end();
-    logger.info(`[${threadId}] released DB connection`);
     throw e;
+  } finally {
+    await releaseDbPoolAndClient();
```
dandelany, line 123: This is correct logically, but IIRC `finally` has some nasty caveats in JS & we should probably avoid it here

dandelany (changes requested):
Reviewed with @joshhaug - looking good except notes below

action-server/src/threads/worker.ts
joshhaug: Thanks, removed in c424db7e1497f7660ff972bf348f97b6df2e6d91

action-server/src/listeners/dbListeners.ts
joshhaug: Agreed, done in 424a2eccc3bc3e88925778f598e59043c23b3641

deployment/hasura/migrations/Aerie/18_action_cancellation/up.sql
```diff
@@ -0,0 +1,32 @@
...
+end$$;
+
+create trigger notify_action_run_cancel_requested
+  after update on actions.action_run
+  for each row
+  when (
+    NEW.canceled = true
+      and OLD.canceled is distinct from NEW.canceled
+    )
+execute function actions.notify_action_run_cancel_requested();
```
Mythicaeda: The condition for sim and scheduling is the following:

```sql
create trigger notify_simulation_workers_cancel
after update of canceled on merlin.simulation_dataset
for each row
when ((old.status != 'success' or old.status != 'failed') and new.canceled)
execute function merlin.notify_simulation_workers_cancel();
```

It allows for the user to send multiple cancel requests, but more importantly, it abstains from sending a cancel request if the job is already complete. I recommend adding that "check job incomplete" part to your trigger.

deployment/hasura/migrations/Aerie/18_action_cancellation/up.sql
```diff
@@ -0,0 +1,32 @@
...
+  security definer
+  language plpgsql as $$
+begin
+  -- Notify if canceled is set to true
+  if NEW.canceled then
+    perform pg_notify('action_run_cancel_requested', json_build_object(
+        'action_run_id', NEW.id,
+        'canceled', NEW.canceled
+    )::text);
+  end if;
```
Mythicaeda: Nonblocking, minor: you only need to send `action_run_id`. The request channel already communicates that `canceled` is `true`. 

Also nonblocking: You also don't need the `if` wrapper since this code is only called if the `if` condition is true. That's the point of using a `when` condition on a `trigger`-- it performs these wrapping pre-checks ahead of function execution.

deployment/hasura/migrations/Aerie/18_action_cancellation/up.sql
```diff
@@ -0,0 +1,32 @@
...
+
+  return null;
+end$$;
+
+create trigger notify_action_run_cancel_requested
+  after update on actions.action_run
+  for each row
+  when (
+    NEW.canceled = true
```
Mythicaeda: Nonblocking: SQL supports standard `if(boolean)`, you don't need to do `if(boolean = true)` :-P

action-server/src/threads/worker.ts
joshhaug: I was running into some trouble attempting to pass the `AbortController` to the worker to be called from within `worker.ts`. The hassle was, surprise surprise, serialization issues. But there's a more elegant solution here, which is just to have the worker throw an `AbortError` -- this kills the worker immediately as far as I can tell. This means we don't even need an `AbortController`, as aborts are triggered via messages passed from the main thread to the worker thread. See relevant commit 6aebca3a0a03dd440c5390ba866827f5ea33db3f.

action-server/src/threads/workerPool.ts
joshhaug: Added in 6aebca3a0a03dd440c5390ba866827f5ea33db3f. Good idea

deployment/hasura/migrations/Aerie/18_action_cancellation/up.sql
joshhaug: Thanks, updated in 1b08f6ad6e84150e753eb9ea90b6997d41d10114

deployment/hasura/migrations/Aerie/18_action_cancellation/up.sql
joshhaug: Good idea, updated in 1b08f6ad6e84150e753eb9ea90b6997d41d10114

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -76,10 +79,29 @@ begin
     select pg_notify('action_run_inserted', json_strip_nulls(row_to_json(payload))::text)
     from payload
   );
-  return null;
```
Mythicaeda: This line looks like it was erroneously removed.

joshhaug: Bruh moment, fixed in 8f85bd676682f650e3844087734b5e237f3b5c21

deployment/postgres-init-db/sql/tables/actions/action_run.sql
```diff
@@ -76,10 +79,29 @@ begin
...
+
+create function actions.notify_action_run_cancel_requested()
+  returns trigger
+  security definer
+  language plpgsql as $$
+begin
+  perform pg_notify('action_run_cancel_requested', json_build_object(
+      'action_run_id', NEW.id
+  )::text);
+end$$;
```
Mythicaeda: Missing `return null;` at the last line. The function says it `returns trigger`, you need to return _something_. (`null` is picked because this is a function for an _after_ trigger, though you could reasonably return `new`.

deployment/postgres-init-db/sql/tables/actions/action_run.sql
joshhaug: Fixed in 47258e45131e463a3125e2e4f8419f807d4fdab0, thanks

action-server/src/threads/workerPool.ts
joshhaug: Left a small comment in 5c797a57f64fc962753d26c39e712ec946098c3b explaining what the transferList does for future reference

action-server/src/threads/worker.ts
```diff
@@ -2,54 +2,134 @@ import { threadId } from "worker_threads";
...
+        if (msg.type === "abort") {
+          logger.info(`[${threadId}] Received abort message`);
+          try {
+            await releaseDbPoolAndClient();
+            logger.info(`[${threadId}] Async cleanup complete`);
+            task.message_port?.close();
+            logger.info(`[${threadId}] Message port closed. Aborting now...`);
+            // This ensures node has flushed stdout buffer before continuing:
+            await new Promise((r) => process.stdout.write("", r));
+            ActionWorkerPool.removeFromMap(task.action_run_id);
```
dandelany: This was formerly in the `workerPool` and has been moved here - was that intentional? I don't think the worker has access to the same instance of the `Map` that the parent thread has

dandelany (changes requested):
One question from looking at this again with @mattdailis

dandelany:
FYI @mattdailis - I chatted about this with @joshhaug today and the most recent approach for aborting "unraveled" a bit once we realized that the new approach doesn't really work for aborting *future, queued tasks*, only actively running ones. So the updated approach (in progress by Josh) is something like:
* check if port is open (if possible), or send a "started" message on message port when worker starts running
* keep track of "running tasks" in the worker pool using ^this
* remove running tasks on 'workerDestroy' (or other relevant message)
* have a conditional when cancelling:
   - if the task is running, send a message & wait for it to cleanup before aborting
   - if not running, call AbortController to abort immediately (remove from queue)

joshhaug:
Updated to address @dandelany and @mattdailis 's point from above regarding accessing `ActionWorkerPool.removeFromMap` from inside the worker threads. I've rebased and updated the following:
* Added support for killing queued actions that have not started. This somehow fell by the wayside in between rounds of updates for this PR. This requires maintaining a `Set<action_run_id>` (I'm calling it `runningActions`) in the main thread in order to determine whether a given action run is running or just queued. 
* Updated the migration index # to 19
* Generally, a partial revert of fc6821a85d5c5b383b0f1a6a11f3ac45c9434815 -- this approach allowed the worker to kill itself but did not allow piscina to kill queued tasks, only running tasks. I've modified the diagram to show the new flow control:

Screenshot 2025-05-13 at 9 51 14 PM

Mythicaeda (commented):
Approving the SQL Code! 👍

dandelany (approved):
Reviewed & tested with Josh, tests passing, good to go! :rocket:

dandelany (changes requested):
Alas, we discovered a last minute backend(?) bug while testing, I'm gonna un-approve this until we figure it out. To reproduce:

1. Load up a workspace with the `basic-action` example
2. Start **three** different action runs, each with a long `sleepMs` time so they keep running
3. Cancel the **second** (queued, not yet started) action
4. Cancel the **first** (currently running) action
5. The third action should start running, but instead stays queued
6. If you start another action run, it will run and also run the queued action, ie. it eventually unbreaks itself.

dandelany:
A few notes here:
* There is still a bug here related to cancelling queued actions, see this comment: https://github.com/NASA-AMMOS/aerie/pull/1675#pullrequestreview-2865893981
  * we haven't been able to resolve this yet & suspect it's a bug in the underlying piscina library, but per discussion with leads and devs, we don't think it's a serious enough bug to block this merge - especially since it resolves itself if you run/queue more actions - we will tackle it after this merges
* Otherwise, this is in the same boat as https://github.com/NASA-AMMOS/aerie/pull/1665#issuecomment-2960844180 - all checks passing & basically ready to go, but will need one final update to migrations after #1689 - then I will approve

mattdailis:
>* There is still a bug here related to cancelling queued actions, see this comment: [Action Cancellation #1675 (review)](https://github.com/NASA-AMMOS/aerie/pull/1675#pullrequestreview-2865893981)   
>  * we haven't been able to resolve this yet & suspect it's a bug in the underlying piscina library, but per discussion with leads and devs, we don't think it's a serious enough bug to block this merge - especially since it resolves itself if you run/queue more actions - we will tackle it after this merges

It seems like the suspicious-looking code in piscina was added in the latest release (v5.0.0). It may be worth running a test with the previous version, `v4.9.2`, and see if cancellation works better

mattdailis (approved):

dandelany (approved):
Tested with Matt on piscina v4.9.2 today with no issues - ready to merge! Thanks all.

PR #1688, opened 2025-06-03
Removed environment variables from the global being passed into the a…

cohansen:
…ction vm, changed it to only allow permitted env vars

* **Tickets addressed:** NA
* **Review:** By commit

## Description
Updated our global copy function inside action server to remove all of the passed environment variables, and then re-add only the permitted environment variables that are prefixed with `PUBLIC_ACTION_`.

## Verification
Manual tests.

## Documentation
TBD, we need to add documentation surrounded our permitted environment variable prefix.

## Future work
We're also going to have non-persisted secrets in the future that will be entangled with some of this work.


dandelany (approved):
Tested with an action that checked env vars before & after this change & confirmed all works correctly.  Thanks!

PR #1695, opened 2025-06-16
Remove deprecated version tag from docker-compose.yml

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

When I run docker compose commands, I get the following warning message:

> WARN[0000] /Users/dailis/projects/aerie/worktrees/feature/external-source-event-attributes/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

This PR heeds the warning and removes the version tag from all of the docker-compose.yml files in this repository.

Docker compose docs: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

## Verification

I have not validated these changes, but my hope is that the CI checks will be sufficient to prove that at least the e2e-test docker compose file is well-formed.

## Documentation

I'm not aware of any docs that talk about docker compose versions - I'm open to suggestions for whether we should document that, and if so, where to put that documentation.

## Future work

Possible future documentation work.

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.7"
 services:
   aerie_action:
     container_name: aerie_action
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.7"
 services:
   aerie_action:
     build:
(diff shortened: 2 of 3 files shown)


dandelany (approved):
👍 will be nice not to see this warning anymore!

PR #1703, opened 2025-06-26
Update to sequence template time fixes

cartermak:
- Cleans up/unifies logic for parsing and serializing times in templates
- Makes the behavior consistent, where previously the behavior of the `add-time` and `subtract-time` helpers was different than the `format-as-date` helper
- Adds/improves some tests
- Ran the formatter (oops, my IDE did it)

### DIFF
--- a/sequencing-server/src/lib/mustache/util/time.ts
+++ b/sequencing-server/src/lib/mustache/util/time.ts
@@ -2,74 +2,57 @@ import { ParsedDoyString, ParsedDurationString, ParsedYmdString, TimeTypes } fro
 import { Temporal } from '@js-temporal/polyfill';
 import { SequencingLanguage } from '../enums/language.js';
 
+/**
+ * Time Helpers
+ */
+export function addTime(startTime: string, duration: string, language: SequencingLanguage): string {
+  return serializeInstant(parseInstant(startTime, language).add(parseDuration(duration)), language);
+}
 
-/////////////// SEQUENCING-SERVER-SPECIFIC HELPERS ///////////////
-export function addTime(startTime: string, duration: string, environment: { language: SequencingLanguage }): string {
-  let date: Temporal.Instant
-  if (environment.language === SequencingLanguage.STOL) {
-    date = STOLToInstant(startTime)
-  }
-  else if (environment.language === SequencingLanguage.TEXT) {
-    date = Temporal.Instant.from(TextToISO8601(startTime))
-  }
-  else {
-    date = SeqNToInstant(startTime)
-  }
+export function subtractTime(startTime: string, duration: string, language: SequencingLanguage): string {
+  return serializeInstant(parseInstant(startTime, language).subtract(parseDuration(duration)), language);
+}
 
-  let dur: Temporal.Duration;
-  if (duration.includes(":")) {
-    dur = Temporal.Duration.from(AERIEDurationToISO8601(duration))
-  }
-  else {
-    dur = Temporal.Duration.from(duration)
-  }
-  date = date.add(dur)
+export function formatTime(time: string, language: SequencingLanguage): string {
+  return serializeInstant(parseInstant(time, language), language);
+}
 
-  if (environment.language === SequencingLanguage.STOL) {
-    return InstanttoSTOL(date)
-  }
-  else if (environment.language === SequencingLanguage.TEXT) {
-    return InstantToText(date)
-  }
-  else { 
-    return InstanttoSeqN(date)
+/**
+ * Utilities for parsing and serializing times used in sequence templates
+ */
+
+function parseInstant(instant: string, language: SequencingLanguage): Temporal.Instant {
+  if (language === SequencingLanguage.STOL) {
+    return STOLToInstant(instant);
+  } else if (language === SequencingLanguage.TEXT) {
+    return Temporal.Instant.from(TextToISO8601(instant));
+  } else if (language === SequencingLanguage.SEQN) {
+    return SeqNToInstant(instant);
+  } else {
+    throw new Error('Unknown language: ' + language)
   }
 }
 
-export function subtractTime(startTime: string, duration: string, environment: { language: SequencingLanguage }): string {
-  let date: Temporal.Instant
-  if (environment.language === SequencingLanguage.STOL) {
-    date = STOLToInstant(startTime)
+function serializeInstant(instant: Temporal.Instant, language: SequencingLanguage): string {
+  if (language === SequencingLanguage.STOL) {
+    return InstanttoSTOL(instant);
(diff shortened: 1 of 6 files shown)


sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -2,91 +2,75 @@ import { ParsedDoyString, ParsedDurationString, ParsedYmdString, TimeTypes } fro
...
-    const result = new Date(convertDoyToYmd(date)).toISOString()
-    return result + (result.includes("Z") ? "" : "Z")
+  if ((date.match(/-/g) || []).length === 2) {
+    // YYYY-MM-DD string
+    const result = new Date(date).toISOString();
+    return result + (result.includes('Z') ? '' : 'Z');
+  } else {
+    // doy string
+    const result = new Date(convertDoyToYmd(date)).toISOString();
+    return result + (result.includes('Z') ? '' : 'Z');
```
cartermak, line 73: This is just formatter changes

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -96,100 +80,103 @@ function checkNumDurationComponents(split: string[]): split is [string, string,
...
-  else {
-    return result + (result.includes("Z") ? "" : "Z");
+  if (result.includes('.') && result.split('.').length == 2) {
+    let split = result.split('.');
+    let rest = split[0];
+    let micros = split[1]?.replace('Z', '');
+    micros = micros?.padEnd(6, '0');
+    return rest + '.' + micros + 'Z';
+  } else {
+    return result + (result.includes('Z') ? '' : 'Z');
```
cartermak, line 176: All formatter changes, basically

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -213,18 +200,18 @@ export function convertDoyToYmd(doyString: string, includeMsecs = true): string
...
     } else {
       // doyString is already in ymd format
       //    just in case - correct and "/" in lieu of a T, i.e. 2025-001/time vs 2025-001Ttime
-      return `${doyString.replace("/", "T")}${doyString.charAt(doyString.length - 1) !== "Z" ? "Z" : ""}`;
+      return `${doyString.replace('/', 'T')}${doyString.charAt(doyString.length - 1) !== 'Z' ? 'Z' : ''}`;
     }
   }
 
-  throw Error(`Given date: ${doyString} is an invalid DOY (or YMD) string.`)
+  throw Error(`Given date: ${doyString} is an invalid DOY (or YMD) string.`);
```
cartermak, line 218: Also formatter

dandelany:
A couple non-blocking questions from me:
* Is this technically a breaking change to the to the template helper API, and if so, can you write a short description of the API change for the upgrade guide?
* Are there any docs/specs we can link to for any of these time formats, maybe in a code comment? would be nice to have a reference for determining correct behavior

cartermak:
@dandelany 

> Is this technically a breaking change to the to the template helper API

I'm going to say no, other than slightly changing how the formats behaved to be more consistent.

> Are there any docs/specs we can link to for any of these time formats

Not yet! I'll draft a PR to include that in the Templates docs.

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -96,100 +80,103 @@ function checkNumDurationComponents(split: string[]): split is [string, string,
...
+    let rest = split[0];
+    let micros = split[1]?.replace('Z', '');
+    micros = micros?.padEnd(6, '0');
+    return rest + '.' + micros + 'Z';
+  } else {
+    return result + (result.includes('Z') ? '' : 'Z');
   }
 }
 
-/////////////// AERIE-UI HELPERS ///////////////
```
pranav-super, line 192: I think we want this comment because it indicates where the helpers are sourced from. They are modified slightly here, but if we want to think about extracting them at some point, I think it's worth leaving these comments in.

sequencing-server/src/lib/mustache/util/time.ts
```diff
@@ -96,100 +80,103 @@ function checkNumDurationComponents(split: string[]): split is [string, string,
...
-    let microseconds = parseInt(split2[1]?.padEnd(6, "0") as string)
-    let microsecondsString = `${microseconds}`.padStart(6, "0")
-    return `${isNegative ? "-" : ""}PT${hours > 0 ? `${hours}H` : ""}${minutes > 0 ? `${minutes}M` : ""}${seconds > 0 && microseconds > 0 ? `${seconds}.${microsecondsString}S` : (seconds > 0) ? `${seconds}$` : (microseconds > 0) ? `0.${microsecondsString}S` : ""}`
-  }
-  else {
-    return `${isNegative ? "-" : ""}PT${hours > 0 ? `${hours}H` : ""}${minutes > 0 ? `${minutes}M` : ""}${seconds > 0 ? `${seconds}S` : ""}`
+    let microseconds = parseInt(split2[1]?.padEnd(6, '0') as string);
+    let microsecondsString = `${microseconds}`.padStart(6, '0');
+    return `${isNegative ? '-' : ''}PT${hours > 0 ? `${hours}H` : ''}${minutes > 0 ? `${minutes}M` : ''}${
+      seconds > 0 && microseconds > 0
```
pranav-super, line 100: My only commentary here is that this code formatting looks a little bizarre, but it seems completely solid!

sequencing-server/test/templating/time.spec.ts
```diff
@@ -26,7 +27,7 @@ describe('Time vector functions', () => {
         let inputTime = '2025-001/12:00:01.0002Z'
         let inputDuration = '1000:00:01.01234' // removed Z
 
-        let result = subtractTime(inputTime, inputDuration, { language: 'STOL' as SequencingLanguage })
+        let result = subtractTime(inputTime, inputDuration, SequencingLanguage.STOL)
```
pranav-super, line 30: This is much better!

pranav-super (commented):
Minus restoring a few comments, this looks good to go!

sequencing-server/src/lib/mustache/util/time.ts
cartermak: ooooh that makes sense. I was confused what was meant by "aerie-ui helpers" but I'll add back a comment with that context. There's already separate work to develop a universal Aerie time library going on: https://github.com/NASA-AMMOS/aerie-time-utils

sequencing-server/src/lib/mustache/util/time.ts
cartermak: Yeah it's just what prettier is configured for

PR #1704, opened 2025-06-26
Delay reporting simulation status until after profile segments have been written

### LINKED ISSUE #1683: Incorrect Visualization of Simulation Results at End of Plan
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

3.4.0

### Describe the bug

After running simulation on a moderately complicated model, I sometimes get what appears to be weird simulation results at the very end of the plan (see screenshot below). This seems to occur randomly, but often enough that I can usually reproduce it after 3-5 forced resimulations. Refreshing the UI seems to solve the problem, so the data itself seems to be fine. 

Image

### Reproduction

I see this issue using this model:
https://github.com/NASA-AMMOS/aerie-orbiter-model

Try loading this plan
https://github.com/NASA-AMMOS/aerie-orbiter-model/blob/main/Example_MarsSat_Plan.json

and this view
https://github.com/NASA-AMMOS/aerie-orbiter-model/blob/main/MarsSat_Overview_View.json

and then hitting simulate. If results seem ok, try changing a simulation configuration parameter and try simulating again. 

### Logs

```shell

```

### System Info

```shell
Local deployment, firefox 128.11.0esr
```

### Severity

Moderate

### LINKED ISSUE #1694: Simulation profile is not available when simulation completes
### Checked for duplicates

No - I haven't checked

### Is this a regression?

Yes - This worked in a previous version
Worked in version 3.0.1

### Version

3.4.0

### Describe the bug

Found this while running automated tests against Aerie 3.4.0. We have a few tests that simulate a plan, wait until the simulation completes (by polling `simulation_dataset` for `success == true`), and then fetch the profile for a resource. In Aerie 3.4.0 if the test immediately requests the profile as soon as the simulation finishes it receives an empty resource. But if I add a delay of a few seconds then the resource is fetched successfully. In past releases the resources were available as soon as the sim finished.

### Reproduction

Contact me for a Clipper integration test that can reproduce this.

### Logs

```shell

```

### System Info

```shell
Chrome
```

### Severity

Major

mattdailis:
* **Tickets addressed:** Fixes #1694 Fixes #1683
* **Review:** By commit  

## Description

Follow-up to https://github.com/NASA-AMMOS/aerie/pull/1640

Contract: simulation status is `success` implies that the simulation dataset is fully written and committed.

We currently violate this contract occasionally.

See more details here: https://github.com/NASA-AMMOS/aerie/issues/1694#issuecomment-3001499156 

After discussion with @Mythicaeda and @joshhaug , we decided to go with "Option 1" from the linked comment. This amounts to making the `computeResults` block until the resource streamer has finished.

This PR replaces the use of `Consumer` for resource streaming with a new `AsyncConsumer`, which is meant to reflect that the work done by the `accept` method may continue in the background after `accept` has returned. Its `close` method is intended to block until all previous calls to `accept` have finished their background processing.

`close` needs to be safe to call twice, because we call it both in the "nominal" case, as well as in the "finally" case.

### Report final simulation time
While doing the verification steps below, I noticed that the final simulation time was never reported. I added one commit to report the simulation extent one more time after exiting the main simulation loop.

## Verification

I followed a similar approach to https://github.com/NASA-AMMOS/aerie/pull/1640

I deliberately slowed down the resource streamer to guarantee that it is not done by the time the simulation status is updated.
```diff
queryQueue.submit(() -> {
+  Thread.sleep(10000); // omitting try/catch for brevity
  queryHandler.uploadResourceProfiles(resourceProfiles);
});
```

Having reproduced the problem on develop, I stashed the above change, checked out this branch, and popped the stash, and observed that simulation would spend longer in the "incomplete" status, and would only switch to "success" after the resources had finished being written.

## Documentation

This is a bug fix; no documentation has been invalidated.

## Future work

Hopefully none 🤞

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
@@ -109,6 +109,8 @@ public final class SimulationDriver {
           simulationExtentConsumer.accept(engine.getElapsedTime());
         }
 
+        simulationExtentConsumer.accept(engine.getElapsedTime()); // Report the final simulation time
+
       } catch (SpanException ex) {
         // Swallowing the spanException as the internal `spanId` is not user meaningful info.
         final var topics = missionModel.getTopics();
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
@@ -564,7 +564,7 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
         postSimulationResults(connection, datasetId, results, SimulationStateRecord.success());
         deleteSimulationExtent(connection, datasetId);
         transactionContext.commit();
-        logger.info("%s updated simulation status".formatted(Duration.microseconds(System.nanoTime() / 1000)));
+        logger.info("%s set simulation status to success".formatted(Duration.microseconds(System.nanoTime() / 1000)));
       } catch (final SQLException ex) {
         throw new DatabaseException("Failed to store simulation results", ex);
       } catch (final NoSuchSimulationDatasetException ex) {
@@ -581,6 +581,7 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
         failSimulation(connection, datasetId, reason);
         deleteSimulationExtent(connection, datasetId);
         transactionContext.commit();
+        logger.info("%s set simulation status to failed".formatted(Duration.microseconds(System.nanoTime() / 1000)));
       } catch (final SQLException ex) {
         throw new DatabaseException("Failed to update simulation state to failure", ex);
       } catch (final NoSuchSimulationDatasetException ex) {
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
@@ -564,6 +564,7 @@ public final class PostgresResultsCellRepository implements ResultsCellRepositor
         postSimulationResults(connection, datasetId, results, SimulationStateRecord.success());
         deleteSimulationExtent(connection, datasetId);
         transactionContext.commit();
+        logger.info("%s updated simulation status".formatted(Duration.microseconds(System.nanoTime() / 1000)));
       } catch (final SQLException ex) {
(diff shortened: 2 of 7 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/SimulationDriver.java
```diff
@@ -109,6 +109,8 @@ public static <Model> SimulationResults simulate(
           simulationExtentConsumer.accept(engine.getElapsedTime());
         }
 
+        simulationExtentConsumer.accept(engine.getElapsedTime()); // Report the final simulation time
```
joshhaug, line 112: Good catch

joshhaug (approved):
I've been able to successfully reproduce Matt's verification results with the stash provided on slack. 
Really clean, approving.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
```diff
@@ -564,6 +564,7 @@ public void succeedWith(final SimulationResults results) {
         postSimulationResults(connection, datasetId, results, SimulationStateRecord.success());
         deleteSimulationExtent(connection, datasetId);
         transactionContext.commit();
+        logger.info("%s set simulation status to success".formatted(Duration.microseconds(System.nanoTime() / 1000)));
```
Mythicaeda, line 567: Why were the logging statements added?

Mythicaeda (approved):
LGTM

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresResultsCellRepository.java
mattdailis: We added them while trying to reproduce the issue so that we could get a better idea of timing. @dandelany asked me to leave them in the code as useful information for someone trying to understand the system's behavior

dandelany (approved):
Notes from review/testing with @mattdailis yesterday:
* We found a reliable way to reproduce the issue - apply this patch to the banananation mission model: [simulation-status-data-race-repro.patch](https://github.com/user-attachments/files/21148800/simulation-status-data-race-repro.patch) and rebuild, ie:
```
git apply patchfile.patch
./gradlew :examples:banananation:clean
./gradlew :examples:banananation:assemble
```
* Confirmed via the UI (which uses the `status` flag to decide when to fetch profiles) that on `develop` ^this reliably fails to show complete profiles, running on this branch works correctly.
* We added a couple additional `info` log messages as they may help diagnose any future timing-related simulation issues
* One outstanding question for @Mythicaeda (I will follow up separately) is whether the same issue could affect Stateless Aerie - since in this PR the `close` method for the `Streamer` in `orchestration-utils` is a no-op.

PR #1700, opened 2025-06-25
Publish procedural-utils and merlin-server

JoelCourtney:
* **Tickets addressed:** none
* **Review:** By commit  
* **Merge strategy:** Merge (no squash)  

## Description

This uncomments the publishing block for the procedural-utils package.

PR #1705, opened 2025-06-26
Remove merlin-server as a dependency of orchestration-utils

mattdailis:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

This is a follow-up to #1700

Currently, users cannot use orchestration-utils without pulling in merlin-server. We don't typically publish merlin-server (#1700 was an exception to that). This PR removes merlin-server from the orchestration-utils dependencies.

There were four things in merlin-server that orchestration-utils needed:
- pgTimestampP
- activityArgumentsP
- simulationArgumentsP
- realDynamicsP
- EventGraphFlattener

This PR creates a new `Parsers.java` file under the `gov.nasa.jpl.aerie.merlin.driver.json` and moves the four parsers above from merlin-server to this new file.

This PR moves `EventGraphFlattener` into `gov.nasa.jpl.aerie.merlin.driver.timeline` alongside the definition of `EventGraph` itself.

## Verification

It compiles!

## Documentation

I couldn't find any docs on stateless aerie - if someone knows otherwise, let me know.

## Future work

There are a number of duplicated parsers and parsing functions across merlin-server and scheduler-server. I think we can take a careful look at them and move some of them to this new Parsers.java class. 

There are a few copy/pasted implementations of the `parseJson` function that I think can be unified by adding a static method to JsonParser in parsing-utilities. I took a stab at that, but then decided that it would be better for this PR to be minimal. That first attempt is in the branch called [`dailis/orchestration-utils-dependency-refactor-take-1`](https://github.com/NASA-AMMOS/aerie/compare/dailis/orchestration-utils-dependency-refactor-take-1)

### DIFF
--- a/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/Parsers.java
+++ b/merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/Parsers.java
@@ -1,0 +1,80 @@
+package gov.nasa.jpl.aerie.merlin.driver.json;
+
+import gov.nasa.jpl.aerie.json.JsonParseResult;
+import gov.nasa.jpl.aerie.json.JsonParser;
+import gov.nasa.jpl.aerie.json.SchemaCache;
+import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.types.Timestamp;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.json.JsonValue;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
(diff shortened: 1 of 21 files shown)


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/json/Parsers.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static gov.nasa.jpl.aerie.json.BasicParsers.doubleP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.mapP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.productP;
+import static gov.nasa.jpl.aerie.json.BasicParsers.stringP;
+import static gov.nasa.jpl.aerie.json.Uncurry.tuple;
+import static gov.nasa.jpl.aerie.json.Uncurry.untuple;
+import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
+
+public class Parsers {
```
Mythicaeda, line 29: Would these Parsers make sense in either `type-utils` or `parsing-utils`?

PR #1709, opened 2025-07-10
Generate subsystem tags for subsystem annotations in mission models

### LINKED ISSUE #934: Extract Subsystem tags from Mission Model

tratsko:
* **Tickets addressed:** Fixes #934
* **Review:** By commit 

## Description

There are two new annotations being defined in the Merlin annotation processor. `@Subsystem` on each activity and @WithSubsystem in the package-info. These are optional annotations that take a string as the name of the subsystem. 

The mission model compiles with these annotations stored in a getSubsystem function on the activity and the getSubsystems function on the mission model. 

Once the compiled model is uploaded to the Merlin server it gets the subsystems and then generates tags for each subsystem if they don't already exist. Once all of the tags are generated it associates each activity that was annotated with a subsystem to that tag.

## Verification

It compiles and picks up tags when the mission model is uploaded with annotations to the database. This verification was done manually.

## Documentation

https://nasa-ammos.github.io/aerie-docs/mission-modeling/activity-types/introduction/#activity-type-metadata
This documentation on putting tags in the javadoc on mission models is invalidated by these changes. This documentation on the subsystem tags was invalid prior to this PR as there is no functionality that associates subsystems to activities from the mission model.

## Future work

Generate random colors for tags
Allow for the annotation to have color as a parameter
Allow for description annotation

### DIFF
--- a/deployment/hasura/migrations/Aerie/27_merlin_tags_permissions/up.sql
+++ b/deployment/hasura/migrations/Aerie/27_merlin_tags_permissions/up.sql
@@ -16,4 +16,4 @@ do $$
   end
 $$;
 
-call migrations.mark_migration_applied('27');
+call migrations.mark_migration_applied(27);
--- a/deployment/hasura/migrations/Aerie/27_merlin_tags_permissions/up.sql
+++ b/deployment/hasura/migrations/Aerie/27_merlin_tags_permissions/up.sql
@@ -11,9 +11,9 @@ do $$
     limit 1
     into merlin_user;
 
-    execute format('revoke select, insert on table tags.tags from %I', merlin_user);
-    execute format('revoke usage on schema tags from %I', merlin_user);
+    execute format('grant usage on schema tags to %I', merlin_user);
+    execute format('grant select, insert on table tags.tags to %I', merlin_user);
   end
 $$;
 
-call migrations.mark_migration_rolled_back('25');
+call migrations.mark_migration_applied('27');
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/DirectiveType.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/DirectiveType.java
@@ -37,9 +37,12 @@ public interface DirectiveType<Model, Arguments, Result> {
   //   the output data is emitted.
   OutputType<Result> getOutputType();
 
-  default Optional<String> getSubsystem() {
-    return Optional.empty();
-  }
+  /**
+   * Gets the subsystem of this directive.
+   *
+   * @return The subsystem of this directive.
+   */
+  Optional<String> getSubsystem();
 
   /**
    * Initializes a {@link Task} operating on the given model given a directive instance.
--- a/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/DirectiveType.java
+++ b/merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/DirectiveType.java
@@ -4,6 +4,7 @@ import gov.nasa.jpl.aerie.merlin.protocol.types.InstantiationException;
 import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
 
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * A type of directive that can be issued to a Merlin model.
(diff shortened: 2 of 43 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSubsystemsAction.java
```diff
@@ -0,0 +1,67 @@
...
+    insert into tags.tags (name, owner, color) values (?, 'Mission Model', '#000000')
+    on conflict (name) do nothing;
+""";
+
+  private static final @Language("SQL") String findTagsSQL = """
+    select name, id
+    from tags.tags as t;
+    """;
+  // where t.name in ?; TODO Check if we need this where clause since it would make it more efficient to only grab tags
+  //                       associated with a subsystem instead of all tags
```
Mythicaeda: This statement isn't used in the action (and if anything should be in a separate action)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSubsystemsAction.java
```diff
@@ -0,0 +1,67 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/*package-local*/ final class InsertSubsystemsAction implements AutoCloseable {
+  private static final @Language("SQL") String insertSQL = """
+    insert into tags.tags (name, owner, color) values (?, 'Mission Model', '#000000')
```
Mythicaeda: `color` should not be getting set in this `insert`

mattdailis: Not disagreeing, but it's hard to understand your rationale. Could you elaborate a bit?

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/ModelType.java
```diff
@@ -48,6 +50,10 @@ public interface ModelType<Config, Model> {
    */
   InputType<Config> getConfigurationType();
 
+  default List<String> getSubsystems() {
+    return Collections.emptyList();
+  }
```
Mythicaeda, line 60: @mattdailis I'll defer to you if you think otherwise, but on Aerie we don't do `default` methods for interfaces. Please remove the "default" implementation:

```suggestion
  List<String> getSubsystems()
```

Mythicaeda: Additionally, please add a doc comment to this method

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/DirectiveType.java
```diff
@@ -36,6 +37,10 @@ public interface DirectiveType<Model, Arguments, Result> {
   //   the output data is emitted.
   OutputType<Result> getOutputType();
 
+  default Optional<String> getSubsystem() {
+    return Optional.empty();
+  }
```
Mythicaeda, line 47: Similar story here:
1) Remove `default` implementation
2) Add doc comment

deployment/hasura/migrations/Aerie/27_merlin_tags_permissions/up.sql
```diff
@@ -0,0 +1,19 @@
...
+      and grantee != (select current_user)
+    limit 1
+    into merlin_user;
+
+    execute format('grant usage on schema tags to %I', merlin_user);
+    execute format('grant select, insert on table tags.tags to %I', merlin_user);
+  end
+$$;
+
+call migrations.mark_migration_applied('27');
```
Mythicaeda: In both the `up` and the `down`, these need to now be numbers instead of strings

mattdailis:
@Mythicaeda I've asked Tim to add the default methods back to the interfaces—my understanding is that this enables existing mission models to continue to work without recompilation. Let me know if you have something else in mind that would help preserve this backwards compatibility

Mythicaeda (approved):
Sorry for the delay! LGTM, default methods do remove the `breaking` part of the change.

PR #1711, opened 2025-07-15
Workspaces Server

Mythicaeda:
* **Review:** By commit  

## Description

- Implements a Workspace Server that stores files inside of a Docker Container
- Updates the DB to remove the User Sequences Table
- Adds Workspace Collaborators
- Refactors the migration script to support non-sql between-migration tasks


dandelany (changes requested):
One note from testing with @mattdailis earlier today:

We think the API for *creating a directory* should basically ignore the `overwrite` flag and always behave the same way: create the requested directory (or directories, if a nested path is given) *if* they don't exist, otherwise return (don't fail if they do exist). This makes it consistent with how Java `createDirectories` API works - having to pass `overwrite` to get this behavior makes it a bit ambiguous & implies an existing directory would be overwritten by an empty one (which should not happen)

More notes to come as we review code!

dandelany:
@Mythicaeda FYI I rebased and had to bump the migration numbers since we merged the External Events PR. I *think* I updated all of the uses of magic number "24" (in aerie_db_migration.py) but let me know if I missed anything.

.github/workflows/publish.yml
```diff
@@ -69,6 +69,9 @@ jobs:
           - image: aerie-action
             context: action-server
             file: action-server/Dockerfile
+          - image: aerie-workspaces
```
dandelany: let's make this singular `aerie-workspace` or `aerie_workspace` everywhere (dash for image, _ for container)

deployment/docker-compose.yml
```diff
@@ -159,6 +159,26 @@ services:
       - aerie_file_store:/usr/src/app/sequencing_file_store
     networks:
       - aerie_net
+  aerie_workspace:
+    container_name: aerie_workspaces
```
dandelany: singular here too

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -0,0 +1,454 @@
...
+
+    String fileName() {
+      return filePath.getFileName().toString();
+    }
+  }
+
+  @Override
+  public void apply(final Javalin javalin) {
+    javalin.routes(() -> {
+      //before("/ws/*", authorize); <- dont want to force auth on the health check.
```
dandelany: This is commented out, but without it, we don't auth users (except on the `createWorkspace` route). We think this should be:
```
before("/ws/*", this::authorize);
```
does that seem right @Mythicaeda ?

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -0,0 +1,454 @@
...
+        context.status(400).result("Workspace location may not contain '/'");
+        return;
+      }
+
+      workspaceLocation = Path.of(bodyJson.getString("workspaceLocation"));
+      workspaceName = bodyJson.containsKey("workspaceName") ? bodyJson.getString("workspaceName") : workspaceLocation.toString();
+    } catch (NullPointerException npe) {
+      context.status(400).result("Mandatory body parameter is null. Request body format is:\n" + helpText);
+      return;
+    } catch (JsonException je) {
```
mattdailis: I'm curious - which line of code above is expected to sometimes throw a null pointer exception? Would it be possible to do some kind of null check instead of catching NullPointerException?

In general I'm of the opinion that catching NPE is something to be avoided because it could easily mask a real bug

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/JWTService.java
```diff
@@ -0,0 +1,80 @@
...
+    final var typeString = jwtInfo.getString("type");
+    final var issuer = jwtInfo.containsKey("iss") ? jwtInfo.getString("iss") : null;
+
+    // Expand on this switch statement as we support more Algorithm types.
+    // Currently, the Gateway only supports HMAC key variants
+    final Algorithm algorithm = switch (typeString) {
+      case "HS256" -> Algorithm.HMAC256(key);
+      case "HS384" -> Algorithm.HMAC384(key);
+      case "HS512" -> Algorithm.HMAC512(key);
+      default -> Algorithm.none();
```
dandelany: To be defensive, we should `throw` if the `typeString` is not a known algorithm - as is, a user could make a typo eg `hs_256` in their config and it would "fail open" - ie. it would work but would silently not authenticate tokens. We could (if needed) also have a "none" case here for explicitly using no algorithm

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Oh, it's commented out because there was no implemented `authorize` method when I drafted it. I'll uncomment it, just be aware that it's currently only checking that an unexpired, valid JWT was provided ("valid" meaning the user can access the role they're using and that the signature is correct), as opposed to a detailed permissions check (ie, "viewer" role can't use the `edit` endpoints).

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -0,0 +1,312 @@
...
+    boolean success = true;
+
+    final var contents = directory.listFiles();
+    if(contents == null) {
+      return rm(directory);
+    }
+
+    for(final var f : contents) {
+      if(Files.isSymbolicLink(f.toPath()) || !f.isDirectory()) {
+        success = success && rm(f);
```
dandelany: The doc comment for this implies that it will try to delete as much as possible, not stopping on failures, but it seems like this `success && ...` pattern will short-circuit and stop on failures, is this intended?

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -0,0 +1,312 @@
...
+  private boolean rm(final File file) {
+    try {
+      return file.delete();
+    } catch (SecurityException se) {
+      return false;
+    }
+  }
+
+  @Override
+  public boolean deleteWorkspace(final int workspaceId) throws NoSuchWorkspaceException, SQLException {
```
dandelany: minor: in general it's a bit confusing to have functions which sometimes `throw` on failure and sometimes `return false` depending on how they fail - would be nice to standardize and maybe always use exceptions. I think the `Files` API does a better job at this than the `File` API

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: L111-L112: 
```java
parcelId = bodyJson.getInt("parcelId");
final var workspaceString = bodyJson.getString("workspaceLocation");
```

When you call `json.get('key')` on a key not in the json object, the method throws an NPE. Since the code needed a `try-catch` anyway to catch JSON parsing errors (such as if `parcelId` was set to a `string`), I decided to catch the thrown exception rather than set up `if` statements to "pre-catch" it.

> In general I'm of the opinion that catching NPE is something to be avoided because it could easily mask a real bug

While I agree with this in general, if the code was to _not_ catch the NPE here, all the user would get in the event that their misformatted body caused an NPE would be a generic `500` status, which is not useful information for them.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    // Converting our API to the Files API
+    final var walkDepth = depth == -1 ? Integer.MAX_VALUE : depth + 1;
+    try(final Stream<Path> walkOutput = Files.walk(path, walkDepth)) {
+      final var walkList = new ArrayList<>(walkOutput.toList());
+      walkList.removeFirst(); // remove the initial path
+      return new DirectoryTree(path, walkList, postgresRepository.getExtensionMapping());
+    } catch (IOException io) {
+      return null;
```
dandelany: We should probably throw this Exception up to the caller to get the message - as-is any IOException will be reported as "directory does not exist" but there could be other reasons for the failure

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/DirectoryTree.java
```diff
@@ -0,0 +1,96 @@
...
+    for(final var path : inputList){
+      if(path.toFile().isDirectory()) {
+        this.root.addChild(new DirectoryNode(path));
+      } else {
+        this.root.addChild(new FileNode(path, RenderType.getRenderType(path.getFileName().toString(), extensionMappings)));
+      }
+    }
+  }
+
+  private static class FileNode {
```
dandelany: Minor/later thing, but a lot of this file helper stuff seems like the kind of thing that might be better to pull from a library & reduce how much we need to maintain. Might look into using the Apache `FileUtils` library later & see how much that buys us.

dandelany (changes requested):
@mattdailis and I did some review on this today, notes above. The Java service is pretty well covered, but we still need to review the migration script (tomorrow). I also pushed some additional fixes to this branch since we're time-constrained - @Mythicaeda please take a look and let me know if you have any issues with these:

* c5f3ee1d7efd861901dc0ad20625ffca6ed71af4 ensures that we don't use `Algorithm.none` by accident for JWTs - if there is a good reason to support it we can add it back as an explicit "none" option, but it shouldn't be the default.
* 00c558e8e615e08f8bb33b971faa99c8c7fb3f95 adds and uses a utility for safely resolving subpaths within workspaces. Without this, we are vulnerable to path traversal attacks, eg. a user could access/modify a relative path outside their workspace like `../otherworkspace/someones.file` or even an absolute path like `/etc/passwd` (because `Path.of("/foo").resolve(Path.of("/etc/passwd")) == "/etc/passwd"`). Also added unit tests for this - lmk if you'd prefer this to be in a separate util file.
* 58e68c9dffc0e323c5fdee570ab4004606725c0b adds slightly more path safety by ensuring metadata files are siblings in case one of the file extension values has `..`/other unsafe things.

There might be a few more of these path-safety-type issues remaining, let me know if you see any.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
mattdailis: > Since the code needed a try-catch anyway to catch JSON parsing errors (such as if parcelId was set to a string), I decided to catch the thrown exception rather than set up if statements to "pre-catch" it.

Thanks, that does make sense, it does make the code easier to write. As a reader, here's what makes me nervous about seeing "catch NullPointerException" anywhere in a program:

These are the possible sources of a null pointer exception in this try catch block:
- reader.readObject() (if reader is null)
- bodyJson.getInt("parcelId") (if bodyJson is null or if the key is missing)
- bodyJson.getString("workspaceLocation")  (if bodyJson is null or if the key is missing)
- workspaceString.contains("/") (if workspaceString is null)
- context.status(400) (if context is null)
- bodyJson.getString("workspaceLocation") (if bodyJson is null or if the key is missing)
- bodyJson.containsKey("workspaceName") (if bodyJson is null or if the key is missing)
- bodyJson.getString("workspaceName") (if bodyJson is null or if the key is missing)
- workspaceLocation.toString() (if workspaceLocation is null)

It's a little overwhelming as a reader of this code to try to figure out where it's expected to come from. This argument generalizes to catching any sort of "unchecked" exception, but null pointer exceptions are particularly insidious because they can come from absolutely anywhere.

From a user perspective, all of the above failures get turned into the same message, which I think misses out on an opportunity to give a specific error message. From a debugging perspective, if there is a bug, there is no stack trace to refer to to figure out which line of code caused the error.

> set up if statements to "pre-catch" it.

I took a couple cracks at it, I'm not sure either of these is pretty, but I do think they give better error messages and don't mask potential bugs:
- https://github.com/NASA-AMMOS/aerie/commit/78cc626b98390c0cb772f3eebed8136fc755f5e8
- https://github.com/NASA-AMMOS/aerie/commit/9466db70f857129b6c0caba1046882a5312b6b38

Any preference between those two, or other ideas we should consider?

Mythicaeda: If you revert the changes done to L120 `workspaceName` and return it to its ternary statement (use if provided, else use the location), I prefer the second one.

Though, please, let me push my pending changes first. If I had realized you guys were going to commit to my branch, I would've pushed last night.

Mythicaeda: Actually, correction, you used the wrong method in the second version (you wanted to do `containsKey`, you did `get`, which will throw an NPE if it's absent), I'll just fix it and commit it, I know what was intended

mattdailis: Sounds good - I have no intention of committing to this branch before consulting you first! I made new branches for those two commits just for demonstration purposes (EDIT: whoops! didn't see Dan's commits from yesterday. I stand corrected)

> you did get, which will throw an NPE if it's absent

Is that true? My read on the code was that it acts like `Map`, which returns `null` if the key is missing https://github.com/javaee/jsonp/blob/4ee2eded959188ff0948950a1bd8cd7b07563310/impl/src/main/java/org/glassfish/json/JsonObjectBuilderImpl.java#L340-L343

I do think `containsKey` is clearer though :+1:

deployment/hasura/migrations/Aerie/25_workspaces_setup/up.sql
```diff
@@ -0,0 +1,224 @@
...
+    set parcel_id = ppw.parcel_id
+    from parcels_per_ws ppw
+    where ws.id = ppw.workspace_id and wrow = 1
+)
+  -- Move user sequences to the new workspaces as needed
+  update sequencing.user_sequence us
+    set workspace_id = iws.new_wid
+  from inserted_ws iws, parcels_per_ws ppw, sequencing.workspace ws
+  where us.workspace_id = ws.id
+    and iws.name = ws.name || ' (' || ppw.pname ||')'
```
dandelany: minor: @Mythicaeda can you clarify if this line is needed or not? Was trying to understand if it was safe to remove & just match on `id`s instead

dandelany: nevermind, we understand why this is needed, but would prefer to keep the old workspace ID in `inserted_ws` & then do a match like `and iws.old_workspace_id = ws.id` here to avoid any string/name matching brittleness

dandelany: on third thought... it doesn't seem like `parcel.name` is guaranteed to be unique, so this may be more of a requirement than a preference, since *not* using IDs here could result in a name collision?

Mythicaeda: The update on L199 fixes any workspace name conflicts.

Mythicaeda: > nevermind, we understand why this is needed, but would prefer to keep the old workspace ID in inserted_ws & then do a match like and iws.old_workspace_id = ws.id here to avoid any string/name matching brittleness

This is impossible; you cannot return an input field from an `insert` statement. Trust me, that was my first approach. I do see the concern about the string matching, however

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceAppDriver.java
```diff
@@ -0,0 +1,125 @@
...
+      config.jetty.multipartConfig.maxInMemoryFileSize(10, SizeUnit.MB); //the maximum file size to handle in memory
+      config.jetty.multipartConfig.maxTotalRequestSize(1, SizeUnit.GB); //the maximum size of the entire multipart request
+
+      config.plugins.enableCors(cors -> cors.add(CorsPluginConfig::anyHost));
+      config.plugins.register(workspaceBindings);
+      config.jetty.server(() -> server);
+    });
+
+    // Start the HTTP server.
+    javalin.start(9999);
```
dandelany: can we change this to `javalin.start(configuration.httpPort());` or even `javalin.start();` here since the connector already defines the port? `9999` seems like it will be overridden/not necessary?

Mythicaeda: I changed it to the first option to align with `AerieAppDriver`

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/DirectoryTree.java
Mythicaeda: Added doc comments to explain why this is not using Apache `FileUtils`

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: Exception is now bubbled-up

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: Fixed it. It should not not short-circuit

dandelany (approved):
#1719 has been merged into this branch, now thoroughly tested (including migrations) and approved. I also updated the `aerie-actions` dependency to `1.0.0` now that it has been published. The security scan should pass on `develop` after the merge. 

Merging this for testing on aerie-dev & release 3.6.0.  Thanks everyone for the hard work on this!

PR #1712, opened 2025-07-22
Workspaces copy/move support

joshhaug:
* **Tickets addressed:** No ticket
* **Review:** By file

## Description
At a high level, this just allows the user to specify one of two POSTs:

```json
{
  "moveTo": "moved/file/location.txt",
  "toWorkspace": 1
}
```

```json
{
  "copyTo": "copied/file/location.txt",
  "toWorkspace": 1
}
```

If `toWorkspace` is not specified, or is the same as the context's `PathInformation.workspaceId`, then these are assumed to be operations within the given context workspace. Otherwise, they are performed across workspaces. 

The `post` method now calls out to `handleCopy` or `handleMove`, which perform some shared tests via `isCopyOrMoveValid`. If those pass, we call the appropriate `(move|copy)(Directory|File)` method. These methods accept both a source and target workspace Id, which prevents duplicated logic between intra- and inter-workspace copy/move ops. 

Additionally, I fixed a typo in `GetExtensionMappingsAction` which was preventing things from working when testing.

### DIFF
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -214,12 +214,12 @@ public class WorkspaceBindings implements Plugin {
       type = context.queryParamAsClass("type", String.class)
                     .allowNullable()
                     .check(Objects::nonNull, "'type' must be provided.")
-                    .check(ts -> ts.equalsIgnoreCase("file") || ts.equalsIgnoreCase("directory"),
+                    .check(ts -> ts == null || ts.equalsIgnoreCase("file") || ts.equalsIgnoreCase("directory"),
                            "'type' must be one of 'file' or 'directory'")
                     .get();
       overwrite = context.queryParamAsClass("overwrite", Boolean.class).getOrDefault(false);
     } catch (ValidationException ve) {
-      context.status(400).result(ve.getMessage());
+      context.status(400).result(ve.getMessage() != null ? ve.getMessage() : "Invalid request");
       return;
     }
 
@@ -234,7 +234,7 @@ public class WorkspaceBindings implements Plugin {
     if ("file".equalsIgnoreCase(type)) {
       // Reject the request if the file isn't provided.
       final var file = context.uploadedFile("file");
-      if (file == null || !pathInfo.fileName().equals(file.filename())) {
+      if (file == null) {
         context.status(400).result("No file provided with the name " + pathInfo.fileName());
         return;
       }
@@ -271,7 +271,7 @@ public class WorkspaceBindings implements Plugin {
       context.status(200).result("Success");
     } catch (Exception e) {
       e.printStackTrace();
-      context.status(400).result("Error: " + e.getMessage());
+      context.status(400).result("Error: " + e.getMessage() != null ? e.getMessage() : "Invalid request");
     }
   }
 
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -14,6 +14,7 @@ import io.javalin.validation.ValidationException;
 import javax.json.Json;
 import javax.json.JsonException;
 import javax.json.JsonObject;
+import javax.json.stream.JsonParsingException;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.StringReader;
@@ -255,7 +256,7 @@ public class WorkspaceBindings implements Plugin {
     }
   }
 
-  private void post(Context context) throws IOException, SQLException, NoSuchWorkspaceException {
+  private void post(Context context) {
 
     try (final var bodyReader = Json.createReader(new StringReader(context.body()))) {
       final var bodyJson = bodyReader.readObject();
@@ -269,9 +270,28 @@ public class WorkspaceBindings implements Plugin {
       }
 
       context.status(200).result("Success");
-    } catch (Exception e) {
+
+    } catch (JsonException e) {
+      // Malformed JSON in request body
+      context.status(400).result("Malformed JSON: " + e.getMessage());
+
+    } catch (IllegalArgumentException e) {
+      // Client passed unsupported keys or invalid logic
+      context.status(400).result("Invalid request: " + e.getMessage());
+
+    } catch (NoSuchWorkspaceException e) {
+      // Domain-specific error: workspace doesn't exist
+      context.status(404).result("Workspace not found: " + e.getMessage());
+
+    } catch (IOException | SQLException e) {
+      // Server-side failure, log and return 500
       e.printStackTrace();
-      context.status(400).result("Error: " + e.getMessage() != null ? e.getMessage() : "Invalid request");
+      context.status(500).result("Internal server error while processing the request.");
+
+    } catch (Exception e) {
+      // Catch-all for unexpected issues
+      e.printStackTrace();
+      context.status(500).result("Unexpected error: " + (e.getMessage() != null ? e.getMessage() : "Unknown error"));
     }
   }
 
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -258,11 +258,16 @@ public class WorkspaceBindings implements Plugin {
     try (final var bodyReader = Json.createReader(new StringReader(context.body()))) {
       final var bodyJson = bodyReader.readObject();
 
+      boolean sameWorkspace;
       // Parse what the post request is for
-      final var moveRq = bodyJson.containsKey("moveTo");
+      if (bodyJson.containsKey("toWorkspace") && pathInfo.workspaceId() != bodyJson.getInt("toWorkspace")) {
+        sameWorkspace = false;
+      } else {
+        sameWorkspace = true;
+      }
 
       // Perform the request
-      if (moveRq) {
+      if (sameWorkspace) {
         final var destination = Path.of(bodyJson.getString("moveTo"));
 
         // Reject if source does not exist
@@ -293,11 +298,51 @@ public class WorkspaceBindings implements Plugin {
             context.status(500).result("Unable to move file. " + e.getMessage());
           }
         }
+      } else {
+        final var destination = Path.of(bodyJson.getString("moveTo"));
+        final int targetWorkspace = bodyJson.getInt("toWorkspace");
+
+        // Reject if source does not exist
+        if (!workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath)) {
+          context.status(404).result(pathInfo.fileName() + " does not exist in the source workspace.");
+          return;
+        }
+
+        // Reject if target workspace does not exist
+        if (!workspaceService.checkFileExists(targetWorkspace, Path.of("/"))) {
+          context.status(404).result("Target workspace with ID "+targetWorkspace+" does not exist.");
+          return;
+        }
+
+        // Return "Conflicted" if destination exists
+        if (workspaceService.checkFileExists(targetWorkspace, destination)) {
+          context.status(409).result(destination.getFileName().toString() + " already exists");
+          return;
+        }
+
+        if (workspaceService.isDirectory(pathInfo.workspaceId, pathInfo.filePath())) {
+          if (workspaceService.moveDirectoryToWorkspace(pathInfo.workspaceId, pathInfo.filePath, targetWorkspace, destination)) {
+            context.status(200);
+          } else {
+            context.status(500).result("Unable to move directory.");
+          }
+        } else {
+          try {
+            if (workspaceService.moveFileToWorkspace(pathInfo.workspaceId, pathInfo.filePath, targetWorkspace, destination)) {
+              context.status(200);
+            } else {
+              context.status(500).result("Unable to move file.");
+            }
+          } catch (SQLException e) {
+            context.status(500).result("Unable to move file. " + e.getMessage());
+          }
+        }
       }
     } catch (JsonException je) {
       final String helpText = """
           {
-            "moveTo": text  // Path to where in the workspace the file should be moved to
+            "moveTo": text,     // Path to where in the workspace the file should be moved to
+            "toWorkspace": int  // Id of the target workspace if different from source workspace (optional)
           }
           """;
       context.status(400).result("Request body is malformed. Request body format is:\n" + helpText);
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -224,8 +224,10 @@ public class WorkspaceBindings implements Plugin {
       return;
     }
 
-    // Report a "Conflict" status if the object already exists and "overwrite" is false
-    if(workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath) && !overwrite) {
+    // Report a "Conflict" status if the file already exists, is a file and not a directory, and "overwrite" is false
+    if(workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath)
+       && !workspaceService.isDirectory(pathInfo.workspaceId, pathInfo.filePath)
+       && !overwrite) {
       context.status(409).result(pathInfo.fileName() + " already exists.");
       return;
     }
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -4,7 +4,6 @@ import com.auth0.jwt.exceptions.JWTVerificationException;
 import gov.nasa.jpl.aerie.workspace.server.postgres.NoSuchWorkspaceException;
 import io.javalin.Javalin;
 import io.javalin.apibuilder.ApiBuilder;
-import io.javalin.http.BadRequestResponse;
 import io.javalin.http.ContentType;
 import io.javalin.http.Context;
 import io.javalin.http.UnauthorizedResponse;
@@ -15,7 +14,6 @@ import javax.json.Json;
 import javax.json.JsonException;
 import javax.json.JsonObject;
 import javax.json.JsonReader;
-import javax.json.stream.JsonParsingException;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.StringReader;
@@ -315,23 +313,28 @@ public class WorkspaceBindings implements Plugin {
     }
   }
 
-  private record CopyMoveValid(int status, String message){};
+  private record CopyMoveValid(int status, String message){}
 
-  private CopyMoveValid isCopyOrMoveValid(int sourceWorkspace, Path sourceFile, int targetWorkspace, Path targetFile)
-  throws NoSuchWorkspaceException {
-    // Reject if source does not exist
-    if (!workspaceService.checkFileExists(sourceWorkspace, sourceFile)) {
-      return new CopyMoveValid(404, sourceFile + " does not exist in the source workspace.");
+  private CopyMoveValid isCopyOrMoveValid(int sourceWorkspace, Path sourceFile, int targetWorkspace, Path targetFile) {
+    try {
+      // Return "Resource Not Found" if sourceFile does not exist
+      if (!workspaceService.checkFileExists(sourceWorkspace, sourceFile)) {
+        return new CopyMoveValid(404, sourceFile + " does not exist in the source workspace.");
+      }
+    } catch (NoSuchWorkspaceException se) {
+      // Return "Resource Not Found" if source workspace does not exist
+      return new CopyMoveValid(404, "Source workspace with ID "+sourceWorkspace+" does not exist.");
     }
 
-    // Reject if target workspace does not exist
-    if (!workspaceService.checkFileExists(targetWorkspace, Path.of("/"))) {
+    try {
+      // Return "Conflicted" if destination exists
+      if (workspaceService.checkFileExists(targetWorkspace, targetFile)) {
+        return new CopyMoveValid(409, targetFile + " already exists");
+      }
+    }
+    catch (NoSuchWorkspaceException se) {
+      // Return "Resource not found" if target workspace does not exist
       return new CopyMoveValid(404, "Target workspace with ID "+targetWorkspace+" does not exist.");
-    }
-
-    // Return "Conflicted" if destination exists
-    if (workspaceService.checkFileExists(targetWorkspace, targetFile)) {
-      return new CopyMoveValid(409, targetFile + " already exists");
     }
 
     return new CopyMoveValid(200, "Success");
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -254,9 +254,6 @@ public class WorkspaceBindings implements Plugin {
   }
 
   private void post(Context context) throws IOException, SQLException, NoSuchWorkspaceException {
-    System.out.println("Request body: " + context.body());
-
-    final var pathInfo = PathInformation.of(context);
 
     try (final var bodyReader = Json.createReader(new StringReader(context.body()))) {
       final var bodyJson = bodyReader.readObject();
@@ -277,8 +274,7 @@ public class WorkspaceBindings implements Plugin {
   }
 
   private boolean isCopyOrMoveValid(Context context, int sourceWorkspace, Path sourceFile, int targetWorkspace, Path targetFile)
-  throws NoSuchWorkspaceException
-  {
+  throws NoSuchWorkspaceException {
     // Reject if source does not exist
     if (!workspaceService.checkFileExists(sourceWorkspace, sourceFile)) {
       context.status(404).result(sourceFile + " does not exist in the source workspace.");
--- a/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
+++ b/workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
@@ -2,6 +2,7 @@ package gov.nasa.jpl.aerie.workspace.server;
 
 import com.auth0.jwt.exceptions.JWTVerificationException;
 import gov.nasa.jpl.aerie.workspace.server.postgres.NoSuchWorkspaceException;
+import gov.nasa.jpl.aerie.workspace.server.postgres.WorkspaceFileOpException;
 import io.javalin.Javalin;
 import io.javalin.apibuilder.ApiBuilder;
 import io.javalin.http.ContentType;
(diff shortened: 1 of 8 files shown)


workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -213,25 +215,27 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
       type = context.queryParamAsClass("type", String.class)
                     .allowNullable()
                     .check(Objects::nonNull, "'type' must be provided.")
-                    .check(ts -> ts.equalsIgnoreCase("file") || ts.equalsIgnoreCase("directory"),
+                    .check(ts -> ts == null || ts.equalsIgnoreCase("file") || ts.equalsIgnoreCase("directory"),
```
Mythicaeda: `ts` physically cannot be null, see L217 `.check(Objects::nonNull, "'type' must be provided.")`. Please remove this change or explain why it is necessary.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -213,25 +215,27 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
                            "'type' must be one of 'file' or 'directory'")
                     .get();
       overwrite = context.queryParamAsClass("overwrite", Boolean.class).getOrDefault(false);
     } catch (ValidationException ve) {
-      context.status(400).result(ve.getMessage());
+      context.status(400).result(ve.getMessage() != null ? ve.getMessage() : "Invalid request");
```
Mythicaeda, line 226: Can you please provide an example case where `ve.getMessage` was `null`? A thrown exception should always have a `message` AFAIK

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -213,25 +215,27 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
       return;
     }
 
-    // Report a "Conflict" status if the object already exists and "overwrite" is false
-    if(workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath) && !overwrite) {
+    // Report a "Conflict" status if the file already exists, is a file and not a directory, and "overwrite" is false
+    if(workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath)
+       && !workspaceService.isDirectory(pathInfo.workspaceId, pathInfo.filePath)
+       && !overwrite) {
```
Mythicaeda: **Major**: `isDirectory` should **not** be added here. This is a check for if a resource with that name already exists in **any** form (as indicated by it returning a [409: Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409)). More critically, this check is happening **BEFORE** we've even distinguished if the user is trying to create a `file` or a `directory`.

You can add a "make sure there's not a directory with this name" check within the file-uploading part of the method (between lines 235 and 241) if you'd like, but please revert the change here.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -213,25 +215,27 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
       context.status(409).result(pathInfo.fileName() + " already exists.");
       return;
     }
 
     if ("file".equalsIgnoreCase(type)) {
       // Reject the request if the file isn't provided.
       final var file = context.uploadedFile("file");
-      if (file == null || !pathInfo.fileName().equals(file.filename())) {
+      if (file == null) {
```
Mythicaeda: Please revert the change here. The second half of this `if` statement ensures that the request was formatted correctly. That is, that the request said "I'm uploading file `XYZ.seq`, here is the file `XYZ.seq`"

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
```diff
@@ -47,15 +47,52 @@ record FileStream(InputStream readingStream, String fileName, long fileSize){}
...
+   * Move a file within a workspace or between workspaces.
+   * @param oldWorkspaceId the id of the source workspace
+   * @param oldFilePath the path, relative to the source workspace root, that the file is currently at
+   * @param newWorkspaceId the id of the target workspace, note that this can be the same as oldWorkspaceId
+   * @param newFilePath the new path of the file, relative to the new workspace root
    * @return true if the file was moved, false otherwise
    */
-  boolean moveFile(final int workspaceId, final Path oldFilePath, final Path newFilePath)
-  throws NoSuchWorkspaceException, SQLException;
+  boolean moveFile(final int oldWorkspaceId, final Path oldFilePath, final int newWorkspaceId, final Path newFilePath)
```
Mythicaeda, line 71: Open for discussion: I think the "between workspaces" case should be an override of `moveFile`/`moveDirectory`, as it saves us a needless DB call in the far more common case of moving files/directories within a workspace.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
```diff
@@ -47,15 +47,52 @@ record FileStream(InputStream readingStream, String fileName, long fileSize){}
...
+
+  /**
+   * Move a directory within a workspace or between workspaces.
+   * @param oldWorkspaceId the id of the source workspace
+   * @param oldDirectoryPath the path, relative to the source workspace root, of the directory
+   * @param newWorkspaceId the id of the target workspace, note that this can be the same as oldWorkspaceId
+   * @param newDirectoryPath the new path of the directory, relative to the new workspace root
+   * @return true if the directory was moved, false otherwise
+   */
+  boolean moveDirectory(final int oldWorkspaceId, final Path oldDirectoryPath, final int newWorkspaceId, final Path newDirectoryPath)
```
Mythicaeda, line 95: Minor: Can you please move `moveDirectory`/`copyDirectory` back to where it was in the file? Prior to this change, the class was organized to have all of the file methods grouped together, and all of the directory methods grouped together, both groups ordered as `RCUD` (which, to be fair, I probably should've ordered them in the more standard `CRUD` order, but still). Putting the directory methods next to the file methods breaks this organization.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -195,17 +281,22 @@ public boolean createDirectory(final int workspaceId, final Path directoryPath)
...
+    final var newPath = newRepoPath.resolve(newDirectoryPath);
 
     // Do not permit the workspace's root directory to be moved
-    // Or to move a directory to replace the root directory
-    if(Files.isSameFile(oldPath, repoPath) || Files.isSameFile(newPath, repoPath)) return false;
+    if(Files.isSameFile(oldPath, oldRepoPath)) throw new WorkspaceFileOpException("Cannot move the workspace root directory.");
+
+    // Do not permit a directory to replace the root directory
+    // TODO: maybe we want to allow this? This check fails anyways because newPath is not a real file at this point
+    // if(Files.isSameFile(newPath, newRepoPath)) return false;
```
Mythicaeda: Three things:
1) `TODO` is left behind
2) `newPath` should absolutely be a real file if it's the other repo's root path, not sure what that you mean by that part of the `TODO`. Could you elaborate?
3) No, we don't want to allow the user to replace the root directory of the new workspace (this is to maintain consistency with the "within workspace" interaction)

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: I included this because this was throwing a failure when performing the `equalsIgnoreCase` check. I'm confused by this as presumably the above `check()` would short-circuit, but I guess the behavior of this is a bit more nuanced than I thought.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: My opinion is that `file.filename()` may not be the same as `pathInfo.fileName()`. For the following action code, for instance:
```ts
  async writeFile(
    name: string,
    contents: string,
    overwrite: boolean = false
  ): Promise<any> {
    const formData = new FormData();
    formData.append("file", new Blob([contents]), name);
    const path = `/ws/${this.workspaceId}/${encodeURIComponent(name)}?type=file&overwrite=${overwrite}`;
    await this.reqWorkspace(
      path,
      'PUT',
      formData
    );
    return { success: true };
  }
```
This check could fail if the `name` is `dir1/dir2/test.json`, because `pathInfo.fileName()` would be `test.json` but `file.filename()` would be `dir1/dir2/test.json`.  Do we consider this to be a incorrectly formatted request?

Mythicaeda: `file.filename()` would be `test.json`. That's the explicit implementation of the method in `PathInfo` (L42 of this file)

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
joshhaug: Done in c5505dfaa7b4262dba069c95358ff92dedcb795d

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
joshhaug: Done in ed2df9329d43e046920f89024c07ad4145ca193b

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: Done in e86d0c325be6a100306a5811660b431da9ff6cfb

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: `file.filename()` comes from `context.uploadedFile("file")`, which is set via the FormData above. From my testing it seems like `filename` is the full path of the file (`name` in the above typescript example).

Mythicaeda: Yes, but that TypeScript request is misformatted.

This line:
`formData.append("file", new Blob([contents]), name);`
should set the file name to just the actual file name, not the complete path.

joshhaug: Addressed in c479eb806b63fcd44bf30e259e160cc5a8856591

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Well, either way, `null` is **not** a permitted value. If an exception is occurring when `ts` is `null`, why not just reverse the check instead?

```suggestion
                    .check(ts -> "file".equalsIgnoreCase(ts) || "directory".equalsIgnoreCase(ts),
```

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: I encountered this during debugging but I don't remember the case that triggered it. I was surprised as well. I can try to dig up the case that triggered this -- it could've potentially been a cascade from another bug.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
joshhaug: Would you mind clarifying? I'm not sure I follow 😕

joshhaug:
Just reviewed these comments with @dandelany , I think we are ready for re-review if you have some time @Mythicaeda

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Comment on L227 was not reverted.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +255,137 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
+        handleCopy(context, bodyJson);
+      } else {
+        throw new IllegalArgumentException("Unsupported operation. Expected 'moveTo' or 'copyTo'.");
+      }
+
+      context.status(200).result("Success");
+    } catch (Exception e) {
+      e.printStackTrace();
+      context.status(400).result("Error: " + e.getMessage() != null ? e.getMessage() : "Invalid request");
+    }
```
Mythicaeda: Javalin already has a generic exception handler. Please replace this block with the actual exceptions you expect to catch (and update the method signature accordingly). Additionally, status code 400 is **not** appropriate as a catch-all error code. `400` means that there was something wrong that in the client's request (ie, their body was misformatted, they forgot a query variable)

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +255,137 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
     }
   }
 
-  private void post(Context context) throws NoSuchWorkspaceException, IOException {
-    final var pathInfo = PathInformation.of(context);
+  private void post(Context context) throws IOException, SQLException, NoSuchWorkspaceException {
```
Mythicaeda: The `post` method used to contain a `helpText` and would respond with the appropriate body format should the user's body be misformatted. The removal of this behavior is a regression and makes the API harder to use for clients who do not already know the appropriate format. Please restore and expand on the helpText.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
Mythicaeda: I'm proposing having the following overrides of `moveFile`:
```java
boolean moveFile(final int workspaceId, final Path oldFilePath, final Path newFilePath)

boolean moveFile(final int oldWorkspaceId, final Path oldFilePath, final int newWorkspaceId, final Path newFilePath)
```

joshhaug:
Note to myself: need to check/verify empty file loading

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +255,137 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
+    // Return "Conflicted" if destination exists
+    if (workspaceService.checkFileExists(targetWorkspace, targetFile)) {
+      return new CopyMoveValid(false, 409, targetFile + " already exists");
+    }
 
-        // Reject if source does not exist
-        if (!workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath)) {
-          context.status(404).result(pathInfo.fileName() + " does not exist in the workspace.");
-          return;
+    return new CopyMoveValid(true, 0, "");
```
Mythicaeda: Since you're using HTTP status codes as the values elsewhere, you should set status to `200` here. Additionally, you don't need the boolean `valid`, since `status` already communicates that.

joshhaug:
> Note to myself: need to check/verify empty file loading

Fix implemented in aerie-actions library, and did not require any changes in backend as I originally thought: https://github.com/NASA-AMMOS/aerie-actions/pull/8#issuecomment-3115104197

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: Good catch, updated in 5b3b18db8f72fc8e905ec6d95846ea80460d9ac0

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
joshhaug: Implemented an override in 5b3b18db8f72fc8e905ec6d95846ea80460d9ac0, but it is not currently being used.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: Ok, updated in 6312c42f1a2dc5f97ad00945021a0c3d38851c93

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: Updated in a14d93e5202b0401531dd381dbaf86cb7b92b417

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
joshhaug: Good idea, updated in 05b35926a70b1061edc8d6461856fd97eb7fa2c7

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +257,172 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
+      // Logical errors or unsupported operations
+      context.status(400).result("Invalid request: " + e.getMessage() + "\n\n" + helpText);
+
+    } catch (NoSuchWorkspaceException e) {
+      // Workspace not found
+      context.status(404).result("Workspace not found: " + e.getMessage());
+
+    } catch (IOException | SQLException e) {
+      // Internal server error
+      e.printStackTrace();
```
Mythicaeda: Occurring in multiple locations and multiple files: The correct way for the server to log an error is by using a `org.slf4j.Logger`. Please update this to make Workspaces consistent with the rest of our services

Specifically, this occurs twice in `WorkspaceBindings` and twice in `WorkspaceFileSystemService`.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -195,17 +288,21 @@ public boolean createDirectory(final int workspaceId, final Path directoryPath)
...
+    final var newRepoPath = postgresRepository.workspaceRootPath(newWorkspaceId);
+    final var newPath = newRepoPath.resolve(newDirectoryPath);
 
     // Do not permit the workspace's root directory to be moved
-    // Or to move a directory to replace the root directory
-    if(Files.isSameFile(oldPath, repoPath) || Files.isSameFile(newPath, repoPath)) return false;
+    if(Files.isSameFile(oldPath, oldRepoPath)) throw new WorkspaceFileOpException("Cannot move the workspace root directory.");
+
+    // Do not permit a directory to replace the root directory
+    if(Files.isSameFile(newPath, oldRepoPath)) throw new WorkspaceFileOpException("Cannot replace the workspace root directory.");
```
Mythicaeda: You are missing a check to prevent replacing the root directory of the other workspace:

```
Files.isSameFile(newPath, newRepoPath)
```

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -159,6 +160,98 @@ public boolean moveFile(final int workspaceId, final Path oldFilePath, final Pat
...
+  {
+    final var sourceRepoPath = postgresRepository.workspaceRootPath(sourceWorkspaceId);
+    final var sourcePath = sourceRepoPath.resolve(sourceFilePath);
+    final var destRepoPath = postgresRepository.workspaceRootPath(destWorkspaceId);
+    final var destPath = destRepoPath.resolve(destFilePath);
+
+    try {
+      // Validate source exists and is a directory
+      if (!Files.exists(sourcePath)) throw new WorkspaceFileOpException("Source directory \"%s\" in workspace %d does not exist.".formatted(sourceFilePath, sourceWorkspaceId));
+      if (!Files.isDirectory(sourcePath)) throw new WorkspaceFileOpException("Source directory \"%s\" in workspace %d is not actually a directory. This is a bug.".formatted(sourceFilePath, sourceWorkspaceId));
```
Mythicaeda: Please exclude the phrase "This is a bug" from the error message.
```suggestion
      if (!Files.isDirectory(sourcePath)) throw new WorkspaceFileOpException("Source directory \"%s\" in workspace %d is not actually a directory.".formatted(sourceFilePath, sourceWorkspaceId));
```

Additionally, this state is not unsalvageable -> you could just go into the `copyFile` with the same args (However, this "additionally" part is nonblocking, I'm just musing on what can be done instead of throwing)

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -159,6 +160,98 @@ public boolean moveFile(final int workspaceId, final Path oldFilePath, final Pat
...
+          final Path target = destPath.resolve(relative);
+          try {
+            if (Files.isDirectory(source)) {
+              Files.createDirectories(target);
+            } else {
+              Files.copy(source, target);
+            }
+          } catch (IOException e) {
+            throw new UncheckedIOException(e);
+          }
```
Mythicaeda, line 238: If you're not going to handle the IOException, you should let it be thrown as opposed to rewrapping it as Unchecked (especially since you catch `IOException` further on in the method).

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +257,172 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
+      context.status(500).result("Internal server error while processing the request.");
```
Mythicaeda:
```suggestion
      context.status(500).result("Internal server error while processing the request." + e.getMessage());
```

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -252,58 +257,172 @@ private void put(Context context) throws NoSuchWorkspaceException, IOException {
...
+  throws NoSuchWorkspaceException {
+    // Reject if source does not exist
+    if (!workspaceService.checkFileExists(sourceWorkspace, sourceFile)) {
+      return new CopyMoveValid(404, sourceFile + " does not exist in the source workspace.");
+    }
+
+    // Reject if target workspace does not exist
+    if (!workspaceService.checkFileExists(targetWorkspace, Path.of("/"))) {
+      return new CopyMoveValid(404, "Target workspace with ID "+targetWorkspace+" does not exist.");
+    }
```
Mythicaeda: Nonblocking: I've mixed feelings on keeping these checks, as they're already checked by the actual move/copy procedure, and quite quickly too (especially checking if the other workspace exists). I don't feel strongly about removing them, however, so feel free to resolve this comment.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
Mythicaeda: Right. I was proposing creating _and using_ it. If it's going to go unused, then I'd rather the change be reverted/dropped.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Since I know that I'm going to be editing this part of the code after we remerge, I'll call this comment resolved, but you ought to know that this is still not the original message. The original message was `Report a "Conflict" status if the object already exists and "overwrite" is false`. In the future, I'd recommend dropping or editing prior commits to rollback a change in order to avoid subtle differences like this.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
dandelany: Fixed in ab8d781

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
dandelany: * Fixed in ab8d781
* I think we *only* care about this check, not `oldRepoPath`, so I replaced line 305
* `Files.isSameFile` throws if either path doesn't exist, which `newPath` won't in the nominal case, so I made it `Files.exists(newPath) && Files.isSameFile(newPath, newRepoPath)`

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
dandelany: Fixed in ab8d781, I think it's OK to throw in this "shouldn't happen" case so we can figure out why it did

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
dandelany: Annoyingly Java doesn't let you throw checked exceptions like `IOException` unless you declare them, which you can’t do with an inline lambda. So it seems like you could either extract this into a separate function, or throw something else like a `RuntimeException`, but that's even less descriptive. Are either of those much of an improvement/more idiomatic than just doing this? `¯\_(ツ)_/¯`

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
dandelany: fixed in ab8d781694b733db717e63c53a2116bf87ea553d

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
dandelany: Notably this doesn't seem to be the case in the `moveDirectory` function, which only checks for root directory stuff, but not existence/replacement of existing directories. I would be in favor of only checking once, if we can audit & guarantee they're really checked everywhere they need to be. But like you said, nonblocking.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: No, if the exception is being wrapped in `Unchecked` to get around lambda restrictions, then `UncheckedIOException` makes the most sense.

workspace_server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
dandelany: Fixed in [876e7ec](https://github.com/NASA-AMMOS/aerie/pull/1712/commits/876e7ec47e930e5747528e90a3070cb5ccb23798) - discussed with @mattdailis & we decided to remove the override and add a check in the `moveFile` implementation to avoid the extra DB check if the IDs are the same.

Mythicaeda: Thank you. You should do the same for `moveDirectory` and `copyFile/copyDirectory` so that the behavior isn't unique to `moveFile`

Mythicaeda (approved):

dandelany:
finished review with @Mythicaeda and [tests pass](https://github.com/NASA-AMMOS/aerie/actions/runs/16605212133/job/46976680437) - will resolve the Publish scan issue separately. merging this.

PR #1714, opened 2025-07-24
Add description on activity types, parameters and resources in mission models

### LINKED ISSUE #1083: Store activity type / mission model description into the database
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

No - I haven't considered

### Related problems

_No response_

Right now we're building description into the compiled mission model, we should also store this into the database.

### LINKED ISSUE #1474: Provide description metadata for resources and activity parameters
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

It should be easier for users to get information about activities and resources. Users may not have access to source code, and cross-referencing resources/params from Aerie to JavaDoc is cumbersome. 

I want to be able to annotate resources and activity parameters with description information, similar to what we can do for units. Supposedly the metadata annotation capability should support this, but i have been unable to get it to work except for the unit case: https://jpl.slack.com/archives/C0163E42UBF/p1702073797959459

tratsko:
* **Tickets addressed:** Closes #1083 and closes #1474
* **Review:** By commit  

## Description

This adds description as an annotation to add on an activity definition in the mission model. @Description on each activity. This is an optional annotation that takes in a string as the description under an activity. 

This annotation can also be added onto parameters in activities to define a description for said parameter.

The mission model compiles with this annotation stored in the activity type and or parameters depending on which the annotation was placed onto.

Also there is a description the user can now add to resources when registering them. This is passed in as another parameter in the function that registers the resources.

Once the compiled model is uploaded to the Merlin server it gets the description and adds it to the activity type and or parameters in the database when each activity is created. Also when the resources are loaded into the database, the description is added to the description column on the resource type if it was defined in the mission model.

## Verification

It compiles and picks up tags when the mission model is uploaded with annotations to the database. This verification was done in the e2e test in MissionModelTests for the activty types, parameters and resources.

## Documentation

https://nasa-ammos.github.io/aerie-docs/mission-modeling/activity-types/introduction/#activity-type-metadata
This documentation on putting tags in the javadoc on mission models is invalidated by these changes. This documentation on the activity annotations is inaccurate so it will need to get changed.

## Future work

We really ought to rethink how unit is handled. Maybe move the Unit annotation from contrib to framework/annotations?


dandelany:
Per discussion with @shaheerk94 and @tratsko today, we decided to:
* Extend this PR before review to add `@description` support to **resources** and **activity parameters**, in order to cover the use cases in #1474 
* Prioritize backend/GraphQL API support for these descriptions. We will add them to the UI also, but if there are any design questions etc., this PR may be merged earlier, independently from UI.

dandelany:
Updates from todays meeting with @mattdailis @tratsko 
* `@Description` annotation has been added to **activity parameters**, following the same pattern from activity types above
* Adding description to resources is more complicated, since they are defined dynamically
* @tratsko  will add an overload of the `discrete` and `real` methods (user-facing functions for resource definitions) with an additional (optional) description *parameter* (instead of annotation), and thread this down to the rest of the code & into the DB
* This will also need to be updated in the `streamline` resource framework since some users use this as their primary method for defining resources.
* References for Tim to follow the logic for resources: 
  - "top of the hill": https://github.com/NASA-AMMOS/aerie/blob/6587e8409220bff9a59de3d91b5594b4faef1441/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/LocalMissionModelService.java#L356
  - registrar: https://github.com/NASA-AMMOS/aerie/blob/43d4357f4e21aeb83926525b3118342533866155/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/Registrar.java#L27

deployment/hasura/migrations/Aerie/28_description_annotations/up.sql
```diff
@@ -0,0 +1,11 @@
...
+add column description text default null;
+comment on column merlin.activity_type.description is e''
+  'The description of this activity type.';
+
+alter table merlin.resource_type
+  add column description text default null;
+comment on column merlin.resource_type.description is e''
+  'The description of this resource type.';
+
+call migrations.mark_migration_applied('28');
```
Mythicaeda: Typo: Input needs to be numeric, not text (applies to up and down)

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/metadata/UnitRegistrar.java
```diff
@@ -19,8 +18,17 @@ public static <T> ValueMapper<T> withUnit(final String unit, final ValueMapper<T
...
     registrar.discrete(name, resource, withUnit(unit, valueMapper));
   }
+
+  public static <T> void discreteResource(final Registrar registrar, final String name, final Resource<T> resource, final ValueMapper<T> valueMapper, final String unit, final String description) {
+    registrar.discrete(name, resource, withUnit(unit, valueMapper), description);
+  }
+
   public static void realResource(final Registrar registrar, final String name, final Resource<RealDynamics> resource, final String unit) {
-    registrar.realWithMetadata(name, resource, "unit", unit, new ValueMapper<String>() {
+    realResource(registrar, name, resource, unit, null);
```
Mythicaeda, line 32: Not disagreeing with the design, but why was `null` chosen to represent a lack of description as opposed to a) Optional.empty() or b) the empty string?

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
```diff
@@ -20,9 +21,10 @@
  */
 @ActivityType("BiteBanana")
 @Subsystem("Eat")
+@Description("Takes a bite out of the banana")
 public final class BiteBananaActivity {
   @Parameter
-
+  @Description("The size of the bite in meters")
   @Banannotation("Specifies the size of bite to take")
```
Mythicaeda: What's the difference between `Description` and `Banannotation` here?

merlin-framework-processor/src/main/java/gov/nasa/jpl/aerie/merlin/processor/metamodel/ParameterRecord.java
```diff
@@ -3,15 +3,18 @@
 import javax.lang.model.element.Element;
 import javax.lang.model.type.TypeMirror;
 import java.util.Objects;
+import java.util.Optional;
 
 public final class ParameterRecord {
   public final String name;
   public final TypeMirror type;
   public final Element element;
+  public final Optional<String> description;
```
Mythicaeda: Only `ParameterRecord` is using `Optional<String>` for its description field. Please resolve the inconsistency.

contrib/src/main/java/gov/nasa/jpl/aerie/contrib/metadata/UnitRegistrar.java
tratsko: I went with null because it keeps resource registration simple since you only have to pass a string instead of wrapping it in an optional.of() every time you want to add a description on a resource. I could change it to use an empty string instead if that is preferable!

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/BiteBananaActivity.java
tratsko: @Description is an Aerie supported annotation; @Banannotation is a model specific custom annotation. They both hold the same kind of data (a description of the parameter), just created in two different ways. I can move one to a different parameter to avoid having both on the same field if that’s clearer.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertSubsystemsAction.java
```diff
@@ -26,7 +26,7 @@
   private final PreparedStatement findStatement;
 
 
-  public InsertSubsystemsAction(final Connection connection) throws SQLException {
+  public  InsertSubsystemsAction(final Connection connection) throws SQLException {
```
Mythicaeda: Minor, non-blocking: the whitespace in this file was accidentally changed:

```suggestion
  public InsertSubsystemsAction(final Connection connection) throws SQLException {
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/InsertActivityTypesAction.java
```diff
@@ -50,12 +51,21 @@ public void apply(final int modelId, Collection<ActivityType> activityTypes, Map
 
 
         if (activityType.subsystem().isPresent()) {
-          int subsystemIndex = mapSubsystemsToIds.get(activityType.subsystem().get());
+          Integer subsystemIndex = mapSubsystemsToIds.get(activityType.subsystem().get());
+          if (subsystemIndex == null) {
+            throw new FailedInsertException("Unknown subsystem '" + activityType.subsystem().get() + "' for model_id=" + modelId + " in merlin.activity_type");
+          }
```
Mythicaeda: A more standard way of doing this check is:
```suggestion
if(!mapSubsystemsToIds.containsKey(activityType.subsystem().get())) {
  throw new FailedInsertException("Unknown subsystem '" + activityType.subsystem().get() + "' for model_id=" + modelId + " in merlin.activity_type");
}
int subsystemIndex = mapSubsystemsToIds.get(activityType.subsystem().get());
```
This makes it clear that:
1. the exception is being thrown because the key is not in the map **and** 
2. that `null` is not an accepted `value` for a `key` in the map.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityTypesAction.java
```diff
@@ -65,7 +66,8 @@ public List<ActivityType> get(final long missionModelId) throws SQLException {
             getJsonColumn(results, "computed_attributes_value_schema", valueSchemaP)
                 .getSuccessOrThrow($ -> new Error("Corrupt activity type computed attribute schema cannot be parsed: "
                                                   + $.reason())),
-            Optional.ofNullable((String) results.getObject("subsystem"))
+            Optional.ofNullable((String) results.getObject("subsystem")),
+            Optional.ofNullable(results.getString("description"))
```
Mythicaeda: Why was `subsystem` using a `.getObject` call?
```suggestion
            Optional.ofNullable(results.getString("subsystem")),
            Optional.ofNullable(results.getString("description"))
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -66,11 +66,12 @@ public static JsonValue serializeParameters(final List<Parameter> parameters) {
...
-    return serializeMap(pair -> Json.createObjectBuilder()
+    return serializeMap(pair -> {
+      final var builder = Json.createObjectBuilder()
             .add("schema", new ValueSchemaJsonParser().unparse(pair.getRight().schema()))
-            .add("order", pair.getLeft())
-            .build(),
-        parameterMap);
+            .add("order", pair.getLeft());
+      return builder.build();
+    }, parameterMap);
```
Mythicaeda: This change is unnecessary and should be reverted

merlin-sdk/src/main/java/gov/nasa/jpl/aerie/merlin/protocol/model/InputType.java
```diff
@@ -170,7 +170,7 @@ T instantiate(Map<String, SerializedValue> arguments)
   List<ValidationNotice> getValidationFailures(T value);
 
   /** A named parameter to an {@link InputType} */
-  record Parameter(String name, ValueSchema schema) {}
+  record Parameter(String name, ValueSchema schema) { }
```
Mythicaeda: This is an unneeded whitespace change and should be reverted:
```suggestion
  record Parameter(String name, ValueSchema schema) {}
```

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/NewDurationParameterActivity.java
Mythicaeda, line 1: Why did you delete this file?

Mythicaeda (commented):
Requesting a couple of small changes before approving

examples/banananation/src/main/java/gov/nasa/jpl/aerie/banananation/activities/NewDurationParameterActivity.java
mattdailis: I asked Tim to delete this file — it's identical to `DurationParameterActivity`. I added it by accident when coming up with test cases for migrating plans from one mission model to another.

Mythicaeda (approved):
LGTM!

PR #1715, opened 2025-07-24
Access resources from mission model in procedures

JoelCourtney:
* **Review:** By commit  

## Description

This PR lets goals and constraints query resources by referencing them through the mission model. Meaning you can write `simResults.resource(model.fruit)` instead of `simResults.resource("/fruit", Real.deserializer())`, where `mission` is a valid instance of the mission model.

- **Vile hack No. 1**: The goal/constraint gets the instance of the model through the `WithModel<M>` interface, which has a mutable static field `modelSingleton`. The scheduler or constraint action creates an instance of the model with the default sim config and sets the sim config at the beginning of the action. Goals/constraints that inherit from `WithModel<M>` can then get that instance with `this.model()`.
- **Vile Hack No. 2**: The only code this adds to the mission model is a `NameableResource` interface and `NamedResource` abstract implementor of that interface for convenience. Concrete resource types can inherit from `NamedResource`, which just lets the `Registrar` call `NameableResource.setName(string)`. This happens during the model constructor, so it happens both during simulation (which is useless) and during scheduling/constraint actions. The timeline library `SimulationResults` then provides a few `resource(...)` overloads that look for things like `NameableResource<RealDynamics>`, `NameableResource<Boolean>`, etc. It calls `NameableResource.getName()` and associates the payload type with a timeline type like `Real`, `Booleans`, etc.

This required some fiddling with ClassLoaders to make sure that the mission model was not loaded twice by different loaders (which isn't just a waste, it also causes a class cast exception).

## Verification

I've made two integration tests, one for scheduling and one for constraints. I've also manually verified that the constraint/goal jars don't include the mission model. (`unzip -l <jar> | grep banana`)

## Documentation

I swear I'll update the guide docs this time.

## Future work

Because Vile Hack No. 1 creates the model with the default sim config, this could cause some issues when we support other configs during scheduling. I'm assuming that the default config corresponds to the full model with no resources left out.

### DIFF
--- a/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/profiles/Numbers.kt
+++ b/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/profiles/Numbers.kt
@@ -223,5 +223,38 @@ data class Numbers<N: Number>(private val timeline: Timeline<Segment<N>, Numbers
       }
       seg.withNewValue(number)
     }) }
+
+    /**
+     * Converts a list of serialized value segments into an integer [Numbers] profile;
+     * for use with [gov.nasa.ammos.aerie.procedural.timeline.plan.Plan.resource].
+     *
+     * @throws ArithmeticException if any of the segment values are not integers
+     */
+    @JvmStatic fun intDeserializer() = { list: List<Segment<SerializedValue>> -> Numbers(list.map { seg ->
+      val bigDecimal = seg.value.asNumeric().orElseThrow { Exception("value was not numeric: $seg") }
+      seg.withNewValue(bigDecimal.intValueExact())
+    }) }
+
+    /**
+     * Converts a list of serialized value segments into a double [Numbers] profile;
+     * for use with [gov.nasa.ammos.aerie.procedural.timeline.plan.Plan.resource].
+     *
+     * @throws ArithmeticException if any of the segment values are not doubles
+     */
+    @JvmStatic fun doubleDeserializer() = { list: List<Segment<SerializedValue>> -> Numbers(list.map { seg ->
+      val bigDecimal = seg.value.asNumeric().orElseThrow { Exception("value was not numeric: $seg") }
+      seg.withNewValue(bigDecimal.toDouble())
+    }) }
+
+    /**
+     * Converts a list of serialized value segments into a long [Numbers] profile;
+     * for use with [gov.nasa.ammos.aerie.procedural.timeline.plan.Plan.resource].
+     *
+     * @throws ArithmeticException if any of the segment values are not longs
+     */
+    @JvmStatic fun longDeserializer() = { list: List<Segment<SerializedValue>> -> Numbers(list.map { seg ->
+      val bigDecimal = seg.value.asNumeric().orElseThrow { Exception("value was not numeric: $seg") }
+      seg.withNewValue(bigDecimal.longValueExact())
+    }) }
   }
 }
--- a/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/SimulationResults.kt
+++ b/procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/SimulationResults.kt
@@ -6,8 +6,11 @@ import gov.nasa.ammos.aerie.procedural.timeline.collections.Directives
 import gov.nasa.ammos.aerie.procedural.timeline.payloads.Segment
 import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.AnyInstance
 import gov.nasa.ammos.aerie.procedural.timeline.collections.Instances
+import gov.nasa.ammos.aerie.procedural.timeline.collections.profiles.*
 import gov.nasa.ammos.aerie.procedural.timeline.ops.SerialSegmentOps
 import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.AnyDirective
+import gov.nasa.jpl.aerie.merlin.framework.resources.NameableResource
+import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics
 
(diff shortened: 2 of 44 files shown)


procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/SimulationResults.kt
```diff
@@ -25,6 +28,13 @@ interface SimulationResults {
    */
   fun <V: Any, TL: SerialSegmentOps<V, TL>> resource(name: String, deserializer: (List<Segment<SerializedValue>>) -> TL): TL
 
+  fun resource(instance: NameableResource<RealDynamics>) = resource(instance.name, Real.deserializer())
+  fun resource(instance: NameableResource<Boolean>) = resource(instance.name, Booleans.deserializer())
+  fun resource(instance: NameableResource<String>) = resource(instance.name, Strings.deserializer())
+  fun resource(instance: NameableResource<Number>) = resource(instance.name, Numbers.deserializer())
```
mattdailis, line 34: Kotlin newbie question: it seems like there are multiple functions defined here with the same name, and whose arguments have the same type erasure—how is this possible?

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/collections/profiles/Numbers.kt
```diff
@@ -223,5 +223,38 @@ data class Numbers<N: Number>(private val timeline: Timeline<Segment<N>, Numbers
...
       seg.withNewValue(number)
     }) }
+
+    /**
+     * Converts a list of serialized value segments into an integer [Numbers] profile;
+     * for use with [gov.nasa.ammos.aerie.procedural.timeline.plan.Plan.resource].
+     *
+     * @throws ArithmeticException if any of the segment values are not integers
+     */
+    @JvmStatic fun intDeserializer() = { list: List<Segment<SerializedValue>> -> Numbers(list.map { seg ->
```
mattdailis, line 233: The new deserializer methods in this class are unused, so I'm not totally sure how they're related to the rest of this PR

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/util/WithModel.kt
```diff
@@ -0,0 +1,29 @@
...
+  fun model(): M {
+    if (modelSingleton == null) {
+      throw IllegalStateException("modelSingleton was not initialized.")
+    }
+
+    return modelSingleton as M
+  }
+
+  companion object {
+    @JvmStatic var modelSingleton: Any? = null
```
mattdailis, line 27: Just for your awareness/amusement, we do employ a similar pattern on the mission model initialization side: https://github.com/NASA-AMMOS/aerie/blob/a28d46e5c44c2bdb9ce122b224415fdd19e90faa/merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ModelActions.java#L15

The reason we do it there is to enable static methods like `delay` and `spawn` to hook into the context object.

Here, I'm less sure that `model()` must be a static method. Would it be much less ergonomic for it to be a member of `EditablePlan`?

Another aside: I recall @Twisol emphasizing that the context linked above was not so much a singleton as a "dynamically scoped variable". This meant that 1) it was scoped to a particular call stack and would definitely not survive beyond it, and 2) it was scoped to a single thread, limiting its "globalness".

scheduler-worker/src/main/java/gov/nasa/jpl/aerie/scheduler/worker/services/SynchronousSchedulerAgent.java
```diff
@@ -129,6 +130,7 @@ public void schedule(
       ensureRequestIsCurrent(specification, request);
       //create scheduler problem seeded with initial plan
       final var schedulerMissionModel = loadMissionModel(planMetadata);
+      WithModel.setModelSingleton(schedulerMissionModel.missionModel.getModel());
```
mattdailis, line 133: If we stick with the singleton approach here, it may be worth making it an AutoCloseable and putting it in a try-with-resources here. That way we can reliably clear it out once we're done with it, so it doesn't stick around hogging memory or possibly being accidentally reused on a future run.

mattdailis (commented):
No changes requested yet, just some clarifying questions so far. I haven't dug into the classloader business yet, that'll be in a follow-up review

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/util/WithModel.kt
JoelCourtney: I think you're right, it probably could be an instance method on EditablePlan. I'll look into it.

procedural/timeline/src/main/kotlin/gov/nasa/ammos/aerie/procedural/timeline/plan/SimulationResults.kt
JoelCourtney: Wow I didn't even realize how unusual this was when I wrote it. You're right, this would not work in Java. Apparently the JVM is able to differentiate overloads with the same argument type erasure if their return types are different, but Java is not. I don't understand how, but some smart people tried to explain it [here](https://stackoverflow.com/questions/42916801/how-does-erasure-work-in-kotlin)

JoelCourtney: Personally even though I don't really understand it, I'm satisfied because the test cases work in Java. And the type inference works correctly, somehow (the test cases use `var`, and it is still able to infer the type).

I guess this means that Java's ability to resolve overloads is more flexible/powerful than what it actually allows you to declare. Which is kinda nuts.

PR #1719, opened 2025-08-04
Feat/workspace improvements admin secret

mattdailis:
This PR exists to make it easier to see the changes I'm proposing to #1711 

Changes include:
- Have the workspaces server check for the presence of a hasura admin secret header as an alternative to JWT authentication
- Minor refactor to the migration python script to avoid calling apply_after in the Hasura constructor
- Update the migration python script to provide a user id along with the admin secret header


deployment/aerie_db_migration.py
```diff
@@ -349,6 +349,12 @@ def __apply_workspaces_migration__(self) -> bool:
     if results.pop(0)[0] != 'id':
       exit_with_error("Error while fetching user sequences from the database.")
 
+    # admin headers for making requests to the workspace service without a JWT
+    workspace_service_headers = {
+      "x-hasura-admin-secret": self.admin_secret,
+      "x-hasura-role": "admin",
+      "x-hasura-user-id": "migration-script"
```
Mythicaeda: Please use `Aerie Legacy` instead of `migration-script`, for two reasons:
1) `Aerie Legacy` is already our "migration user", as in, when we want to set a user during a migration, we use `Aerie Legacy`
2) `migration-script` is not a real user in the system and will lead to errors if used in a column that's fk'd to the `user` table. I know that that is not happening in the workspaces migration, but it will trip up anybody who uses these headers for reference in the future.

deployment/aerie_db_migration.py
```diff
@@ -349,6 +349,12 @@ def __apply_workspaces_migration__(self) -> bool:
     if results.pop(0)[0] != 'id':
       exit_with_error("Error while fetching user sequences from the database.")
 
+    # admin headers for making requests to the workspace service without a JWT
+    workspace_service_headers = {
+      "x-hasura-admin-secret": self.admin_secret,
+      "x-hasura-role": "admin",
```
Mythicaeda: Typo: `aerie_admin` is the name of Aerie's admin role, not `admin`:
```suggestion
      "x-hasura-role": "aerie_admin",
```

deployment/aerie_db_migration.py
```diff
@@ -190,8 +190,8 @@ def reload_metadata(self):
     """
     Apply and reload the metadata.
     """
-    self.execute('metadata apply')
-    self.execute('metadata reload')
+    self.execute(f'metadata apply {self.metadata_suffix} {self.command_suffix}')
+    self.execute(f'metadata reload {self.metadata_suffix} {self.command_suffix}')
```
Mythicaeda: Two things:
1) `command_suffix` is already added in `self.execute`
2) `self.execute` has a parameter for passing a string of flags

Correct syntax (based on how I designed the code) is:
```suggestion
    self.execute('metadata apply', self.metadata_suffix)
    self.execute('metadata reload', self.metadata_suffix)
```

Mythicaeda:
Aside from the three quick things with the migration script, only other thing that stands out to me is that the kubernetes file `aerie-workspaces-deployment.yaml` wasn't given the new env var (and I just noticed that the kubernetes files for workspaces need to be renamed).

dandelany:
Thanks @Mythicaeda 
* Your changes above have been addressed
* After testing @mattdailis & I made two usability improvements:
  - we changed `.seq.user` to `.seq` as the default extension for migrated sequences - we don't know they are SeqN necessarily, but this gives a better user experience for existing users by showing them as a sequence file
  - 99% of the time, sequence names will be unique, so we updated the migration logic to have them use their existing filenames *unless* they collide, and only then add `_{id}` to the file to disambiguate

We'll re-test with this branch, lmk if you have any questions.

Mythicaeda (approved):
Once this passes testing (including adding example data that will force the conflict resolution code to trigger), this LGTM!

dandelany:
OK, i have tested this branch pretty thoroughly now on my local machine and everything is working as expected as of [19e1b1a](https://github.com/NASA-AMMOS/aerie/pull/1719/commits/19e1b1a5c692ba69b0d64186c9d562a98c249671). I tested the migration by starting with an older version and upgrading + migrating to confirm file creation works as expected - both with a minimal test dataset and with a dump of the database from `aerie-dev`. Merging this :tada:

PR #1729, opened 2025-08-12
Wrapped exception causes in an extensions object so they flow through…

### LINKED ISSUE #1717: Exception Serializers don't conform to Hasura expectations
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

3.5.1

### Describe the bug

Error messages from Merlin bindings include the "message" but not the "cause" field of the serialized exception. In the below screenshot, you can see "input mismatch exception", but what you can't see is "plan with id 195 has not yet been simulated at its current revision"

Image

This is (as far as I can tell) the line of code throwing the exception: https://github.com/NASA-AMMOS/aerie/blob/e03f0965632804f8b9f22216eebf5c00ba5e0465/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/ConstraintAction.java#L74

and this is the code serializing the exception: https://github.com/NASA-AMMOS/aerie/blob/e03f0965632804f8b9f22216eebf5c00ba5e0465/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java#L495

Note that the "message" field is what's shown in the Aerie UI, but that the "cause" field is not. I suspect this is because Hasura treats the "message" field specially - https://hasura.io/docs/2.0/actions/action-handlers/#returning-an-error-response

### Reproduction

I hit this scenario when running constraints on an out-of-date plan

### Logs

```shell

```

### System Info

```shell
Firefox, Mac M2, aerie-dev
```

### Severity

Moderate

cohansen:
… hasura actions correctly

* **Tickets addressed:** Closes #1717 
* **Review:** By commit

## Description
Our exception serializer was just appending `cause` which wasn't flowing through the Hasura action. Any additional data that needs to be preserved in the response needs to be inside of an `extentions` object. 

https://hasura.io/docs/2.0/actions/action-handlers/#returning-an-error-response

## Verification
Manual testing.

## Documentation
NA

## Future work
Any other requests that need to pass additional data to the front-end and go through Hasura actions need to wrap that additional data in `extensions`.


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
```diff
@@ -493,14 +493,26 @@ public static JsonValue serializeNoSuchActivityTypeException(final MissionModelS
   public static JsonValue serializeInputMismatchException(final InputMismatchException ex) {
     return Json.createObjectBuilder()
                .add("message", "input mismatch exception")
-               .add("cause", ex.getMessage())
+               .add("extensions", serializeCauseAsExtension(ex.getMessage()))
```
goetzrrGit, line 496: Is this the only place where we are handling exceptions via a Hasura action? I'm curious if we can apply this same error-handling pattern to other responses.

cohansen: Yes I'm sure there are other cases where we might want the `cause` field to bubble up. I just added it to the 2 known spots for now.

goetzrrGit (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/ResponseSerializers.java
goetzrrGit: Got it.

dandelany:
fixed e2e tests, merging

PR #1734, opened 2025-09-03
Fix: Make Workspaces Extension Matching Ignore Case

Mythicaeda:
* **Review:** By commit  

## Description

Changes the `Map` generated from the `GetExtensionMappings` from a HashMap to a case-insensitive TreeMap. This fixes the bug where a file with a differently capitalized extension is treated as having an "unknown" file extension. (ie `seqN.txt` vs `seqn.txt`)

Also provides a value to `PUBLIC_WORKSPACE_CLIENT_URL` to all of the backend's docker compose files (needed or else the UI can't find the workspace server when developing)

## Verification

I made a workspace, and uploaded:
1) a file with no extension (`ExampleFile`) (having no extension, it will return as an "unknown")
2) a file with an extension that matched exactly with the string in the db (`ExampleFile.seqN.txt`)
3) a file with an extension that had a different capitalization than the string in the db (`ExampleFile copy.seqn.txt`)

On this branch, both 2 and 3 are correctly considered to be sequencing files

On `develop`, 3 is considered to be an "text" file type, since it fails to match `.seqn.txt` but succeeds on matching `.txt`

Image of this branch:
bugfixbranch

Image of develop:
devbranch

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -193,6 +193,7 @@ services:
       PUBLIC_HASURA_CLIENT_URL: http://localhost:8080/v1/graphql
       PUBLIC_HASURA_SERVER_URL: http://hasura:8080/v1/graphql
       PUBLIC_HASURA_WEB_SOCKET_URL: ws://localhost:8080/v1/graphql
+      PUBLIC_WORKSPACE_CLIENT_URL: http://localhost:28000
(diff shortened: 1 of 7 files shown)


joshhaug:
Changes lgtm, pulling to test now

joshhaug (approved):

PR #1736, opened 2025-09-04
Small action server test improvements

dandelany:
Two small fixes to action server:

* Updated the logic for replacing secrets - if this code was called with empty values for any secrets, it would add "*****" between every character of the logs. In general we should just ignore empty secrets and bypass the "censorship" logic
* @Mythicaeda noticed that the action server test outputs were not being written to the CI artifact like the other test reports were - updated to write to a file and include it in the artifact.

### DIFF
--- a/action-server/package.json
+++ b/action-server/package.json
@@ -8,7 +8,7 @@
     "clean": "rm -rf build",
     "start": "npm run clean; npm run build-quiet; node build/app.js",
     "dev": "npm run clean; npm run build; nodemon --watch  src/**/*  -e ts --exec 'node --inspect=0.0.0.0:9229 --require ts-node/register build/app.js'",
-    "test": "node -r ts-node/register tests/*.test.ts",
+    "test": "bash -c 'node -r ts-node/register tests/*.test.ts | tee action-server-test-report.log; exit ${PIPESTATUS[0]}'",
     "lint": "eslint src/**/*.ts",
     "reformat": "npx prettier --config ./.prettierrc.json -w ./src"
   },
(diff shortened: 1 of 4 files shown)


action-server/package.json
```diff
@@ -8,7 +8,7 @@
     "clean": "rm -rf build",
     "start": "npm run clean; npm run build-quiet; node build/app.js",
     "dev": "npm run clean; npm run build; nodemon --watch  src/**/*  -e ts --exec 'node --inspect=0.0.0.0:9229 --require ts-node/register build/app.js'",
-    "test": "node -r ts-node/register tests/*.test.ts",
+    "test": "bash -c 'node -r ts-node/register tests/*.test.ts | tee action-server-test-report.log; exit ${PIPESTATUS[0]}'",
```
Mythicaeda, line 11: Oh, is `exit PIPESTATUS[0]` how we preserve the exit status from the tests? Neat! 

Quick clarifying question:

1) Are the commands wrapped in a `bash -c` because of a quirk of npm?

Mythicaeda (approved):
LGTM, I just have the one question. I am seeing the test artifact in the uploaded `e2eTest` results!

action-server/package.json
dandelany: > is exit PIPESTATUS[0] how we preserve the exit status 

yup!

> wrapped in a bash -c because of a quirk of npm?

it's because PIPESTATUS is a thing in bash but not *all* shells, so this makes sure it uses bash. it's a little funky but at least keeps it a one-liner.

Issue #1737, opened 2025-09-11
@WithMappers does not work in procedural scheduling


adrienmaillard:
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

develop

### Describe the bug

I was trying to add a record parameter to an `e2e.procedural.scheduling` scheduling procedure to test a new feature. I added a custom value mapper class for that new record parameter and added it with a `@WithMappers` annotation in the package-info.java 

When trying to build e2e tests, I got an error 
`AllMappers.value() is defined in an inaccessible class or interface@WithMappers(BasicValueMappers.class)`

Then, I removed this new record parameter and tried removing both `@WithMappers` annotations from the `package-info.java` and it compiled. This means that the value mappers for the procedure have been found somewhere else and that the annotations are not taken into account.

### Reproduction

In `e2e.procedural.scheduling` `package.info.java`, remove the `@WithMappers` annotation and try compiling. It will compile.

### Logs

```shell

```

### System Info

```shell
Mac, Intellij
```

### Severity

Moderate

JosephVolosin:
Also seeing this through a Gradle error when compiling after trying to use a custom mapper. FWIW the mapper _does_ work in the context of the model, just not in procedural scheduling. Same `is defined in an inaccessible class or interface` error.

PR #1738, opened 2025-09-12
Add item ordering in record value mapper to allow ordering of activity/scheduling goal parameters

adrienmaillard:
* **Tickets addressed:** Yet to be written
* **Review:** By commit  

## Description

Metadata containing item order is now attached to every `RecordValueMapper` so it can be used by the UI to display parameters in the order it has been defined in the scheduling procedures.

Changes in the UI are on branch https://github.com/NASA-AMMOS/aerie-ui/tree/ordering-scheduling-goal-params

<a id="verification"></a>
## Verification

With tests in `RecordValueMapperTests`. I also performed an end to end test with the UI code and activities/procedural goals. For ordering at the top level, I could use procedural goals but because of issue https://github.com/NASA-AMMOS/aerie/issues/1737 that I have found while developing, I had to test the nested ordering with activity parameters on the bananation model. 

## Documentation

Docs about parameters of mission models/procedural goals and constraints will have to mention that the way of ordering parameters is to go through records and use the default record value mapper. Custom value mappers for records will not produce the ordering.

## Future work

None.


mattdailis (approved):
Looks good to me based on reading the code and manual testing

dandelany:
Tested with Matt, merging this + https://github.com/NASA-AMMOS/aerie-ui/pull/1773 - Thanks!

PR #1742, opened 2025-09-22
Track file_extension_content_type table in Hasura

Mythicaeda:
* **Review:** By commit  

## Description

While the metadata for this table was written correctly, it was not added to the `tables.yaml` file.
This PR adds it to that file.

It additionally updates the default values of `AERIE_DATABASE_URL` in our docker composes to include the `ui` schema, which is needed in order to insert or update `file_extension_content_type` (because Hasura doesn't schema-prefix types).

## Verification

I went into the Hasura console and saw that table was tracked. I also confirmed that there were no other unexpected untracked tables across the DB.

I also tested that the following mutation worked:
```GQL
mutation UpdateFileExtensionMapping {
  update_file_extension_content_type(where: {file_extension: {_eq: ".seq.json"}}, _set: {content_type: "JSON"}) {
        affected_rows
  }
}
```

### DIFF
--- a/deployment/docker-compose.yml
+++ b/deployment/docker-compose.yml
@@ -203,7 +203,7 @@ services:
     container_name: hasura
     depends_on: ["postgres"]
     environment:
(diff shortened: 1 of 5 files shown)


mattdailis (approved):
Looks reasonable to me. @dandelany the change to `AERIE_DATABASE_URL` will need to be called out in release notes since many users manage their own docker compose files

PR #1744, opened 2025-09-24
Default parameters for procedural constraints/scheduling goals 

adrienmaillard:
* **Tickets addressed:** ?
* **Review:** By commit  

## Description

Copy/paste what has been done for the mission model annotations. There's one less type of default parameters as we allow only records when writing procedural goals/constraints.

Same for the endpoints for getting effective arguments. Very similar. 

## Verification

New tests. I noticed that procedural constraints did not have tests.

## Documentation

Documentation about the new annotations and endpoints need to be added 

## Future work

UI work to use the effective arguments endpoints

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -102,6 +102,18 @@ type Query {
 }
 
 type Query {
+  getConstraintProcedureEffectiveArgumentsBulk(
+    arguments: [ProcedureEffectiveArgumentsInput!]!
+  ): [ProcedureEffectiveArgumentsResponse!]!
+}
+
+type Query {
+  getSchedulingProcedureEffectiveArgumentsBulk(
+    arguments: [ProcedureEffectiveArgumentsInput!]!
+  ): [ProcedureEffectiveArgumentsResponse!]!
+}
+
+type Query {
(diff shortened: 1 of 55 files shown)


dandelany (changes requested):
Tested with @mattdailis & seems to be working correctly - though we still need UI support to show the defaults correctly. This would be safe to merge without UI though, as a hidden feature until UI is implemented.

We found a couple small things in the e2e tests which we'd like to fix before release:
* `TemplateDefaultsTests.java` & `TemplateParametersTest.java` seem to be testing the same procedure the same way - one of these should probably test DumbRecurrenceGoal instead, & should fail if the required arg is not provided in that case
* `BasicConstraintTests.java` - less important, but we could check violations or arguments on constraint run to confirm args were passed - currently it just checks that the constraint ran successfully
		
@mattdailis will look into these. thanks!

mattdailis:
> @mattdailis will look into these. thanks!

Done :+1: See last two commits: " Test both versions of DumbRecurrenceGoal" and "Add assertions to BasicConstraintTests".

dandelany (approved):
Thanks @mattdailis for updating tests - this is ready to merge.

For posterity - this still needs UI support, but we decided to merge the backend for now & merge the UI portion separately, since this doesn't break anything & works on the backend.

Issue #1732, opened 2025-08-28
Implement common HTTP error message formats for all Aerie services


dandelany:
## Background

@AaronPlave @duranb and I have been working on improving error messaging across Aerie. One major impediment to this is the fact that there is no common standard format/structure for errors returned from HTTP services - some just respond with a string, others use JSON but with different fields, etc.

## Proposal

We propose that all HTTP endpoints on Aerie services (with **some exceptions below**) should respond in the following way when they encounter an error/fatal problem:

* They should respond with an **error status code** (4xx or 5xx, not 200)
* The response body should contain a **single JSON object** with the following properties:
  - **required fields:**
    - `type`: very short, semi-human-readable string representing the category/class/type of error, in all caps and underscores, eg. `“INVALID_SIMULATION_ID”`
    - `message`: short (1-2 sentences) human-readable string explaining the cause of the error
    - `timestamp`: ISO 8601 UTC string timestamp at the time the error happened
  - **optional fields:**
    - `service`: optional string identifying the backend service that threw the error
    - `cause`: longer human-readable string, explaining detailed cause of error & any recommendations to fix
    - `trace`: stack trace of error (in any language), to be printed in monospace font in the UI & collapsible
    - `data`: optional unstructured data object with any additional useful error data

### Example
```json
{
  "type": "INVALID_ACTIVITY_PARAMETER_TYPE",
  "message": "The value \"seven\" is not a valid value for activity parameter \"dataRate\" (expected number)"
  "timestamp": "2025-08-28T00:15:03.678Z"
  "service": "merlin-worker",
  "cause": "Activity parameter values must match the type assigned to them in the mission model, defined by ...."
  "trace": "Exception in thread \"main\" java.lang.NullPointerException: Cannot invoke \"String.length()\" because \"str\" is null\n    at com.example.MyClass.myMethod(MyClass.java:10)\n    at com.example.MyClass.main(MyClass.java:5)"
  "data": {"value": "seven"}
}
```

#### Endpoints accessed by Hasura Actions
A slightly different error format is required for any Aerie service endpoints which are **accessed by Hasura Actions** (as opposed to directly via HTTP). This is because [Hasura does its own thing to wrap errors](https://hasura.io/docs/2.0/actions/action-handlers/#returning-an-error-response), and when doing so it *ignores properties other than 'message' and 'extensions'*. Therefore we propose for these endpoints:

* responses should return a JSON object with the format: `{message: "...", {extensions: ...}}`
[...]

PR #1757, opened 2025-10-22
Add Workspace Permissions

### LINKED ISSUE #1733: Workspace service Permissions for owner/collaborators - backend
## Background
Currently, the Workspace service only has basic **role-based permissions,** users either have write access to **all** workspaces, or none, depending on their user role. Users would like to have more fine-grained permissions than this, to enable giving particular users access to certain workspaces, without needing to give them an elevated role. This was discussed in a previous discussion here (Permissions section): https://github.com/NASA-AMMOS/aerie-team/discussions/72

The Postgres schema already supports ownership and collaborator membership (`workspace_collaborators`), but the Workspace service is not yet wired to use this data. We also don't yet have a UI for setting owners/collaborators, but that will be covered in another ticket - this is backend only.

## Description

Update Workspace Service to check the user's `userId` (from auth token) for membership in the accessed workspace's `collaborators` or `owner`, and enforce the proper permissions/privileges for that user. Permission specs were discussed in https://github.com/NASA-AMMOS/aerie-team/discussions/72, copied here with slight reorganization:

 * List workspace contents - Viewer role
 * Create Workspace (including picking parcel) - User role
 * Delete or rename workspace - Owner 
 * Changing parcel that is assigned to a workspace - Owner 
 * CRUD files and directories - Owner & collaborators 
 * Upload actions to workspace - Owner & collaborators 
 * Edit actions settings - Owner & collaborators 
 * Run actions - Owner & collaborators 
 * Remove actions - Owner & collaborators 
 * Setting filter - Owner 
 * Updating list of collaborators - Owner and collaborators 
 * Updating owner - Owner 

Some of the above may require changes in action server too, which falls under this ticket, but may be a separate PR.

## Implementation plan
Implement checks in Workspace service : 
- Update shared Java `PermissionsService` code to support checking a user's membership role for a given workspace (Owner, Collaborator, or None).  
- Update Workspace Service to import and call `PermissionsService`, and enforce the permissions for owner/collaborator roles outlined above.
- Ideally: restructure call to `authorize` in the service's `before` method to reuse the user & not call `authorize` twice

Mythicaeda:
* **Tickets addressed:** Closes #1733 
* **Review:** By commit  

## Description

Note: users using their own version of `docker-compose` will now need to specify `HASURA_GRAPHQL_URL` on the Workspace server. This is because the Permissions service uses GQL to get permissions information.

Adds role-based permissions to the Workspaces Server. Now every endpoint properly checks that the user has permission to run the specific action on the Workspace, as opposed to before, where any user could perform any action so long as they had a valid JWT or used the Hasura Admin Secret.

The role-based permissions are stored in the `user_role_permissions` table, in a new column `workspace_permissions`. The set of permissions are:
```
'create_workspace', 
'delete_workspace',
'list_workspace_contents', 
'delete_file_directory', 
'read_file_directory', 
'write_file_directory'
```

Aside from `create_workspace`, which is a coarse-grained permission and must only either be unset or set to `NO_CHECK`, the remaining permissions are fine-grained. They can be configured to one of the following values:

```
'NO_CHECK', 
'OWNER',
'COLLABORATOR',
'OWNER_COLLABORATOR'
```

The code that validates the `permissions` columns for the `user_role_permissions` table has been split out into several functions to aid with readability, but the baseline behavior for `action_permissions` and `function_permissions` hasn't changed.

Additionally, this PR updates the workspace server error responses to be in line with the schema proposed in #1732.

## Verification

- Adds DB tests for the new permissions column.
- Adds E2E tests for the new authorization behavior
- Outlines but does not implement a more complete Workspace Bindings test suite. 

## Documentation

Our custom role docs need to be updated with information about the new column, and what keys and values it accepts.

## Future work

- Complete E2E tests
- Create API docs
- Refactor PermissionsService so that it uses a DB connection instead of a GQL connection

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PermissionsTest.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/PermissionsTest.java
@@ -8,17 +8,23 @@ import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import java.io.IOException;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@SuppressWarnings("SqlSourceToSinkFlow")
 public class PermissionsTest {
   private static final String testRole = "testRole";
   private enum FunctionPermissionKey {
@@ -52,6 +58,7 @@ public class PermissionsTest {
     PLAN_OWNER_SOURCE, PLAN_COLLABORATOR_SOURCE, PLAN_OWNER_COLLABORATOR_SOURCE,
     PLAN_OWNER_TARGET, PLAN_COLLABORATOR_TARGET, PLAN_OWNER_COLLABORATOR_TARGET
   }
+  private enum WorkspacePermission { NO_CHECK, OWNER, COLLABORATOR, OWNER_COLLABORATOR }
 
   private DatabaseTestHelper helper;
   private MerlinDatabaseTestHelper merlinHelper;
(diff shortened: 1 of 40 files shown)


mattdailis (approved):

dandelany (approved):
Reviewed and tested with Matt, made a few small changes:
* Updated the action_run hasura yaml file to enforce permissions for action runs - only workspace owners and collaborators are able to run actions within a given workspace.
* Updated the error timestamp format in `FormattedError` to use date+time (instant) instead of date.
Looks good otherwise, thanks!

Mythicaeda:
I'm fine with both changes, however I believe the initial choice of using `BASIC_ISO_DATE` was done to match the ticket. Does `ISO_INSTANT` also meet that format, but with more precision?

mattdailis:
> I'm fine with both changes, however I believe the initial choice of using `BASIC_ISO_DATE` was done to match the ticket. Does `ISO_INSTANT` also meet that format, but with more precision?

During our manual testing we noticed that the timestamp didn't include hours, minutes, or seconds (it only included the date, not the time). We figured that wasn't granular enough. Which ticket are you referring to? Maybe we should add a note there as well

Mythicaeda:
> During our manual testing we noticed that the timestamp didn't include hours, minutes, or seconds (it only included the date, not the time). We figured that wasn't granular enough. Which ticket are you referring to? Maybe we should add a note there as well

This ticket: https://github.com/NASA-AMMOS/aerie/issues/1732

And, ah, I must not have checked the date thoroughly then, if it was missing the time part.

PR #1760, opened 2025-10-30
Create Bulk Endpoint for Workspaces

### LINKED ISSUE #1749: Support bulk file operations in Workspace Service
## Background

We plan to implement https://github.com/NASA-AMMOS/aerie-ui/issues/1791 to support selecting **multiple files** in the Sequence Editor workspace. The context menu items **move**, **copy**, **delete**, and **move to workspace** must work on multiple selected items at once. Workspace Service is the backend for these operations, but some endpoints are single-path only, so we need bulk file support.

## Goals

* Accept multiple paths per request for **move, copy, delete, and move-to-workspace**
* Support **file** paths, **directory** paths, or a mix of both
* Predictable behavior with clear partial-failure reporting.
* Be consistent with current **overwrite protection** behavior. Don't overwrite files **unless** a "force" parameter is passed
* Use **"best effort"** approach in case of partial failures & attempt to operate on as many files as possible
* Report partial failures clearly so the UI can surface them to the user
* Ideally keep old single-path requests working for backward compat.

### Partial failure cases

Since we are performing multiple file operations in one request, some of them may fail & others succeed. Ideally (IMO) these would be more **atomic** operations, and either completely succeed or completely fail, but this is not easy to implement right now & comes with usability challenges. Instead we plan to implement "best effort" operations, eg. move as many files as possible, but also report any failures.

In general, if the *entire operation* fails, the server should respond with an appropriate error code. If there is a partial failure, it should respond with the status code [207 "Multi-Status Response"](https://www.rfc-editor.org/rfc/rfc4918#section-13) & include details about which parts failed and which succeeded. (This code is a new part of the proposal & open to feedback)

**Some** of these errors may be the result of our rules refusing to overwrite an existing file by the same name - these cases should be clearly communicated, as the UI will present the user with the option to "proceed anyway", which will result in another request to the same endpoint with the "force" parameter set to `true`.

### Concurrent file operations

Some of these bulk file operations may be dangerous if they are performed concurrently with other operations on the same files. We need to consider these cases and have some form of protection against dangerous/inconsistent behavior due to concurrency. This may mean having global or path-based file locks to prevent concurrent operations. Requirements for this need to be fleshed out further. It may be useful to refer to the [WebDAV protocol](https://en.wikipedia.org/wiki/WebDAV), which has similar goals, for inspiration on handling these cases.

Mythicaeda:
* **Tickets addressed:** Closes #1749 
* **Review:** By commit  

## Description

**Note: depends on #1757** 

Adds in bulk endpoints for file/directory level `PUT`, `POST`, and `DELETE` requests.

Updates single-item `PUT` to accept `folder` as a valid value for `type`. It is treated as equivalent to `directory`.

Refactors `handleMove` and `handleCopy`. They now take in all the values they would extract from `context` and output a `results` object that contains the HTTP status code and response object. This allows the bulk version of `POST` to call these methods in a loop and create its response array from the output.

## Verification

Basic E2E Tests were created. Edge case E2E Tests need to be created

## Documentation

No docs were invalidated.

## Future work

- Create API docs for Workspace Server
- Finish Workspace Binding E2E tests
- Concurrent File Operation Protections

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
@@ -2485,6 +2485,7 @@ public class BindingsTests {
       class BulkPost {
         private int workspaceId;
         private int otherWorkspaceId;
+        private final Path destinationPath = Path.of("./destination_dir");
 
         @BeforeEach
         void beforeEach() throws IOException {
@@ -2502,9 +2503,11 @@ public class BindingsTests {
               new BulkPutItem.FileBulkPutItem("top_dir/sub_file.txt", "file within a directory"),
               new BulkPutItem.FileBulkPutItem("other_dir/other_file.txt", "another file within a directory"),
               new BulkPutItem.FileBulkPutItem("top_dir/nested_dir/nested_file.txt", "file within a nested directory"),
-              new BulkPutItem.DirectoryBulkPutItem("destination_dir")
+              new BulkPutItem.DirectoryBulkPutItem("destination_dir"),
+              new BulkPutItem.DirectoryBulkPutItem("destination")
           );
           wsServer.bulkPut(ownerToken, workspaceId, wsContents);
+          wsServer.bulkPut(ownerToken, otherWorkspaceId, List.of(new BulkPutItem.DirectoryBulkPutItem("destination_dir")));
         }
 
         @AfterEach
@@ -2524,7 +2527,6 @@ public class BindingsTests {
         @ParameterizedTest
         @MethodSource("bulkPostBasicCasesArgs")
         void bulkMoveSameWorkspaceBasicCases(List<Path> inputs) {
-          final var destinationPath = Path.of("./destination_dir");
           final var resp = wsServer.bulkMove(
               ownerToken,
               workspaceId,
@@ -2575,7 +2577,6 @@ public class BindingsTests {
         @ParameterizedTest
         @MethodSource("bulkPostBasicCasesArgs")
         void bulkMoveBtwnWorkspaceBasicCases(List<Path> inputs) {
-          final var destinationPath = Path.of(".");
           final var resp = wsServer.bulkMove(
               ownerToken,
               workspaceId,
@@ -2626,7 +2627,6 @@ public class BindingsTests {
         @ParameterizedTest
         @MethodSource("bulkPostBasicCasesArgs")
         void bulkCopySameWorkspaceBasicCases(List<Path> inputs) {
-          final var destinationPath = Path.of("./destination_dir");
           final var resp = wsServer.bulkCopy(
               ownerToken,
               workspaceId,
@@ -2679,7 +2679,6 @@ public class BindingsTests {
         @ParameterizedTest
         @MethodSource("bulkPostBasicCasesArgs")
         void bulkCopyBtwnWorkspaceBasicCases(List<Path> inputs) {
-          final var destinationPath = Path.of(".");
           final var resp = wsServer.bulkCopy(
               ownerToken,
               workspaceId,
@@ -2729,6 +2728,7 @@ public class BindingsTests {
           final var nestedFileInput = Path.of("other_dir/other_file.txt");
           final var topDirInput = Path.of("top_dir");
           final var nestedDirInput = Path.of("other_dir/nested_dir");
+          final var siblingNameInput = Path.of("destination");
 
           return Stream.of(
               Arguments.arguments(named("Top Level File Single Bulk POST", List.of(topFileInput))),
@@ -2739,7 +2739,8 @@ public class BindingsTests {
               Arguments.arguments(named("Multiple Directories Bulk POST", List.of(topDirInput, nestedDirInput))),
               Arguments.arguments(named(
                   "Mixed Files and Directories Bulk POST",
-                  List.of(topFileInput, nestedFileInput, nestedDirInput, topDirInput)))
+                  List.of(topFileInput, nestedFileInput, nestedDirInput, topDirInput))),
+              Arguments.arguments(named("Sibling Directory with Subset Name", List.of(siblingNameInput)))
           );
         }
 
@@ -2855,7 +2856,7 @@ public class BindingsTests {
           /**
            * Generate arguments to test basic upload cases.
            */
-          private static Stream<Arguments> bulkPostBasicCasesArgs() {
+          private static Stream<Arguments> overwriteCasesArgs() {
             final var topFile = new ConflictItem(
                 Path.of("top_file.txt"),
                 "top level file",
@@ -2892,7 +2893,7 @@ public class BindingsTests {
            * With overwrite unset, a conflict is returned. This tests for both within and between workspaces
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkMoveOverwriteUnset(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
@@ -2945,7 +2946,7 @@ public class BindingsTests {
            * With overwrite set to false, a conflict is returned. This tests for both within and between workspaces
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkMoveOverwriteFalse(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
@@ -2998,7 +2999,7 @@ public class BindingsTests {
            * With overwrite set to true, no conflict occurs. This tests for moving within a workspace
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkMoveOverwriteTrueWithinWS(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
@@ -3037,7 +3038,7 @@ public class BindingsTests {
            * With overwrite set to true, no conflict occurs. This tests for moving between workspaces
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkMoveOverwriteTrueBetweenWS(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
@@ -3075,7 +3076,7 @@ public class BindingsTests {
            * With overwrite unset, a conflict is returned. This tests for both within and between workspaces
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkCopyOverwriteUnset(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
@@ -3128,7 +3129,7 @@ public class BindingsTests {
            * With overwrite set to false, a conflict is returned. This tests for both within and between workspaces
            */
           @ParameterizedTest
-          @MethodSource("bulkPostBasicCasesArgs")
+          @MethodSource("overwriteCasesArgs")
           void bulkCopyOverwriteFalse(List<ConflictItem> inputs) {
             final var paths = inputs.stream().map(i -> i.originalPath).toList();
             final var destination = Path.of("./destination_dir");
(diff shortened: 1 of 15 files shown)


workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -108,8 +127,17 @@ public void apply(final Javalin javalin) {
...
     javalin.exception(NumberFormatException.class,
                       (ex, ctx) -> ctx.status(400).json(new FormattedError(ex)));
+    javalin.exception(SecurityException.class,
+                      (ex, ctx) -> ctx.status(500).json(new FormattedError(ex)));
+    javalin.exception(Exception.class, (ex, ctx) -> {
+      // Catch-all for unexpected issues
+      logger.error("Unexpected error processing workspace request", ex);
+      final var message = ex.getMessage() != null ? ex.getMessage() : "Unknown error.";
+      ctx.status(500).json(new FormattedError("UNKNOWN_ERROR", message, ex));
+    });
```
dandelany, line 149: I did some testing on error cases and was able to hit this catch-all a couple times - good addition 👍 A few small recommendations:
* It's fine to not log noisy user problems like 404's, but we should add `logger.warn` or `logger.error` to handlers which indicate an "us problem", to help us debug what happened. Eg. at least `IOException` & `SQLException` currently return 500 but don't log.  Maybe `SecurityException` too since it either indicates a malicious user or a script gone wrong.
* Minor, but Javalin will sometimes throw specific HTTP exceptions which have already set HTTP status codes, and the catch-all handler will overwrite these as 500. I couldn't come up with a repro case without commenting out other handlers, but I think something like this would be a good defensive thing to put at the top of the catch-all handler:
```
  if (ex instanceof HttpResponseException hre) {
    ctx.status(hre.getStatus())
       .json(new FormattedError(hre));
    return;
  }
  ```

File: workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java [hunk 2]
```diff
@@ -144,6 +172,37 @@ private JWTService.UserSession authorize(Context context) {
     }
   }
 
+  /**
+   * Check that the request meets the permissions to perform the given action on the workspace.
+   * If it does not, format the context into an appropriate error state.
+   * @return true, if the user passes the permissions check. false otherwise
+   */
+  private boolean checkPermissions(Context context, int workspaceId, WorkspaceAction action) {
```
dandelany, line 192: No action needed, but i wonder if a better pattern here could be to `throw` on failed permission checks rather than mutating context and returning a boolean. That would centralize error handling, keep this helper more pure, and avoid callers needing to remember to `return` after the `false` case.

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -608,40 +654,464 @@ private void delete(Context context) throws NoSuchWorkspaceException, IOExceptio
...
+        // Create file
+        try {
+          if (workspaceService.saveFile(workspaceId, item.path(), file)) {
+            response.add("status", 200)
+                    .add("response", "File " + item.path().getFileName() + " uploaded to " + item.path());
+          } else {
+            response.add("status", 500)
+                    .add("response", new FormattedError("Could not save file.").toJson());
+          }
+        } catch (IOException ioe) {
```
dandelany: There is a decent amount of duplicated logic between this function and the single-upload `put` handler... Might be nice to abstract them into a single `handleUpload` like you've done for `handleCopy` and `handleMove`. Or if that feels too big - it would be good to compare this side-by-side with that function and make sure they handle the same cases - eg. this handles `IOException` and `"Unsupported item upload type"` but I don't think the single `put` does.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
```diff
@@ -1484,7 +1486,7 @@ void nonEmptyWorkspaceWithDepth() {
     }
 
     /**
-     * Tests for the /ws/{workspaceId}/<filePath> routes.
+     * Tests for the /ws/{workspaceId}/<path> routes.
```
dandelany, line 1489: Noticed above on line 1276 (couldn't leave a comment there since not a new change, but figured worth mentioning) - I'm a bit surprised that calling `deleteMockCommandDictionary` *before* `deleteMockParcel` doesn't throw/complain, since I thought the parcel had a foreign key for the command dictionary. If I try to delete a dictionary before deleting its parcel via UI, I get an error.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
```diff
@@ -1898,6 +1900,1677 @@ void conflictedIfDestinationExists() {
       }
     }
 
+    /**
+     * Tests for the /ws/bulk/{workspaceId}/ routes.
+     */
+    @Nested
+    class BulkWSRoutes {
+      @Nested
+      class BulkPut {
```
dandelany, line 1909: Small/nonblocker, but it would help with readability if we could standardize naming conventions and name handlers (& their associated helpers) for what they *do* rather than the HTTP verb. Eg. we say `BulkPut` in some places and `BulkCreate` in others, same with `BulkPost` vs. `BulkMoveCopy`

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -44,6 +47,65 @@ Path resolveSubPath(final Path rootPath, final Path filePath) {
...
+        throw new WorkspaceFileOpException("Path segment '"+ segment+ "' has illegal characters: "+forbiddenMatcher.group());
+      }
+
+      // Check that the segment is not a reserved filenames:
+      final var name = segment.split("\\.")[0];
+      if(Arrays.asList(reservedFilenames).contains(name)){
+        throw new WorkspaceFileOpException("Path segment '"+ segment+ "' contains reserved name: "+name);
+      }
+    }
+  }
```
dandelany, line 124: Thanks for catching these cases! My only nonblocking architectural complaint here is that devs always have to remember to call *both* resolve/validate functions in order to get a safe path to use, ie.
```
final var path = resolveSubPath(repoPath, directoryPath);
validatePath(path);
```
Feels like it would be safer to bake validation into `resolveSubPath` or some parse function which calls both?

dandelany (approved):
Thanks @Mythicaeda ! Looks good, and I appreciate the additional e2e test coverage on these endpoints. All of my local testing worked as expected, added some notes on potential improvements below but none of them are blockers. This is OK to merge, or lmk if you push any changes for my comments & I'll take another look.

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: I made this method a `boolean` that mutates the context to avoid having duplicate `catch` blocks everywhere that calls the method, but I suppose having it throw an exception that contains the status code and formatted error would work (though that wouldn't necessarily remove the need for a return statement, unless you're proposing to have none of the endpoints catch the new exception)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
Mythicaeda: So the reason it works is because `deleteParcel` is effectively doing nothing. The db relationship on `parcel` says that if it's `command_dictionary` is deleted, that delete should cascade and delete the `parcel` as well. I just tested in the UI and this worked as I said. In your UI test, was the parcel in use on a workspace? If so, the workspace would stop the parcel from being deleted, which would stop the cdict from being deleted.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
Mythicaeda: The reason the handlers were named for their HTTP verb is because they're handling ALL requests of that type. For example, we currently support two types of `POST` requests: move and copy. Once we know which of these requests we're handling, we pass it to a method that's named for the specific action its doing (ie `handleMove`). 

I'd prefer to keep the direct handlers named according to what endpoint they're handling and pull out the uploading/deleting code into sub-methods that the handlers call (ie make a `handleUpload` method that `put` and `bulkPut` called).

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: Lemme see about making a separate function to call both. The reason it's not baked into `resolveSubPath` is because we only need to validate the path when we're going to try and write to it (at least, as far as I'm aware), but we need to resolve sub paths way more often (ie checking if a file exists, check whether its a directory, etc).

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
dandelany: Ah that makes sense - yes i was using the parcel in a workspace.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
dandelany: > they're handling ALL requests of that type

But they're not - they're handling all requests of that type *to* a particular endpoint, which is not really described by the function name. Eg. the function just called `post` doesn't tell me anything about what's being posted/where & doesn't scale well to multiple endpoints. `bulkPost` is better but my preference remains something descriptive, eg. `BulkMoveCopy`, it just gives readers more information when looking at the endpoint -> handler bindings. In Express I often reserve the prefix `handle` specifically for direct handlers to distinguish them from the sub-methods/helpers doing the actual work

Either way, we have an inconsistent mix today, so it would be good to land on some convention here (though ultimately minor!):
```
path("/ws/bulk/{workspaceId}", () -> {
  ApiBuilder.put(this::bulkUpload);
  ApiBuilder.post(this::bulkPost);
  ApiBuilder.delete(this::bulkDelete);
});
path("/ws/{workspaceId}/<path>",
     () -> {
       ApiBuilder.get(this::get);
       ApiBuilder.put(this::put);
       ApiBuilder.delete(this::delete);
       ApiBuilder.post(this::post);
});
path("/ws/{workspaceId}", () -> {
  ApiBuilder.get(this::listWorkspaceContents);
  ApiBuilder.delete(this::deleteWorkspace);
});
```

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
dandelany: Yep fair, didn't think of that subtlety. Thanks.

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: As discussed, I'm going to make a separate small PR that puts that exception into the PermissionsService and updates `MerlinBindings`, `SchedulerBindings`, and `WorkspaceBindings` accordingly.

Edit: [here's the ticket for that work](https://github.com/NASA-AMMOS/aerie/issues/1771)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
Mythicaeda: I can rename `put` to `singleItemUpload` and so on, though I'd like to argue in favor of keeping the word `post` in the `post` handlers, as 1) I find `MoveCopy` clunkier and 2) I would not be surprised if we support more types of POST requests as workspaces matures.

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: Added a new commit to address this: [Create method to resolve a path for writing and validate it](https://github.com/NASA-AMMOS/aerie/pull/1760/commits/6290a9ada4fe72f944b19da2e389edd1e74d4d31)

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Added a commit to address this: [Log when a 500 is returned by the server](https://github.com/NASA-AMMOS/aerie/pull/1760/commits/bec83d3c5e1412062802a4aa29d82b539aaa8559)

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: Addressed in this commit: [Extract Duplicate Logic for Upload and Delete](https://github.com/NASA-AMMOS/aerie/pull/1760/commits/9542e744a19135198785d205ba532c4341889623)

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/BindingsTests.java
Mythicaeda: Addressed in this commit, though if you have different names you'd prefer than please say so and I'll update them: [Rename handler methods](https://github.com/NASA-AMMOS/aerie/pull/1760/commits/1a76ccbce6a3a9fb0183383cff1025ef1bd223d1)

dandelany:
👍 reviewed final round of changes with @AaronPlave, this is good to go. Merging.

PR #1762, opened 2025-11-11
Actions improvements - JWT, user info, adaptations in actions

dandelany:
## Description
Improvements to action server to support new features:

* Update `aerie-actions` to support loading adaptations in actions, param descriptions & other features
 - [ ] TODO before merge: [merge & publish](https://github.com/NASA-AMMOS/aerie-actions/pull/13/files) `aerie-actions@1.1.0` and update dependency
* User authentication when passing transient secrets to actions
* Extract JWT from secrets requests & pass it into action context, so actions may send authed requests to Hasura/other apps
* Pass username and action run id to actions, per user feedback
* Fix issue with action runs accidentally overwriting action-server's `process.env` reference by changing how we pass global context down to actions.
* Separate `app.js` from `server.js` for better testability, add test for authenticating user JWT

### DIFF
--- a/action-server/src/config.ts
+++ b/action-server/src/config.ts
@@ -34,7 +34,7 @@ export const configuration = (): Config => {
     ACTION_WORKER_NUM: env.ACTION_WORKER_NUM ?? "1",
     ACTION_MAX_WORKER_NUM: env.ACTION_MAX_WORKER_NUM ?? "1",
     HASURA_GRAPHQL_ADMIN_SECRET: env.HASURA_GRAPHQL_ADMIN_SECRET ?? "",
-    HASURA_GRAPHQL_JWT_SECRET: env.HASURA_GRAPHQL_JWT_SECRET ?? "",
+    HASURA_GRAPHQL_JWT_SECRET: env.HASURA_GRAPHQL_JWT_SECRET ?? "{ \"type\": \"HS256\", \"key\": \"examplekey\" }",
     JWT_ALGORITHMS: parseArray(env.JWT_ALGORITHMS, ['HS256']),
     LOG_FILE: env.LOG_FILE ?? "console",
     LOG_LEVEL: env.LOG_LEVEL ?? "debug",
--- a/action-server/src/config.ts
+++ b/action-server/src/config.ts
@@ -10,7 +10,6 @@ export interface Config {
   ACTION_WORKER_NUM: string;
   ACTION_MAX_WORKER_NUM: string;
   HASURA_GRAPHQL_JWT_SECRET: string;
-  JWT_ALGORITHMS: Algorithm[];
   LOG_FILE: string;
   LOG_LEVEL: string;
   MERLIN_GRAPHQL_URL: string;
@@ -33,7 +32,6 @@ export const configuration = (): Config => {
     ACTION_WORKER_NUM: env.ACTION_WORKER_NUM ?? "1",
     ACTION_MAX_WORKER_NUM: env.ACTION_MAX_WORKER_NUM ?? "1",
     HASURA_GRAPHQL_JWT_SECRET: env.HASURA_GRAPHQL_JWT_SECRET ?? "{ \"type\": \"HS256\", \"key\": \"examplekey\" }",
-    JWT_ALGORITHMS: parseArray(env.JWT_ALGORITHMS, ['HS256']),
     LOG_FILE: env.LOG_FILE ?? "console",
     LOG_LEVEL: env.LOG_LEVEL ?? "debug",
     MERLIN_GRAPHQL_URL: env.MERLIN_GRAPHQL_URL ?? "http://localhost:8080/graphql",
@@ -43,19 +41,3 @@ export const configuration = (): Config => {
     WORKSPACE_BASE_URL: env.WORKSPACE_BASE_URL ?? "http://aerie_workspace:28000",
   };
 };
-
-/**
- * Parse generically typed environment variable into an array.
- * Returns the default value if parse fails.
- */
-function parseArray<T = string>(value: string | undefined, defaultValue: T[]): T[] {
-  if (typeof value === 'string') {
-    try {
-      return JSON.parse(value);
-    } catch (e) {
-      console.error(e);
-      return defaultValue;
-    }
-  }
-  return defaultValue;
-}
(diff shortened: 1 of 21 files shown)


action-server/src/type/types.ts
```diff
@@ -23,10 +23,12 @@ export type ConsoleOutput = {
 
 export type ActionConfig = {
   ACTION_FILE_STORE: string;
+  ACTION_RUN_ID: string;
   SEQUENCING_FILE_STORE: string;
   SECRETS?: Record<string, string> | undefined;
   WORKSPACE_BASE_URL: string;
   HASURA_GRAPHQL_ADMIN_SECRET: string;
+  USERNAME?: string;
```
Mythicaeda, line 30: Given that we're now passing along the token, I wonder if the mandatory `HASURA_GRAPHQL_ADMIN_SECRET` couldn't be replaced with `AUTHORIZATION_HEADERS`, which contains a map of the applicable auth headers (ie `x-hasura-admin-secret`, `Authorization`, `x-hasura-role`)

docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
     depends_on: [ "postgres" ]
     environment:
       HASURA_GRAPHQL_ADMIN_SECRET: "${HASURA_GRAPHQL_ADMIN_SECRET}"
+      HASURA_GRAPHQL_JWT_SECRET: "${HASURA_GRAPHQL_JWT_SECRET}"
```
Mythicaeda, line 10: We discussed this during the code walkthrough, but leaving a comment so I don't forget: don't forget to add this to the other `docker-compose`s (most critically, the one in `deployment`)

action-server/src/config.ts
```diff
@@ -30,6 +34,8 @@ export const configuration = (): Config => {
     ACTION_WORKER_NUM: env.ACTION_WORKER_NUM ?? "1",
     ACTION_MAX_WORKER_NUM: env.ACTION_MAX_WORKER_NUM ?? "1",
     HASURA_GRAPHQL_ADMIN_SECRET: env.HASURA_GRAPHQL_ADMIN_SECRET ?? "",
+    HASURA_GRAPHQL_JWT_SECRET: env.HASURA_GRAPHQL_JWT_SECRET ?? "{ \"type\": \"HS256\", \"key\": \"examplekey\" }",
+    JWT_ALGORITHMS: parseArray(env.JWT_ALGORITHMS, ['HS256']),
```
Mythicaeda: I know this envvar is here to mirror the Gateway, but why do we have a separate `JWT_ALGORITHMS` array instead of using the algorithm specified in the JWT secret? (For context, WS server uses the latter method)

docker-compose.yml
dandelany: Done, thanks 👍

action-server/src/config.ts
dandelany: Yeah good idea, I think this makes more sense & should always work. Updated to use algorithm from the `type` field on `HASURA_GRAPHQL_JWT_SECRET` and throw if not present, and removed `JWT_ALGORITHMS` from configuration.

action-server/src/type/types.ts
dandelany: Yes, good idea! Well..... **maybe** haha.

I went ahead and made this change in [d6a764d](https://github.com/NASA-AMMOS/aerie/pull/1762/commits/d6a764dd10b9f5db3df4f887c11be8a20b318fe8) and tested it locally, but now i'm having some second thoughts. This touches on a core question re: the mental model/philosophy of actions:

Does an action generally operate as the **user who ran it** or as a **special admin user**?  If an action tries to eg. write to a part of the DB that the current user doesn’t have access to, should the current user be able to run it?

Reasons to prefer running actions as **user**:
* More secure by default, follows [Principle of Least Privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege). 
* Consistency, users have the same permissions with actions as they do in the rest of Aerie

Reasons to prefer running actions with **admin secret**:
* Actions have more power & flexibility re: what users can do with them
* Actions are **uploaded by admins only** which already prevents non-admins from writing malicious actions. Actions **could** be considered “the blessed way” to give non-admins specific admin-like powers.
* Traceability/repeatability. If admin secret is used, every action run with the same parameters & settings will produce the same output (if written to be deterministic). Otherwise, the *user who ran the action* is an additional "input" to the action that changes its behavior, so eg. another user could not "re-run" the action and expect exactly the same results

Let me sleep on it and check in with @parkerabercrombie and come back to this question tomorrow...

dandelany:
Thanks @Mythicaeda I addressed all your notes, with one outstanding on-the-fence question i'll make a decision about tomorrow, ready for another look.

Ignore the failing tests for the moment - they were passing and will pass again after I publish a new version of the aerie-actions API with the updated config etc - I'll make sure they pass before merging.

action-server/src/config.ts
Mythicaeda: Obviously this is not part of this PR, but I think we should also apply this change to the Gateway itself.

action-server/src/type/types.ts
parkerabercrombie: For all of the action use case we currently have on Clipper I want the action to run as the user who initiated the action, not as an admin. If i had to pick one, I'd definitely prefer the action to run as the user (and especially to provide the action with the username, so we can attach that username to products generated by the action).

I suppose I can imagine cases where you might want to run as admin for the "blessed way" to run a certain workflow, but I don't have any actual mission need for that at the moment. I don't see any harm is providing a hook in the action API to allow the action code to grab the admin token if it needs to assume the admin role.

dandelany: Thanks for the input @parkerabercrombie - we'll go with this (running with user token) as the default and think about adding some kind of "admin mode" hook in the future if needed.

Mythicaeda (approved):
Presuming tests start passing once the aerie-actions code has been updated, this LGTM

dandelany:
@duranb and I tested this again yesterday with the the UI from https://github.com/NASA-AMMOS/aerie-ui/pull/1817 with no issues. Added a few more mini commits to update to the newly published `aerie-actions` library, fix a CodeQL issue, and disposition vulnerabilities from the security scan - no functional changes. Tests are passing and this is ready to merge :tada:

PR #1763, opened 2025-11-13
Change SimulationResultsWriter to write more directly to file

DavidLegg:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Changes SimulationResultsWriter to write output JSON more directly to the chosen output file (or stdout).
It no longer builds in-memory JsonValue nor String representations of large chunks of the output.
Instead, the resultsGenerator is connected directly to the chosen output stream, and build* subroutines to build in-memory JsonValue objects are replaced by corresponding write* subroutines to write that section of the output directly to file.

When using a ResourceFileStreamer, this means opening only one resource temp file at a time and streaming data from that file to the final output file, never loading the full resource profile into memory.

This makes stateless Aerie capable of handling very large simulation results.

## Verification
This was tested with a Clipper plan that generated a nearly 14GB output file.

## Documentation
N/A - purely a performance-oriented refactor

## Future work
There are potentially more performance gains to be had by changing the output format to one that can be written incrementally. Something like a JSON Lines file where each line is a result, in time order, and a result can be any update from the simulation: activity start or end, event on a topic, or change in resource dynamics.
Such an output file could be written as the simulation runs. Combined with the results-streaming interface that's already in place, and we could write the final output file in parallel with simulation, with no need to run a post-processing step.

### DIFF
--- a/orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
+++ b/orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
@@ -1,37 +1,26 @@
 package gov.nasa.jpl.aerie.orchestration.simulation;
 
-import gov.nasa.jpl.aerie.types.ActivityInstance;
-import gov.nasa.jpl.aerie.types.ActivityInstanceId;
+import gov.nasa.jpl.aerie.json.JsonParser;
 import gov.nasa.jpl.aerie.merlin.driver.SimulationResults;
-import gov.nasa.jpl.aerie.merlin.driver.UnfinishedActivity;
-import gov.nasa.jpl.aerie.merlin.driver.engine.EventRecord;
 import gov.nasa.jpl.aerie.merlin.driver.resources.ResourceProfile;
-import gov.nasa.jpl.aerie.merlin.driver.timeline.EventGraph;
-import gov.nasa.jpl.aerie.merlin.protocol.types.RealDynamics;
-import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
-import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
 
 import javax.json.Json;
-import javax.json.JsonObject;
 import javax.json.JsonReader;
-import javax.json.JsonValue;
 import javax.json.stream.JsonGenerator;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
 import java.io.StringReader;
-import java.io.StringWriter;
+import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.temporal.ChronoUnit;
-import java.util.List;
 import java.util.Map;
-import java.util.concurrent.RecursiveTask;
+
 import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
 import gov.nasa.jpl.aerie.merlin.server.remotes.postgres.EventGraphFlattener;
 import gov.nasa.jpl.aerie.types.Plan;
 import gov.nasa.jpl.aerie.types.Timestamp;
-import org.apache.commons.lang3.tuple.Pair;
-import org.apache.commons.lang3.tuple.Triple;
 
 import static gov.nasa.jpl.aerie.merlin.driver.json.SerializedValueJsonParser.serializedValueP;
 import static gov.nasa.jpl.aerie.merlin.driver.json.ValueSchemaJsonParser.valueSchemaP;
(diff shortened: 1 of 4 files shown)


orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
```diff
@@ -174,333 +82,266 @@ public void writeResults(CanceledListener canceledListener) {
...
-                .add("startOffset", startOffset)
-                .add("duration", act.duration().toString())
-                .add("attributes", serializedValueP.unparse(act.computedAttributes()))
-                .add("arguments", activityArgumentsP.unparse(act.arguments()))
-                .add("startTime", act.start().toString())
-                .add("endTime", endTime);
-
-      // Append to the array builder
-      simulatedActivitiesBuilder.add(actBuilder);
+      resultsGenerator
```
Mythicaeda: Minor, nonblocking: the initial order of fields for directives were chosen deliberately so that all of the ID fields were next to each other at the top of the entry. Could you restore this order?

orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
```diff
@@ -174,333 +82,266 @@ public void writeResults(CanceledListener canceledListener) {
...
+      // We expect RFS made a temp file where each line is a profile segment
+      var resourceTempFile = Path.of(resourceFileStreamer.getFileName(profileName));
+      try (final var stream = Files.lines(resourceTempFile)) {
         stream.forEach(s -> {
           if (!s.isBlank()) {
+            // s is a JSON object for a single segment, write it as a value to the results generator
+            // Sadly, this requires reading the object into a JsonValue, just to write it back out (!)
             try (final JsonReader jr = Json.createReader(new StringReader(s))) {
-              segmentsBuilder.add(jr.readObject());
+              resultsGenerator.write(jr.readValue());
```
Mythicaeda, line 188: Just a comment, no changes: Shame there's no `writeRaw` method like in Jackson's implementation. That said, the reader and JSON Value only exist for the length of the loop, so memory should be fine.

orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
Mythicaeda, line 350: Could you update the JSON Schema to reflect the new pattern for topics and events?

Mythicaeda (commented):
Thanks for improving the memory use! I think we can revisit parallelization/streaming as we go once that starts being an issue with stateless sim.

I'm only holding off on approval because of the schema docs. It's not the best place to have them, I know, but they ought to stay up-to-date while we find them a new home.

Additionally for @dandelany: while it can still only sim, the change to events/topics is the one change that I really wanted done before we "announced" Stateless Aerie/took it out of "beta"/whatever the appropriate phrase is.

orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
DavidLegg: Agreed - I guess it would be hard to have the JSON generator guarantee a consistent internal state without basically parsing the JSON string anyways, but it still feels wasteful here.

orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
DavidLegg: Ah, that makes a lot of sense. Restored!

orchestration-utils/src/main/java/gov/nasa/jpl/aerie/orchestration/simulation/SimulationResultsWriter.java
DavidLegg: Good catch! Should be up to date now.

stateless-aerie/src/test/resources/simpleFooPlanResults.json
```diff
@@ -758,15 +754,14 @@
                     }
                 },
                 "startTime": "2024-07-01T02:27:15.059Z",
-                "endTime": "2024-07-01T02:27:17.059Z"
+                "endTime": "2024-07-01T02:27:17.059Z",
+                "childIds": [
+                ],
+                "directiveId": 4,
+                "parentId": null
```
Mythicaeda, line 761: I'm surprised this isn't causing the tests to fail. Unless `assertEquals` with JSON objects are smart enough to be key-order agnostic?

Mythicaeda (approved):
LGTM! The only question I have left is regarding the reference file for tests, but there's nothing wrong with it so long as the object comparison is in fact key-order-agnostic.

stateless-aerie/src/test/resources/simpleFooPlanResults.json
dandelany: looked into this because I was curious: we assert equality on `JsonObject`s which are `Map`s under the hood, so it is in fact key-order agnostic.

Mythicaeda: Thanks! I poked at this a little as well, so for posterity: javax's JsonObject specifically uses a `LinkedHashMap` under the hood, which, as you said, overrides `.equals` to check in a key-order agnostic way. 

I thought there was one `Map` implementation that _didn't_ do this and would return `false` if the keys were ordered differently. This was because I remembered reviewing a special `assertMapEquals` method (or something similar) somewhere in the Scheduler tests. However, I can't find it on a quick search, so either a) the helper I'm thinking of was for a niche case like comparing a map of `Activities` (which have a unique definition of "equality" in tests), or b) the code I'm thinking of has since been removed or replaced.

PR #1775, opened 2026-01-21
Treat only record components in procedures as parameters

### LINKED ISSUE #1776: Issues with Procedural Goal mapper generation
Discovered with @AaronPlave and @mattdailis while testing procedural goals on the [aerie-orbiter-model](https://github.com/NASA-AMMOS/aerie-orbiter-model) 

1. If you update the model to PlanDev 4.0.0 and try to run ` ./gradlew scheduling:buildAllProcedureJars` it fails with the following:
```
    /Users/delany/Documents/jpl/repos/aerie-orbiter-model/scheduling/build/generated/sources/annotationProcessor/java/main/scheduling/generated/procedures/AddPeriapsesMapper.java:95: error: cannot find symbol
      arguments.put("VERSIONED_KERNELS_ROOT_DIRECTORY", this.mapper_VERSIONED_KERNELS_ROOT_DIRECTORY.serializeValue(input.VERSIONED_KERNELS_ROOT_DIRECTORY()));
                                                                                                                     ^
  symbol:   method VERSIONED_KERNELS_ROOT_DIRECTORY()
  location: variable input of type AddPeriapses
```
2. It now generates files called eg. `AddApoapsisMapper.jar` instead of `AddApoapsis.jar` like before

The first issue (we believe) is fixed by changing [this line of ProcedureProcessor](https://github.com/NASA-AMMOS/aerie/blob/e4b9061397bbec62fd08fce15138dc56889a5dcf/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/ProcedureProcessor.java#L269) to use `ElementKind.RECORD_COMPONENT` instead of `FIELD` but would like to verify this further.

mattdailis:
* **Tickets addressed:** Fixes #1776 
* **Review:** By commit  

## Description

@dandelany and @AaronPlave noticed that the annotation processor generates invalid code for scheduling procedures that contain static fields.

The cause was identified as filtering by FIELD (which includes static fields) instead of by RECORD_COMPONENT (which excludes static fields).

## Verification

@dandelany made the change locally and tested that the orbiter model compiles and runs successfully

## Documentation

This is a bugfix; no documentation has been invalidated

## Future work

- We observed that jar files now have the `Mapper` suffix, when in previous versions they didn't. Not sure what's up with that

### DIFF
--- a/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/ProcedureProcessor.java
+++ b/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/ProcedureProcessor.java
@@ -266,7 +266,7 @@ public final class ProcedureProcessor implements Processor {
   private List<ParameterRecord> getExportParameters(final TypeElement exportTypeElement)
   {
     return exportTypeElement.getEnclosedElements().stream()
-                            .filter(e -> e.getKind() == ElementKind.FIELD) // Element must be a field
+                            .filter(e -> e.getKind() == ElementKind.RECORD_COMPONENT) // Element must be a record component
                             .map(e -> new ParameterRecord(e.getSimpleName().toString(), e.asType(), e))
                             .toList();
   }


Mythicaeda (approved):

dandelany (approved):
👍 thanks - we may have to revisit this if we ever want to support classes, but I confirmed with @Mythicaeda  we currently only support `Record` type procedural goals & constraints, so this should work for all cases.

Sidenote - some of this confusion came from the docs which had a `class`-type constraint, so @Mythicaeda [updated the docs](https://github.com/NASA-AMMOS/plandev-docs/pull/5) to correct the `Record` (pun intended)

PR #1777, opened 2026-02-03
Format Java Backend Exceptions

### LINKED ISSUE #1732: Implement common HTTP error message formats for all Aerie services

### LINKED ISSUE #1771: Reduce the Types of Exceptions PermissionsService throws
See this comment chain:
https://github.com/NASA-AMMOS/aerie/pull/1760#discussion_r2628439527

In short, make a single exception in the Permissions Service that contains:
1. The HTTP status code the context should be set to 
2. The actual exception thrown

Mythicaeda:
* **Tickets addressed:** Closes #1771 and closes #1732 
* **Review:** By commit  

## Description

This PR does three things:
1. Make PermissionsService only throw a singular exception to reduce duplicated code in the services that use it
2. Update how Exceptions thrown in the Java endpoints are formatted to be united.
3. Ensure that Hasura returns the entire formatted exception in the `extensions` object

Number 2) is accomplished by pulling FormattedError out into the `parsing-utils` library, then creating service-specific subclasses to handle service-specific exceptions.

## Verification

E2E Tests were updated to reflect the updates to the error messages.

I did some manual tests with the UI. In these tests, I started a debugger in the backend with a breakpoint set at the start of the Hasura Action Handler code. After submitting the request via the UI, I would go in and remove data from the database to force the backend pre-checks to fail. I'd then resume the backend and observe that 1) the backend was sending the information correctly and 2) how the UI was parsing the response.

What I observed was:
- for Hasura Action endpoints, the backend is correctly putting the formatted exception in the `extensions` object
- the UI is reading this `extensions` object and is able to extract `data`, `service`, and `timestamp` out of it. It's unclear if it's using the top-level `message` field or the one in the `extensions` object, but that doesn't matter since they contain the same string.
- the UI is **not** reading the `type` field out of the `extensions` object, instead using a generic `ERROR` type. The `type` field is present in `extensions` object
- if scheduling, simulation, or constraints throw an exception (rather than failing), the exception is placed under the generic `LOGS` tab rather than being sorted into the correct category.

Example:
Server Response:
```json
{
	"errors": [
		{
			"extensions": {
				"data": {
					"specification_id": 659
				},
				"message": "Could not check permissions on scheduling specification 659: specification does not exist.",
				"service": "aerie_permissions",
				"timestamp": "2026-05-18T16:10:49.084229462Z",
				"trace": <TRACE REMOVED BY ME FOR CONCISENESS>,
				"type": "NO_SUCH_SCHEDULING_SPECIFICATION"
			},
			"message": "Could not check permissions on scheduling specification 659: specification does not exist."
		}
	]
}
```

UI Logs:
image


dandelany:
@Mythicaeda will do some smoke testing of this against the current version of the UI to make sure the new message format is fully supported

dandelany:
Discussed this one with @Mythicaeda yesterday and we realized that the endpoints used by Hasura queries (listed in `deployment/hasura/metadata/actions.yaml`) need to have their error payloads wrapped in the `extensions` structure as noted in #1732 (see "Endpoints accessed by Hasura Actions")

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
```diff
@@ -73,6 +81,55 @@ public void apply(final Javalin javalin) {
...
+      //  MissionModelLoader.MissionModelLoadException.class, (ex, ctx) ->
+        //    ctx.status(500).json(new MerlinFormattedError(ex)));
+    javalin.exception(
+        HttpResponseException.class, (ex, ctx) ->
+            ctx.status(ex.getStatus()).json(new FormattedError(AerieService.SCHEDULER_SERVER, "HTTP_RESPONSE_EXCEPTION", ex)));
+    javalin.exception(Exception.class, (ex, ctx) -> {
+      // Catch-all for unexpected issues
+      final var message = ex.getMessage() != null ? ex.getMessage() : "Unknown error.";
+      final var fe = new FormattedError(AerieService.SCHEDULER_SERVER, "UNKNOWN_ERROR", message, ex);
+      logger.error("Unexpected error processing workspace request {}", fe);
```
AaronPlave: Should this be "Unexpected error processing scheduling request {}"

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/http/MerlinBindings.java
```diff
@@ -112,26 +125,68 @@ public void apply(final Javalin javalin) {
...
+    javalin.exception(MissionModelLoadException.class, (ex, ctx) ->
+        ctx.status(500).json(new MerlinFormattedError(ex)));
+    javalin.exception(
+        HttpResponseException.class, (ex, ctx) ->
+            ctx.status(ex.getStatus()).json(new FormattedError(FormattedError.AerieService.MERLIN_SERVER, "HTTP_RESPONSE_EXCEPTION", ex)));
+    javalin.exception(Exception.class, (ex, ctx) -> {
+      // Catch-all for unexpected issues
+      final var message = ex.getMessage() != null ? ex.getMessage() : "Unknown error.";
+      final var fe = new FormattedError(FormattedError.AerieService.MERLIN_SERVER, "UNKNOWN_ERROR", message, ex);
+      logger.error("Unexpected error processing workspace request {}", fe);
```
AaronPlave: Should this be "Unexpected error processing merlin request {}"?

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/http/SchedulerBindings.java
Mythicaeda: Oops! I've removed the word "workspace" from all three locations, since which service is throwing the error is already included in the log message.

scheduler-server/src/main/java/gov/nasa/jpl/aerie/scheduler/server/exceptions/SchedulerFormattedError.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public SchedulerFormattedError(SpecificationLoadException ex) {
+    super(
+        AerieService.SCHEDULER_SERVER,
+        "SPECIFICATION_LOAD_EXCEPTION",
+        ex,
+        SchedulingCompilationError.schedulingErrorJsonP.unparse(ex.errors));
+  }
+
+  public SchedulerFormattedError(DatabaseException ex) {
```
AaronPlave, line 54: Should this be a db exception or a scheduling exception since the service here is still the scheduler?

Mythicaeda: I have it under the main service's exception because a `DatabaseException` is thrown specifically when the service has a networking issue with the DB. If there's an actual issue with the transaction (for example, no rows are returned from the lookup or an insert fails), then a different exception type is thrown (ie `NoSuchPlanException`, `FailedInsertException`).

AaronPlave: Oh wait right, the service here is merlin server.. is it odd to be hitting scheduling (since that is where the request is ultimately running) and get "service = merlin, db exception"? Or is that the right way? I guess whatever it is we should be consistent about it when we have these interacting services (you likely already are doing this, just talking out loud).

Mythicaeda: Oh the service being MERLIN_SERVICE is a typo, lemme fix that

dandelany:
FYI @Mythicaeda I merged `develop` into this branch and fixed some conflicts with the "simultaneous edit" work in [bd10a8b](https://github.com/NASA-AMMOS/plandev/pull/1777/commits/bd10a8bbea4b5e7d32d9dbf03ac763968a24e784) and [8959452](https://github.com/NASA-AMMOS/plandev/pull/1777/commits/8959452c9d8fe300abc63eb588f1a2fcac10d778), tests are passing - though scans are failing again (should be fixed by #1850). I'm working on doing the same with the UI branch & smoke testing the merged branches

dandelany (approved):
Thanks @Mythicaeda - this required some more changes on the UI than we originally thought, but I think we're finally there. This has been reviewed and tested with the latest https://github.com/NASA-AMMOS/plandev-ui/pull/1933 and they are ready to merge (together)

PR #1780, opened 2026-02-09
Fix Generated Procedure Names

### LINKED ISSUE #1776: Issues with Procedural Goal mapper generation

Mythicaeda:
* **Tickets addressed:** fixes #1776  
* **Review:** By commit  

## Description

The generated Procedures now match the name of the user code procedure.

Remove bandaid fix done to `e2e-tests` package

## Verification

Checked the output of compilation when running `./gradlew clean assemble buildAllProcedureJars`. The Jars and compiled files no longer have `Mapper` attached to the end

## Documentation

Only docs update needed is unrelated to this change in specific: Adrien's changes are breaking (procedures compiled on versions prior to 4.0 will not run on 4.0), and our upgrade guide should call that out.

### DIFF
--- a/e2e-tests/build.gradle
+++ b/e2e-tests/build.gradle
@@ -108,7 +108,7 @@ tasks.create("generateProcedureJarTasks") {
   }
 
   files.toList().each { file ->
-    final nameWithoutExtension = file.name.replace(".java", "").replace("Mapper", "")
+    final nameWithoutExtension = file.name.replace(".java", "")
     final taskName = "buildProcedureJar_${nameWithoutExtension}"
 
     println "Generating ${taskName} task, which will build ${nameWithoutExtension}.jar"
--- a/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/AllStaticallyDefinedMethodMaker.java
+++ b/procedural/processor/src/main/java/gov/nasa/ammos/aerie/procedural/processor/AllStaticallyDefinedMethodMaker.java
@@ -24,7 +24,7 @@ import java.util.stream.Collectors;
 
   @Override
   public MethodSpec makeInstantiateMethod() {
-    final var activityTypeName = inputType.declaration().getQualifiedName().toString();
+    final var procedureName = inputType.declaration().getQualifiedName().toString();
 
     var methodBuilder = MethodSpec.methodBuilder("instantiate")
         .addModifiers(Modifier.PUBLIC)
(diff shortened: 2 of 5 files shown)


dandelany (commented):
This looks mostly good - just one question - what's the reason for changing to `getQualifiedName()` instead of SimpleName? Just better debug-ability/less confusion? Any chance this could cause any issues re: backwards compatibility, updating procedures, etc? Seems safe but just double checking

Mythicaeda:
> This looks mostly good - just one question - what's the reason for changing to `getQualifiedName()` instead of SimpleName?

`getSimpleName` just returns the name of the class, without any of the package info. `getQualifiedName` returns the fully specified name of the class. Because we want the Mapper class to share a name with the user code class it instantiates, it **must** refer to the user code class with its fully qualified name. I think this would be clearer if `activityTypeName` was renamed from the copy-paste value in the mission model generator to `procedureName`.

dandelany (approved):
👍 thanks!

PR #1787, opened 2026-02-24
Fix MerlinServiceException when Scheduling Posts Sim Results with Unfinished Activities

Mythicaeda:
* **Review:** By commit  

## Description

There was a bug in how the scheduler was posting spans for unfinished activities to Hasura, where it would try to set to value of its duration to the _string_ `"null"` instead of the _value_ `null`. This PR fixes this by only including the `duration` field in the JSON version of the span if the Java version has a duration.

## Verification

I added a new test and example procedure to the Scheduling E2E Tests. If you try and schedule this procedure on `develop`, you will see the `MerlinServiceException` when the scheduler attempts to post simulation results. Meanwhile, the results will post successfully if scheduled against this branch.


dandelany (approved):
👍 reviewed and tested with @AaronPlave - makes sense and works, thanks!

PR #1788, opened 2026-02-24
Update default SeqJSON association to JSON

Mythicaeda:
## Description

As per the title.

This intentionally does not have a migration to avoid impacting end users who want `seq.json` to be a `SEQUENCE` type. If users want to update their own deployments, they can update the value in the table `ui.file_extension_content_type`.

This change should be called out in the upgrade guide.


dandelany (approved):
👍 looks good, thanks!

PR #1789, opened 2026-02-25
Add External Event Visibility to Procedural Constraints

pranav-super:
* **Tickets addressed:** --
* **Review:** By commit  

## Description
This pull request introduces visibility of external events in procedural constraints. Conceptually, the approach is extremely similar to that taken when introducing external events to procedural scheduling, and the implementation largely piggybacks off of that implementation. 

That being said, this PR introduces external events to `PlanRepository`, `PlanService`, and other related entities like `EvaluationEnvironment` that previously lacked visibility to external events. These are the analogues in procedural constraints to `MerlinDatabaseService` (and other smaller items like `SchedulerToProcedurePlanAdapter` and `SynchronousSchedulerAgent`). But common entities are still shared, like the Kotlin timeline representations of External Events and Sources. 

## Verification
A few new end-to-end tests were introduced in `ExternalEventConstraintTests` under `e2e-tests`, to verify that several constraints leveraging External Events worked as expected.

## Documentation
No documentation was invalidated, though examples can/should now be added to the "Examples" page under the [Procedural Constraints section of the documentation](https://nasa-ammos.github.io/plandev-docs/scheduling-and-constraints/procedural/constraints/introduction/). As a matter of fact, that page already claims/suggests that external events can be leveraged by the constraints engine!

## Future work
None anticipated.

(Small note: significant issues with the build process arose when trying to organize the procedures in `src/main/java` under `e2e-tests` into separate `constraints` and `scheduling` folders. This organization would be useful but has worked inconsistently when building.)

### DIFF
--- a/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventPresenceConstraint.java
+++ b/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventPresenceConstraint.java
@@ -1,0 +1,57 @@
+package gov.nasa.jpl.aerie.e2e.procedural.scheduling.procedures;
+
+import gov.nasa.ammos.aerie.procedural.constraints.Constraint;
+import gov.nasa.ammos.aerie.procedural.constraints.Violation;
+import gov.nasa.ammos.aerie.procedural.constraints.Violations;
+import gov.nasa.ammos.aerie.procedural.constraints.annotations.ConstraintProcedure;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.EventQuery;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.Plan;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.SimulationResults;
+import org.jetbrains.annotations.NotNull;
+
+// TODO: Make parameters Lists of Strings/SourceQueries for more intricate EventQuery testing in
+//        ExternalEventConstraintsTests when ValueMappers are able to be added, following the resolution of:
+//          https://github.com/NASA-AMMOS/plandev/issues/1737
+//          "@WithMappers does not work in procedural scheduling".
+@ConstraintProcedure
+public record ExternalEventPresenceConstraint(
+    String eventType,
+    String derivationGroup,
+    String sourceKey
+  ) implements Constraint {
+  @NotNull
+  @Override
+  public Violations run(@NotNull Plan plan, @NotNull SimulationResults simResults) {
+    // handle omitted sourceKey/derivationGroup/eventType (setting them to null manually, lest they be considered
+    //    missing arguments
+    final String NULL_VALUE = "NULL";
+    final var updatedEventType = eventType.equals(NULL_VALUE) ? null : eventType;
+    final var updatedDerivationGroup = derivationGroup.equals(NULL_VALUE) ? null : derivationGroup;
+    final var sourceQuery =
+        sourceKey.equals(NULL_VALUE) || derivationGroup.equals(NULL_VALUE)
+            ? null
+            : new EventQuery.SourceQuery(sourceKey, derivationGroup);
+
+    // make the query
+    var events = plan.events(
+        new EventQuery(
+            updatedDerivationGroup,
+            updatedEventType,
+            sourceQuery
+        )
+    );
+
+    // if we have any events that suit the above query
+    if (!events.collect().isEmpty()) {
+      // ...mark those events as violating
+      return new Violations(
+          events.collectIntervals()
+                .stream()
+                .map(e -> new Violation(e.getInterval()))
+                .toList()
+      );
+    }
+    // otherwise, return an empty violations object.
+    return new Violations();
+  }
+}
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
@@ -1,0 +1,425 @@
+package gov.nasa.jpl.aerie.e2e.procedural.constraints;
+
+import gov.nasa.jpl.aerie.e2e.procedural.scheduling.ProceduralTestingSetup;
+import gov.nasa.jpl.aerie.e2e.types.ConstraintInvocationId;
+import gov.nasa.jpl.aerie.e2e.types.ConstraintResult;
+import gov.nasa.jpl.aerie.e2e.utils.GatewayRequests;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+import javax.json.JsonValue;
+import java.io.IOException;
+import java.util.stream.Stream;
+
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.ADDITIONAL_DERIVATION_GROUP;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.ADDITIONAL_EVENT_TYPE;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.ADDITIONAL_SOURCE_KEY;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.DERIVATION_GROUP;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.EVENT_TYPE;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.SOURCE_KEY;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.SOURCE_TYPE;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.uploadExternalSourceEventTypes;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.uploadExternalSources;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Named.named;
+
+public class ExternalEventConstraintTests extends ProceduralTestingSetup {
+  private ConstraintInvocationId procedureId;
+  private final static String NULL_VALUE = "NULL";
+
+  @BeforeAll
+  void localBeforeAll() throws IOException {
+    // Upload some External Events
+    uploadExternalSourceEventTypes(playwright);
+    uploadExternalSources(playwright);
+  }
+
+  @BeforeEach
+  void localBeforeEach() throws IOException {
+    // make associations between plans and derivation groups
+    hasura.insertPlanDerivationGroupAssociation(planId, DERIVATION_GROUP);
+    hasura.insertPlanDerivationGroupAssociation(planId, ADDITIONAL_DERIVATION_GROUP);
+  }
+
+  @AfterAll
+  void localAfterAll() throws IOException {
+    // External Event deletion
+    hasura.deleteExternalSource(SOURCE_KEY, DERIVATION_GROUP);
+    hasura.deleteExternalSource(ADDITIONAL_SOURCE_KEY, ADDITIONAL_DERIVATION_GROUP);
+    hasura.deleteDerivationGroup(DERIVATION_GROUP);
+    hasura.deleteDerivationGroup(ADDITIONAL_DERIVATION_GROUP);
+    hasura.deleteExternalSourceType(SOURCE_TYPE);
+    hasura.deleteExternalEventType(EVENT_TYPE);
+    hasura.deleteExternalEventType(ADDITIONAL_EVENT_TYPE);
+  }
+
+  @AfterEach
+  void localAfterEach() throws IOException {
+    hasura.deleteConstraint(procedureId.id());
+
+    hasura.deletePlanDerivationGroupAssociation(planId, DERIVATION_GROUP);
+    hasura.deletePlanDerivationGroupAssociation(planId, ADDITIONAL_DERIVATION_GROUP);
+  }
+
+  void uploadConstraint(String constraintName, int planId) throws IOException{
+    try (final var gateway = new GatewayRequests(playwright)) {
+      final int constraintJarId = gateway.uploadJarFile("build/libs/" + constraintName + ".jar");
+      // Add Scheduling Procedure
+      procedureId = hasura.createConstraintSpecProcedure(
+          constraintName + ".jar",
+          constraintJarId,
+          planId
+      );
+    }
+  }
+
+  void uploadConstraint(String constraintName) throws IOException{
+    uploadConstraint(constraintName, planId);
+  }
+
+  // check that plan.events() functions correctly if nothing is attached
+  @Test
+  void verifyAbsenceOfExternalEvents() throws IOException{
+    // in this case, create a plan with no derivation group associations
+    var newPlanId = hasura.createPlan(
+        modelId,
+        "No Derivation Groups".formatted(this.getClass().getSimpleName()),
+        "48:00:00",
+        planStartTimestamp);
+
+    // upload the constraint
+    uploadConstraint("ExternalEventAbsenceConstraint", newPlanId);
+
+    // run it
+    hasura.awaitSimulation(newPlanId);
+    final var resp = hasura.checkConstraints(newPlanId);
+
+    // check that there are no violations, and no errors
+    assertEquals(1, resp.constraintsRun().size());
+    assertEquals(0, resp.constraintsRun().getFirst().errors().size());
+    assertEquals(0, resp.constraintsRun().getFirst().result().get().violations().size());
+
+    // manually delete the plan
+    hasura.deletePlan(newPlanId);
+  }
+
+  // verify the presence and location of events of a given type
+  @Test
+  void verifyPresenceOfExternalEventByEventType() throws IOException {
+    // upload the constraint
+    uploadConstraint("ExternalEventPresenceConstraint");
+
+    // update constraint arguments
+    final var args = Json.createObjectBuilder()
+                         .add("eventType", "TestType")
+                         .add("derivationGroup", NULL_VALUE)
+                         .add("sourceKey", NULL_VALUE)
+                         .build();
+    hasura.updateConstraintArguments(procedureId.invocationId(), args);
+
+    // run it
+    hasura.awaitSimulation(planId);
+    final var resp = hasura.checkConstraints(planId);
+
+    // checks that external events of this type are present (expect one violation per event, so 4 total)
+    assertEquals(1, resp.constraintsRun().size());
+    assertEquals(0, resp.constraintsRun().getFirst().errors().size());
+    assertEquals(4, resp.constraintsRun().getFirst().result().get().violations().size());
+
+    // check the windows of those violations, ensuring they line up with the events
+    var violations = resp.constraintsRun().getFirst().result().get().violations();
+
+    // violation 1: single window, for event from 01:00:00 to 02:00:00
+    var firstViolation = violations.getFirst().windows();
+    assertEquals(1, firstViolation.size());
+    assertEquals(
+        new ConstraintResult.Interval(3600000000L, 7200000000L),
+        firstViolation.getFirst()
+    );
+    // violation 2: single window, for event from 03:00:00 to 04:00:00
+    var secondViolation = violations.get(1).windows();
+    assertEquals(1, secondViolation.size());
+    assertEquals(
+        new ConstraintResult.Interval(10800000000L, 14400000000L),
+        secondViolation.getFirst()
+    );
+    // violation 3: single window, for event from 05:00:00 to 06:00:00
+    var thirdViolation = violations.get(2).windows();
+    assertEquals(1, thirdViolation.size());
+    assertEquals(
+        new ConstraintResult.Interval(18000000000L, 21600000000L),
+        thirdViolation.getFirst()
+    );
+    // violation 4: single window, for event from 25:00:00 to 26:00:00
+    var fourthViolation = violations.get(3).windows();
+    assertEquals(1, fourthViolation.size());
+    assertEquals(
+        new ConstraintResult.Interval(90000000000L, 93600000000L),
+        fourthViolation.getFirst()
+    );
+  }
+
+  // verify the presence and location of events from a given derivation group
+  @Test
+  void verifyPresenceOfExternalEventByDerivationGroup() throws IOException {
+    // upload the constraint
+    uploadConstraint("ExternalEventPresenceConstraint");
+
+    // update constraint arguments
+    final var args = Json.createObjectBuilder()
+                         .add("derivationGroup", "TestGroup")
+                         .add("eventType", NULL_VALUE)
+                         .add("sourceKey", NULL_VALUE)
+                         .build();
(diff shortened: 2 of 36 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanExternalEventsAction.java
```diff
@@ -0,0 +1,129 @@
...
+
+/*package-local*/ final class GetPlanExternalEventsAction implements AutoCloseable {
+  private final @Language("SQL") String sql = """
+      select
+        ee.event_key,
+        ee.event_type_name,
+        ee.source_key,
+        ee.derivation_group_name,
+        ee.start_time,
+        extract(epoch from ee.duration) as duration,
```
Mythicaeda, line 32: Proposal:

```suggestion
        extract(epoch from ee.duration)*1e6 as duration_micros,
```

This would allow you to just do `Duration.of(resultSet.getLong("duration_micros"), Duration.MICROSECOND);` on L68, rather than casting Double to a Long

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanExternalEventsAction.java
```diff
@@ -0,0 +1,129 @@
...
+      //        but I am unsure if the design would appreciate multiple queries, or perhaps an Action referencing
+      //        another Action (i.e. we have a GetExternalSourceAttributesAction invoked by this action).
+      final var sourceAttributes = new SerializedValueJsonParser().parse(Json
+          .createReader(
+              new StringReader(Objects.requireNonNull(resultSet.getObject("source_attributes", PGobject.class).getValue()))
+          )
+          .readObject())
+         .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)))
+         .asMap()
+         .get();
```
Mythicaeda, line 93: Heads up: this is not standard practice for how we get JsonObject fields from the DB in Aerie. Instead we use the helper method `PostgresParsers.getJsonColumn`. 

To fix this code, you need to perform 2 steps:

### Step 1: Create a Parser for "Event Attributes"
Add the following line to `gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers` (I specifically recommend adding it around L120):

```java
public static final JsonParser<Map<String, SerializedValue>> eventAttributesP = mapP(serializedValueP);
```

^I'm assuming the event attributes are a Map of SerializedValues based on the use of `SerializedValueJsonParser`.

### Step 2: Use `PostgresParsers.getJsonColumn`
These massive statements then simplify into

```java
// get event attributes
final var eventAttributes = getJsonColumn(resultSet, "attributes", eventAttributesP)
          .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)));
      
// get source attributes
// (long comment skipped for brevity)
final var sourceAttributes = getJsonColumn(resultSet, "source_attributes", eventAttributesP)
          .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetPlanExternalEventsAction.java
```diff
@@ -0,0 +1,129 @@
...
+          .readObject())
+          .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)))
+          .asMap()
+          .get();
+
+      // get source attributes
+      // NOTE: in the query, we get the source attributes with each event. This is pretty repetitive, which I don't
+      //        quite like, especially as there could be a _lot_ of events. A fix to this could entail a second query,
+      //        but I am unsure if the design would appreciate multiple queries, or perhaps an Action referencing
+      //        another Action (i.e. we have a GetExternalSourceAttributesAction invoked by this action).
```
Mythicaeda, line 85: WRT this comment, something that bothers me is that you're creating a new External Source object for every single External Event, even if those events share a Source. (It's definitely a new instance, as ExternalSource.kt does not use [the `@JvmRecord` annotation that would make it a record](https://kotlinlang.org/docs/jvm-records.html#declare-records-in-kotlin))

To avoid this, what you should do is have this method accept a `Map<Key, ExternalSource> sources` as a parameter and then use the `source_key` column from the `resultSet` to look up the source.

You would then add a `GetExternalSourcesMap` Action (and corresponding `get` method in the `ProfileRepository` class). Finally, you should update `ProfileRepository.getExternalEvents` (the caller of this method) to call `GetExternalSourcesMap` and feed the resulting map into this method.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
```diff
@@ -149,6 +152,17 @@ static List<ProfileSegment<Optional<SerializedValue>>> getDiscreteProfileSegment
     }
   }
 
+  static Map<String, List<ExternalEvent>> getExternalEvents(
```
Mythicaeda, line 155: Can you please explain why this method is in `ProfileRepository` rather than `PostgresPlanRepository`?

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
```diff
@@ -271,6 +275,22 @@ public List<Pair<Duration, ProfileSet>> getExternalDatasets(
...
+    try (final var connection = this.dataSource.getConnection()) {
+        return ProfileRepository.getExternalEvents(connection, planId, horizonStart);
+    } catch (final SQLException ex) {
+      throw new DatabaseException(
+          "Failed to get external events for plan with id `%s`".formatted(planId), ex);
+    } catch (final InvalidEntityException in) {
+      throw new RuntimeException(
+          ("Failed to get external events for plan with id `%s; "
+           + "failed to parse jsonb for external event/source attributes`").formatted(planId), in);
+    }
```
Mythicaeda, line 291: Something I've been looking into improving is reducing the number of "Silent Runtime Exceptions" thrown by our code. If this method were to properly throw its `SQLException` and `InvalidEntityException` rather than cast them to `RuntimeException`, would that effect user code or just the Constraint Action? 

(Please note that I'm not asking for a change here, just trying to gauge the impact for some future work.)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
```diff
@@ -19,9 +21,11 @@
 import org.apache.commons.lang3.tuple.Pair;
 
 import javax.sql.DataSource;
+import java.io.IOException;
```
Mythicaeda, line 24: This added import statement is unused

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/mocks/InMemoryPlanRepository.java
```diff
@@ -15,6 +16,8 @@
 import gov.nasa.jpl.aerie.types.Timestamp;
 import org.apache.commons.lang3.tuple.Pair;
 
+import java.io.IOException;
```
Mythicaeda, line 19: This added import statement is unused

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/services/PlanService.java
```diff
@@ -14,6 +15,8 @@
 import gov.nasa.jpl.aerie.types.Timestamp;
 import org.apache.commons.lang3.tuple.Pair;
 
+import java.io.IOException;
```
Mythicaeda, line 18: This added import statement is unused (`ConstraintId` on L8 is also unused, but that's not something you added)

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/models/ReadonlyPlan.java
```diff
@@ -4,6 +4,8 @@
 import gov.nasa.ammos.aerie.procedural.timeline.collections.Directives;
 import gov.nasa.ammos.aerie.procedural.timeline.collections.ExternalEvents;
 import gov.nasa.ammos.aerie.procedural.timeline.ops.SerialSegmentOps;
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.ExternalEvent;
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.ExternalSource;
```
Mythicaeda, line 8: `ExternalSource`, `ArrayList`, and `Collectors` are all added but unused imports.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/utils/ExternalEventUtils.java
```diff
@@ -0,0 +1,98 @@
+package gov.nasa.jpl.aerie.e2e.utils;
+
+import com.microsoft.playwright.Playwright;
+
+import javax.json.Json;
+import javax.json.JsonObject;
+import java.io.IOException;
+
+/**
+ * Extracted the following functions into a separate "util" file, as they are used by constraints and scheduling tests.
```
Mythicaeda, line 10: Minor: This comment reads like a Git Commit Message and not an actual comment describing this class. Could you please reword it?

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventAttributeQuantityConstraint.java
```diff
@@ -0,0 +1,36 @@
+package gov.nasa.jpl.aerie.e2e.procedural.scheduling.procedures;
+
+import gov.nasa.ammos.aerie.procedural.constraints.Constraint;
+import gov.nasa.ammos.aerie.procedural.constraints.Violation;
```
Mythicaeda, line 4: Unused Import statement

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
```diff
@@ -0,0 +1,173 @@
+package gov.nasa.jpl.aerie.e2e.procedural.constraints;
+
+import gov.nasa.jpl.aerie.e2e.procedural.scheduling.ProceduralTestingSetup;
+import gov.nasa.jpl.aerie.e2e.types.ConstraintInvocationId;
+import gov.nasa.jpl.aerie.e2e.utils.GatewayRequests;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
```
Mythicaeda, line 7: `BeforeAll` and `JsonObjectBuilder` are unused imports

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
```diff
@@ -0,0 +1,173 @@
...
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.SOURCE_KEY;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.SOURCE_TYPE;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.uploadExternalSourceEventTypes;
+import static gov.nasa.jpl.aerie.e2e.utils.ExternalEventUtils.uploadExternalSources;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ExternalEventConstraintTests extends ProceduralTestingSetup {
+  private ConstraintInvocationId procedureId;
+
+  // TODO: beforeAll?
```
Mythicaeda, line 29: Please remove this TODO

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventPresenceConstraint.java
```diff
@@ -0,0 +1,27 @@
...
+import gov.nasa.ammos.aerie.procedural.constraints.Violation;
+import gov.nasa.ammos.aerie.procedural.constraints.Violations;
+import gov.nasa.ammos.aerie.procedural.constraints.annotations.ConstraintProcedure;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.EventQuery;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.Plan;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.SimulationResults;
+import org.jetbrains.annotations.NotNull;
+
+@ConstraintProcedure
+public record ExternalEventPresenceConstraint() implements Constraint {
```
Mythicaeda, line 13: This Constraint should be written as inverted. That is, it returns a Violation if it detects an external event. This way, it is obvious when reading the test `verifyPresenceOfExternalEvent` that the Constraint saw an External Event of type `TestType`.

You could additionally make this a paramaterized constraint and also run it with an `eventType` argument that _isnt_ present (ie `fakeType`) to show that `plan.events` is successfully filtering out undesired types

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
```diff
@@ -0,0 +1,173 @@
...
+    final var args = Json.createObjectBuilder().add("quantity", 3).build();
+    hasura.updateConstraintArguments(procedureId.invocationId(), args);
+
+    // run it
+    hasura.awaitSimulation(planId);
+    final var resp = hasura.checkConstraints(planId);
+
+    // check that there are violations in two places (where the existing events are)
+    assertEquals(1, resp.constraintsRun().size());
+    assertEquals(2, resp.constraintsRun().getFirst().result().get().violations().size());
```
Mythicaeda, line 100: Asserting that there are two violations is *necessary but not sufficient*  to prove that those violations are "(where the existing events are)". You need additional assertions confirming that these are `window` violations and not `activityDirective` violations, and that the windows correspond to the expected intervals.

e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/ExternalEventAttributeQuantityConstraint.java
```diff
@@ -0,0 +1,36 @@
...
+import gov.nasa.ammos.aerie.procedural.constraints.Violations;
+import gov.nasa.ammos.aerie.procedural.constraints.annotations.ConstraintProcedure;
+import gov.nasa.ammos.aerie.procedural.timeline.collections.Windows;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.EventQuery;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.Plan;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.SimulationResults;
+import org.jetbrains.annotations.NotNull;
+
+@ConstraintProcedure
+public record ExternalEventAttributeQuantityConstraint(int quantity) implements Constraint {
```
Mythicaeda, line 14: Suggestion: A more meaningful parameter for the purpose of testing would be `attribute` rather than `quantity`, with the constraint returning a violation for each external event with that attribute. After all, this constraint is meant to be testing that constraints have access to the attributes of external events.

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
```diff
@@ -0,0 +1,173 @@
...
+
+    // check that there are violations
+    assertEquals(1, resp.constraintsRun().size());
+
+    var violations = resp.constraintsRun().getFirst().result().get().violations();
+    assertEquals(1, violations.size());
+
+    // get length of violation (expect 30 minutes)
+    var violation = violations.get(0).windows().get(0);
+    assertEquals(30, (violation.end() - violation.start())/(60*1000000));
```
Mythicaeda, line 127: You need additional assertions to prove that the violation window actually aligns with where the activity and external event overlap

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // ensure external event (with given attributes) overlaps with activity for total of M seconds
+  @Test
+  void testExternalEventActivityOverlapForSeconds() throws IOException {
+    // upload the constraint
```
Mythicaeda, line 133: What is this testing with regards to the visibility of External Events in Constraints that was not covered by `testExternalEventActivityOverlap`?

Mythicaeda (changes requested):
Marking this as "Request Changes" because of the `GetExternalEventAction` code.

With regards to the tests, I noticed that you _only_ tested that the `eventType` query param of the `plan.events` method works, not `derivationGroup` or `source`. Tests that prove that these other two filters work, and that they work in combination, would be valuable.

Additionally, I'm requesting a test added where a constraint that runs `plan.events` is checked against a plan that has no external events.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
pranav-super: Yeah! The reasoning here was to mirror the implementations for `ExternalDatasets`, as external events are also external entities that can be associated with plans. The implementation can be found split across these two files, much like `ExternalDatasets` have `getExternalDatasets` in `PostgresPlanRepository`, which invokes `getPlanDatasetsForPlan`. Wholly implementing `getExternalEvents` in `PostgresPlanRepository` seemed a little misleading, as external events are a separate entity that can be associated with a plan but are partly independent from them. However, moving it entirely there if it suits the PlanDev design better could definitely be done!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/PostgresPlanRepository.java
pranav-super: I am hesitant to say it wouldn't affect user code, but I don't think it would if the exception type changes. Nothing on our end really depends on or expects these exception types though!

If your question is also inclusive of what happens while code runs, I'm not entirely sure. I have not encountered this exception being thrown at all yet, although it could be if parsing fails (although, were parsing to fail (L76 and L89 of `GetPlanExternalEventsAction`), an error would have been encountered much earlier, at source upload.

Mythicaeda: Understood, thanks!

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/ProfileRepository.java
Mythicaeda: TL;DR: thank you for the explanation! Based on it, please inline the implementation into the `PlanRepository`. 

Full Response:
While its true that both External Datasets and External Events are external entities associated with Plans, the key difference here is that External Datasets are made out of Profiles--that's why the code for fetching them is in the *Profile*Repository. Since External Events are not comprised of Profiles, it does not make sense for the `ProfileRepository` to be responsible for fetching them. A better mirror would be to put the code in a new `EventRepository`. However, since this is our only `Events` Action currently, I'm fine with keeping it in `PlanRepository`.

pranav-super: Got it. I am planning to introduce an `EventRepository` class with static getters for external events, much like `ProfileRepository`, and pointing the invocation in `PlanRepository` to that, but otherwise changing nothing. If that sounds incorrect, however, (i.e. functionality in `PlanRepository` *should* be changed if an `EventRepository` is introduced), I can inline it.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetExternalSourcesMapAction.java
```diff
@@ -0,0 +1,69 @@
...
+    final var result = new HashMap<String, ExternalSource>();
+    this.statement.setLong(1, planId.id());
+    final var resultSet = statement.executeQuery();
+
+    while (resultSet.next()) {
+      // get source key
+      final String key = resultSet.getString("key");
+      // get source attributes
+      final var attributes = getJsonColumn(resultSet, "source_attributes", eventAttributesP)
+          .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)));
```
Mythicaeda: Minor typo here:


```suggestion
      final var attributes = getJsonColumn(resultSet, "attributes", eventAttributesP)
          .getSuccessOrThrow(reason -> new InvalidEntityException(List.of(reason)));
```

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetExternalSourcesMapAction.java
```diff
@@ -0,0 +1,69 @@
...
+/*package-local*/ final class GetExternalSourcesMapAction implements AutoCloseable {
+  private final @Language("SQL") String sql = """
+      select
+        s.key,
+        s.attributes
+      from merlin.plan_derivation_group as pdg
+        join merlin.external_source as s
+          on pdg.derivation_group_name = s.derivation_group_name
+      where pdg.plan_id = ?;
+    """;
```
Mythicaeda, line 27: 1 major thing and 1 minor thing:

Major: You forgot to return `derivation_group_name` as a column

Minor: _This is completely optional_, but since the column you're joining on has the same name in each table, you have the option of using `using` instead of `on`. The advantage to using `using` is that you only end up with a single `derivation_group_name` column after the join, rather than two columns with the same name and data. This allows you to skip needing to specify which tables each column comes from.

```suggestion
  private final @Language("SQL") String sql = """
      select
        key,
        attributes,
        derivation_group_name
      from merlin.plan_derivation_group as pdg
      join merlin.external_source as s using (derivation_group_name)
      where pdg.plan_id = ?;
    """;
```

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/constraints/ExternalEventConstraintTests.java
pranav-super: The idea was just to check that not only would a violation be flagged were there any overlap, but a violation could be filtered if the overlap window was a specific duration. However, if we precisely check the violation window size in `testExternalEventActivityOverlap`, this would be covered. The idea here was to provide an additional check that does an implied check of the window length without explicitly accessing it from the violation. That being said, if the above window check is made, this one should be obsolete, and can be removed.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetExternalSourcesMapAction.java
pranav-super: Updated during testing.

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetExternalSourcesMapAction.java
pranav-super: Oops! Updated during testing.

Mythicaeda (approved):
LGTM! Thank you for implementing this!

PR #1794, opened 2026-02-25
Add another false positive CVE for gosu to trivyignore

AaronPlave:
Addresses recent trivy CVE flag for gosu that is a false positive, see https://github.com/tianon/gosu/pull/173#issuecomment-3936440684


duranb (approved):

duranb:
👍🏻 Since this is the same CVE that we ignored on the UI

PR #1796, opened 2026-02-27
Fix action server crash from unhandled duplicate secrets

AaronPlave:
- Handle duplicate secrets gracefully by logging a warning and returning
- Add .catch() to the floating addActionSecret promise in the /secrets endpoint

* **Review:** By commit  

### DIFF
--- a/action-server/src/app.ts
+++ b/action-server/src/app.ts
@@ -3,6 +3,7 @@ import { configuration } from "./config";
 import {authMiddleware, corsMiddleware, jsonErrorMiddleware} from "./middleware";
 import { ActionRunner } from "./type/actionRunner";
 import { extractCookies } from "./utils/auth";
+import logger from "./utils/logger";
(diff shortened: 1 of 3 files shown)


dandelany:
I've been looking at this a bit and it does fix the issue, however I want to do a bit more digging to understand why the global `app.use(jsonErrorMiddleware);` didn't handle this and instead caused a crash... I wouldn't expect that but maybe there is a catch here re: unhandled promise rejections vs. thrown errors. Full review coming soon

dandelany (commented):
@AaronPlave  and I worked on this yesterday and fixed the issue in a better way, & also fixed the general JSON error handler middleware which was not working before. Looks like we broke e2e tests though - investigating.

dandelany (approved):
👍 reviewed and tested with @AaronPlave , fixed tests.

PR #1798, opened 2026-03-04
stringify + parse action results before returning them, to sanitize to only serializable values

dandelany:
## Description
Fixes an issue with actions, where in certain cases, if the Action returns results with certain unserializable values, action server will fail mysteriously. In particular, this happens if an action tries to return the adaptation (from `loadAdaptation`).

This deserves some future investigation (why does this fail **silently**?) - but ultimately we decided results should only be serializable values anyway (it doesn't make much sense for an action to return a function or a class in results). For truly unserializable values, this should cause a *loud* error which will end up in logs, rather than a silent one.

## Verification

1. Run this version of action server
2. Upload an action which calls `loadAdaptation` & returns it in results

for example:

```
import type { SeqJson } from "@nasa-jpl/seq-json-schema/types.js";
import {
  seqJsonLanguage,
  seqnLanguage,
  type OutputLanguage,
  type PhoenixAdaptation,
  type PhoenixContext,
  seqJsonToSeqn,
  seqnToSeqJson,
  seqnParser,
} from "@nasa-jpl/aerie-sequence-languages";

function toInputFormat(output: string) {
  const seqJson = JSON.parse(output) as SeqJson;
  return seqJsonToSeqn(seqJson);
}

function toOutputFormat(input: string, context: PhoenixContext, name: string) {
  const tree = seqnParser.parse(input);
  const seqJson = seqnToSeqJson(tree, input, context.commandDictionary, name);
  return JSON.stringify(seqJson, null, 2);
}

const seqJsonOutputLanguage: OutputLanguage = {
  ...seqJsonLanguage,
  toOutputFormat,
  toInputFormat,
};

// Create the main phoenix adaptation and export it for consumers
const adaptation: PhoenixAdaptation = {
  input: seqnLanguage,
  outputs: [seqJsonOutputLanguage]
}

export { adaptation };
```

### DIFF
--- a/action-server/src/utils/codeRunner.ts
+++ b/action-server/src/utils/codeRunner.ts
@@ -134,7 +134,11 @@ export const jsExecute = async (
     const actionsAPI = new ActionsAPI(client, workspaceId, actionConfig);
     const results = await context.main(parameters, settings, actionsAPI);
 
-    return { results, console: logBuffer, errors: null };
+    // clone + serialize results returned from action
+    // to sanitize unserializable things in object (todo: investigate more)
+    const cleanResults = JSON.parse(JSON.stringify(results));
+
+    return { results: cleanResults, console: logBuffer, errors: null };
   } catch (error: any) {
     // wrap `throw 10` into a `new throw(10)`
     let errorResponse: Error;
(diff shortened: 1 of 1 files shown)


AaronPlave (approved):

dandelany:
@AaronPlave I made one additional change after we talked (based on test results): [756081d](https://github.com/NASA-AMMOS/aerie/pull/1798/commits/756081dfb834c565c03c3b70cd9e4e783de55112)

This fixes support for actions with no `return` (ie. that return `undefined` - which can't be JSON.parsed). We should **allow** returning results but I don't think users should be **forced** to.

PR #1801, opened 2026-03-09
Split ActivityTypes by activity for improved incremental compilation

joshhaug:
* **Review:** By file  

## Overview
Fix incremental compilation by doing the following:
- Change generateActivityTypes() to return List<JavaFile> instead of single JavaFile
- Generate one `ActivityTypes_X.java` file per activity (~20 lines each)
- Generate master `ActivityTypes.java` that uses reflection to load all activity types
- Master only depends on package, not individual activities
- This enables incremental compilation: changing one activity only regenerates that one file
- Expected build time improvement scales with model size: 85%+ improvement for incremental builds on Clipper translation

## Description
The Clipper Model translation task uses a rather large Aerie model with 739 activities. We've been running into long compile times (3:30+) and have tried the obvious stuff -- increasing heap size, allowing parallel gradle tasks, enabling incremental compilations, etc. Interestingly, incremental compilations resulted in zero speedup if any change was made to the codebase. The reason for this seems to be dependency bottlenecks from a monolithic `ActivityTypes` file (2,978 lines) that imports all 739 activity mappers as static fields. `Mission.java` imports `ActivityTypes`, and 362 source files import `Mission.java`. So a single-line comment change in one activity --> invalidation of that activity's mapper --> invalidation of ActivityTypes --> invalidation of Mission.java --> recompile of 362+ files.

The proposed solution is to generate separate `ActivityTypes_X.java` files for each activity, rather than dumping all mapper references in a monolithic `ActivityTypes` file. We then generate a master `ActivityTypes` class that loads all activity types via reflection. I know, I know, this is a runtime-rather-than-compile-time tradeoff. I had Claude write a unit test to check this: `ActivityTypesPerformanceTest.java` and the results are negligible for Banananation (<1ms initialization, ~83ns per lookup) and surprisingly similar for Clipper translation (initial load time < 1ms,  ~0.015ms per activity lookup). I can say that the results are worth it for the clipper model; incremental compile time went down from a full recompile (3:30) to <0:30 when activity-level changes are made.

### DIFF
--- a/examples/banananation/src/test/java/gov/nasa/jpl/aerie/banananation/ActivityTypesPerformanceTest.java
+++ b/examples/banananation/src/test/java/gov/nasa/jpl/aerie/banananation/ActivityTypesPerformanceTest.java
@@ -1,0 +1,85 @@
+package gov.nasa.jpl.aerie.banananation;
+
+import gov.nasa.jpl.aerie.banananation.generated.ActivityTypes;
+import org.junit.jupiter.api.Test;
+
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
+import java.lang.management.MemoryUsage;
+
+public class ActivityTypesPerformanceTest {
+
+  @Test
+  public void testActivityTypesLoadTime() {
+    // Warm up JVM
+    System.gc();
+    
+    long startTime = System.nanoTime();
+    var directiveTypes = ActivityTypes.directiveTypes;
+    long endTime = System.nanoTime();
+    
+    long durationMs = (endTime - startTime) / 1_000_000;
+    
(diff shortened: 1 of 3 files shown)


dandelany:
Thanks @joshhaug ! We discussed this yesterday and devs are generally in favor of this, feedback was:

* make sure to look at as many edge cases as possible + add tests where appropriate
  - any issues with file names which differ from class names? or files which have the same name at different levels of folder hierarchy? things like that
* Josh will do some more testing with UI + clipper model
* using reflection here adds complexity, but we feel it's worth it for the benefit - add some code comments re: why we decided to do this
* look into microbenchmark harness for perf testing (if useful - not necessary)

PR #1803, opened 2026-03-16
Action versioning and archiving

AaronPlave:
* **Tickets addressed:** See UI PR https://github.com/NASA-AMMOS/aerie-ui/pull/1892
* **Review:** By file

## Description
Introduces versioning and archiving for action definitions. Instead of mutating action_definition in place when a new action file is uploaded, each upload now creates an immutable action_definition_version row with an auto-incrementing revision number.

Main changes:
* New actions.action_definition_version table holds per-version data (action_file_id, parameter_schema, settings_schema, author, archived)
* action_definition is simplified to hold only shared metadata (name, description, settings, archived, workspace_id) — file and schema columns moved to the version table
* action_run now records action_definition_revision so runs are pinned to the version that executed them. Defaults to the latest revision if not provided.
* DB triggers (pg_notify) updated to fire on action_definition_version inserts instead of action_definition inserts
* Action server listener updated to write schema extraction results to the version table and strip stale settings keys from the parent definition when a new version's schema drops them
* Hasura metadata and permissions added for the new table. Existing permissions updated to reflect the column changes
* Migration 31_action_versioning includes both up.sql and down.sql

Areas I'd appreciate closer review on:

- The migration SQL — especially the data-migration step that backfills version 0 from existing rows and the column drop ordering. Want to make sure the up/down are safe and complete.
- The auto-increment revision trigger (action_definition_version_set_revision)
- The stale-settings-stripping query in dbListeners.ts — this prunes settings keys from the parent action_definition when a new version's schema no longer includes them. Want to make sure the SQL logic is right and doesn't accidentally wipe valid settings.

## Verification
- [ ] Write tests
- [x] Verify behaviors with UI

### DIFF
--- a/action-server/src/listeners/dbListeners.ts
+++ b/action-server/src/listeners/dbListeners.ts
@@ -5,11 +5,10 @@ import type { PoolClient } from "pg";
 import { configuration } from "../config";
 import { ActionsDbManager } from "../db";
 import { ActionWorkerPool } from "../threads/workerPool";
-import type { ActionDefinitionInsertedPayload, ActionResponse, ActionRunInsertedPayload } from "../type/types";
+import { ActionRunner } from "../type/actionRunner";
+import type { ActionDefinitionInsertedPayload, ActionResponse, ActionRunCancellationRequestPayload, ActionRunInsertedPayload } from "../type/types";
 import { extractSchemas } from "../utils/codeRunner";
 import logger from "../utils/logger";
-import { ActionRunCancellationRequestPayload } from "../type/types";
-import { ActionRunner } from "../type/actionRunner";
 
 let listenClient: PoolClient | undefined;
 
@@ -30,11 +29,11 @@ async function refreshActionDefinitionSchema(payload: ActionDefinitionInsertedPa
 
   const pool = ActionsDbManager.getDb();
   const query = `
-    UPDATE actions.action_definition
+    UPDATE actions.action_definition_version
     SET
       parameter_schema = $1::jsonb,
       settings_schema = $2::jsonb
-    WHERE id = $3
+    WHERE action_definition_id = $3 AND revision = $4
       RETURNING *;
   `;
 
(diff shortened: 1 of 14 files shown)


dandelany (changes requested):
This is coming along well! Did some testing and review with @AaronPlave and identified the following backend issues:

1. FIXED ~If you upload multiple versions of an action, and then you run the action without specifying a version, the **UI defaults to the latest** version, but the backend actually **runs version 0** rather than correctly defaulting to latest. If you use the dropdown to specify a particular version (other than latest) it works correctly.~
2. FIXED ~If you are running an **older version** of an action, and the **settings schema has changed**, the action run modal will ask you to specify the settings for this run - OK so far. But if one or more of these settings was a **`file` or `fileList` type** setting, the file dropdown is not correctly populated with the files in the workspace, and the default setting values are not correctly set.~

dandelany:
After thinking about it, I realized that **I caused** issue 1 in the last comment, due to [this change](https://github.com/NASA-AMMOS/aerie/pull/1803/changes/1f753e658551abc7f27b6b0d5386f74a3d0c9976#diff-0795764cb861eeedef2f8642c95c785ee78d15b4f62b6a399fe7254cf1c01ba4L66) and [this change](https://github.com/NASA-AMMOS/aerie/pull/1803/changes/c80de321f8db064361684add8fd954e37a256a8d) (though you helped me do it @AaronPlave 😛). We meant to just simplify it but actually broke the behavior for future inserts - it needs to *not* have a default in order to correctly hit the `action_run_set_default_revision` trigger. Reverted to basically what you had + an extra comment and that seems to have fixed it.

AaronPlave:
> 2\. If you are running an **older version** of an action, and the **settings schema has changed**, the action run modal will ask you to specify the settings for this run - OK so far. But if one or more of these settings was a **`file` or `fileList` type** setting, the file dropdown is not correctly populated with the files in the workspace, and the default setting values are not correctly set.

Fixed in the UI PR.

Mythicaeda (approved):

dandelany (approved):
👍 all issues resolved, ready to merge with UI branch

PR #1807, opened 2026-03-25
Add Workspace Metadata

### LINKED ISSUE #1802: Metadata for SeqDev Workspace files - Protocol and API
## Summary

Introduce a metadata protocol for workspace files in SeqDev. Metadata is stored in hidden sibling JSON files and accessed exclusively through a Metadata API in the Workspace Service.

Metadata files are created lazily by the workspace service and follow the lifecycle of their associated workspace file - eg. when the underlying file is deleted or copied, their metadata file should also be deleted or copied.

This ticket implements the metadata storage protocol and API. Enforcement of metadata-controlled behavior (e.g., read-only files) is covered in a separate ticket.

---

# Metadata Storage Protocol

Each workspace file may have a metadata file stored as a hidden sibling, named as `.<filename>.meta.seqdev`. (Note: this will require a DB migration to add this as an allowed metadata file extension)

For example:

```
foo.seqn             // existing file
.foo.seqn.meta.seqdev  // metadata file
```

## Rules
- Metadata files are **not guaranteed to exist** for every file.
- Metadata files are created when metadata is first written.
- **Any non-metadata** workspace files (not directories) may have a metadata file.
- If a file with metadata is **moved, renamed, copied, or deleted**, its associated metadata file should be too.
- Combination file/metadata ops should be done as *atomically* as is feasible, ie. both succeed or both fail. Where atomicity is not possible, prefer solutions which allow the main file operation to succeed even if metadata operations fail.
- We should take any steps possible to prevent "orphaned" metadata files (without an associated regular file) for any operations **done through our API**, though orphans will be possible if user changes files directly.
 

## Metadata File JSON Structure

Metadata files will be JSON files with the following structure:

```json
{
  "version": "1",
  "readOnly": false,
  "createdBy": "username",
  "createdAt": "timestamp",
  "lastEditedBy": "username",
  "lastEditedAt": "timestamp",
  "user": {
    "arbitrary": "values"
  }
}
```

### Top-level fields

| Field | Type | Description |
|------|------|-------------|
| version | string | Metadata format version (service-controlled) |
| readOnly | boolean | Indicates file should be treated as read-only |
| createdBy | string | User who created metadata |
| lastEditedBy | string | Last user who modified file or metadata |
| lastEditedAt | string | Timestamp of last modification |
| user | object | Arbitrary user metadata |

### Rules

- Top-level fields in the metadata (besides `user`) are read/written to directly by SeqDev and have a controlled schema.
  - `readOnly` can be updated by the user, & is used internally by workspace service
  - `version`, `createdBy`, `createdAt`, `lastEditedBy`, & `lastEditedAt` are not directly editable by the user via the API, and are managed by SeqDev internally.
- The `user` field is an arbitrary nested JSON object controlled by the user. The keys can be anything, but they *cannot* contain dots (ie. `"user.name"` is an invalid key).

---

# Metadata API

All endpoints with a path reference the **main file path**, not the metadata file path.

---

## Get Metadata

Returns the metadata file for the file at `<filePath>`

```
GET /metadata/<workspaceId>/<filePath>
```

Behavior

| Condition | Response |
|----------|----------|
| File missing | 404 error |
| File path is a directory | 400 error (only files have metadata) |
| Metadata missing | `{ "version": "1" }` |
| Metadata valid | Metadata JSON file contents |

**NOTE:** that this will return the contents of the metadata file **AS-IS** -  if the JSON is malformed, the client will need to detect/handle it.

---

## Set Metadata (Upsert)

Set values in the metadata file given an update object

```
POST /metadata/<workspaceId>/<filePath>
```

Example request

```json
{
  "readOnly": false,
  "user": {
    "status": "draft"
  }
}
```

Rules

- Creates metadata if missing.
- Performs a **shallow merge** of top-level fields that the user is allowed to edit (namely `readOnly`).
- Also performs a **shallow merge** of any provided `user` object (so that the entire `user` object isn't overwritten).
- Unknown top-level keys rejected.
- If metadata is already malformed, this endpoint throws a 500 with an appropriate error message
- If user passes a malformed/non-object for the `user` object (including one with dots in the keys), respond with a 400 error.

---

## Unset Metadata Keys

Unsets values in the metadata file for the given list of keys

```
POST /metadata/unset/<workspaceId>/<filePath>
```

Request body

```json
["readOnly", "user.status", "user.info.name"]
```

Rules

- Removes keys from metadata.
- Supports dot-path syntax inside `user` with any level of depth.
- Removing missing keys is a no-op.

---

## Delete Metadata

```
DELETE /metadata/<workspaceId>/<filePath>
```

Behavior

| Condition | Result |
|----------|--------|
| File missing | 404 |
| Metadata missing | 200 |

---

# Changes to Existing Workspace API

We want to modify the existing workspace directory listing API to optionally return metadata associated with the files. Specifically, this endpoint:

```
GET /ws/<workspaceId>/<folderPath>
```

Should accept a new query parameter `withMetadata=true`. When enabled, each file entry in the response should include two additional keys, `metadata` (the contents of their metadata JSON file) and `metadataStatus` to indicate any issue with a malformed metadata file.

```

Example

```json
{
  "name": "plan.seqn",
  "metadata": {...},
  "metadataStatus": "ok"
}
```

Status values

| Status | Meaning |
|------|--------|
| ok | Metadata exists and parsed |
| missing | Metadata file absent |
| malformed | Metadata JSON invalid |

- If metadata is malformed, the `metadata` value should be `null`
- Metadata **files** themselves should **not listed by default** by this endpoint.

---

# Restrictions on Metadata Files

Metadata files cannot be accessed through generic file APIs.

Workspace Service must reject

- Reading metadata files directly
- Writing metadata files directly
- Moving/renaming metadata files
- Deleting metadata files
- Uploading files matching metadata naming pattern

All metadata operations must go through the Metadata API.

---

# Out of Scope

- Metadata enforcement (read-only behavior)
- Automatic audit field updates
- Actions API support
- Metadata indexing/search
- Bulk metadata operations
- Schema migrations
- Handling filesystem drift between file and metadata

---

### LINKED ISSUE #1804: Metadata for Seqdev Workspaces - Managed fields and Read-only flag
## Summary

Follow up to #1802, which established the base SeqDev metadata structure + API. Implement enforcement of `readOnly` metadata flag and automatic management of managed metadata fields.

Includes

- Read-only file enforcement
- Automatic updates to managed metadata (`createdBy`/`createdAt`/`lastEditedBy`/`lastEditedAt`)
- Integration with Workspace Service file operations

---

## Read-Only Enforcement

Existing workspace file operations must be updated to check if the file metadata's `readOnly` flag is set before executing the operation. Specifically, if a file's metadata contains:

```json
"readOnly": true
```

- Workspace Service must **reject** file operations that mutate the file.
- This includes: File **write**, **rename**, **move** or **delete**, either individually or as a part of a bulk operation.
- These operations should still be **allowed** for files with this flag: Read file, Read metadata, Update metadata (including disabling `readOnly`), Copy file, or anything else which does not modify the underlying file.
- `readOnly` can be assumed to be `false` if the file has no metadata or the `readOnly` key is absent.

---

## Managed Metadata Fields

The Workspace Service should automatically manage the following metadata fields. These fields are service-controlled and cannot be modified through the Metadata API.

| Field | Description |
|------|-------------|
| createdBy | Username of user who created the file |
| createdAt | Timestamp of file creation |
| lastEditedBy | Last user who modified file  |
| lastEditedAt | Timestamp of last modification |

- All timestamps should be ISO 8601 timestamp format eg. `2026-03-17T00:09:34.075Z`
- `createdBy` and `createdAt` should be set automatically when the file is first created or uploaded. 
  - `createdBy` should be the username from the user's auth token
  - `createdAt` should be the creation time (of the underlying file, not the metadata file)
- Likewise, `lastEditedBy`/`lastEditedAt` are updated whenever the file is updated.
  - These fields are updated only when the file contents change, not when metadata is modified.
- Clients should not be able set these fields directly in a metadata API call, they should be entirely managed by us.

---

## Metadata Creation

Metadata files are created lazily, and may not exist yet when these operations are called. We should automatically create metadata files whenever:

- A new file is created - create metadata file populated with managed fields and `readOnly: false`.
- A file is modified - if metadata is missing, create a new metadata file with managed fields and fill them in as if the file was just created (`createdAt: <now>`)
- Metadata is set/updated - if metadata is missing, same thing, create it with fresh timestamp.

For V1, consistency between file and metadata during rename/move/copy/delete is best-effort only. We should try hard not to fail in a way that orphans metadata files, but ultimately paired file + metadata operations are not 100% atomic or crash-safe.

Mythicaeda:
* **Tickets addressed:** Closes #1802 and closes #1804
* **Review:** By commit  

## Description

- Adds new endpoints for handling metadata files and update existing file endpoints to reject directly updating metadata files
- Updates the reserved Metadata extension from `.aerie` to `.meta.seqdev`
- Updates `listContents` endpoint to hide `METADATA`-type files by default, and to provide the _contents_ of the metadata files if the `withMetadata` flag is provided
- Updates file move/copy/delete so that metadata files are consistently handled _before_ their corresponding file. Before, it varied based on endpoint
- Restricts move/copy/save/delete file endpoints to reject with a 423 Locked status if the file is marked as "readOnly"

## Verification

I'd like to add some tests that focus on the contents of the metadata files, and how they're reconstructed after deletion, as well as basic bindings tests for the endpoints.

## Documentation

No docs are invalidated, but we should add documenting these new endpoints to writing docs for the other workspace server endpoints


dandelany (changes requested):
@duranb and I have been testing this branch together with [the UI branch](https://github.com/NASA-AMMOS/plandev-ui/pull/1901) - overall things are working well! We discovered two blocker issues which should be fixed on the backend:

1. Deleting a read-only file is **correctly blocked** by the backend, but **deleting a folder containing a read-only file** is not. This can be hidden on the UI but would be best to fix on backend. For example, if you have a folder named "parent" containing a readonly file, you can send a `DELETE` to `ws/bulk/1` with the payload `["parent"]` and it will delete the folder & its contents. This should instead respond with an error if any folder contents (descendant nodes) are read-only.
2. Copying a read-only file is **incorrectly blocked** by the backend, but **should be allowed**. This one was discussed & I thought we demo'ed a fix but does not appear to be working now... Eg. if I have a read-only file called "hello.txt" and I try to copy it to a folder called "newparent", the UI `POST`s the following request to `ws/bulk/1`:
```{"copyTo":"./newparent","items":[{"path":"hello.txt"}],"overwrite":false}```
and gets the following error response:
```
[
    {
        "item": "hello.txt",
        "status": 423,
        "response": {
            "type": "INTERNAL_ERROR",
            "message": "Unable to copy 'hello.txt' in Workspace 1 to './newparent/hello.txt' in Workspace 1.",
            "timestamp": "2026-04-14T00:52:02.467623336Z",
            "service": "aerie_workspace",
            "cause": "File is currently marked as read only."
        }
    }
]
```

@Mythicaeda note that in both cases above, the UI uses the **bulk** endpoint instead of the **single** [copy/move] endpoints, but the fix should apply to both bulk & single cases.

**Comment by Mythicaeda** [comment]
For (1), blocking folder delete if there's a readonly file within the folder would implicitly mean that you cannot delete a workspace if it has a readonly file within it, which feels unintentional. Additionally, even for the normal folder case, the spec feels too vague. There's four approaches I can take to this:
1. delete up until encountering a `readOnly`, then stop when one is reached
2. search the folder (and subfolder) contents for a `readOnly` before deleting and stop if any are found
3. delete everything that isn't `readOnly` -- this one actually aligns with the behavior of `rm` and would be my preference
4. One of 1-3, as well as adding a `force` flag to delete that will skip the `readOnly` check

For (2), it looks like I forgot to push the change to `copy` after the demos -- please repull and retest

**Comment by dandelany** [comment]
Thanks @Mythicaeda - options 2 and 3 above both seem pretty reasonable. I think option 2 (check ahead of time, don't delete anything if readonly is found) is preferable for a couple reasons - despite being different from how the OS does it, it aligns with our principle of having these ops be as atomic as possible & either completely succeed or fail. It also lines up with the Clipper feedback we've gotten re: wanting as many guardrails as possible to stop them from accidentally deleting a folder they didn't mean to delete. 

If possible the error message returned in this case should also call out the paths of any read-only files it finds, to help the user track them down if they do intend to delete.

File: workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/DirectoryTree.java [hunk 1]
```diff
@@ -49,16 +59,131 @@ private static class FileNode {
...
+        Instant.parse(createdAt);
+
+        metadata.getString("lastEditedBy");
+        final var lastEditedAt = metadata.getString("lastEditedAt");
+        Instant.parse(lastEditedAt);
+
+
+        // Validate that the optional keys have the correct type, if they're present
+        if (metadata.containsKey("readOnly")) {
+          readOnly = metadata.getBoolean("readOnly");
```
dandelany, line 157: It feels weird to be setting this important piece of state only as a side-effect of a method called `validate` - can we move this up to the constructor after the `validateMetadata` call?

e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/utils/UserSetup.java
```diff
@@ -0,0 +1,43 @@
+package gov.nasa.jpl.aerie.e2e.utils;
```
dandelany, line 1: let's add a comment to this file to note it gets auto-loaded by the launcher session listener file

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -1165,12 +1324,267 @@ private JsonArray handleBulkDelete(int workspaceId, List<String> toDelete) {
...
+   * }
+   *
+   * If the metadata files contents are malformed, returns a 500 error response
+   * If the user passes a malformed set of keys (including non-existent top-level keys or a non-json object "user" field), returns a 400 error.
+   *
+   * Also takes in a query param "mergeBehavior", which can be one of "deep", "shallow", and "overwrite". Behavior is as follows:
+   *  - "deep": deep merge JSON Object params (combine nested properties)
+   *  - "shallow": shallow merge of JSON Object params (combine only top-level properties)
+   *  - "overwrite": replace JSON Object params with the new value, if provided
+   * Defaults to "deep" if not provided
```
dandelany, line 1395: should say 'defaults to shallow'

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -1165,12 +1324,267 @@ private JsonArray handleBulkDelete(int workspaceId, List<String> toDelete) {
...
+   *   "readOnly": false
+   *   "user": {
+   *     "status": "draft"
+   *   }
+   * }
+   *
+   * If the metadata files contents are malformed, returns a 500 error response
+   * If the user passes a malformed set of keys (including non-existent top-level keys or a non-json object "user" field), returns a 400 error.
+   *
+   * Also takes in a query param "mergeBehavior", which can be one of "deep", "shallow", and "overwrite". Behavior is as follows:
```
dandelany, line 1391: note that this part is only applicable to the user object, and eg. doesn't overwrite the entire metadata object

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -1165,12 +1324,267 @@ private JsonArray handleBulkDelete(int workspaceId, List<String> toDelete) {
...
+  public void deleteMetadata(final Context context) throws NoSuchWorkspaceException {
+    // Permissions Check
+    final var pathInfo = PathInformation.of(context);
+    if (!checkPermissions(context, pathInfo.workspaceId, WorkspaceAction.delete_file_directory)) {
+      return;
+    }
+
+    // Check that the underlying file exists
+    if (!workspaceService.checkFileExists(pathInfo.workspaceId, pathInfo.filePath)) {
+      context.status(404).json(new FormattedError(new NoSuchFileException(pathInfo.workspaceId, pathInfo.filePath)));
```
dandelany, line 1570: let's remove the check for the underlying file here, since this could be a useful endpoint for cleaning up "orphan" metadata files

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
```diff
@@ -106,8 +122,85 @@ boolean moveDirectory(final int oldWorkspaceId, final Path oldDirectoryPath, fin
...
+   * @throws NoSuchWorkspaceException If the specified workspace does not exist
+   * @throws WorkspaceFileOpException If "filePath" refers to a metadata file or directory, or if the metadata file for "filePath" is a directory
+   * @throws IOException If the metadata file cannot be opened for reasons other than nonexistence
+   * @throws JsonException If the file's metadata is malformed
+   */
+  boolean isReadOnly(final int workspaceId, final Path filePath)
+  throws NoSuchWorkspaceException, WorkspaceFileOpException, IOException, JsonException;
+
+  /**
+   * Get the list of readOnly files contained in the folder.
```
dandelany, line 143: note that this is for all descendants, not just children

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -216,86 +269,233 @@ public RenderType getFileType(final Path filePath) throws SQLException {
...
+  {
+    final var repoPath = postgresRepository.workspaceRootPath(workspaceId);
+    final var path = resolveReadingPath(repoPath, filePath);
+    final var file = path.toFile();
+    final var metadataFile = resolveMetadataPath(repoPath, filePath).toFile();
+
+    // Delete the metadata file if it exists
+    if(metadataFile.exists()) {
+      if(!rm(metadataFile)){
+        return false;
```
dandelany, line 420: log a warning here since this is a "shouldn't ever really happen"

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -216,86 +269,233 @@ public RenderType getFileType(final Path filePath) throws SQLException {
...
+    if (Files.exists(newPath) && Files.isSameFile(newPath, newRepoPath)) {
+      throw new WorkspaceFileOpException("Cannot replace the workspace root directory.");
+    }
+
+    // Do not try to move a directory into itself
+    if(oldWorkspaceId == newWorkspaceId && newPath.startsWith(oldPath)){
+      throw new WorkspaceFileOpException("Cannot move a directory into itself.");
+    }
+
+    return oldPath.toFile().renameTo(newPath.toFile());
```
dandelany, line 497: eventually we may change this to use `Files` API for consistency with others

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -341,87 +541,390 @@ public boolean copyDirectory(final int sourceWorkspaceId, final Path sourceFileP
...
     final var repoPath = postgresRepository.workspaceRootPath(workspaceId);
-    final var path = resolveReadingPath(repoPath, filePath);
-    final var file = path.toFile();
+    final var metadataFile = resolveMetadataPath(repoPath, filePath).toFile();
+    final var metadataFileContents = readMetadataFile(metadataFile);
+    return metadataFileContents.getBoolean("readOnly", false);
+  }
 
-    boolean success = true;
+  private boolean isReadOnly(final Path repoPath, final Path filePath)
```
dandelany, line 562: dedupe this with the previous function & have one call the other

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -341,87 +541,390 @@ public boolean copyDirectory(final int sourceWorkspaceId, final Path sourceFileP
...
+
+      // Add version
+      contents.version().ifPresentOrElse(
+          v -> generator.write("version", v),
+          () -> generator.write("version", "1"));
+
+      // Fill in "created" information, using the "metadataLastEdited" information as a fallback
+      contents.createdBy().ifPresentOrElse(
+          c -> generator.write("createdBy", c),
+          () -> generator.write("createdBy", contents.metadataLastEditedBy()));
```
dandelany, line 667: try and use `lastEditedBy` as fallback before `metadataLastEditedBy`

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -341,87 +541,390 @@ public boolean copyDirectory(final int sourceWorkspaceId, final Path sourceFileP
...
+   * @throws JsonException If the current contents are malformed.
+   */
+  private MetadataUpdates.Builder generateUpdatedMetadataFile(
+      final JsonObject currentContents,
+      MetadataUpdates updates,
+      MetadataMergeBehavior mergeBehavior
+  ) throws JsonException {
+    final var mergedBuilder = new MetadataUpdates.Builder(updates.metadataLastEditedBy(), updates.metadataLastEditedAt());
+
+    // Upsert the fields, skipping "lastEditedAt" and "lastEditedBy" (as they're already set),
```
dandelany, line 703: should say "skipping metadataLastEditedBy and metadataLastEditedBy"

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -341,87 +541,390 @@ public boolean copyDirectory(final int sourceWorkspaceId, final Path sourceFileP
...
+   * @param mergeBehavior How to merge the `user` object
+   * @return The new contents of the metadata file, expressed as a MetadataUpdates object
+   * @throws JsonException If the current contents are malformed.
+   */
+  private MetadataUpdates.Builder generateUpdatedMetadataFile(
+      final JsonObject currentContents,
+      MetadataUpdates updates,
+      MetadataMergeBehavior mergeBehavior
+  ) throws JsonException {
+    final var mergedBuilder = new MetadataUpdates.Builder(updates.metadataLastEditedBy(), updates.metadataLastEditedAt());
```
dandelany, line 701: add comment/note that these are fallbacks only representing the metadata file changed time

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -341,87 +541,390 @@ public boolean copyDirectory(final int sourceWorkspaceId, final Path sourceFileP
...
+        new MetadataUpdates.Builder(userId).build(),
+        MetadataMergeBehavior.deepMerge);
+
+    // Unset the specified keys
+    for (final var key : keysToUnset) {
+      if (key.startsWith("user.")) {
+        final var userObject = fileContentsBuilder.getUser();
+        if (userObject == null) continue;
+
+        final var path = new ArrayList<>(Arrays.asList(key.split("\\.")));
```
dandelany, line 864: let's have a comment somewhere (eg. in service interface file) that explains that nested keys are supported by that API

dandelany (changes requested):
Reviewed with @Mythicaeda - minor comments added.

Still in progress: reviewing Test files and doing some more backend testing locally

Mythicaeda:
Comments have been addressed in the 6 newest commits

dandelany (approved):
Thanks @Mythicaeda - @AaronPlave and I have reviewed the remaining code and tests & done some more local smoke testing and all looks good. Ready to merge with UI

PR #1810, opened 2026-03-27
De/serialization performance improvements and span computation improvements

joshhaug:
* **Review:** By file  

### 1. Jackson Streaming Migration
- Moved away from the legacy `javax.json` parser, testing shows `Jackson` is substantially faster. 
- Added `writeJson(JsonGenerator)` to `ValueMapper` and `writeTo(JsonGenerator)` to `SerializedValue` for direct unparsing
- In-mem objects now serialize directly to JSON streams, bypassing intermediate `JsonValue` or `JsonNode` allocations.

### 2. `SerializedValue` Optimization
- Added native `double` support to `SerializedValue` to prevent  `BigDecimal` overhead for double-precision resource samples and params. 

### 3. Simulation Engine Speedups
- `SimulationEngine` now pre-indexes serializable topics using an `IdentityHashMap` which means O(1) topic indexing. I think this is ok, but I'm curious to know what the Experts have to say. 
-  Replaced the $O(N)$ topic scan in `computeSpanInfo` with constant-time lookups

### 4. Result I/O Improvements
- `SimulationResultsWriter` is refactored to stream large results directly to disk/network using `JsonGenerator`.
- `PostgresProfileQueryHandler` now streams `jsonb` data directly

**Note:** `SerializedValue.Visitor` now requires implementation of `onDouble(double)`.

## Verification
- `SerializationBenchmark`: Micro-benchmarks comparing legacy vs. streaming paths.
- `JacksonStreamingTortureTest`: Verification of complex nested record serialization.

### DIFF
--- a/merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/JacksonStreamingTortureTest.java
+++ b/merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/JacksonStreamingTortureTest.java
@@ -1,0 +1,358 @@
+package gov.nasa.jpl.aerie.merlin.driver;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.BooleanValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.DoubleValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.IntegerValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.ListValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.LongValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.RecordValueMapper;
+import gov.nasa.jpl.aerie.contrib.serialization.mappers.StringValueMapper;
+import gov.nasa.jpl.aerie.merlin.driver.json.JsonEncoding;
+import gov.nasa.jpl.aerie.merlin.driver.resources.DeferredSerializedValue;
+import gov.nasa.jpl.aerie.merlin.framework.ValueMapper;
+import gov.nasa.jpl.aerie.merlin.protocol.model.OutputType;
+import gov.nasa.jpl.aerie.merlin.protocol.types.SerializedValue;
+import gov.nasa.jpl.aerie.merlin.protocol.types.ValueSchema;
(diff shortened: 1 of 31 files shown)


merlin-driver/src/test/java/gov/nasa/jpl/aerie/merlin/driver/JacksonStreamingTortureTest.java
```diff
@@ -0,0 +1,358 @@
...
+/**
+ * Torture test exercising the Jackson streaming serialization pipeline at scale.
+ *
+ * Verifies:
+ * 1. Correctness: writeJson output matches legacy serializeValue path for all mapper types
+ * 2. Scale: handles 200+ resources × 500 segments without issues
+ * 3. DeferredSerializedValue: lazy caching + direct streaming
+ * 4. Large composite types: deeply nested records with collections
+ */
+class JacksonStreamingTortureTest {
```
mattdailis, line 38: Any objection to calling it a "stress test"? Torture is a little harsh sounding

merlin-framework/src/main/java/gov/nasa/jpl/aerie/merlin/framework/ValueMapper.java
```diff
@@ -30,4 +33,12 @@ public interface ValueMapper<T> {
...
    */
   SerializedValue serializeValue(T value);
+
+  /**
+   * Writes a value directly to a Jackson {@link JsonGenerator}, bypassing intermediate SerializedValue allocations.
+   * The default implementation falls back to {@link #serializeValue(T)} followed by streaming write.
+   */
+  default void writeJson(T value, JsonGenerator gen) throws IOException {
+    serializeValue(value).writeTo(gen);
+  }
```
mattdailis, line 43: Some context: the purpose of SerializedValue is to provide a format-agnostic interface (so you could just as easily write to avro, orc, parquet, etc).

I'm trying to decide whether or not `writeJson` fits that goal. I suppose I would defend it a couple ways:
- it's a default method, which means you don't have to implement it unless you need to for performance reasons
- JSON is currently the only format we use in practice

mattdailis: A thought occurred to me: it seems to me the key innovation of this PR is not the ability to bypass the SerializedValue interface and write JSON directly, but the ability to serialize without building up a parallel in-memory data structure.

I think it's worth exploring an interface that does effectively the same thing as this PR at runtime, but at compile time abstracts away the "JSON" part. I think it would look very similar to writeJson, but we'd wrap `JsonGenerator` in an interface

PR #1811, opened 2026-04-01
Split E2E Test tasks into multiple subtasks

Mythicaeda:
* **Tickets addressed:** N/A
* **Review:** By commit  

## Description

Currently, the E2E Test task takes 15 minutes to execute, which is absurdly high. This PR aims to improve that number without making the tests flaky.

Breakdown of E2E Suite test speed locally (with services already setup):
- `:e2eTest:e2eTest:`: ~6 mins
- `:dbTest:e2eTest:`: ~2 mins
- `:sequencing-server:e2eTest:`: ~1 min
- `:action-server:e2eTest:`: ~5 sec

The first approach is to split the tests by suite into multiple parallel jobs. The suites should already be independent (especially as they're a mixture of Java and TS tests), but by running them in separate jobs we avoid any issues where two suites are trying to alter the DB at the same time. Additionally, by having them as separate jobs, it will be clearer which suite in particular is failing, as opposed to needing to go through the collected folder.

Breakdown of tests after this approach:
- `unitTests`: 6 min (2 min running tests, 3.5 min assembling)
- `:e2eTest:e2eTest:`: 11 min (5 min running tests, 3.5 min assembling)
- `:dbTest:e2eTest:`: 4.5 min (4 min running tests)
- `:sequencing-server:e2eTest:`: 7 min (1.5 min running tests, 3.5 min assembling the Java backend)
- `:action-server:e2eTest:`: 5 min (10s running tests, 3 min assembling the Java backend)

After that, I added the following improvements
- Removed the backend from the action server. While it requires the envvars from the `e2e-test` environment, it doesn't actually require the backend to be running
- Configure the gradle cache to cache build artifacts. It will now cache build artifacts generated by tests run on `develop`
- Cache `node_modules` for `action-server` and `sequencing-server`. These seemed to not be cached by the gradle caching task, so they're now being manually cached

Breakdown of tests after all of this:
- `unitTests`: 4.5 min (2 min running tests, 2 min assembling)
- `:e2eTest:e2eTest:`: 9 min (5 min running tests, 2 min assembling)
- `:dbTest:e2eTest:`: 4.5 min (4 min running tests)
- `:sequencing-server:e2eTest:`: 5 min (1.5 min running tests, 2 min assembling)
- `:action-server:e2eTest:`: 30s min (10s running tests)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,61 +33,13 @@ env:
   PR_BODY: "${{github.event.pull_request.body}}"
 
 jobs:
-  compile-check:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout Repo
-        uses: actions/checkout@v6
-
-      - name: Cache Sequencing NPM Packages
-        uses: actions/cache@v5
-        with:
-          path: ./sequencing-server/node_modules
-          key: sequencing-${{ runner.os }}-${{ hashFiles('./sequencing-server/package-lock.json') }}
-      - name: Cache Action NPM Packages
-        uses: actions/cache@v5
-        with:
-          path: ./action-server/node_modules
-          key: action-${{ runner.os }}-${{ hashFiles('./action-server/package-lock.json') }}
-
-      - name: Setup Java
-        uses: actions/setup-java@v5
-        with:
-          distribution: "temurin"
-          java-version: "21"
-      - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v5
-        with:
-          # This action defaults to matching on "main" and "master" as the branches allowed to write to the cache
-          cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
-      - name: Assemble
-        # Don't generate docs for tests
-        run: ./gradlew assemble -x javadoc -x javadocJar -x generateDocumentation --parallel --profile
-      - name: Upload Profile
-        if: always()
-        uses: actions/upload-artifact@v7
-        with:
-          name: ProfileResults
-          path: |
-            /home/runner/work/aerie/aerie/build/reports/profile/profile*
-
   unit-test:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v6
-
-      - name: Cache Sequencing NPM Packages
-        uses: actions/cache@v5
-        with:
-          path: ./sequencing-server/node_modules
-          key: sequencing-${{ runner.os }}-${{ hashFiles('./sequencing-server/package-lock.json') }}
-      - name: Cache Action NPM Packages
-        uses: actions/cache@v5
-        with:
-          path: ./action-server/node_modules
-          key: action-${{ runner.os }}-${{ hashFiles('./action-server/package-lock.json') }}
-
+      - name: Load NPM Cache
+        uses: ./.github/actions/load-node-cache
       - name: Setup Java
         uses: actions/setup-java@v5
         with:
@@ -98,7 +50,6 @@ jobs:
         with:
           # This action defaults to matching on "main" and "master" as the branches allowed to write to the cache
           cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
-
       - name: Assemble
         # Don't generate docs for tests
         run: ./gradlew assemble -x javadoc -x javadocJar -x generateDocumentation --parallel
@@ -145,15 +96,15 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v6
+      - name: Setup Postgres Client (psql)
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes postgresql-client
       - name: Setup Java
         uses: actions/setup-java@v5
         with:
           distribution: "temurin"
           java-version: "21"
-      - name: Setup Postgres Client (psql)
-        run: |
-          sudo apt-get update
-          sudo apt-get install --yes postgresql-client
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
         with:
@@ -176,18 +127,8 @@ jobs:
           name: DB Test Results
           path: |
             **/db-tests/build/reports/tests/e2eTest
-      - name: Print Logs for Services
-        if: always()
-        run: docker compose -f ./e2e-tests/docker-compose-test.yml logs -t
-      - name: Stop Services
-        if: always()
-        run: |
-          docker ps -a
-          docker compose -f ./e2e-tests/docker-compose-test.yml down
-          docker ps -a
-      - name: Prune Volumes
-        if: always()
-        run: docker volume prune --force
+      - name: Shutdown Backend
+        uses: ./.github/actions/shutdown-test-backend
 
   sequencing-e2e-tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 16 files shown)


.github/workflows/test.yml
```diff
@@ -33,60 +33,272 @@ env:
...
+          # This action defaults to matching on "main" and "master" as the branches allowed to write to the cache
+          cache-read-only: ${{ github.ref != 'refs/heads/develop' }}
+      - name: Assemble
+        # Don't generate docs for tests
+        run: ./gradlew assemble -x javadoc -x javadocJar -x generateDocumentation --parallel --profile
+      - name: Upload Profile
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: ProfileResults
```
dandelany: discussed with @Mythicaeda & decided to split this into it's own workflow `profile-compilation` - it's useful for testing/profiling but not to run on every test

.github/workflows/test.yml
```diff
@@ -33,60 +33,272 @@ env:
...
+      - name: Cache Sequencing NPM Packages
+        uses: actions/cache@v5
+        with:
+          path: ./sequencing-server/node_modules
+          key: sequencing-${{ runner.os }}-${{ hashFiles('./sequencing-server/package-lock.json') }}
+      - name: Cache Action NPM Packages
+        uses: actions/cache@v5
+        with:
+          path: ./action-server/node_modules
+          key: action-${{ runner.os }}-${{ hashFiles('./action-server/package-lock.json') }}
```
dandelany: Let's update this cache key to also include `'./action-server/package.json'` on the off chance a developer checks in a package.json change but *doesn't* commit package-lock. This "shouldn't happen" but might cause very confusing results if it did

.github/workflows/test.yml
```diff
@@ -33,60 +33,272 @@ env:
...
+    environment: e2e-test
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v6
+      - name: Setup NPM
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: ./action-server/.nvmrc
+          cache: npm
+          cache-dependency-path: ./action-server/package-lock.json
```
dandelany, line 78: same here re: including package.json as a part of the cache key

.github/workflows/test.yml
```diff
@@ -33,60 +33,272 @@ env:
...
+        uses: actions/cache@v5
+        with:
+          path: ./action-server/node_modules
+          key: action-${{ runner.os }}-${{ hashFiles('./action-server/package-lock.json') }}
+
+      # Start Hasura and Postgres ahead of assembling the Java code, as the Java code takes a few minutes to compile.
+      # This means Hasura will have time to complete its start-up tasks without needing an extra "sleep 30" command.
+      - name: Start Hasura and Postgres
+        run: |
+            docker compose -f ./e2e-tests/docker-compose-test.yml up -d postgres hasura
```
dandelany: add `--quiet-pull`

.github/workflows/test.yml
```diff
@@ -33,60 +33,272 @@ env:
...
+        if: always()
+        run: docker volume prune --force
+
+  java-e2e-test:
+    runs-on: ubuntu-latest
+    environment: e2e-test
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v6
+      - name: Cache Sequencing NPM Packages
```
dandelany: maybe worth looking into: is there a way to save yml steps as variables/snippets at the top & then reference them to avoid having to copy/paste these steps

dandelany (changes requested):
A few small comments below from reviewing with @Mythicaeda

.github/workflows/test.yml
Mythicaeda: I'm gonna push back on this after reading more on the `setup-node` action. `cache-dependency-path` is explicitly looking for lockfiles, it just permits you to hand it multiple lockfiles in the event of a mono-repo like our own.

I get that impression from [this section](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#running-without-a-lockfile) of the action's advanced-usage docs:
> If you choose not to use a lockfile, you must ensure that caching is disabled. The cache feature relies on the lockfile to generate a unique key for the cache entry.

.github/workflows/test.yml
Mythicaeda: I've added several [composite actions](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) to split off repeated subtasks

.github/workflows/test.yml
Mythicaeda: Done

.github/workflows/test.yml
Mythicaeda: Done

.github/workflows/test.yml
Mythicaeda: Done

.github/workflows/test.yml
dandelany: Welp I'm slightly down the rabbit hole on this one but... By my reading of docs I *do* think my proposed change would work, but it is not recommended because there is a **better way** to get what we're going for (protection against `package`/`lock` file mismatches) - using `npm ci` instead of `npm i` in CI workflows. 

This is the recommended way to do it in CI environments (for a number of reasons), it's what we do in the UI repo workflows, and `ci` helpfully throws a loud error if there is any mismatch between the `package` and `lock` files (whereas `npm i` will "helpfully" "fix" the mismatch)

For action server this is easy, for the other ones using Gradle it is [slightly more work but is a supported/recommended pattern in the gradle-node-plugin docs](https://github.com/node-gradle/gradle-node-plugin/blob/main/docs/faq.md?utm_source=chatgpt.com#how-do-i-use-npm-ci-instead-of-npm-install). I will push a minimal commit to this PR for you to peek at @Mythicaeda

dandelany (approved):
@Mythicaeda I pushed two more commits to this branch for your review:

* [8a773ed](https://github.com/NASA-AMMOS/plandev/pull/1811/commits/8a773edc5f4b6719f6478a5e24457a9ee2e6b1f7) is explained in [this comment](https://github.com/NASA-AMMOS/plandev/pull/1811#discussion_r3083253194)
* [9d4eaac](https://github.com/NASA-AMMOS/plandev/pull/1811/commits/9d4eaac0ebad354cc9c7b8bb9322e698b59e2bbd) changes the node cache used for gradle builds to be more aligned with what `setup-node` does, which is to use the **global `~/.npm` cache** rather than caching individual `node_modules` folders which are more fragile/disposable. In particular, `npm ci` *throws away* the node_modules folder but *does* make use of the global cache. Even though our services use different Node versions, it's OK for them to share this cache! The [setup-node docs mention this](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data) - "The action follows [actions/cache](https://github.com/actions/cache/blob/main/examples.md#node---npm) guidelines, and caches global cache on the machine instead of node_modules, so cache can be reused between different Node.js versions."

Let me know if you have any concerns - if not this is 👍 to merge.

Mythicaeda:
The one thing I'd like to confirm before merging is that sequencing server and action server are able to use the `~/.npm` loaded cache, but I can't seem to manually run the compilation speed check. If you're able to run this, please do and send the run my way, else I'll probably just push a temporary commit to force it to run here and then remove it before we merge

dandelany:
> I can't seem to manually run the compilation speed check

Yeah, doesn't seem like it's possible in the Github UI until some version of the workflow has been merged to main :-/ 

Feel free to push a temp commit or you may also be able to trigger it via a GH API call

dandelany:
I updated the required checks on the `develop` branch to be the new test suites, this is now 🟢 to merge.
(@Mythicaeda told me on Slack she approved)

PR #1822, opened 2026-04-30
Fix Edge Case in Anchor Validation

### LINKED ISSUE #1270: Anchored directive validation failing to catch plan bound violations in some cases
### Checked for duplicates

Yes - I've already checked

### Is this a regression?

No - This is a new bug

### Version

2.1.0

### Describe the bug

Found this issue when testing https://github.com/NASA-AMMOS/aerie-ui/pull/1066, see https://github.com/NASA-AMMOS/aerie-ui/pull/1066#issuecomment-1870555797 for the original comment. 

An anchored directive is not correctly reported as outside of plan time bounds in the following case:
1. Negatively anchor directive A to a plan which should (with https://github.com/NASA-AMMOS/aerie-ui/pull/1066) be reported as out of plan bounds.
2. Anchor directive B to directive A with a positive offset to directive A. Notice that directive B is not reported as out of bounds.
3. Negatively offset directive B to directive A. Notice that directive B will be correctly reported as out of bounds.

### Reproduction

See above

### Logs

_No response_

### System Info

```shell
N/A
```

### Severity

Moderate

Mythicaeda:
* **Tickets addressed:** Closes #1270 
* **Review:** By commit  

## Description

Previously, the "activity is before plan start" validation check would skip activities that did not have a negative start offset. This would cause it to fail to mark activities as "out of bounds" in the case where it had a positive offset to its out-of-bounds parent anchor, but not enough of a positive offset to put it back inside plan bounds. The logic has been fixed to check **all** start time anchors, not just negative start time anchors.

## Verification

A new test was added to AnchorTests that goes over the formerly-buggy edge case: `NetNegativePlanStartStatus.indirectlyInvalidChild`

Additionally, the AnchorTests have been refactored, with a focus on making them more direct and better commented. See the second commit for a more thorough list of changes.

### DIFF
--- a/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
+++ b/db-tests/src/test/java/gov/nasa/jpl/aerie/database/AnchorTests.java
@@ -492,7 +492,7 @@ public class AnchorTests {
       final var baseValidation = getValidationStatus(planId, baseActId).reasonInvalid;
       final var childValidation = getValidationStatus(planId, childActId).reasonInvalid;
 
-      // Only Base should both have a warning message
+      // Only Base should have a warning message
       assertTrue(unrelatedValidation.isEmpty());
       assertTrue(parentValidation.isEmpty());
       assertEquals("Activity Directive " +baseActId +" has a net negative offset "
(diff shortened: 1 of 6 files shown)


AaronPlave (approved):
@dandelany and I tested this locally and added one e2e test.

dandelany:
Merging so i can get tests running on #1825 - thanks!

PR #1825, opened 2026-05-01
Make Parcel Names Unique

### LINKED ISSUE #1751: Enforce unique names on sequence parcels
### Checked for duplicates

No - I haven't checked

### Is this a regression?

No - This is a new bug

### Version

v3.5.2

### Describe the bug

Turns out you can create multiple parcels with the same name, which is confusing because there's no way to tell the difference when creating a sequence/workspace.

The DB should enforce a uniqueness condition for parcel names.

### Reproduction

Create two parcels with the same name.

### Logs

```shell

```

### System Info

```shell
Clipper Aerie
```

### Severity

Minor

Mythicaeda:
* **Tickets addressed:** Closes #1751 
* **Review:** By commit  

## Description

Adds a uniqueness constraint to parcel names in the DB.

The up migration will automatically deduplicate parcel names, repeating the process until there are no more duplicate names. 

Note: I've set this PR's migration to one higher than #1822, as I expect that PR to be merged first.

## Verification

I manually set up a DB where there with multiple parcels with unique names, multiple parcels named `parcel`, and a parcel named `parcel(1)` (which is the deduplication naming pattern). I then ran the DB migration via the Python script and observed that conflicting names were correctly deduplicated.

### DIFF
--- a/deployment/hasura/migrations/Aerie/34_unique_parcel_names/down.sql
+++ b/deployment/hasura/migrations/Aerie/34_unique_parcel_names/down.sql
@@ -1,0 +1,2 @@
+alter table sequencing.parcel drop constraint parcel_name_key;
+call migrations.mark_migration_rolled_back(34);
--- a/deployment/hasura/migrations/Aerie/34_unique_parcel_names/up.sql
+++ b/deployment/hasura/migrations/Aerie/34_unique_parcel_names/up.sql
@@ -1,0 +1,31 @@
+-- Data migration: deduplicate parcel names:
+do $$
+begin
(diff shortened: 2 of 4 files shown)


dandelany (approved):
Looks good, tested including local migration (up AND down, since we made a mistake the first time!) Only (minor, nonbreaking) issue is UI - for future reference:
* We should have a better error message for this case to make it clear why it failed - just says "Failed"
* Ideally should revert the unsaved text to the last saved name and not gray out the save button on unsuccessful save.

I'll wrap these into a future UI issue

PR #1827, opened 2026-05-05
Small fixes to action server port parsing, github workflows

dandelany:
Sweeping up a couple tiny fixes I had in my notes:

* Action server was doing `parseInt(AERIE_DB_PORT, 5432),` like it was a fallback, in fact second argument is *radix* and this always returns NaN. This caused it to fallback to a default (5432) in another layer and **never** use custom `AERIE_DB_PORT` if passed
* Publish workflow should require  `init` step since it uses output from it on line 144. Also updated to make security scan scan the images from the PR branch instead of `develop`
* Pgcmp workflow had wrong paths for scripts, this only matters for auto-running the workflow if/when any of those scripts change

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -140,7 +140,7 @@ jobs:
 
   scan:
     runs-on: ubuntu-latest
-    needs: containers
+    needs: [init, containers]
     if: needs.init.outputs.SHOULD_PUBLISH_IMAGES == 'true'
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -140,7 +140,7 @@ jobs:
 
   scan:
     runs-on: ubuntu-latest
-    needs: containers
+    needs: [init, containers]
```
Mythicaeda, line 143: I'm fine with this change, but am surprised it was needed. I'd expect it to be implicitly dependent on `init`, since `containers` depends on that task.

action-server/src/threads/worker.ts
```diff
@@ -20,7 +20,7 @@ function getDbPool() {
 
   dbPool = new pg.Pool({
     host: AERIE_DB_HOST,
-    port: parseInt(AERIE_DB_PORT, 5432),
+    port: parseInt(AERIE_DB_PORT, 10),
```
Mythicaeda: Proposal: since `radix` is optional parameter, why not just exclude it?

Mythicaeda (approved):
Approving as neither of my comments are blocking

action-server/src/threads/worker.ts
dandelany: Good call - fixed

PR #1832, opened 2026-05-28
New scan workflow + sonar project config file to enable sonarqube scans

dandelany:
Per MGSS/AMMOS requirements, we need to use a combination of CodeQL + SonarQube for security scans moving forward, via GH actions. This will ultimately be implemented in all PlanDev repos & will replace the existing CodeQL scans.

This is affected by [this `nasa-scrub` issue](https://github.com/nasa/scrub/issues/118) and contains a temporary workaround to avoid it, which we can easily remove later when fixed upstream.

JPL internal docs with details: https://wiki.jpl.nasa.gov/pages/viewpage.action?spaceKey=AmmosArch&title=Code+Vulnerability+Scanning+Set+Up+for+AMMOS+Open+Source+Software

See also: https://github.com/NASA-AMMOS/plandev-ui/pull/1935


dandelany:
Known issues with this (non-blockers):
1. GH workflows are scanned but do not show the correct file source in Sonar - difficult to fix

Mythicaeda (approved):

PR #1833, opened 2026-06-01
Allow Plan Bounds to be Changed

### LINKED ISSUE #1616: Change plan start and end bounds
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

### Related problems

_No response_

I want to be able to change the plan start and end a plan after the plan has been created. This use case has come up for extending a plan, but I can imagine cases where you might want to change the plan start as well.

Mythicaeda:
* **Tickets addressed:** Closes #1616 
* **Review:** By commit  

## Description

Adds a trigger to the Plan table to cascade updates to plan bounds to dependent tables 
   - specifically: `activity_directive`, `simulation`, `simulation_dataset`, and `plan_dataset`
   - a snapshot is automatically taken when this trigger fires

Adds columns to the Plan Snapshots table to track plan boundaries at the time of snapshot creation
   - updates the Create and Restore Snapshot functions to restore plan boundaries

Adds input validation to `merlin.get_snapshot_history_from_plan(starting_plan_id)`. The function will now raise an exception if `starting_plan_id` does not refer to an existing plan.

## Verification

Database tests have been added in `PlanCollaborationTests` and `PlanBoundsTests`

Additionally, reusable code within `MerlinDatabaseTests`, `ActivityDirectiveChangelogTests`, and `PlanCollaborationTests` has been pulled in `MerlinDatabaseTestHelper`

Tests in `MerlinDatabaseTests` that formerly looked up the plan's simulation specification by spec id now use the plan id.

### DIFF
--- a/deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
+++ b/deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
@@ -812,4 +812,4 @@ begin
 end
 $$;
 
-call migrations.mark_migration_applied(35);
+call migrations.mark_migration_applied(36);
--- a/deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
+++ b/deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
@@ -206,10 +206,17 @@ create or replace procedure merlin.restore_from_snapshot(_plan_id integer, _snap
 				from diff) a;
 
 		-- Remove any added activities
-  delete from merlin.activity_directive ad
+    delete from merlin.activity_directive ad
 		using diff d
 		where (ad.id, ad.plan_id) = (d.activity_id, _plan_id)
 			and d.in_snapshot is false;
+
+    -- Update model_id and bounds of the plan
+    update merlin.plan
+    set model_id = _model_id,
+        start_time = _plan_start_time,
+        duration = _plan_duration
+    where id = _plan_id;
 
 		-- Upsert the rest
 		insert into merlin.activity_directive (
@@ -258,13 +265,6 @@ create or replace procedure merlin.restore_from_snapshot(_plan_id integer, _snap
 			where pts.snapshot_id = _snapshot_id
 			on conflict (activity_id, plan_id)
 			do update	set preset_id = excluded.preset_id;
-
-    -- Update model_id and bounds of the plan
-    update merlin.plan
-    set model_id = _model_id,
-        start_time = _plan_start_time,
-        duration = _plan_duration
-    where id = _plan_id;
 
 		-- Clean up
 		drop table diff;
--- a/deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
+++ b/deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
@@ -129,7 +129,6 @@ end$$;
 create trigger increment_plan_revision_on_directive_update_trigger
 after update on merlin.activity_directive
 for each row
-when (pg_trigger_depth() < 1)
 execute function merlin.increment_plan_revision_on_directive_update();
 
 create function merlin.increment_plan_revision_on_directive_delete()
--- a/deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
+++ b/deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
@@ -129,6 +129,7 @@ end$$;
 create trigger increment_plan_revision_on_directive_update_trigger
 after update on merlin.activity_directive
 for each row
+when (pg_trigger_depth() < 1)
 execute function merlin.increment_plan_revision_on_directive_update();
 
 create function merlin.increment_plan_revision_on_directive_delete()
--- a/deployment/postgres-init-db/sql/tables/merlin/plan.sql
+++ b/deployment/postgres-init-db/sql/tables/merlin/plan.sql
@@ -157,43 +157,56 @@ execute function util_functions.increment_revision_update();
 create function merlin.cascade_plan_bounds_update()
   returns trigger
   language plpgsql as $$
+declare
+  old_plan_end timestamptz;
+  new_plan_end timestamptz;
+  sim_start_horizon timestamptz;
+  sim_end_horizon timestamptz;
+  start_time_difference interval;
+  end_time_difference interval;
 begin
   -- prevent adjustment if the plan is locked
   if old.is_locked then
     raise exception 'Cannot adjust bounds of locked plan.';
   end if;
 
+  -- Set variables
+  old_plan_end := old.start_time + old.duration;
+  new_plan_end := new.start_time + new.duration;
+  start_time_difference := old.start_time - new.start_time;
+  end_time_difference := old_plan_end - new_plan_end;
+
   -- Take a backup snapshot
   perform merlin.create_snapshot(
       old.id,
       'Plan Bound Adjustment',
       'Automatic snapshot made before adjusting plan bounds from ' ||
-      '['|| old.start_time ||' - '|| old.start_time + old.duration || '] to ' ||
-      '[' || new.start_time || ' - ' || new.start_time + new.duration || ']',
+      '['|| old.start_time ||' - '|| old_plan_end || '] to ' ||
+      '[' || new.start_time || ' - ' || new_plan_end || ']',
       null);
 
   -- Update activities that are anchored to the plan bounds
   update merlin.activity_directive ad
-  set start_offset = start_offset + (new.start_time - old.start_time)
+  set start_offset = start_offset + start_time_difference
   where anchor_id is null
     and anchored_to_start -- anchored to plan start
     and ad.plan_id = old.id;
 
   update merlin.activity_directive ad
-  set start_offset = start_offset + (new.duration - old.duration)
+  set start_offset = start_offset + end_time_difference
   where anchor_id is null
     and not anchored_to_start -- anchored to plan end
     and ad.plan_id = old.id;
 
   -- Update associated dataset offsets (simulation and plan)
   update merlin.simulation_dataset
-  set offset_from_plan_start = offset_from_plan_start + (new.start_time - old.start_time)
+  set offset_from_plan_start = offset_from_plan_start + start_time_difference
   from merlin.simulation sim_spec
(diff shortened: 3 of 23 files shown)


dandelany:
Todo: write a test that checks that the activity relative time gets updated to maintain absolute time when plan bounds change

deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
```diff
@@ -129,6 +129,7 @@ end$$;
 create trigger increment_plan_revision_on_directive_update_trigger
 after update on merlin.activity_directive
 for each row
+when (pg_trigger_depth() < 1)
```
dandelany: Per review discussion, we are a bit nervous that preventing by depth is a "blunt instrument" that might have unintended consequences for other/future triggers. @Mythicaeda will look into other ways of targeting & preventing only the trigger we care about instead of any with depth > 1... If trigger relId or trigger tableName identifies it as coming from the plans table, we can use that to target

deployment/postgres-init-db/sql/tables/merlin/plan.sql
```diff
@@ -134,6 +134,58 @@ for each row
...
+    and not anchored_to_start -- anchored to plan end
+    and ad.plan_id = old.id;
+
+  -- Update associated dataset offsets (simulation and plan)
+  update merlin.simulation_dataset
+  set offset_from_plan_start = offset_from_plan_start + (new.start_time - old.start_time)
+  from merlin.simulation sim_spec
+  where simulation_id = sim_spec.id
+    and sim_spec.plan_id = old.id;
+
```
dandelany, line 228: Per discussion:
* we want to update sim bounds here (when updating plan bounds) - they are stored as absolute times
*  if sim bounds match plan bounds before update - update them to match new plan bounds
* else:
  - trim sim bounds so that neither begin or end of sim bounds is outside of plan bounds
  - if sim bounds are entirely outside of plan bounds after trimming, just reset sim bounds to the new plan bounds

deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
```diff
@@ -0,0 +1,660 @@
...
+-- Data Migration: fill in default info for these columns
+update merlin.plan_snapshot
+set plan_start_time = p.start_time,
+    plan_duration = p.duration
+from merlin.plan p
+where p.id = plan_id;
+
+-- Add not null argument to new columns
+alter table merlin.plan_snapshot
+alter column plan_start_time set not null,
```
dandelany, line 92: We currently don't delete plan snapshots associated with a plan when the plan is deleted. As a result, if this migration is run on a deployment where "orphan" snapshots exist, these `not null` columns will be set to null and fail... @Mythicaeda will consider best solution

dandelany (changes requested):
Generally looks good - a few changes below. Hardest thing is figuring out to do with the cascading trigger problem mentioned below... @Mythicaeda will look into options

deployment/postgres-init-db/sql/tables/merlin/plan.sql
```diff
@@ -134,6 +134,58 @@ for each row
...
+  perform merlin.create_snapshot(
+      old.id,
+      'Plan Bound Adjustment',
+      'Automatic snapshot made before adjusting plan bounds from ' ||
+      '['|| old.start_time ||' - '|| old.start_time + old.duration || '] to ' ||
+      '[' || new.start_time || ' - ' || new.start_time + new.duration || ']',
+      null);
+
+  -- Update activities that are anchored to the plan bounds
+  update merlin.activity_directive ad
```
dandelany, line 210: @Mythicaeda - I was testing locally with @AaronPlave today and we are seeing a bug that I think is an issue with this core offset math:

1. Create a plan with bounds of eg. Day 100 - 110 of 2026 (in DOY format)
2. Add an activity directive on day 103 (ie. anchored to Plan Start with an offset of 3 days)
3. Now change plan bounds to be Day 90 - 110 (ie widen it by 10 days at the beginning)
4. I expect the activity directive to **still** be at day 103 (same absolute time), and to have a new offset of 13 days. Instead it is outside of the new plan bounds, at day 83 and has an offset of -7 days.

Is this supposed to be `(old.start_time - new.start_time)` instead?

In any case, this is the core functionality of this ticket, so we should have some tests that verify the absolute time is preserved in these cases.

AaronPlave:
Activities anchored to the end of the plan don't appear to retain their absolute times

Mythicaeda:
> Activities anchored to the end of the plan don't appear to retain their absolute times

This bug has been fixed.

deployment/hasura/migrations/Aerie/36_change_plan_bounds/up.sql
Mythicaeda: A step has been added to the up migration to first try and recover the snapshot's bounds by using related plans. For example, if snapshot's plan has been deleted, but that deleted plan had a branch that still exists, we can use the branch's bounds as the snapshot's bounds, since a branch and parent **must** have had the same bounds.

As per discussion with @AaronPlave, if **every** plan related to that snapshot has been deleted, the bounds will be set to the following fallback values: 
`plan_start_time = '1950-01-01 T00:00:00Z', plan_duration = '00:00:00'`
These values were picked so that it is obvious that this is a leftover snapshot.

deployment/postgres-init-db/sql/tables/merlin/plan.sql
Mythicaeda: Core offset math has been adjusted and tests have been added to verify that it is properly working.

deployment/postgres-init-db/sql/tables/merlin/activity_directive/activity_directive.sql
Mythicaeda: In order to bypass this change, I moved `cascade_plan_bounds_update` from a `BEFORE UPDATE` trigger to an `AFTER UPDATE` trigger. As a consequence, I had to move the call to `take_snapshot` to a new trigger that still runs `BEFORE UPDATE` (otherwise the snapshot would have the updated plan boundaries).

Note: none of the trigger related fields identified what called the change. They all solely referred to the `activity_directive` table that the trigger runs on. What these fields are probably useful for is if we have a single trigger function that we use on multiple tables (ie, `util_functions.increment_revision`) and we wanted to determine which table's trigger called the function (ie, for logging purposes)

deployment/postgres-init-db/sql/tables/merlin/plan.sql
Mythicaeda: Sim spec bounds are now being updated as specified. Tests have been added to verify that this "trim to new plan bounds" behavior only occurs when the sim spec bounds fall outside of the new plan bounds (and only for the relevant boundaries)

deployment/postgres-init-db/sql/tables/merlin/plan.sql
```diff
@@ -154,6 +154,138 @@ for each row
...
+      'Plan Bound Adjustment',
+      'Automatic snapshot made before adjusting plan bounds from ' ||
+      '['|| old.start_time ||' - '|| old_plan_end || '] to ' ||
+      '[' || new.start_time || ' - ' || new_plan_end || ']',
+      null);
+  return new;
+end;
+$$;
+
+create trigger take_snapshot_before_plan_bounds_update
```
dandelany, line 183: I think there's a bug hiding around here related to restoring plan snapshots after changing plan bounds. 

To reproduce:
1. Make a plan, with plan bounds eg. 2026-150 to 155
2. Add some activities anchored to plan start with 1d and 2d offsets, so their absolute dates are eg. 151 and 152
3. Go to Plan Metadata and change the start date to 149, 1 day earlier.
4. All good so far - activities now have offsets of 2d and 3d, correctly maintaining their absolute dates, 151 and 152.
5. Now, look at the Plan Snapshot that was created. Click "Preview" on it, everything still looks correct.
6. Click "Restore" on the Plan Snapshot - the old plan bounds are restored, but the activities are **updated incorrectly**, they now have offsets 0d and 1d, and **changed absolute dates** to 150 and 151.

It's not immediately obvious what's causing this but I'll do some more poking, just wanted to call it out ASAP since it's a blocker. It's almost like the correct math/adjustment is being applied twice instead of once, but not sure.

Also note that the same issue occurs if you change the **end date** of the plan with **end-anchored** activities (+ then restore the snapshot)

dandelany (changes requested):
I've been testing a lot locally, and confirmed the issues we saw re: maintaining absolute activity time have been fixed 🎉  However, there's a another more subtle (but equally important) bug I'm seeing re: activity times when restoring plan snapshots - details below.

deployment/postgres-init-db/sql/tables/merlin/plan.sql
Mythicaeda: I know what's causing it.

Originally, I was updating the plan bounds **before** replacing all of the activities in `restore_snapshot`. This meant that, in the function, you'd have the following flow:
1. the plan bounds would be updated,
2. because of step 1, all of the activities on the plan have their offsets adjusted
3. activities that do not exist on the snapshot are deleted
4. activities that exist on the snapshot are overwritten by their snapshotted versions.
5. the union of the set of the activities on the plan in step 2 and updated or inserted by step 4 have their anchor validations refreshed

However, this was causing issues when restoring a plan where the bounds have changed _and_ activities have been added to the plan since the snapshot was created. Specifically, the issue was in step 5 when it would try to update the validation for activities that were removed during step 3. 

Initially, I decided to fix this by moving the `update plan bounds` step to be after step 4, but that causes the issue you saw, where the updated activities then have their start offsets updated. Instead, I'm going to move the `update plan bounds` step to happen between steps 3 and 4. This will fix the bug you saw without reintroducing the bug I was trying to fix.

Mythicaeda: Fix has been applied and the `restoresPlanBounds` test in `RestorePlanSnapshotTests` has been updated to show that it is effective.

dandelany: 👍 thanks, re-tested and confirmed this is working

dandelany (approved):
👍 looking good! Did some more testing with @AaronPlave and no remaining issues found. Ready to merge along with the UI PR

PR #1836, opened 2026-06-12
Add simultaneous-edit protection to workspace file saves

AaronPlave:
* **Tickets addressed:** UI Ticket: https://github.com/NASA-AMMOS/plandev-ui/issues/1789
* **Review:** By File
* **Merge strategy:** Squash

## Description
Adds simultaneous-edit protection to sequencing-workspace file saves so one user can't silently overwrite another user's changes. UI PR [here](https://github.com/NASA-AMMOS/plandev-ui/pull/1949) – adds UI support for warning user of the conflict and allowing them to select the desired resolution.

The approach in the backend uses standard HTTP `ETag`/`If-Match`:
- `GET /ws/{id}/{path}` now returns an `ETag` (a hash of the file's content performed at the time of the GET). The client sends this `ETag` back as `If-Match` when saving.
- On `PUT`, if `If-Match` no longer matches the file on disk (via `ETag` comparison), the save is rejected with HTTP `412` and a small response body describing the conflict (current tag, last editor, time), or `reason: "deleted"` if the file is gone. `If-Match: *` (or none) forces an overwrite.

A few notes:
- A present `If-Match` takes precedence over the existing "already exists" HTTP `409` so normal edit-saves run the version check.
- `saveFile` hashes the upload as it streams to disk and returns the new tag so we never re-read the file to produce it.
- The hash lives in one place so it can later be backed by a git object id without changing the API.
- Bulk uploads are unchanged (they send no `If-Match` and keep their existing overwrite/`409` behavior).

## Verification
- additional unit tests in `WorkspaceFileSystemServiceTest.java`
- full e2e tests left to the UI PR


workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
```diff
@@ -431,12 +431,13 @@ private void getFileDirectory(Context context) throws NoSuchWorkspaceException {
       }
 
       try {
-        final var fileStream = workspaceService.loadFile(pathInfo.workspaceId, pathInfo.filePath());
-        final var inputStream = fileStream.readingStream();
+        final var fileContent = workspaceService.loadFileWithETag(pathInfo.workspaceId, pathInfo.filePath());
```
AaronPlave: Rename back to `loadFile`

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -299,15 +303,56 @@ public boolean deleteWorkspace(final int workspaceId) throws NoSuchWorkspaceExce
 
   //region File Operations
   @Override
-  public FileStream loadFile(final int workspaceId, final Path filePath) throws IOException, NoSuchWorkspaceException {
+  public FileContent loadFileWithETag(final int workspaceId, final Path filePath)
```
AaronPlave: Revert back to using an input stream record and compute etag in a streamable way.

https://stackoverflow.com/questions/304268/getting-a-files-md5-checksum-in-java

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
```diff
@@ -19,13 +23,43 @@
  * An interface that defines how the Aerie system can interact with the Workspaces backend.
  */
 public interface WorkspaceService {
+  /**
+   * Compute a byte array into a strong Entity Tag (lowercase hex) using the SHA-256 algorithm.
+   */
+  static String computeETag(final byte[] content) {
+    return eTagFromDigest(newSHA256Digest().digest(content));
```
dandelany, line 31: Reading through this with @AaronPlave - @Mythicaeda can you help us understand this part? The parameter name implies that the actual content of the file should be passed in & this function would compute the hash. However, where this is called (in `getETag()` within `WorkspaceFileSystemService.java`), the caller seems to also be computing a digest and passing it to this function. Are we doing a hash of a hash here?

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
```diff
@@ -299,15 +302,68 @@ public boolean deleteWorkspace(final int workspaceId) throws NoSuchWorkspaceExce
...
+    }
+
+    return new FileStream(new FileInputStream(file), file.getName(), Files.size(file.toPath()), getETag(path));
+  }
+
+  @Override
+  public String getETag(final int workspaceId, final Path filePath)
+  throws IOException, NoSuchWorkspaceException, NoSuchFileException, WorkspaceFileOpException {
+    final var path = resolveReadingPath(workspaceId, filePath);
+    if(filePath.toFile().isDirectory()) {
```
AaronPlave: Should we use `Files.isDirectory(path)` instead here and below since i think `filePath` always resolves to CWD?

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
dandelany: I think this check can be removed now that `loadFile` checks for the file's existence and throws the `NoSuchFileException` if it's missing

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceService.java
Mythicaeda: Whoops, you're correct, I've fixed it. Doing a hash of a hash was **not** the intent, I just forgot to update those usages to directly call `eTagFromDigest`

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: We **should** use `Files.isDirectory(path)` here because `path` is the actual, resolved path we want to be checking. I've corrected that bug and updated other usages of `path.toFile().isDirectory()` to use `Files.isDirectory()` for consistency

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
dandelany, line 321: Sonar scan complains here about opening a FileInputStream without a fallback to close it - reading more I'm not sure we're ever closing it. Should we be, and if so, when/where? This could be silently leaking resources/file-descriptors which may not be an immediate problem but could show up at scale/with long-running instances.

dandelany (commented):
Pushed one small comment & two questions below - @Mythicaeda let me know if you think the file-closing thing should be a blocker or something we fix more widely in a separate PR.

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceFileSystemService.java
Mythicaeda: Javalin auto-closes the `FileInputStream` when its done using it, so we don't have to worry about it leaking resources. I've added comments clarifying this to the `getFile` and `getMetadataFile` endpoints.

(Wish I could point you at the line of Javalin's docs that _says_ it does this automatically, but they don't say it explicitly, so... it's called at the end of the [use block](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/use.html) on L99 of `JavalinServlet.kt`)

workspace-server/src/main/java/gov/nasa/jpl/aerie/workspace/server/WorkspaceBindings.java
Mythicaeda: That makes sense, the extra check has been removed.

Mythicaeda:
> Pushed one small comment

Thanks for adding the comment! I tweaked the wording of it slightly to better explain why Javalin can't autogenerate the etag for files -- it's not a static vs dynamic file issue, it's specifically to do with the use of an `InputStream` as opposed to a `JSONObject` or `String`.

I've also added the requested test that proves that the 1MB chunking does not impact ETag generation. (Related: let me know if you have an opinion on the chunking size, I picked 1MB kind of arbitrarily)

dandelany (approved):
All notes above have been addressed - re-tested with the latest UI and everything is working as expected. Ready to merge!

PR #1837, opened 2026-06-17
Optionally Disable Expansion if Constraints Violated

pranav-super:
* **Tickets addressed:** N/A; LRO MOT Request
* **Review:** By commit  

## Description
The issue in question involves [adding a warning to the UI](https://github.com/NASA-AMMOS/plandev-ui/pull/1956) to notify planners when there are violations on their plan before expansion.

Here, a 'headless' analogue was added, which throws an error when there are constraint violations during a call to `expandAllTemplates` if they are not explicitly bypassed.

The default is to bypass this check, so as to avoid interference with past behavior.

Currently, erroring is only thrown when there are violations for a given simulation ID. Checks that are made on the UI side, like the simulation being up to date, unchecked constraints, and the like, are not done here as erroring for those other issues is not considered necessary.

## Verification
They were tested locally, both with and without the UI updates (where removal of the UI update and explicitly setting `bypassConstraints` to false activated this feature).

## Documentation
None.

## Future work
None anticipated.

### DIFF
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -78,7 +78,7 @@ type Mutation {
     seqIds: [String!]!,
     simulationDatasetId: Int!,
     modelId: Int!,
-    bypassConstraints: Bool, // (optional)
+    bypassConstraints: Boolean,
   ): ExpandAllSequencesResponse
 }
 
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -77,7 +77,8 @@ type Mutation {
   expandAllTemplates(
     seqIds: [String!]!,
     simulationDatasetId: Int!,
-    modelId: Int!
+    modelId: Int!,
+    bypassConstraints: Bool, // (optional)
   ): ExpandAllSequencesResponse
 }
 
--- a/deployment/hasura/metadata/actions.graphql
+++ b/deployment/hasura/metadata/actions.graphql
@@ -70,7 +70,7 @@ type Mutation {
 }
 
 type Mutation {
-  expandAllActivities(expansionSetId: Int!, simulationDatasetId: Int!): ExpandAllActivitiesResponse
+  expandAllActivities(expansionSetId: Int!, simulationDatasetId: Int!, bypassConstraints: Boolean): ExpandAllActivitiesResponse
 }
 
 type Mutation {
--- a/sequencing-server/src/routes/command-expansion.ts
+++ b/sequencing-server/src/routes/command-expansion.ts
@@ -1,7 +1,7 @@
 import type { UserCodeError } from '@nasa-jpl/aerie-ts-user-code-runner';
 import pgFormat from 'pg-format';
 import type { Context } from '../app.js';
-import { db, piscina, promiseThrottler, typeCheckingCache } from './../app.js';
+import { db, graphqlClient, piscina, promiseThrottler, typeCheckingCache } from './../app.js';
 import { Result } from '@nasa-jpl/aerie-ts-user-code-runner/build/utils/monads.js';
 import express from 'express';
 import { serializeWithTemporal } from './../utils/temporalSerializers.js';
@@ -25,6 +25,7 @@ import { stringifyActivity } from '../lib/mustache/util/activity.js';
 import { stolBuilder } from '../builders/stolBuilder.js';
 import { concatBuilder } from "../builders/concatBuilder.js";
 import { SequencingLanguage } from '../lib/mustache/enums/language.js';
+import { gql } from 'graphql-request';
 
 const logger = getLogger('app');
 
@@ -359,6 +360,7 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
 
   //  0. Extract stuff from request
   // needed to uniquely identify sequence templates, along with activity type
+  const bypassConstraints = req.body.bypass !== undefined ? req.body.bypass as boolean : true;
   const modelId = req.body.input.modelId as number;
   const simulationDatasetId = req.body.input.simulationDatasetId as number;
   const seqIds = (req.body.input.seqIds as number[]).filter((val, index, arr) => arr.indexOf(val) == index); // remove duplicates, if they're even possible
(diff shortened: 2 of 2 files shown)


deployment/hasura/metadata/actions.graphql
```diff
@@ -77,7 +77,8 @@ type Mutation {
   expandAllTemplates(
     seqIds: [String!]!,
     simulationDatasetId: Int!,
-    modelId: Int!
+    modelId: Int!,
+    bypassConstraints: Bool, // (optional)
```
AaronPlave: this comment seems to be preventing successful compilation

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -359,6 +360,7 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
 
   //  0. Extract stuff from request
   // needed to uniquely identify sequence templates, along with activity type
+  const bypassConstraints = req.body.bypass !== undefined ? req.body.bypass as boolean : true;
```
AaronPlave: Should this be `req.body.input. bypassConstraints`?

deployment/hasura/metadata/actions.graphql
AaronPlave: And shouldn't it be `Boolean`?

sequencing-server/src/routes/command-expansion.ts
AaronPlave: And should it default to true or false? Opt-in vs opt-out?

AaronPlave:
Any reason not to also have this gating on non-template expansion?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -367,6 +369,52 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
...
+            }
+          }
+        }
+      `,
+      {
+        simulationDatasetId: simulationDatasetId
+      },
+    );
+  
+    console.log(simulationDatasetId, JSON.stringify(constraint_run), "\n", constraint_run.at(constraint_run.length-1), "\n", constraint_run.at(constraint_run.length-1)?.results.results.violations);
```
AaronPlave: leftover log?

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -367,6 +369,52 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
...
+    if (constraint_run.length === 0) {
+      throw new Error(
+        `POST /command-expansion/expand-all-sequence-templates: Expansion for simulation dataset ${simulationDatasetId} failed, as violations haven't been checked yet.`
+      );
+    } 
+    if (constraint_run.at(constraint_run.length-1)?.results.results.violations.length) {
+      const numViolations = constraint_run.at(constraint_run.length-1)?.results.results.violations.length;
+      throw new Error(
+        `POST /command-expansion/expand-all-sequence-templates: Expansion for simulation dataset ${simulationDatasetId} failed, as there ${numViolations ?? 0 > 1 ? "are" : "is"} still ${numViolations} violation${numViolations ?? 0 > 1 ? "s" : ""}.`
```
AaronPlave: Think you need parens around this clause:
`(numViolations ?? 0) > 1`

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -367,6 +369,52 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
     simulationDatasetId
   }
 
+  //  0b. [OPTIONAL] Verify that for the given simulationDatasetId, constraints are up to date
+  if (!bypassConstraints) {
+    // not checking if simulation is out of date, as that leads to a lot of extra Hasura calls and processing here, and this expansion call is for a specific simulation dataset.
+    // additionally, not checking unchecked cosntraints, as we shouldn't throw an error if anything is unchecked.
+    // we only care about if there are violated constraints:
+    const { constraint_run } = await graphqlClient.request<{
```
AaronPlave: So each constraint has an independent run (scoped to a sim dataset) so we need to check validation status of _all_ constraint runs for the latest constraint request. Something like this ish:


```
//  0b. [OPTIONAL] Verify constraints are satisfied for the given simulationDatasetId.
if (!bypassConstraints) {
  // We only block on *violations*. We intentionally don't fail on stale sims,
  // unchecked constraints, or constraint errors — only on confirmed violations
  // in the most recent constraint request for this simulation dataset.
  const { constraint_request } = await graphqlClient.request<{
    constraint_request: {
      constraints_run: {
        results: {
          results: { gaps: object[]; violations: object[] };
        };
      }[];
    }[];
  }>(
    gql`
      query GetLatestConstraintViolations($simulationDatasetId: Int!) {
        constraint_request(
          where: { simulation_dataset_id: { _eq: $simulationDatasetId } }
          order_by: { requested_at: desc }
          limit: 1
        ) {
          constraints_run {
            results {
              results
            }
          }
        }
      }
    `,
    { simulationDatasetId },
  );

  const latestRequest = constraint_request[0];
  if (latestRequest === undefined) {
    throw new Error(
      `POST /command-expansion/expand-all-sequence-templates: Expansion for simulation dataset ${simulationDatasetId} failed, as constraints haven't been checked yet.`,
    );
  }

  const numViolations = latestRequest.constraints_run.reduce(
    (total, run) => total + (run.results?.results?.violations?.length ?? 0),
    0,
  );

  if (numViolations > 0) {
    throw new Error(
      `POST /command-expansion/expand-all-sequence-templates: Expansion for simulation dataset ${simulationDatasetId} failed, as there ${
        numViolations > 1 ? 'are' : 'is'
      } still ${numViolations} violation${numViolations > 1 ? 's' : ''}.`,
    );
  }
}
```

AaronPlave (changes requested):

sequencing-server/src/routes/command-expansion.ts
pranav-super: It should default to opt-out, meaning bypass would be set to true. We'd only want to fail on constraints in the sequencing server if explicitly asked!

deployment/hasura/metadata/actions.graphql
pranav-super: This has been corrected!

sequencing-server/src/routes/command-expansion.ts
pranav-super: This modification has been made, and also duplicated for normal expansion as well.

pranav-super (commented):
(Responded)

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -367,6 +369,64 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
...
+    const numViolations = latestRequest.constraints_run.reduce(
+      (total, run) => total + (run.results?.results?.violations?.length ?? 0),
+      0,
+    );
+
+    if (numViolations > 0) {
+      throw new Error(
+        `POST /command-expansion/expand-all-sequence-templates: Expansion for simulation dataset ${simulationDatasetId} failed, as there ${
+          (numViolations > 1) ? 'are' : 'is'
+        } still ${numViolations} violation${(numViolations > 1) ? 's' : ''}.`,
```
Mythicaeda, line 424: Extremely minor wording nitpick: why the `still` here? Assuming you don't change your constraints to fit your sim results (which seems like bad practice), you'd need a separate sim run for those violations to not apply, wouldn't you?


```suggestion
        } ${numViolations} violation${(numViolations > 1) ? 's' : ''}.`,
```

sequencing-server/src/routes/command-expansion.ts
Mythicaeda, line 356: Please update this comment to include the added argument


```suggestion
/**
  * ARGUMENTS
  * {
  *    modelId: Int!,
  *    simulationDatasetId: Int!,
  *    seqIds: [Int!]!,
  *    bypassConstraints: Boolean
  * }
  */
```

sequencing-server/src/routes/command-expansion.ts
```diff
@@ -359,6 +360,7 @@ commandExpansionRouter.post('/expand-all-sequence-templates', async (req, res, n
 
   //  0. Extract stuff from request
   // needed to uniquely identify sequence templates, along with activity type
+  const bypassConstraints = req.body.input.bypassConstraints !== undefined ? (req.body.input.bypassConstraints as boolean) : true;
```
Mythicaeda, line 363: Given that the **default** behavior is to bypass constraints, it makes more sense logically for this argument to be inverted. That is, it's called something like `referenceConstraints` or `enforceConstraints` and defaults to `false`. As a user, it'd be very weird to have to input `variable = false` to activate new behavior.

sequencing-server/src/routes/command-expansion.ts
Mythicaeda, line 1: I'd like to request six e2e tests, which are each of these updated endpoints run under the following conditions:

1. there is a constraint run with violations on the sim dataset
2. there is a constraint run without violations on the sim dataset
3. there are no constraints run on the sim dataset

In all three cases, the endpoint is set to be checking that there are no constraint violations.

dandelany:
@Mythicaeda is picking the backend work here and will go through [the slides linked here](https://github.com/NASA-AMMOS/plandev-ui/issues/1971#issuecomment-5051706405) (especially Conclusions slide) and determine what changes need to be made on the backend. Generally we're thinking we'll keep the existing constraint violations check (with flag renamed to `force` and defaulted to `false`), and add any additional checks related to simulation status being not successful or out of date.

PR #1839, opened 2026-06-30
Fix `Cascade` Delete Failing when handling Anchored Activities

### LINKED ISSUE #1797: Scheduler Fails to `Cascade` Delete When Placed Activities are Anchored to Each Other
This bug occurs when a procedural scheduling goal is set to delete activities it placed in a prior scheduling with the anchor resolution strategy set to `Cascade`. The scheduler deletes goals from a prior run by iterating over the activities placed by the goal and deleting them one-by-one, handling anchors as it deletes the anchoring activity. This means that if one of the later activities is anchored to an earlier one, when the earlier one is removed, the later one is deleted in that same iteration of the loop. When this deletion occurs, the "placed by previous run" iterator isn't updated and the scheduler will attempt to delete that later activity a second time, throwing an exception when it can't find the already-deleted activity.

The method `Procedure#shouldDeletePastCreations` needs to be updated to properly handle anchored activities.

Reproduction steps:
Run scheduling twice using the following procedural goal with the Bananation model:
```java
@SchedulingProcedure
public record AnchorCascadeDeleteGoal() implements Goal {

  @Override
  public ActivityAutoDelete shouldDeletePastCreations(
      @NotNull final Plan plan,
      @Nullable final SimulationResults simResults) {
    // Delete activities created by previous runs of this goal, using Cascade to
    // handle anchored activities
    return new ActivityAutoDelete.AtBeginning(DeletedAnchorStrategy.Cascade, false);
  }

  @Override
    final var ids = new ActivityDirectiveId[3];

    ids[0] = plan.create(
        "BiteBanana",
        new DirectiveStart.Absolute(Duration.HOUR),
        Map.of("biteSize", SerializedValue.of(1))
    );
    ids[1] = plan.create(
        "BiteBanana",
        new DirectiveStart.Anchor(ids[0], Duration.HOUR, DirectiveStart.Anchor.AnchorPoint.End),
        Map.of("biteSize", SerializedValue.of(2))
    );
    ids[2] = plan.create(
        "BiteBanana",
        new DirectiveStart.Anchor(ids[1], Duration.HOUR, DirectiveStart.Anchor.AnchorPoint.Start),
        Map.of("biteSize", SerializedValue.of(3))
    );

    plan.commit();
  }
}
```

Mythicaeda:
* **Tickets addressed:** Closes #1797 
* **Review:** By commit  

## Description

A new method has been added to the `EditablePlan` interface: `deleteIfExists`. Unlike `delete`, if the activity does not exist, then the method silently returns instead of throwing an Exception. 

`Procedure.deletePastCreations` has been updated to call this new method instead of `delete`. Additionally, the way the method gets the set of ids to be deleted has been simplified: rather than iterating through every activity on the plan and checking that a) it has a corresponding entry in the `sourceSchedulingGoals` map and b) that entry corresponds to the `Procedure`'s `goalId`, it now directly filters the `sourceSchedulingGoals` for the keys that have the `Procedure`'s `goalId` as a value.

## Verification

A test going through the reproduction steps on the ticket were added.

### DIFF
--- a/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/AnchorCascadeDeleteGoal.java
+++ b/e2e-tests/src/main/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/procedures/AnchorCascadeDeleteGoal.java
@@ -1,0 +1,53 @@
+package gov.nasa.jpl.aerie.e2e.procedural.scheduling.procedures;
+
+import gov.nasa.ammos.aerie.procedural.scheduling.ActivityAutoDelete;
+import gov.nasa.ammos.aerie.procedural.scheduling.Goal;
+import gov.nasa.ammos.aerie.procedural.scheduling.annotations.SchedulingProcedure;
+import gov.nasa.ammos.aerie.procedural.scheduling.plan.DeletedAnchorStrategy;
+import gov.nasa.ammos.aerie.procedural.scheduling.plan.EditablePlan;
+import gov.nasa.ammos.aerie.procedural.timeline.payloads.activities.DirectiveStart;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.Plan;
+import gov.nasa.ammos.aerie.procedural.timeline.plan.SimulationResults;
+import gov.nasa.jpl.aerie.merlin.protocol.types.Duration;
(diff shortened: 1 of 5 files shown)


dandelany (approved):
Thanks @Mythicaeda ! I've tested this manually with a few different variations of procedures and I haven't seen any errors. Will merge after rebase/tests 🚀

PR #1841, opened 2026-07-02
Reuse visited set in NamingContext

DavidLegg:
Opening on behalf of Brad, who wrote this code with the help of Claude.

This code intends to reduce memory allocations by reusing the context set in Naming.


DavidLegg:
After reading this code carefully, I believe it's safe and correct, but it's not clear to me how much, if any, memory it saves. Certainly it reduces allocations and copying, which likely reduces runtime a bit.
However, the NamingContext is an ephemeral object we use to compute a name, so they shouldn't be contributing to retained memory.

PR #1845, opened 2026-07-06
Timeline trimming: release old SlabList entries during simulation

DavidLegg:


merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/timeline/LiveCells.java
```diff
@@ -65,6 +65,39 @@ private <State> Cell<State> getCell(final Query<State> query) {
...
+
+  /** Returns the set of all queries known to this LiveCells (local + parent). */
+  private Iterable<Query<?>> allQueries() {
+    if (this.parent == null) return this.cells.keySet();
+    // Combine local and parent queries
+    final var all = new HashMap<Query<?>, LiveCell<?>>(this.cells);
+    for (final var query : this.parent.allQueries()) {
+      all.putIfAbsent(query, null); // null value just to collect keys
+    }
+    return all.keySet();
```
DavidLegg, line 98:
```suggestion
    final var all = new HashSet<Query<?>, LiveCell<?>>(this.cells.keySet());
    all.putAll(this.parent.allQueries());
    return all;
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SlabList.java
```diff
@@ -115,4 +116,46 @@ public SlabList<T> duplicate() {
...
+   */
+  public int advanceHead(final int slabsToTrim) {
+    int trimmed = 0;
+    for (int i = 0; i < slabsToTrim; i++) {
+      final var nextSlab = this.head.next().getValue();
+      if (nextSlab == null) break;
+      trimmed += this.head.elements().size();
+      // Clear old slab's references to help GC
+      this.head.elements().clear();
+      this.head.next().setValue(null);
```
DavidLegg, line 133: I doubt the GC needs this help. This removes references from unused object to used objects. The GC is capable of understanding the direction of references and ignoring these itself.
```suggestion
```

merlin-driver/src/main/java/gov/nasa/jpl/aerie/merlin/driver/engine/SlabList.java
DavidLegg, line 58: Minor point, but if we make `head` a mutable variable, we should update these methods not to depend on it. Reverting to object-identity by deleting these methods is likely good enough.

DavidLegg (commented):
I'd like to discuss this PR with @mattdailis. Fundamentally, it's a change we've wanted to make for a long time, trimming the event graph as we go. I want to double check my understanding of the event graph system before we go forward, though. Bugs in that system can be subtle (often too subtle for the testing to catch quickly) and have far-reaching, hard-to-debug consequences.

That said, if we can safely trim down the event graph as we go, I believe that could have a dramatic effect on retained memory.

DavidLegg:
This PR is superceded by https://github.com/NASA-AMMOS/plandev/pull/1847 - we should close this PR if/when that one goes through.

Copying my message on that PR:
> This replaces https://github.com/NASA-AMMOS/plandev/pull/1845 - After Matt and I talked through the existing Aerie code and Claude's work, we decided that starting from scratch and using Matt's prior work and Claude's work as references would give us the cleanest results the fastest.
> 
> This still needs to be profiled to make sure it actually saves on retained memory. We don't expect speed improvements.

DavidLegg:
Closing since #1847 was merged instead.

PR #1846, opened 2026-07-06
Fix `IllegalStateException` when Simulating during Scheduling

### LINKED ISSUE #1786: `IllegalStateException` when Simulating during Scheduling
Under specific circumstances, the method `PlanSimCorrespondence#equalsWithIdMap` in the Scheduler's `SimulationFacade` interface can throw an `IllegalStateException`. This method performs an id-agnostic comparison between two plans. It's used as part of an optimization in the `CheckpointSimulationFacade#simulateWithResults` method to skip resimulating the plan if it is equivalent to a previously-simulated version of the plan.

Specifically, `equalsWithIdMap` will throw an `IllegalStateException` (specifically a `Duplicate Key Violation`) when one of the plans being compared contains two activities with the same arguments and start times (which I'll be calling "duplicate activities" in this ticket). 

This can be fixed in one of two ways:
1. Update how the comparison is done to handle duplicate activities.
2. If duplicate activities are detected, skip the comparison and return `false`.

To reproduce the exception:
1. Create a new Bananation plan
2. Put the following Scheduling Goal on the plan's scheduling specification
3. Schedule the plan twice (alternatively, put the goal on the spec twice and schedule the plan once)

```java
@SchedulingProcedure
public record BugTestSchedulingGoal() implements Goal
{
  @Override
    final var ids = new ActivityDirectiveId[3];

    ids[0] = plan.create(
        "parent",
        new DirectiveStart.Absolute(Duration.HOUR),
        Map.of()
    );
    ids[1] = plan.create(
        "child",
        new DirectiveStart.Anchor(ids[0], Duration.HOUR, DirectiveStart.Anchor.AnchorPoint.End),
        Map.of()
    );
    ids[2] = plan.create(
        "grandchild",
        new DirectiveStart.Anchor(ids[1], Duration.HOUR, DirectiveStart.Anchor.AnchorPoint.Start),
        Map.of()
    );

    plan.simulate();
    plan.commit();
    //plan.simulate();
  }
}
```

Mythicaeda:
* **Tickets addressed:** Fixes #1786 
* **Review:** By commit  

## Description

The method `PlanSimCorrespondence#equalsWithIdMap` has been reworked. Before, it was implicitly assuming that every activity on the plan was distinct in some way. It no longer makes that assumption, instead making the value in the inverted maps a `List` of `ActivityDirectiveId`s instead of just a single `ActivityDirectiveId`.

Additionally, the method now fast-fails when the number of directives in the two `PlanSimCorrespondence`s differ. Before, if the only difference between two `PlanSimCorrespondence`s was a new activity, the method would only realize that _after_ it had done all the work of populating the `result` map.

## Verification

A new e2e test was added that schedules the procedural goal `DumbRecurrenceGoal` twice. `DumbRecurrenceGoal` is a goal that will **always** trigger this bug when it is scheduled more than once. This test fails on `develop` but passes on this branch.

A manual test can be done for this bug with the following Verification steps:
1. Deploy the `develop` branch
2. Run `./gradlew buildAllProcedureJars --parallel` to build the test JARs
3. Create a new plan and add the `DumbRecurrenceGoal` to its scheduling spec. Be sure to give it an argument for the `biteSize` parameter
4. Schedule the plan twice. Expected result: the second run should fail with an `IllegalStateException`
5. Checkout this branch, rebuild and redeploy (I recommend doing this without taking down the volumes, but if you do take them down, then repeat steps 2 and 3)
6. Schedule the plan as many times as you wish. There will be no more exception and each run will add more `BiteBanana` activities

### DIFF
--- a/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/BasicSchedulingTests.java
+++ b/e2e-tests/src/test/java/gov/nasa/jpl/aerie/e2e/procedural/scheduling/BasicSchedulingTests.java
@@ -10,6 +10,7 @@ import org.junit.jupiter.api.Test;
 
 import javax.json.Json;
 import java.io.IOException;
+import java.util.Comparator;
 import java.util.Objects;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -34,7 +35,8 @@ public class BasicSchedulingTests extends ProceduralTestingSetup {
         "Test Scheduling Procedure 1",
         dumbRecurrenceGoalJarId,
         specId,
-        0
+        0,
+        true
     );
   }
 
@@ -184,7 +186,7 @@ public class BasicSchedulingTests extends ProceduralTestingSetup {
    * Run a spec that includes unfinished activities
    */
   @Test
-  void IncludesUnfinishedActivities() throws IOException {
+  void includesUnfinishedActivities() throws IOException {
     // Disable the recurrence goal to simplify scheduling results
     hasura.updateSchedulingSpecEnabled(dumbRecurrenceGoalId.invocationId(), false);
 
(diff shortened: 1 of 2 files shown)


dandelany (approved):
👍 Reviewed this with @Mythicaeda and @mattdailis and all the code changes make sense - tested locally today and confirmed this works as expected. Thanks for fixing this! Merging

PR #1847, opened 2026-07-06
Drop references to Slabs that have already been processed

mattdailis:
Paired with @DavidLegg to re-implement the changes in https://github.com/NASA-AMMOS/plandev/tree/perf/timeline-trim in a way that made sense to us.

Key changes:
- Maintain SpanInfo from start to finish of simulation
- Remove `head` field from SlabList, asserting that any iterators are created before any items are inserted
- Add method to step up all cells, moving the iterators up and (hopefully) dropping the last references to Slabs, allowing the GC to collect them


DavidLegg:
This replaces https://github.com/NASA-AMMOS/plandev/pull/1845 - After Matt and I talked through the existing Aerie code and Claude's work, we decided that starting from scratch and using Matt's prior work and Claude's work as references would give us the cleanest results the fastest.

This still needs to be profiled to make sure it actually saves on retained memory. We don't expect speed improvements.

PR #1848, opened 2026-07-06
FPrime dictionary parsing - support for aliases, don't change stem names

joswig:
This PR updates the parsing of FPrime command dictionaries. There is an inprogress follow up [pr](https://github.com/NASA-AMMOS/aerie-sequence-languages/pull/26) adding support for the fprime commanding language. Dictionary language and sequencing language are not coupled in our implementation

- [x] Updated dictionary schema pulled from https://github.com/nasa/fpp/blob/main/compiler/tools/fpp-to-dict/test/dictionary.schema.json
- [x] Regenerated dictionary parsing code which now support alias element added to schema
- [x] Removed command stem munging, as best as anyone can recall that was added to allow FPrime sequencing to use seqn.grammar, but that requirement is obsolete and by not changing stems we can utilize format that is native to fprime tool chain
- [x] Added test dictionary

Testing

1. Startup plandev with the following environment variable under aerie-sequencing DICTIONARY_PARSER_PLUGIN: "/fprime-dictionary-parser/fprime-parser.js"
2. In plandev-ui
  a. Upload one of the test fprime json dictionaries
  b. create a parcel with this dictionary
  c. create a workspace with the parcel with this dictionary
  d. open the workspace
  e. create a sequence
  f. go to the dictionary view panel and ensure expected commands are and arguments are present


sequencing-server/plugins/fprime-dictionary-parser/schema/fprime-schema.json
```diff
@@ -1,488 +1,590 @@
...
+                        }
+                    },
+                    "id": {
+                        "type": "number"
+                    },
+                    "format": {
+                        "type": "string"
+                    },
+                    "throttle": {
+                        "type": "object",
```
Mythicaeda, line 410: Will `throttle`'s type changing in the schema break existing sequences?

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
```diff
@@ -1,15 +1,15 @@
...
   version: '1.0.0',
-  author: 'Ryan Goetz',
+  author: 'AMMOS SeqDev',
   parseDictionary(dictionaryString: string): {
-    commandDictionary?: CommandDictionary,
-    channelDictionary?: ChannelDictionary,
-    parameterDictionary?: ParameterDictionary
+    commandDictionary?: CommandDictionary;
+    channelDictionary?: ChannelDictionary;
+    parameterDictionary?: ParameterDictionary;
```
Mythicaeda, line 12: What is the semantic difference behind changing the `,`s to `;`s?

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
```diff
@@ -93,7 +93,7 @@ function commandToAMPCSCommand(
   return {
     ...argsToAMPCSArgs(fppCommand.formalParams, typeDefinitions),
     description: fppCommand.annotation ?? '',
-    stem: fppCommand.name.replaceAll('.', '_'),
+    stem: fppCommand.name,
```
Mythicaeda, line 96: Why were we initially changing `.`'s to `_`s? Was `.` a restricted character?

sequencing-server/plugins/fprime-dictionary-parser/test/dictionary/MathDeploymentTopologyDictionary.json
Mythicaeda, line 1: What's the original source of this test dictionary? I'm not seeing it in the `fpp` repo or the `fprime-examples` repo.

sequencing-server/plugins/fprime-dictionary-parser/test/test.ts
Mythicaeda, line 1: Given that this PR adds `alias` support, could you add a test that checks that sequencing can actually drill down to the underlying type from the alias? (ie, `FwSizeType`, which is a `PlatformSizeType`, which is `64-bit integer`)

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
Mythicaeda: Additionally, I notice that argument names have the same call to replace `.`s with `_`s. Do those instances need to be updated as well?

sequencing-server/plugins/fprime-dictionary-parser/schema/fprime-schema.json
joswig: In theory yes, but since we haven't merged the language plugin for fprime and no missions have communicated using it yet, but it is a requested new capability, the risk is very low.

We also don't own the schema, this is merely a copy paste of the canonical version the fprime team maintains at https://github.com/nasa/fpp/blob/main/compiler/tools/fpp-to-dict/test/dictionary.schema.json

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
joswig: I checked with @goetzrrGit on the initial implementation here, the best he could recall was to allow fprime dictionaries to be used with the SeqN grammar which prohibits periods in stem names.

As we now have a language plugin mechanism, my preference is to use the existing externally defined fprime sequencing language there's a WIP PR [fprime highlighting and tooltips](https://github.com/NASA-AMMOS/aerie-sequence-languages/pull/26) which will implement that

joswig: regarding the other removing the other replaces, I think that's good from a consistency standpoint. If you're onboard I'll take those out as well

sequencing-server/plugins/fprime-dictionary-parser/test/dictionary/MathDeploymentTopologyDictionary.json
joswig: I generated it by following the [F Prime MathComponent Tutorial](https://github.com/fprime-community/fprime-tutorial-math-component)

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
joswig: I'll remove the name munging from the arguments and any others

sequencing-server/plugins/fprime-dictionary-parser/src/fprime-parser.ts
joswig: "You can use , or ; to separate the properties" per https://www.typescriptlang.org/docs/handbook/2/everyday-types.html.

We more commonly use ; in our types, though this one was courtesy of our an auto correct

Mythicaeda (approved):

PR #1852, opened 2026-07-14
Add upload and download simulation dataset feature

remy-rabideau:
* **Review:** By commit

___REQUIRES_GATEWAY_PR___="148"
[NASA-AMMOS/plandev-gateway#148](https://github.com/NASA-AMMOS/plandev-gateway/pull/148)

## Description

Adds a matched pair of Hasura actions for moving simulation datasets in and out of PlanDev:

- **`uploadSimulationDataset`** (mutation) — accepts a simulation results JSON payload in `SimulationResultsWriter` format and persists it to the database against an existing plan, returning the new `simulationDatasetId`.
- **`downloadSimulationDataset`** (query) — reads an existing simulation dataset back out and returns it in that same JSON format.

The two are deliberately symmetric: the format `downloadSimulationDataset` emits is the format `uploadSimulationDataset` accepts, so a dataset can be downloaded from one plan and re-uploaded to another with round-trip fidelity. Two serializer tests (`testSerializeEmptyResultsRoundTrips`, `testSerializeFullResultsRoundTrips`) and one e2e test (`validRoundTrip`) pin that property.

The primary use case for upload is transferring a simulation dataset generated through procedural stateless scheduling into the PlanDev database for viewing in the UI; ingesting results from external simulators is also supported. Download covers the reverse direction — exporting results for offline analysis, archival, or transfer between environments. The `SimulationResultsWriter` format is also produced by aerie-cli via `plans download-simulation-full-results`.

Pipeline (upload): Hasura receives the mutation → forwards the action envelope to merlin-server → `SimulationResultsParser` parses the JSON → `PostgresPlanRepository.uploadSimulationDataset` validates activity types, then writes the dataset, profiles, segments, activity spans, topics, and events in a single transaction → returns the new `simulationDatasetId`.

Pipeline (download): Hasura receives the query → merlin-server loads the dataset record, profiles, spans, topics, events, and simulation arguments → `ResponseSerializers.serializeSimulationResultsForDownload` emits the `SimulationResultsWriter` JSON shape.

### Reviewer notes

Three things worth attention:

1. **Permissions differ between the two actions, intentionally.** `uploadSimulationDataset` is gated on `simulate`, because uploading produces a simulation dataset — the same artifact a simulation run produces — rather than an external dataset. `downloadSimulationDataset` is gated on `resource_samples`, the permission that already governs reading simulated resource data; the `viewer` role holds it as `NO_CHECK`, so viewers can export results from plans they do not own. The e2e tests `forbidden` and `viewerMayDownload` cover both sides of this.
2. **The Javalin max request size was raised to 256 MB globally**, not scoped to the upload route.
3. **`SimulationResults` gained a `simulationArguments` field.** The existing 8-argument constructor is retained and delegates with `Map.of()`, so no existing call site changes.

Changes:

- **deployment/hasura/metadata/actions.graphql** — adds the `uploadSimulationDataset` mutation and `downloadSimulationDataset` query, their `UploadSimulationDatasetResponse` / `DownloadSimulationDatasetResponse` types, and the `SimulationResultsInput` / `SimulationResultsOutput` scalars
- **deployment/hasura/metadata/actions.yaml** — registers both action handlers against merlin-server with a 300 s timeout; upload permits `aerie_admin` and `user`, download additionally permits `viewer`
- **merlin-server/.../SimulationResultsParser.java** *(new)* — parses the `SimulationResultsWriter` JSON format: DOY timestamps, real/discrete profiles with named segments, simulated and unfinished activities, simulation arguments, topics, and events. Parse-only — the unparse direction throws `UnsupportedOperationException` pointing at `ResponseSerializers.serializeSimulationResultsForDownload`, which is the real serializer
- **merlin-server/.../ResponseSerializers.java** — adds `serializeSimulationResultsForDownload` (profiles, spans, simulation arguments, topics, and events, with event graphs flattened via `EventGraphFlattener`) and `serializeCreatedSimulationDatasetId`
- **merlin-server/.../MerlinBindings.java** — registers `POST /uploadSimulationDataset` and `POST /downloadSimulationDataset`; parses the Hasura action envelope, applies the permission checks described above, and maps failures to status codes
- **merlin-server/.../HasuraParsers.java** — adds `hasuraUploadSimulationDatasetActionP` and `hasuraDownloadSimulationDatasetActionP`
- **merlin-server/.../HasuraAction.java** — adds the `UploadSimulationDatasetInput` and `DownloadSimulationDatasetInput` records
- **merlin-server/.../InvalidSimulationDatasetException.java** *(new)* — carries the set of activity types present in an uploaded dataset but absent from the plan's mission model
- **merlin-server/.../MerlinFormattedError.java** — maps that exception to an `INVALID_SIMULATION_DATASET` formatted error
- **merlin-server/.../PlanService.java / LocalPlanService.java** — adds `uploadSimulationDataset` and `downloadSimulationDataset` to the service interface and implementation
- **merlin-server/.../PlanRepository.java / PostgresPlanRepository.java / InMemoryPlanRepository.java** — adds both methods to the repository interface. The PostgreSQL implementation writes the full dataset in a single transaction on upload (validating activity types against the mission model first) and reassembles a `SimulationResults` on download; the in-memory mock throws `UnsupportedOperationException` for both
- **merlin-server/.../PostgresResultsCellRepository.java** — widens `getSimulationTopics`, `getSimulationEvents`, `getActivities`, `insertSimulationTopics`, `insertSimulationEvents`, and `postActivities` from `private` to package-private so `PostgresPlanRepository` can reuse them instead of duplicating the queries
- **merlin-driver/.../SimulationResults.java** — adds the `simulationArguments` field and a 9-argument constructor; the existing 8-argument constructor delegates with an empty map
- **merlin-server/.../AerieAppDriver.java** — raises the Javalin max request size to 64 GB per request by Clipper

## Verification

Automated coverage added:

- **`UploadSimulationDatasetParserTest`** (9 tests) — real profiles, activities, simulation arguments, topics and events, invalid timestamps, missing fields, and absence defaults
- **`DownloadSimulationDatasetParserTest`** (7 tests) — valid envelopes, missing `planId` / `simulationDatasetId`, non-numeric and beyond-int-range dataset ids, and missing session variables
- **`DownloadSimulationDatasetSerializerTest`** (10 tests) — real and discrete profiles, simulated and unfinished activities, simulation arguments, topics and events, concurrent events within one transaction, plus the two round-trip tests
- **`UploadSimulationDatasetTests`** (e2e, 8 tests) — successful upload with empty, discrete, real, and mixed profiles; invalid plan id (404); missing required field (400); unauthorized user (403); unknown activity types rejected (400)
- **`DownloadSimulationDatasetTests`** (e2e, 7 tests) — invalid plan id, missing and nonexistent dataset id, forbidden role, empty profiles, viewer access, and a full upload → download round trip

## Documentation

No existing documentation covers these endpoints. The `SimulationResultsWriter` format they read and write is defined in the `aerie-cli` repository (see [NASA-AMMOS/aerie-cli#187](https://github.com/NASA-AMMOS/aerie-cli/pull/187)).

## Future work

- Return 404 rather than 500 for a nonexistent `simulationDatasetId` on download — the lookup currently throws a bare `RuntimeException` that falls through to the catch-all handler (asserted as-is in `nonexistentSimulationDatasetId`)
- Read simulation arguments through a typed repository action instead of the inline `select arguments from merlin.simulation_dataset` statement in `PostgresPlanRepository.downloadSimulationDataset`
- Stream download responses rather than buffering the full dataset in memory
- plandev-gateway and plandev-ui changes (separate PRs) are required to expose these endpoints to end users


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
```diff
@@ -114,6 +114,7 @@ public static void main(final String[] args) {
     final var javalin = Javalin.create(config -> {
       config.showJavalinBanner = false;
       if (configuration.enableJavalinDevLogging()) config.plugins.enableDevLogging();
+      config.http.maxRequestSize = 64L * 1024 * 1024 * 1024;
```
remy-rabideau, line 117: Per request by Clipper for large simulation result datasets

PR #1855, opened 2026-07-27
Upgrade (some) backend Node services to Node 24

dandelany:
Some of our backend Node services/libraries are using older versions of Node - we need to bring them up to Node 24 for long term support + future security patches. Right now we're mainly targeting Constraints and Scheduling services, more to come.


duranb (approved):
Should we also add an .npmrc file to enforce the correct node versions where appropriate?

```
# .npmrc

engine-strict=true
```

I'll go ahead and approve for either case. Thanks!

dandelany:
> Should we also add an .npmrc file to enforce the correct node versions where appropriate?

I don't think so - this entirely blocks running/installing if the Node version is wrong and I'm not super confident we (or users) don't/won't want to do that in some cases. It *is* reasonable to want to enforce that for devs, and it seems like package.json has a new [**devEngines**](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) field for that which I will look into for next round...

Gonna merge this for now - will handle security scan fixes (unrelated) in future PR

PR #1860, opened 2026-08-05
Update Action Server to Node 24, update Dockerfile

dandelany:
Follow-on from #1855, this updates Action Server from node 22 -> 24. I also:
* added a couple missing dependencies to the package.json (they were only working due to peer dependencies)
* upgraded a couple dependencies - eslint-sheriff and piscina - to fix warnings & take bug fixes
* Updated the way we call eslint-sheriff config to match latest docs
* Updated the Dockerfile and .dockerignore to a more modern multi-stage Dockerfile format, matching those in #1855 more closely and resulting in a smaller, more secure image


duranb (approved):
lgtm!

PR #1861, opened 2026-08-05
Preserve activity names

mattdailis:
Procedural scheduling tracks activity names, but those names get lost when they pass through `gov.nasa.jpl.aerie.types.ActivityDirective`.

This PR adds a name field to `gov.nasa.jpl.aerie.types.ActivityDirective`.

### DIFF
--- a/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
+++ b/merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
@@ -19,7 +19,8 @@ import static gov.nasa.jpl.aerie.merlin.server.remotes.postgres.PostgresParsers.
       ceil(extract(epoch from a.start_offset) * 1000*1000) as start_offset_in_micros,
       a.arguments,
       a.anchor_id,
-      a.anchored_to_start
+      a.anchored_to_start,
+      a.name
     from merlin.activity_directive as a
     where a.plan_id = ?
     """;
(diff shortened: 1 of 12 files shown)


merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
```diff
@@ -44,7 +45,8 @@ public List<ActivityDirectiveRecord> get(final long planId) throws SQLException
                 getJsonColumn(results, "arguments", activityArgumentsP)
                     .getSuccessOrThrow($ -> new Error("Corrupt activity arguments cannot be parsed: " + $.reason())),
                 (Integer)results.getObject("anchor_id"),
-                results.getBoolean("anchored_to_start"))
+                results.getBoolean("anchored_to_start"),
+                (String)results.getObject("name"))
```
DavidLegg: Why is this `getObject` cast to String, and not `getString`?

DavidLegg (approved):

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/remotes/postgres/GetActivityDirectivesAction.java
mattdailis: The intent was to handle `null` correctly. After checking just now, `getString` will do what I want, so you're right  :+1: I pushed a commit to change it.

Getters for primitive types (long, integer) are where you want to use `getObject`—otherwise you get `0` instead of `null`. String isn't primitive, so we're fine.

That's what I get for not checking the docs! Thanks for calling it out

PR #1868, opened 2026-08-10
[WIP] Rename PlanDev packages

Mythicaeda:
* **Review:** By commit  


dandelany (commented):
Discussed this briefly with @Mythicaeda and we had some concerns about how files show up as deletions/renames instead of moves - I investigated today and propose we use #1870 as an equivalent but slightly-improved version of this PR. @Mythicaeda please give that one a read when you have a chance & let me know what you think

PR #1869, opened 2026-08-10
Stop Exposing Hasura Action Endpoints

Mythicaeda:
* **Review:** By commit  

## Description

Before, the Merlin and Scheduling server containers exposed the main Javalin port by default for the sake of health checks. Similarly, the Sequencing server exposed the main Express port for the same reason. This theoretically could allow a user to directly hit the Hasura Action endpoints in any of these servers. As all of these servers expected to receive communication over these endpoints solely from Hasura, they were not setup to validate the JWT or user headers they received.

For Merlin and Scheduler, this behavior has been fixed by adding a second Javalin instance that solely contains a Health Check endpoint on a separate port, internal port 8080. This Javalin instance is set to be brought up and shut down by the "main" instance, so it should still accurately reflect the state of the container. For Sequencing, this behavior has been fixed by adding a second Express server that solely contains Health Check endpoint on internal port 8080. The `dev` and `deployment` docker composes have been updated to use `8080` instead of the old port.

Because of this change, the tests had to be split into two suites: tests that require accessing the "internal" Javalin instance in Merlin Server and Scheduler Server, and the rest of the tests. The rest of this PR is this refactor, alongside a few other changes to the tests:
 - Test procedures are now organized into `constraints` and `goals`, rather than having them intermixed. They've also been up a couple of packages (from `procedural.scheduling.procedures` to just `procedural`)
 - External Event tests have been reworked and can now be safely run in parallel. The former implementation reused the same primary key values for tests in multiple files. This should fix the "flakiness" we were seeing with the External Event tests
 - Other duplicated primary key values have been deduplicated (Simulation Tests, Workspace Metadata Route Tests)
 - Tags have been added to all test classes so that we can easily create more suites in the future.
 - E2E Test parallelization is now configured in the JUnit engine itself, rather than at the Gradle task level. DB Tests have also been updated to run in parallel.

## Verification

The main changes were validated by trying to directly `GET` the Hasura action endpoints while using the new docker configuration. For all endpoints, the server returned that it was unable to find such a route. In comparison, trying to directly `GET` the Hasura action endpoints while using the old docker configuration would result in being told that the endpoints _do_ exist but don't support `GET`.

For Sequencing Server, I temporarily added the following `GET` handler to the main Sequencing App, then performed the same tests. As expected, this `GET` response was only accessible when using the old configuration.
```js
app.get('/put-dictionary', (_: Request, res: Response) => {
  res.send('Got put-dictionary');
});
```

## Documentation

`DEVELOPER.MD` and the e2e-tests' `README.MD` have been updated to:
1. Clarify that in order to locally run _all_ of the e2e-tests, `docker-compose-test.yml` must be used to bring up PlanDev
2. Add the mandatory `buildAllProcedureJars` step to the e2eTest instructions. Without this step, tests that use procedural goals or constraints will not have the JARs they need.

## Future work

It'd be great if, rather than having two Javalin instances, we can have one serving two different "webapps" on different ports: the Hasura one and the healthcheck. Jetty servers (the server type that Javalin uses) seem to support this concept with [virtual hosts](https://jetty.org/docs/jetty/12.1/operations-guide/deploy/index.html#virtual-hosts-port), but I'm not sure how to configure this in the Javalin context.


.github/actions/shutdown-test-backend/action.yml
```diff
@@ -1,19 +1,28 @@
...
+    required: false
+    default: 'docker-compose.yml'
 runs:
   using: composite
   steps:
     - name: Print Logs for Services
       if: always()
       shell: bash
-      run: docker compose -f ./e2e-tests/docker-compose-test.yml logs -t
+      run: docker compose -f $INPUT_DOCKER_FILE logs -t
```
dandelany: shell vars should always be double-quoted in situations like this ie. `docker compose -f "$INPUT_DOCKER_FILE"`. Without quotes, if this value ever contains a space, it will be treated like separate command line arguments

dandelany (changes requested):
Reviewed with @Mythicaeda & this is very close to ready, a few notes:
* Please update from `develop` & resolve the conflict
* From a dev perspective, without the context of this PR, its not clear what the intention of the new "Routes tests" e2e tests vs the others are - can you add a comment explaining eg. "routes tests are integration tests which run *within* the internal docker network to test internal routes, other e2e tests run only against ports which are exposed to the outside world" or similar?
* One remaining concern was being confident that, if the main *app* server goes down, the *healthcheck* will also go down. It seems like we are doing everything we can on the Java side to ensure this, but would be good to test the JS server (sequencing) & ensure that if an uncaught error brings down the app server, the healthcheck isn't still reporting "healthy"

merlin-server/src/main/java/gov/nasa/jpl/aerie/merlin/server/AerieAppDriver.java
```diff
@@ -101,26 +101,57 @@ public static void main(final String[] args) {
...
+      config.showJavalinBanner = false;
+    });
+
+    // Set up the health check routes
+    healthCheckServer.get("", ctx -> ctx.status(200));
+    healthCheckServer.get("health", ctx -> ctx.status(200));
+
+    // Tie the health checker into the merlin server health
+    merlinServer.events(listener -> {
+      listener.serverStarting(() -> healthCheckServer.start(8080));
```
dandelany: I think this should be `serverStarted` rather than `serverStarting`, otherwise the healthcheck server may still run if merlin server tries-but-fails to fully start up.

dandelany (approved):
👍 ready to merge when tests pass

PR #1870, opened 2026-08-14
Rename java packages in two commits

dandelany:
This branch was created because I (+ @mattdailis & @Mythicaeda ) had concerns about #1868 - in the Github UI, a large number of changes which are essentially "rename file X and change its imports to the new namespace" appear instead as "file X was deleted and file Y was created with new contents". We were concerned that this would have an effect on git history and that `git blame` would no longer show useful results on these supposedly "new" files.

I spent some time investigating this today and found:
* The issue is **not as bad** as I first thought - the Github UI does not use the same heuristic as `git blame`
* However, some files (~72) **are affected negatively** and their `git blame` output shows the rename as the "blame commit" for some/all of the file
* I put together this branch as a test to see what happens if we insert a "file rename only" commit before the commit which changes file contents, and:
  * it does improve `git blame` for those 72 files
  * and it is provably identical to #1876

So overall it's a smaller improvement than I initially thought, but probably still worth keeping this branch instead of #1876, if you agree @Mythicaeda. You can confirm it reaches the exact same end result by checking out this branch and running `git diff origin/feat/rename-plandev-packages HEAD`. Details:

## Github UI vs. Git Blame

I assumed that since eg. [ActivityInstance.java](https://github.com/NASA-AMMOS/plandev/pull/1868/changes#diff-6c5c1630caf5563ebcc618ec857e498dcf4812d678038611b67b585509a47db3) appears as a "new file" in the GH PR interface on #1868, this also meant its `git blame` history would be changed/set to the rename commit, but this is not always the case: if you run `git blame constraints/src/main/java/gov/nasa/ammos/plandev/constraints/model/ActivityInstance.java` on `feat/rename-plandev-packages` (@Mythicaeda 's branch), it **does** show proper git history from before the rename.

However, this is **not true** of all files, eg. I checked another and`git blame examples/model-migration-2/src/main/java/gov/nasa/ammos/plandev/examples/model/migration/activities/GrowBananaActivity.java` shows the rename commit as the entire `git blame` history of the file. So there is at least some history "lost" (the commits aren't lost, but `blame` is).

## This Branch

The idea here was to make a new commit before Theresa's commit which *only* moved files to their new locations, and then apply @Mythicaeda 's changes as a 2nd commit - to help Git's heuristics understand the history. Here's what I did:
```
# make a local backup copy of @Mythicaeda 's rename branch to work from
git branch backup/rename-plandev-packages origin/feat/rename-plandev-packages
# make a new "clean" branch starting from the same base commit where the other branch forked off from `develop`
base=$(git merge-base origin/develop backup/rename-plandev-packages)
git switch -c feat/rename-plandev-packages-clean "$base"

# run a script to move all the java files to their new locations & commit the moves only
node move-java-packages.mjs --apply
git commit -m "moved files"

# restore the changes from the other branch into the working tree & commit them after the file moves
git restore --source backup/rename-plandev-packages --staged --worktree -- .
git commit -m "rename java packages"

# diff against the original branch to confirm exactly the same - returns nothing
git diff origin/feat/rename-plandev-packages HEAD   
```

## Is it any better?

As you can see, the Github PR diff looks exactly the same 😅 *But:*

* The PR review is easier since we can look at the two commits separately and they are simple on their own
* and `git blame` **is improved** for the relatively small # of files that were affected. Eg. `git blame examples/model-migration-2/src/main/java/gov/nasa/ammos/plandev/examples/model/migration/activities/GrowBananaActivity.java` shows correct history on this branch

Overall a quick hacky script I wrote to compare shows "**750 git blame lines recovered, 72 files improved**, 1044 files unchanged, 0 worse". If I had realized the impact would be that small at the beginning I might not have gone to the trouble 😅 But now that it's done, it's probably worth keeping!

### Appendix: move script

I didn't commit the move script since it was just a throwaway, but here it is if you care:
<details>
<summary>move-java-packages.mjs</summary>
```javascript
#!/usr/bin/env node

import { execFileSync } from 'node:child_process';
import { mkdirSync } from 'node:fs';
import { dirname } from 'node:path';

const apply = process.argv.includes('--apply');
const from = 'gov/nasa/jpl/aerie';
const to = 'gov/nasa/ammos/plandev';

const files = execFileSync('git', ['ls-files', '-z'], {
  encoding: 'utf8',
  maxBuffer: 10 * 1024 * 1024,
})
  .split('\0')
  .filter(Boolean);

const moves = files
  .filter(file => file.includes(from))
  .map(oldPath => ({
    oldPath,
    newPath: oldPath.replace(from, to),
  }));

for (const { oldPath, newPath } of moves) {
  console.log(`${oldPath} -> ${newPath}`);

  if (apply) {
    mkdirSync(dirname(newPath), { recursive: true });
    execFileSync('git', ['mv', '--', oldPath, newPath]);
  }
}

console.log(
  `\n${moves.length} files ${apply ? 'moved' : 'would be moved'}`,
);
```
</details>


Mythicaeda (approved):
Approval contingent on Tests passing

e2e-tests/src/test/java/gov/nasa/ammos/plandev/e2e/utils/GatewayRequests.java
```diff
@@ -26,15 +26,15 @@ public GatewayRequests(Playwright playwright) throws IOException {
...
+   * Auto-login the PlanDevE2ETests user, whose token is used to upload mission model JARs
    */
   private void login() throws IOException {
     if (token != null) return;
     final var response = request.post(
         "/auth/login", RequestOptions.create()
                                      .setHeader("Content-Type", "application/json")
                                      .setData(Json.createObjectBuilder()
-                                                  .add("username", "AerieE2eTests")
+                                                  .add("username", "PlanDevE2ETests")
```
Mythicaeda, line 37: This can be done in the other PR, but on recheck, the Sequencing test helper `login.js` has a test user named "AerieE2ESequencingTests" that ought to be renamed

Mythicaeda (approved):

e2e-tests/src/test/java/gov/nasa/ammos/plandev/e2e/utils/GatewayRequests.java
dandelany: 👍 thanks, i'll make sure this gets covered in the other PR

PR #1872, opened 2026-08-17
Fix issue with getting correct tag IDs in `begin_merge` procedure

dandelany:
## Summary

@Mythicaeda and I have been working on https://github.com/NASA-AMMOS/plandev/pull/1870 which is supposed to *only* rename the Java packages (& related backend things), with no functional changes. However, the backend DB tests started failing on that branch for some reason - specifically two tests in `PlanCollaborationTests:TagsTest`. When they are run on their own they still pass, but they fail in the test suite. 

I eventually tracked it down to this issue in the `begin_merge` procedure which has been there for at least two years, maybe more - we didn't notice it because the test incorrectly passed due to a coincidence of snapshot vs. activity directive IDs.

### Core Issue
In the `begin_merge` function, line 293 (the changed line) is meant to be getting the tag IDs associated with a **snapshot ID**, but it called `tag_ids_activity_directive(dd.activity_id, psa.snapshot_id)`. The second param of `tag_ids_activity_directive` is supposed to be a plan ID, but a `snapshot_id` is being passed. I changed this to call `tag_ids_activity_snapshot` which I think was the intention all along

I also added an up/down migration since this modifies a core DB procedure.

### Why did tests miss it?
This bug was hard to detect because the incorrect tag lookup usually produced behavior that was indistinguishable from correct behavior. `begin_merge` passed a **snapshot ID** to `tags.tag_ids_activity_directive, which interpreted it as a **plan ID,** instead of calling tags.tag_ids_activity_snapshot. Most activities have no tags, and most conflicting modifications also differ in fields other than tags, so the incorrect tag value generally had no effect on the merge result. The bug only becomes significant when both plans modify the same activity and tags are necessary to distinguish the two versions.

### Why did the rename cause tests to start failing?
The package rename in #1870 changed the fully qualified names of the nested `PlanCollaborationTests` classes, which **changed JUnit’s ["deterministic but intentionally non-obvious"](https://docs.junit.org/6.0.1/writing-tests/test-execution-order.html) test order**. Because database cleanup does not reset IDs, this also changed the plan and snapshot IDs assigned during the tests. Under the new ordering, a snapshot ID incorrectly used as a plan ID collided with the receiving plan’s ID, causing the merge logic to compare the receiving tags to themselves and incorrectly conclude that the plans were identical.

### DIFF
--- a/deployment/hasura/migrations/Aerie/37_fix_merge_tags/down.sql
+++ b/deployment/hasura/migrations/Aerie/37_fix_merge_tags/down.sql
@@ -1,7 +1,7 @@
 create or replace procedure merlin.begin_merge(_merge_request_id integer, review_username text)
   language plpgsql as $$
-declare
-validate_id integer;
+  declare
+    validate_id integer;
     validate_status merlin.merge_request_status;
     validate_non_no_op_status merlin.activity_change_type;
     snapshot_id_supplying integer;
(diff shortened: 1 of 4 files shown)


Mythicaeda (approved):
Still don't know what caused the tests to finally start failing, but I'm glad this bug has finally been caught and fixed!

PR #1879, opened 2026-09-10
DB migration script: wrap `shlex.quote` around command line args to ensure they are properly quoted

dandelany:
## Summary

Fixes shell escaping in `plandev_db_migration.py` by using Python's (built-in standard-library) `shlex.quote()` for values passed to Hasura CLI commands.

This prevents migration failures when paths or configuration values contain spaces, quotes or other shell-significant characters. We encountered this during upgrade testing when the deployment directory was named `deployment 2`, which caused the Hasura command to be parsed incorrectly and eventually resulted in an opaque `IndexError`.

This change applies quoting consistently to:

* Hasura project and `.env` paths
* Database source name
* Hasura endpoint
* Hasura admin secret

### DIFF
--- a/deployment/plandev_db_migration.py
+++ b/deployment/plandev_db_migration.py
@@ -4,6 +4,7 @@
 import os
 import argparse
 import sys
+import shlex
 import shutil
 import subprocess
 from dotenv import load_dotenv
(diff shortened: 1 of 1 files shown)


mattdailis (approved):
Seems reasonable to me

PR #1886, opened 2026-09-16
DB changes for non-executable models/read only plans

AaronPlave:
WIP starting point for non-executable plan db changes. Mission model table changed + migrations, no other real fallout addressed.

Main issue: https://github.com/NASA-AMMOS/plandev/issues/1885


dandelany:
@Mythicaeda this was the DB outcome of our discussion this afternoon - we want to get agreement on DB structure as first priority so that we can work independently knowing what the data will look like. This is ready for a review from you and, once we get agreement/tests passing, I think it can be merged - since it won't break any existing behavior on the backend for existing models/plans.

After discussion, we prefer keeping an `is_read_only` flag on the `plan` in addition to the `is_executable` flag on the model as implemented here - while the former would be derivable from the latter, it feels cleaner and more future proof to have them as separate flags/concepts (but open to being convinced otherwise)

deployment/postgres-init-db/sql/tables/merlin/mission_model.sql
```diff
@@ -9,7 +9,9 @@ create table merlin.mission_model (
   default_view_id integer default null,
 
   owner text,
-  jar_id integer not null,
+  jar_id integer,
```
Mythicaeda: Are we hard-locked into the decision of nullable `jar_id`? I'd prefer renaming this column to something like `definition_file_id` and saving the plan.json++ file as the model's definition, for two big reasons and one minor reason:

Big reasons:
1. Traceability. If there's a question as to why the stub model generated a certain way, we have the original file for reference.
2. We discussed in the future wanting to support `model.json` files that could have executable code in them. Renaming this column and saving the definition file feels closer to that eventual extension.

Minor reason:
1. There's no risk of introducing NPEs when dealing with Mission Models on either the backend or frontend. Additionally, in the event that the backend incorrectly attempts to load a non-executable mission model, the backend's `failed to load mission model from "model.json"` is a far clearer error message than an NPE

deployment/postgres-init-db/sql/tables/merlin/plan.sql
```diff
@@ -12,6 +12,7 @@ create table merlin.plan (
     on update cascade,
 
   is_locked boolean not null default false,
+  is_read_only boolean not null default false,
```
Mythicaeda, line 15: For this feature, this flag is planned to always be equal to the value of `mission_model.is_executable`. Unless we are planning to introduce "archived plans" (plans that are readonly but are using an executable model), then I propose simplifying this by using the model's `executable` column.

deployment/postgres-init-db/sql/tables/merlin/mission_model.sql
Mythicaeda: Having the file saved _does_ make me wonder if we could kick the parsing off to the backend instead of the gateway. Hasura automatically attempts to call `refreshActivityTypes`, `refreshModelParameters`, and `refreshResourceTypes` when a mission model is added, so those methods will already need to be updated in some way to handle these stub models (which could just be to ignore them). 

Pros:
- having the backend do the parsing uses our existing pipeline for how Activity and Resource Types get into the DB, rather than having different services handle different model types

Cons:
- if the backend's doing the parsing, it will need another Event to parse out the sim results (if present) after the `refresh` Events have run (or a watcher thread waiting for the `refresh`es to complete). This is obviously more complicated than having the Gateway parse the entire `plan.json++` at once

deployment/postgres-init-db/sql/tables/merlin/plan.sql
dandelany: Discussed & I think we want to keep this readonly flag on the plan too

deployment/postgres-init-db/sql/tables/merlin/mission_model.sql
dandelany: Discussed & decided to:
* Change `jar_id` to `definition_file_id`
* Still parse the JSON file in Gateway
* When doing so, separate the "model" portion from the rest of the plan and save this in a JSON file
* Set that JSON file ID to the `definition_file_id` in the DB

dandelany (approved):
Updated during review, this is now ready 👍

Mythicaeda:
Pushed a couple commits to update the other locations in the backend that referred to `jar_id` to fix the tests. The UI will also need to do this rename on its "upload mission model JAR" method, but the Gateway should be clear, since it doesn't currently create a Mission Model.

PR #1887, opened 2026-09-17
Non-Executable Plans

### LINKED ISSUE #1885: Support Model-Free Non-Executable Plan Import, Viewing, and Template Expansion
Allow users to import a self-contained JSON file into PlanDev, inspect its activities and datasets, and perform template expansion without a mission model.

Target: End of September 2026. Both @AaronPlave and @joshhaug have worked on prototypes of this functionality - the requirements and remaining decisions below should guide the implementation.

## Requirements

* Support a JSON format capable of representing everything currently supported by `plan.json`, plus spans (simulated activities) and imported resource datasets. This must also include supporting fields like activity definitions and resource types.
* Create a plan from this file without a mission model or other supporting files like external datasets. Users should *not* be allowed to select a mission model to associate with these model-free plans
* When importing, validate this file for format & consistency, ideally against a JSON schema that can provide specific contextual error messages if invalid
* We refer to these as "model-free" plans because they do not need a user-provided Java mission model - but they may use a "stub" model row in the database that is hidden from the user
* Model-free plans should be importable by non-admin users, ie. the same group of users that are allowed to create a normal plan
* Model free plans should support activity directives, spans, and resources - though *none* of these are required
* Ensure the UI can display and inspect activities and resources in model-free plans
* Display the imported plan in read-only mode: activities cannot be modified, and simulation and scheduling cannot run.
* Ensure that template expansion still works for these model-free plans
* Ensure that procedural constraints can still be run on model-free plans

## Questions & Decisions

* **Activity definitions:** Must the file include an activity type dictionary with parameter schemas, or are definitions optional? Which metadata is necessary for viewing and template expansion? Where should these definitions be stored?
  * **Decision:** For now, we will require the user to provide them in the JSON file (proposed schema forthcoming).
  * This is the easiest way to proceed now since the UI expects them. _If we have time_, at the end, make the UI tolerant of missing activity/parameter definitions and update the schema to make them optional
* **Model association:** Should these plans have no model association, or use a stub model row? How will existing model-dependent queries and functionality behave? If we do have a stub model, is it at all visible to the user, or completely hidden?
  * **Decision:** We will automatically create a mostly-hidden non-executable model row in the DB when the user uploads a non-executable plan. Too many things rely on some model being present to truly have *no model* row for these plans.
* **Resource datasets:** Should datasets imported with a model-free plan be considered `simulation_dataset`s (associated with a model) or just plain `dataset`s (like an external profile)?
  * **Decision:** they will be `simulation_dataset`s, as this requires the least work and is still maps reasonably well to what the dataset represents 
* **Read-only scope:** Which operations remain available on a model-free plan - for example, changing plan metadata, managing external datasets, configuring views, and associating constraints/goals/external events with the plan? How is immutability enforced through the API?
  * **Decision:** see [matrix in this slack convo](https://jpl.enterprise.slack.com/docs/T024LMMEZ/F0C28RKNV3P) (team only). 
  * Generally we are disallowing all changes to activities/parameters & simulation. Ideally would like to **allow** plan metadata changes (eg. name/description/tags) but only if feasible

**Notes**

* **Exporting** these files from existing plans is not *yet* a requirement, but will be in the future & we should keep it in mind.
* Existing plans can survive deletion of their mission model and still be viewable. This should be supported as well as possible, and will likely be the same behavior as an imported model-free plan *without* activity definitions

**Success Condition**

Import a representative file into an instance with no mission model, inspect its activities and resources, expand a known template, and verify the expected output. Reload the plan to confirm persistence, and verify that unsupported mutations, simulation, and scheduling are blocked.

Mythicaeda:
- **Tickets addressed:** Closes #1885
- **Review:** By commit 
- **Merge strategy:** Merge (no squash) 

## Description

Feature branch for Non-Executable Plans


dandelany:
This is from the feature branch into `develop` - I thought the intent was to merge into the feature branch so we could review changes independently before going to develop. Can you reopen this as a PR from `some-other-branch` into `feat/non-executable-plans`?

Mythicaeda:
I'm not quite sure what you mean. This _is_ the feature branch that the backend changes will be merged into (ie, my next set of db changes will be into this branch here). It's open as a draft PR for visibility and so that automated tests run as we merge into it. 

Unless you're confused because the changes match those in `feat/non-executable-model-db`? If that's the case, the reason the changes are the same are because I merged that branch into this one (`feat/non-executable-plans`) yesterday, as that PR was approved and passing tests (though I did have to push a commit to get the tests working).

I've also added the `DON'T MERGE` tag to the PR for clarity.

